PR #22, opened 2024-10-23
Add equalsToken to ShorthandPropertyAssignment

sandersn:
To make it match the tsc AST.

### DIFF
--- a/internal/compiler/parser.go
+++ b/internal/compiler/parser.go
@@ -5142,13 +5142,15 @@ func (p *Parser) parseObjectLiteralElement() *Node {
 	//     IdentifierReference[?Yield] Initializer[In, ?Yield]
 	// this is necessary because ObjectLiteral productions are also used to cover grammar for ObjectAssignmentPattern
 	var node *Node
+	var equalsToken *Node
 	isShorthandPropertyAssignment := tokenIsIdentifier && p.token != SyntaxKindColonToken
 	if isShorthandPropertyAssignment {
 		var initializer *Expression
-		if p.parseOptional(SyntaxKindEqualsToken) {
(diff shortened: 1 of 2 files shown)


internal/compiler/parser.go
```diff
@@ -5142,13 +5142,15 @@ func (p *Parser) parseObjectLiteralElement() *Node {
 	//     IdentifierReference[?Yield] Initializer[In, ?Yield]
 	// this is necessary because ObjectLiteral productions are also used to cover grammar for ObjectAssignmentPattern
 	var node *Node
+	var equalsToken *Node
 	isShorthandPropertyAssignment := tokenIsIdentifier && p.token != SyntaxKindColonToken
 	if isShorthandPropertyAssignment {
 		var initializer *Expression
-		if p.parseOptional(SyntaxKindEqualsToken) {
+		equalsToken = p.parseOptionalToken(SyntaxKindEqualsToken)
```
jakebailey, line 5149: Just `equalsToken := ...` here? The value isn't used below.

sandersn: this was copied from day 2 of my using Go =)

sandersn:
@ahejlsberg says this is an intentional difference so I'll normalise it in the baseliner instead.

jakebailey:
Intentional or not... we're going to have to end up with some scheme to walk tokens in the future, so hopefully your script is able to handle that in the future too.

sandersn:
I'd like for it happen via an API change so that a normal visitor can access tokens, instead of some internal hackery.

Emit will test the whole thing comprehensively anyway -- especially since we no longer care about exact AST equivalence.

jakebailey:
Or, we could have two visit functions generated, one that manifests tokens up, one that doesn't.

PR #23, opened 2024-10-23
Use stringer generation for ModuleResolutionKind

andrewbranch:


internal/compiler/types.go
```diff
@@ -1,6 +1,6 @@
 package compiler
 
-//go:generate go run golang.org/x/tools/cmd/stringer -type=SyntaxKind,SignatureKind,ScriptKind,ScriptTarget,LanguageVariant,Tristate -output=stringer_generated.go
+//go:generate go run golang.org/x/tools/cmd/stringer -type=LanguageVariant,ModuleResolutionKind,ScriptKind,ScriptTarget,SignatureKind,SyntaxKind,Tristate -output=stringer_generated.go
```
andrewbranch, line 3: Big diff because I alphabetized these

jakebailey: Yeah, I just had them in source order 😄

PR #21, opened 2024-10-23
Port single-file baseline writer

andrewbranch:
This adds `baselines.RunBaseline` which compares content against baselines in `testdata/baselines/reference`, and writes changes to `testdata/baselines/local`. Baselines can be accepted with `hereby baseline-accept`, copied from the TypeScript repo.

PR #24, opened 2024-10-23
Add parser AST tests

sandersn:
Super simple parser baselines. I use these to find differences between tsc and tsgo. I'm also going to use them to debug JSDoc parsing.
I don't think we'll want to keep these around, so I haven't learned how to turn them into a test case. If we *do* want to keep them around, I'll want to use Andrew's baseline code.

Edit: From the latest commit:

Go now has two tests:
  1. Normal local/reference baselines.
  2. Local/tsc baselines.

I stored the tsc baselines in a separate location, although I don't know
if that's the right choice. Ideally I want to be able to compare against
tsc output, then switch over to comparing against the previous commit's
output.

However, I don't think either reference/*.ast or gold/*.ast should be
checked in, since our current tsc test suite covers parsing quite well.
This is just needed until that test suite is up and running.

Notes:
- Dumps to tests/baselines/local (tsgo) and tests/baselines/gold (tsc). Directories are flattened.
- Format is an indented tree of syntax kinds. Identifiers also dump their FullText. I may add span indices if I need them.
- A couple of tests -- binary expression stress tests -- are skipped in tsc because they blow the stack of the basic visitor code.
- Normalises a number of differences between the ASTs so far:

1. NodeArray becomes explicit *List types; I skip these in the tsgo output.
2. OmittedExpression is represented as a BindingElement with all-nil properties in tsgo.
3. Very clunky code to avoid printing aliases from the tsc SyntaxKind enum.


andrewbranch:
Related: #21

internal/compiler/parser_test.go
```diff
@@ -18,3 +20,72 @@ func BenchmarkParse(b *testing.B) {
 		ParseSourceFile(fileName, sourceText, ScriptTargetESNext)
 	}
 }
+
+func TestParseAndPrintNodes(t *testing.T) {
+	root := "../../_submodules/TypeScript"
+	local := "../../tests/baselines/local"
```
jakebailey: We'll definitely want to use the approach from Andrew's PR for storing these; the paths here won't work as you expect once generalized, and we do not want to have a top level `tests` dir with a bunch of contents.

internal/harness/baselineAST.js
```diff
@@ -0,0 +1,155 @@
+// @ts-check
```
jakebailey: Is the point of this script to run and then compare? I assume the eventual goal is to do `os/exec.Command` or something to run this...

internal/compiler/parser_test.go
sandersn: Switched to baseline.Run and repo.TypeScriptSubmodulePath

internal/compiler/parser_test.go
```diff
@@ -2,7 +2,11 @@ package compiler
 
 import (
 	"fmt"
+	"github.com/microsoft/typescript-go/internal/repo"
+	"github.com/microsoft/typescript-go/internal/testutil/baseline"
```
jakebailey: These imports are missorted; they should be appearing below the stdlib.

gofmt won't do this by itself; I need to update CI and dprint to be stricter about this, but just hitting save in your editor I think will "fix" this.

sandersn: Weird, I have autosave turned on. Does a manual save command do something extra?

edit: yes. This is a problem because I never press Ctrl-S anymore.

jakebailey: Yes; but you can configure that behavior in `settings.json`. I set it to only work on explicit saves as a conservative choice (I personally hit Ctrl+S when I know I want the editor to do something).

I'm surprised auto import didn't do this, but maybe it doesn't. I'll have to figure that out.

internal/compiler/parser_test.go
```diff
@@ -18,3 +23,66 @@ func BenchmarkParse(b *testing.B) {
...
 	}
 }
+
+func TestParseAndPrintNodes(t *testing.T) {
+	program := NewProgram(ProgramOptions{
+		RootPath: repo.TypeScriptSubmodulePath,
+		Options:  &CompilerOptions{Target: ScriptTargetESNext, ModuleKind: ModuleKindNodeNext}})
+	for _, sourceFile := range program.SourceFiles() {
+		path := filepath.ToSlash(sourceFile.fileName)
+		path = path[len(repo.TypeScriptSubmodulePath)+1:]
```
jakebailey: This is dubious; the variable is a filepath encoded string, while path is a slash encoded string.

internal/harness/baselineAST.js
```diff
@@ -0,0 +1,159 @@
...
```
jakebailey: This should go in `compiler/testdata` instead of creating a ghost package

internal/compiler/parser_test.go
```diff
@@ -18,3 +23,66 @@ func BenchmarkParse(b *testing.B) {
 		ParseSourceFile(fileName, sourceText, ScriptTargetESNext)
 	}
 }
+
+func TestParseAndPrintNodes(t *testing.T) {
+	program := NewProgram(ProgramOptions{
+		RootPath: repo.TypeScriptSubmodulePath,
+		Options:  &CompilerOptions{Target: ScriptTargetESNext, ModuleKind: ModuleKindNodeNext}})
+	for _, sourceFile := range program.SourceFiles() {
```
jakebailey: Rather than looping the test, construct the test like this:

```go
func TestParseAndPrintNodes(t *testing.T) {
	t.Parallel()

	for _, sourceFile := range files {
		b.Run(sourceFile.Name(), func(t *testing.T) {
			t.Parallel()
			// ...
		})
	}
```

You probably should just do the FS walk in this test instead and parse each file separately; it's only temporary that `NewProgram` adds all files.

jakebailey: Doing it separately here will also mean that parsing is done in parallel per test; _technically_ `NewProgram` does it in parallel too, but will likely be less parallel as we start implementing it as finding deps and so on, whereas these tests are totally parallelizable.

internal/harness/baselineAST.js
sandersn: So tests like testdata are nested inside the compiler directory in Go?

jakebailey: Normally, you'd stick testdata in the package you're working on, but since we have packages which will have to share test files, having them at the root isn't a big deal. But for one-off artifacts like this... I would probably just stick them within the packages's testdata. Or, we can make a `scripts` dir at the repo root?

jakebailey: (The scripts dir also makes one more thing for Go to scan, but it's sort of whatever)

jakebailey: (But in general I would say that within `internal`, `cmd`, etc, we should avoid making dirs that _aren't_ packages, except for `testdata`, `_*`, `.*`, which won't be packages either.)

internal/compiler/parser_test.go
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+	path, err := filepath.Rel(repo.TypeScriptSubmodulePath, fileName)
+	if err != nil {
+		t.Errorf("%s is outside of the TypeScript submodule", fileName)
+	}
+	return strings.ReplaceAll(path, string(filepath.Separator), "_") + ".ast"
+}
+
+// prefix specifies the directory to write the baseline
+func printAST(sourceFile *SourceFile) string {
+	sb := strings.Builder{}
```
jakebailey:
```suggestion
	var sb strings.Builder
```

This is a nit, but I pretty much always suggest writing values like this if you're not initializing them; my thinking is that this form doesn't require me to remember anything about the type I'm trying to make, as in I don't need to remember that I can't write `s := SyntaxKind{}` because it's not a struct, but I do have to write `strings.Builder{}` because it is. Just declaring the variable without using `:=` gets you there no matter what type you're writing.

internal/compiler/parser_test.go
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+	var visit func(node *Node, indentation int) bool
+	visit = func(node *Node, indentation int) bool {
+		offset := 1
+		skind := node.kind.String()[len("SyntaxKind"):]
```
jakebailey:
```suggestion
		skind, _ := strings.CutPrefix(node.kind.String(), "SyntaxKind")
```

Feels a little safer. Hoping to fork and pull in `stringer` so we can do our own thing with it later, though.

internal/compiler/parser_test.go
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+		switch node.kind {
+		case SyntaxKindModifierList, SyntaxKindTypeParameterList, SyntaxKindTypeArgumentList, SyntaxKindSyntaxList:
+			offset = 0
+		case SyntaxKindIdentifier:
+			indent := strings.Repeat("  ", indentation)
```
jakebailey: From the perf perspective, all of these repeats will end up being inefficient because this will create fresh string values over and over for each element.

internal/compiler/parser_test.go
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+		t.Fatalf("Error walking the path %q: %v", repo.TypeScriptSubmodulePath, err)
+	}
+}
+
+func TestParseAgainstTSC(t *testing.T) {
+	t.Parallel()
+	goldDir := "../../testdata/baselines/gold"
+	entries, err := os.ReadDir(goldDir)
+	if err != nil || len(entries) == 0 {
+		cmd := exec.Command("node", "testdata/baselineAST.js", "../../_submodules/TypeScript/", goldDir)
```
jakebailey: I think this isn't really the workflow I was imagining. I was more thinking of having no files at all, and instead have Go iterate all files, executing the dumper script for each and capturing its output via stdout and failing the test with a diff if it mismatches.

Is your thinking that you want the files on disk for some reason? Or is that other method slow?

internal/compiler/parser_test.go
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+			return nil
+		}
+		ext := filepath.Ext(fileName)
+		if (ext == ".ts" || ext == ".js" || ext == ".tsx" || ext == ".jsx") &&
+			!(strings.HasSuffix(fileName, "binderBinaryExpressionStress.ts") ||
+				strings.HasSuffix(fileName, "binderBinaryExpressionStress.js") ||
+				strings.HasSuffix(fileName, "binderBinaryExpressionStressJs.ts") ||
+				strings.HasSuffix(fileName, "binderBinaryExpressionStressJs.js")) {
+			testName, _ := filepath.Rel(repo.TypeScriptSubmodulePath, fileName)
+			t.Run(testName, func(t *testing.T) {
```
jakebailey: Just from a code perspective, I think this would be more cleanly written:

```go
testName, _ := filepath.Rel(repo.TypeScriptSubmodulePath, fileName)
t.Run(testName, func(t *testing.T) {
	t.Parallel()

	if isIgnoredTest(testName) {
		t.Skip("some reason")
	}
})

File: internal/compiler/parser_test.go [hunk 12]
```diff
@@ -18,3 +28,121 @@ func BenchmarkParse(b *testing.B) {
...
+				t.Parallel()
+				sourceText, err := os.ReadFile(fileName)
+				if err != nil {
+					t.Errorf("Failed to read file %s: %v", fileName, err)
+					return
+				}
```
jakebailey: You'll probably just want to write:

```
assert.NilError(t, err)
```

Which will handle everything for you.

internal/compiler/parser_test.go
sandersn: Yeah, I want to use Beyond Compare to diff. It's more usable than console output for a few failures, and there's no comparison at all for many failures -- currently there are around 400.

sandersn: Oh, also the dumper script is slow, although in parallel it would be much faster.

internal/compiler/parser_test.go
```diff
@@ -18,3 +30,140 @@ func BenchmarkParse(b *testing.B) {
...
+	path, err := filepath.Rel(repo.TypeScriptSubmodulePath, fileName)
+	if err != nil {
+		t.Errorf("%s is outside of the TypeScript submodule", fileName)
+	}
+	return strings.ReplaceAll(path, string(filepath.Separator), "_") + ".ast"
+}
+
+var (
+	indentationCache map[int]string = make(map[int]string)
+	mutex            sync.Mutex
```
jakebailey: call this `indentationCacheMu`; this is in the global scope of the package so that's a non-specific name.

internal/compiler/parser_test.go
jakebailey: I think that if we want to have a mode which dumps out the files, we should do that behind a flag; it's possible to create test-only flags with the `flag` package in a `_test` file.

If we think this test is going to slow to run from scratch, we should consider using `testing.Short()` and them modifying our test scripts to pass in `-short` or something when needed. Or only enable these tests in general behind a flag / CI check.

But the way I would think about this is that someone should be able to clone the repo and then just do `go test ./...` and it passes, no extra files created, etc.

sandersn:
A couple of other ideas:

1. The toplevel test could run the baseline JS code in the mode that writes to gold/ before running the rest.
2. The parallel tests could check gold/ for an up-to-date baseline and skip running the baseline JS if it's there already.

sandersn:
Third idea: if this doesn't get used for more than a month or two, should it just be a script that people can run locally?
But if it's not a test in the compiler_test namespace, will it need to wait until we make a public API?

PR #13, opened 2024-10-22
Generate String methods for enums with stringer

jakebailey:
Using the oft used: https://cs.opensource.google/go/x/tools/+/refs/tags/v0.26.0:cmd/stringer/stringer.go

This has some rough edges:

- Bitset enums are not handled
	- This seems fixable upstream with another flag.
- `-trimprefix` can't be a list, so we can't generate a single file for all enums and also use that flag.
	- This seems fixable upstream; commas can't have been a legal `trimprefix` before.
- The "first" value of duplicate values always wins.
	- This seems desirable?
- Many calls to `go run` in `go generate` is slow because there's no binary caching, which will be fixed with `go tool stringer` hopefully in an upcoming version

PR #25, opened 2024-10-23
Port of `--pretty` diagnostics reporting

DanielRosenwasser:
Prints errors with code blocks, line numbers, and red squiggles when `--pretty` is provided.

Currently this doesn't faithfully port over non-pretty error reporting (and just uses what's already there), and doesn't toggle the default based on whether we're in a TTY or not.

One intentional deviation here is that related spans are not indented. Stylistically, this looked a bit awkward (maybe because the indent was not consistent between the code block and the message itself), so I removed it.

### DIFF
--- a/internal/compiler/path.go
+++ b/internal/compiler/path.go
@@ -224,7 +224,7 @@ func getPathFromPathComponents(pathComponents []string) string {
 	if len(pathComponents) == 0 {
 		return ""
 	}
-	
+
 	root := pathComponents[0]
 	if root != "" {
 		root = ensureTrailingDirectorySeparator(root)


internal/compiler/utilities.go
```diff
@@ -164,6 +167,20 @@ func addRelatedInfo(diagnostic *Diagnostic, relatedInformation ...*Diagnostic) *
 	return diagnostic
 }
 
+func DiagnosticCategoryName(category diagnostics.Category) string {
```
jakebailey: This could be declared in the diagnostics package or as a method.

internal/compiler/program.go
```diff
@@ -389,3 +393,192 @@ func (p *Program) collectModuleReferences(file *SourceFile, node *Statement, inA
...
+	firstLine, firstChar := GetLineAndCharacterOfPosition(file, pos)
+	var relativeFileName string
+	if formatOpts != nil {
+		relativeFileName = ConvertToRelativePath(file.path, formatOpts.CurrentDirectory, formatOpts.GetCanonicalFileName)
+	} else {
+		relativeFileName = file.path
+	}
+
+	writeWithStyleAndReset(output, relativeFileName, foregroundColorEscapeCyan)
+	output.WriteRune(':')
```
jakebailey: Micro optimization, but `WriteByte` directly writes the byte to the output. Not sure it practically matters, though, because WriteRune does do that too if the rune is small enough and it may all be inlinable anyway and all disappear.

internal/compiler/diagnostics/generate.go
```diff
@@ -43,6 +43,20 @@ const (
 	CategoryMessage
 )
 
+func (category Category) Name() string {
```
jakebailey: This is now in a proper file, so you don't need to use a generator for it.

internal/compiler/diagnostics/diagnostics.go
```diff
@@ -13,6 +13,20 @@ const (
 	CategoryMessage
 )
 
+func (category Category) Name() string {
```
gabritto, line 16: Jake added some codegen that generates a similar function: https://github.com/microsoft/typescript-go/pull/13/files#diff-76ea5b324faf8f11be7f729765c6a3870ac109f5605a7602a0e826431ac25d66R21

However, I now noticed that the stringer generated function generates something like "CategoryWarning" instead of "warning", so maybe we need to fix that.

internal/compiler/program.go
```diff
@@ -389,3 +393,192 @@ func (p *Program) collectModuleReferences(file *SourceFile, node *Statement, inA
...
+		} else {
+			// Squiggle the entire line.
+			writer.WriteString(strings.Repeat("~", len(lineContent)))
+		}
+
+		writer.WriteString(resetEscapeSequence)
+	}
+}
+
+func WriteFlattenedDiagnosticMessage(writer *strings.Builder, diagnostic *Diagnostic, newline string, level int) {
```
gabritto: I think `level` will never be anything other than `0` for callers.

internal/compiler/utilities.go
```diff
@@ -3135,6 +3138,15 @@ func extensionIsTs(ext string) bool {
 	return ext == ExtensionTs || ext == ExtensionTsx || ext == ExtensionDts || ext == ExtensionMts || ext == ExtensionDmts || ext == ExtensionCts || ext == ExtensionDcts || len(ext) >= 7 && ext[:3] == ".d." && ext[len(ext)-3:] == ".ts"
 }
 
+func ConvertToRelativePath(absoluteOrRelativePath, basePath string, getCanonicalFileName func(fileName string) string) string {
```
gabritto: I think this doesn't matter too much, but the original function seems to be in `path.ts` in the old codebase instead of in `utilities`.

internal/compiler/program.go
```diff
@@ -389,3 +393,192 @@ func (p *Program) collectModuleReferences(file *SourceFile, node *Statement, inA
 		}
 	}
 }
+
+type DiagnosticsFormattingOptions struct {
+	CurrentDirectory     string
```
gabritto: This was a function in the original, so I assumed it could change throughout the life of the formatting host. Is that not true?

DanielRosenwasser: Technically, but error reporting is one-short, so I don't expect us to rely on changes. Even if it does, someone can always mutate `CurrentDirectory` under the hood if we need.

gabritto: This is the usage that looked suspicious to me: https://github.dev/microsoft/TypeScript/blob/2ac4cb78d6930302eb0a55d07f154a2b0597ae32/src/compiler/watch.ts#L126
It wasn't clear to me that, using watch mode (which calls the above function to create a diagnostics reporter), the current directory would not change.

internal/compiler/utilities.go
```diff
@@ -38,7 +38,7 @@ func (t TextRange) End() int {
 	return int(t.end)
 }
 
-func (t TextRange) Len() int {
+func (t TextRange) Length() int {
```
jakebailey: FWIW `Len()` is more idiomatic in Go than `Length()` (Not sure why this is changing in this PR?)

internal/compiler/program.go
DanielRosenwasser: For the purposes of error reporting, I would really hope the system is not changing the CWD midway through. I would consider that behavior extremely spooky.

cmd/tsgo/main.go
```diff
@@ -57,12 +65,71 @@ func main() {
...
+	runes := make([]rune, len(str))
+	for i, c := range str {
+		upper := unicode.ToUpper(c)
+		if upper == c {
+			runes[i] = unicode.ToLower(c)
+		} else {
+			runes[i] = upper
+		}
+	}
+	return string(runes)
```
jakebailey: FWIW this could just be a `strings.Map`

cmd/tsgo/main.go
```diff
@@ -57,12 +65,69 @@ func main() {
...
+		program.PrintTypeAliases()
+	}
+
 	fmt.Printf("Files:         %v\n", len(program.SourceFiles()))
 	fmt.Printf("Types:         %v\n", program.TypeCount())
 	fmt.Printf("Compile time:  %v\n", compileTime)
 	fmt.Printf("Memory used:   %vK\n", memStats.Alloc/1024)
 }
+
+func isFileSystemCaseSensitive() bool {
```
jakebailey, line 106: At some point, we should just make a package called `platform` or `system` or something that has like:

```go
package platform

import "sync"

var IsFileSystemCaseSensitive = sync.OnceValue(func() bool {
	if runtime.GOOS == "windows" {
		return false
	}

	// If the current executable exists under a different case, we must be case-insensitve.
	if _, err := os.Stat(swapCase(os.Args[0])); os.IsNotExist(err) {
		return false
	}
	return true
})

func swapCase(str string) string {
	return strings.Map(func(r rune) rune {
		upper := unicode.ToUpper(r)
		if upper == r {
			return unicode.ToLower(r)
		} else {
			return upper
		}
	}, str)
}
```

Though, maybe not... given accessing `os.Args` outside of the `main` package is just very dubious and it's not like there's a rando file we can assume exists. Though, I guess we could just use `os.Executable()`.

cmd/tsgo/main.go
```diff
@@ -57,12 +65,69 @@ func main() {
...
+			} else {
+				getCanonicalFileName = strings.ToLower
+			}
+
+			currentDirectory, err := os.Getwd()
+			if err != nil {
+				currentDirectory = ""
+			}
+
+			output := strings.Builder{}
```
jakebailey: Style nit; I prefer writing:

```go
var output strings.Builder
```

Rather than the `{}` syntax, as `{}` only works for struct values and so you would have to remember which types are structs and which aren't. `var output T` works for any type.

cmd/tsgo/main.go
```diff
@@ -57,12 +65,69 @@ func main() {
...
+				getCanonicalFileName = func(path string) string {
+					return path
+				}
+			} else {
+				getCanonicalFileName = strings.ToLower
+			}
+
+			currentDirectory, err := os.Getwd()
+			if err != nil {
+				currentDirectory = ""
```
jakebailey: This seems cursed. Are you sure we shouldn't kill the program if this fails?

internal/compiler/path.go
```diff
@@ -16,7 +16,7 @@ func hasTrailingDirectorySeparator(path string) bool {
 }
 
 func combinePaths(paths ...string) string {
-	return path.Join(paths...)
+	return filepath.Join(paths...)
```
jakebailey: I don't think this is correct; my understanding is that all paths in this file are assumed to be `/` separated (which is what we do internally in TS). May need to check with @andrewbranch.

internal/compiler/scanner.go
```diff
@@ -1750,3 +1750,36 @@ func getEndLinePosition(sourceFile *SourceFile, line int) int {
...
+}
+
+func ComputePositionOfLineAndCharacter(lineStarts []TextPos, line int, character int) TextPos {
+	/// !!! debugText, allowEdits
+	if line < 0 || line >= len(lineStarts) {
+		// if (allowEdits) {
+		//     // Clamp line to nearest allowable value
+		//     line = line < 0 ? 0 : line >= lineStarts.length ? lineStarts.length - 1 : line;
+		// }
+		panic(fmt.Sprintf("Bad line number. Line: %d, lineStarts.length: %d.", line, len(lineStarts)))
```
jakebailey, line 1765: _Technically_ it's convention for go errors/panics to start with a lowercase number, but I don't see me being able to convince the team to do that just yet 😄

internal/compiler/path.go
```diff
@@ -252,3 +252,13 @@ func ensureTrailingDirectorySeparator(path string) string {
 func (p Path) ensureTrailingDirectorySeparator() Path {
 	return Path(ensureTrailingDirectorySeparator(string(p)))
 }
+
+func ConvertToRelativePath(absoluteOrRelativePath, basePath string, getCanonicalFileName func(fileName string) string) string {
+	/// !!!
+	/// May have some differences with original,
+	/// don't necessarily want dependencies on 'filepath'.
+	if resolvedRelativePath, err := filepath.Rel(basePath, getCanonicalFileName(absoluteOrRelativePath)); err == nil {
```
jakebailey: Here too; I am not sure if using filepath is correct in this file.

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,337 @@
...
+		} else {
+			lineEnd = sourceFile.loc.end
+		}
+
+		lineContent := strings.TrimRightFunc(sourceFile.text[lineStart:lineEnd], unicode.IsSpace) // trim from end
+		lineContent = strings.ReplaceAll(lineContent, "\t", " ")                                  // convert tabs to single spaces
+
+		// Output the gutter and the actual contents of the line.
+		writer.WriteString(gutterStyleSequence)
+		fmt.Fprintf(writer, "%*d", gutterWidth, i+1)
```
jakebailey, line 116: Using a feature of `fmt` I've never used 😄

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,337 @@
...
+
+		lineStart := GetPositionOfLineAndCharacter(sourceFile, i, 0)
+		var lineEnd TextPos
+		if i < lastLineOfFile {
+			lineEnd = GetPositionOfLineAndCharacter(sourceFile, i+1, 0)
+		} else {
+			lineEnd = sourceFile.loc.end
+		}
+
+		lineContent := strings.TrimRightFunc(sourceFile.text[lineStart:lineEnd], unicode.IsSpace) // trim from end
```
jakebailey, line 111: Annoying that `TrimSpace` exists but not `TrimRightSpace` and so on... TrimSpace is optimized for ASCII whereas the func variants are not.

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,337 @@
...
+		// Output the gutter and the error span for the line using tildes.
+		writer.WriteString(gutterStyleSequence)
+		fmt.Fprintf(writer, "%*s", gutterWidth, "")
+		writer.WriteString(resetEscapeSequence)
+		writer.WriteString(gutterSeparator)
+		writer.WriteString(squiggleColor)
+		if i == firstLine {
+			// If we're on the last line, then limit it to the last character of the last line.
+			// Otherwise, we'll just squiggle the rest of the line, giving 'slice' no end position.
+			lastCharForLine := ifElse(i == lastLine, lastLineChar, len(lineContent))
```
jakebailey: This doesn't seem super needed compared to just writing the if, honestly

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,337 @@
...
+
+			// Fill with spaces until the first character,
+			// then squiggle the remainder of the line.
+			writer.WriteString(strings.Repeat(" ", firstLineChar))
```
jakebailey, line 140: Any of these `writer.WriteString(strings.Repeat(...))` calls will be wasteful as they'll produce the intermediate string; you'd be better off writing a `func writeRepeated(b *strings.Builder, r rune, count int)` helper or something.

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,337 @@
...
+	writer.WriteString(diagnostic.Message())
+
+	for _, chain := range diagnostic.messageChain {
+		flattenDiagnosticMessageChain(writer, chain, newline, 1 /*level*/)
+	}
+}
+
+func flattenDiagnosticMessageChain(writer *strings.Builder, chain *MessageChain, newLine string, level int) {
+	writer.WriteString(newLine)
+	for i := 0; i < level; i++ {
```
jakebailey: Nit, but `for i := range level`. A lint pass will eventually show these cases.

cmd/tsgo/main.go
DanielRosenwasser: This is how it works in `sys` right now, and I decided to kick that a little down the road.

cmd/tsgo/main.go
DanielRosenwasser: Probably! I think the way this is modeled through the hosts is that you can end up with a `""` directory.

internal/compiler/scanner.go
DanielRosenwasser: I couldn't find any lowercase numbers to start the message with. 😞

internal/compiler/path.go
DanielRosenwasser: Unfortunately there is no `path.Rel`. I didn't want to scaffold all of that just yet.

internal/compiler/path.go
andrewbranch: Yes, this was intentional—TS uses forward slashes across all systems, so it’s important that we never use `filepath` outside of test infrastructure.

DanielRosenwasser: Okay.

Ughhhhhhhhhhhhhhhhhhhhhhh

jakebailey: Going forward I think things are going to be super annoying unless we really ensure to never use the wrong kind of path, and assert. Especially if we use `fs.FS` as our FS abstraction, where it _doesn't_ take rooted paths at all so relative paths are just relative to the root, so that will just "break" if we don't also make all paths absolute...

cmd/tsgo/main.go
```diff
@@ -57,12 +65,69 @@ func main() {
...
+		} else {
+			for _, diagnostic := range diagnostics {
+				printDiagnostic(diagnostic, 0)
+			}
 		}
 	}
+
+	if printTypes {
+		program.PrintTypeAliases()
+	}
```
jakebailey: Where is this call coming from? Isn't this handled above?

internal/compiler/path.go
```diff
@@ -143,21 +220,22 @@ func (p Path) getDirectoryPath() Path {
...
-
 func getPathFromPathComponents(pathComponents []string) string {
 	if len(pathComponents) == 0 {
 		return ""
 	}
 
+	// !!!
+	// TODO: This code and the original might have a
+	// fast path to just join the components if the
+	// first component doesn't have a directory separator.
```
DanielRosenwasser: https://github.com/microsoft/TypeScript/pull/60371

internal/compiler/diagnostics/diagnostics.go
andrewbranch: There’s an option to strip a prefix; it just needs to be added to the comment.

jakebailey: The trimprefix flag doesn't work when you use the other flag to generate multiple types. We'd need to fork `stringer` to make that work, which wouldn't be all that bad.

andrewbranch:
Are we close to merging this? If not, can we split out the path/core changes?

cmd/tsgo/main.go
jakebailey: This was marked resolved but I still see two calls in this function. Am I missing something?

internal/compiler/path.go
DanielRosenwasser: Sorry, I forgot to push.

DanielRosenwasser:
I'd just like one more approval before I merge.

internal/compiler/error_reporting.go
```diff
@@ -0,0 +1,342 @@
...
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+)
+
+type DiagnosticsFormattingOptions struct {
+	CurrentDirectory     string
+	NewLine              string
```
jakebailey, line 16: I really do hope we just say no to CRLF in outputs... And drop all of this code.

PR #27, opened 2024-10-24
Fix case of BigInt, previously Bigint

sandersn:


jakebailey:
The prototype for `bigint` is `BigInt`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

(My preference is `BigInt`, personally)

sandersn:
Based on discussion in the design meeting, I'm going to merge this now that it's up to date and passing CI.

PR #28, opened 2024-10-24
Add AST baselines

sandersn:
This is the same as https://github.com/microsoft/typescript-go/pull/24, but without the TS AST dumper that I don't want to turn into a test case.

1. Still needs baselines. I'll probably put those in at the end so as not to break github UI.
2. I'm still not sure we'll want to keep this format forever, but I do need it for adding JSDoc (and probably fixing other bugs in the parser).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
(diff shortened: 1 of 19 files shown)


jakebailey:
I think the direction we want to take here is to _not_ check in the baselines, but just error using assert/cmp or something to print out an error.

Additionally, the PR should run `node` in a parallel test over every test file and check them. If that takes too long (I suspect it will), we can make it conditional on `testing.Short()`, or just have them only run when an env var is set.

(Also the test script should itself be in a `testdata` dir, probably within the compiler package)

sandersn:
OK, let's use #24 instead of this PR then.

PR #29, opened 2024-10-24
Add insertion ordered map/set

jakebailey:
This adds a fully tested implementation of an insertion ordered map, deduped by key, just like JS's `Map`.

I thought I could swap out `orderedMap` in `utilities.go`, but it's actually not what I thought it was; that type doesn't store off the order of keys, only the order of values and whether or not it's been added or not.

Probably that type deserves to be written out in another file, which I may just add to this PR separately.

TODO:

- JSON implementation based on https://pkg.go.dev/github.com/go-json-experiment/json#example-package-OrderedObject
- Benchmarks

### DIFF
--- a/internal/collections/map.go
+++ b/internal/collections/map.go
@@ -123,7 +123,7 @@ func (m *Map[K, V]) Size() int {
 	return len(m.keys)
 }
 
-// Clone returns a copy of the map. Entries are shallow-copied.
+// Clone returns a shallow copy of the map.
 func (m *Map[K, V]) Clone() *Map[K, V] {
 	m2 := m.clone()
 	return &m2
--- a/internal/collections/set.go
+++ b/internal/collections/set.go
@@ -2,43 +2,53 @@ package collections
 
 import "iter"
 
+// Set an insertion ordered set.
 type Set[T comparable] struct {
 	m Map[T, struct{}]
 }
 
+// NewSetWithSizeHint creates a new Set with a hint for the number of elements it will contain.
 func NewSetWithSizeHint[T comparable](hint int) *Set[T] {
 	return &Set[T]{
 		m: newMapWithSizeHint[T, struct{}](hint),
 	}
 }
 
+// Add adds a value to the set.
 func (s *Set[T]) Add(value T) {
 	s.m.Set(value, struct{}{})
 }
 
+// Has returns true if the set contains the value.
 func (s *Set[T]) Has(value T) bool {
 	return s.m.Has(value)
 }
 
+// Delete removes a value from the set.
 func (s *Set[T]) Delete(value T) bool {
 	_, ok := s.m.Delete(value)
 	return ok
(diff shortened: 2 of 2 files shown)


internal/collections/insertion_ordered_map.go
```diff
@@ -0,0 +1,127 @@
...
+	} else {
+		m.keys = slices.Delete(m.keys, i, i+1)
+	}
+
+	return v, true
+}
+
+// Keys returns an iterator over the keys in the map.
+// A slice of the keys can be obtained by calling `slices.Collect`.
+func (m *InsertionOrderedMap[K, V]) Keys() iter.Seq[K] {
```
jakebailey: I'm doing the "new thing" and using iterators here. I could return slices or similar for keys and just say "do not modify", but that may not be a good idea and technically limits the implementation.

internal/collections/insertion_ordered_map.go
```diff
@@ -0,0 +1,127 @@
+package collections
+
+import (
+	"iter"
+	"maps"
+	"slices"
+)
+
+// InsertionOrderedMap is an insertion ordered map.
+type InsertionOrderedMap[K comparable, V any] struct {
```
jakebailey: I can also add `InsertionOrderedSet`... which would basically just be `InsertionOrderedSet[T, struct{}]`

jakebailey: Also, this name is long, I'm not sure it's worth calling it anything other than `Map`, though.

internal/collections/insertion_ordered_map_test.go
```diff
@@ -0,0 +1,134 @@
...
+	assert.Assert(t, !clone.Has(start))
+	assert.Equal(t, len(slices.Collect(clone.Keys())), 0)
+	assert.Equal(t, len(slices.Collect(clone.Values())), 0)
+}
+
+func padInt(n int) string {
+	return fmt.Sprintf("%10d", n)
+}
+
+func TestMapWithSizeHint(t *testing.T) {
```
andrewbranch: Is there a convention or rule of thumb for when to break into multiple Test functions vs. combine multiple assertions in one function?

jakebailey: The test I've written above is not particularly elegant. I should probably split it up, it's just something I had written a bit ago to hack on things (I was actually testing different implementations of this to see how fast we could go).

internal/collections/insertion_ordered_map.go
jakebailey: I went back to Map. 😄

jakebailey:
Alternatively, we use https://pkg.go.dev/github.com/wk8/go-ordered-map/v2 for this. It has a different implementation under the hood (a linked list) with a different API (which I've written a wrapper for already since I was perf testing a few weeks ago)

PR #32, opened 2024-10-26
Type instantiation and indexed access types

ahejlsberg:
Ports logic for indexed access type resolution and core logic for instantiation of type parameters, object types (including array and tuple types), union types, intersection types, and indexed access types. Includes slight reworking of tuple type representation and more consistent object type creation (`newObjectType` now uses `objectFlags` to determine which object type variant to allocate).

Also, new file `mapper.go` that contains type mapper implementations. Mappers use the same struct-with-sidecar layout as nodes and types (although there currently aren't any shared fields in the base struct beyond the `data` interface).

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -852,7 +852,8 @@ func (c *Checker) checkExpressionWorker(node *Node, checkMode CheckMode) *Type {
 		return c.getFreshTypeOfLiteralType(c.getStringLiteralType(node.AsNoSubstitutionTemplateLiteral().text))
 	case SyntaxKindNumericLiteral:
 		// !!! checkGrammarNumericLiteral(node as NumericLiteral)
-		value, _ := strconv.ParseFloat(node.AsNumericLiteral().text, 64)
+		// !!! Revise this to handle NaN, Infinity, etc. in the same manner as JS
+		value := stringToNumber(node.AsNumericLiteral().text)
 		return c.getFreshTypeOfLiteralType(c.getNumberLiteralType(value))
 	case SyntaxKindBigintLiteral:
 		// !!! checkGrammarBigIntLiteral(node as BigIntLiteral);
@@ -7131,7 +7132,7 @@ func (c *Checker) getPropertyTypeForIndexType(originalObjectType *Type, objectTy
 			}
 		}
 		if everyType(objectType, c.isTupleType) && isNumericLiteralName(propName) {
-			index, _ := strconv.ParseFloat(propName, 64)
+			index := stringToNumber(propName)
 			if accessNode != nil && everyType(objectType, func(t *Type) bool {
 				return t.TargetTupleType().combinedFlags&ElementFlagsVariable == 0
 			}) && accessFlags&AccessFlagsAllowMissing == 0 {
@@ -7375,7 +7376,7 @@ func indexTypeLessThan(indexType *Type, limit int) bool {
 		if t.flags&TypeFlagsStringOrNumberLiteral != 0 {
 			propName := getPropertyNameFromType(t)
 			if isNumericLiteralName(propName) {
-				index, _ := strconv.ParseFloat(propName, 64)
+				index := stringToNumber(propName)
 				return index >= 0 && index < float64(limit)
 			}
 		}
--- a/internal/compiler/utilities.go
+++ b/internal/compiler/utilities.go
@@ -3,6 +3,7 @@ package compiler
 import (
 	"fmt"
 	"maps"
+	"math"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -3855,7 +3856,7 @@ func getPropertyNameFromType(t *Type) string {
 	case t.flags&TypeFlagsStringLiteral != 0:
 		return t.AsLiteralType().value.(string)
 	case t.flags&TypeFlagsNumberLiteral != 0:
-		return strconv.FormatFloat(t.AsLiteralType().value.(float64), 'g', -1, 64)
+		return numberToString(t.AsLiteralType().value.(float64))
 	case t.flags&TypeFlagsUniqueESSymbol != 0:
 		return t.AsUniqueESSymbolType().name
 	}
@@ -3884,8 +3885,7 @@ func isNumericLiteralName(name string) bool {
 	// Note that this accepts the values 'Infinity', '-Infinity', and 'NaN', and that this is intentional.
 	// This is desired behavior, because when indexing with them as numeric entities, you are indexing
 	// with the strings '"Infinity"', '"-Infinity"', and '"NaN"' respectively.
-	value, err := strconv.ParseFloat(name, 64)
-	return err == nil && strconv.FormatFloat(value, 'g', -1, 64) == name
+	return numberToString(stringToNumber(name)) == name
 }
 
 func isPropertyName(node *Node) bool {
@@ -3926,4 +3926,19 @@ func getTextOfJsxNamespacedName(node *Node) string {
 
 func isThisProperty(node *Node) bool {
 	return (isPropertyAccessExpression(node) || isElementAccessExpression(node)) && getAccessedExpression(node).kind == SyntaxKindThisKeyword
+}
+
+func numberToString(f float64) string {
+	// !!! This function should behave identically to the expression `"" + f` in JS
+	return strconv.FormatFloat(f, 'g', -1, 64)
+}
+
+func stringToNumber(s string) float64 {
+	// !!! This function should behave identically to the expression `+s` in JS
+	// This includes parsing binary, octal, and hex numeric strings
+	value, err := strconv.ParseFloat(s, 64)
+	if err != nil {
+		return math.NaN()
+	}
+	return value
 }
--- a/internal/compiler/printer.go
+++ b/internal/compiler/printer.go
@@ -1,7 +1,6 @@
 package compiler
 
 import (
-	"strconv"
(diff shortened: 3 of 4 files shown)


internal/compiler/mapper.go
```diff
@@ -0,0 +1,207 @@
+package compiler
+
+// TypeMapper
+
+type TypeMapper struct {
+	data TypeMapperData
+}
```
jakebailey, line 7: Does the performance of these kinds of mappers actually end up mattering? Would have thought a plain interface would be sufficient. I'm not sure the scale of these or how they're referenced, though. I was thinking that they aren't nearly large enough or referenced enough to where it would end up mattering.

internal/compiler/mapper.go
```diff
@@ -0,0 +1,207 @@
...
+
+type CompositeTypeMapper struct {
+	TypeMapper
+	c  *Checker
+	m1 *TypeMapper
+	m2 *TypeMapper
+}
+
+func newCompositeTypeMapper(c *Checker, m1 *TypeMapper, m2 *TypeMapper) *TypeMapper {
+	m := &CompositeTypeMapper{}
```
jakebailey, line 193: Optimization idea at some point in the future; if these chains get long and we combine multiple composite mappers, we should just make `CompositeTypeMapper` hold a slice of other mappers and then write a loop out, rather than recursing.

internal/compiler/types.go
```diff
@@ -903,6 +907,20 @@ const (
 
 type AccessFlags uint32
 
+const (
+	AccessFlagsNone                       AccessFlags = 0
+	AccessFlagsIncludeUndefined           AccessFlags = 1 << 0
```
jakebailey, line 912: My porting script was not smart enough to rewrite these sorts of things to `iota`; would be good in the future to replace these given we don't actually care what they are as long as they are different bits.

internal/compiler/utilities.go
```diff
@@ -617,6 +651,13 @@ func replaceElement[T any](slice []T, i int, t T) []T {
 	return result
 }
 
+func identical[T any](s1 []T, s2 []T) bool {
```
jakebailey, line 655: Not in this PR (or maybe never), but the way the stdlib chooses to write these is:

```go
func identical[Slice ~[]E, E any](s1, s2 Slice) bool { ... }
```

I don't think we have any named slice types at the moment where this matters, though.

internal/compiler/utilities.go
```diff
@@ -2223,6 +2272,14 @@ func getBodyOfNode(node *Node) *Node {
 	return nil
 }
 
+func getFlowNodeOfNode(node *Node) *FlowNode {
+	flowNodeData := node.FlowNodeData()
+	if flowNodeData != nil {
```
jakebailey, line 2278: Style nit you don't need to apply in this PR or anything, but:

```suggestion
	if flowNodeData := node.FlowNodeData(); flowNodeData != nil {
```

(There are lots of examples of this, but, it's just ported code, so it's okay)

internal/compiler/utilities.go
```diff
@@ -3739,3 +3832,98 @@ func getContainingFunction(node *Node) *Node {
...
+	//
+	// Here, we test whether 'ToString(ToNumber(name))' is exactly equal to 'name'.
+	// The '+' prefix operator is equivalent here to applying the abstract ToNumber operation.
+	// Applying the 'toString()' method on a number gives us the abstract ToString operation on a number.
+	//
+	// Note that this accepts the values 'Infinity', '-Infinity', and 'NaN', and that this is intentional.
+	// This is desired behavior, because when indexing with them as numeric entities, you are indexing
+	// with the strings '"Infinity"', '"-Infinity"', and '"NaN"' respectively.
+	value, err := strconv.ParseFloat(name, 64)
+	return err == nil && strconv.FormatFloat(value, 'g', -1, 64) == name
```
jakebailey: Can you leave a comment or something here noting that we'll have to replace this code?

Go's number parsing is not the same as JSs in a number of ways, such that we can't actually use it and get reliable results for edge cases.

(Not sure where else we've used `strconv` in load-bearing code; this is the first one I've noticed when skimming the code.)

internal/compiler/utilities.go
```diff
@@ -3739,3 +3832,98 @@ func getContainingFunction(node *Node) *Node {
...
+
+/**
+ * Gets the symbolic name for a member from its type.
+ */
+func getPropertyNameFromType(t *Type) string {
+	switch {
+	case t.flags&TypeFlagsStringLiteral != 0:
+		return t.AsLiteralType().value.(string)
+	case t.flags&TypeFlagsNumberLiteral != 0:
+		return strconv.FormatFloat(t.AsLiteralType().value.(float64), 'g', -1, 64)
```
jakebailey: Ah this is another case; I don't think we'll end up being able to use strconv.

internal/compiler/checker.go
```diff
@@ -3868,6 +3927,83 @@ func (c *Checker) resolveAnonymousTypeMembers(t *Type) {
 	}
 }
 
+// The mappingThisOnly flag indicates that the only type parameter being mapped is "this". When the flag is true,
+// we check symbols to see if we can quickly conclude they are free of "this" references, thus needing no instantiation.
+func (c *Checker) createInstantiatedSymbolTable(symbols []*Symbol, m *TypeMapper) SymbolTable {
+	if len(symbols) == 0 {
+		return nil
+	}
+	result := make(SymbolTable)
```
jakebailey, line 3937: The linter will eventually suggest this, but consider writing:

```suggestion
	result := make(SymbolTable, len(symbols))
```

Which will allocate all of the memory needed here up-front.

internal/compiler/checker.go
```diff
@@ -3868,6 +3927,83 @@ func (c *Checker) resolveAnonymousTypeMembers(t *Type) {
...
+	return result
+}
+
+// The mappingThisOnly flag indicates that the only type parameter being mapped is "this". When the flag is true,
+// we check symbols to see if we can quickly conclude they are free of "this" references, thus needing no instantiation.
+func (c *Checker) instantiateSymbolTable(symbols SymbolTable, m *TypeMapper, mappingThisOnly bool) SymbolTable {
+	if len(symbols) == 0 {
+		return nil
+	}
+	result := make(SymbolTable)
```
jakebailey, line 3950: Similarly here.
```suggestion
	result := make(SymbolTable, len(symbols))
```

internal/compiler/checker.go
```diff
@@ -5292,7 +5893,28 @@ func (c *Checker) newLiteralType(flags TypeFlags, value any, regularType *Type)
 }
 
 func (c *Checker) newObjectType(objectFlags ObjectFlags, symbol *Symbol) *Type {
-	t := c.newType(TypeFlagsObject, objectFlags, &AnonymousType{})
+	var data TypeData
+	switch {
+	case objectFlags&ObjectFlagsClassOrInterface != 0:
+		data = &InterfaceType{}
```
jakebailey, line 5900: Maybe not for this PR, but I suspect we'll eventually end up making allocators for each of these.

internal/compiler/checker.go
```diff
@@ -6380,131 +6999,453 @@ func (c *Checker) getCrossProductUnionSize(types []*Type) int {
...
+							case indexType.flags&TypeFlagsStringLiteral != 0:
+								errorInfo = NewMessageChain(nil, diagnostics.Property_0_does_not_exist_on_type_1, indexType.AsLiteralType().value, c.typeToString(objectType))
+							case indexType.flags&TypeFlagsNumberLiteral != 0:
+								errorInfo = NewMessageChain(nil, diagnostics.Property_0_does_not_exist_on_type_1, indexType.AsLiteralType().value, c.typeToString(objectType))
+							case indexType.flags&(TypeFlagsNumber|TypeFlagsString) != 0:
+								errorInfo = NewMessageChain(nil, diagnostics.No_index_signature_with_a_parameter_of_type_0_was_found_on_type_1, c.typeToString(indexType), c.typeToString(objectType))
+							}
+							diagnostic := c.error(accessExpression, diagnostics.Element_implicitly_has_an_any_type_because_expression_of_type_0_can_t_be_used_to_index_type_1, c.typeToString(fullIndexType), c.typeToString(objectType))
+							if errorInfo != nil {
+								diagnostic.addMessageChain(errorInfo)
```
jakebailey, line 7251: This is probably ported but I do hope that we one day flip this all to be a more return-early style rather than nested so deeply.

internal/compiler/mapper.go
ahejlsberg: A plain interface would work, but we store a *lot* of references to mappers (every instantiated type has one) so pointers keep the footprint down. For example, when resolving the members of an instantiated object type, the mapper is propagated to all of the member symbols which in turn propagates it to the member types.

internal/compiler/types.go
ahejlsberg: Agreed.

internal/compiler/utilities.go
ahejlsberg: Yeah, I'll keep it in mind.

internal/compiler/utilities.go
ahejlsberg: Sure, I'll add a comment.

ahejlsberg:
Latest commit funnels of our string-to-number and number-to-string conversions into two functions, `stringToNumber` and `numberToString`, with comments that we need these functions to have JS semantics.

PR #34, opened 2024-10-28
Add JSDoc nodes to AST

sandersn:
1. No usage yet.
2. No tests, but there aren't any changes in my AST baselines yet because there's no usage.
3. I left parameters not using NodeArray substitutes but all the other NodeArrays in TS I translated to custom *List nodes.
4. Every HasJSDoc node visits JSDoc. We might want to make that configurable if JSDoc parsing is off -- but `visit(nil)` should be fast, I hope?

Edit: I'm adding my usage analysis of JSDoc here because it's going to affect all my subsequent JS PRs.

## Background
`top32js` is filtered from the top 150 or so JS projects on github. Of these 150, 32 compiled successfully and found JSDoc in the parse tree:

Projects that compiled and use JSDoc:
	- Chart.js
	- cypress
	- uni-app
	- fabric.js
	- prettier
	- monaco-editor (just a vendored copy of typescriptServices.js)
	- pdf.js
	- parcel (small)
	- svelte
	- todomvc (one (1) tag)
	- javascript-algorithms
	- video.js
	- webpack
This list is heavily (entirely?) weighted towards TS-via-JSDoc projects. That's our target audience, but I don't want to leave our audience of never-heard-of-TS behind, necessarily. So I took the same 150 or so JS projects and extracted all the JS files in them and their dependencies. I grouped them into npm packages and excluded those without a publish in the last 2 years. This gave me `npm6522`, a set of 6522 packages. The JS files in here represent what people might randomly open in VSCode. It includes lots of TS emit and some flow code.

Here's what I *imagine* about TS-via-JSDoc users: 
- They want Typescript semantics without Typescript syntax. 
- They are willing to write JSDoc that Typescript understands. 
- They adopt new, Typescript-only tags quickly, even though only Typescript understands them (see `@import`).

Here's what I *imagine* about random-JS users:
- They want quick info and completions as much as possible.
- Completion lists should be longer rather than shorter.
- Quick info should be as simple as possible, but no simpler.
- They dislike red and orange squigglies if they're not correct or they have no way to fix them.

So, my approach to cutting JSDoc features is to remove features that TS-via-JSDoc projects don't use much, particularly if it's only one project. Even If a feature is still used widely in JS shipped by NPM, I'll remove if it I think that the loss won't hurt quick info or completions much. Because random-JS users will only see Typescript syntax errors, we can drop features that would add semantic errors. Fortunately, unknown tags parse just fine, so we can drop whole tags too.

## Results

I'm going to drop the following syntax in this PR. Later PRs will of course drop the semantics (if any).

- NonNullableType, postfix version -- prefix NonNullableType is emitted by the Closure Compiler, so might be OK to drop too.
- UnknownType -- drop: no usage in top32, low usage outside. Random-JS quickinfo will give `any` either way.
- NamepathType -- no usage in top32, almost no usage outside. It was only added to avoid parse errors, but people have stopped using this syntax.
- NullableType, postfix version -- prefix NullableType is used in Flow code, which people may open in VSCode. The semantics should probably change to `T | undefined` if they're not already that.

- ClassTag - almost no usage in top32, semantics are redundant with assignment declaration inference. It's used a lot in random-JS, but it's because TS emits it for ES5.
- EnumTag - no (correct) usage in top32, low usage outside. Semantics are confusing wrt TS `enum`.
- AuthorTag - no usage in top32, low usage outside. No semantics; it's just to avoid parsing an email address as a tag. Dropping it only means that the quickinfo it's in will be mis-formatted with an extra tag.
- ThrowsTag - low usage in top32. No semantics.

Honourable mentions that I will *not* cut, but are hardly used at all:
- OverloadTag, SatisfiesTag, ReadonlyTag, PublicTag, OverrideTag, ImplementsTag, ThisTag - these are all equivalent to TS, and many are new. Still not used very much though!
- VariadicType - low usage in top32, low usage outside. Dropping this would add parse errors. I plan to simplify semantics to just be equal to `T[]`, with no fancy checking whether it's used only for `arguments`.
- FunctionType - low usage in top32, fairly low usage outside. Dropping this would add parse errors. Later, I'll check whether semantics can be simplified in some way.

## Summary
Drop the following types and tags:
- [x] UnknownType
- [x] NamepathType
- [x] postfix NonNullableType
- [x] postfix NullableType (parse as OptionalType regardless of context)
- [x] ClassTag
- [x] ThrowsTag
- [x] EnumTag
- [x] AuthorTag

Note that dropping a tag will result it being parsed as JSDocUnknownTag, which API consumers should still be able to retrieve, just less conveniently.


internal/compiler/ast.go
```diff
@@ -1066,10 +1232,6 @@ func (f *NodeFactory) NewTryStatement(tryBlock *BlockNode, catchClause *CatchCla
 	return f.NewNode(SyntaxKindTryStatement, data)
 }
 
-func (node *TryStatement) Kind() SyntaxKind {
-	return SyntaxKindTryStatement
-}
```
jakebailey: Why did these all get deleted?

internal/compiler/ast.go
```diff
@@ -602,6 +746,28 @@ type FlowNodeBase struct {
 
 func (node *FlowNodeBase) FlowNodeData() *FlowNodeBase { return node }
 
+type HasJSDocBase struct {
+	jsDocs *JSDocListNode
```
jakebailey: Most nodes getting a pointer bigger, sigh

Would be nice to perf test the difference here (my additional benchmark PR isn't ready/merged)

internal/compiler/ast.go
sandersn: These were the few that were left behind after the switch to using `Node.kind` instead of `Node.Kind()`. Which I *think* happened.

jakebailey: I was just trying to decide if this was on purpose for perf (since the kinds are statically known for each individual node type)

internal/compiler/ast.go
sandersn: I talked to Anders about this and he suggested a Node flag for HasJSDoc, plus a lookaside map Node->JSDocListNode.

sandersn: This is in now.

internal/ast/ast.go
```diff
@@ -815,10 +991,6 @@ func (f *NodeFactory) NewTypeParameterDeclaration(modifiers *ModifierListNode, n
 	return f.NewNode(KindTypeParameter, data)
 }
 
-func (node *TypeParameterDeclaration) Kind() Kind {
-	return KindTypeParameter
-}
```
jakebailey, line 867: These `Kind()` changes seem unrelated to your PR.

internal/ast/ast.go
```diff
@@ -725,6 +852,55 @@ type FlowNodeBase struct {
...
 
+// if you provide nil for file, this code will walk to the root of the tree to find the file
+func (node *Node) JSDoc(file *SourceFile) *JSDocListNode {
+	if node.Flags&NodeFlagsHasJSDoc == 0 {
+		return nil
+	}
+	if file == nil {
+		current := node
+		parent := node.Parent
+		for parent != nil {
```
jakebailey: I think we have a helper for this?

internal/ast/ast.go
```diff
@@ -725,6 +852,55 @@ type FlowNodeBase struct {
...
+	file.jsdocCacheMu.Lock()
+	defer file.jsdocCacheMu.Unlock()
+	if jsdocs, ok := file.jsdocCache[node]; ok {
+		return jsdocs
+	}
+	return nil
+}
+func (node *Node) SetJSDoc(file *SourceFile, jsDocs *JSDocListNode) {
+	if node.Flags&NodeFlagsHasJSDoc == 0 {
+		node.Flags &= NodeFlagsHasJSDoc
```
jakebailey, line 906: Given this modifies the node, it's not clear to me that it's even worth having a mutex on `jsdocCache`, no? Since this could race too?

Making our AST safe for concurrent use when nodes are modified seems like it'll require more changes, so probably not needed right now.

internal/ast/ast.go
sandersn: They are, they are long leftovers of having a Kind() method instead of a Kind property.

internal/ast/ast.go
sandersn: Yeah, that looks like it could race. 
To check my understanding: it sounds to me like you're suggesting dropping the mutex from this PR entirely, right?

jakebailey: Yes given we haven't locked anything else on the AST yet

internal/ast/ast.go
```diff
@@ -764,6 +889,44 @@ type FlowNodeBase struct {
...
+
+func (f *NodeFactory) NewJSDocList(jsDocs []*Node) *Node {
+	data := &JSDocList{}
+	data.jsDocs = jsDocs
+	return f.NewNode(KindJSDocList, data)
+}
+
+func (node *JSDocList) ForEachChild(v Visitor) bool {
+	return visitNodes(v, node.jsDocs)
+}
```
jakebailey: These list types should go away, I think, now that we have NodeList and no more Nodes that are just lists.

internal/ast/kind.go
```diff
@@ -184,6 +184,9 @@ const (
 	KindModifierList
 	KindTypeParameterList
 	KindTypeArgumentList
+	KindJSDocList
+	KindJSDocCommentList
+	KindJSDocTagList
```
jakebailey: Same here; probably want to merge from main.

internal/ast/ast.go
```diff
@@ -772,6 +885,29 @@ type FlowNodeBase struct {
...
+	if jsdocs, ok := file.jsdocCache[node]; ok {
+		return jsdocs
+	}
+	return nil
+}
+func (node *Node) SetJSDoc(file *SourceFile, jsDocs []*Node) {
+	if node.Flags&NodeFlagsHasJSDoc == 0 {
+		node.Flags &= NodeFlagsHasJSDoc
+	}
+	file.jsdocCache[node] = jsDocs
```
jakebailey, line 908: It seems like this file also needs the `GetSourceFileOfNode` change? Or, is it guaranteed that setters have the source file anyway?

sandersn: I intend this to be called much less than the getter -- basically just the parser which definitely has the source file. So I'd prefer to be stricter for the setter, and require the source file.

PR #38, opened 2024-10-29
Index types, union/intersection properties, and call, construct and index signatures

ahejlsberg:
This PR ports the following:

* Resolution and instantiation of index types (`keyof`).
* Creation of union and intersection properties.
* Resolution and instantiation of call, construct, and index signatures.

### DIFF
--- a/internal/compiler/utilities.go
+++ b/internal/compiler/utilities.go
@@ -924,10 +924,6 @@ func isOuterExpression(node *Node, kinds OuterExpressionKinds) bool {
 	return false
 }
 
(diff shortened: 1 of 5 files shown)


internal/compiler/utilities.go
```diff
@@ -3826,6 +3790,34 @@ func (m *orderedMap[K, V]) add(key K, value V) {
 	m.values = append(m.values, value)
 }
 
+type set[T comparable] struct {
```
jakebailey, line 3793: Would be nice to move this into another package, but not required for this PR of course.

PR #40, opened 2024-10-29
Add package.json parsing

andrewbranch:
Adds a `packagejson` package with types that implement custom unmarshaling to help handle nulls, object key order, polymorphic structures like `exports`, and unexpected field types.

### DIFF
--- a/internal/collections/map.go
+++ b/internal/collections/map.go
@@ -194,16 +194,16 @@ func (m *Map[K, V]) UnmarshalJSONV2(dec *jsontext.Decoder, opts json2.Options) e
 	if err != nil {
 		return err
 	}
-	if token.Kind() == jsontext.Null.Kind() {
+	if token.Kind() == 'n' { // jsontext.Null.Kind()
 		// By convention, to approximate the behavior of Unmarshal itself,
 		// Unmarshalers implement UnmarshalJSON([]byte("null")) as a no-op.
 		// https://pkg.go.dev/encoding/json#Unmarshaler
 		return nil
 	}
-	if token.Kind() != jsontext.ObjectStart.Kind() {
+	if token.Kind() != '{' { // jsontext.ObjectStart.Kind()
 		return fmt.Errorf("cannot unmarshal non-object JSON value into Map")
 	}
-	for dec.PeekKind() != jsontext.ObjectEnd.Kind() {
+	for dec.PeekKind() != '}' { // jsontext.ObjectEnd.Kind()
 		var key K
 		var value V
 		if err := json2.UnmarshalDecode(dec, &key, opts); err != nil {
--- a/internal/compiler/packagejson/jsonvalue.go
+++ b/internal/compiler/packagejson/jsonvalue.go
@@ -21,6 +21,25 @@ const (
 	JSONValueTypeObject
 )
 
+func (t JSONValueType) String() string {
+	switch t {
+	case JSONValueTypeNull:
+		return "null"
+	case JSONValueTypeString:
+		return "string"
+	case JSONValueTypeNumber:
+		return "number"
+	case JSONValueTypeBoolean:
+		return "boolean"
+	case JSONValueTypeArray:
+		return "array"
+	case JSONValueTypeObject:
+		return "object"
+	default:
+		return fmt.Sprintf("unknown(%d)", t)
+	}
+}
+
 type JSONValue struct {
 	Type  JSONValueType
 	Value any
@@ -50,8 +69,7 @@ func (v *JSONValue) UnmarshalJSONV2(dec *jsontext.Decoder, opts json2.Options) e
 
 func unmarshalJSONValue[T any](v *JSONValue, data []byte) error {
 	if string(data) == "null" {
-		v.Value = nil
-		v.Type = JSONValueTypeNull
+		*v = JSONValue{Type: JSONValueTypeNull}
 	} else if data[0] == '"' {
 		v.Type = JSONValueTypeString
 		return json.Unmarshal(data, &v.Value)
@@ -84,7 +102,7 @@ func unmarshalJSONValue[T any](v *JSONValue, data []byte) error {
 
 func unmarshalJSONValueV2[T any](v *JSONValue, dec *jsontext.Decoder, opts json2.Options) error {
 	switch dec.PeekKind() {
-	case jsontext.Null.Kind():
+	case 'n': // jsontext.Null.Kind()
 		if _, err := dec.ReadToken(); err != nil {
 			return err
 		}
@@ -120,7 +138,7 @@ func unmarshalJSONValueV2[T any](v *JSONValue, dec *jsontext.Decoder, opts json2
 		}
 		v.Type = JSONValueTypeObject
 		v.Value = &object
-	case jsontext.True.Kind(), jsontext.False.Kind():
+	case 't', 'f': // jsontext.True.Kind(), jsontext.False.Kind()
 		v.Type = JSONValueTypeBoolean
 		if err := json2.UnmarshalDecode(dec, &v.Value, opts); err != nil {
 			return err


internal/compiler/packagejson/packagejson.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type PathFields struct {
+	TSConfig      Expected[string] `json:"tsconfig"`
+	Main          Expected[string] `json:"main"`
+	Types         Expected[string] `json:"types"`
+	Typings       Expected[string] `json:"typings"`
+	TypesVersions JSONValue        `json:"typesVersions"`
+	Imports       JSONValue        `json:"imports"`
+	Exports       Exports          `json:"exports"`
```
andrewbranch, line 20: I may make dedicated types for `TypesVersions` and `Imports` and/or expand the `Exports` type to be suitable for those as I start to use this in module resolution.

internal/collections/map.go
```diff
@@ -1,9 +1,16 @@
...
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
 	"iter"
 	"maps"
+	"reflect"
 	"slices"
+
+	jsonExp "github.com/go-json-experiment/json"
```
jakebailey: camelCase names aren't really standard for package names, including aliases; I would probably switch this to be `json2` everywhere if anything.

internal/collections/map.go
```diff
@@ -135,3 +148,68 @@ func (m *Map[K, V]) clone() Map[K, V] {
 		mp:   maps.Clone(m.mp),
 	}
 }
+
+func (m *Map[K, V]) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
+		return nil
+	}
```
jakebailey: This strikes me as odd; shouldn't this zero the map or something?

internal/collections/map.go
```diff
@@ -135,3 +148,68 @@ func (m *Map[K, V]) clone() Map[K, V] {
...
+		if key, ok := nameToken.(K); ok {
+			var valueBytes json.RawMessage
+			if err := dec.Decode(&valueBytes); err != nil {
+				return err
+			}
+			var value V
+			if err := json.Unmarshal(valueBytes, &value); err != nil {
+				return err
+			}
+			m.Set(key, value)
```
jakebailey, line 184: One thing to consider here is that unmarshalling the same thing twice will just continually add entries into the same map. I'm not entirely certain that's the right thing to do.

internal/collections/map.go
```diff
@@ -135,3 +148,68 @@ func (m *Map[K, V]) clone() Map[K, V] {
 		mp:   maps.Clone(m.mp),
 	}
 }
+
+func (m *Map[K, V]) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
+		return nil
+	}
+	dec := json.NewDecoder(bytes.NewReader(data))
```
andrewbranch, line 159: I couldn’t think of another way to do this. I included JSON V1 implementations to prove that we aren’t taking a hard dependency on an experimental package, but this is a lot slower than `UnmarshalJSONV2`.

internal/collections/map.go
```diff
@@ -135,3 +148,68 @@ func (m *Map[K, V]) clone() Map[K, V] {
...
+			var value V
+			if err := json.Unmarshal(valueBytes, &value); err != nil {
+				return err
+			}
+			m.Set(key, value)
+		} else {
+			return fmt.Errorf("cannot unmarshal key into Map[%v, ...]", reflect.TypeFor[K]())
+		}
+	}
+	return nil
```
jakebailey, line 189: Does this need to error out if the decoder has any remaining tokens? (Maybe an invariant of the interface is it's not possible)

internal/collections/map.go
```diff
@@ -135,3 +148,68 @@ func (m *Map[K, V]) clone() Map[K, V] {
...
+}
+
+func (m *Map[K, V]) UnmarshalJSONV2(dec *jsontext.Decoder, opts jsonExp.Options) error {
+	token, err := dec.ReadToken()
+	if err != nil {
+		return err
+	}
+	if token.Kind() == jsontext.Null.Kind() {
+		return nil
+	}
```
jakebailey: Sort of the same comments here about  null and overwriting the map. Sort of think the map has to be zero'd.

internal/collections/map.go
jakebailey: Or, error, because it's not an object? Presumably null is handled elsewhere?

internal/compiler/packagejson/expected.go
```diff
@@ -0,0 +1,20 @@
...
+	"encoding/json"
+)
+
+type Expected[T any] struct {
+	Null  bool
+	Valid bool
+	Value T
+}
+
+func (e *Expected[T]) UnmarshalJSON(data []byte) error {
```
jakebailey, line 13: Are things slow here because this one doesn't have a V2 interface?

internal/compiler/packagejson/packagejson_test.go
```diff
@@ -0,0 +1,55 @@
...
+						b.Fatal(err)
+					}
+				}
+			})
+		})
+
+		b.Run("ParseJSONText", func(b *testing.B) {
+			b.Run(f.Name(), func(b *testing.B) {
+				for i := 0; i < b.N; i++ {
+					compiler.ParseJSONText(f.Name(), string(content))
```
jakebailey, line 50: I was going to say "this is a cycle" but then I remembered this is a `_test` package so that isn't a problem, nice

internal/compiler/packagejson/packagejson.go
```diff
@@ -0,0 +1,41 @@
...
+type DependencyFields struct {
+	Dependencies         Expected[map[string]string] `json:"dependencies"`
+	PeerDependencies     Expected[map[string]string] `json:"peerDependencies"`
+	OptionalDependencies Expected[map[string]string] `json:"optionalDependencies"`
+}
+
+type Fields struct {
+	HeaderFields
+	PathFields
+	DependencyFields
```
jakebailey, line 32: Do these actually have to be embeddings? The other types aren't referenced anywhere (so far)

internal/collections/map_test.go
```diff
@@ -156,3 +158,29 @@ func TestMapWithSizeHint(t *testing.T) {
 
 	assert.Assert(t, allocs < 10, "allocs = %v", allocs)
 }
+
+func TestMapUnmarshalJSON(t *testing.T) {
+	t.Parallel()
+
+	testMapUnmarshalJSON(t, json.Unmarshal)
+	testMapUnmarshalJSON(t, func(in []byte, out any) error { return json2.Unmarshal(in, out) })
```
jakebailey: Consider using `t.Run` here instead to give these names and let them fail individually.

internal/compiler/packagejson/packagejson.go
andrewbranch: There will be other places in the future when we don't need to parse out all the fields.

internal/compiler/packagejson/expected.go
andrewbranch: I tried to make a V2 implementation here and discovered an annoyance. My implementation was

```go
func (e *Expected[T]) UnmarshalJSONV2(dec *jsontext.Decoder, opts json2.Options) error {
  if json2.UnmarshalDecode(dec, &e.Value, opts) != nil {
    e.Valid = false
  }
  return nil
}
```

But when the `e.Valid = false` branch is taken, the decoder also marks itself as in a bad state and errors for all subsequent properties within the same JSON object. With this approach, it seems you _have_ to inspect the decoder tokens and avoid errors when delegating to other types' unmarshalling implementations. It's not a huge problem to do that, but to answer your question, I _think_ we aren't losing out an anything here since we're currently only using `Expected` at top-level fields... I can try out a benchmark to validate that, though.

internal/collections/map.go
andrewbranch: There’s a bit in the Go docs that says

> By convention, to approximate the behavior of [Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) itself, Unmarshalers implement `UnmarshalJSON([]byte("null"))` as a no-op.

The content at the link says a little more:

> The JSON null value unmarshals into an interface, map, pointer, or slice by setting that Go value to nil. Because null is often used in JSON to mean “not present,” unmarshaling a JSON null into any other Go type has no effect on the value and produces no error.

I find this a strange choice, and we could choose to deviate from convention, but it also feels like it could be an antipattern to unmarshal into a non-zero struct in the first place?

internal/collections/map.go
andrewbranch: Isn’t it? Last-in wins?

internal/collections/map.go
andrewbranch: No, the decoder is valid over the whole top-level JSON string, so there _will_ be remaining tokens unless we happened to be unmarshaling the top-level JSON object. The work of this function is done when it has seen a balanced closing brace, and reading any more tokens would leave the decoder in a bad state for subsequent fields.

internal/collections/map.go
jakebailey: I guess if that's the suggestion... I just think it sounds error prone, especially when we're more strongly handling this with `Expected`.

andrewbranch: Yeah, I think at least `Expected` should zero the value since it recognizes `null` as a distinct value.

internal/collections/map.go
jakebailey: True, yeah: https://go.dev/play/p/RB7-Sqv4jJ5

v2 docs say

> Unmarshaling a JSON value into a non-empty Go value always merges if the input is an object, and otherwise replaces.

internal/compiler/packagejson/expected.go
jakebailey: Manually looking at the tokens is what I would expectish, yeah.

internal/compiler/packagejson/expected.go
```diff
@@ -0,0 +1,23 @@
...
+	if string(data) == "null" {
+		var value T
+		e.Valid = false
+		e.Null = true
+		e.Value = value
+	} else if json.Unmarshal(data, &e.Value) == nil {
+		e.Valid = true
+	}
+	return nil
```
jakebailey: At this point, I would actually suggest writing the code like this:

```suggestion
	if string(data) == "null" {
		*e = Expected[T]{Null: true}
		return nil
	}
	if json.Unmarshal(data, &e.Value) == nil {
		e.Valid = true
	}
	return nil
```

It's convention to try and return early, but, what you've written is okay too. The `*e` thing is a separate change.

internal/compiler/packagejson/expected.go
```diff
@@ -0,0 +1,23 @@
...
+type Expected[T any] struct {
+	Null  bool
+	Valid bool
+	Value T
+}
+
+func (e *Expected[T]) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
+		var value T
+		e.Valid = false
```
jakebailey: Why `false` here? Or does null not imply valid?

internal/compiler/packagejson/expected.go
andrewbranch: I actually don’t even know how we can implement this without reflecting on `T`... this is probably something we need to give feedback about on the experimental code.

internal/compiler/packagejson/expected.go
andrewbranch: It’s arbitrary, since you can’t express whether you expect `null` as a valid expression of `T`. We could instead express this as a tristate of [null, unexpected type, expected type]

internal/compiler/packagejson/jsonvalue.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type JSONValue struct {
+	Type  JSONValueType
+	Value any
+}
+
+func (v *JSONValue) AsObject() *collections.Map[string, *JSONValue] {
+	if v.Type != JSONValueTypeObject {
+		panic(fmt.Sprintf("expected object, got %v", v.Type))
```
jakebailey, line 50: `JSONValueType` does not have a `String()` method; this will probably just print things like `JSONValueType(3)`.

internal/compiler/packagejson/jsonvalue.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (v *JSONValue) UnmarshalJSONV2(dec *jsontext.Decoder, opts json2.Options) error {
+	return unmarshalJSONValueV2[JSONValue](v, dec, opts)
+}
+
+func unmarshalJSONValue[T any](v *JSONValue, data []byte) error {
+	if string(data) == "null" {
+		v.Value = nil
+		v.Type = JSONValueTypeNull
```
jakebailey: Just as a style thing here (sorry, I keep giving you these), I'd probably have written this myself in a form which just assigns over all of `*v` at once just to be sure that I haven't forgotten to set a case. Maybe it's not a bug deal here; there's just two things to set in v so it's not too bad to vet each branch.

internal/compiler/packagejson/jsonvalue.go
```diff
@@ -0,0 +1,135 @@
...
+	} else {
+		v.Type = JSONValueTypeNumber
+		return json.Unmarshal(data, &v.Value)
+	}
+	return nil
+}
+
+func unmarshalJSONValueV2[T any](v *JSONValue, dec *jsontext.Decoder, opts json2.Options) error {
+	switch dec.PeekKind() {
+	case jsontext.Null.Kind():
```
jakebailey: I think this can just be `'n'` and then this whole switch case will be static.

internal/compiler/packagejson/jsonvalue.go
```diff
@@ -0,0 +1,135 @@
...
+		v.Type = JSONValueTypeArray
+		v.Value = elements
+	case '{':
+		var object collections.Map[string, *T]
+		if err := json2.UnmarshalDecode(dec, &object, opts); err != nil {
+			return err
+		}
+		v.Type = JSONValueTypeObject
+		v.Value = &object
+	case jsontext.True.Kind(), jsontext.False.Kind():
```
jakebailey: Similarly here for `'t', 'f'`: https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20240815175050-ebd3a8989ca1/jsontext#Kind

PR #42, opened 2024-10-30
Port text writer abstraction for using in printing ASTs

weswigham:
This PR just adds the `NewTextWriter` constructor and a `SingleLineStringWriter` singleton for use in printing ASTs. I've refrained from including any of my partial work on the emitter/printer itself in this PR, since module structure necessitates a big pile of renaming or additional helper methods on AST nodes, which probably aren't best included here. <sub>Though I do want to know - what's the preferred direction here? Do I yield to the existing mess and throw everything into the same folder/file (I'd rather not perpetuate that, personally)? Do I add adhoc/codegen accessor methods for all the AST node members (`kind` included)? Do I churn and capitalize every field in `ast.go` so they're accessible in subfolders (probably best done as a separate refactor)?</sub>


go.mod
```diff
@@ -7,9 +7,12 @@ require (
 	gotest.tools/v3 v3.5.1
 )
 
+require github.com/google/go-cmp v0.6.0 // indirect
```
jakebailey: This file got wonky; you probably just want to revert to what is in current main. (now that the dep is real)

internal/compiler/string_util/pkg.go
```diff
@@ -0,0 +1,2 @@
+// Exports common rune utilities for parsing and emitting javascript
```
jakebailey: Since this file only has one file, you can probably just move this into the other file.

Also, convention says to start the comment like `// Package stringutil blah blah blah`.

internal/compiler/string_util/util.go
```diff
@@ -0,0 +1,113 @@
+package string_util
```
jakebailey: Package names in go generally are just lowercase letters, no underscores, etc; `stringutil` is the more conventional name.

internal/compiler/string_util/util.go
```diff
@@ -0,0 +1,113 @@
...
+func IsOctalDigit(ch rune) bool {
+	return ch >= '0' && ch <= '7'
+}
+
+func IsHexDigit(ch rune) bool {
+	return ch >= '0' && ch <= '9' || ch >= 'A' && ch <= 'F' || ch >= 'a' && ch <= 'f'
+}
+
+func IsASCIILetter(ch rune) bool {
+	return ch >= 'A' && ch <= 'Z' || ch >= 'a' && ch <= 'z'
```
jakebailey: Am I wrong to think that a lot of this is already present in the `unicode` package?

internal/compiler/string_util/util.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+		}
+	}
+	result = append(result, TextPos(lineStart))
+	return result
+}
+
+// TextPos
+
+type TextPos int32
```
jakebailey: I don't know that this should be the eventual place for this to live, if you're just trying to break cycles. Probably we should end up with a `textpos` package with  a `textpos.Pos` type or something.

internal/compiler/printer/emittextwriter.go
```diff
@@ -0,0 +1,32 @@
+package printer
+
+import "github.com/microsoft/typescript-go/internal/compiler"
+
+// These methods are all local to this module, so outside this module an `EmitTextWriter` is basically opaque
```
jakebailey: This will attach the comment to the declaration for hovers; not sure that's desirable.

internal/compiler/printer/pkg.go
```diff
@@ -0,0 +1,2 @@
+// Package exports a Printer for pretty-printing TS ASTs and writer interfaces and implementations for using them
```
jakebailey: Here too; probably just `printer.go` is okay for this.

I only had `pkg.go` in the `diagnostics` package at the time because there were no non-generated files, but I ended up moving code so `pkg.go` for a doc wasn't required.

internal/compiler/printer/printer_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/compiler"
+	"gotest.tools/v3/assert"
+)
+
+func TestSanityCheckParsePrintRoundtrip(t *testing.T) {
+	samples := [](string){
```
jakebailey:
```suggestion
	samples := []string{
```

internal/compiler/printer/printer_test.go
```diff
@@ -0,0 +1,42 @@
...
+		// }`,
+		// 		`namespace Foo {
+		// 	export function bar(x?: string[]) {
+		// 		return null as any as (A extends B ? true : never)
+		// 	}
+		// }`,
+	}
+
+	for _, sample := range samples {
+		t.Run(sample, func(t *testing.T) {
```
jakebailey: This is all commented out, but you probably will need to come up with test names here given this just uses random text which will not be easily selectable from `go test`.

internal/compiler/printer/singlelinestringwriter.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/microsoft/typescript-go/internal/compiler/string_util"
+)
+
+var SingleLineStringWriter EmitTextWriter = &singleLineStringWriter{}
+
+type singleLineStringWriter struct {
+	str string
+}
+
+// clear implements EmitTextWriter.
```
jakebailey: These functions all seem to not actually say anything about the behavior; I would probably just omit them. (I think gopls generates these? Or my memory is bad.)

internal/compiler/printer/textwriter.go
```diff
@@ -0,0 +1,222 @@
...
+func (w *textWriter) writeSymbol(text string, symbol compiler.Symbol) {
+	w.write(text)
+}
+
+// writeTrailingSemicolon implements EmitTextWriter.
+func (w *textWriter) writeTrailingSemicolon(text string) {
+	w.write(text)
+}
+
+func NewTextWriter(newLine string) EmitTextWriter {
```
jakebailey, line 198: Generally speaking, it's an anti-pattern to return an interface rather than the concrete type, e.g. "accept interfaces, return structs". The rationale is that you can assert that something implements an interface another way, but returning the concrete type will allow callers to directly call methods without potential indirection, and an implementation of an interface will often have more methods than the interface.

I would consider just writing `TextWriter` as a type and then returning `*TextWriter` here.

Though, the interface consists entirely of unexported methods, so it is in general a little weird... So maybe it's fine I guess.

internal/compiler/string_util/util.go
jakebailey: Ah, this is straight code movement from the `compiler` package.

internal/compiler/printer/textwriter.go
jakebailey: Is this ever actually called anywhere? Maybe this should just be unexported. I'm only seeing it used locally at the moment, but this is of course only the abstraction so far.

internal/compiler/printer/singlelinestringwriter.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// isAtStartOfLine implements EmitTextWriter.
+func (w singleLineStringWriter) isAtStartOfLine() bool {
+	return false
+}
+
+// rawWrite implements EmitTextWriter.
+func (w *singleLineStringWriter) rawWrite(s string) {
+	w.str += s
```
jakebailey: All of this string `+=`ing seems spooky. A builder is probably the right thing to do here, right? Or am I misunderstanding the usecase?

jakebailey:
Left some comments, less so about the actual functionality at the moment and more Go-ish stuff.

> Though I do want to know - what's the preferred direction here? Do I yield to the existing mess and throw everything into the same folder/file (I'd rather not perpetuate that, personally)?

I'm honestly not sure where to draw the line at this point. It may be easier to just dump this into the `compiler` package for now until stuff is shifted around. Ideally, it would be nice for this to be in a separate package.

> Do I add adhoc/codegen accessor methods for all the AST node members (kind included)?

We'll probably just capitalize them in the future.

> Do I churn and capitalize every field in ast.go so they're accessible in subfolders (probably best done as a separate refactor)?

I think we're going to do that in a separate change, for sure; my next task is generating an `ast` package.

weswigham:
Yeah... I think we should be breaking stuff out ASAP- the tech debt incurred by not doing so initially seems like something we'd almost never pay off, and the longer we wait, the thicker the dependency morass gets and the more work disentangling each component is, so the more inertia the tangled structure has.

Starting with stuff broken out should make it easy to keep it that way, we just need to enforce/swap over to conventions that allow it.

internal/compiler/printer/singlelinestringwriter.go
weswigham: You're right - I swapped to a builder for the non-single-line one but didn't swap this one over. This has been corrected.

internal/compiler/printer/textwriter.go
weswigham: I do think we want to keep it this way - I put an example of the intended ultimate usage for the package into the package docstring, which should help clarify that `NewTextWriter` augha get called externally to be provided as an input for the printer itself (to help configure how low-level parts of the print is done). There's also a semicolon-deferring text writer I need to port for quickfixes, but I figured I could defer on that for a bit.

internal/compiler/printer/singlelinestringwriter.go
weswigham: Yeah, `gopls` generated these - I've removed them.

internal/compiler/printer/printer_test.go
weswigham: I've given these more greppable names~

internal/compiler/string_util/util.go
weswigham: I've extracted this type into a separate package. Right now it's `textpos.TextPos` which is a bit redundant. I almost want to rename the package to `text` and the type to `Pos` - since we'll also probably want to move `TextRange` as well, eventually, and `text.Range` is a lot more reasonable, too.

jakebailey: Yeah, those are good names, though, maybe a little _too_ good 😄

internal/compiler/printer/singlelinestringwriter.go
jakebailey: Yeah, not my favorite feature. Normally it's okay because "make your interfaces small" is the way to do it in Go, but, coming from TS we've obviously not followed that scheme.

internal/compiler/printer/textwriter.go
```diff
@@ -0,0 +1,197 @@
...
+	lastWritten             string
+	indent                  int
+	lineStart               bool
+	lineCount               int
+	linePos                 int
+	hasTrailingCommentState bool
+}
+
+func (w *textWriter) clear() {
+	// Is it worth reusing the old string builder?
```
jakebailey: I think the answer is no; you technically don't want to reuse it so that it's able to give ownership of its buffer up. But, I'm not sure that even works when it's been wrapped up in a struct and isn't just plain on a stack.

internal/compiler/printer/textwriter.go
```diff
@@ -0,0 +1,197 @@
...
+	w.builder.Reset()
+	w.lastWritten = ""
+	w.indent = 0
+	w.lineStart = true
+	w.lineCount = 0
+	w.linePos = 0
+	w.hasTrailingCommentState = false
```
jakebailey: Could this just be `*w = textWriter{newLine: w.newLine, lineStart: true}`? It'd also be nice if the zero value were useful, but `lineStart` is flipped (though not set in `newTextWriter`?) and newLine is required...

internal/compiler/printer/emittextwriter.go
```diff
@@ -0,0 +1,32 @@
...
+	writeComment(text string)
+	writeKeyword(text string)
+	writeOperator(text string)
+	writePunctuation(text string)
+	writeSpace(text string)
+	writeStringLiteral(text string)
+	writeParameter(text string)
+	writeProperty(text string)
+	writeSymbol(text string, symbol compiler.Symbol)
+	writeLine(force ...bool)
```
rbuckton: a slice of bools seems overkill for an optional parameter. I'd just either make the parameter mandatory or add a second method.

internal/compiler/printer/textwriter.go
```diff
@@ -0,0 +1,190 @@
...
+func (w *textWriter) writeSymbol(text string, symbol compiler.Symbol) {
+	w.write(text)
+}
+
+func (w *textWriter) writeTrailingSemicolon(text string) {
+	w.write(text)
+}
+
+func NewTextWriter(newLine string) EmitTextWriter {
+	return &textWriter{newLine: newLine, lineStart: true}
```
jakebailey: I'd just write:

```go
var w textWriter
w.clear()
return &w
```

To keep these in sync.

internal/compiler/printer/textwriter.go
weswigham: > Could this just be

Sure - I had that in an earlier version, but thought that maybe the string builder would reuse the backing buffer between resets. I just looked and it doesn't, so I suppose there isn't a reason not to.

> though not set in newTextWriter?

Good catch - corrected. Doesn't actually change anything at the default 0 indentation (at least not when we don't try to emit a redundant linebreak at the start of a string anywhere), but good for consistency.

internal/compiler/printer/emittextwriter.go
weswigham: Hm, making it mandatory seems rough since it's only used rarely, so extra method it is, I guess.

internal/compiler/printer/emittextwriter.go
```diff
@@ -0,0 +1,33 @@
...
+	writeKeyword(text string)
+	writeOperator(text string)
+	writePunctuation(text string)
+	writeSpace(text string)
+	writeStringLiteral(text string)
+	writeParameter(text string)
+	writeProperty(text string)
+	writeSymbol(text string, symbol compiler.Symbol)
+	writeLine()
+	writeLineForce(force bool)
```
rbuckton, line 19: Do we still need the argument?

weswigham: Yea - all the usages of the `force` param pass in a variable, not a literal, and since we don't have ternaries switching the callee between `writeLine` and `writeLineRaw` at every use is gnarly.

jakebailey:
I will say that I'm not like mega excited about the massive interface, but, it's just what we already had in TS, so...

weswigham:
> Until we shift stuff around, it may be the case that this needs to move into the compiler package or something? I am not sure what the next stage of usage here is

Yeah, my intent is that we were already planning on shifting the definition of `Symbol` into a common type definition package (just like the AST), so once that's done the dependency will be fixed - just didn't need to do that as part of this just quite yet.

jakebailey:
I was hoping the AST move would happen along with the code gen, but I've more or less paused that to work on other stuff to unblock people, so maybe that just needs to be its own PR to move the existing AST types out first.

PR #43, opened 2024-10-31
isIdentifierPart should not allow colon or dash

sandersn:
tsgo incorrectly lexes non-ascii identifiers followed by colon as a single identifier -- in non-TS files only. That is, 'x:' instead of 'x' ':' (where x is non-ascii). This is incorrect in some, probably most, cases. The place I noticed it was an object literal:

```ts
f({ 県: '大阪府' });
```

That's because `isIdentifierPart` allows colon in identifiers for JSX when languageVariant is provided.
However, in tsc, languageVariant is optional and is only provided in 3 calls from services. All other calls disallow colon, even in JSX files. In tsgo, languageVariant is always provided, and non-TS files are treated as JSX.

isIdentifierPart should not allow colons and dashes, and the calls from services should check for JSX another way. It's likely that the checks are unneeded in the first place.

I haven't checked in a test with this PR. A type or symbol baseline will easily display it when we have those, and the test case I provided above shows it off.

### DIFF
--- a/internal/compiler/scanner.go
+++ b/internal/compiler/scanner.go
@@ -818,7 +818,7 @@ func (s *Scanner) ReScanSlashToken() SyntaxKind {
 		}
 		for {
 			ch, size := s.charAndSize()
-			if size == 0 || !isIdentifierPart(ch, s.languageVersion, s.languageVariant) {
+			if size == 0 || !isIdentifierPart(ch, s.languageVersion) {
 				break
 			}
 			s.pos += size
@@ -978,7 +978,7 @@ func (s *Scanner) scanIdentifier(prefixLength int) bool {
 		for {
 			s.pos += size
 			ch, size = s.charAndSize()
-			if !isIdentifierPart(ch, s.languageVersion, s.languageVariant) {
+			if !isIdentifierPart(ch, s.languageVersion) {
 				break
 			}
 		}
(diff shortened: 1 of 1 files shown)


sandersn:
I read the tsc code. Those uses in services are not buggy because they only pass in names that have already been parsed. But really, they've modified what should be a private scanner function to allow for JSX. The right fix is for there to be a utility in services that *maybe* calls the scanner and then adds JSX on top. Even more likely is that the checks are pointless because they're from already-parsed identifiers.

```
src/services/codefixes/fixAddMissingMember.ts:767:88 - error TS2554: Expected 2 arguments, but got 3.

767     return filter(targetProps, targetProp => isIdentifierText(targetProp.name, target, LanguageVariant.JSX) && !((targetProp.flags & SymbolFlags.Optional || getCheckFlags(targetProp) & CheckFlags.Partial) || seenNames.has(targetProp.escapedName)));
                                                                                           ~~~~~~~~~~~~~~~~~~~

src/services/completions.ts:5416:40 - error TS2554: Expected 2 arguments, but got 3.

5416     if (isIdentifierText(name, target, jsxIdentifierExpected ? LanguageVariant.JSX : LanguageVariant.Standard) || symbol.valueDeclaration && isPrivateIdentifierClassElementDeclaration(symbol.valueDeclaration)) {
                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/services/inlayHints.ts:355:75 - error TS2554: Expected 2 arguments, but got 3.

355         if (!isIdentifierText(name, getEmitScriptTarget(compilerOptions), getLanguageVariant(file.scriptKind))) {
```

PR #45, opened 2024-11-01
Port path tests (and fixes to path functions)

DanielRosenwasser:


internal/compiler/path_test.go
jakebailey, line 1: I wish these were table tests for better splitting/debugging, but it's okay

PR #46, opened 2024-11-01
Port error baseline generation

gabritto:


internal/compiler/path.go
```diff
@@ -410,3 +413,152 @@ func getRelativePathToDirectoryOrUrl(directoryPathOrUrl string, relativeOrAbsolu
...
+		}
+	}
+
+	return int32(cmp.Compare(len(aComponents), len(bComponents)))
+}
+
+// Gets the portion of a path following the last (non-terminal) separator (`/`).
+// Semantics align with NodeJS's `path.basename` except that we support URL's as well.
+// If the base name has any one of the provided extensions, it is removed.
+// ```
```
gabritto: I don't know the proper way of having code in doc comments in Go.

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+	NewLine:              harnessNewLine,
+}
+
+type TestFile struct {
+	unitName    string
+	content     string
+	fileOptions map[string]string
+}
+
+var diagnosticsLocationPrefix = regexp.MustCompile(`(?im)^(lib.*\.d\.ts)\(\d+,\d+\)`)
```
gabritto, line 40: Should I be using `makeRegexp` here (and everywhere else)?

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+		count := 1 + dupeCase[resultName]
+		dupeCase[resultName] = count
+		resultName = fmt.Sprintf("%s.dupe%d", resultName, count)
+	} else {
+		dupeCase[resultName] = 0
+	}
+	return resultName
+}
+
+func flattenDiagnosticMessage(d *compiler.Diagnostic, newLine string) string {
```
gabritto, line 275: For this and the function below, given how the existing TS code works, I thought it would be very inconvenient to pass a string builder for the function instead of having the function return a string, because the original code calls `flattenDiagnosticMessage()` and then does a bunch of transformations on the string before adding that to the result, so I created those wrappers.
I also couldn't create a generic wrapper because I think that's just not possible in Go, but maybe it is.

gabritto:
I tried writing a test but we don't have any file system mocking implemented yet, so it would be a lot of work. I'm working on porting that next, so hopefully then I can add a test for this.

internal/testutil/baseline/error_baseline.go
jakebailey: You're doing it right; just having globals like this is the way to go. The `makeRegexp` IMO is probably not worth it long-term.

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+	errorsReported = 0
+
+	// 'merge' the lines of each input file with any errors associated with it
+	dupeCase := map[string]int{}
+	nonEmptyFiles := compiler.Filter(inputFiles, func(f *TestFile) bool { return len(f.content) > 0 })
+	for _, inputFile := range nonEmptyFiles {
+		// Filter down to the errors in the file
+		fileErrors := compiler.Filter(diagnostics, func(e *compiler.Diagnostic) bool {
+			return e.File() != nil &&
+				compiler.ComparePaths(removeTestPathPrefixes(e.File().FileName(), false), removeTestPathPrefixes(inputFile.unitName, false), "", true) == compiler.ComparisonEqual
```
DanielRosenwasser: As a TODO after we get this PR in, maybe we can use `getErrorSummary` to avoid the walks for `globalErrors` plus each file (though I guess most tests only have 1 file anyway).

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+		})
+
+		// Header
+		outputLines.WriteString(newLine())
+		outputLines.WriteString("==== ")
+		outputLines.WriteString(removeTestPathPrefixes(inputFile.unitName, false))
+		outputLines.WriteString(" (")
+		outputLines.WriteString(fmt.Sprintf("%d errors) ====", len(fileErrors)))
```
DanielRosenwasser:
```suggestion
		fmt.Fprintf(outputLines,
			"%s==== %s (%d errors) ====,
			newline(),
			removeTestPathPrefixes(inputFile.unitName, false),
			len(fileErrors),
		)
```

internal/compiler/utilities.go
```diff
@@ -478,7 +478,7 @@ func formatMessage(message *diagnostics.Message, args ...any) string {
 	return text
 }
 
-func filter[T any](slice []T, predicate func(T) bool) []T {
+func Filter[T any](slice []T, predicate func(T) bool) []T {
```
jakebailey: At some point we should move all of these helpers out of this package; it's odd for this to be the place for them long-term.

internal/compiler/utilities.go
```diff
@@ -573,6 +573,12 @@ func every[T any](array []T, predicate func(T) bool) bool {
 	return true
 }
 
+func ForEach[T any](array []T, action func(T)) {
+	for _, value := range array {
+		action(value)
+	}
+}
```
jakebailey: I haven't gotten to the code which uses this yet, but I suspect that it should really just be written as a loop, not functionally.

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+
+	"github.com/microsoft/typescript-go/internal/compiler"
+)
+
+// IO
+const harnessNewLine = "\r\n"
+
+var lineDelimiter = regexp.MustCompile("\r?\n")
+var nonWhitespace = regexp.MustCompile(`\S`)
+var tsExtension = regexp.MustCompile(`\.tsx?$`)
```
jakebailey: Style nit; these blocks of vars can just be combined into one.

```suggestion
var (
	lineDelimiter = regexp.MustCompile("\r?\n")
	nonWhitespace = regexp.MustCompile(`\S`)
	tsExtension   = regexp.MustCompile(`\.tsx?$`)
)
```

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+type TestFile struct {
+	unitName    string
+	content     string
+	fileOptions map[string]string
+}
+
+var diagnosticsLocationPrefix = regexp.MustCompile(`(?im)^(lib.*\.d\.ts)\(\d+,\d+\)`)
+var diagnosticsLocationPattern = regexp.MustCompile(`(?i)(lib.*\.d\.ts):\d+:\d+`)
+
+func DoErrorBaseline(t *testing.T, baselinePath string, inputFiles []*TestFile, errors []*compiler.Diagnostic, pretty bool) {
```
jakebailey:
```suggestion
func DoErrorBaseline(t assert.TestingT, baselinePath string, inputFiles []*TestFile, errors []*compiler.Diagnostic, pretty bool) {
```

This package a one place I'd avoid referencing `testing`; it's very possible that we'd want to run this from a benchmark test (e.g. we could very reasonably convert future compiler tests into benchmarks by just calling the same code from a benchmark func rather than a test func).

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	return strings.Join(outputLines, "")
+}
+
+func iterateErrorBaseline(t *testing.T, inputFiles []*TestFile, inputDiagnostics []*compiler.Diagnostic, pretty bool) []string {
+	t.Helper()
+	diagnostics := slices.Clone(inputDiagnostics)
+	slices.SortFunc(diagnostics, compiler.CompareDiagnostics)
+
+	outputLines := strings.Builder{}
```
jakebailey: Style nit, but I recommend writing:

```suggestion
	var outputLines strings.Builder
```

You did that above, but I recommend this in general just because it means you don't need to remember that the type is a struct.

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+	outputErrorText := func(error *compiler.Diagnostic) {
+		message := flattenDiagnosticMessage(error, harnessNewLine)
+
+		errLines := strings.Split(removeTestPathPrefixes(message, false), "\n")
+		errLines = compiler.Mapf(errLines, func(s string) string {
+			lastRune, _ := utf8.DecodeLastRuneInString(s)
+			if lastRune == '\r' {
+				return s[:len(s)-1]
+			}
+			return s
```
jakebailey: This could just be:
```suggestion
			return strings.TrimSuffix(s, "\r")
```

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+		})
+		errLines = compiler.Filter(errLines, func(s string) bool { return len(s) > 0 })
+		errLines = compiler.Mapf(errLines, func(s string) string {
+			return fmt.Sprintf("!!! %s TS%d: %s", error.Category().String(), error.Code(), s)
+		})
```
jakebailey: This section strikes me as "too functional"; this could just be a for loop, like:

```go
var errLines []string

for _, line := range strings.Split(removeTestPathPrefixes(message, false), "\n") {
	line = strings.TrimSuffix(line, "\r")
	if len(line) < 0 {
		continue
	}
	out := fmt.Sprintf("!!! %s TS%d: %s", error.Category().String(), error.Code(), line)
	errLines = append(errLines, out)
}
```

internal/testutil/baseline/error_baseline.go
```diff
@@ -0,0 +1,280 @@
...
+
+	topDiagnostics := minimalDiagnosticsToString(diagnostics, pretty)
+	topDiagnostics = removeTestPathPrefixes(topDiagnostics, false)
+	topDiagnostics = diagnosticsLocationPrefix.ReplaceAllString(topDiagnostics, "$1(--,--)")
+
+	result = append(result, topDiagnostics+harnessNewLine+harnessNewLine)
+
+	// Report global errors
+	globalErrors := compiler.Filter(diagnostics, func(d *compiler.Diagnostic) bool { return d.File() == nil })
+	compiler.ForEach(globalErrors, outputErrorText)
```
jakebailey:
```suggestion
	for _, d := range diagnostics {
		if d.File() == nil {
			outputErrorText(d)
		}
	}
```

To me would be clearer; I don't think we need to be filtering (and creating extra arrays) or introducing a `ForEach`.

jakebailey:
Did a first pass over things (just for Go code stuff)

internal/testutil/baseline/error_baseline.go
gabritto: I was using `assert.TestingT`, but then I changed it because that type doesn't have a `.Helper()` method. Should I use `testing.TB` instead?

jakebailey: Ah, yeah `testing.TB` is fine.

internal/compiler/checker.go
```diff
@@ -2201,7 +2201,7 @@ func (c *Checker) addDuplicateDeclarationError(node *Node, message *diagnostics.
 		leadingMessage := createDiagnosticForNode(adjustedNode, diagnostics.X_0_was_also_declared_here, symbolName)
 		followOnMessage := createDiagnosticForNode(adjustedNode, diagnostics.X_and_here)
 		if len(err.relatedInformation) >= 5 || some(err.relatedInformation, func(d *Diagnostic) bool {
-			return compareDiagnostics(d, followOnMessage) == 0 || compareDiagnostics(d, leadingMessage) == 0
+			return CompareDiagnostics(d, followOnMessage) == 0 || CompareDiagnostics(d, leadingMessage) == 0
```
jakebailey, line 2831: We really do need to start splitting this package a part a little; exporting everything from the compiler doesn't seem like the best time... 😄

jakebailey: Oh, I'm just repeating myself from another comment

internal/compiler/path.go
```diff
@@ -12,6 +13,9 @@ type Path string
 const directorySeparator = '/'
 const urlSchemeSeparator = "://"
 
+// check path for these segments: ”, '.'. '..'
```
jakebailey: This is an annoying go gotcha, but Go comments get reformatted without `"` because they're supposed to be special readable things. I don't like it, but for this comment you're going to have to write these out in an indented code block to stop gofmt from screwing with the quotes.

internal/compiler/path.go
jakebailey: You can just indent the code one extra level; the code fences unfortunately won't work.

internal/testutil/baseline/util.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func sanitizeTestFilePath(name string) string {
+	path := testPathCharacters.ReplaceAllString(name, "_")
+	path = compiler.NormalizeSlashes(path)
+	path = testPathDotDot.ReplaceAllString(path, "__dotdot/")
+	path = string(compiler.ToPath(path, "", canonicalizeForHarness))
+	if strings.HasPrefix(path, "/") {
+		return path[1:]
+	}
```
jakebailey:
```suggestion
	path = strings.TrimPrefix(path, "/")
```

internal/testutil/baseline/util.go
```diff
@@ -0,0 +1,55 @@
...
+	fileName := compiler.GetBaseFileName(filePath, nil, false)
+	return strings.HasPrefix(fileName, "lib.") && strings.HasSuffix(fileName, compiler.ExtensionDts)
+}
+
+func isBuiltFile(filePath string) bool {
+	return strings.HasPrefix(filePath, libFolder) || strings.HasPrefix(filePath, compiler.EnsureTrailingDirectorySeparator(builtFolder))
+}
+
+func isTsConfigFile(path string) bool {
+	return strings.Contains(path, "tsconfig") && strings.Contains(path, "json")
```
jakebailey, line 44: Shocking that this is actually how the original code does it too 😄

internal/testutil/baseline/util.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+var testPathPrefix = regexp.MustCompile(`(?:(file:\/{3})|\/)\.(?:ts|lib|src)\/`)
+var testPathCharacters = regexp.MustCompile(`[\^<>:"|?*%]`)
+var testPathDotDot = regexp.MustCompile(`\.\.\/`)
+
+// This is done so tests work on windows _and_ linux
+var canonicalizeForHarness = strings.ToLower
+
+var libFolder = "built/local/"
```
jakebailey, line 18: We'll have to see where this ends up, given we don't have a real place for this at the moment.

internal/compiler/checker.go
gabritto: Where would this function go? It's neither `path` nor `sliceutils`, but compiler `utilities`. There's a few others like this that I had to export from the compiler package, like `ComputeLineStarts` from the scanner.

jakebailey: This one's probably fine where it is, since a Diagnostic is specific to the compiler, kinda, but probably we'd actually come up with a new package just for that, or put it in the existing diagnostics package. This was just the first one in the diff, I think the other helpers are more easily movable 😄

internal/compiler/path.go
```diff
@@ -419,3 +422,152 @@ func getRelativePathToDirectoryOrUrl(directoryPathOrUrl string, relativeOrAbsolu
...
+// getBaseFileName("file:///path/to/file.ext") == "file.ext"
+// getBaseFileName("file:///path/to/") == "to"
+// getBaseFileName("file:///") == ""
+// getBaseFileName("file://") == ""
+// getBaseFileName("/path/to/file.ext", ".ext", true) == "file"
+// getBaseFileName("/path/to/file.js", ".ext", true) == "file.js"
+// getBaseFileName("/path/to/file.js", [".ext", ".js"], true) == "file"
+// getBaseFileName("/path/to/file.ext", ".EXT", false) == "file.ext"
+// ```
+func GetBaseFileName(path string, extensions []string, ignoreCase bool) string {
```
andrewbranch: Literally only one place uses the `extensions` parameter—let’s remove it. Whenever we’re porting overloaded functions, I think we should be skeptical of the value that rarely-used signatures add. In this case, it just tacks on completely unrelated functionality that’s trivial for the caller to perform separately.

internal/compiler/utilities.go
```diff
@@ -510,6 +511,158 @@ func formatMessage(message *diagnostics.Message, args ...any) string {
...
 
+func filter[T any](slice []T, predicate func(T) bool) []T {
+	result, _ := sameFilter(slice, predicate)
+	return result
+}
+
+func sameFilter[T any](slice []T, predicate func(T) bool) ([]T, bool) {
+	for i, value := range slice {
+		if !predicate(value) {
+			result := slice[0:i]
```
jakebailey: This all looks like the result of a bad merge

gabritto: This is what I get for trying to use vscode's merge editor.

internal/compiler/error_reporting.go
```diff
@@ -341,3 +341,22 @@ func prettyPathForFileError(file *SourceFile, fileErrors []*Diagnostic, formatOp
...
+
+func WriteFormatDiagnostic(output *strings.Builder, diagnostic *Diagnostic, formatOpts *DiagnosticsFormattingOptions) {
+	if diagnostic.file != nil {
+		line, character := GetLineAndCharacterOfPosition(diagnostic.file, diagnostic.loc.Pos())
+		fileName := diagnostic.file.fileName
+		relativeFileName := tspath.ConvertToRelativePath(fileName, formatOpts.CurrentDirectory, formatOpts.GetCanonicalFileName)
+		output.WriteString(fmt.Sprintf("%s(%d,%d): ", relativeFileName, line+1, character+1))
+	}
+
+	output.WriteString(fmt.Sprintf("%s TS%d: ", diagnostic.Category().String(), diagnostic.Code()))
```
jakebailey:
```suggestion
	fmt.Fprintf(output, "%s TS%d: ", diagnostic.Category().String(), diagnostic.Code())
```

internal/compiler/error_reporting.go
```diff
@@ -341,3 +341,22 @@ func prettyPathForFileError(file *SourceFile, fileErrors []*Diagnostic, formatOp
...
+		WriteFormatDiagnostic(output, diagnostic, formatOpts)
+	}
+}
+
+func WriteFormatDiagnostic(output *strings.Builder, diagnostic *Diagnostic, formatOpts *DiagnosticsFormattingOptions) {
+	if diagnostic.file != nil {
+		line, character := GetLineAndCharacterOfPosition(diagnostic.file, diagnostic.loc.Pos())
+		fileName := diagnostic.file.fileName
+		relativeFileName := tspath.ConvertToRelativePath(fileName, formatOpts.CurrentDirectory, formatOpts.GetCanonicalFileName)
+		output.WriteString(fmt.Sprintf("%s(%d,%d): ", relativeFileName, line+1, character+1))
```
jakebailey:
```suggestion
		fmt.Fprintf(output, "%s(%d,%d): ", relativeFileName, line+1, character+1)
```

internal/utils/utils.go
jakebailey, line 1: This dir got moved so the PR needs to be updated.

internal/core/core.go
```diff
@@ -225,3 +225,36 @@ func EquateStringCaseInsensitive(a, b string) bool {
...
+
+type Comparison = int32
+
+const (
+	ComparisonLessThan    Comparison = -1
+	ComparisonEqual       Comparison = 0
+	ComparisonGreaterThan Comparison = 1
+)
+
+func CompareStringsCaseInsensitive(a string, b string) Comparison {
```
jakebailey, line 244: Seems like @andrewbranch's PR also introduces some of this; we'll have to disambiguate them.

internal/core/core.go
```diff
@@ -225,3 +225,36 @@ func EquateStringCaseInsensitive(a, b string) bool {
...
 }
+
+func GetStringEqualityComparer(ignoreCase bool) func(a, b string) bool {
+	if ignoreCase {
+		return EquateStringCaseInsensitive
+	}
+	return EquateStringCaseSensitive
+}
+
+type Comparison = int32
```
jakebailey: Is this intentionally an alias and not a named type?

internal/compiler/error_reporting.go
```diff
@@ -342,3 +342,22 @@ func prettyPathForFileError(file *SourceFile, fileErrors []*Diagnostic, formatOp
...
+
+func WriteFormatDiagnostic(output *strings.Builder, diagnostic *Diagnostic, formatOpts *DiagnosticsFormattingOptions) {
+	if diagnostic.file != nil {
+		line, character := GetLineAndCharacterOfPosition(diagnostic.file, diagnostic.loc.Pos())
+		fileName := diagnostic.file.fileName
+		relativeFileName := tspath.ConvertToRelativePath(fileName, formatOpts.CurrentDirectory, formatOpts.GetCanonicalFileName)
+		fmt.Fprintf(output, "%s(%d,%d): ", relativeFileName, line+1, character+1)
+	}
+
+	fmt.Fprintf(output, "%s TS%d: ", diagnostic.Category().String(), diagnostic.Code())
```
jakebailey, line 360: No need to change this, but `%s` will actually call `String()` if present.

internal/core/core.go
jakebailey: Other PRs use `int` for this and I think the stdlib comparisons also require int, so I would probably go for `int` too.

internal/core/core.go
andrewbranch: Let’s get this in first and I’ll fix conflicts.

internal/core/core.go
andrewbranch: I think a named type would be a pain here, for interop with the standard library. An alias seems fine to me, but not really necessary. It should match the standard lib comparison return types, though.

PR #47, opened 2024-11-01
Add skipTrivia and scanConflictMarkerTrivia

sandersn:
Plus supporting functions.

### DIFF
--- a/internal/compiler/binder.go
+++ b/internal/compiler/binder.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/utils"
 )
 
 type ContainerFlags int32
@@ -990,7 +991,7 @@ func (b *Binder) hasExportDeclarations(node *Node) bool {
 			statements = body.AsModuleBlock().statements
 		}
 	}
-	return some(statements, func(s *Node) bool {
+	return utils.Some(statements, func(s *Node) bool {
 		return isExportDeclaration(s) || isExportAssignment(s)
 	})
 }
@@ -1719,7 +1720,7 @@ func (b *Binder) checkUnreachable(node *Node) bool {
 				//   On the other side we do want to report errors on non-initialized 'lets' because of TDZ
 				isError := unreachableCodeIsError(b.options) && node.flags&NodeFlagsAmbient == 0 && (!isVariableStatement(node) ||
 					getCombinedNodeFlags(node.AsVariableStatement().declarationList)&NodeFlagsBlockScoped != 0 ||
-					some(node.AsVariableStatement().declarationList.AsVariableDeclarationList().declarations, func(d *Node) bool {
+					utils.Some(node.AsVariableStatement().declarationList.AsVariableDeclarationList().declarations, func(d *Node) bool {
 						return d.AsVariableDeclaration().initializer != nil
 					}))
 				b.errorOnEachUnreachableRange(node, isError)
@@ -1762,7 +1763,7 @@ func (b *Binder) errorOnEachUnreachableRange(node *Node, isError bool) {
 func (b *Binder) isExecutableStatement(s *Node) bool {
 	// Don't remove statements that can validly be used before they appear.
 	return !isFunctionDeclaration(s) && !b.isPurelyTypeDeclaration(s) && !(isVariableStatement(s) && getCombinedNodeFlags(s)&NodeFlagsBlockScoped == 0 &&
-		some(s.AsVariableStatement().declarationList.AsVariableDeclarationList().declarations, func(d *Node) bool {
+		utils.Some(s.AsVariableStatement().declarationList.AsVariableDeclarationList().declarations, func(d *Node) bool {
 			return d.AsVariableDeclaration().initializer == nil
 		}))
 }
@@ -2082,7 +2083,7 @@ func (b *Binder) bindSwitchStatement(node *Node) {
 	b.preSwitchCaseFlow = b.currentFlow
 	b.bind(stmt.caseBlock)
 	b.addAntecedent(postSwitchLabel, b.currentFlow)
-	hasDefault := some(stmt.caseBlock.AsCaseBlock().clauses, func(c *Node) bool {
+	hasDefault := utils.Some(stmt.caseBlock.AsCaseBlock().clauses, func(c *Node) bool {
 		return c.kind == SyntaxKindDefaultClause
 	})
 	if !hasDefault {
@@ -2524,7 +2525,7 @@ func (b *Binder) addDeclarationToSymbol(symbol *Symbol, node *Node, symbolFlags 
 	if symbol.declarations == nil {
 		symbol.declarations = b.newSingleDeclaration(node)
 	} else {
-		symbol.declarations = appendIfUnique(symbol.declarations, node)
+		symbol.declarations = utils.AppendIfUnique(symbol.declarations, node)
 	}
 	// On merge of const enum module with class or function, reset const enum only flag (namespaces will already recalculate)
 	if symbol.constEnumOnlyModule && symbol.flags&(SymbolFlagsFunction|SymbolFlagsClass|SymbolFlagsRegularEnum) != 0 {
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/utils"
 )
 
 // CheckMode
@@ -502,7 +503,7 @@ func (c *Checker) mergeModuleAugmentation(moduleName *Node) {
 			// the pattern ('*.foo'), so that 'getMergedSymbol()' on a.foo gives you
 			// all the exports both from the pattern and from the augmentation, but
 			// 'getMergedSymbol()' on *.foo only gives you exports from *.foo.
-			if some(c.patternAmbientModules, func(module PatternAmbientModule) bool {
+			if utils.Some(c.patternAmbientModules, func(module PatternAmbientModule) bool {
 				return mainModule == module.symbol
 			}) {
 				merged := c.mergeSymbol(moduleAugmentation.symbol, mainModule, true /*unidirectional*/)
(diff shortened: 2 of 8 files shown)


internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,178 @@ func TokenToString(token SyntaxKind) string {
...
+		return true
+	case '#':
+		// Only if its the beginning can we have #! trivia
+		return pos == 0
+	default:
+		return ch > maxAsciiCharacter
+	}
+}
+
+func skipTrivia(text string, pos int, stopAfterLineBreak bool, stopAtComments bool, inJSDoc bool) int {
```
DanielRosenwasser: Since every current caller sets the last 3 to `false, false, false`, I would suggest creating something like

```go
type SkipTriviaOptions struct {
	stopAfterLineBreak bool
	stopAtComments bool
	inJSDoc bool
}
```

and then people can always just pass the default-initialized value in.

jakebailey: And make it `skipTriviaOptions`, since this is an unexported function.

sandersn: Is that the usual pattern or is it more common to create a `skipTriviaFull` with 5 (or 3, in the form of options) parameters and `skipTrivia` with the usual 2?

jakebailey: In this case, I think an options struct is the most reasonable. If the only difference between two functions is how many args they have, then there's probably not good unique names they should have (full or not full is not a very good distinction). The options bag at least lets you specify the options you need.

ahejlsberg: Regarding naming of functions with optional parameters, I'm using the pattern of `getMyCoffee` for the short form and `getMyCoffeeEx` for the extended form with every parameter. I would just do that here as well.

jakebailey: I would really avoid doing that except for simple cases; named parameters are very commonly written in Go with options structs like this. They should not be any less efficient on the whole thanks to inlining or pass-by-value semantics (or even fine when passed as a non-escaping address).

internal/compiler/scanner.go
```diff
@@ -1707,7 +1707,13 @@ func couldStartTrivia(text string, pos int) bool {
 	}
 }
 
-func skipTrivia(text string, pos int, stopAfterLineBreak bool, stopAtComments bool, inJSDoc bool) int {
+type SkipTriviaOptions struct {
```
jakebailey: `skipTrivia` is unexported, so this should be called `skipTriviaOptions`. Types are not uppercased in Go.

(Also, this type wouldn't be usable as an exported type because all of its fields are unexported)

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,184 @@ func TokenToString(token SyntaxKind) string {
...
-func skipTrivia(sourceText string, pos int) int {
-	s := NewScanner()
-	s.text = sourceText
-	s.pos = pos
-	s.Scan()
-	return s.tokenStart
+func couldStartTrivia(text string, pos int) bool {
+	// Keep in sync with skipTrivia
+	ch := text[pos]
+	switch ch {
```
jakebailey: Tip, you can write:

```suggestion
	switch ch := text[pos] {
```

internal/compiler/scanner.go
sandersn: 👍🏼 
this is going to take a long time to learn

internal/compiler/scanner.go
jakebailey: Oops, meant `ch := text[pos]; ch`

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,183 @@ func TokenToString(token SyntaxKind) string {
...
+		ch := text[pos]
+		switch ch {
+		case '\r':
+			if text[pos+1] == '\n' {
+				pos++
+			}
+			fallthrough
+		case '\n':
+			pos++
+			if options.stopAfterLineBreak {
```
jakebailey, line 1738: All references to `options` need to be guarded with a nil check, or the function needs to reassign `options = &skipTriviaOptions{}` at the top to provide a default.

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,183 @@ func TokenToString(token SyntaxKind) string {
...
+		ch := text[pos]
+		switch ch {
+		case '\r':
+			if text[pos+1] == '\n' {
+				pos++
+			}
+			fallthrough
+		case '\n':
+			pos++
+			if options != nil && options.stopAfterLineBreak {
```
jakebailey: Shouldn't this be:

```suggestion
			if options == nil || options.stopAfterLineBreak {
```

and for the others? As in, default? I would honestly have just done the defaulting at the top instead of this; it is not expensive at runtime.

jakebailey: Seems like no, it is supposed to be `options?.stopAfterLineBreak` which is what you wrote.

sandersn: Defaulting at the top is a great idea.

The code in this commit is correct, though, because if you don't pass options, they should all default to false, not true.

internal/compiler/scanner.go
DanielRosenwasser: Is it better here to always take a pointer to an options bag, or can it just always be passed by value to avoid this sort of thing?

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,186 @@ func TokenToString(token SyntaxKind) string {
...
+	return false
+}
+
+func scanConflictMarkerTrivia(text string, pos int, error func(diag *diagnostics.Message, pos int, len int, args ...any)) int {
+	if error != nil {
+		error(diagnostics.Merge_conflict_marker_encountered, pos, mergeConflictMarkerLength)
+	}
+
+	ch := text[pos]
+	len := len(text)
```
jakebailey: FWIW I would avoid shadowing builtin names like this; this prevents any code below from using the len function.

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,186 @@ func TokenToString(token SyntaxKind) string {
...
+	}
+}
+
+type skipTriviaOptions struct {
+	stopAfterLineBreak bool
+	stopAtComments     bool
+	inJSDoc            bool
+}
+
+func skipTrivia(text string, pos int, options *skipTriviaOptions) int {
```
jakebailey: I know I suggested the options pattern here, but seeing just how many callers are passing in `nil` directly makes me think that it actually is worth splitting into two functions, leaving `skipTrivia` as a wrapper which calls `skipTriviaEx`.

internal/compiler/scanner.go
jakebailey: I've flip flopped on this case... every caller passes in nil still

jakebailey: (but, the options bag is better, just, make a wrapper that passes nil)

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,189 @@ func TokenToString(token SyntaxKind) string {
...
+				continue
+			}
+		case '*':
+			if canConsumeStar {
+				pos++
+				canConsumeStar = false
+				continue
+			}
+		default:
+			if ch > maxAsciiCharacter && (isWhiteSpaceLike(rune(ch))) {
```
jakebailey:
```suggestion
			if ch > maxAsciiCharacter && isWhiteSpaceLike(rune(ch)) {
```

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,189 @@ func TokenToString(token SyntaxKind) string {
...
+			pos++
+			continue
+		case '/':
+			if options.stopAtComments {
+				break
+			}
+			if text[pos+1] == '/' {
+				pos += 2
+				for pos < len(text) {
+					if isLineBreak(rune(text[pos])) {
```
jakebailey: All of these `rune(text[pos])` make me nervous; maybe it doesn't matter for this case but this is how you accidentally chop up a unicode character.

internal/compiler/scanner.go
sandersn: booo gofmt

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,189 @@ func TokenToString(token SyntaxKind) string {
...
+var mergeConflictMarkerLength = len("<<<<<<<")
+var maxAsciiCharacter byte = 127
+
+func isConflictMarkerTrivia(text string, pos int) bool {
+	if pos < 0 {
+		panic("pos < 0")
+	}
+
+	// Conflict markers must be at the start of a line.
+	if pos == 0 || isLineBreak(rune(text[pos-1])) {
```
sandersn: A maximally correct isLineBreak check would cover both `\r \n` and LineSeparator and ParagraphSeparator:`
```suggestion
	prev, _ := utf8.DecodeLastRuneInString(text[:pos - 2])
	if pos == 0 || isLineBreak(prev) || isLineBreak(rune(text[pos-1])) {
```
I don't know if that's worth supporting.

internal/compiler/scanner.go
sandersn: OK, I posted a commit that is maximally fancy in that I switched to utf8.DecodeInString wherever a check against utf8 characters might happen, and kept `text[pos]` where it can't.

internal/compiler/scanner.go
```diff
@@ -1657,13 +1690,193 @@ func TokenToString(token SyntaxKind) string {
...
+			}
+		}
+		return pos
+	}
+}
+
+// All conflict markers consist of the same character repeated seven times.  If it is
+// a <<<<<<< or >>>>>>> marker then it is also followed by a space.
+var mergeConflictMarkerLength = len("<<<<<<<")
+var maxAsciiCharacter byte = 127
```
jakebailey, line 1806: If you drop `byte` here, this const wont' have to be converted a few lines above.

sandersn: unfortunately, it's also used in couldStartTrivia, which is comparing bytes, not runes.

sandersn:
Yeah, I don't think we'll know whether this really works until there are some tests that depend on trivia. I think we'll have to shake bugs out at that point.

internal/compiler/scanner.go
jakebailey: Constants don't have to be typed; leaving it off makes it work anywhere.

PR #50, opened 2024-11-04
Type relations and property access expression checking

ahejlsberg:
This PR ports/implements the following:

* Resolution of class and interface base types
* Generalized `KeyBuilder`
* Simplification of type object hierarchy
* Accessors for common type properties
* Checking of property access expressions
* Type relations for object types, type parameters, union types, and intersection types

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -72,7 +72,7 @@ func main() {
 	runtime.GC()
 	runtime.GC()
 	runtime.ReadMemStats(&memStats)
-	if !quiet {
+	if !quiet && len(diagnostics) != 0 {
 		if pretty {
 			var getCanonicalFileName func(path string) string
 			if useCaseSensitiveFileNames {


internal/compiler/printer.go
```diff
@@ -37,25 +37,31 @@ func (c *Checker) symbolToString(s *Symbol) string {
 }
 
 func (c *Checker) typeToString(t *Type) string {
-	p := c.newPrinter()
+	return c.typeToStringEx(t, nil, TypeFormatFlagsNone)
+}
+
+func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *Node, flags TypeFormatFlags) string {
+	_ = enclosingDeclaration
```
jakebailey: Params can be unused so this isn't strictly required (not sure if you're working around the `unusedparameter` analyzer or not)

internal/compiler/checker.go
```diff
@@ -237,13 +265,21 @@ type Checker struct {
...
 	lastGetCombinedNodeFlagsNode       *Node
 	lastGetCombinedNodeFlagsResult     NodeFlags
 	lastGetCombinedModifierFlagsNode   *Node
 	lastGetCombinedModifierFlagsResult ModifierFlags
+	subtypeRelation                    *Relation
+	strictSubtypeRelation              *Relation
+	assignableRelation                 *Relation
+	comparableRelation                 *Relation
+	identityRelation                   *Relation
+	enumRelation                       *Relation
```
jakebailey, line 282: Given these are all initialized anyway, you could probably just make these `Relation` instead of `*Relation`? Unless you're optimizing for having to reference these as values often?

internal/compiler/checker.go
```diff
@@ -3144,154 +4180,274 @@ func (c *Checker) getDeclaredTypeOfClassOrInterface(symbol *Symbol) *Type {
...
+						return false
+					}
+				}
+			}
+		}
+	}
+	return true
+}
+
+type KeyBuilder struct {
```
jakebailey, line 4217: Nit, but we'll probably eventually unexport this given it's only ever used internally (same with the methods).

internal/compiler/checker.go
```diff
@@ -4546,40 +6071,39 @@ func (c *Checker) isThisless(symbol *Symbol) bool {
...
+				sig = c.createSignatureInstantiation(baseSig, c.fillMissingTypeArguments(typeArguments, baseSig.typeParameters, minTypeArgumentCount))
+			} else {
+				sig = c.cloneSignature(baseSig)
+			}
+			sig.typeParameters = classType.AsInterfaceType().LocalTypeParameters()
+			sig.resolvedReturnType = classType
+			if isAbstract {
+				sig.flags |= SignatureFlagsAbstract
+			} else {
+				sig.flags &= ^SignatureFlagsAbstract
```
jakebailey: FWIW you can use `&^=` instead of `&= ^`; Go has that as an operator.

(The most recent code output by ts-to-go does that transform automatically)

internal/compiler/checker.go
jakebailey: (that being said we have like, so many types we've declared as exported in this package that definitely wouldn't be if we weren't just copying code, so, nevermind)

internal/compiler/checker.go
ahejlsberg: We do a lot of checking of which relation we're holding a reference to, e.g. `relation == c.subtypeRelation`, and don't like having to use `&` there.

PR #52, opened 2024-11-04
Prep work for emitter port

rbuckton:
There are a number of complexities in the current emitter that are a result of some of our AST nodes serving multiple purposes. The `EmitHint` enum, which is plumbed throughout the emitter and transformers, was originally introduced to easily differentiate between expression and non-expression _Identifier_ nodes without requiring a walk of `.parent` pointers, which is unreliable when transforming or synthesizing a subtree for the purposes of emit, diagnostics, or generating quick-fixes. 

Given that the native compiler's AST already has a number of differences from our TS-based compiler, now would be the best time to address other "breaking" changes to the AST structure.

This PR seeks to introduce the following changes to our AST nodes:

- Replace `Identifier` with `IdentifierName`, `IdentifierReference`, `BindingIdentifier` and `LabelIdentifier`
- Replace the current `Block` node with `Block` and `FunctionBody`

In addition, due to the limitations on static typing with the current direction of our AST, this PR also introduces runtime assertions into the various AST factory methods so as to catch incorrect usages early during development and testing.

# Splitting up `Identifier`

In the ECMA-262 specification, there are four distinct forms of "identifiers" (not including `PrivateIdentifier`):

- [`IdentifierName`](https://tc39.es/ecma262/#prod-IdentifierName) - The lexical token that represents any potential identifier (including potential reserved and contextual keywords) and is used in the grammar to represent property/member names (i.e., `obj.x`, `get x() {}`, etc.) as well as aliased imports (`import { x as y } from "..."`), export aliases (`export { y as x }`), and re-exports (`export { x } from "..."`) 
- [`IdentifierReference`](https://tc39.es/ecma262/#prod-IdentifierReference) - Indicates an identifier that is a reference to a declared name. In JS, an `IdentifierReference` is always an expression. In TS, an `IdentifierReference` is the root of an `EntityName` (so long as it is not part of an `import()` type), and thus may also refer to a type parameter. As they are references, these are often rewritten during emit to handle reading or writing to imports, exports, namespace variables, etc.
- [`BindingIdentifier`](https://tc39.es/ecma262/#prod-BindingIdentifier) - Indicates an identifier that declares a new binding, such as a variable (`let x`), a binding pattern (`let { x }`), an import `import { x } from "..."`, a function `function x() {}`, a class `class x {}`, a parameter `function f(x) {}`, etc.
- [`LabelIdentifier`](https://tc39.es/ecma262/#prod-LabelIdentifier) - Indicates an identifier used in a `LabeledStatement`, `break`, or `continue`.

The use cases, semantics, and emit for each of these kinds of identifier differ wildly and we often have to use fairly complex functions like `isExpressionNode` or enums like `EmitHint` to disambiguate them today during bind, check, and emit. By breaking out `Identifier` to its individual use cases, we can avoid some overly complex and unnecessary work in the process.

Not all of these identifiers are expressions, and not all of them need flow nodes. However, for the time being I've opted to limit this change to only introducing `SyntaxKind` entries for each individual identifier kind, while reusing the `Identifier` struct as-is. This is to keep most of the existing code paths simple that would today depend on either `node.AsIdentifier()` or the `isIdentifier()` node test function. The assertions added to each of the node factory functions will help us to identify the cases where we need more specificity as we continue to port existing compiler functionality to Go.

# Splitting up `Block`

Much like `Identifier`, ECMA-262 has several forms of `Block` that we generally conflate in our AST:

- [`BlockStatement`](https://tc39.es/ecma262/#prod-BlockStatement) - Any `Block` used in a statement position
- [`TryStatement`](https://tc39.es/ecma262/#prod-TryStatement)
- [`Catch`](https://tc39.es/ecma262/#prod-Catch)
- [`FunctionBody`](https://tc39.es/ecma262/#prod-FunctionBody)
- [`AsyncFunctionBody`](https://tc39.es/ecma262/#prod-AsyncFunctionBody)
- [`GeneratorBody`](https://tc39.es/ecma262/#prod-GeneratorBody)
- [`AsyncGeneratorBody`](https://tc39.es/ecma262/#prod-AsyncGeneratorBody)
- [`ClassStaticBlockBody`](https://tc39.es/ecma262/#prod-ClassStaticBlockBody)

While we also have our own `ModuleBlock` which has its own AST node, despite behaving similar to a `Block`.

Due to this conflation, we regularly employ functions like `isFunctionBlock` to walk the `parent` of a `Block` to determine if it is the body of a function or `static {}` block.

In this PR, I propose carving off `FunctionBody` from `Block` in the same fashion as proposed for `Identifier`, by introducing a unique `SyntaxKind` for that case. For the time being, I've opted not to carve off `ClassStaticBlockBody` due to its similarity to `FunctionBody`, though it may be worth considering in the future. We may also want to consider carving of `BlockStatement` from `Block`, and leaving `Block` to `try..catch..finally`, but due to the overwhelming similarity between `Block` and `BlockStatement`, I've opted to leave that as-is.


jakebailey:
It's not clear if we actually want to do that split yet (at least at this stage of porting; @ahejlsberg); is there a separable portion of this PR which just does the asserting and such?

(I am hoping to do something similar to your asserts in my AST code gen change, but that is happening more slowly than I would have liked.)

rbuckton:
I can separate out the asserts, but this would remove a fair amount of complexity when porting the emitter. I primarily added the asserts to validate this specific change.

jakebailey:
I do not think we should be doing this split at this stage; just porting, even if it's less convenient. And if we aren't going to make this structural change in the existing JS code, I think it'll only make interop harder to have syntaxkinds that don't exist.

ahejlsberg:
I really don't want to take on this change right now. I think our top priority needs to be a port that is as close as possible to the current codebase, and this is a rather significant change from an API compatibility point of view.

The asserts are less problematic, yet I don't think they aid the porting effort. I mean, we already know that the parser constructs correct trees, and presumably this is the case too for a port of the emitter. It would actually be more useful to have consistent comments in the (generated) AST that show the exact sets of types that can occur for each child property. I was trying to maintain something like that through the aliases for `Node` with comments that indicate what unions they represent, but it really ought to be machine generated.

rbuckton:
> ... rather significant change from an API compatibility point of view.

One of the motivations for this change was that we are already breaking API compatibility in quite a few places in our AST, but I agree with your point regarding keeping the port as close as possible for now.

> The asserts are less problematic, yet I don't think they aid the porting effort. I mean, we already know that the parser constructs correct trees, and presumably this is the case too for a port of the emitter. It would actually be more useful to have consistent comments in the (generated) AST that show the exact sets of types that can occur for each child property. I was trying to maintain something like that through the aliases for `Node` with comments that indicate what unions they represent, but it really ought to be machine generated.

I disagree that they do not aid in the porting effort. There are already several significant changes to the AST that make a direct port less than reliable (`NodeArray` replaced with new `Node` types or just plain slices, `hasTrailingComma` is missing or replaced with a different parse involving `OmittedExpression`, etc.), so having these assertions will help us catch minor differences. While we cannot have the same level of type safety we had in TypeScript due to the lack of union types and assertions, we are not even using interfaces or struct subtypes in many of these APIs and have swung the pendulum nearly all the way to being completely untyped. This is further complicated by the fact that some of the `NodeFactory` methods take slightly different inputs than the current compiler, or have different order to the arguments.  

The Go language service does not help us here either since it eagerly resolves all of the type aliases to `Node`, so we lose a lot of information when simply using signature help or quick info.

I believe adding these assertions will help us catch these inconsistencies since we cannot rely on type safety.

If we are concerned about the runtime overhead of these assertions, I did add a comment about a Go-specific mechanism by which we could conditionally enable or disable assertions at build time such that we might only run them when testing. 

For now, I'm going to close this PR in favor of #57, which contains only the assertions.

PR #53, opened 2024-11-05
Port semver `Version` and `VersionRange`

DanielRosenwasser:
Reimplements `Version` and `VersionRange` in the new compiler.

### DIFF
--- a/internal/semver/version.go
+++ b/internal/semver/version.go
@@ -198,7 +198,7 @@ type SemverParseError struct {
 }
 
 func (e *SemverParseError) Error() string {
-	return "Could not parse version string from " + e.origInput
+	return fmt.Sprintf("Could not parse version string from %q", e.origInput)
 }
 
 func TryParseVersion(text string) (Version, error) {


internal/semver/semver.go
```diff
@@ -0,0 +1,252 @@
...
+	major      uint32
+	minor      uint32
+	patch      uint32
+	prerelease []string
+	build      []string
+}
+
+const ComparisonLessThan = -1
+const ComparisonEqualTo = 0
+const ComparisonGreaterThan = 1
```
jakebailey: I don't think these need to be exported.

DanielRosenwasser:
There are still a few things I need to get done (e.g. swap over to raw strings, add `String()` for `VersionRange`), but otherwise I think this is ready for review.

internal/semver/semver.go
```diff
@@ -0,0 +1,677 @@
...
+	build      []string
+}
+
+var _zeroPrerelease = []string{"0"}
+
+var versionZero = Version{
+	prerelease: _zeroPrerelease,
+}
+
+func incrementMajor(v Version) Version {
```
jakebailey: I think these could just be methods.

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+	"cmp"
+	"fmt"
+	"math"
+	"regexp"
+	"slices"
+	"strconv"
+	"strings"
+)
+
+// > A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative
```
rbuckton, line 13:
```suggestion
// https://semver.org/#spec-item-2
// > A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative
```
Missing spec reference from original comment block

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+// > integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor
+// > version, and Z is the patch version. Each element MUST increase numerically.
+//
+// NOTE: We differ here in that we allow X and X.Y, with missing parts having the default
+// value of `0`.
+var versionRegexp = regexp.MustCompile(`^(0|[1-9]\d*)(?:\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*)(?:-([a-zA-Z0-9-.]+))?(?:\+([a-zA-Z0-9-.]+))?)?)?$`)
```
rbuckton: To be consistent with our current implementation, shouldn't we use flags? A prefix `(?i)...` is essentially the same as `/.../i`

```suggestion
var versionRegexp = regexp.MustCompile(`(?i)^(0|[1-9]\d*)(?:\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*)(?:-([a-z0-9-.]+))?(?:\+([a-z0-9-.]+))?)?)?$`)
```

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+
+// https://semver.org/#spec-item-9
+// > A pre-release version MAY be denoted by appending a hyphen and a series of dot separated
+// > identifiers immediately following the patch version. Identifiers MUST comprise only ASCII
+// > alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers
+// > MUST NOT include leading zeroes.
+var prereleaseRegexp = regexp.MustCompile(`^(?:0|[1-9]\d*|[a-zA-Z-][a-zA-Z0-9-]*)(?:\.(?:0|[1-9]\d*|[a-zA-Z-][a-zA-Z0-9-]*))*$`)
+var prereleasePartRegexp = regexp.MustCompile(`^(?:0|[1-9]\d*|[a-zA-Z-][a-zA-Z0-9-]*)$`)
```
rbuckton:
```suggestion
var prereleaseRegexp = regexp.MustCompile(`(?i)^(?:0|[1-9]\d*|[a-z-][a-z0-9-]*)(?:\.(?:0|[1-9]\d*|[a-zA-Z-][a-zA-Z0-9-]*))*$`)
var prereleasePartRegexp = regexp.MustCompile(`(?i)^(?:0|[1-9]\d*|[a-z-][a-z0-9-]*)$`)
```

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+
+// https://semver.org/#spec-item-10
+// > Build metadata MAY be denoted by appending a plus sign and a series of dot separated
+// > identifiers immediately following the patch or pre-release version. Identifiers MUST
+// > comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty.
+var buildRegExp = regexp.MustCompile(`^[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$`)
+var buildPartRegExp = regexp.MustCompile(`^[a-zA-Z0-9-]+$`)
```
rbuckton:
```suggestion
var buildRegExp = regexp.MustCompile(`(?i)^[a-z0-9-]+(?:\.[a-z0-9-]+)*$`)
var buildPartRegExp = regexp.MustCompile(`(?i)^[a-z0-9-]+$`)
```

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+
+// https://semver.org/#spec-item-9
+// > Numeric identifiers MUST NOT include leading zeroes.
+var numericIdentifierRegExp = regexp.MustCompile(`^(?:0|[1-9]\d*)$`)
+
+type Version struct {
+	// !!!
+	// These should probably all be float64
```
rbuckton: Why? Each segment must be integral. I could see `uint64`, but `float64` would be incorrect.

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+	major      uint32
+	minor      uint32
+	patch      uint32
+	prerelease []string
+	build      []string
+}
+
+var _zeroPrerelease = []string{"0"}
```
rbuckton: Does reusing this save us anything? Slices are mutable, so wouldn't reusing it have the potential to backfire if someone were to parse a version and change the value?

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+			return comparisonLessThan
+		}
+		if !leftIsNumeric {
+			return comparisonGreaterThan
+		}
+
+		// https://semver.org/#spec-item-11
+		// > identifiers consisting of only digits are compared numerically
+		leftAsNumber := stringToNumber(left)
+		rightAsNumber := stringToNumber(right)
```
rbuckton: While the original implementation uses `+` to convert, the RegExp that guards this ensures that `left` and `right` can only be integral. Rather than converting these to `float64`, I would convert them to `int32` instead.

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+	// > a difference is found [...]
+	return slices.CompareFunc(left, right, comparePreReleaseIdentifier)
+}
+
+func comparePreReleaseIdentifier(left, right string) int {
+	// https://semver.org/#spec-item-11
+	// > Precedence for two pre-release versions with the same major, minor, and patch version
+	// > MUST be determined by comparing each dot separated identifier from left to right until
+	// > a difference is found [...]
+	r := strings.Compare(left, right)
```
rbuckton: This check only cares about equality. Wouldn't `==` be more efficient?

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+		// https://semver.org/#spec-item-11
+		// > identifiers consisting of only digits are compared numerically
+		leftAsNumber := stringToNumber(left)
+		rightAsNumber := stringToNumber(right)
+		return cmp.Compare(leftAsNumber, rightAsNumber)
+	}
+
+	// https://semver.org/#spec-item-11
+	// > identifiers with letters or hyphens are compared lexically in ASCII sort order.
+	return strings.Compare(left, right)
```
rbuckton: You already perform this check on L146. I'd either return `r` here if L146 remains as-is, or leave this as-is if L146 is changed to use `==`.

internal/semver/version.go
```diff
@@ -0,0 +1,257 @@
...
+			return result, &SemverParseError{origInput: text}
+		}
+
+		result.build = strings.Split(buildStr, ".")
+	}
+
+	return result, nil
+}
+
+func stringToNumber(s string) float64 {
```
rbuckton: This function shouldn't be necessary based on above comments.

internal/semver/version_range.go
```diff
@@ -0,0 +1,424 @@
...
+//
+// partial      ::= xr ( '.' xr ( '.' xr qualifier ? )? )?
+// xr           ::= 'x' | 'X' | '*' | nr
+// nr           ::= '0' | ['1'-'9'] ( ['0'-'9'] ) *
+// qualifier    ::= ( '-' pre )? ( '+' build )?
+// pre          ::= parts
+// build        ::= parts
+// parts        ::= part ( '.' part ) *
+// part         ::= nr | [-0-9A-Za-z]+
+var partialRegExp = regexp.MustCompile(`^([xX*0]|[1-9]\d*)(?:\.([xX*0]|[1-9]\d*)(?:\.([xX*0]|[1-9]\d*)(?:-([a-zA-Z0-9-.]+))?(?:\+([a-zA-Z0-9-.]+))?)?)?$`)
```
rbuckton:
```suggestion
var partialRegExp = regexp.MustCompile(`(?i)^([x*0]|[1-9]\d*)(?:\.([x*0]|[1-9]\d*)(?:\.([x*0]|[1-9]\d*)(?:-([a-z0-9-.]+))?(?:\+([a-z0-9-.]+))?)?)?$`)
```

internal/semver/version_range.go
```diff
@@ -0,0 +1,424 @@
...
+
+// https://github.com/npm/node-semver#range-grammar
+//
+// hyphen       ::= partial ' - ' partial
+var hyphenRegExp = regexp.MustCompile(`^\s*([a-zA-Z0-9-+.*]+)\s+-\s+([a-zA-Z0-9-+.*]+)\s*$`)
```
rbuckton:
```suggestion
var hyphenRegExp = regexp.MustCompile(`(?i)^\s*([a-z0-9-+.*]+)\s+-\s+([a-z0-9-+.*]+)\s*$`)
```

internal/semver/version_range.go
```diff
@@ -0,0 +1,424 @@
...
+
+// https://github.com/npm/node-semver#range-grammar
+//
+// simple       ::= primitive | partial | tilde | caret
+// primitive    ::= ( '<' | '>' | '>=' | '<=' | '=' ) partial
+// tilde        ::= '~' partial
+// caret        ::= '^' partial
+var rangeRegExp = regexp.MustCompile(`^([~^<>=]|<=|>=)?\s*([a-zA-Z0-9-+.*]+)$`)
```
rbuckton:
```suggestion
var rangeRegExp = regexp.MustCompile(`(?i)^([~^<>=]|<=|>=)?\s*([a-z0-9-+.*]+)$`)
```

internal/semver/version_range.go
```diff
@@ -0,0 +1,424 @@
...
+
+func formatDisjunction(sb *strings.Builder, alternatives [][]versionComparator) {
+	for i, alternative := range alternatives {
+		if i > 0 {
+			sb.WriteString(" || ")
+		}
+		formatAlternative(sb, alternative)
+	}
+
+	if sb.Len() == 0 {
```
rbuckton: Is this guaranteed to be equivalent to the original source? The original source only cared if the string constructed by this function was empty, but this function accepts the string builder as an argument and assumes the string builder is empty when invoked. While that's true currently, if the npm `semver` package decided to add new range syntax that we wanted to match, we'd need to be sure we checked this line.

I'd recommend we either just check `len(alternatives) == 0`, or compare the length after writing to the length prior to writing.

internal/semver/version.go
DanielRosenwasser: Maybe it's overkill, but my rationale was that today we can parse out large integral values that can be represented in float64, but not uint32, or will end up turning into `Infinity`. So the closest thing I could probably do is to just say if parsing ever fails, I'll just set the values to the max `uint32` value.

internal/semver/version.go
```diff
@@ -0,0 +1,260 @@
...
+type Version struct {
+	major      uint32
+	minor      uint32
+	patch      uint32
+	prerelease []string
+	build      []string
+}
+
+var versionZero = Version{
+	prerelease: []string{"0"},
```
jakebailey, line 49: You could turn this into:

```go
func zeroPrerelease() []string {
	return []string{"0"}
}
```

And then call that everywhere to ensure that it stays the same, if you wanted.

internal/semver/version.go
rbuckton: > Maybe it's overkill, but my rationale was that today we can parse out large integral values that can be represented in float64, but not uint32, or will end up turning into `Infinity`. So the closest thing I could probably do is to just say if parsing ever fails, I'll just set the values to the max `uint32` value.

semver.org doesn't specify a limit, just to ["use good judgement"](https://semver.org/#does-semver-have-a-size-limit-on-the-version-string). Realistically, if a version is larger than the max `uint32` then it's most likely crafted intentionally to break something and we should reject it.

internal/semver/version.go
DanielRosenwasser: The original intent was just to share some of the same slices, but it's probably overkill anyway. For now I think it's fine as-is.

internal/semver/version.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return sb.String()
+}
+
+type SemverParseError struct {
+	origInput string
+}
+
+func (e *SemverParseError) Error() string {
+	return "Could not parse version string from " + e.origInput
```
jakebailey: I'd actually suggest:

```suggestion
	return fmt.Sprintf("Could not parse version string from %q", e.origInput)
```

To get nice quoted text.

It's also convention that error messages _not_ start with capital letters, given they can be wrapped by other errors and a string like `Something bad: Some other error: Whoops it broke` looks odd.

andrewbranch: @jakebailey, do you want to push a change while Daniel is gone? I need to replace my third-party dependency with this code for my module resolution PR.

rbuckton:
Should I add a commit to resolve my concerns as well, or just push up a separate PR to address once this is merged?

andrewbranch:
Sorry, I didn’t realize you still had comments that hadn’t been addressed. I think Daniel is out the whole week and it would be helpful to me if we could get this in one way or another.

jakebailey:
Yeah I thought everything had been addressed through later changes

internal/semver/version.go
jakebailey: I can though now I'll wait for Ron to push onto this branch 😄

rbuckton:
> Yeah I thought everything had been addressed through later changes

It looks like it was, the various force pushes made it hard to follow.

PR #54, opened 2024-11-05
Add update methods to NodeFactory

rbuckton:
This adds Update methods to `NodeFactory` for existing nodes in preparation for porting the emitter. In the long term, I think these would make more sense on the individual `Node` subtypes, but given that creating nodes requires a `NodeFactory` at present, this seemed the best place to put them for now.


rbuckton:
Closing in favor of #195

PR #55, opened 2024-11-06
Move common utility functions to new `utils` package

ahejlsberg:
Moves common utility functions from `compiler` package to new `utils` package.

### DIFF
--- a/internal/compiler/parser.go
+++ b/internal/compiler/parser.go
@@ -79,6 +79,7 @@ func ParseJSONText(fileName string, sourceText string) *SourceFile {
 	p.nextToken()
 	pos := p.nodePos()
 	var expressions []*Node
+
 	for p.token != SyntaxKindEndOfFile {
 		var expression *Node
 		switch p.token {
@@ -117,7 +118,9 @@ func ParseJSONText(fileName string, sourceText string) *SourceFile {
 		p.finishNode(arr, pos)
 		statement = p.factory.NewExpressionStatement(arr)
 	}
+
 	p.finishNode(statement, pos)
+	p.parseExpectedToken(SyntaxKindEndOfFile)
 	node := p.factory.NewSourceFile(p.sourceText, p.fileName, []*Node{statement})
 	p.finishNode(node, pos)
 	result := node.AsSourceFile()
@@ -216,6 +219,10 @@ func (p *Parser) parseSourceFileWorker() *SourceFile {
 	}
 	pos := p.nodePos()
 	statements := p.parseList(PCSourceElements, (*Parser).parseStatement)
+	eof := p.parseTokenNode()
+	if eof.kind != SyntaxKindEndOfFile {
+		panic("Expected end of file token from scanner.")
+	}
 	node := p.factory.NewSourceFile(p.sourceText, p.fileName, statements)
 	p.finishNode(node, pos)
 	result := node.AsSourceFile()


internal/funcs/funcs.go
```diff
@@ -0,0 +1,210 @@
+// Package funcs contains useful functions.
+package funcs
```
jakebailey: I think this package should be "sliceutils" or something; helper only packages like this aren't idiomatic. https://github.com/microsoft/typescript-go/pull/56/files I think is closer to what I was envisioning.

internal/funcs/funcs.go
```diff
@@ -0,0 +1,210 @@
...
+	result[i] = t
+	return result
+}
+
+func InsertSorted[T any](slice []T, element T, cmp func(T, T) int) []T {
+	i, _ := slices.BinarySearchFunc(slice, element, cmp)
+	return slices.Insert(slice, i, element)
+}
+
+var regexps = make(map[string]*regexp.Regexp)
```
jakebailey: Long term I'd like to get rid of this in favor of plain global regexes; that's make this file all slice stuff.

internal/funcs/funcs.go
ahejlsberg: A couple of observations. One, I really don't like long package names like "sliceutils", particularly for packages with lots of functions that unnecessarily make referencing expressions longer and harder to read. Shorter names like "maps", "fs", "os", just make for a much better coding experience. Two, "mixed-bag" packages may not be idiomatic, but they're practical. The key objective here is having a core package that other packages _in our project_ can share. We're _not_ building a collections framework. Having all of our helpers spread out over several packages just adds more conceptual noise.

jakebailey: The flip side is having to remember where to find helpers associated with specific types like slices; the name `funcs` doesn't say anything about what could be "in the box" or if searching for a function there will net the helper you're looking for (I would probably think that the package contains stuff to do with functions), unless we dump everything in there again and then it's just back to square one. At least in the TS namespace, everything was dumped into one namespace so you couldn't possibly lose something, but if we're breaking things apart, I'm not personally sold on this being the way it's sliced.

gabritto: Regarding long names, you could always do a qualified import of `sliceutils` as `su` or something, but I agree with @jakebailey's point that having a more descriptive name helps finding things. But I'm not strongly against dumping everything in a "tsutils" package.

jakebailey: If we're going to have to do this, then I would prefer just calling it `utils` to make it clear that it's utils, at least.

I would not go for qualified imports just because it means files are inconsistent about naming; the only time I think they should be used is when the package path's basename differs from the package name, for clarity. e.g., packages out there which are at `github.com/blah/go-something` and declare package `something`.

gabritto: Ok, then +1 for `utils` if we choose to dump everything in one package.

ahejlsberg: I'm fine with `utils` instead of `funcs`. My preference is one single package with shared stuff. I could live with multiple packages as long as the package names are short, but honestly I think we're just wasting navel gazing time sorting them into groups that aren't in any sense complete because they just provide the extra stuff that our project so happens to need.

ahejlsberg: For now, I'll rename `funcs` to `utils`.

DanielRosenwasser: I do like the idea of general utility functions going in a single package rather than being split across some for maps, slices, and strings, but long-term "big bag of stuff" package has a bigger compilation/testing cost if you're working on non-core-compiler stuff, and it adds up quickly. 

So I think some of the stuff in @gabritto's PR make sense as well (e.g. it would be ideal if `tspath` had its own package long-term, as would something like `sys`).

jakebailey: I don't mind having `utils` be little helpers and such, but I do think complete concepts like "path handling" and "platform handling" are definitely ones that should go in other boxes. I don't think this PR moves path stuff, but Gabriela's does in a way I think is correct.

internal/funcs/funcs.go
```diff
@@ -0,0 +1,210 @@
...
+func MakeRegexp(s string) *regexp.Regexp {
+	regexpMutex.Lock()
+	rx, ok := regexps[s]
+	if !ok {
+		rx = regexp.MustCompile(s)
+		regexps[s] = rx
+	}
+	regexpMutex.Unlock()
+	return rx
+}
```
gabritto: I think we're missing `appendIfUnique` here.

PR #56, opened 2024-11-06
Move path to its own package

gabritto:


gabritto:
Going to modify this PR to just the paths changes, since we have https://github.com/microsoft/typescript-go/pull/55

internal/sliceutil/util.go
```diff
@@ -0,0 +1,207 @@
...
+				}
+				i++
+			}
+			return result, false
+		}
+	}
+	return slice, true
+}
+
+func Mapf[T, U any](slice []T, f func(T) U) []U {
```
jakebailey: This can just be "Map" now that it isn't a keyword.

internal/sliceutil/util.go
```diff
@@ -0,0 +1,207 @@
...
+	for i := len(slice) - 1; i >= 0; i-- {
+		value := slice[i]
+		if predicate(value) {
+			return i
+		}
+	}
+	return -1
+}
+
+func FindInMap[K comparable, V any](m map[K]V, predicate func(V) bool) V {
```
jakebailey: This isn't a slice function unlike the others; not sure it should be here or if we need someone other clever name for the package...

jakebailey:
> Going to modify this PR to just the paths changes, since we have #55

I think your PR is actually closer to what is idiomatic for the language, though the map helpers do mean there should be another package ...

gabritto:
> > Going to modify this PR to just the paths changes, since we have #55
> 
> I think your PR is actually closer to what is idiomatic for the language, though the map helpers do mean there should be another package ...

Ok, then I'll leave it up.

internal/sliceutil/util.go
gabritto: Moved this back to compiler for now. It's not super clear to me what we'll do about maps, if we're gonna have more map utils... We can create that package when we need it, I guess. A part I find confusing is that we also have "collections" with a custom ordered map and some functions for that.

jakebailey:
Should probably reconcile the discussions between here and #55 before merging either

internal/stringutil/util.go
```diff
@@ -1,4 +1,4 @@
-package compiler
+package stringutil
```
DanielRosenwasser: The functions in this should be renamed to not mention `Strings` since they are in `stringutil`.

jakebailey:
#55 is merged so this should be updated for the path stuff.

internal/tspath/path.go
```diff
@@ -42,16 +42,16 @@ func isDiskPathRoot(path string) bool {
...
-//	pathIsAbsolute("file:///path/to/file.ext") === true
+//	PathIsAbsolute("file:///path/to/file.ext") === true
 //	// Non-absolute
-//	pathIsAbsolute("path/to/file.ext") === false
-//	pathIsAbsolute("./path/to/file.ext") === false
+//	PathIsAbsolute("path/to/file.ext") === false
+//	PathIsAbsolute("./path/to/file.ext") === false
 //	```
-func pathIsAbsolute(path string) bool {
+func PathIsAbsolute(path string) bool {
```
jakebailey, line 54: Technically this should be `IsAbsolute` since it's written as `tspath.IsAbsolute`, but, not super problematic and I wouldn't bother changing it right now

PR #57, opened 2024-11-06
Add common node test assertions

rbuckton:
This pulls out the assertions from #52.

As is, this currently catches a minor case during parse where we produce an invalid `FunctionTypeNode` when trying to parse type arguments as part of parsing a left-shift `<<`.

I'd also note that this will mostly replace passing in optional `test` callbacks to node visitors since Go does not support overloads, but will provide the same coverage and capability.

### DIFF
--- a/internal/compiler/debug.go
+++ b/internal/compiler/debug.go
@@ -6,6 +6,8 @@ import (
 	"runtime"
 	"slices"
 	"strings"
+
+	"github.com/microsoft/typescript-go/internal/utils"
 )
 
 // TODO(rbuckton): If we want some of these assertions to only be part of a checked/debug build, we could
@@ -115,7 +117,7 @@ func assertNode(node *Node, fns ...NodeTest) {
 				return
 			}
 		}
-		assertFail("Node of %v did not pass the expected node tests: %v", node.kind, strings.Join(mapf(fns, getFunctionName), ", "))
+		assertFail("Node of %v did not pass the expected node tests: %v", node.kind, strings.Join(utils.Map(fns, getFunctionName), ", "))
 	}
 }
 


internal/compiler/ast.go
```diff
@@ -749,13 +773,17 @@ type TypeParameterDeclaration struct {
 	NodeBase
 	DeclarationBase
 	ModifiersBase
-	name        *IdentifierNode // Identifier
-	constraint  *TypeNode       // Optional
-	defaultType *TypeNode       // Optional
-	expression  *Node           // For error recovery purposes
+	name        *IdentifierNode // IdentifierNode
```
sandersn, line 776: what is the point of the `// IdentifierNode` comment here and elsewhere?

internal/compiler/types.go
```diff
@@ -389,6 +389,7 @@ const (
 	SyntaxKindSyntaxList
 	// Transformation nodes
 	SyntaxKindNotEmittedStatement
+	SyntaxKindNotEmittedTypeElement
```
sandersn, line 392: is this intended to be part of this PR?

internal/compiler/utilities.go
```diff
@@ -1911,10 +2036,26 @@ func isPushOrUnshiftIdentifier(node *Node) bool {
 	return text == "push" || text == "unshift"
 }
 
+func isThisExpression(node *Node) bool {
+	return node.kind == SyntaxKindTrueKeyword
```
sandersn:
```suggestion
	return node.kind == SyntaxKindThisKeyword
```

jakebailey:
Unfortunately this leads to a 10% perf loss in parse:

```
goos: linux
goarch: amd64
Parse/empty.ts-20                                       306.6n ± 1%   303.9n ± 4%        ~ (p=0.247 n=10)
Parse/checker.ts-20                                     43.39m ± 2%   48.62m ± 1%  +12.04% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             14.30m ± 1%   16.20m ± 1%  +13.27% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 599.1µ ± 2%   660.3µ ± 2%  +10.21% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   228.6µ ± 1%   261.4µ ± 1%  +14.36% (p=0.000 n=10)
geomean                                                 482.1µ        528.7µ        +9.66%

Parse/empty.ts-20                                         704.0 ± 0%     704.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     23.59Mi ± 0%   24.07Mi ± 0%  +2.04% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             8.537Mi ± 0%   9.062Mi ± 0%  +6.15% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 337.4Ki ± 0%   342.9Ki ± 0%  +1.65% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   153.2Ki ± 0%   162.0Ki ± 0%  +5.78% (p=0.000 n=10)
geomean                                                 375.9Ki        387.5Ki       +3.10%
¹ all samples are equal

                                                      │   old.txt   │                new.txt                │
                                                      │  allocs/op  │  allocs/op   vs base                  │
Parse/empty.ts-20                                        4.000 ± 0%    4.000 ± 0%        ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     241.0k ± 0%   329.1k ± 0%  +36.52% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             100.0k ± 0%   147.8k ± 0%  +47.78% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 3.788k ± 0%   4.919k ± 0%  +29.86% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   1.441k ± 0%   2.029k ± 0%  +40.80% (p=0.000 n=10)
geomean                                                 3.502k        4.546k       +29.83%
¹ all samples are equal
```

jakebailey:
That being said, it could just be the overhead of the varargs slices, which would be fixed by further manual inlining / code generation

internal/compiler/ast.go
rbuckton: It's a workaround for limitations in the Go language service, for example:

image

image

The quick info here eagerly resolves the `TypeNode` type alias to `Node`, leaving no useful information about the property aside from the comment.

internal/compiler/types.go
rbuckton: Yes. It's used in utilities.go. I'd much rather have the kind and the check in that function than to overlook it later.

rbuckton:
> Unfortunately this leads to a 10% perf loss in parse: [...]

> That being said, it could just be the overhead of the varargs slices, which would be fixed by further manual inlining / code generation

I can remove the vararg slices, as I was concerned about those. If the runtime overhead even without them is problematic, then I would suggest we consider using build tags for checked/debug builds as suggested here: https://github.com/microsoft/typescript-go/pull/57/files#diff-3f5dd66d5afa8d935ca9fab29bd4d6ab2bc8fddd805350a88744ffb4dc6b3405R11

rbuckton:
What are you using to run this benchmark?

jakebailey:
It's:

```console
# Capture before
$ go test -run=- -bench=. -benchmem -count=10 ./internal/compiler | tee old.txt
# Capture after
$ go test -run=- -bench=. -benchmem -count=10 ./internal/compiler | tee new.txt
# Compare
$ benchstat old.txt new.txt
```

Where `benchstat` came from `go install golang.org/x/perf/cmd/benchstat@latest`, but can also just be run direct with `go run golang.org/x/perf/cmd/benchstat@latest old.txt new.txt`.

Docs at: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat

Can replace `| tee out.txt` with `> out.txt` if you aren't on a unix system.

rbuckton:
When I run the first command I just get

```
no Go files in D:\dev\typescript-go
```

jakebailey:
Did you leave off `./internal/compiler`? That should select the package to test; it seems like it's trying to run right at the root or something...

rbuckton:
> Did you leave off `./internal/compiler`? That should select the package to test; it seems like it's trying to run right at the root or something...

Nope, it's there. I am running the command within Powershell on Windows. If I toss that command into a task in herebyfile.mjs I can run it with hereby/gulp, but it doesn't seem to want to run from Powershell.

rbuckton:
It's not just the varargs slices. I expect that means no asserts unless we decide we want checked/release builds.

I can put up a PR that just adds the missing type aliases to assist with quickinfo/signature help. If the emitter ends up being unable to reuse any of the existing transform logic, then I'm going to need to rely on that more than other parts of the compiler that are roughly being ported 1:1.

jakebailey:
Yeah, feel free to send that; I do have some ideas on how to make these checks faster but I am definitely wary about having two different builds... But we could theoretically build tag that and then ensure local dev work happens with debug flags enabled or something... It's tricky because you sorta want to be benchmarking it without the checks, but you also want them on when doing tests...

rbuckton:
Deprecating in favor of #60

PR #58, opened 2024-11-07
Adds `NodeVisitor` and `ParenthesizerRules`

rbuckton:
This adds `ParenthesizerRules` and `NodeVisitor`. `NodeVisitor` encompasses the current `visitNode`, `visitNodes`, and `visitEachChild` functions in a way that is more conducive to Golang, avoiding the numerous overloads and optional parameters, and will help to cut down on closures we would have other generated when visiting/transforming:

```js
// js
function visitor(node: Node): VisitResult<Node> {
  switch (node.kind) {
    case SyntaxKind.ClassExpression: return visitClassExpression(node as ClassExpression);
    ...
  }
}
function visitClassExpression(node: ClassExpression) { ... }
```

```go
struct MyNodeVisitor {
  NodeVisitorBase
}
func (v *MyNodeVisitor) VisitClassExpression(node *ClassExpression) *Node { ... }
```

Note that this simplifies `VisitNode` as it no longer accepts a `test` callback. This validation would instead be handled more consistently in the various `New` methods on `NodeFactory`.

Also note that this PR is currently a draft. It intends to depend on #57 and the version of `debug.go` in that PR, and will be rebased once #57 and #54 are merged.

### DIFF
--- a/internal/compiler/ast.go
+++ b/internal/compiler/ast.go
@@ -22,10 +22,22 @@ func visitNodes(v Visitor, nodes []*Node) bool {
 	return false
 }
 
+// NodeTest
+
+type NodeTest func(*Node) bool
+
 // NodeFactory
 
 type NodeFactory struct {
-	identifierPool Pool[Identifier]
+	parenthesizerRules ParenthesizerRules
+	identifierPool     Pool[Identifier]
+}
+
+func (f *NodeFactory) ParenthesizerRules() ParenthesizerRules {
+	if f.parenthesizerRules == nil {
+		f.parenthesizerRules = NewParenthesizerRules(f)
+	}
+	return f.parenthesizerRules
 }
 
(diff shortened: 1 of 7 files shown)


jakebailey:
The main ergonomic annoyance with this pattern is that there isn't a form you can write that's a closure; you have to declare the type, then put all of the state you need into the visitor explicitly.

The Go compiler itself doesn't use this pattern, instead doing it more like we do today: https://pkg.go.dev/go/ast#Walk

rbuckton:
> The main ergonomic annoyance with this pattern is that there isn't a form you can write that's a closure; you have to declare the type, then put all of the state you need into the visitor explicitly.
> 
> The Go compiler itself doesn't use this pattern, instead doing it more like we do today: https://pkg.go.dev/go/ast#Walk

I plan to rebase this PR and bring it closer to what we currently do, even if only for the purpose of discussion.

rbuckton:
Closing in favor of #195

PR #59, opened 2024-11-08
More type checking

ahejlsberg:
This PR ports

* Base constraints
* More type relations
* Discriminated union type reduction
* Excess property checking
* Object literal expression checking

Also, the PR simplifies the very complicated logic we had for pruning and suppressing messages in error pyramids.

In between `checker.go` and `relater.go`, we now have about 14,000 lines of checker code ported.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -212,6 +212,8 @@ type Checker struct {
 	typeParameterLinks                 LinkStore[*Symbol, TypeParameterLinks]
 	interfaceTypeLinks                 LinkStore[*Symbol, InterfaceTypeLinks]
 	typeAliasLinks                     LinkStore[*Symbol, TypeAliasLinks]
+	spreadLinks                        LinkStore[*Symbol, SpreadLinks]
+	patternForType                     map[*Type]*Node
 	anyType                            *Type
 	autoType                           *Type
 	wildcardType                       *Type
@@ -251,6 +253,11 @@ type Checker struct {
 	anyFunctionType                    *Type
 	noConstraintType                   *Type
 	circularConstraintType             *Type
+	markerSuperType                    *Type
+	markerSubType                      *Type
+	markerOtherType                    *Type
+	markerSuperTypeForCheck            *Type
+	markerSubTypeForCheck              *Type
 	enumNumberIndexInfo                *IndexInfo
 	patternAmbientModules              []PatternAmbientModule
 	patternAmbientModuleAugmentations  SymbolTable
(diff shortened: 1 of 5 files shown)


internal/compiler/relater.go
```diff
@@ -448,20 +509,82 @@ func getRecursionIdentity(t *Type) RecursionId {
...
-	// propType := keyPropertyName && c.getTypeOfPropertyOfType(t, keyPropertyName)
-	// return propType && c.getConstituentTypeForKeyType(unionType, propType)
+func (c *Checker) findBestTypeForInvokable(source *Type, unionTarget *Type, kind SignatureKind) *Type {
+	if len(c.getSignaturesOfType(source, kind)) != 0 {
+		return utils.Find(unionTarget.Types(), func(t *Type) bool { return len(c.getSignaturesOfType(t, kind)) != 0 })
+	}
+	return nil
+}
+
+func (c *Checker) findMostOverlappyType(source *Type, unionTarget *Type) *Type {
```
jakebailey, line 555: I love how we have this function

internal/compiler/checker.go
```diff
@@ -1828,6 +1872,532 @@ func (c *Checker) checkReferenceExpression(expr *Node, invalidReferenceMessage *
...
+	return s.flags&(TypeFlagsAny|TypeFlagsNonPrimitive|TypeFlagsObject|TypeFlagsInstantiableNonPrimitive) != 0 ||
+		s.flags&TypeFlagsUnionOrIntersection != 0 && utils.Every(s.Types(), c.isValidSpreadType)
+}
+
+func (c *Checker) getUnionIndexInfos(types []*Type) []*IndexInfo {
+	sourceInfos := c.getIndexInfosOfType(types[0])
+	var result []*IndexInfo
+	for _, info := range sourceInfos {
+		indexType := info.keyType
+		if utils.Every(types, func(t *Type) bool { return c.getIndexInfoOfType(t, indexType) != nil }) {
```
DanielRosenwasser, line 2230: I feel like this gets a little hard to read. I'd rather stuff like this get lifted into its own variable.

```suggestion
		allTypesHaveIndexInfo := utils.Every(types, func(t *Type) bool {
			return c.getIndexInfoOfType(t, indexType) != nil
		})
		if allTypesHaveIndexInfo {
```

ahejlsberg: Yeah, once we have the port done we'll want to do a sweep where we optimize this, both for readability and performance. Unless this is inlined, we're creating a closure here that could be avoided by manually looping.

PR #60, opened 2024-11-08
Add missing type aliases to improve quickinfo

rbuckton:
This adds a number of missing type aliases that make it easier to reason over the new AST structure. 

This also adds comments that repeat the type of the field as a workaround for an issue with Go's language service where it eagerly resolves type aliases in quick info:

image
image

This also removes a duplicate `kind` field on `ForInOrOfStatement` as that results in unnecessary extra space reserved in the struct.

Closes #57

### DIFF
--- a/internal/compiler/ast.go
+++ b/internal/compiler/ast.go
@@ -568,8 +568,8 @@ type UnionOrIntersectionTypeNode = Node // UnionTypeNode | IntersectionTypeNode
 type TemplateLiteralLikeNode = Node     // TemplateHead | TemplateMiddle | TemplateTail
 type TemplateMiddleOrTail = Node        // TemplateMiddle | TemplateTail
 type TemplateLiteral = Node             // TemplateExpression | NoSubstitutionTemplateLiteral
-type TypePredicateParameterName = Node  // IdentifierReferenceNode | ThisTypeNode
-type ImportAttributeName = Node         // IdentifierName | StringLiteral
+type TypePredicateParameterName = Node  // Identifier | ThisTypeNode
+type ImportAttributeName = Node         // Identifier | StringLiteral
 type LeftHandSideExpression = Node      // subset of Expression
 
 // Aliases for node singletons
@@ -1117,7 +1117,7 @@ func (node *CaseBlock) ForEachChild(v Visitor) bool {
 type CaseOrDefaultClause struct {
 	NodeBase
 	expression          *Expression  // Expression. nil in default clause
-	statements          []*Statement // Statement[]
+	statements          []*Statement // []Statement
 	fallthroughFlowNode *FlowNode
 }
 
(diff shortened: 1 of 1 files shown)


jakebailey:
> a workaround for an issue with Go's language service where it eagerly resolves type aliases in quick info

To be clear, I don't think this is a bug. Type aliases don't really "exist" in Go, they're redirections intended for refactoring code and leaving old names behind. We're misusing them for something they aren't designed for, so it's really on us that things don't look great in the editor.

internal/compiler/ast.go
```diff
@@ -1092,12 +1110,12 @@ func (node *CaseBlock) ForEachChild(v Visitor) bool {
 
 type CaseOrDefaultClause struct {
 	NodeBase
-	expression          *Node // nil in default clause
-	statements          []*Statement
+	expression          *Expression  // Expression. nil in default clause
+	statements          []*Statement // Statement[]
```
jakebailey: Given all of these are types in Go, it doesn't feel right to write the array types in TS syntax, at least in my opinion.

I feel like if this whole thing is just here to make hover show something, that it should just be purely copy/paste of the type annotation to the left.

rbuckton:
> To be clear, I don't think this is a bug. Type aliases don't really "exist" in Go, they're redirections intended for refactoring code and leaving old names behind. We're misusing them for something they aren't designed for, so it's really on us that things don't look great in the editor.

Either quickinfo is a bug, or signature help is a bug, because signature help does not resolve to `Node`:

image
image

jakebailey:
Huh, weird... Definitely not what I expected

rbuckton:
Also compare quickinfo for a type:

image

vs. quickinfo for a property on that type:

image

rbuckton:
Are there any other thoughts on this one? I expect we'll also get these generated eventually, but it'd be nice to have better quick info now while I'm working on the printer.

internal/compiler/ast.go
```diff
@@ -549,21 +549,33 @@ type MemberName = Node                  // Identifier | PrivateIdentifier
...
+type JsxTagNameExpression = Node        // IdentifierReference | KeywordExpression | JsxTagNamePropertyAccess | JsxNamespacedName
 type ClassLikeDeclaration = Node        // ClassDeclaration | ClassExpression
 type AccessorDeclaration = Node         // GetAccessorDeclaration | SetAccessorDeclaration
 type LiteralLikeNode = Node             // StringLiteral | NumericLiteral | BigIntLiteral | RegularExpressionLiteral | TemplateLiteralLikeNode | JsxText
 type LiteralExpression = Node           // StringLiteral | NumericLiteral | BigIntLiteral | RegularExpressionLiteral | NoSubstitutionTemplateLiteral
 type UnionOrIntersectionTypeNode = Node // UnionTypeNode | IntersectionTypeNode
 type TemplateLiteralLikeNode = Node     // TemplateHead | TemplateMiddle | TemplateTail
 type TemplateMiddleOrTail = Node        // TemplateMiddle | TemplateTail
+type TemplateLiteral = Node             // TemplateExpression | NoSubstitutionTemplateLiteral
+type TypePredicateParameterName = Node  // IdentifierReferenceNode | ThisTypeNode
```
ahejlsberg: That should be `IdentifierNode`

PR #61, opened 2024-11-11
Make all AST members public

weswigham:
This way packages (folders) that are *not* the one the AST is defined in can actually inspect and manipulate the AST. This is basically a prerequisite for any kind of reasonable data structure organization, imo, and is absolutely required to start breaking the cycles in the data structure definitions. We probably want to push this through sooner rather than later, since making members public like this is a big rename operation.

### DIFF
--- a/internal/compiler/ast.go
+++ b/internal/compiler/ast.go
@@ -4221,11 +4221,11 @@ type SourceFile struct {
 	DeclarationBase
 	LocalsContainerBase
 	Text                        string
-	fileName                    string
-	path                        string
+	FileName                    string
+	Path                        string
 	Statements                  []*Statement
-	diagnostics                 []*Diagnostic
-	bindDiagnostics             []*Diagnostic
+	Diagnostics                 []*Diagnostic
+	BindDiagnostics             []*Diagnostic
 	BindSuggestionDiagnostics   []*Diagnostic
 	LineMap                     []textpos.TextPos
 	LanguageVersion             ScriptTarget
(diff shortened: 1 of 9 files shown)


internal/compiler/ast.go
```diff
@@ -42,460 +42,460 @@ func (f *NodeFactory) NewNode(kind SyntaxKind, data NodeData) *Node {
 // interface valued properties either store a true nil or a reference to a non-nil struct.
 
 type Node struct {
-	kind   SyntaxKind
-	flags  NodeFlags
-	loc    TextRange
-	id     NodeId
-	parent *Node
-	data   NodeData
+	Kind   SyntaxKind
```
rbuckton, line 45: While I understand `node.Kind == ...` seems preferable, `node.Kind()` seems more Go-like, especially when you look at something like the `reflect` package's definition of `type.Kind()`. I'm still a bit concerned that we're making public and writable something we'd generally prefer be immutable.

internal/compiler/ast.go
```diff
@@ -42,460 +42,460 @@ func (f *NodeFactory) NewNode(kind SyntaxKind, data NodeData) *Node {
...
-func (n *Node) Locals() SymbolTable                       { return n.data.LocalsContainerData().locals }
-func (n *Node) FunctionLikeData() *FunctionLikeBase       { return n.data.FunctionLikeData() }
-func (n *Node) Parameters() []*ParameterDeclarationNode   { return n.data.FunctionLikeData().parameters }
-func (n *Node) ReturnType() *TypeNode                     { return n.data.FunctionLikeData().returnType }
-func (n *Node) ClassLikeData() *ClassLikeBase             { return n.data.ClassLikeData() }
-func (n *Node) BodyData() *BodyBase                       { return n.data.BodyData() }
+func (n *Node) Pos() int                                  { return n.Loc.Pos() }
+func (n *Node) End() int                                  { return n.Loc.End() }
+func (n *Node) ForEachChild(v Visitor) bool               { return n.Data.ForEachChild(v) }
+func (n *Node) GetName() *DeclarationName                 { return n.Data.GetName() }
```
rbuckton, line 60: This feels like a case where if we left the field as `name`, then the public `Name()` method would be correct both on `Node` and its subtypes. With the exception of setting `symbol`, pretty much all of these fields are immutable outside of the factory.

weswigham:
This is gonna be superseded by #66

PR #62, opened 2024-11-11
Variance and signature type relations

ahejlsberg:
Ports logic for variance computation, signature type relations, and type predicates.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -1027,8 +1027,7 @@ func (c *Checker) checkExpressionWorker(node *Node, checkMode CheckMode) *Type {
 		return c.checkPropertyAccessExpression(node, checkMode, false /*writeOnly*/)
 	case SyntaxKindBinaryExpression:
 		return c.checkBinaryExpression(node, checkMode)
-	case SyntaxKindTypeAssertionExpression:
-	case SyntaxKindAsExpression:
+	case SyntaxKindTypeAssertionExpression, SyntaxKindAsExpression:
 		return c.checkAssertion(node, checkMode)
 	}
 	return c.anyType // !!!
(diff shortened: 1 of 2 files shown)


internal/compiler/checker.go
```diff
@@ -993,6 +1027,9 @@ func (c *Checker) checkExpressionWorker(node *Node, checkMode CheckMode) *Type {
 		return c.checkPropertyAccessExpression(node, checkMode, false /*writeOnly*/)
 	case SyntaxKindBinaryExpression:
 		return c.checkBinaryExpression(node, checkMode)
+	case SyntaxKindTypeAssertionExpression:
+	case SyntaxKindAsExpression:
```
jakebailey: This seems like a bug; should be:

```suggestion
	case SyntaxKindTypeAssertionExpression, SyntaxKindAsExpression:
```

Right?

internal/compiler/relater.go
```diff
@@ -949,6 +975,739 @@ func (c *Checker) typeCouldHaveTopLevelSingletonTypes(t *Type) bool {
...
+				}
+				c.outofbandVarianceMarkerHandler = oldHandler
+				if unmeasurable || unreliable {
+					if unmeasurable {
+						variance |= VarianceFlagsUnmeasurable
+					}
+					if unreliable {
+						variance |= VarianceFlagsUnreliable
+					}
+				}
```
jakebailey:
```suggestion
				if unmeasurable {
					variance |= VarianceFlagsUnmeasurable
				}
				if unreliable {
					variance |= VarianceFlagsUnreliable
				}
```

(I think this is a straight port though)

internal/compiler/checker.go
jakebailey: I think there's a lint rule we can enable which will force us to write a comment in a switch/case that explains why the case is empty (which would help this in the future)

ahejlsberg:
I'll fix, give me a couple of minutes.

internal/compiler/checker.go
ahejlsberg: Yeah, we need a lint rule here or something.

jakebailey: I can't find it, so it may just be a custom rule we have to invent.

PR #63, opened 2024-11-11
Rename utils to core

jakebailey:
Pending us putting more stuff in here such that `utils` is a bad name. Hopefully this gets broken down further later.


internal/core/core.go
```diff
@@ -1,5 +1,4 @@
-// Package funcs contains useful functions.
```
jakebailey, line 1: Not really much use in describing an omnibus package 😄

PR #66, opened 2024-11-12
Move AST and its' dependents into packages, make AST members public

weswigham:
This PR moves `ast.go` and its' dependencies into a standalone `ast` package that only depends on the `core` package, which makes it much easier to write code that manipulates AST nodes that doesn't have to live inside the `internal/compiler` folder.

### DIFF
--- a/internal/compiler/utilities.go
+++ b/internal/compiler/utilities.go
@@ -3713,8 +3713,8 @@ func getExports(symbol *ast.Symbol) ast.SymbolTable {
 func getLocals(container *ast.Node) ast.SymbolTable {
 	data := container.LocalsContainerData().Locals()
 	if data == nil {
-		container.LocalsContainerData().SetLocals(make(ast.SymbolTable))
-		return container.LocalsContainerData().Locals()
+		data = make(ast.SymbolTable)
+		container.LocalsContainerData().SetLocals(data)
 	}
 	return data
 }


jakebailey:
Moving every enum to its own package I think is overkill; I think we can do fine with:

- `SyntaxKind` -> `syntax.Kind`
- `NodeFlags` -> `ast.NodeFlags`
- The rest in `core`

(If we do want to split them into packages, they should go in some `enums` dir I guess, but I do not think this level of granularity is desirable)

internal/scripttarget/pkg.go
```diff
@@ -0,0 +1,23 @@
+package scripttarget
+
+//go:generate go run golang.org/x/tools/cmd/stringer -type=Type -output=stringer_generated.go
+
+type Type int32
```
jakebailey: I really do not like this pattern of a non-descriptive name like this. I'd rather have a redundant name than a type called "type". Or really, just `core.Whatever`

weswigham: Even though the package name is part of the type name and thus is totally descriptive as long as it's outside this file? And we already know we're going to hear complaints in the other direction that `scripttarget.ScriptTarget` is very repetitive, and was a _driving reason_ for the "just throw it all into the god namespace" pattern? Because from that POV there's a good reason to be even more terse and make the usage `scripttarget.T` - and clearly names like that aren't entirely foreign in `go`, since the testing suite uses them. If it's not obvious, I've scripted most of this stage, so it's pretty easy to change it either way.

jakebailey: `scripttarget.T` is definitely worse... I would not copy `testing` (I don't think they would have named their types like that if they were to introduce testing today).

I just think of reading `scripttarget.Type` or hovering on something and it being moderately weird.

But, I don't even think we should be putting these into their own packages at all. `syntax.Kind` is just because the name is clever and people seemed to like it better than writing `ast.SyntaxKind`... the others are going to used so much less

jakebailey: (And, adding packages for these was not in the TODO we talked about as far as I remember, just `syntax.Kind`)

ahejlsberg: I agree with Jake. Let's have `syntax.Kind`, `ast.NodeFlags`, put the rest in `core`.

weswigham:
> but I do not think this level of granularity is desirable

It keeps the namespaces a _lot_ cleaner though.

> Moving every enum to its own package I think is overkill; 

I'm just thinking about keeping everything `ast` needs to depend on out of `core` and in its' own thing, so `ast` doesn't need to depend on `core` and is truly standalone (it does just need some enums and the pool allocator which are logically separate, and the symbol and symboltable definitions to move with it) and `core` is actual compiler utilities and not "dumping-ground of nice-to-haves literally everything depends on". Since the AST's datastructure descriptions clearly don't depend on `Filter` and `Map` functions (or basically anything currently in `core`), for example. Likewise, `CompilerOptions` is another user of these many of enums and I kinda doubt it'll need any collection utilities when it's teased out to support its' standalone parsing packages. I *could* put all or most of these enums into one package, but I think separate packages like this the usages line up a lot more closely with the original TS code and the editor behaviors work pretty well (given how the LS automatically adds and removes imports as-needed as you pick through completions and save the file). 

Putting them into a folder is absolutely a good idea, though. I'll definitely move them under a common header folder.

I still need to add `SymbolFlags`, `CheckFlags`, and `Tristate` to the set of moved enums the AST needs. Tristate is actually a fun one since it wasn't migrated following the same conventions as the rest, so I need to migrate it in a slightly more bespoke way.

One moderately annoying thing I've already taken note of though: the regexp package has a `syntax` sub-package that somehow ranks higher in completions than the local `syntax` package when you start typing `syntax`. Presumably because its' import path is shorter/fewer segments, as a builtin package. That seems wrong. We should maybe file a bug on that.

jakebailey:
> One moderately annoying thing I've already taken note of though: the regexp package has a `syntax` sub-package that somehow ranks higher in completions than the local `syntax` package when you start typing `syntax`. Presumably because its' import path is shorter/fewer segments, as a builtin package. That seems wrong. We should maybe file a bug on that.

Yikes, that is a bad one. Maybe it does need to be `syntaxkind.Kind`, `syntaxkind.Identifier`, etc.

weswigham:
> Maybe it does need to be syntaxkind.Kind, syntaxkind.Identifier, etc.

I've normalized it so that the package name is the old full type name, and the new type name is `Kind` if it ended in `Kind`, `Flags` if it ended in `Flags`, and `Type` otherwise. So `syntaxkind.Kind`, `checkflags.Flags`, and `tristate.Type`.

This should be all the enums that have to be extracted to move `ast.go` into its' own package. It's just the pool allocator (which is only referenced in two or three places) which also needs its' own package (since it's used by the ast's factory functions), then renaming all the fields in the ast to be public again (so any existing references continue to work) and fixing up all the references to nodes/symbol/symboltable to refer to the `ast` package.

ahejlsberg:
Pool allocator should just be in `core`.

ahejlsberg:
I'm not a fan of all these enum packages. We have hundreds and hundreds of enums, seems way overkill to have a package for each of them as we move towards a more granular state.

I think the best solution is to just move all of this stuff into the `ast` package. So, `ast.SyntaxKind`, `ast.Node`, etc. This other stuff just doesn't seem idiomatic.

ahejlsberg:
And anything that isn't AST related should move to `core`. Such as `core.Pool`, `core.Tristate`, `core.CompilerOptions`.

jakebailey:
I am still also definitely not for these little enum packages; `SyntaxKind` was the only one I was really okay with. Maybe `tristate`. But we're going to have a lot of them to the point where I don't feel it's going to be worth it.

That being said, I do think we will eventually move `pool` elsewhere as I have some ideas on how to allocate with differing growth strategies, but `core` seems totally fine for it for now.

(I am not sure why we are re-bikeshedding this; I thought we had the general scheme together that in the short term, we were just moving `ast`, other stuff to `core`, and _maybe_ adding `syntaxkind.Kind`.)

ahejlsberg:
To summarize

* Enums moving to `core`: `LanguageVariant`, `ScriptKind`, `ScriptTarget`, `Tristate`.
* Enums moving to `ast`: `CheckFlags`, `NodeFlags`, `SymbolFlags`, `SyntaxKind`.

If we find `ast.SyntaxKind`, `ast.SyntaxKindIdentifier` etc. to be too long, we could consider renaming all `SyntaxKind` to just `Kind`. We'd then have `ast.Kind`, `ast.KindIdentifier`, etc. and naming pattern would still be consistent with all other enums.

I think we'll also want to change all of the `ast.isXXX(node *Node)` check functions to methods on node such that we have `node.IsIdentifier()` and `node.AsIdentifier()` etc. This can happen in a separate PR if need be.

internal/ast/checkflags.go
```diff
@@ -0,0 +1,35 @@
+package ast
+
+// CheckFlags
+
+type CheckFlags = uint32
```
jakebailey, line 5: This seems to be a typo in the original code; this should not be an alias, it should be a named type.

internal/ast/pkg.go
jakebailey, line 1: This can be deleted.

internal/compiler/printer/printer.go
```diff
@@ -30,371 +30,371 @@ func PrintNode(node *compiler.Node) string {
 // func printNode(node *compiler.Node, writer EmitTextWriter) {
 // 	// TODO: Port emitter.ts's emit code more faithfully
 // 	switch node.Kind {
-// 	case compiler.SyntaxKindUnknown, compiler.SyntaxKindEndOfFile, compiler.SyntaxKindConflictMarkerTrivia, compiler.SyntaxKindNonTextFileMarkerTrivia:
+// 	case compiler.ast.KindUnknown, compiler.ast.KindEndOfFile, compiler.ast.KindConflictMarkerTrivia, compiler.ast.KindNonTextFileMarkerTrivia:
```
jakebailey: All of these seem to have been replaced incorrectly.

weswigham: Yeah, but this comment as a whole can probably be removed - I went through the effort of making the switch to put together a quick pretty printer, but Ron's on that now and probably already has the same or better, so I'm not stressing over the mangled comment.

internal/ast/symbol.go
```diff
@@ -0,0 +1,31 @@
+package ast
+
+// Ids
+
+type NodeId uint32
```
jakebailey: From a Go style perspective, it sure does hurt my eyes to see `Id` instead of `ID` everywhere

internal/ast/symbol.go
```diff
@@ -0,0 +1,31 @@
...
+// Symbol
+
+type Symbol struct {
+	Flags                        SymbolFlags
+	CheckFlags                   CheckFlags // Non-zero only in transient symbols created by Checker
+	ConstEnumOnlyModule          bool       // True if module contains only const enums or other modules with only const enums
+	IsReplaceableByMethod        bool
+	Name                         string
+	Declarations                 []*any
+	ValueDeclaration             *any
```
jakebailey: I assume this is temporary because the AST hasn't moved

internal/compiler/ast.go
```diff
@@ -46,7 +46,7 @@ func (f *NodeFactory) NewNode(kind ast.Kind, data NodeData) *Node {
 type Node struct {
 	Kind   ast.Kind
 	Flags  ast.NodeFlags
-	Loc    TextRange
+	Loc    textpos.TextRange
```
jakebailey: I think this should just move to core.

ahejlsberg: I agree, move `TextPos` and `TextRange` to `core`.

rbuckton:
> I think we'll also want to change all of the `ast.isXXX(node *Node)` check functions to methods on node such that we have `node.IsIdentifier()` and `node.AsIdentifier()` etc. This can happen in a separate PR if need be.

We _frequently_ use these node test methods as callbacks throughout the compiler. Won't that introduce hundreds or thousands of closures? We also have more than a few aggregate node tests like `isMemberName`, `isClassElement`, etc., that we use regularly as well. I'm concerned that would make completions on `node` somewhat cluttered.

internal/compiler/printer/printer.go
rbuckton: I would just remove the comment

internal/ast/symbol.go
rbuckton: It depends on whether you consider "id" to be shorthand for "Identity", in which case `Id` makes sense, or an acronym for "Identity Document" (like a driver's license), in which case `ID` makes sense. I generally see it as shorthand for "Identity", but that's a personal preference.

internal/compiler/ast.go
```diff
@@ -775,102 +777,102 @@ type TypeParameterDeclaration struct {
 	NodeBase
 	DeclarationBase
 	ModifiersBase
-	name        *IdentifierNode // IdentifierNode
-	constraint  *TypeNode       // TypeNode. Optional
-	defaultType *TypeNode       // TypeNode. Optional
-	expression  *Expression     // Expression. Optional, For error recovery purposes
+	Name_        *IdentifierNode // IdentifierNode
```
rbuckton: I'm not sure I'm a fan of having a public `Name_` field if we can already access it though `.Name()`. The only thing that ever sets this should be the factory method. I wonder if we're being far too coarse grained with these changes.

What was wrong with using methods as accessors again? I can see how for cases like setting `Symbol` we would want that field to be public so that it's writable. However, it seems like in most other cases it might be far cleaner to have specific accessor methods on node subtypes which in some cases might override a more general purpose method on `Node` (like `Name()`, `Text()`, etc.).

jakebailey:
> We _frequently_ use these node test methods as callbacks throughout the compiler. Won't that introduce hundreds or thousands of closures? We also have more than a few aggregate node tests like `isMemberName`, `isClassElement`, etc., that we use regularly as well. I'm concerned that would make completions on `node` somewhat cluttered.

It wouldn't, but the syntax one would have to use is `(*ast.Node).IsIdentifier`, which is quite annoying. The comment was sent before we talked about it in person, we just never replied here to say "nevermind, don't do that". So I think we'll just stick with top level funcs for now.

internal/compiler/ast.go
jakebailey: > What was wrong with using methods as accessors again?

I would personally like to use accessors, but for this PR, I don't think we should do so to keep the other code similar. It's not that hard to refactor the code to change it later. This movement was supposed to be a quick move, not a giant refactor.

The downsides are just:

- More parens. I personally don't think this is a big deal.
- The parser/binder sometimes assign directly onto the AST to finish things off. That would become more challenging.

internal/ast/symbol.go
jakebailey: I remember this argument in relation to some TC39 thing, but the convention in Go is to write `ID`, not `Id`. https://go.dev/wiki/CodeReviewComments#:~:text=This%20rule%20also%20applies%20to%20%E2%80%9CID%E2%80%9D%20when%20it%20is%20short%20for%20%E2%80%9Cidentifier%E2%80%9D

(This is a page I had meant to more broadly link people to for guidance for loads of stuff)

internal/ast/diagnostic.go
```diff
@@ -0,0 +1,76 @@
+package ast
+
+import (
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+// Diagnostic
+
+type Diagnostic struct {
```
jakebailey, line 10: I really don't like how this is in `ast`; it has nothing to do with the AST. It has to because it references `SourceFile`, presumably, but can we leave this in `compiler` and have things be fine?

internal/ast/checkflags.go
jakebailey: Can we fix this now without breaking anything?

internal/core/text.go
```diff
@@ -0,0 +1,32 @@
...
+
+// TextPos
+
+type TextPos int32
+
+// TextRange
+
+type TextRange struct {
+	Pos_ TextPos
+	End_ TextPos
```
jakebailey: Do we ever actually refer to these?

Our use of int32s and ints and methods and not methods with this type are, in general, weird.

internal/core/set.go
```diff
@@ -0,0 +1,29 @@
+package core
+
+type Set[T comparable] struct {
```
jakebailey: Now we have `core.Set` and `collections.Set`, bleh. Ok, a later problem I guess. Didn't realize this had to move?

internal/core/pool.go
```diff
@@ -0,0 +1,26 @@
...
+func (p *Pool[T]) New() *T {
+	if len(p.data) == cap(p.data) {
+		p.data = make([]T, 0, NextPoolSize(len(p.data)))
+	}
+	index := len(p.data)
+	p.data = p.data[:index+1]
+	return &p.data[index]
+}
+
+func NextPoolSize(size int) int {
```
jakebailey: Does this actually need to be exported?

internal/compiler/utilities.go
```diff
@@ -84,194 +42,124 @@ var nextNodeId atomic.Uint32
...
-	return NewMessageChain(message, args...).addMessageChain(details)
+func chainDiagnosticMessages(details *ast.MessageChain, message *diagnostics.Message, args ...any) *ast.MessageChain {
+	return NewMessageChain(message, args...).AddMessageChain(details)
 }
 
 type OperatorPrecedence int
 
 const (
-	// Expression:
+	// ast.Expression:
```
jakebailey: All of these comment modifications look like victims of regex.

internal/compiler/relater.go
```diff
@@ -1888,13 +1889,13 @@ func (r *Relater) hasExcessProperties(source *Type, target *Type, reportErrors b
...
-					if isJsxAttributes(r.errorNode) || isJsxOpeningLikeElement(r.errorNode) || isJsxOpeningLikeElement(r.errorNode.parent) {
+					if ast.IsJsxAttributes(r.errorNode) || isJsxOpeningLikeElement(r.errorNode) || isJsxOpeningLikeElement(r.errorNode.Parent) {
 						// !!!
-						// // JsxAttributes has an object-literal flag and undergo same type-assignablity check as normal object-literal.
+						// // ast.JsxAttributes has an object-literal flag and undergo same type-assignablity check as normal object-literal.
 						// // However, using an object-literal error message will be very confusing to the users so we give different a message.
 						// if prop.valueDeclaration && isJsxAttribute(prop.valueDeclaration) && getSourceFileOfNode(errorNode) == getSourceFileOfNode(prop.valueDeclaration.name) {
 						// 	// Note that extraneous children (as in `<NoChild>extra</NoChild>`) don't pass this check,
-						// 	// since `children` is a SyntaxKind.PropertySignature instead of a SyntaxKind.JsxAttribute.
+						// 	// since `children` is a ast.Kind.PropertySignature instead of a ast.Kind.JsxAttribute.
```
jakebailey: Seems like redundant dots, also a victim of regexes

internal/compiler/program.go
```diff
@@ -262,10 +265,10 @@ func (p *Program) collectExternalModuleReferences(file *SourceFile) {
 	// 	}
 	// }
 	// /** Returns a token if position is in [start-of-leading-trivia, end), includes JSDoc only in JS files */
-	// function getNodeAtPosition(sourceFile: SourceFile, position: number): Node {
-	// 	let current: Node = sourceFile;
-	// 	const getContainingChild = (child: Node) => {
-	// 		if (child.pos <= position && (position < child.end || (position === child.end && (child.kind === SyntaxKind.EndOfFileToken)))) {
+	// function getNodeAtPosition(sourceFile: ast.SourceFile, position: number): ast.Node {
```
jakebailey: Also modified comment

internal/compiler/parser.go
```diff
@@ -1917,30 +1918,30 @@ func (p *Parser) parseImportSpecifier() *Node {
...
+	// ast.ImportSpecifier:
 	//   BindingIdentifier
-	//   ModuleExportName as BindingIdentifier
-	// ExportSpecifier:
-	//   ModuleExportName
-	//   ModuleExportName as ModuleExportName
+	//   ast.ModuleExportName as BindingIdentifier
+	// ast.ExportSpecifier:
+	//   ast.ModuleExportName
+	//   ast.ModuleExportName as ast.ModuleExportName
```
jakebailey: Also seems like an oops

internal/compiler/parser.go
```diff
@@ -2744,81 +2745,81 @@ func (p *Parser) parseTypeParameter() *Node {
...
 
-func (p *Parser) parseParameters(flags ParseFlags) []*Node {
+func (p *Parser) parseParameters(flags ParseFlags) []*ast.Node {
 	// FormalParameters [Yield,Await]: (modified)
 	//      [empty]
 	//      FormalParameterList[?Yield,Await]
 	//
 	// FormalParameter[Yield,Await]: (modified)
-	//      BindingElement[?Yield,Await]
+	//      ast.BindingElement[?Yield,Await]
```
jakebailey: Here and below (I'll stop commenting on these)

internal/core/pool.go
weswigham: Yea, there's a utility function that uses it for predicting some sizes

internal/core/set.go
weswigham: Yeah, blame `SourceFile` for referencing a lot of types it probably shouldn't XD

internal/core/text.go
weswigham: Yes we do refer to them, sometimes. Most times, even. I'd call the method accessors less-used. Unifying the accessor-vs-method style is, imo, something we'd need to apply enforcement on with followup work.

internal/core/pool.go
jakebailey: That's very silly but not something to fix in this PR I guess

internal/ast/diagnostic.go
weswigham: It cannot be left in `compiler` - it and `SourceFile` are a matched pair as they reference one another, and since diagnostics are in mutable slices, dealing with converting `[]any` to `[]*Diagnostic` is a massive PITA.

jakebailey: What about `core`, given we'll want to use it for other things, and then we make the File field `any` or some other interface that represents why this field would even be used?

internal/compiler/parser.go
weswigham: Removed.

internal/ast/checkflags.go
weswigham: Yes, but shouldn't that just be a quick 2 character diff followup?

internal/ast/diagnostic.go
weswigham: It's probably a reasonable followup to reduce the couling between the diagnostic datastructures and the AST. While you're thinking about it, though, know that the diagnostic constructor functions are still in `compiler` because they _use the scanner_, so you might need to tease apart quite a bit to actually perfectly isolate all diagnostic-related stuff.

jakebailey:
Theoretically this all looks fine, but my remaining concerns are:

- `Diagnostic` moving to `ast` because of the cycle. I really don't know if this is a good idea. I'd almost prefer less safety (make `Diagnostic` hold onto a `SourceFile` by reference), or come up with an interface which encapsulates what we even need the `SourceFile` for.
  - Diagnostic also brings along chaining 😦
- The underscore suffixed fields are a shame; it's probably fine for now but we need to settle on a better method for this overall IMO.

internal/core/set.go
jakebailey: I don't mind `core.Set`; a self-initializing set is sort of sweet. Gotta move sometime. https://github.com/microsoft/typescript-go/pull/70 would resolve the naming.

weswigham:
Yeah, I'd rather not have `Diagnostic` in `ast` either, since its' coupling is one of accidental convenience, but reworking it seems like it could pretty easily be a followup - it aughta have been in the `diagnostics` package already, more like than not, but was in `compiler`, likely for the same reason I'm moving it with the `ast`.

weswigham:
> The underscore suffixed fields are a shame; it's probably fine for now but we need to settle on a better method for this overall IMO.

Yeah, I'm hoping most of those go away, either because we start lint-enforcing and codegen'ing methods, or because we settle and agree on using struct members instead of methods. I don't care which, but we're *very* mixed right now.

jakebailey:
For `Diagnostic`, a solution suggested by @ahejlsberg is to drop the diags off of SourceFile and instead we store them on the program, returning from parse a result struct with both. That'd break the cycle, at the cost of Program needing to hold onto those results.

weswigham:
Yeah, except it's not quite so easy because `checker` doesn't know about programs, and grammar checking looks at source files to see if they have a parse error to skip the grammar errors... They kinda have to live on `SourceFile` unless we do some pretty significant movements and refactors.

weswigham:
`Diagnostic` should absolutely move away from `ast`, but moving it is going to be a substantive, functional change to multiple existing APIs - it should not be part of this PR, it should absolutely be a separate change to remove that coupling, since it's going to touch a few places and APIs to retain current functionality.

ahejlsberg:
I'm fine with `Diagnostic` and `MessageChain` being in `ast`.

We should move the `NewDiagnostic` and `NewMessageChain` functions to `ast` as well. They don't depend on the scanner, and once we move them, there's really no reason for the properties to be public.

I definitely don't like the underscored property names. We should change those to private properties with a `SetXXX()` method where necessary.

ahejlsberg:
Specifically, regarding field names, we should either have an exported field `Foo` with no accessors, or a private field `foo` with `Foo()` and `SetFoo(...)` methods (or no setter if the data is immutable). For now, I would just apply this to the places where you added underscores to the field names.

internal/ast/ast.go
```diff
@@ -661,13 +661,13 @@ func (node *ModifiersBase) Modifiers() *ModifierListNode { return node.modifiers
 // LocalsContainerBase
 
 type LocalsContainerBase struct {
-	locals        SymbolTable // Locals associated with node (initialized by binding)
-	nextContainer *Node       // Next container in declaration order (initialized by binding)
+	Locals_       SymbolTable // Locals associated with node (initialized by binding)
```
jakebailey: This one may have been missed.

internal/ast/ast.go
```diff
@@ -4316,18 +4327,22 @@ func (node *SourceFile) Path() string {
...
+}
+
 func (node *SourceFile) Diagnostics() []*Diagnostic {
-	return node.diagnostics
+	return node.Diagnostics_
 }
 
 func (node *SourceFile) BindDiagnostics() []*Diagnostic {
-	return node.bindDiagnostics
+	return node.BindDiagnostics_
```
jakebailey: This one too, maybe.

internal/compiler/binder.go
```diff
@@ -78,52 +79,52 @@ const (
...
-	result.flags = flags
-	result.name = name
+	result.Flags = flags
+	result.Name = name
 	return result
 }
 
 /**
- * Declares a Symbol for the node and adds it to symbols. Reports errors for conflicting identifier names.
+ * Declares a ast.Symbol for the node and adds it to symbols. Reports errors for conflicting identifier names.
```
jakebailey: A few more comments got changed that should be reverted too.

jakebailey:
Oops, seems like the locals change is broken due to the map not being initialized.

weswigham:
> Oops, seems like the locals change is broken due to the map not being initialized.

The ouroboros of helper methods and pointers is probably pointing at the wrong thing somewhere. Lemme try and rework it.

internal/compiler/utilities.go
```diff
@@ -3711,7 +3711,7 @@ func getExports(symbol *ast.Symbol) ast.SymbolTable {
 }
 
 func getLocals(container *ast.Node) ast.SymbolTable {
-	return getSymbolTable(&container.LocalsContainerData().Locals_)
+	return container.Locals()
```
ahejlsberg: This isn't going to work, `getSymbolTable` actually initializes the referenced location with a new map.

ahejlsberg: I'm fine with losing the `getSymbolTable` helper and just inlining the initialization using a call to `SetLocals`.

weswigham: Yeah, with how we initialize this lazily, I think this really wants to be a public field.

internal/compiler/utilities.go
```diff
@@ -3711,7 +3711,12 @@ func getExports(symbol *ast.Symbol) ast.SymbolTable {
 }
 
 func getLocals(container *ast.Node) ast.SymbolTable {
-	return getSymbolTable(&container.LocalsContainerData().Locals_)
+	data := container.LocalsContainerData().Locals()
+	if data == nil {
+		container.LocalsContainerData().SetLocals(make(ast.SymbolTable))
+		return container.LocalsContainerData().Locals()
+	}
+	return data
```
jakebailey:
```suggestion
	data := container.LocalsContainerData().Locals()
	if data == nil {
		data = make(ast.SymbolTable)
		container.LocalsContainerData().SetLocals(data)
	}
	return data
```

PR #67, opened 2024-11-12
Add utils for module resolution

andrewbranch:
I extracted this out of my module resolution PR to make it easier to review, and less likely to keep collecting merge conflicts as I finish it up.

### DIFF
--- a/internal/tspath/path_test.go
+++ b/internal/tspath/path_test.go
@@ -298,6 +298,7 @@ func TestGetRelativePathToDirectoryOrUrl(t *testing.T) {
 }
 
 func TestToFileNameLowerCase(t *testing.T) {
+	t.Parallel()
 	assert.Equal(t, toFileNameLowerCase("/user/UserName/projects/Project/file.ts"), "/user/username/projects/project/file.ts")
 	assert.Equal(t, toFileNameLowerCase("/user/UserName/projects/projectß/file.ts"), "/user/username/projects/projectß/file.ts")
 	assert.Equal(t, toFileNameLowerCase("/user/UserName/projects/İproject/file.ts"), "/user/username/projects/İproject/file.ts")


cmd/tsgo/main.go
```diff
@@ -74,25 +74,16 @@ func main() {
...
 			}
 
 			var output strings.Builder
 			formatOpts := ts.DiagnosticsFormattingOptions{
-				NewLine:              "\n",
-				CurrentDirectory:     currentDirectory,
-				GetCanonicalFileName: getCanonicalFileName,
+				NewLine:                   "\n",
+				CurrentDirectory:          currentDirectory,
+				UseCaseSensitiveFileNames: useCaseSensitiveFileNames,
```
andrewbranch: I’m proposing we stop passing around `getCanonicalFileName` and just use `ignoreCase` or `useCaseSensitiveFileNames` everywhere. Is there any issue with that?

internal/tspath/path.go
```diff
@@ -421,3 +438,137 @@ func getRelativePathToDirectoryOrUrl(directoryPathOrUrl string, relativeOrAbsolu
...
 }
+
+func PathIsRelative(path string) bool {
+	return core.MakeRegexp(`^\.\.?(?:$|[\\/])`).MatchString(path)
+}
+
+type ComparePathsOptions struct {
+	IgnoreCase       bool
+	CurrentDirectory string
+}
```
andrewbranch: These are common to many functions that need to check for equality in paths

cmd/tsgo/main.go
jakebailey: What if the filesystem interface had a func on it which canonicalized filenames, would that be less irritating, or more?

internal/compiler/pkg.go
```diff
@@ -1,2 +1,5 @@
 // Package compiler implements the TypeScript compiler.
 package compiler
+
+const VersionMajorMinor = "6.0"
+const Version = "6.0.0-dev"
```
jakebailey: I assume this is needed for typesVersions resolution?

internal/core/core.go
```diff
@@ -225,3 +225,20 @@ func EquateStringCaseInsensitive(a, b string) bool {
...
+	if a == b {
+		return 0
+	}
+	if a == "" {
+		return -1
+	}
+	if b == "" {
+		return 1
+	}
+	return strings.Compare(strings.ToLower(a), strings.ToLower(b))
```
jakebailey: lowercase string comparisons 🙈

cmd/tsgo/main.go
andrewbranch: I’m not sure where `getCanonicalFileName` originally came from, but it feels very suspicious to me that we let hosts define that function _and_ we also rely on `useCaseSensitiveFileNames` all over the place. If an invariant of our API usage is that `getCanonicalFileName` should only be an identity function or a to-lower function, then passing it around is clearly redundant, and we should pass around a `bool` instead. If `getCanonicalFileName` is allowed to be any string→string mapping, then we have _so_ many false assumptions littered throughout our current codebase.

internal/compiler/pkg.go
andrewbranch: Yep

jakebailey: Probably should be in `core` then, maybe `core/version.go`.

jakebailey: (Or, some other package maybe)

internal/testutil/baseline/baseline.go
```diff
@@ -14,50 +15,55 @@ type Options struct {
...
-	localFileName := localPath(relativeFileName, opts.Subfolder)
-	referenceFileName := referencePath(relativeFileName, opts.Subfolder)
+	localFileName = localPath(relativeFileName, opts.Subfolder)
+	referenceFileName = referencePath(relativeFileName, opts.Subfolder)
 	expected := getExpectedContent(relativeFileName, opts)
 	if _, err := os.Stat(localFileName); err == nil {
 		if err := os.Remove(localFileName); err != nil {
-			return fmt.Errorf("Failed to remove the local baseline file %s: %w", localFileName, err)
+			t.Fatal(fmt.Errorf("failed to remove the local baseline file %s: %w", localFileName, err))
+			return
```
jakebailey: No need to return after `t.Fatal` unless you're trying to do some special optimization; this is guaranteed to exit.

internal/testutil/baseline/error_baseline.go
```diff
@@ -24,10 +24,13 @@ var (
 	tsExtension   = regexp.MustCompile(`\.tsx?$`)
 )
 
+// Ensure tests work on windows and linux
+var useCaseSensitiveFileNames = false
```
jakebailey: How does this ensure tests work in both?

andrewbranch: I don’t know; this was adapted from https://github.com/microsoft/typescript-go/pull/67/files#diff-762b54c7dd8d0945b1c8044aa225380fda7237a163dbcebced59b08b06a068efL15-L16

internal/compiler/error_reporting.go
```diff
@@ -199,7 +199,7 @@ func WriteLocation(output *strings.Builder, file *SourceFile, pos int, formatOpt
 	firstLine, firstChar := GetLineAndCharacterOfPosition(file, pos)
 	var relativeFileName string
 	if formatOpts != nil {
-		relativeFileName = tspath.ConvertToRelativePath(file.path, formatOpts.CurrentDirectory, formatOpts.GetCanonicalFileName)
+		relativeFileName = tspath.ConvertToRelativePath(file.path, tspath.ComparePathsOptions{IgnoreCase: !formatOpts.UseCaseSensitiveFileNames, CurrentDirectory: formatOpts.CurrentDirectory})
```
jakebailey: Should `formatOpts` just contain `ComparePathsOptions`?

internal/testutil/baseline/error_baseline.go
andrewbranch: But, it makes sense that you don’t want it to actually be dependent on the platform that tests are running

jakebailey: Oh, like, this is a default value instead of a default implementation of using ToLower.

I guess it's fine for now but may as well be deleted given that's the default formatting option anyway...

internal/compiler/error_reporting.go
andrewbranch: We tend to expose `UseCaseSensitiveFileNames` in high-level types and `IgnoreCase` in low-level functions. Using `IgnoreCase` doesn’t make a lot of sense here at this level. I wonder if `ComparePathsOptions` should have `UseCaseSensitiveFileNames` instead of `IgnoreCase`...

jakebailey: Whatever you'd like, not a big deal at this moment, just opining.

gabritto: I think ultimately it would be best if all the booleans were either all `IgnoreCase` or all `UseCaseSensitiveFileNames`, so we don't confuse the two, but maybe that would be annoying during porting.

andrewbranch: That would have implications (possibly good?) for the zero value of `ComparePathsOptions{}` though...

jakebailey: You'll have to update the literals to the right syntax for embedding, it seems

andrewbranch: @gabritto I think anything that’s explicitly dealing with paths can use `UseCaseSensitiveFileNames`, and we only swap to `IgnoreCase` by the time we get to functions like `GetStringComparer`, which isn’t file-name-specific. Latest commits move in that direction.

PR #68, opened 2024-11-13
TypeFacts, enum types, constant evaluator, more type relations

ahejlsberg:
This PR ports TypeFacts, enum types, constant evaluator, and more type relations.

### DIFF
--- a/internal/compiler/relater.go
+++ b/internal/compiler/relater.go
@@ -1542,11 +1542,12 @@ func (c *Checker) sliceTupleType(t *Type, index int, endSkipCount int) *Type {
 
 func (c *Checker) getKnownKeysOfTupleType(t *Type) *Type {
 	fixedLength := t.TargetTupleType().fixedLength
-	keys := make([]*Type, fixedLength)
+	keys := make([]*Type, fixedLength+1)
 	for i := range fixedLength {
 		keys[i] = c.getStringLiteralType(strconv.Itoa(i))
 	}
-	return c.getUnionType(append(keys, c.getIndexType(ifElse(t.TargetTupleType().readonly, c.globalReadonlyArrayType, c.globalArrayType))))
+	keys[fixedLength] = c.getIndexType(ifElse(t.TargetTupleType().readonly, c.globalReadonlyArrayType, c.globalArrayType))
+	return c.getUnionType(keys)
 }
 
 func (c *Checker) getRestArrayTypeOfTupleType(t *Type) *Type {
@@ -1750,8 +1751,7 @@ func (c *Checker) compareSignaturesIdentical(source *Signature, target *Signatur
 			}
 		}
 	}
-	targetLen := c.getParameterCount(target)
-	for i := 0; i < targetLen; i++ {
+	for i := range c.getParameterCount(target) {
 		s := c.getTypeAtPosition(source, i)
 		t := c.getTypeAtPosition(target, i)
 		related := compareTypes(t, s)


internal/compiler/relater.go
```diff
@@ -2359,13 +2505,72 @@ func (r *Relater) restoreErrorState(e errorState) {
...
+	}
 	if result != TernaryFalse {
 		r.restoreErrorState(saveErrorState)
 	}
 	return result
 }
 
+func (r *Relater) isSourceIntersectionNeedingExtraCheck(source *Type, target *Type) bool {
+	return source.flags&TypeFlagsIntersection != 0 && r.c.getApparentType(source).flags&TypeFlagsStructuredType != 0 &&
+		!utils.Some(source.Types(), func(t *Type) bool {
```
jakebailey:
```suggestion
		!core.Some(source.Types(), func(t *Type) bool {
```

internal/compiler/relater.go
```diff
@@ -3338,7 +3678,130 @@ func (r *Relater) signaturesIdenticalTo(source *Type, target *Type, kind Signatu
 }
 
 func (r *Relater) indexSignaturesRelatedTo(source *Type, target *Type, sourceIsPrimitive bool, reportErrors bool, intersectionState IntersectionState) Ternary {
-	return TernaryTrue // !!!
+	if r.relation == r.c.identityRelation {
+		return r.indexSignaturesIdenticalTo(source, target)
+	}
+	indexInfos := r.c.getIndexInfosOfType(target)
+	targetHasStringIndex := utils.Some(indexInfos, func(info *IndexInfo) bool { return info.keyType == r.c.stringType })
```
jakebailey:
```suggestion
	targetHasStringIndex := core.Some(indexInfos, func(info *IndexInfo) bool { return info.keyType == r.c.stringType })
```

internal/compiler/relater.go
```diff
@@ -3338,7 +3678,130 @@ func (r *Relater) signaturesIdenticalTo(source *Type, target *Type, kind Signatu
...
+	return TernaryFalse
+}
+
+/**
+ * Return true if type was inferred from an object literal, written as an object type literal, or is the shape of a module
+ * with no call or construct signatures.
+ */
+func (c *Checker) isObjectTypeWithInferableIndex(t *Type) bool {
+	if t.flags&TypeFlagsIntersection != 0 {
+		return utils.Every(t.Types(), c.isObjectTypeWithInferableIndex)
```
jakebailey:
```suggestion
		return core.Every(t.Types(), c.isObjectTypeWithInferableIndex)
```

internal/compiler/utilities.go
```diff
@@ -4023,3 +4038,144 @@ func isObjectLiteralElementLike(node *Node) bool {
...
+	evaluateTemplateExpression := func(expr *Node, location *Node) EvaluatorResult {
+		result := expr.AsTemplateExpression().head.Text()
+		resolvedOtherFiles := false
+		hasExternalReferences := false
+		for _, span := range expr.AsTemplateExpression().templateSpans {
+			spanResult := evaluate(span.Expression(), location)
+			if spanResult.value == nil {
+				return evaluatorResult(nil, true /*isSyntacticallyString*/, false, false)
+			}
+			result += anyToString(spanResult.value)
```
jakebailey: Not sure if it'll matter but this is a string concat in a loop which may blow up on larger template strings.

internal/compiler/checker.go
```diff
@@ -417,6 +510,7 @@ func NewChecker(program *Program) *Checker {
 	c.comparableRelation = &Relation{}
 	c.identityRelation = &Relation{}
 	c.enumRelation = &Relation{}
+	c.getGlobalNonNullableTypeAlias = utils.Memoize(func() *Symbol { return c.getGlobalSymbol("NonNullable", SymbolFlagsTypeAlias, nil /*diagnostic*/) })
```
```suggestion
	c.getGlobalNonNullableTypeAlias = core.Memoize(func() *Symbol { return c.getGlobalSymbol("NonNullable", SymbolFlagsTypeAlias, nil /*diagnostic*/) })
```

I will note that `sync.OnceValue` is also an option, though that one is slightly heavier just because it also ensures that panics are handled.

internal/compiler/relater.go
```diff
@@ -2539,11 +2744,110 @@ func (r *Relater) structuredTypeRelatedToWorker(source *Type, target *Type, repo
...
+			}
+		}
+		if reportErrors {
+			originalErrorChain = nil
+		}
 	case target.flags&TypeFlagsIndex != 0:
+		targetType := target.AsIndexType().target
+		// A keyof S is related to a keyof T if T is related to S.
+		if source.flags&TypeFlagsIndex != 0 {
+			result = r.isRelatedTo(targetType, source.AsIndexType().target, RecursionFlagsBoth /*reportErrors*/, false)
```
jakebailey:
```suggestion
			result = r.isRelatedTo(targetType, source.AsIndexType().target, RecursionFlagsBoth, false /*reportErrors*/)
```

ts-to-go is updated to do that too if you wanted to work from a newer conversion.

internal/compiler/relater.go
```diff
@@ -3338,7 +3678,130 @@ func (r *Relater) signaturesIdenticalTo(source *Type, target *Type, kind Signatu
...
+		var related Ternary
+		switch {
+		case r.relation != r.c.strictSubtypeRelation && !sourceIsPrimitive && targetHasStringIndex && targetInfo.valueType.flags&TypeFlagsAny != 0:
+			related = TernaryTrue
+		case r.c.isGenericMappedType(source) && targetHasStringIndex:
+			related = r.isRelatedTo(r.c.getTemplateTypeFromMappedType(source), targetInfo.valueType, RecursionFlagsBoth, reportErrors)
+		default:
+			related = r.typeRelatedToIndexInfo(source, targetInfo, reportErrors, intersectionState)
+		}
+		if !(related != 0) {
```
jakebailey: I'll try and fix this in the converter in the future 😅

ahejlsberg:
@jakebailey Feedback addressed.

internal/compiler/relater.go
```diff
@@ -1532,6 +1540,15 @@ func (c *Checker) sliceTupleType(t *Type, index int, endSkipCount int) *Type {
 	return c.createTupleTypeEx(c.getTypeArguments(t)[index:endIndex], target.elementInfos[index:endIndex], false /*readonly*/)
 }
 
+func (c *Checker) getKnownKeysOfTupleType(t *Type) *Type {
+	fixedLength := t.TargetTupleType().fixedLength
+	keys := make([]*Type, fixedLength)
+	for i := range fixedLength {
+		keys[i] = c.getStringLiteralType(strconv.Itoa(i))
+	}
+	return c.getUnionType(append(keys, c.getIndexType(ifElse(t.TargetTupleType().readonly, c.globalReadonlyArrayType, c.globalArrayType))))
```
jakebailey: This is making the linter unhappy because this _looks_ like an accidental side-effect (though in this case, it immediately returns); I would write this append on the previous line.

internal/compiler/relater.go
```diff
@@ -1691,8 +1708,139 @@ func (c *Checker) isResolvingReturnTypeOfSignature(signature *Signature) bool {
...
+				related := compareTypes(sourceThisType, targetThisType)
+				if related == TernaryFalse {
+					return TernaryFalse
+				}
+				result &= related
+			}
+		}
+	}
+	targetLen := c.getParameterCount(target)
+	for i := 0; i < targetLen; i++ {
```
jakebailey: This is a style thing since `for i := range targetLen` is equivalent; if you think this is going to be too overbearing for the port, we can disable this lint (it's just good practice).

You can run linting yourself locally by using the config options in the new template and installing the tools with `hereby install-tools` and `hereby lint`.

jakebailey: (I should also be able to rewrite loops of this form in ts-to-go too)

ahejlsberg: Actually, I like int ranges, just hadn't caught that one.

internal/compiler/relater.go
jakebailey: Oh, you're right, this was actually a bugish, in that you preallocated and assigned and then the append was a wasted copy. Nice, not a bad lint then

PR #69, opened 2024-11-13
Port of `SourceMapGenerator`, adds `sourcemap` package

rbuckton:
This is a port of the `SourceMapGenerator` from the TS compiler codebase to Go. I've opted to put it into its own subpackage as it is fairly isolated from other functionality. In the future we may likely also port `DocumentPositionMapper` to support the language service, which will also live in this subpackage.

### DIFF
--- a/internal/collections/ordered_map.go
+++ b/internal/collections/ordered_map.go
@@ -14,27 +14,27 @@ import (
 	"github.com/go-json-experiment/json/jsontext"
 )
 
-// Map is an insertion ordered map.
-type Map[K comparable, V any] struct {
+// OrderedMap is an insertion ordered map.
+type OrderedMap[K comparable, V any] struct {
 	keys []K
 	mp   map[K]V
 }
 
-// NewMapWithSizeHint creates a new Map with a hint for the number of elements it will contain.
-func NewMapWithSizeHint[K comparable, V any](hint int) *Map[K, V] {
+// NewOrderedMapWithSizeHint creates a new OrderedMap with a hint for the number of elements it will contain.
+func NewOrderedMapWithSizeHint[K comparable, V any](hint int) *OrderedMap[K, V] {
 	m := newMapWithSizeHint[K, V](hint)
 	return &m
 }
 
-func newMapWithSizeHint[K comparable, V any](hint int) Map[K, V] {
-	return Map[K, V]{
+func newMapWithSizeHint[K comparable, V any](hint int) OrderedMap[K, V] {
+	return OrderedMap[K, V]{
 		keys: make([]K, 0, hint),
 		mp:   make(map[K]V, hint),
 	}
 }
 
 // Set sets a key-value pair in the map.
-func (m *Map[K, V]) Set(key K, value V) {
+func (m *OrderedMap[K, V]) Set(key K, value V) {
 	if m.mp == nil {
 		m.mp = make(map[K]V)
 	}
@@ -46,24 +46,24 @@ func (m *Map[K, V]) Set(key K, value V) {
 }
 
 // Get retrieves a value from the map.
-func (m *Map[K, V]) Get(key K) (V, bool) {
+func (m *OrderedMap[K, V]) Get(key K) (V, bool) {
 	v, ok := m.mp[key]
 	return v, ok
 }
 
 // GetOrZero retrieves a value from the map, or returns the zero value of the value type if the key is not present.
-func (m *Map[K, V]) GetOrZero(key K) V {
+func (m *OrderedMap[K, V]) GetOrZero(key K) V {
 	return m.mp[key]
 }
 
 // Has returns true if the map contains the key.
-func (m *Map[K, V]) Has(key K) bool {
+func (m *OrderedMap[K, V]) Has(key K) bool {
 	_, ok := m.mp[key]
 	return ok
 }
 
 // Delete removes a key-value pair from the map.
-func (m *Map[K, V]) Delete(key K) (V, bool) {
+func (m *OrderedMap[K, V]) Delete(key K) (V, bool) {
 	v, ok := m.mp[key]
 	if !ok {
 		var zero V
@@ -90,7 +90,7 @@ func (m *Map[K, V]) Delete(key K) (V, bool) {
 
 // Keys returns an iterator over the keys in the map.
 // A slice of the keys can be obtained by calling `slices.Collect`.
-func (m *Map[K, V]) Keys() iter.Seq[K] {
+func (m *OrderedMap[K, V]) Keys() iter.Seq[K] {
 	return func(yield func(K) bool) {
 		for _, key := range m.keys {
 			if !yield(key) {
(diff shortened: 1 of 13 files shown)


internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,356 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type SourceIndex int
+type NameIndex int
```
jakebailey, line 13: These two appear to not be used in the package's public API, so could be unexported.

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func NewSourceMapGenerator(file string, sourceRoot string, sourcesDirectoryPath string, currentDirectory string, getCanonicalFileName func(string) string) *SourceMapGenerator {
+	gen := &SourceMapGenerator{}
+	gen.file = file
+	gen.sourceRoot = sourceRoot
+	gen.sourcesDirectoryPath = sourcesDirectoryPath
+	gen.currentDirectory = currentDirectory
+	gen.getCanonicalFileName = getCanonicalFileName
+	return gen
```
jakebailey: I would prefer writing code like:

```go
return &SourceMapGenerator{
    // ...
}
```

The AST only does it the other way due to the complication of embedding.

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,356 @@
...
+	Version        int        `json:"version"`
+	File           string     `json:"file"`
+	SourceRoot     string     `json:"sourceRoot,omitempty"`
+	Sources        []string   `json:"sources"`
+	Names          []string   `json:"names,omitempty"`
+	Mappings       string     `json:"mappings"`
+	SourcesContent []*string  `json:"sourcesContent,omitempty"`
+}
+
+func NewSourceMapGenerator(file string, sourceRoot string, sourcesDirectoryPath string, currentDirectory string, getCanonicalFileName func(string) string) *SourceMapGenerator {
```
jakebailey: Can you use an options struct for these input? Reading the tests, it's not mega clear what is what from just reading it, so an options struct would be nice for reading.

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+type RawSourceMap struct {
+	Version        int        `json:"version"`
+	File           string     `json:"file"`
+	SourceRoot     string     `json:"sourceRoot,omitempty"`
+	Sources        []string   `json:"sources"`
+	Names          []string   `json:"names,omitempty"`
+	Mappings       string     `json:"mappings"`
+	SourcesContent []*string  `json:"sourcesContent,omitempty"`
```
jakebailey: You could probably use Andrew's Optional struct here if that's what you're emulating. (But a pointer is fine if it's achieving the right result with Go's JSON lib)

internal/sourcemap/sourcemap_generator.go
jakebailey: No, I'm wrong, sorry; just noticed them.

rbuckton: They are though? `AddSource` and `AddName` return them, and `AddMappingSource`/`AddMappingSourceName` accept them.

internal/sourcemap/sourcemap_generator.go
rbuckton: It's working as intended currently.

internal/sourcemap/sourcemap_generator.go
jakebailey: Yeah, see above (sorry)

internal/sourcemap/sourcemap_generator.go
rbuckton: I switched it to use `tspath.ComparePathOptions` for `currentDirectory`/`getCanonicalFileName`. For the other three parameters, is it really worth introducing a struct to create a struct?

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,353 @@
...
+		gen.lastGeneratedLine != gen.pendingGeneratedLine ||
+		gen.lastGeneratedCharacter != gen.pendingGeneratedCharacter ||
+		gen.lastSourceIndex != gen.pendingSourceIndex ||
+		gen.lastSourceLine != gen.pendingSourceLine ||
+		gen.lastSourceCharacter != gen.pendingSourceCharacter ||
+		gen.lastNameIndex != gen.pendingNameIndex
+}
+
+func (gen *SourceMapGenerator) appendMappingCharCode(charCode rune) {
+	gen.mappings.Grow(1)
```
jakebailey: No need to grow here; it's only by one so that'll happen anyway with the rune write. Usually grow is for "I know ahead of time roughly how long the entire string will be"

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,353 @@
...
+
+	var sourcesContent []*string
+	if gen.sourcesContent != nil {
+		sourcesContent = append([]*string{}, gen.sourcesContent...)
+	}
+
+	var names []string
+	if gen.names != nil {
+		names = append([]string{}, gen.names...)
+	}
```
jakebailey: I think these can all just be `slices.Clone`; the case where you don't want `nil` (`sources`) can just be an if statement after the clone to make it empty instead.

rbuckton: Isn't this essentially what `slices.Clone` does behind the scenes?

jakebailey: Yes, which is why I wouldn't write it out by hand

jakebailey: Using the name `slices.Clone` also makes it clear what the intent is here, rather than remembering that append on an empty list gives back a copy....

internal/sourcemap/sourcemap_generator.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// Gets the source map as a `RawSourceMap` object
+func (gen *SourceMapGenerator) RawSourceMap() *RawSourceMap {
+	gen.commitPendingMapping()
+	return &RawSourceMap{
+		Version:        3,
+		File:           gen.file,
+		SourceRoot:     gen.sourceRoot,
+		Sources:        append([]string{}, gen.sources...),
```
jakebailey: This random one not being a clone seems odd; I would have probably written it as a conditional above here to say "if cloning really is nil, set it to empty", or something... But I guess it's okay

internal/sourcemap/sourcemap_generator.go
jakebailey: I guess not; I was also trying to decide if it was even worth having a `New` func at all, and whether or not you could simply make the fields exported and require people to construct the value directly.

rbuckton: I'd rather leave the constructor as we also do some tracing in the TS compiler that gets set up in the constructor. I plan to add that later.

PR #71, opened 2024-11-13
Move CompilerOptions to core

andrewbranch:


internal/compiler/pkg.go
```diff
@@ -1,2 +1,5 @@
 // Package compiler implements the TypeScript compiler.
 package compiler
+
+const VersionMajorMinor = "6.0"
+const Version = "6.0.0-dev"
```
jakebailey: Bad merge?

internal/compiler/parser.go
```diff
@@ -3711,20 +3711,20 @@ func (p *Parser) isParenthesizedArrowFunctionExpression() Tristate {
 		// ERROR RECOVERY TWEAK:
 		// If we see a standalone => try to parse it as an arrow function expression as that's
 		// likely what the user intended to write.
-		return TSTrue
+		return core.TSTrue
```
jakebailey: I do kinda wish we would make a `tristate` package, give how often we're writing `core.TSTrue`, which would be nice to read as `tristate.True`. Obviously this doesn't scale, but this may be one of a few that would be nice. Not hard to change later though.

andrewbranch:
Either order is fine with me. The only difference I see here is I’ve grouped all the `CompilerOptions` property types together in one file, and generated a single stringer file for the whole `core` package.

weswigham:
> The only difference I see here is I’ve grouped all the CompilerOptions property types together in one file, and generated a single stringer file for the whole core package.

I liked per-file enums and stringer lines since it makes it very clear which enums have stringer support already and which do not when you're reading their source. Kinda like a decorator, but, y'know, in a comment.

jakebailey:
I generally agree with more than one file, but we may end up with a lot of files. I personally do not care about that so don't mind, but others might.

andrewbranch:
No strong opinion, though another advantage of multiple files is the ability to remove prefixes.

jakebailey:
Generated code needs an update but otherwise LGTM

PR #72, opened 2024-11-13
Add compiler test case parsing function

gabritto:
Adds the function that parses file options and file names from compiler tests, minus the part that detects symlinks and the part that requires parsing tsconfig files, which are commented out for now.


internal/testutil/baseline/test_case_parser.go
jakebailey, line 1: Is `baseline` the correct package for this? It seems like it could be its own thing.

internal/testutil/baseline/test_case_parser.go
```diff
@@ -0,0 +1,177 @@
...
+type testCaseContent struct {
+	settings             compilerSettings
+	testUnitData         []*testUnit
+	tsConfig             any // !!!
+	tsConfigFileUnitData any // !!!
+	symlinks             any // !!!
+}
+
+// Regex for parsing options in the format "@Alpha: Value of any sort"
+var optionRegex = regexp.MustCompile(`(?m)^\/{2}\s*@(\w+)\s*:\s*([^\r\n]*)`) // multiple matches on multiple lines
```
jakebailey: Is this how the old code used to do it? `extractCompilerSettings` seems to just call it on the entire code, so would this parse out after the files start being loaded below?

internal/testutil/baseline/test_case_parser.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Given a test file containing // @FileName directives, return an array of named units of code to be added to an existing compiler instance
+func makeUnitsFromTest(code string, fileName string, settings compilerSettings) testCaseContent {
```
jakebailey: This should probably be exported to be usable elsewhere, right?

internal/testutil/baseline/test_case_parser.go
gabritto: This function will be used by the test runners only and I think it's not enough to be its own package. I'll move it to the `runner` package -- I have created that package in the follow-up code I'm working on.

internal/testutil/baseline/test_case_parser.go
gabritto: I moved it to the package where it'll be used later, so probably it will stay unexported.

PR #74, opened 2024-11-14
Run tests with gotestsum if available, split bench tests out

jakebailey:
https://github.com/gotestyourself/gotestsum is a nice `go test` wrapper; use it if present and install it with `install-tools`.

The tool can't handle benchmark tests, so split that out. I also made it so that `hereby test` does not run every benchmark one time, which will long-term not be needed for local dev work. But `hereby test:all` will do that.

I can also _not_ use colons in task names, and go back to dashes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
 
       - run: npm ci
 
-      - run: npx hereby install-tools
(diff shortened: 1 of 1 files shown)


jakebailey:
Hm, this adds a bunch of time to CI tests because installing the linter with install-tools takes a while too. I'll do something more straightforward for that case I guess...

PR #75, opened 2024-11-14
Lightweight type writer activated by `-t` flag

ahejlsberg:
This PR improves the `-t` flag to activate a lightweight type writer that prints all types in `main.ts`. This is quite similar to our type writer for tests and makes it a lot easier to check that the right types are being generated.

The PR also ensures that `core.TextPos` is used only as a storage format (i.e. for the private properties of `TextRange` and the `[]core.TextPos` used to store line starts). In all other instances, we now consistently use `int` to index and slice strings, be they source text or not.

I did investigate using `core.TextPos` in parameter and local variable types, but it devolves into hundreds of casts and leads to odd statements like `s.pos += core.TextPos(count)` to increment a `TextPos` property by an `int` value. IMO, the noise level isn't worth it and it really doesn't make anything more type safe. After all, our source text string is no different than any other string, and the language itself uses `int` to index strings.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -1399,6 +1399,10 @@ func IsArrayBindingPattern(node *Node) bool {
 	return node.Kind == KindArrayBindingPattern
 }
 
+func IsBindingPattern(node *Node) bool {
+	return node.Kind == KindArrayBindingPattern || node.Kind == KindObjectBindingPattern
+}
+
 // ParameterDeclaration
 
 type ParameterDeclaration struct {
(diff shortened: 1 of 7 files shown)


internal/compiler/scanner.go
```diff
@@ -1065,8 +1065,8 @@ func (s *Scanner) scanString(jsxAttributeString bool) string {
 
 func (s *Scanner) scanTemplateAndSetTokenValue(shouldEmitInvalidEscapeError bool) ast.Kind {
 	startedWithBacktick := s.char() == '`'
-	start := s.pos
 	s.pos++
+	start := s.pos
```
jakebailey, line 1069: Aha, that is how it was before, oops.

jakebailey:
Annoyingly `hereby lint --fix` does not have a fixer for the `unconvert` lint :(

PR #76, opened 2024-11-15
Move some ast utilities from `compiler` to `ast`

rbuckton:
This moves a number of common `ast`-related functions out of `compiler` and into `ast` so that they can be more readily used from other subpackages (such as the emitter)

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -194,7 +194,8 @@ func isFunctionLikeDeclarationKind(kind Kind) bool {
 
 // Determines if a node is function-like (but is not a signature declaration)
 func IsFunctionLikeDeclaration(node *Node) bool {
-	return isFunctionLikeDeclarationKind(node.Kind)
+	// TODO(rbuckton): Move `node != nil` test to call sites
+	return node != nil && isFunctionLikeDeclarationKind(node.Kind)
 }
 
 func isFunctionLikeKind(kind Kind) bool {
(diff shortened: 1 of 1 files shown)


jakebailey:
Er, well, it fails tests 😄

internal/compiler/utilities.go
```diff
@@ -1369,47 +1080,40 @@ func isEmptyObjectLiteral(expression *ast.Node) bool {
 
 func isFunctionSymbol(symbol *ast.Symbol) bool {
 	d := symbol.ValueDeclaration
-	return d != nil && (ast.IsFunctionDeclaration(d) || ast.IsVariableDeclaration(d) && isFunctionLike(d.AsVariableDeclaration().Initializer))
+	if d != nil {
```
jakebailey, line 1083: Hm, how did this end up needing to change if this is a pure code move?

rbuckton: I'd removed the `nil` test as we are fairly inconsistent with those in the original TS codebase. In the long term, I think it'd be better to `nil`-test beforehand for many of these as it isn't always necessary, and we certainly don't always guard everywhere. I can add it back in to `IsFunctionLike` for the time being, but I think it's better to leave the early guards in as well.

internal/ast/utilities.go
```diff
@@ -0,0 +1,477 @@
...
+
+func IsTypeNode(node *Node) bool {
+	return IsTypeNodeKind(node.Kind)
+}
+
+func isJSDocTypeAssertion(node *Node) bool {
+	return false // !!!
+}
+
+type OuterExpressionKinds int16
```
weswigham, line 376: Is there a reason this is plural?

rbuckton: Because it's plural in the TS codebase.

weswigham: ...hah, I never noticed. Weird.

PR #77, opened 2024-11-15
Add ParameterList Node to support emitter

rbuckton:
This adds a `ParameterList` node much like the `TypeParameterList` and `ModifierList` nodes. The `ParameterList` node acts much like `NodeArray` did in the old compiler, providing `pos` and `end` information relating to the parentheses surrounding the parameter list, which the emitter uses to differentiate between a parenthesized arguments list or a single binding identifier in arrow functions: `a => {}` vs `(a) => {}`.


ahejlsberg:
Let's discuss first next time. I considered this as I ported the code and I'd really like to avoid this change. You can already tell the difference with a simple `Parameters == nil` check. We use a `nil` value to indicate a missing parameter list vs. an empty (but non-nil) slice to indicate an empty parameter list. Since pretty much every function-like node has a (possibly empty) parameter list, introducing a new node and associated indirection is costly both in time and space.

rbuckton:
> You can already tell the difference with a simple `Parameters == nil` check. 

I think you may be thinking of `NewExpression`? That's not what this indicates, and a `nil` check isn't the problem. The issue is needing to differentiate between a single-unparenthesized parameter (`a => {}`) vs zero or more parenthesized parameters (`(a) => {}`). If anything, we don't really need list nodes for modifiers or type parameters, as modifiers are never bracketed and type parameters are always bracketed, but we do need them for parameters because they may be either bracketed or unbracketed, which is a property of the list itself.

Without a node, I will have to rescan the source to find the `(` and `)` to properly print leading and trailing comments, for every arrow function that we emit that takes a single parameter

ahejlsberg:
The reason for the dedicated modifiers, type parameters, and type argument nodes is that it reduces the overhead of _not_ having them from 24 bytes (an empty slice) to 8 bytes (a nil pointer). In the vast majority of cases these nodes are missing and we really don't want to spend 24 bytes to indicate something isn't there.

Regarding un-parenthesized single parameters, it would the trivial to add a `NodeFlags` bit.

PR #79, opened 2024-11-15
Move module resolver to package, implement most of ResolveModuleName

andrewbranch:
- Gets module resolution API close to something I'm happy with
  - Moved things to appropriate packages
  - Removed many superfluous types
  - Gave things more Go-y names
- Implements most of `ResolveModuleName`
  - package.json `imports`/`exports` is still missing
- Adds tests generated from TS compiler tests
  - Captured resolution calls/results from all compiler tests except those using `--moduleResolution classic`
  - Ones using `node10` are forced to `bundler`
  - Tests that don't pass yet are skipped

### DIFF
--- a/internal/compiler/module/resolver.go
+++ b/internal/compiler/module/resolver.go
@@ -525,18 +525,17 @@ func (r *resolutionState) createResolvedModuleWithFailedLookupLocationsHandlingS
 
 func (r *resolutionState) createResolvedModuleWithFailedLookupLocations(resolved *resolved, isExternalLibraryImport bool) *ResolvedModuleWithFailedLookupLocations {
 	if r.resultFromCache != nil {
+		var result *ResolvedModuleWithFailedLookupLocations
 		if !r.resolver.cache.isReadonly() {
-			r.resultFromCache.FailedLookupLocations = updateResolutionField(r.resultFromCache.FailedLookupLocations, r.failedLookupLocations)
-			r.resultFromCache.AffectingLocations = updateResolutionField(r.resultFromCache.AffectingLocations, r.affectingLocations)
-			r.resultFromCache.ResolutionDiagnostics = updateResolutionField(r.resultFromCache.ResolutionDiagnostics, r.resultFromCache.ResolutionDiagnostics)
-			return r.resultFromCache
+			result = r.resultFromCache
 		} else {
-			result := *r.resultFromCache
-			result.FailedLookupLocations = initializeResolutionFieldForReadonlyCache(result.FailedLookupLocations, r.failedLookupLocations)
-			result.AffectingLocations = initializeResolutionFieldForReadonlyCache(result.AffectingLocations, r.affectingLocations)
-			result.ResolutionDiagnostics = initializeResolutionFieldForReadonlyCache(result.ResolutionDiagnostics, r.resultFromCache.ResolutionDiagnostics)
-			return &result
+			cloned := *r.resultFromCache
+			result = &cloned
 		}
+		result.FailedLookupLocations = slices.Concat(result.FailedLookupLocations, r.failedLookupLocations)
+		result.AffectingLocations = slices.Concat(result.AffectingLocations, r.affectingLocations)
+		result.ResolutionDiagnostics = slices.Concat(result.ResolutionDiagnostics, r.resultFromCache.ResolutionDiagnostics)
+		return result
 	}
 	var resolvedModule ResolvedModule
 	if resolved != nil {
@@ -939,7 +938,7 @@ func (r *resolutionState) loadNodeModuleFromDirectoryWorker(ext extensions, cand
 		pathPatterns := tryParsePatterns(versionPaths.GetPaths())
 		if result := r.tryLoadModuleUsingPaths(ext, moduleName, candidate, versionPaths.GetPaths(), pathPatterns, loader, onlyRecordFailuresForPackageFile); result.stop {
 			if result.value.packageId.Name != "" {
-				// TODO: are these asserts really necessary?
+				// !!! are these asserts really necessary?
 				panic("expected packageId to be empty")
 			}
 			return result.value
@@ -949,7 +948,7 @@ func (r *resolutionState) loadNodeModuleFromDirectoryWorker(ext extensions, cand
 	if packageFile != "" {
 		if packageFileResult := loader(ext, candidate, onlyRecordFailuresForPackageFile); packageFileResult != nil {
 			if packageFileResult.packageId.Name != "" {
-				// TODO: are these asserts really necessary?
+				// !!! are these asserts really necessary?
 				panic("expected packageId to be empty")
 			}
 			return packageFileResult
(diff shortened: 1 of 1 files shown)


testdata/baselines/reference/module/resolver/allowJsCrossMonorepoPackage.trace.json
jakebailey, line 1: Are these files comparable to something we can run based on the TypeScript source in the submodule to verify, or is that not possible?

internal/compiler/diagnostics/diagnostics.go
```diff
@@ -44,3 +46,11 @@ func (m *Message) Message() string                    { return m.text }
 func (m *Message) ReportsUnnecessary() bool           { return m.reportsUnnecessary }
 func (m *Message) ElidedInCompatabilityPyramid() bool { return m.elidedInCompatabilityPyramid }
 func (m *Message) ReportsDeprecated() bool            { return m.reportsDeprecated }
+
+func Format(message *Message, args ...any) string {
```
jakebailey: Should this be a method on `Message`? Should `FormatStringFromArgs` be just here? Not sure that it's ever actually used other than for diags...

internal/compiler/error_reporting.go
```diff
@@ -236,20 +236,20 @@ func WriteErrorSummaryText(output *strings.Builder, allDiagnostics []*ast.Diagno
 	if totalErrorCount == 1 {
 		// Special-case a single error.
 		if len(errorSummary.GlobalErrors) > 0 {
-			message = formatMessage(diagnostics.Found_1_error)
+			message = diagnostics.Format(diagnostics.Found_1_error)
```
jakebailey: With If `Format` were a method, this could be `diagnostics.Found_1_error.Format()` which would be cool

internal/compiler/module/types.go
```diff
@@ -0,0 +1,175 @@
...
+	Options *core.CompilerOptions
+}
+
+type ResolvedProjectReference struct {
+	CommandLine ParsedCommandLine
+	SourceFile  *ast.SourceFile
+	References  []*ResolvedProjectReference
+}
+
+type PerDirectoryResolutionCache[T any] interface {
```
jakebailey, line 42: Are these types ever used outside the package such that they should be unexported? All methods are unexported which is usually red flag.

internal/compiler/module/resolver.go
```diff
@@ -0,0 +1,1313 @@
...
+	path                     string
+	extension                string
+	packageId                PackageId
+	originalPath             string
+	resolvedUsingTsExtension bool
+}
+
+type resolutionKindSpecificLoader = func(extensions extensions, candidate string, onlyRecordFailures bool) *resolved
+
+// !!! this seems unnecessary
```
jakebailey, line 30: Probably 😄 

Wonder if you can get away with an `ok bool` return piped down.

internal/compiler/module/resolver.go
```diff
@@ -0,0 +1,1313 @@
...
+		features = NodeResolutionFeaturesNode16Default
+	case core.ModuleResolutionKindNodeNext:
+		features = NodeResolutionFeaturesNodeNextDefault
+	case core.ModuleResolutionKindBundler:
+		features = NodeResolutionFeaturesBundlerDefault
+	}
+	if options.ResolvePackageJsonExports == core.TSTrue {
+		features |= NodeResolutionFeaturesExports
+	} else if options.ResolvePackageJsonExports == core.TSFalse {
+		features &= ^NodeResolutionFeaturesExports
```
jakebailey: For all of the cases like `a &= ^b`, you can write `a &^= b` (a bit unset operator)

internal/compiler/module/resolver.go
```diff
@@ -0,0 +1,1313 @@
...
+	return currentDirectory
+}
+
+type parsedPatterns struct {
+	matchableStringSet collections.OrderedSet[string]
+	patterns           []core.Pattern
+}
+
+func tryParsePatterns(paths map[string][]string) parsedPatterns {
+	// TODO?: TS has a weakmap cache
```
jakebailey: We won't be able to use a weakmap cache in Go; in Go slices/maps aren't comparable so it's unlike we'd be able to key off of anything. There are other ways we could cache things, though, e.g. to hash the input and build a map on that, but even that wouldn't be weak so would instead turn into a more traditional TTL-based cache.

internal/compiler/module/resolver.go
```diff
@@ -0,0 +1,1313 @@
...
+	return append(to, value...)
+}
+
+func initializeResolutionFieldForReadonlyCache[T any](fromCache []T, value []T) []T {
+	if len(fromCache) == 0 {
+		return value
+	}
+	var result []T
+	copy(result, fromCache)
+	return append(result, value...)
```
jakebailey: Probably this could just be `slices.Concat` if you wanted to avoid writing the whole func here.

jakebailey:
LGTM though some minor comments. Things will have to be changed with the new FS abstraction of course.

I also expect our "hosts" to be broken down into concrete elements, given it's going to be challenging to combine them arbitrarily without making some "god objects".

internal/compiler/diagnostics/diagnostics.go
andrewbranch: I didn’t make it a method purely because I thought it would be hard to read. 

```go
diagnostics.Cannot_find_name_0_Do_you_need_to_install_type_definitions_for_a_test_runner_Try_npm_i_save_dev_types_Slashjest_or_npm_i_save_dev_types_Slashmocha_and_then_add_jest_or_mocha_to_the_types_field_in_your_tsconfig.Format(
  "jest",
)
// vs
diagnostics.Format(
  diagnostics.Cannot_find_name_0_Do_you_need_to_install_type_definitions_for_a_test_runner_Try_npm_i_save_dev_types_Slashjest_or_npm_i_save_dev_types_Slashmocha_and_then_add_jest_or_mocha_to_the_types_field_in_your_tsconfig,
  "jest",
)
```

Though I guess

```go
diagnostics.Cannot_find_name_0_Do_you_need_to_install_type_definitions_for_a_test_runner_Try_npm_i_save_dev_types_Slashjest_or_npm_i_save_dev_types_Slashmocha_and_then_add_jest_or_mocha_to_the_types_field_in_your_tsconfig
  .Format("jest")
```

is fine, but for some reason it feels nicer to me not to chain off of the thing that’s going to be wider than my editor.

internal/compiler/module/types.go
andrewbranch: This was some of the earliest code I ported, and I was doing more of a direct port without yet thinking too much about where things get used—it was all in the compiler package and at the time I was just using TS casing. It’s likely that this will end up needing to be properly exported eventually so the Project system can provide a ModuleResolutionCache implementation that can be shared between Programs with compatible options, but we’re pretty far away from that right now.

internal/compiler/module/resolver.go
andrewbranch: Definitely (though I think `stop` is better). The thing that’s infuriating is that some callers of functions that return a `SearchResult` ignore the directive to stop looking. I wanted to have a chance to examine whether all these functions actually need to wrap their results or not before refactoring.

internal/compiler/module/resolver.go
andrewbranch: I feel like it would be fine to have this cache on `Resolver` (and subsequently tie `Resolver` to the life of the `Program`). The work is not expensive, but it could be done many times during the construction of a single program.

testdata/baselines/reference/module/resolver/allowJsCrossMonorepoPackage.trace.json
andrewbranch: I have a branch of TS that generates these; they differ from the ones that would be generated by `main` because everything uses `--moduleResolution node10`

internal/compiler/diagnostics/diagnostics.go
jakebailey: That's true, yeah. I'm not sure how to balance things better, our vars here are already absurdly long 😄 

Fine to leave this of course.

andrewbranch: Eh, the elimination of the extra `diagnostics.` is pretty nice. I’ll try this

PR #80, opened 2024-11-15
Add FS implementation

jakebailey:
This adds a filesystem implementation, wrapping `io/fs.FS`.

The filesystem works on "normalized paths", which in our parlance means "absolute paths, normalized to `/` slashes regardless of OS". These paths look like:

- `/home/jake/mycode/wow.ts`
- `c:/Users/Jake/project/hello.ts`

`io/fs.FS` is not a "traditional" abstraction. All paths must be `/` separated. Good news, that's how we already did it. But, each `fs.FS` itself is _already rooted_ and expects to be given paths _relative_ to that root.. For example, I can do `os.DirFS("/")` or `os.DirFS("c:/")`, then pass in `"home/jake/mycode/wow.ts"`. This poses a problem on Windows machines because there's no common root between all drives. But, good news again, we already have `GetEncodedRootLength` which is able to perform the splitting we need.

The FS in this PR just splits paths with `GetEncodedRootLength` (rejecting relative paths), and then opens the relevant FS by root, passing in the rest of the path to perform file operations.

The FS also supports wrapping a single `fs.FS`, which is useful for using `testing/fstest.Map`. Since this is _not_ multi-root, paths like `c:/blah` are treated as `/c:/blah`.

This FS differs from the original TS "System" interface in a few ways:

- `ReadFile` returns `(contents string, ok bool)` instead of `string | undefined`.
- Instead of providing `readDirectory`, which in TS is a recursive function that also performs exclude/include globs, I'm just giving `WalkDir`, which is the same as `fs.WalkDir`, `filepath.WalkDir`, etc. Globbing can be done externally.
- `Realpath` is implemented by hand and is a noop in the virtual side of the FS. I _think_ my implementation is okay, but, the real thing to use would be `fs.Readlink` which is not yet available in Go.

This PR also eliminates all uses of `path` and `path/filepath` outside of the VFS, `tspath`, tests, and code gen.

### DIFF
--- a/internal/vfs/vfs.go
+++ b/internal/vfs/vfs.go
@@ -69,7 +69,8 @@ var _ FS = (*vfs)(nil)
 // For paths like `c:/foo/bar`, fsys will be used as though it's rooted at `/` and the path is `/c:/foo/bar`.
 func FromIOFS(useCaseSensitiveFileNames bool, fsys fs.FS) FS {
 	return &vfs{
-		readSema:                  osReadSema,
+		readSema: osReadSema,
+		// !!! The passed in FS may not actually respect case insensitive file names.
 		useCaseSensitiveFileNames: useCaseSensitiveFileNames,
 		rootFor: func(root string) fs.FS {
 			if root == "/" {


internal/vfs/vfs.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	return dirs
+}
+
+func (a *adapter) WalkDir(root string, walkFn WalkDirFunc) error {
+	fsys, rest := a.rootAndPath(root)
+	if fsys == nil {
+		return nil
+	}
+	return fs.WalkDir(fsys, rest, walkFn)
```
jakebailey: This is probably wrong due to not joining the root path

internal/vfs/vfs.go
```diff
@@ -0,0 +1,258 @@
...
+	"strings"
+	"sync"
+	"unicode"
+	"unicode/utf16"
+
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+// FS is a file system abstraction.
+type FS interface {
```
jakebailey, line 20: Given there's only one concrete implementation of this interface (which then itself wraps a more generic abstraction), I may just eliminate this entirely.

internal/vfs/vfs.go
```diff
@@ -0,0 +1,258 @@
...
+	// UseCaseSensitiveFileNames returns true if the file system is case-sensitive.
+	UseCaseSensitiveFileNames() bool
+
+	// GetCurrentDirectory returns the current directory.
+	GetCurrentDirectory() string
```
EricCornelson: could this return a `Path` instead of string (similar notes for the other methods)?

internal/compiler/module/types.go
```diff
@@ -9,17 +9,12 @@ import (
 	"github.com/microsoft/typescript-go/internal/compiler/packagejson"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/tspath"
+	"github.com/microsoft/typescript-go/internal/vfs"
 )
 
 type ResolutionHost interface {
-	FileExists(fileName string) bool
-	ReadFile(fileName string) string
+	FS() vfs.FS
```
andrewbranch, line 16: Just curious, why add a method instead of embed?

internal/vfs/vfs.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (v *vfs) FileExists(path string) bool {
+	stat := v.stat(path)
+	return stat != nil && !stat.IsDir()
+}
+
+func (v *vfs) ReadFile(path string) (contents string, ok bool) {
+	if v.readSema != nil {
+		v.readSema <- struct{}{}
+		defer func() { <-v.readSema }()
```
andrewbranch, line 197: I didn’t get this far in the Go book yet 😆

internal/vfs/vfs.go
andrewbranch: It’s very important that it doesn’t—the conversion from string to `Path` is lossy. We would never be able to use this to construct any user-facing strings if it returned a `Path`. `Path` should only ever be used when a comparison is involved (e.g. do two file names refer to the same file; do they share a common ancestor directory)

internal/compiler/module/types.go
jakebailey: That would then imply that we have to create a struct which embeds a `vfs.FS` value and then assign the object there in order to make something which implements `ResolutionHost`. Then, if you need to construct another host object that references the FS of another host, you can't just pass the previous FS by value, you'd have to pass in the _entire_ previous interface and do an "I to I conversion". If it's an accessor like this, then all you do is grab one host's FS and then pass it as part of the next one.

But, maybe that's not the right approach, and we actually _do_ want to construct giant hosts which implement loads of methods. We have a lot of hosts, and that may help us determine "optional" functionality by doing speculative type conversions...

jakebailey:
> What are all the `Path` related `!!!`?

They're all places where I find it mega awkward that I have to convert to string and back, or where I am unsure if I even did it right.

Are our path helpers designed to operate on `Path`s? Like, could I instead make:

```go
func ConvertToRelativePath[S ~string](p S, opt ComparePathsOptions) S { ... }
```

So that it worked with both strings and `Path`?

internal/compiler/program.go
```diff
@@ -41,14 +41,14 @@ func NewProgram(options ProgramOptions) *Program {
 	}
 	p.host = options.Host
 	if p.host == nil {
-		p.host = NewCompilerHost(p.options, options.SingleThreaded)
+		p.host = NewCompilerHost(p.options, options.SingleThreaded, vfs.FromOS(""))
```
jakebailey: Here I would have liked to pass in a cwd, but I'm not sure if that's supposed to be an argument to createprogram or not; I need to check the original code.

internal/compiler/program.go
```diff
@@ -57,6 +57,29 @@ func NewProgram(options ProgramOptions) *Program {
...
+			info, err := d.Info()
+			if err != nil {
+				return err //nolint:wrapcheck
+			}
+			fileInfos = append(fileInfos, FileInfo{Name: path, Size: info.Size()})
+		}
+		return nil
+	})
+	if err != nil {
+		fmt.Println(err)
```
jakebailey, line 76: The old code did this but I'd like to not leave a print in here.

internal/compiler/program.go
```diff
@@ -66,16 +89,16 @@ func (p *Program) parseSourceFiles(fileInfos []FileInfo) {
 	for i := range fileInfos {
 		p.host.RunTask(func() {
 			fileName := fileInfos[i].Name
-			text, _ := p.host.ReadFile(fileName)
+			text, _ := p.host.FS().ReadFile(fileName)
 			sourceFile := ParseSourceFile(fileName, text, p.options.GetEmitScriptTarget())
-			path, _ := filepath.Abs(fileName)
+			path := tspath.ToPath(fileName, p.host.FS().GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
```
jakebailey: I'm thinking there should be a func on `vfs` which does this operation, rather than passing this each time.

internal/vfs/vfs.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+// FS is a file system abstraction.
+type FS interface {
+	// UseCaseSensitiveFileNames returns true if the file system is case-sensitive.
+	UseCaseSensitiveFileNames() bool
+
+	// GetCurrentDirectory returns the current directory.
+	GetCurrentDirectory() string
```
jakebailey: I actually think I should drop these methods entirely; they only carry state which I think is actually better passed through other hosts? The values are not used internally at all. Thoughts?

andrewbranch:
> Are our path helpers designed to operate on `Path`s? Like, could I instead make:
>
> ```go
> func ConvertToRelativePath[S ~string](p S, opt ComparePathsOptions) S { ... }
> ```
>
> So that it worked with both strings and `Path`?

It depends. Unfortunately, you just have to reason about whether the outputs maintain the Pathyness of the inputs. I think this example may be safe, but you have to check whether there’s any way that the `CurrentDirectory string` from `ComparePathsOptions` could contribute any path components to the output. If both the current directory and absolute path inputs were already `Path`s, then it would be safe, but also `UseCaseSensitiveFileNames` option would be unnecessary. So it’s not always totally straightforward.

internal/vfs/vfs.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (v *vfs) UseCaseSensitiveFileNames() bool {
+	return v.useCaseSensitiveFileNames
+}
+
+func rootLength(p string) int {
+	l := tspath.GetEncodedRootLength(p)
+	if l <= 0 {
+		panic("expected absolute path, got: " + p)
```
jakebailey: I added this assert; the FS doesn't have a CWD so cannot itself resolve paths. That's a downside of it not having a CWD...

jakebailey:
I believe the PR is now fully ready for a review; I know @andrewbranch already approved it but I pushed one last change that fixed error reporting paths to now use the non-lowercased paths which I think is correct but want eyes on.

internal/vfs/vfs.go
jakebailey: FWIW TS's system has `GetCurrentDirectory`, I am wondering if we should restore it just so paths are relative? But, I kinda want us to actually name a `NormalizedPath` type or something... It's probably ok right now.

internal/vfs/vfs.go
jakebailey: This is semaphore implemented in terms of a channel; it allows a fixed number of concurrent work. Normally this is not a problem in Go, but performance benchmarking shows that not storming the filesystem with reads improves performance and that's also what gopls does.

jakebailey:
Realized I did not add unit tests, will do that quick.

internal/compiler/module/types.go
DanielRosenwasser: It's nice to implement these big hosts when you only need a subset of functionality, but I'm not against this change.

jakebailey: Yeah, declaring that you only need an FS with `ReadFile` is definitely something not handled here. It's pretty easy to change this, though.

PR #81, opened 2024-11-17
Use `NodeList` with position information for all node lists in AST

ahejlsberg:
This PR implements changes that more closely aligns us with the existing TS codebase by storing position information in node lists and eliminating the new discrete node kinds for modifier, type argument, and type parameter node lists. Specifically:

* Introduces new `NodeList` struct that combines a `TextRange` and a `[]*Node`.
* Changes all `[]*Node` properties in AST to use `NodeList`
* Changes `TypeArgumentList` and `TypeParameterList` to just be a `NodeList`.
* Changes `ModifierList` to be a `NodeList` with an extra `ModifierFlags` property.
* Introduces pool allocation of slices stored in `NodeList`.
* Introduces pool allocation of `Token` nodes.
* Uses stack allocated slices when parsing lists.

Overall, these changes improve parsing speed by 3-10%. Memory size remains roughly equal, but the number of allocations is reduced by 25-40% (because of pool allocation).

In the process of developing this PR, I did experiment with storing modifier lists, type argument lists, and type parameter lists inline (even when they're empty). This had a significant adverse effect on memory consumption and parse times (on the order of 5% for both), so the PR now uses `*ModifierList` and `*NodeList` for optional lists.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -720,7 +720,7 @@ func IsLocalsContainer(node *Node) bool {
 
 type FunctionLikeBase struct {
 	LocalsContainerBase
-	TypeParameters *NodeList // NodeList[*TypeParameterDeclarationNode]
+	TypeParameters *NodeList // NodeList[*TypeParameterDeclarationNode]. Optional
 	Parameters     *NodeList // NodeList[*ParameterDeclarationNode]
 	ReturnType     *TypeNode // Optional
 }
@@ -1556,8 +1556,8 @@ type ClassLikeBase struct {
 	ExportableBase
 	ModifiersBase
 	name            *IdentifierNode // IdentifierNode
-	TypeParameters  *NodeList       // NodeList[*TypeParameterDeclarationNode]
-	HeritageClauses *NodeList       // NodeList[*HeritageClauseNode]
+	TypeParameters  *NodeList       // NodeList[*TypeParameterDeclarationNode]. Optional
+	HeritageClauses *NodeList       // NodeList[*HeritageClauseNode]. Optional
 	Members         *NodeList       // NodeList[*ClassElement]
 }
 
@@ -1644,8 +1644,8 @@ type InterfaceDeclaration struct {
 	ExportableBase
 	ModifiersBase
 	name            *IdentifierNode
-	TypeParameters  *NodeList
-	HeritageClauses *NodeList // NodeList[*HeritageClauseNode]
+	TypeParameters  *NodeList // NodeList[*TypeParameterDeclarationNode]. Optional
+	HeritageClauses *NodeList // NodeList[*HeritageClauseNode]. Optional
 	Members         *NodeList // NodeList[*TypeElement]
 }
 
@@ -1679,7 +1679,7 @@ type TypeAliasDeclaration struct {
 	ModifiersBase
 	LocalsContainerBase
 	name           *IdentifierNode // IdentifierNode
-	TypeParameters *NodeList       // TypeParameterListNode. Optional
+	TypeParameters *NodeList       // NodeList[*TypeParameterDeclarationNode]. Optional
 	TypeNode       *TypeNode       // TypeNode
 }
 
@@ -2901,7 +2901,7 @@ type CallExpression struct {
 	ExpressionBase
 	Expression       *Expression // Expression
 	QuestionDotToken *TokenNode  // TokenNode
-	TypeArguments    *NodeList   // NodeList[*TypeNode]
+	TypeArguments    *NodeList   // NodeList[*TypeNode]. Optional
 	Arguments        *NodeList   // NodeList[*Expression]
 }
 
@@ -2929,8 +2929,8 @@ func IsCallExpression(node *Node) bool {
 type NewExpression struct {
 	ExpressionBase
 	Expression    *Expression // Expression
-	TypeArguments *NodeList   // NodeList[*TypeNode]. Possibly nil
-	Arguments     *NodeList   // NodeList[*Expression]. Possibly nil
+	TypeArguments *NodeList   // NodeList[*TypeNode]. Optional
+	Arguments     *NodeList   // NodeList[*Expression]. Optional
 }
 
 func (f *NodeFactory) NewNewExpression(expression *Expression, typeArguments *NodeList, arguments *NodeList) *Node {
@@ -3067,7 +3067,7 @@ type TaggedTemplateExpression struct {
 	ExpressionBase
 	Tag              *Expression      // Expression
 	QuestionDotToken *TokenNode       // TokenNode. For error reporting purposes only
-	TypeArguments    *NodeList        // NodeList[*TypeNode]
+	TypeArguments    *NodeList        // NodeList[*TypeNode]. Optional
 	Template         *TemplateLiteral // TemplateLiteral
 }
 
@@ -3488,7 +3488,7 @@ func IsIndexedAccessTypeNode(node *Node) bool {
 type TypeReferenceNode struct {
 	TypeNodeBase
 	TypeName      *EntityName // EntityName
-	TypeArguments *NodeList   // NodeList[*TypeNode]
+	TypeArguments *NodeList   // NodeList[*TypeNode]. Optional
 }
 
 func (f *NodeFactory) NewTypeReferenceNode(typeName *EntityName, typeArguments *NodeList) *Node {
@@ -3663,7 +3663,7 @@ func (node *ImportAttributes) ForEachChild(v Visitor) bool {
 type TypeQueryNode struct {
 	TypeNodeBase
 	ExprName      *EntityName // EntityName
-	TypeArguments *NodeList   // NodeList[*TypeNode]
+	TypeArguments *NodeList   // NodeList[*TypeNode]. Optional
 }
 
 func (f *NodeFactory) NewTypeQueryNode(exprName *EntityName, typeArguments *NodeList) *Node {
@@ -4048,7 +4048,7 @@ func IsJsxNamespacedName(node *Node) bool {
 type JsxOpeningElement struct {
 	ExpressionBase
 	TagName       *JsxTagNameExpression // JsxTagNameExpression (Identifier | KeywordExpression | JsxTagNamePropertyAccess | JsxNamespacedName)
-	TypeArguments *NodeList             // NodeList[*TypeNode]
+	TypeArguments *NodeList             // NodeList[*TypeNode]. Optional
 	Attributes    *JsxAttributesNode    // JsxAttributesNode
 }
 
@@ -4073,7 +4073,7 @@ func IsJsxOpeningElement(node *Node) bool {
 type JsxSelfClosingElement struct {
 	ExpressionBase
 	TagName       *JsxTagNameExpression // JsxTagNameExpression (IdentifierReference | KeywordExpression | JsxTagNamePropertyAccess | JsxNamespacedName)
-	TypeArguments *NodeList             // NodeList[*TypeNode]
+	TypeArguments *NodeList             // NodeList[*TypeNode]. Optional
 	Attributes    *JsxAttributesNode    // JsxAttributesNode
 }
 


internal/ast/ast.go
```diff
@@ -677,12 +704,11 @@ func IsLocalsContainer(node *Node) bool {
 
 type FunctionLikeBase struct {
 	LocalsContainerBase
-	typeParameters *TypeParameterListNode // Optional
-	Parameters     []*ParameterDeclarationNode
+	TypeParameters *NodeList // NodeList[*TypeParameterDeclarationNode]
+	Parameters     NodeList  // NodeList[*ParameterDeclarationNode]
```
jakebailey: Rather than using mixed pointer-ness, does it work to just make both pointers and document that one is always present? I'm thinking it may get a little awkward later to have to take pointers to this field to pass it to a function which wants a NodeList (and it would make these objects smaller).

internal/ast/utilities.go
```diff
@@ -475,3 +475,59 @@ func FindAncestorOrQuit(node *Node, callback func(*Node) FindAncestorResult) *No
 	}
 	return nil
 }
+
+func ModifierToFlag(token Kind) ModifierFlags {
```
jakebailey, line 479: Should this just be a method on `Kind`?

internal/ast/utilities.go
```diff
@@ -475,3 +475,59 @@ func FindAncestorOrQuit(node *Node, callback func(*Node) FindAncestorResult) *No
...
+
+func ModifiersToFlags(modifiers []*Node) ModifierFlags {
+	var flags ModifierFlags
+	for _, modifier := range modifiers {
+		flags |= ModifierToFlag(modifier.Kind)
+	}
+	return flags
+}
+
+func GetModifierFlags(node *Node) ModifierFlags {
```
jakebailey: Should this just be a method on `*Node`?

internal/compiler/binder.go
```diff
@@ -1299,9 +1299,9 @@ func (b *Binder) getStrictModeIdentifierMessage(node *ast.Node) *diagnostics.Mes
 	return diagnostics.Identifier_expected_0_is_a_reserved_word_in_strict_mode
 }
 
-func (b *Binder) updateStrictModeStatementList(statements []*ast.Node) {
+func (b *Binder) updateStrictModeStatementList(statements ast.NodeList) {
```
jakebailey: Feels moderately strange to pass this by value, though it is read-only I guess here...

internal/compiler/parser.go
```diff
@@ -247,22 +248,33 @@ func (p *Parser) parseList(kind ParsingContext, parseElement func(p *Parser) *as
...
 }
 
 // Return a non-nil (but possibly empty) slice if parsing was successful, or nil if parseElement returned nil
-func (p *Parser) parseDelimitedList(kind ParsingContext, parseElement func(p *Parser) *ast.Node) []*ast.Node {
+func (p *Parser) parseDelimitedList(kind ParsingContext, parseElement func(p *Parser) *ast.Node) ast.NodeList {
+	pos := p.nodePos()
 	saveParsingContexts := p.parsingContexts
 	p.parsingContexts |= 1 << kind
-	list := []*ast.Node{}
+	list := make([]*ast.Node, 0, 16)
```
jakebailey, line 261: Ah, there's that big allocation I was trying to fix; confusingly this didn't seem to help when I tried it which made no sense...

internal/compiler/parser.go
```diff
@@ -247,22 +248,33 @@ func (p *Parser) parseList(kind ParsingContext, parseElement func(p *Parser) *as
...
 	for {
 		if p.isListElement(kind, false /*inErrorRecovery*/) {
 			startPos := p.nodePos()
 			element := parseElement(p)
 			if element == nil {
 				p.parsingContexts = saveParsingContexts
-				// Return nil list to indicate parseElement failed
-				return nil
+				// Return list with (-1, -1) position to indicate parseElement failed
+				return ast.NewNodeList(core.NewTextRange(-1, -1), nil)
```
jakebailey: Given `TextRange` is a value type and `-1 -1` is commonly used, it'd be good to extract this to a var in core one day.

internal/ast/ast.go
ahejlsberg: I agree it would be more consistent, but at the cost of an extra pointer and an extra indirection. But perhaps that's fine.

internal/ast/utilities.go
ahejlsberg: Hmm, possibly. Though not really a fan of methods on enums from a discoverability point of view.

internal/ast/utilities.go
ahejlsberg: Yeah, that seems reasonable.

internal/compiler/binder.go
ahejlsberg: No stranger than passing the `[]*ast.Node` by value. `NodeList` just adds a `TextRange`. But again, for consistency it may make more sense to always dynamically allocate node lists and pass `*NodeList`.

internal/compiler/parser.go
ahejlsberg: It only works if you make sure the optimizer can "see" that the list doesn't escape the stack. That's the reason `Pool.NewSlice` takes a size and returns an empty slice. At first I had it do the copying, but the optimizer couldn't see though it and things just got worse.

internal/compiler/parser.go
ahejlsberg: Yeah, agreed.

ahejlsberg:
With latest commit we now pool allocate `NodeList` instances and always store `*NodeList` in AST nodes. Performance is a wash, but as expected it consumes a bit more memory (about 1-2%). I do like it for the consistency though.

internal/ast/utilities.go
jakebailey: What makes them less discoverable than other methods?

(In pkg.go.dev the constants are separated from the type so that wouldn't hurt things)

internal/ast/ast.go
```diff
@@ -37,6 +53,38 @@ func (f *NodeFactory) NewNode(kind Kind, data NodeData) *Node {
...
+	list.Loc = loc
+	list.Nodes = nodes
+	return list
+}
+
+func (list *NodeList) Pos() int { return list.Loc.Pos() }
+func (list *NodeList) End() int { return list.Loc.End() }
+
+func (list *NodeList) HasTrailingComma() bool {
+	return len(list.Nodes) > 0 && list.Nodes[len(list.Nodes)-1].End() < list.End()
```
rbuckton, line 74: Will this make it difficult for the language service to synthesize a trailing comma when performing edits? A synthetic `NodeList` would need to have its `End` set greater than the last element, regardless as to whether the last element is existing (with `End >= 0`) or synthetic (with `End == -1`). Changing the `End` to properly simulate a trailing comma could result in the loss of leading or trailing comments for the list

jakebailey:
For reference, main vs this PR:

```
goos: linux
goarch: amd64
                                                      │   old.txt    │               new.txt                │
                                                      │    sec/op    │    sec/op     vs base                │
Bind/empty.ts-20                                        620.9n ± 15%   642.9n ±  6%        ~ (p=0.436 n=10)
Bind/checker.ts-20                                      20.97m ±  7%   21.35m ± 10%        ~ (p=0.353 n=10)
Bind/dom.generated.d.ts-20                              8.581m ±  9%   7.995m ±  1%   -6.84% (p=0.001 n=10)
Bind/Herebyfile.mjs-20                                  336.0µ ±  3%   330.5µ ±  6%        ~ (p=0.796 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20    174.2µ ±  4%   166.3µ ± 11%        ~ (p=0.063 n=10)
Parse/empty.ts-20                                       381.6n ±  2%   639.4n ±  1%  +67.54% (p=0.000 n=10)
Parse/checker.ts-20                                     45.22m ±  1%   42.25m ±  2%   -6.56% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             14.98m ±  2%   14.01m ±  1%   -6.52% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 642.3µ ±  1%   589.6µ ±  4%   -8.21% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   249.6µ ±  1%   225.4µ ±  1%   -9.68% (p=0.000 n=10)
geomean                                                 439.8µ         444.8µ         +1.13%

                                                      │   old.txt    │                 new.txt                 │
                                                      │     B/op     │     B/op      vs base                   │
Bind/empty.ts-20                                         1008.0 ± 0%    1008.0 ± 0%         ~ (p=1.000 n=10) ¹
Bind/checker.ts-20                                      8.410Mi ± 0%   8.410Mi ± 0%         ~ (p=0.896 n=10)
Bind/dom.generated.d.ts-20                              5.386Mi ± 0%   5.385Mi ± 0%    -0.01% (p=0.000 n=10)
Bind/Herebyfile.mjs-20                                  174.5Ki ± 0%   174.5Ki ± 0%         ~ (p=0.084 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20    117.3Ki ± 0%   117.3Ki ± 0%         ~ (p=0.897 n=10)
Parse/empty.ts-20                                         752.0 ± 0%    2128.0 ± 0%  +182.98% (p=0.000 n=10)
Parse/checker.ts-20                                     23.59Mi ± 0%   24.30Mi ± 0%    +3.00% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             8.537Mi ± 0%   8.554Mi ± 0%    +0.20% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 337.4Ki ± 0%   352.6Ki ± 0%    +4.52% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   153.2Ki ± 0%   154.9Ki ± 0%    +1.08% (p=0.000 n=10)
geomean                                                 308.0Ki        344.7Ki        +11.92%
¹ all samples are equal

                                                      │  allocs/op   │  allocs/op   vs base                  │
Bind/empty.ts-20                                          3.000 ± 0%    3.000 ± 0%        ~ (p=1.000 n=10) ¹
Bind/checker.ts-20                                       13.69k ± 0%   13.68k ± 0%   -0.07% (p=0.000 n=10)
Bind/dom.generated.d.ts-20                               15.32k ± 0%   15.30k ± 0%   -0.13% (p=0.000 n=10)
Bind/Herebyfile.mjs-20                                    310.0 ± 0%    310.0 ± 0%        ~ (p=1.000 n=10) ¹
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20      258.0 ± 0%    258.0 ± 0%        ~ (p=1.000 n=10) ¹
Parse/empty.ts-20                                         5.000 ± 0%    6.000 ± 0%  +20.00% (p=0.000 n=10)
Parse/checker.ts-20                                      241.0k ± 0%   157.9k ± 0%  -34.51% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             100.04k ± 0%   72.84k ± 0%  -27.19% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                  3.786k ± 0%   2.606k ± 0%  -31.17% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20    1442.0 ± 0%    895.0 ± 0%  -37.93% (p=0.000 n=10)
geomean                                                  1.419k        1.232k       -13.16%
¹ all samples are equal
```

The empty case is probably larger due to the over-allocation for the top-level statement list. Not important in practice, of course.

I think this PR looks good, but I'll leave it to Ron to double check that this is the layout that he's needing.

internal/ast/ast.go
jakebailey: Yeah, my impression was that we were going to be setting bits on a field on the NodeList...

ahejlsberg: Sure, we can store a discrete flag but it'll cost an extra 8 bytes per `NodeList`. It all adds up.

rbuckton: We may not need to do that for the time being, but it's likely we will need to do that if/when we port the `TextChanges` API for any language service operations that need to be moved to native.

ahejlsberg:
@jakebailey @rbuckton Good to merge?

PR #82, opened 2024-11-20
More type checking

ahejlsberg:
This PR ports

* Template type resolution, instantiation and relations,
* String mapping type resolution, instantiation and relations,
* Deferred type references, and
* Property access chains.

### DIFF
--- a/internal/compiler/relater.go
+++ b/internal/compiler/relater.go
@@ -1949,7 +1949,7 @@ func (c *Checker) inferFromLiteralPartsToTemplateLiteral(sourceTexts []string, s
 		if len(delim) > 0 {
 			s := seg
 			p := pos
-			for true {
+			for {
 				d := strings.Index(getSourceText(s)[p:], delim)
 				if d >= 0 {
 					p += d


internal/compiler/relater.go
```diff
@@ -1844,6 +1844,194 @@ func (c *Checker) getEffectiveConstraintOfIntersection(types []*Type, targetIsUn
...
+	sourceEnd := source.texts[len(source.texts)-1]
+	targetEnd := target.texts[len(target.texts)-1]
+	startLen := min(len(sourceStart), len(targetStart))
+	endLen := min(len(sourceEnd), len(targetEnd))
+	return sourceStart[:startLen] != targetStart[:startLen] || sourceEnd[len(sourceEnd)-endLen:] != targetEnd[len(targetEnd)-endLen:]
+}
+
+func (c *Checker) isTypeMatchedByTemplateLiteralType(source *Type, target *TemplateLiteralType) bool {
+	inferences := c.inferTypesFromTemplateLiteralType(source, target)
+	if inferences != nil {
```
jakebailey, line 1860: Although this was a direct port more or less, I'm a little wary about `nil` comparisons with slices, given how nil and empty are more or less interchangable; would this function work out if it were instead a `types []*Type, ok bool` return?

internal/compiler/checker.go
```diff
@@ -9081,7 +9241,7 @@ func (c *Checker) getTypeFromArrayOrTupleTypeNode(node *ast.Node) *Type {
 		if target == c.emptyGenericType {
 			links.resolvedType = c.emptyObjectType
 		} else if !(node.Kind == ast.KindTupleType && core.Some(node.AsTupleTypeNode().Elements.Nodes, c.isVariadicTupleElement)) && c.isDeferredTypeReferenceNode(node, false) {
-			if node.Kind == ast.KindTupleType && len(node.AsTupleTypeNode().Elements.Nodes) != 0 {
+			if node.Kind == ast.KindTupleType && len(node.AsTupleTypeNode().Elements.Nodes) == 0 {
```
jakebailey, line 9244: Is this one of the bugs I caused? 😅

internal/compiler/relater.go
ahejlsberg: We're fine here because template literal types always have at least one element in the `.types` slice. In general I agree that comparing slices to `nil` is a bit suspect (even if the difference between `nil` and an empty slice is indeed a documented aspect of the language).

PR #83, opened 2024-11-20
Port grammar checks

DanielRosenwasser:
This brings along most grammar checks into the new compiler in a new file called `grammarchecks.go`.

Most of this file just contains `Checker` methods, so it really does continue to be an extension of `checker.go`. I think that long-term, it would be nice if we could make these functions entirely independent of the type-checker and run them as part of the binder. This would interleave less redundant work with each visitor function in the type-checker. But that's going beyond the spirit of a port, and placing these methods in their own file can make that a little bit easier.

Not all functions are actually called at the moment. Specifically, the following don't get invoked anywhere from the checker because they are invoked from `check*` functions which are not immediately invoked by `checkSourceElement`:

* `checkGrammarRegularExpressionLiteral`
* `checkGrammarDecorator`
* `checkGrammarClassLikeDeclaration`
* `checkGrammarTaggedTemplateChain`
* `checkGrammarClassDeclarationHeritageClauses`
* `checkGrammarJsxElement`
* `checkGrammarJsxName`
* `checkGrammarJsxExpression`
* `checkGrammarMetaProperty`
* `checkGrammarImportCallExpression`

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -1371,14 +1371,14 @@ func (c *Checker) checkParameter(node *ast.Node) {
 }
 
 func (c *Checker) checkPropertyDeclaration(node *ast.Node) {
-	// // Grammar checking
+	// Grammar checking
 	if !c.checkGrammarModifiers(node) && !c.checkGrammarProperty(node) {
 		c.checkGrammarComputedPropertyName(node.Name())
 	}
 	c.checkVariableLikeDeclaration(node)
 
-	// // !!!
-	// // c.setNodeLinksForPrivateIdentifierScope(node)
+	// !!!
+	// c.setNodeLinksForPrivateIdentifierScope(node)
 
 	// property signatures already report "initializer not allowed in ambient context" elsewhere
 	if hasSyntacticModifier(node, ast.ModifierFlagsAbstract) && ast.IsPropertyDeclaration(node) {
--- a/internal/compiler/grammarchecks.go
+++ b/internal/compiler/grammarchecks.go
@@ -1184,7 +1184,7 @@ func (c *Checker) checkGrammarJsxElement(node *ast.Node, jsxCommon struct {
 }) bool {
 	c.checkGrammarJsxName(jsxCommon.tagName)
 	c.checkGrammarTypeArguments(node, jsxCommon.typeArguments)
-	seen := make(map[string]bool)
+	var seen core.Set[string]
 
 	for _, attrNode := range jsxCommon.attributes.AsJsxAttributes().Properties.Nodes {
 		if attrNode.Kind == ast.KindJsxSpreadAttribute {
@@ -1195,8 +1195,8 @@ func (c *Checker) checkGrammarJsxElement(node *ast.Node, jsxCommon struct {
 		name := attr.Name()
 		initializer := attr.Initializer
 		textOfName := name.Text()
-		if !seen[textOfName] {
-			seen[textOfName] = true
+		if !seen.Has(textOfName) {
+			seen.Add(textOfName)
 		} else {
 			return c.grammarErrorOnNode(name, diagnostics.JSX_elements_cannot_have_multiple_attributes_with_the_same_name)
 		}


DanielRosenwasser:
Most of what's missing now is related to JSDoc - stuff like grabbing the *effective* return type annotation and the like. I'll probably need to sync with @sandersn on where to stub out some of these helpers just to avoid stepping on each other's toes.

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2193 @@
...
+				if node.Kind == ast.KindTypeParameter {
+					return c.grammarErrorOnNode(modifier, diagnostics.X_0_modifier_cannot_appear_on_a_type_parameter, TokenToString(modifier.Kind))
+				}
+			}
+			switch modifier.Kind {
+			case ast.KindConstKeyword:
+				if node.Kind != ast.KindEnumDeclaration && node.Kind != ast.KindTypeParameter {
+					return c.grammarErrorOnNode(node, diagnostics.A_class_member_cannot_have_the_0_keyword, TokenToString(ast.KindConstKeyword))
+				}
+				parent := (isJSDocTemplateTag(node.Parent) && getEffectiveJSDocHost(node.Parent)) || node.Parent
```
sandersn: I haven't written this yet, but my intent is to create synthetic nodes in the binder that convert template tags to type parameters in the tree (but keep a pointer to the original template tag) so that this code isn't necessary. Something like SyntheticTypeParameter.

There are a couple of problems already apparent to me:
1. Synthetic generation + grammar checks in the same pass may or may not work.
2. There's no variance in go, so TypeParameter and SyntheticTypeParameter may not be nearly as interchangeable as they would in TS.

Regardless, I'd like for code like this line to go away and be replaced with `node.Parent`

DanielRosenwasser: One concern I have with something like that is that we'll often write code that assumes something about textual contents of a node (e.g. `<` or `>` existing in a type parameter list), or we'll want to give a better error based on the tag name of JSDoc. So I'm curious if the approach you're taking makes it hard to run into those bad assumptions, and allows us to still probe on concrete syntax.

It's not a hard requirement, it would just be nice to have.

Regarding the subtyping, we'll have to do a lot of `*ast.Node` like we do everywhere now.

internal/compiler/checker.go
```diff
@@ -5542,6 +5557,38 @@ func (c *Checker) getCombinedNodeFlagsCached(node *ast.Node) ast.NodeFlags {
...
+		return c.tryGetNameFromType(c.getTypeOfExpression(node.Expression()))
+	default:
+		return effectiveName, false
+	}
+}
+
+func (c *Checker) tryGetNameFromType(t *Type) (name string, ok bool) {
+	switch {
+	case t.flags&TypeFlagsUniqueESSymbol != 0:
+		return (t.AsUniqueESSymbolType()).name, true
```
jakebailey: This must be converted code from an older ts-to-go; these parens are redundant.

internal/compiler/checker.go
```diff
@@ -119,6 +119,18 @@ type StringMappingKey struct {
 
 type InferenceContext struct{}
 
+type DeclarationMeaning uint32
+
+const (
+	DeclarationMeaningGetAccessor = 1 << iota
```
```suggestion
	DeclarationMeaningGetAccessor DeclarationMeaning = 1 << iota
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+	// 		panic("Expected to rescan RegularExpressionLiteral")
+	// 	}
+	// 	return lastError != nil
+	// }
+	// return false
+}
+
+func (c *Checker) checkGrammarPrivateIdentifierExpression(privId ast.PrivateIdentifier) bool {
+	privIdAsNode := privId.AsNode()
+	if !(getContainingClass(privId.AsNode()) != nil) {
```
jakebailey:
```suggestion
	if getContainingClass(privId.AsNode()) == nil {
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+		return c.grammarErrorOnNode(privId.AsNode(), diagnostics.Private_identifiers_are_not_allowed_outside_class_bodies)
+	}
+
+	if !ast.IsForInStatement(privId.Parent) {
+		if !isExpressionNode(privIdAsNode) {
+			return c.grammarErrorOnNode(privIdAsNode, diagnostics.Private_identifiers_are_only_allowed_in_class_bodies_and_may_only_be_used_as_part_of_a_class_member_declaration_property_access_or_on_the_left_hand_side_of_an_in_expression)
+		}
+
+		isInOperation := ast.IsBinaryExpression(privId.Parent) && privId.Parent.AsBinaryExpression().OperatorToken.Kind == ast.KindInKeyword
+		if !(c.getSymbolForPrivateIdentifierExpression(privIdAsNode) != nil) && !isInOperation {
```
jakebailey:
```suggestion
		if c.getSymbolForPrivateIdentifierExpression(privIdAsNode) == nil && !isInOperation {
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+func (c *Checker) checkGrammarMappedType(node *ast.MappedTypeNode) bool {
+	if len(node.Members.Nodes) > 0 {
+		return c.grammarErrorOnNode(node.Members.Nodes[0], diagnostics.A_mapped_type_may_not_declare_properties_or_methods)
+	}
+	return false
+}
+func (c *Checker) checkGrammarDecorator(decorator *ast.Decorator) bool {
+	sourceFile := ast.GetSourceFileOfNode(decorator.AsNode())
+	if !c.hasParseDiagnostics(sourceFile) {
+		var node *ast.Expression = decorator.Expression
```
jakebailey: Type declaration here seems redundant?

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+	isInAppropriateContext := node.Parent.Kind == ast.KindSourceFile || node.Parent.Kind == ast.KindModuleBlock || node.Parent.Kind == ast.KindModuleDeclaration
+	if !isInAppropriateContext {
+		c.grammarErrorOnFirstToken(node, errorMessage)
+	}
+	return !isInAppropriateContext
+}
+
+func (c *Checker) checkGrammarModifiers(node *ast.Node /*Union[HasModifiers, HasDecorators, HasIllegalModifiers, HasIllegalDecorators]*/) bool {
+	quickResult := c.reportObviousDecoratorErrors(node)
+	if quickResult == core.TSTrue {
```
jakebailey: Not for this PR, but I do wonder if we'd have an easier time reading these comparisons if they were like `quickResult.IsTrue()`...

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+	if ast.IsParameter(node) && parameterIsThisKeyword(node) {
+		return c.grammarErrorOnFirstToken(node, diagnostics.Neither_decorators_nor_modifiers_may_be_applied_to_this_parameters)
+	}
+
+	var blockScopeKind ast.NodeFlags
+	if ast.IsVariableStatement(node) {
+		blockScopeKind = node.AsVariableStatement().DeclarationList.Flags & ast.NodeFlagsBlockScoped
+	} else {
+		blockScopeKind = ast.NodeFlagsNone
+	}
```
jakebailey:
```suggestion
	blockScopeKind := ast.NodeFlagsNone
	if ast.IsVariableStatement(node) {
		blockScopeKind = node.AsVariableStatement().DeclarationList.Flags & ast.NodeFlagsBlockScoped
	}
```

You don't have to do this, but know that the compiler is totally happy to optimize code written like this too.

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+				}
+			} else if c.compilerOptions.LegacyDecorators == core.TSTrue && (node.Kind == ast.KindGetAccessor || node.Kind == ast.KindSetAccessor) {
+				accessors := c.getAllAccessorDeclarationsForDeclaration(node)
+				if hasDecorators(accessors.firstAccessor) && node == accessors.secondAccessor {
+					return c.grammarErrorOnFirstToken(node, diagnostics.Decorators_cannot_be_applied_to_multiple_get_Slashset_accessors_of_the_same_name)
+				}
+			}
+
+			// if we've seen any modifiers aside from `export`, `default`, or another decorator, then this is an invalid position
+			if flags & ^(ast.ModifierFlagsExportDefault|ast.ModifierFlagsDecorator) != 0 {
```
jakebailey:
```suggestion
			if flags &^ (ast.ModifierFlagsExportDefault|ast.ModifierFlagsDecorator) != 0 {
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+					err.AddRelatedInfo(createDiagnosticForNode(firstDecorator, diagnostics.Decorator_used_before_export_here))
+					return true
+				}
+				return false
+			}
+
+			flags |= ast.ModifierFlagsDecorator
+
+			// if we have not yet seen a modifier, then these are leading decorators
+			if !(flags&ast.ModifierFlagsModifier != 0) {
```
jakebailey:
```suggestion
			if flags&ast.ModifierFlagsModifier == 0 {
```

Must be an old ts-to-go?

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+				ast.KindPrivateKeyword:
+				text := visibilityToString(ast.ModifierToFlag(modifier.Kind))
+
+				if flags&ast.ModifierFlagsAccessibilityModifier != 0 {
+					return c.grammarErrorOnNode(modifier, diagnostics.Accessibility_modifier_already_seen)
+				} else if flags&ast.ModifierFlagsOverride != 0 {
+					return c.grammarErrorOnNode(modifier, diagnostics.X_0_modifier_must_precede_1_modifier, text, "override")
+				} else if flags&ast.ModifierFlagsStatic != 0 {
+					return c.grammarErrorOnNode(modifier, diagnostics.X_0_modifier_must_precede_1_modifier, text, "static")
+				} else if flags&ast.ModifierFlagsAccessor != 0 {
```
jakebailey: So many ifs; seems like they could be `switch { case ...: }`?

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+				} else {
+					container = node.Parent.Parent
+				}
+				if container.Kind == ast.KindModuleDeclaration && !isAmbientModule(container) {
+					return c.grammarErrorOnNode(modifier, diagnostics.A_default_export_can_only_be_used_in_an_ECMAScript_style_module)
+				} else if blockScopeKind == ast.NodeFlagsUsing {
+					return c.grammarErrorOnNode(modifier, diagnostics.X_0_modifier_cannot_appear_on_a_using_declaration, "default")
+				} else if blockScopeKind == ast.NodeFlagsAwaitUsing {
+					return c.grammarErrorOnNode(modifier, diagnostics.X_0_modifier_cannot_appear_on_an_await_using_declaration, "default")
+				} else if !(flags&ast.ModifierFlagsExport != 0) {
```
jakebailey:
```suggestion
				} else if flags&ast.ModifierFlagsExport == 0 {
```

BTW I use this to fix these up in ts-to-go:

```console
$ ast-grep -U -p '!($A != $B)' -r '$A == $B' <path>
$ ast-grep -U -p '!($A == $B)' -r '$A != $B' <path>
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+}
+
+func (c *Checker) checkGrammarParameterList(parameters *ast.NodeList) bool {
+	seenOptionalParameter := false
+	parameterCount := len(parameters.Nodes)
+
+	for i := 0; i < parameterCount; i++ {
+		parameter := parameters.Nodes[i].AsParameterDeclaration()
+		if parameter.DotDotDotToken != nil {
+			if i != (parameterCount - 1) {
```
jakebailey:
```suggestion
			if i != parameterCount-1 {
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+				return c.grammarErrorOnNode(parameter.DotDotDotToken, diagnostics.A_rest_parameter_must_be_last_in_a_parameter_list)
+			}
+			if !(parameter.Flags&ast.NodeFlagsAmbient != 0) {
```
jakebailey:
```suggestion
			if parameter.Flags&ast.NodeFlagsAmbient == 0 {
```

internal/compiler/grammar_checks.go
```diff
@@ -0,0 +1,2249 @@
...
+			}
+		}
+	}
+
+	if node.ExclamationToken != nil && (node.Parent.Parent.Kind != ast.KindVariableStatement || node.TypeNode == nil || node.Initializer != nil || nodeFlags&ast.NodeFlagsAmbient != 0) {
+		var message *diagnostics.Message
+		switch {
+		case node.Initializer != nil:
+			message = diagnostics.Declarations_with_initializers_cannot_also_have_definite_assignment_assertions
+		case !(node.TypeNode != nil):
```
jakebailey:
```suggestion
		case node.TypeNode == nil:
```

I'll stop commenting on these now 😄

internal/ast/ast.go
```diff
@@ -332,7 +334,7 @@ func (n *Node) Initializer() *Node {
 	case KindPropertyDeclaration:
 		return n.AsPropertyDeclaration().Initializer
 	case KindPropertySignature:
-		return n.AsPropertyDeclaration().Initializer
+		return n.AsPropertySignatureDeclaration().Initializer
```
jakebailey, line 337: oops

internal/ast/utilities.go
```diff
@@ -531,6 +561,80 @@ func ModifiersToFlags(modifiers []*Node) ModifierFlags {
...
+		KindPropertyAssignment,
+		KindShorthandPropertyAssignment,
+		KindMissingDeclaration,
+		KindNamespaceExportDeclaration:
+		return true
+	}
+	return false
+}
+
+func CanHaveModifiers(node *Node) bool {
```
jakebailey, line 592: Should check Ron's PR to see if we're duplicating anything (I assume so, and also assume that it's inevitable with so much parallel work)

internal/compiler/checker.go
```diff
@@ -1107,6 +1119,9 @@ func (c *Checker) getSymbol(symbols ast.SymbolTable, name string, meaning ast.Sy
 func (c *Checker) checkSourceFile(sourceFile *ast.SourceFile) {
 	links := c.sourceFileLinks.get(sourceFile)
 	if !links.typeChecked {
+		// Grammar checking
```
jakebailey, line 1122: I always found these comments to be redundant in the old code (well yeah, it's a grammar check, it says so right there), but, it's ok

internal/compiler/checker.go
```diff
@@ -1334,51 +1349,112 @@ func (c *Checker) checkDeferredNode(node *ast.Node) {
...
+	// or if its FunctionBody is strict code(11.1.5).
+	c.checkGrammarModifiers(node)
+
 	// !!!
 	node.ForEachChild(c.checkSourceElement)
 }
 
 func (c *Checker) checkPropertyDeclaration(node *ast.Node) {
-	// !!!
+	// // Grammar checking
```
jakebailey: Still some leftover double comments in places it seems

internal/compiler/utilities.go
```diff
@@ -2022,6 +2060,39 @@ func isPartOfPossiblyValidTypeOrAbstractComputedPropertyName(node *ast.Node) boo
...
+		ast.KindMethodDeclaration:
+		// if this method has a body and its parent is a class declaration, this is a valid target.
+		return node.BodyData() != nil && parent != nil && (ast.IsClassDeclaration(parent) || !useLegacyDecorators && ast.IsClassExpression(parent))
+	case ast.KindParameter:
+		// TODO(rbuckton): Parameter decorator support for ES decorators must wait until it is standardized
+		if !useLegacyDecorators {
+			return false
+		}
+		// if the parameter's parent has a body and its grandparent is a class declaration, this is a valid target.
+		return parent != nil && parent.BodyData() != nil && (parent.BodyData()).Body != nil && (parent.Kind == ast.KindConstructor || parent.Kind == ast.KindMethodDeclaration || parent.Kind == ast.KindSetAccessor) && getThisParameter(parent) != node && grandparent != nil && grandparent.Kind == ast.KindClassDeclaration
```
jakebailey, line 2090: holy line length batman

internal/compiler/grammarchecks.go
```diff
@@ -0,0 +1,2224 @@
...
+}
+
+func (c *Checker) checkGrammarJsxElement(node *ast.Node, jsxCommon struct {
+	tagName       *ast.JsxTagNameExpression
+	typeArguments *ast.NodeList
+	attributes    *ast.JsxAttributesNode
+}) bool {
+	c.checkGrammarJsxName(jsxCommon.tagName)
+	c.checkGrammarTypeArguments(node, jsxCommon.typeArguments)
+	seen := make(map[string]bool)
```
jakebailey: I would just use `core.Set[string]` here (especially rather than a map to bool)

PR #84, opened 2024-11-20
Add basic reference/module resolution to program.go

EricCornelson:
Adds parsing reference directives from source files, and wires up basic module resolution in program.go.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -4387,10 +4387,9 @@ type FileReference struct {
 }
 
 type PragmaArgument struct {
+	core.TextRange
 	Name  string
 	Value string
-	Pos   int
-	End   int
 }
 
 type Pragma struct {
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -54,12 +54,12 @@ func NewProgram(options ProgramOptions) *Program {
 	if p.options == nil {
 		p.options = &core.CompilerOptions{}
 	}
-	p.filesByPath = make(map[string]*ast.SourceFile)
+	p.filesByPath = make(map[string]*ast.SourceFile, len(options.RootNames))
 
 	p.resolvedModules = make(map[string]*module.ResolvedModuleWithFailedLookupLocations)
 	//p.maxNodeModuleJsDepth = p.options.MaxNodeModuleJsDepth
 
-	// TODO: tracing?
+	// TODO(ercornel): !!! tracing?
 	// tracing?.push(tracing.Phase.Program, "createProgram", { configFilePath: options.configFilePath, rootDir: options.rootDir }, /*separateBeginAndEnd*/ true);
 	// performance.mark("beforeProgram");
 
@@ -68,13 +68,13 @@ func NewProgram(options ProgramOptions) *Program {
 		p.host = NewCompilerHost(p.options, options.SingleThreaded)
 	}
 
-	// TODO: SKIPPING FOR NOW :: default lib
+	// TODO(ercornel): !!!: SKIPPING FOR NOW :: default lib
 
 	p.currentDirectory = p.host.GetCurrentDirectory()
 	p.resolver = module.NewResolver(p.host, nil, p.options)
 
 	if len(options.ProjectReferences) > 0 {
-		// TODO: project references
+		// TODO(ercornel): !!!: project references
 	}
 
 	for i, rootName := range options.RootNames {
@@ -117,7 +117,7 @@ func (p *Program) findSourceFile(candidate string, reason FileIncludeReason) *as
 		}
 	}
 
-	// TODO: how to make this async
+	// TODO(ercornel): !!!: how to make this async
 	file := p.parseSourceFile(candidate)
 	if file != nil {
 		p.filesByPath[candidate] = file
@@ -130,7 +130,7 @@ func (p *Program) findSourceFile(candidate string, reason FileIncludeReason) *as
 func (p *Program) parseSourceFile(fileName string) *ast.SourceFile {
 	text, _ := p.host.ReadFile(fileName)
 	sourceFile := ParseSourceFile(fileName, text, p.options.GetEmitScriptTarget())
-	path, _ := filepath.Abs(fileName)
+	path := tspath.GetNormalizedAbsolutePath(fileName, p.currentDirectory)
 	sourceFile.SetPath(path)
 	p.files = append(p.files, sourceFile)
 	return sourceFile
@@ -143,9 +143,9 @@ func (p *Program) processReferencedFiles(file *ast.SourceFile) {
 }
 
 func (p *Program) getResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile {
-	directory := filepath.Dir(currentSourceFile.Path())
+	directory := tspath.GetDirectoryPath(currentSourceFile.Path())
 	if tspath.IsExternalModuleNameRelative(moduleReference) {
-		return p.findSourceFile(filepath.Join(directory, moduleReference), FileIncludeReason{Import, 0})
+		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference), FileIncludeReason{Import, 0})
 	}
 	return p.findNodeModule(moduleReference)
 }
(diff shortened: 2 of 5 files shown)


internal/ast/ast.go
```diff
@@ -4359,3 +4364,62 @@ func (node *SourceFile) ForEachChild(v Visitor) bool {
...
+	FileName       string
+	ResolutionMode core.ResolutionMode
+	Preserve       bool
+}
+
+type PragmaArgument struct {
+	Name  string
+	Value string
+	Pos   int
+	End   int
```
andrewbranch: Is there a reason not to embed `core.TextRange` here?

internal/compiler/module/resolver.go
```diff
@@ -164,7 +164,7 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
 
 		switch moduleResolution {
 		case core.ModuleResolutionKindNode16, core.ModuleResolutionKindNodeNext, core.ModuleResolutionKindBundler:
-			state := newResolutionState(moduleName, containingFile, resolutionMode, compilerOptions, r)
+			state := newResolutionState(moduleName, containingDirectory, resolutionMode, compilerOptions, r)
```
andrewbranch, line 167: Oops, thanks!

internal/compiler/program.go
```diff
@@ -39,48 +54,40 @@ func NewProgram(options ProgramOptions) *Program {
 	if p.options == nil {
 		p.options = &core.CompilerOptions{}
 	}
+	p.filesByPath = make(map[string]*ast.SourceFile)
```
andrewbranch: Maybe use the number of root files as a capacity hint? (Jake would know better than I if this is a good idea)

internal/compiler/program.go
```diff
@@ -39,48 +54,40 @@ func NewProgram(options ProgramOptions) *Program {
...
+
+	p.resolvedModules = make(map[string]*module.ResolvedModuleWithFailedLookupLocations)
+	//p.maxNodeModuleJsDepth = p.options.MaxNodeModuleJsDepth
+
+	// TODO: tracing?
```
andrewbranch: Unfortunately, there are many TODO comments in the Strada codebase, most of which are getting copied over. Anders has started using `// !!!` as an indicator for porting to-do.

internal/compiler/program.go
```diff
@@ -92,23 +99,120 @@ func (p *Program) bindSourceFiles() {
...
+		p.processReferencedFiles(file)
+		p.processImportedModules(file)
+	}
+	return file
+}
+
+func (p *Program) parseSourceFile(fileName string) *ast.SourceFile {
+	text, _ := p.host.ReadFile(fileName)
+	sourceFile := ParseSourceFile(fileName, text, p.options.GetEmitScriptTarget())
+	path, _ := filepath.Abs(fileName)
```
andrewbranch:
```suggestion
	path := tspath.GetNormalizedAbsolutePath(fileName)
```

internal/compiler/program.go
```diff
@@ -92,23 +99,120 @@ func (p *Program) bindSourceFiles() {
...
+}
+
+func (p *Program) processReferencedFiles(file *ast.SourceFile) {
+	for _, ref := range file.ReferencedFiles {
+		p.processSourceFile(ref.FileName, FileIncludeReason{ReferenceFile, 0})
+	}
+}
+
+func (p *Program) getResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile {
+	directory := filepath.Dir(currentSourceFile.Path())
```
andrewbranch:
```suggestion
	directory := tspath.GetDirectoryPath(currentSourceFile.Path())
```

internal/compiler/program.go
```diff
@@ -92,23 +99,120 @@ func (p *Program) bindSourceFiles() {
...
+	if tspath.IsExternalModuleNameRelative(moduleReference) {
+		return p.findSourceFile(filepath.Join(directory, moduleReference), FileIncludeReason{Import, 0})
```
andrewbranch:
```suggestion
		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference), FileIncludeReason{Import, 0})
```

internal/compiler/types.go
```diff
@@ -1103,3 +1103,46 @@ var LanguageFeatureMinimumTarget = LanguageFeatureMinimumTargetMap{
...
+	Jsx                   = ".jsx"
+	Json                  = ".json"
+	TsBuildInfo           = ".tsbuildinfo"
+	Mjs                   = ".mjs"
+	Mts                   = ".mts"
+	Dmts                  = ".d.mts"
+	Cjs                   = ".cjs"
+	Cts                   = ".cts"
+	Dcts                  = ".d.cts"
+)
```
andrewbranch: These exist at `tspath.ExtensionTs` etc.

internal/compiler/program.go
```diff
@@ -92,23 +99,120 @@ func (p *Program) bindSourceFiles() {
...
+			// - noResolve is falsy
+			// - module name comes from the list of imports
+			// - it's not a top level JavaScript module that exceeded the search max
+
+			//const elideImport = isJsFileFromNodeModules && currentNodeModulesDepth > maxNodeModuleJsDepth;
+
+			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
+			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
+
+			shouldAddFile := resolvedFileName != ""
```
andrewbranch: This is fine, but equivalent to `resolution.IsResolved()` which reads a little nicer

internal/compiler/program.go
```diff
@@ -92,23 +99,120 @@ func (p *Program) bindSourceFiles() {
...
 
-func (p *Program) getResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile {
-	directory := filepath.Dir(currentSourceFile.Path())
-	if tspath.IsExternalModuleNameRelative(moduleReference) {
-		return p.findSourceFile(filepath.Join(directory, moduleReference))
-	}
-	return p.findNodeModule(moduleReference)
+func (p *Program) processRootFile(fileName string, reason FileIncludeReason /*, isDefaultLib bool, ignoreNoDefaultLib bool, */) {
+	normalizedPath := tspath.NormalizePath(fileName)
+	p.processSourceFile(normalizedPath, reason)
```
andrewbranch, line 104: This probably needs design discussion consensus, but I would really love to go ahead and make an absolute path from here (which would be a divergence from today’s tsc invoked without a tsconfig)

internal/ast/ast.go
EricCornelson: nope! just missed it

internal/compiler/program.go
EricCornelson: Sounds reasonable to me

internal/compiler/program.go
EricCornelson: added !!! and my name to my todos

internal/compiler/types.go
EricCornelson: removed

internal/compiler/program.go
EricCornelson: I was talking with Jake a bit about the path stuff, and how it would be nice to have an abstract representation of paths that internally store them in some canonical normalized way (possibly with a reference back to the original string if needed), and then you just render them out however you want them when needed. There's lots of ad hoc path manipulation going on in various places, and it's really hard to tell what form a path is in at any given time

jakebailey: I think the main thing I was confused about was "absolute" versus "absolute and lowercased"; I was assuming we could get away with just the latter but was wrong.

So, at that point, we're back to three path types where one is "idk", one is "absolute", and one is "absolute and lowercased for use as map keys", which is what we had in old TS too (more or less). Not super clear if we should name all of these types or just the final one?

But I do think that absoluting every and all paths ever provided is the best idea; that invariant is very helpful and I don't see how it could backfire.

andrewbranch: I suspect that doing it here can eliminate _many_ subsequent `getNormalizedAbsolutePath` calls elsewhere, since today we can never rely on `sourceFile.fileName` being absolute (which is bananas)

PR #87, opened 2024-11-21
add optionDeclarations

iisaduan:
Adds: 
- package `tsoptions` which will contain the various option declarations, command line parsing, and tsconfig parsing
  - `commandLineOption` struct
  - `optionDeclarations` (in `declscompiler.go`)
  - `optionsForBuild` (in `declsbuild.go`)
- some parts of `compilerOptions`
- single line functions to create ordered maps and sets from lists or items

Left to do:
- I'm going to look into script-generating the rest of the command line option definitions (like the watch and build options).
- separate options-- in the ts compiler, the lists of `commandLineOptions` are often reused and recombined to make up the complete lists `optionsDeclarations`, `optionsForBuild`, etc... 
- `interface TsConfigOnlyOption` not yet implemented

### DIFF
--- a/internal/options/commandlineoption.go
+++ b/internal/options/commandlineoption.go
@@ -15,7 +15,7 @@ const (
 	CommandLineOptionTypeObject        CommandLineOptionKind = "object"
 	CommandLineOptionTypeList          CommandLineOptionKind = "list"
 	CommandLineOptionTypeListOrElement CommandLineOptionKind = "listOrElement"
-	CommandLineOptionTypeEnum          CommandLineOptionKind = "enum" //map
+	CommandLineOptionTypeEnum          CommandLineOptionKind = "enum" // map
 )
 
 type CommandLineOption struct {
@@ -59,7 +59,7 @@ type CommandLineOption struct {
 	// todo: revisit to see if this can be reduced to boolean
 	transpileOptionValue core.Tristate
 
-	// used in listtype
+	// used for CommandLineOptionTypeList
 	listPreserveFalsyValues bool
 }
 
--- a/internal/options/declscompiler.go
+++ b/internal/options/declscompiler.go
@@ -267,7 +267,7 @@ var optionsDeclarations = []CommandLineOption{
 	{
 		Name:                     "target",
 		shortName:                "t",
-		Kind:                     CommandLineOptionTypeEnum, //targetOptionMap
+		Kind:                     CommandLineOptionTypeEnum, // targetOptionMap
 		affectsSourceFile:        true,
 		affectsModuleResolution:  true,
 		affectsEmit:              true,
@@ -282,7 +282,7 @@ var optionsDeclarations = []CommandLineOption{
 	{
 		Name:                     "module",
 		shortName:                "m",
-		Kind:                     CommandLineOptionTypeEnum, //moduleOptionMap
+		Kind:                     CommandLineOptionTypeEnum, // moduleOptionMap
 		affectsSourceFile:        true,
 		affectsModuleResolution:  true,
 		affectsEmit:              true,
@@ -297,7 +297,7 @@ var optionsDeclarations = []CommandLineOption{
 		Kind: CommandLineOptionTypeList,
 		// elements: &CommandLineOption{
 		// 	name:                    "lib",
-		// 	kind:                   CommandLineOptionTypeEnum, //libMap,
+		// 	kind:                   CommandLineOptionTypeEnum, // libMap,
 		// 	defaultValueDescription: core.TSUnknown,
 		// },
 		affectsProgramStructure:  true,
@@ -329,7 +329,7 @@ var optionsDeclarations = []CommandLineOption{
 	},
 	{
 		Name:                    "jsx",
-		Kind:                    CommandLineOptionTypeEnum, //jsxOptionMap,
+		Kind:                    CommandLineOptionTypeEnum, // jsxOptionMap,
 		affectsSourceFile:       true,
 		affectsEmit:             true,
 		affectsBuildInfo:        true,
@@ -944,7 +944,7 @@ var optionsDeclarations = []CommandLineOption{
 	},
 	{
 		Name:                    "newLine",
-		Kind:                    CommandLineOptionTypeEnum, //newLineOptionMap,
+		Kind:                    CommandLineOptionTypeEnum, // newLineOptionMap,
 		affectsEmit:             true,
 		affectsBuildInfo:        true,
 		category:                diagnostics.Emit,


internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+	CommandLineOptionTypeNumber        CommandLineOptionKind = "number"
+	CommandLineOptionTypeBoolean       CommandLineOptionKind = "boolean"
+	CommandLineOptionTypeObject        CommandLineOptionKind = "object"
+	CommandLineOptionTypeList          CommandLineOptionKind = "list"
+	CommandLineOptionTypeListOrElement CommandLineOptionKind = "listOrElement"
+	CommandLineOptionTypeCustom        CommandLineOptionKind = "custom" //map
+)
+
+type CommandLineOption struct {
+	kind            CommandLineOptionKind
```
gabritto: I think at the very least `kind` and `name` should be public. Some code in the test harness uses those fields.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+	affectsBindDiagnostics,
+	affectsSourceFile,
+	affectsModuleResolution,
+	affectsEmit,
+
+	allowJsFlag,
+	strictFlag bool
+
+	// transpileoptions worker
+	transpileOptionValue core.Tristate // i think this can be reduced to boolean
```
gabritto: I think you could leave a `// TODO` here to revisit this.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return CommandLineOptionDeprecated[option.name]
+}
+func (option *CommandLineOption) TypeMap() *collections.OrderedMap[string, string] {
+	if option.kind != CommandLineOptionTypeCustom {
+		return nil
+	}
+	return CommandLineOptionCustomType[option.name]
+}
+func (option *CommandLineOption) Elements() *CommandLineOption {
```
gabritto: Why are those methods on `CommandLineOption`? How is it going to be used?

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/microsoft/typescript-go/internal/collections"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+type mapEntry[K comparable, V any] struct {
+	key   K
+	value V
+}
+
+func NewOrderedMapFromList[K comparable, V any](items []mapEntry[K, V]) *collections.OrderedMap[K, V] {
```
gabritto: Is that needed because there's no nice constructor for `OrderedMap`? I guess that might be useful in other places, so maybe it should be moved to the ordered map file.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+	{"es2015", string(core.ModuleKindES2015)},
+	{"es2020", string(core.ModuleKindES2020)},
+	{"es2022", string(core.ModuleKindES2022)},
+	{"esnext", string(core.ModuleKindESNext)},
+	{"node16", string(core.ModuleKindNode16)},
+	{"nodenext", string(core.ModuleKindNodeNext)},
+	{"preserve", string(core.ModuleKindPreserve)},
+})
+
+// TODO: cleanup??
```
gabritto: Maybe add your name to the `TODO` as well? As in `TODO(iisaduan)`.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+	{"react-jsx", string(core.JsxEmitReactJSX)},
+	{"react-jsxdev", string(core.JsxEmitReactJSXDev)},
+})
+
+var newLineOptionMap = NewOrderedMapFromList([]mapEntry[string, string]{
+	{"crlf", string(rune(0))},
+	{"lf", string(rune(1))},
+})
+
+// TODO: figure out where these definitions should be, or if they're still needed
```
gabritto: I looked this up yesterday, and I think those need to be added to `compilerOptions`.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+var moduleDetectionOptionMap = NewOrderedMapFromList([]mapEntry[string, string]{
+	{"auto", string(ModuleDetectionKindAuto)},
+	{"legacy", string(ModuleDetectionKindLegacy)},
+	{"force", string(ModuleDetectionKindForce)},
+})
+
+// todo: remove? mark deprecated??
+// type ImportsNotUsedAsValues int32
```
gabritto: Probably @andrewbranch knows this.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+	{"esnext", string(core.ModuleKindESNext)},
+	{"node16", string(core.ModuleKindNode16)},
+	{"nodenext", string(core.ModuleKindNodeNext)},
+	{"preserve", string(core.ModuleKindPreserve)},
+})
+
+// TODO: cleanup??
+type ModuleDetectionKind int32
+
+const (
```
gabritto: Should that be in `compileroptions.go`? If so, maybe also add a `ModuleDetectionKindNone = 0` too, so that we can represent an unspecified module detection (equivalent to `undefined` in TS).

internal/options/optionsdeclarations.go
```diff
@@ -0,0 +1,1210 @@
+package options
+
+import (
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+var optionsDeclarations = []CommandLineOption{
+	//******* commandOptionsWithoutBuild *******
```
gabritto: Will this slice be split up later? If not, what strategy will we use for the places in ts-ts where we use e.g. the `commandOptionsWithoutBuild` array?

internal/options/errors.go
```diff
@@ -0,0 +1,85 @@
...
+		stringNames = formatCustomTypeKeys(namesOfType)
+	}
+	optName := fmt.Sprintf(`--%v`, opt.name)
+	return ast.NewDiagnostic(nil, loc, diagnostics.Argument_for_0_option_must_be_Colon_1, optName, stringNames)
+}
+
+func formatCustomTypeKeys(keys []string) string {
+	var output strings.Builder
+
+	fmt.Fprintf(&output, "Invalid custom type: '%v'", keys[0])
```
gabritto: Why do we print this?

internal/options/commandlineoption.go
gabritto: Now that I've read the code in `nameMap`, maybe `name` won't be needed after all. `kind` may also not be needed externally if this package exports a function that takes in a string and a command line option and returns a command line value for that option.

internal/options/elementmap.go
andrewbranch: Yes, remove

internal/options/errors.go
iisaduan: This for error message formatting, it prints to `output` (not on the console)

internal/options/optionsdeclarations.go
iisaduan: Yes, I'm planning on re-splitting things up later, when I get to generating things via script. I didn't want to block you or anyone else who needs `optionsDeclarations` any more than needed

internal/options/commandlineoption.go
iisaduan: They are for optional properties (example https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/compiler/types.ts#L7759)

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// used in output in serializing and generate tsconfig
+	category *diagnostics.Message
+
+	// defined once
+	extraValidation *func(value CompilerOptionsValue) (d *diagnostics.Message, args []string)
+
+	// true or undefined
+	// used for configDirTemplateSubstitutionOptions
+	allowConfigDirTemplateSubstitution,
```
jakebailey: Just as a matter of style, I would write each of these out as `something bool`; the comma syntax is really only used when multiple things are very very related, e.g. you might have `X, Y, Z int` on a vector struct or something. Not so much these bools IMO.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// transpileoptions worker
+	transpileOptionValue core.Tristate // i think this can be reduced to boolean
+	// options[option.name] = option.transpileOptionValue;
+
+	// used in listtype
+	listPreserveFalsyValues bool
+}
+
+func (option *CommandLineOption) DeprecatedKeys() *map[string]bool {
```
jakebailey: As a Go style, these method receivers are usually short names, i.e. `o`, not `option`.

jakebailey: `*map[string]bool` feels odd; `maps` can already be nil, so I don't think these need to be like this?

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+	"module":           moduleOptionMap,
+	"target":           targetOptionMap,
+	"moduleDetection":  moduleDetectionOptionMap,
+	"jsx":              jsxOptionMap,
+	"newLine":          newLineOptionMap,
+}
+
+// deprecatedKeys map[string]bool
+var CommandLineOptionDeprecated = map[string]*(map[string]bool){
+	"moduleResolution": &map[string]bool{"node": true},
```
jakebailey: I wouldn't leave these as pointers; maps are already nillable. Any time we care about "present or not present", we should consider using the two-return form where we include a boolean that indicates whether or not something was read.

internal/options/elementmap.go
jakebailey: I would agree, though the OrderedMap type itself does not have any associated "pair" type that would be convenient. Perhaps it should especially if we are doing this often, which I suspect we will be doing in the future. So, yeah, I would move this functionality over to the other pacakge.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
+package options
```
jakebailey: I suspect this package name will come back to bite us; `options` is a convenient variable name and will probably conflict with something...

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+})
+
+var moduleResolutionOptionMap = NewOrderedMapFromList([]mapEntry[string, string]{
+	{"node16", core.ModuleResolutionKindNode16.String()},
+	{"nodenext", core.ModuleResolutionKindNodeNext.String()},
+	{"bundler", core.ModuleResolutionKindBundler.String()},
+})
+
+var targetOptionMap = NewOrderedMapFromList([]mapEntry[string, string]{
+	{"es3", string(core.ScriptTargetES3)},
```
jakebailey: I suspect these conversions are wrong; the script targets are numbers, so this conversion is effectively converting those random numbers to `rune`s. So like, `0` is NUL, etc etc

You probably need a function that prints these out by name instead.

internal/options/errors.go
```diff
@@ -0,0 +1,85 @@
...
+
+func createDiagnosticForInvalidCustomType(opt *CommandLineOption, loc core.TextRange) *ast.Diagnostic {
+	namesOfType := slices.Collect(opt.TypeMap().Keys())
+	stringNames := ""
+	if opt.deprecatedKeys() != nil {
+		stringNames = formatCustomTypeKeys(core.Filter(namesOfType, func(k string) bool { return (*opt.deprecatedKeys())[k] }))
+	} else {
+		stringNames = formatCustomTypeKeys(namesOfType)
+	}
+	optName := fmt.Sprintf(`--%v`, opt.name)
```
jakebailey: I would suggest `%s` for most of these, since if it's _not_ a string a linter will complain about it, versus silently allowing the wrong thing.

internal/options/optionsdeclarations.go
jakebailey, line 1: meta point but so happy to have this particular thing in its own file and not in a file I actually have to read often like it is in TS

internal/options/commandlineoption.go
iisaduan: I wasn't really sure what the best way to use a map to make a set is (that is what I am trying to do here)

internal/options/commandlineoption.go
iisaduan: Do you think it's worth combining all of these to a single flag?

internal/options/commandlineoption.go
iisaduan: the `commandLineOption` struct is to show the shape of the definition of the options, and theoretically shouldn't know anything about other options besides the current one. I've thought about doing a lookup for _all_ the options, since the names are unique, but I dont know about the usecases atm

internal/options/commandlineoption.go
jakebailey: You can use `core.Set[string]`. Or `collections.OrderedSet[string]` if insertion order matters.

internal/options/commandlineoption.go
jakebailey: No, I think separate bools is fine.

Making them flags is a separate optimization that differs from the original TS that we could do later.

internal/options/elementmap.go
iisaduan: The type in TS was supposed to be `Map<string, string|number>`. I took another look through the TS codebase, and it seems like a vast majority of these custom option maps are to numbers, so I think I will refactor the ones that are strings to be enums/numbers as well

internal/options/commandlineoption.go
gabritto: I also remembered the test harness defines its own command line option things, which are specific to tests...

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"github.com/microsoft/typescript-go/internal/collections"
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+type CommandLineOptionKind string
+
+const (
+	CommandLineOptionTypeString        CommandLineOptionKind = "string"
```
DanielRosenwasser, line 12: Does it make sense to only use these declarations in each of the `type` fields?

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+	CommandLineOptionTypeNumber        CommandLineOptionKind = "number"
+	CommandLineOptionTypeBoolean       CommandLineOptionKind = "boolean"
+	CommandLineOptionTypeObject        CommandLineOptionKind = "object"
+	CommandLineOptionTypeList          CommandLineOptionKind = "list"
+	CommandLineOptionTypeListOrElement CommandLineOptionKind = "listOrElement"
+	CommandLineOptionTypeCustom        CommandLineOptionKind = "custom" //map
```
DanielRosenwasser:
```suggestion
	CommandLineOptionTypeCustom        CommandLineOptionKind = "custom" // mapped to the appropriate type
```

I might also just name these "enum", but the original code calls these "custom" types.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (option *CommandLineOption) DisallowNullOrUndefined() bool {
+	return option.name == "extends"
+}
+
+// elements *CommandLineOption
+var CommandLineOptionElements = map[string]*CommandLineOption{
+	"lib": {
+		name:                    "lib",
+		kind:                    "custom", //libMap,
```
DanielRosenwasser: Make sure there's always a space after the beginning of a comment

```suggestion
		kind:                    "custom", // libMap,
```

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+		defaultValueDescription: core.TSUnknown,
```
DanielRosenwasser, line 94: Feels like this should be formatted consistently with the above. Do you have Go's format-on-save on?

```suggestion
		defaultValueDescription:  core.TSUnknown,
```

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,141 @@
...
+		name: "condition",
+		kind: "string",
+	},
+	"plugins": {
+		name: "plugin",
+		kind: "object",
+	},
+}
+
+// typeMap *map[string]string
```
DanielRosenwasser: What's the purpose of these comments? I'm not exactly sure what they refer to.

internal/options/elementmap.go
```diff
@@ -0,0 +1,204 @@
...
+	{"preserve", string(core.JsxEmitPreserve)},
+	{"react-native", string(core.JsxEmitReactNative)},
+	{"react", string(core.JsxEmitReact)},
+	{"react-jsx", string(core.JsxEmitReactJSX)},
+	{"react-jsxdev", string(core.JsxEmitReactJSXDev)},
+})
+
+var newLineOptionMap = NewOrderedMapFromList([]mapEntry[string, string]{
+	{"crlf", string(rune(0))},
+	{"lf", string(rune(1))},
```
DanielRosenwasser: In the original, these mapped to a specific enum declaration. I wouldn't do something ad-hoc like this, especially what's being used here since the value `string(rune(0))` is the string `\x00` (a string with a `NUL` character), and it wouldn't be human readable if displayed.

I'd go in one of two directions:

1. Change the enums to be string enums (these might be a bigger change in the compiler).
2. Split custom command line types into two kinds of types: integral options and strings. `CommandLineOptionCustomType` would be split into two different maps.

Not sure what others think of this.

internal/options/commandlineoption.go
iisaduan: I like that too, certainly less confusing than "custom"

internal/options/elementmap.go
iisaduan: https://github.com/microsoft/typescript-go/pull/87#discussion_r1852934405 I'm going to change them all to be numbers, so we can have them all be number enums. And yup, this was very ad hoc for now 😅, I just wanted it to compile so others could get a reference for what the struct is going to be

internal/core/compileroptions.go
```diff
@@ -180,6 +188,13 @@ func (m ModuleResolutionKind) String() string {
 	}
 }
 
+type NewLineKind int32
+
+const (
+	NewLineKindCRLF = 0
```
jakebailey: These are missing their types.

internal/options/commandlineoption.go
iisaduan: I improved the comments, they were to document what they corresponded to from the original types

internal/options/elementmap.go
iisaduan: update: changed the type to any

navya9singh:
Looks good for tsconfig parsing!

internal/options/elementmap.go
navya9singh: `parsingoptions` maybe?

internal/collections/ordered_map.go
```diff
@@ -33,6 +33,19 @@ func newMapWithSizeHint[K comparable, V any](hint int) OrderedMap[K, V] {
 	}
 }
 
+type MapEntry[K comparable, V any] struct {
+	Key   K
+	Value V
+}
+
+func NewOrderedMapFromList[K comparable, V any](items []MapEntry[K, V]) *OrderedMap[K, V] {
```
jakebailey, line 41: I'm not sure this matters much, but is there a usability reason why this isn't a `...` arg as opposed to a single arg?

I assume it's because you'd have to fully qualify the names, or, just make a slice anyway to avoid that?

jakebailey, line 41: One thing to consider here is the size of `[]MapEntry` vs `[]*MapEntry`, but I'm not sure how large this ends up being.

Perhaps this should actually take `iter.Seq2[K, V]` and be passed in `slices.All(entries)` where it's used?

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+	// used in output
+	description              *diagnostics.Message
+	defaultValueDescription  any
+	showInSimplifiedHelpView bool
+
+	// used in output in serializing and generate tsconfig
+	category *diagnostics.Message
+
+	// defined once
+	extraValidation *func(value CompilerOptionsValue) (d *diagnostics.Message, args []string)
```
jakebailey: No need to be `*` here; `func` is already nillable.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+		Kind: CommandLineOptionTypeString,
+	},
+	"plugins": {
+		Name: "plugin",
+		Kind: CommandLineOptionTypeObject,
+	},
+}
+
+// CommandLineOption.EnumMap()
+var commandLineOptionEnumMap = map[string]*(collections.OrderedMap[string, any]){
```
jakebailey: I think these parens are redundant:

```suggestion
var commandLineOptionEnumMap = map[string]*collections.OrderedMap[string, any]{
```

Go types parse left to right so usually avoid precedence issues that other languages have with pointers.

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+	"moduleResolution": moduleResolutionOptionMap,
+	"module":           moduleOptionMap,
+	"target":           targetOptionMap,
+	"moduleDetection":  moduleDetectionOptionMap,
+	"jsx":              jsxOptionMap,
+	"newLine":          newLineOptionMap,
+}
+
+// CommandLineOption.DeprecatedKeys()
+var commandLineOptionDeprecated = map[string](map[string]bool){
```
jakebailey:
```suggestion
var commandLineOptionDeprecated = map[string]map[string]bool{
```

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+	"moduleResolution": map[string]bool{"node": true},
```
jakebailey: Can we just use our own `Set[string]`? `map[T]struct{}` is more common than `map[t]bool` for sets just because the values don't take up any space.

internal/collections/ordered_map.go
jakebailey: I guess this is only ever used with slice literals so that optimization is probably not needed.

internal/options/enummaps.go
```diff
@@ -0,0 +1,165 @@
+package options
+
+import (
+	"github.com/microsoft/typescript-go/internal/collections"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+var libMap = collections.NewOrderedMapFromList([]collections.MapEntry[string, any]{
```
jakebailey, line 8: Maybe not for this PR, but I would be interested in what interface based abstraction we could come up with here to avoid `any`. You can imagine defining an interface like:

```go
type OptionValue interface {
	Something() string
}
```

Then making named wrapper types which encapsulate the real values, preventing us from accidentally sticking anything and everything into these.

But, I don't think that's strictly required.

internal/options/enummaps.go
```diff
@@ -0,0 +1,165 @@
...
+	// JavaScript only
+	{Key: "es5", Value: "lib.es5.d.ts"},
```
jakebailey, line 10: If you wanted to avoid writing this repeatedly, theoretically you could write:

```go
func mapEntry[K comparable, V any](k K, v V) collections.MapEntry[K, V] {
	return collections.MapEntry[K, V]{Key: k, Value v}
}

var libMap = collections.NewOrderedMapFromList([]collections.MapEntry[string, any]{
	mapEntry("es6", "lib.es2015.d.ts"),
	mapEntry("es2015", "lib.es2015.d.ts"),
	mapEntry("es7", "lib.es2016.d.ts"),
	///
}
```

But, this existing code is not really bad. (Trying to decide if the current code is actually better because it's "static" but I do not know if the Go compiler is smart enough to produce one object at compile time for this.)

internal/collections/ordered_map.go
iisaduan: We could add the pointer version if needed later

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
+package options
+
+import (
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+var optionsDeclarations = []CommandLineOption{
```
jakebailey, line 8: Just to confirm my understanding, is it the case that we know _exactly_ which of these options will map to CompilerOptions and which will not?

If so, I'm wondering if we can write a `reflect` based test which does something like:

```go
typ := reflect.TypeFor[CompilerOptions]()

for i := range typ.NumField() {
	field := typ.Field(i)
    // assert that field is declared in options
}

for _, opt := range optionDeclarations {
   // assert that opt is in CompilerOptions
}
```

internal/options/enummaps.go
iisaduan: Should I change it now? I will likely end up adjusting the definitions as they exist in this PR (I was considering generating them), once I figure out what declarations we actually still need/need to define/how we should organize them

jakebailey: No, I don't think it's a requirement for checking in this PR.

internal/options/elementmap.go
jakebailey: This is the package with options declaration, so, `optiondecls`? `tsoptions`? Not sure. It's possible we don't end up with a variable named `options`, though, could just be `opts` or `compilerOptions`.

internal/options/declscompiler.go
iisaduan: Yes, something like that would be nice. In the TS compiler, there is a is a variable that contains all the ones that translate to `CompilerOptions` properties, but we haven't implemented it in go yet

internal/collections/ordered_map.go
iisaduan: Yup, just as you mentioned, it just happened to be easier to write this way. I guess we could add a `...` arg onto the end, or make a different function that only has a `...` arg (perhaps later)

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+type CommandLineOptionKind string
+
+const (
+	CommandLineOptionTypeString        CommandLineOptionKind = "string"
+	CommandLineOptionTypeNumber        CommandLineOptionKind = "number"
+	CommandLineOptionTypeBoolean       CommandLineOptionKind = "boolean"
+	CommandLineOptionTypeObject        CommandLineOptionKind = "object"
+	CommandLineOptionTypeList          CommandLineOptionKind = "list"
+	CommandLineOptionTypeListOrElement CommandLineOptionKind = "listOrElement"
+	CommandLineOptionTypeEnum          CommandLineOptionKind = "enum" //map
```
DanielRosenwasser:
```suggestion
	CommandLineOptionTypeEnum          CommandLineOptionKind = "enum" // map
```

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
...
+		description:             diagnostics.Print_names_of_files_that_are_part_of_the_compilation_and_then_stop_processing,
+		defaultValueDescription: false,
+	},
+
+	// Basic
+	// targetOptionDeclaration,
+	{
+		Name:                     "target",
+		shortName:                "t",
+		Kind:                     CommandLineOptionTypeEnum, //targetOptionMap
```
DanielRosenwasser:
```suggestion
		Kind:                     CommandLineOptionTypeEnum, // targetOptionMap
```

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
...
+		category:                 diagnostics.Language_and_Environment,
+		description:              diagnostics.Set_the_JavaScript_language_version_for_emitted_JavaScript_and_include_compatible_library_declarations,
+		defaultValueDescription:  core.ScriptTargetES5,
+	},
+
+	// moduleOptionDeclaration,
+	{
+		Name:                     "module",
+		shortName:                "m",
+		Kind:                     CommandLineOptionTypeEnum, //moduleOptionMap
```
DanielRosenwasser:
```suggestion
		Kind:                     CommandLineOptionTypeEnum, // moduleOptionMap
```

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
...
+		category:                 diagnostics.Modules,
+		description:              diagnostics.Specify_what_module_code_is_generated,
+		defaultValueDescription:  core.TSUnknown,
+	},
+	{
+		Name: "lib",
+		Kind: CommandLineOptionTypeList,
+		// elements: &CommandLineOption{
+		// 	name:                    "lib",
+		// 	kind:                   CommandLineOptionTypeEnum, //libMap,
```
DanielRosenwasser:
```suggestion
		// 	kind:                   CommandLineOptionTypeEnum, // libMap,
```

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
...
+		affectsSemanticDiagnostics: true,
+		affectsBuildInfo:           true,
+		showInSimplifiedHelpView:   true,
+		category:                   diagnostics.JavaScript_Support,
+		description:                diagnostics.Enable_error_reporting_in_type_checked_JavaScript_files,
+		defaultValueDescription:    false,
+	},
+	{
+		Name:                    "jsx",
+		Kind:                    CommandLineOptionTypeEnum, //jsxOptionMap,
```
DanielRosenwasser:
```suggestion
		Kind:                    CommandLineOptionTypeEnum, // jsxOptionMap,
```

internal/options/declscompiler.go
```diff
@@ -0,0 +1,1195 @@
...
+		Kind:                    CommandLineOptionTypeBoolean,
+		affectsEmit:             true,
+		affectsBuildInfo:        true,
+		category:                diagnostics.Emit,
+		description:             diagnostics.Emit_a_UTF_8_Byte_Order_Mark_BOM_in_the_beginning_of_output_files,
+		defaultValueDescription: false,
+	},
+	{
+		Name:                    "newLine",
+		Kind:                    CommandLineOptionTypeEnum, //newLineOptionMap,
```
DanielRosenwasser:
```suggestion
		Kind:                    CommandLineOptionTypeEnum, // newLineOptionMap,
```

internal/options/commandlineoption.go
```diff
@@ -0,0 +1,142 @@
...
+	affectsEmit                bool
+
+	allowJsFlag bool
+	strictFlag  bool
+
+	// used in transpileoptions worker
+	// todo: revisit to see if this can be reduced to boolean
+	transpileOptionValue core.Tristate
+
+	// used in listtype
```
DanielRosenwasser:
```suggestion
	// used for CommandLineOptionTypeList
```

internal/options/commandlineoption.go
iisaduan: not sure why my formatting isn't catching these 😅

jakebailey: Comments without a space are special and treated as "directives", so gofmt does not reformat them.

There's a linter plugin that would enforce this for all comments that don't have the form `//blah:something` which cover `//go:generate`, `//nolint:errcheck`, etc, but it's not enabled.

PR #89, opened 2024-11-21
Remove duplicate numberToString/stringToNumber, restore TODOs

jakebailey:
#88 moved these functions but not all uses; it also didn't retain the TODOs since we haven't actually verified that these functions do the same thing that JS does.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -294,14 +294,14 @@ func IfElse[T any](b bool, whenTrue T, whenFalse T) T {
 
 // This function should behave identically to the expression `"" + f` in JS
 func NumberToString(f float64) string {
-	// TODO: verify that this is actaully the same as JS.
+	// !!! verify that this is actually the same as JS.
 	return strconv.FormatFloat(f, 'g', -1, 64)
 }
 
 // This function should behave identically to the expression `+s` in JS, including parsing binary, octal, and hex
 // numeric strings
 func StringToNumber(s string) float64 {
-	// TODO: verify that this is actaully the same as JS.
+	// !!! verify that this is actually the same as JS.
 	value, err := strconv.ParseFloat(s, 64)
 	if err != nil {
 		return math.NaN()


internal/core/core.go
```diff
@@ -294,12 +294,14 @@ func IfElse[T any](b bool, whenTrue T, whenFalse T) T {
 
 // This function should behave identically to the expression `"" + f` in JS
 func NumberToString(f float64) string {
+	// TODO: verify that this is actaully the same as JS.
```
ahejlsberg: Can we use `!!!` instead of `TODO` since that's what we have everywhere else.

jakebailey:
```suggestion
	// !!! verify that this is actually the same as JS.
```

internal/core/core.go
```diff
@@ -294,12 +294,14 @@ func IfElse[T any](b bool, whenTrue T, whenFalse T) T {
...
 	return strconv.FormatFloat(f, 'g', -1, 64)
 }
 
 // This function should behave identically to the expression `+s` in JS, including parsing binary, octal, and hex
 // numeric strings
 func StringToNumber(s string) float64 {
+	// TODO: verify that this is actaully the same as JS.
```
jakebailey:
```suggestion
	// !!! verify that this is actually the same as JS.
```

internal/core/core.go
rbuckton: > Can we use `!!!` instead of `TODO` since that's what we have everywhere else.

Does this have any special meaning in Go tooling? I've seen plenty of tooling that indexes on `// TODO:` in various editors and extensions for other languages (i.e., C# in Visual Studio will show `// TODO` comments in a list for your project), but I haven't seen `// !!!` used frequently elsewhere. Also, we have TODOs in comments in existing code, should we be changing those over to `!!!` as well?

ahejlsberg: The `!!!` markers specifically relate to the porting effort. We should keep that separate from TODOs that are already in the code.

jakebailey: Yeah, that's the main distinction (though I am using a normal todo in my own PRs so I can burn them down)

PR #90, opened 2024-11-21
Port `checkSourceElement`, `checkExpression`, `checkBinaryExpression`

ahejlsberg:
This PR ports `checkSourceElement` and `checkExpression` and scaffolds out all of the functions they call. Also ports `checkBinaryExpression` and associated helpers.


jakebailey:
Seems like there's a conflict after the scanner PRs went in.

PR #97, opened 2024-11-22
Port large portion of pretty printer

rbuckton:
This ports a large chunk of the pretty printer from the TS compiler. While a fair bit has been migrated there are a few pieces that are still in progress, such as Parenthesization, JSX emit, JSDoc emit, Comment emit, Source map emit, temp variable name generation, etc.. There are also a few functions that have been stubbed out for later implementation as they will depend on whether we end up with a transforming emitter or bespoke emitters.


internal/compiler/parser.go
```diff
@@ -2017,7 +2017,7 @@ func (p *Parser) parseModuleExportName(disallowKeywords bool) *ast.Node {
 
 func (p *Parser) tryParseImportAttributes() *ast.Node {
 	if (p.token == ast.KindWithKeyword || p.token == ast.KindAssertKeyword) && !p.hasPrecedingLineBreak() {
-		return p.parseImportAttributes(p.token, true /*skipKeyword*/)
+		return p.parseImportAttributes(p.token, false /*skipKeyword*/)
```
jakebailey, line 2029: Ah, good catch!

internal/core/core.go
```diff
@@ -292,6 +293,25 @@ func IfElse[T any](b bool, whenTrue T, whenFalse T) T {
...
+
+// Returns `a` if `a` is not `nil`; Otherwise, returns `b`. Coalesce is roughly analogous to `??` in JS, except that it
+// non-shortcutting, so it is advised to only use a constant or precomputed value for `b`
+func Coalesce[T any](a T, b T) T {
+	if IsNil(a) {
+		return b
+	} else {
+		return a
+	}
+}
```
jakebailey: Reading the uses, I don't actually think we need to use reflection here; the zero value seems to be sufficient.

```suggestion
// Returns `a` if `a` is not a zero value; Otherwise, returns `b`. Coalesce is roughly analogous to `??` in JS, except that it
// non-shortcutting, so it is advised to only use a constant or precomputed value for `b`
func Coalesce[T any](a T, b T) T {
	var zero T
	if a == zero {
		return b
	} else {
		return a
	}
}
```

internal/printer/printer.go
```diff
@@ -0,0 +1,4296 @@
...
+		panic(fmt.Sprintf("unexpected node: %v", node.Kind))
+	}
+	return scanner.GetSourceTextOfNodeFromSourceFile(p.currentSourceFile, node, includeTrivia)
+}
+
+//
+// Low-level writing
+//
+
+type WriteKind int
```
jakebailey, line 165: Is `WriteKind` ever used by the public API? I don't see it being used outside the package. Same with `QuoteChar`. Not getting the case right isn't a strict requirement at the moment (loads of the checker get the exported-ness wrong), but something to think about.

internal/printer/printer.go
```diff
@@ -0,0 +1,4296 @@
...
+	LFTypeArguments                     ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFAngleBrackets | LFOptional
+	LFTypeParameters                    ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFAngleBrackets | LFOptional
+	LFParameters                        ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFParenthesis
+	LFSingleArrowParameter              ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine
+	LFIndexSignatureParameters          ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFIndented | LFSquareBrackets
+	LFJSDocComment                      ListFormat = LFMultiLine | LFAsteriskDelimited
+	/** @deprecated */ LFImportClauseEntries ListFormat = LFImportAttributes
+)
+
+var brackets = map[ListFormat][2]string{
```
jakebailey: FWIW you could totally just use a struct here instead of an array; they'll have the same layout but you'll be able to give things names. The long-form struct literal is technically optional, and for cases where all values are provided I think no linter will get mad. But, it's not a big deal or anything.

internal/printer/printer.go
```diff
@@ -0,0 +1,4296 @@
...
+	LFHeritageClauseTypes               ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine
+	LFSourceFileStatements              ListFormat = LFMultiLine | LFNoTrailingNewLine
+	LFDecorators                        ListFormat = LFMultiLine | LFOptional | LFSpaceAfterList
+	LFTypeArguments                     ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFAngleBrackets | LFOptional
+	LFTypeParameters                    ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFAngleBrackets | LFOptional
+	LFParameters                        ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFParenthesis
+	LFSingleArrowParameter              ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine
+	LFIndexSignatureParameters          ListFormat = LFCommaDelimited | LFSpaceBetweenSiblings | LFSingleLine | LFIndented | LFSquareBrackets
+	LFJSDocComment                      ListFormat = LFMultiLine | LFAsteriskDelimited
+	/** @deprecated */ LFImportClauseEntries ListFormat = LFImportAttributes
```
jakebailey:
```suggestion
    // Deprecated: < some comment >
	LFImportClauseEntries ListFormat = LFImportAttributes
```

Will render in gopls / pkgsite if you want.

internal/core/core.go
ahejlsberg: An even shorter way of writing the zero test in generic code is `if a == *new(T)`. Saves you from having to create an unused variable.

ahejlsberg: But I agree, *definitely* don't want to use reflection here.

jakebailey: Reading again, it seems like this is explicitly making it _not_ treat zero values of string, int, etc, as zero and passing them through, but the callers are only pointers. It seems like this could just be a pointer constraint?

jakebailey: (Or, just write the if statement?)

jakebailey: That'd be:

```suggestion
// Returns `a` if `a` is not a zero value; Otherwise, returns `b`. Coalesce is roughly analogous to `??` in JS, except that it
// non-shortcutting, so it is advised to only use a constant or precomputed value for `b`
func Coalesce[T *U, U any](a T, b T) T {
	if a == nil {
		return b
	} else {
		return a
	}
}
```

internal/printer/utilities.go
```diff
@@ -0,0 +1,654 @@
...
+		if v != nil {
+			return v.End(), true
+		}
+	case (*core.TextRange):
+		if v != nil {
+			return v.End(), true
+		}
+	default:
+		// worst case, use reflection
+		if !core.IsNil(node) {
```
jakebailey: Hm, how is this compiling? Isn't `node` a non-pointer here?

jakebailey: Oh, this is still using reflection, IsNil wasn't removed

jakebailey: Can this just panic instead?

rbuckton: Why would be a non-pointer? It could potentially be a pointer to a derived node type

rbuckton: Also, `IsNil` has `[T any]`, it doesn't care if it's a pointer.

jakebailey: Right, I didn't realize you replaced `Coalesce` with a version without reflection but left `IsNil` and so my wires got crossed.

rbuckton: I left it as it was still used by `greatestEnd`, though I did modify `greatestEnd` to only use it as a last resort.

internal/printer/utilities.go
```diff
@@ -0,0 +1,654 @@
...
+		// If a NoSubstitutionTemplateLiteral appears to have a substitution in it, the original text
+		// had to include a backslash: `not \${a} substitution`.
+		var b strings.Builder
+		text := node.TemplateLiteralLikeData().Text
+		rawText := node.TemplateLiteralLikeData().RawText
+
+		// Write leading quote character
+		switch node.Kind {
+		case ast.KindNoSubstitutionTemplateLiteral:
+			b.Grow(2)
```
jakebailey: FWIW, I'm not sure that `Grow` is worth it in these cases; what will most likely happen is that the buffer is grown to handle 2 more characters, but then the text will be written and use anything that was added. What I think might be the most reasonable is to optimize for the most common case where nothing is modified, and just grow to be that size at the start.

internal/printer/utilities.go
jakebailey: I feel like I'd still like to see reflection not used at all; it sure seems like there's a closed set of things we could call on...

rbuckton: It'd be nice if Go didn't require reflection to check for `nil` for interfaces, or that Go had `??` or `?.`, as that is what `greatestEnd` and `Coalesce` are replacing from the existing code. I expect to use `greatestEnd` and `Coalesce` in other places once we get to the actual transformation/downleveling part of the emitter, and there's no guarantee that will be limited to just `Node`, `NodeList`, `TextRange`, and `ModifierList` as it could also be any of the `Node` subtypes, or potentially other types like `SourceMapRange` if necessary. Issuing a `panic()` here seems more like whack-a-mole, where I'll have to keeping checking for a crash as I add/port functionality, vs having a slower fallback for less frequent corner cases that can be optimized to a fast path when necessary.

internal/printer/utilities.go
rbuckton: I'm not sure I understand what you mean by "optimize for the most common case where nothing is modified". It's rare for a string to be empty, even for template literal segments, so the "nothing is modified" case seems like it is not common. 

In the code that immediately follows this `switch` I call `Grow` prior to writing any chunk of text, so the extra space requested won't be immediately consumed. If the incremental `Grow` calls aren't efficient, would it be better to precompute the length of the string ahead of time (essentially running the same `escapeString` algorithm twice) to get the total length?

My understanding of `Grow` is that it checks the buffer capacity and if there won't be enough space it allocates a new buffer with double the capacity + `n`, so it shouldn't necessarily be resizing each time `Grow` is called, though it certainly is more efficient to be able to estimate the length up front so that an incremental `Grow` doesn't unnecessarily double the capacity towards the end of writing.

Another approach might be estimating the worst case capacity for an escaped string given the provided `getLiteralTextFlags` and growing to that size, i.e., for a given JS string, multiply the length by 12 (for `\u####\u####`, to account for surrogate pairs) plus the size for the quote characters, or for a JSX encoded string multiply the length by 10 (for `&#x######;`, to account for the maximum encoded unicode character). That seems potentially wasteful, however, given that the majority of characters don't need escaping.

internal/printer/utilities.go
jakebailey: I don't see that wart ever being fixed; nil values as interfaces is a totally legal and useful construct.

It would make more sense to me to instead do nil checks on the specific implementations of the interface, letting them default to `0` or `-1` or something.

It just seems like a sisyphean task to attempt to port TS syntax exactly 1:1 to the point of using reflection to emulate `| undefined` from TS, or to introduce helpers to emulate `??` or `?.` when an if statement would be more typical for Go.

internal/printer/utilities.go
jakebailey: > I'm not sure I understand what you mean by "optimize for the most common case where nothing is modified". It's rare for a string to be empty, even for template literal segments, so the "nothing is modified" case seems like it is not common.

Maybe I'm misunderstanding this code. My interpretation is that `getLiteralText` reads some sort of value out of the AST and turns it back into a string. My guess was that it'd be relatively efficient to just grow that ahead of time to roughly the "unescaped" (ish) form, with the assumption that most of the time, that's enough.

> If the incremental `Grow` calls aren't efficient, would it be better to precompute the length of the string ahead of time (essentially running the same `escapeString` algorithm twice) to get the total length?

Memory wise, yes, that would be the most efficient, but the 2*n growth pattern is typically sufficient to make this not matter.


> My understanding of `Grow` is that it checks the buffer capacity and if there won't be enough space it allocates a new buffer with double the capacity + `n`, so it shouldn't necessarily be resizing each time `Grow` is called

Hm, reading the implementation, yeah. I don't understand why when they grow `n` more bytes, they actually grow to `2*cap(slice) + n`...



> That seems potentially wasteful, however, given that the majority of characters don't need escaping.

Right, that's why I was suggesting just growing once to the "typical" expected size of the string, letting the other rare escaped cases needing to grow automatically. Then you don't need to worry about calling grow except that one single time.

internal/printer/utilities.go
rbuckton: > It would make more sense to me to instead do nil checks on the specific implementations of the interface, letting them default to `0` or `-1` or something.

`0` and `-1` have specific meanings and reaching for something further like `-2` or otherwise seems completely undiscoverable.

> It just seems like a sisyphean task to attempt to port TS syntax exactly 1:1 to the point of using reflection to emulate `| undefined` from TS, or to introduce helpers to emulate `??` or `?.` when an if statement would be more typical for Go.

The `if` statements necessary would be overly complex and far harder to maintain than the simple loop used by `greatestEnd`. I can avoid the `IsNil` for this, though, and just handle corner cases as they come.

internal/printer/utilities.go
rbuckton: > > I'm not sure I understand what you mean by "optimize for the most common case where nothing is modified". It's rare for a string to be empty, even for template literal segments, so the "nothing is modified" case seems like it is not common.
> 
> Maybe I'm misunderstanding this code. My interpretation is that `getLiteralText` reads some sort of value out of the AST and turns it back into a string. My guess was that it'd be relatively efficient to just grow that ahead of time to roughly the "unescaped" (ish) form, with the assumption that most of the time, that's enough.

I think my confusion here was that by "nothing is modified" you meant "nothing requires escaping", while I had read your comment as meaning "nothing is modifying *the builder*".

internal/printer/utilities.go
```diff
@@ -0,0 +1,656 @@
...
+			return v.End(), true
+		}
+	case (*core.TextRange):
+		if v != nil {
+			return v.End(), true
+		}
+	case (core.TextRange):
+		return v.End(), true
+	default:
+		panic("unhandled type: " + reflect.TypeOf(node).Name())
```
jakebailey:
```suggestion
	default:
		panic(fmt.Sprintf("unhandled type: %T", node))
```

This will remove the need for direct reflection.

internal/ast/utilities.go
```diff
@@ -28,6 +28,44 @@ func NodeKindIs(node *Node, kinds ...Kind) bool {
 	return slices.Contains(kinds, node.Kind)
 }
 
+func IsModifierKind(token Kind) bool {
```
DanielRosenwasser, line 31: Can you make sure that you replace the other `isModifierKind` in `compiler`?

internal/ast/utilities.go
```diff
@@ -279,6 +317,21 @@ func IsClassElement(node *Node) bool {
...
+func IsTypeElement(node *Node) bool {
+	switch node.Kind {
+	case KindConstructSignature,
+		KindCallSignature,
+		KindPropertySignature,
+		KindMethodSignature,
+		KindIndexSignature,
+		KindGetAccessor,
+		KindSetAccessor:
+		// KindNotEmittedTypeElement
```
DanielRosenwasser: Maybe add a `!!!` here.

```suggestion
		// !!! KindNotEmittedTypeElement
```

internal/printer/utilities.go
```diff
@@ -0,0 +1,655 @@
...
+	b.WriteString("&#x")
+	b.WriteString(hexCharCode)
+	b.WriteString(";")
+}
+
+func encodeUtf16EscapeSequence(b *strings.Builder, charCode rune) {
+	hexCharCode := strings.ToUpper(strconv.FormatUint(uint64(charCode), 16))
+	b.WriteString(`\u`)
+	for i := len(hexCharCode); i < 4; i++ {
+		b.WriteRune('0')
```
DanielRosenwasser: I think you can just use `WriteByte` in these cases to directly hit a fast path.

```suggestion
		b.WriteByte('0')
```

internal/printer/utilities.go
```diff
@@ -0,0 +1,655 @@
...
+	'\u2028': `\u2028`, // lineSeparator
+	'\u2029': `\u2029`, // paragraphSeparator
+	'\u0085': `\u0085`, // nextLine
+}
+
+func encodeJsxCharacterEntity(b *strings.Builder, charCode rune) {
+	hexCharCode := strings.ToUpper(strconv.FormatUint(uint64(charCode), 16))
+	b.WriteString("&#x")
+	b.WriteString(hexCharCode)
+	b.WriteString(";")
```
DanielRosenwasser:
```suggestion
	b.WriteByte(';')
```

internal/printer/utilities.go
```diff
@@ -0,0 +1,655 @@
...
+					b.WriteString("&#0;")
+				} else if match, ok := jsxEscapedCharsMap[ch]; ok {
+					b.WriteString(match)
+				} else {
+					encodeJsxCharacterEntity(b, ch)
+				}
+
+			default:
+				// Template strings preserve simple LF newlines, still encode CRLF (or CR)
+				if ch == '\r' && quoteChar == quoteCharBacktick && i+1 < len(s) && s[i+1] == '\n' {
```
DanielRosenwasser: I was confused about the comment here, so here's how I'd rephrase:

```suggestion
				if ch == '\r' && quoteChar == quoteCharBacktick && i+1 < len(s) && s[i+1] == '\n' {
				// Template strings can simple LF newlines, but will
				// encode CRLF and CR as LF. So we need to explicitly
				// write the escapes out if we've got anything other than LF.
```

DanielRosenwasser: Also, does this correctly handle a bare `\r`? I don't see that here.

rbuckton: That comment is not correct. In a template literal, `\r` *will* be escaped per the case on line 97-98. `\n` will *not* be escaped per the case on line 99-103. If these two cases were handled independently, then we would escape CRLF as `"\\r\n"`. This case explicitly handles escaping the combined CRLF as `"\\r\\n"`.

I can change the comment to read:

> ```
> Template strings preserve simple LF newlines, but still must escape CRLF. Left alone, the 
> cases for `\r` and `\n` would inadvertently escape CRLF as two independent characters.
> ```

rbuckton: > Also, does this correctly handle a bare `\r`? I don't see that here.

See lines 97-98

internal/core/core.go
rbuckton: Which is better here, `a == *new(T)` or using `[T *U, U any]` and comparing `a == nil`? I've noticed that since I posted this PR that Anders added `OrElse` for pretty much the exact same purpose. While I understand the comparisons differ slightly (the first checks for zero values, while the second checks specifically for `nil`), the cases where `OrElse` is now used would also apply to `Coalesce`.

jakebailey: I feel like both are acceptable, though I would have named `OrElse` something like `ElseIfZero` or something (making `Coalesce` be `ElseIfNil`, but I don't mind `Coalesce`, really). I have a preference for your function to exist and be preferred solely because it represents `??` where `undefined` is a `nil` pointer.

internal/ast/ast.go
```diff
@@ -2895,9 +2983,12 @@ func (f *NodeFactory) NewKeywordExpression(kind Kind) *Node {
 // LiteralLikeBase
 
 type LiteralLikeBase struct {
-	Text string
+	Text       string
+	TokenFlags TokenFlags
```
jakebailey, line 2987: For what it's worth, I perf tested this and there was no CPU usage change, and like a +0.03% memory usage increase (if I remember correctly from last week), so I don't see this being a problem.

internal/core/core.go
rbuckton: I'm not a fan of `ElseIfNil` as a name. The `??` operator is the [null coalesce operator](https://en.wikipedia.org/wiki/Null_coalescing_operator), as named in C# ("null coalesce"), JS ("nullish coalesce"), and even T-SQL (as `COALESCE`). `ElseIfNil` is a more esoteric name, IMO.

rbuckton: Part of my concern over `a == *new(T)` as it implies memory allocation for a zero comparison, only to dereference it with `*` to perform the comparison. I wonder if `OrElse` could be rewritten to avoid using `new`, or whether that actually matters to Go:

```go
func OrElse[T comparable](value T, defaultValue T) T {
	var zero T
	if value != zero {
		return value
	}
	return defaultValue
}
```

jakebailey: With:

```go
func isZero1[T comparable](v T) bool {
    var zero T
    return v == zero
}

func isZero2[T comparable](v T) bool {
    return v == *new(T)
}
```

Putting that into https://godbolt.org/z/PMdza7YPe with some calls using `int`, and removing Go-specific metadata, `isZero1` is:

```asm
        TEXT    command-line-arguments.isZero1[int](SB), DUPOK|NOSPLIT|WRAPPER|NOFRAME|ABIInternal, $0-8
        MOVQ    32(R14), R12
        TESTQ   R12, R12
        JNE     command-line-arguments_isZero1[int]_pc17
command-line-arguments_isZero1[int]_pc9:
        NOP
        NOP
        XCHGL   AX, AX
        TESTQ   AX, AX
        SETEQ   AL
        RET
command-line-arguments_isZero1[int]_pc17:
        LEAQ    8(SP), R13
        CMPQ    (R12), R13
        JNE     command-line-arguments_isZero1[int]_pc9
        MOVQ    SP, (R12)
        JMP     command-line-arguments_isZero1[int]_pc9
```


While `isZero2` is:

```asm
        TEXT    command-line-arguments.isZero2[int](SB), DUPOK|NOSPLIT|WRAPPER|NOFRAME|ABIInternal, $0-8
        MOVQ    32(R14), R12
        TESTQ   R12, R12
        JNE     command-line-arguments_isZero2[int]_pc17
command-line-arguments_isZero2[int]_pc9:
        NOP
        NOP
        XCHGL   AX, AX
        TESTQ   AX, AX
        SETEQ   AL
        RET
command-line-arguments_isZero2[int]_pc17:
        LEAQ    8(SP), R13
        CMPQ    (R12), R13
        JNE     command-line-arguments_isZero2[int]_pc9
        MOVQ    SP, (R12)
        JMP     command-line-arguments_isZero2[int]_pc9
```

So they thankfully compile down to the same thing.

jakebailey: (In general, allocating things that appear to be on the stack versus allocating things that appear to be on the heap a-la `new`, are not actually treated any differently; it's escape analysis which determines that something _can't_ be on the stack)

internal/printer/printer.go
```diff
@@ -0,0 +1,4307 @@
...
+}
+
+type ListFormat int
+
+const (
+	LFNone ListFormat = 0
+
+	// Line separators
+	LFSingleLine    ListFormat = 0      // Prints the list on a single line (default).
+	LFMultiLine     ListFormat = 1 << 0 // Prints the list on multiple lines.
```
jakebailey, line 4196: I assume these are all converted, but, theoretically they can all be using:

```go
LFMultiLine ListFormat = 1 << iota
```

But this is in no way a strict requirement.

internal/printer/utilities.go
DanielRosenwasser: That comment is a bit clearer, thanks!

internal/printer/printer.go
rbuckton: My only concern with using `iota` for flags is that it's harder to see how close you are to running out of bits at a glance.

DanielRosenwasser: Fair, but you will get an error when you do hit the limit though. You won't silently exceed an integral limit.

```
const (
	A Foo = 1 << iota
	B
	C
	D
	E
	F
	G
	H
//      ~
// cannot use 1 << iota (untyped int constant 128) as Foo value in constant declaration (overflows)
)
```

image

jakebailey: Hovering also gets you the value. The errors that one can accidentally introduce without iota are IMO much worse. That and the ergonomics of adding/removing are just not great; you'd only want to handwrite them if they need to be stable for some reason (which may occur for interop).

rbuckton: Either way, I'm not sure `iota` works well here due to how the flags are logically grouped. I'd have to reorder the flags by value and put the masks at the end, rather than having useful explanatory values at use sites. For example, just changing `LFMultiLine` to `1 << iota` gives it the value `4` instead of `1`.

PR #98, opened 2024-11-22
Implement `ResolveTypeReferenceDirective`

andrewbranch:
Also cleans up the resolution tests a lot. I think trace tests will start passing once I add a cache implementation.


testdata/baselines/reference/module/resolver/allowJsCrossMonorepoPackage.trace.json
```diff
@@ -75,89 +75,5 @@
     "======== Module name 'pkg' was successfully resolved to '/node_modules/pkg/index.d.ts'. ========",
     "File '/node_modules/pkg/package.json' does not exist according to earlier cached lookups.",
     "File '/node_modules/package.json' does not exist.",
-    "File '/package.json' does not exist.",
-    "======== Resolving module '@typescript/lib-es5' from '/packages/main/__lib_node_modules_lookup_lib.es5.d.ts__.ts'. ========",
```
andrewbranch, line 79: I deleted these calls from the instrumented tests; they’re just massively duplicated in every test.

internal/compiler/module/resolver_test.go
```diff
@@ -21,303 +20,268 @@ import (
 )
 
 var skip = []string{
-	"nodeModulesAllowJsConditionalPackageExports(module=node16).ts",
```
andrewbranch, line 25: These were in a random order; I sorted them so we can see a real diff of what starts passing in the future.

internal/compiler/module/resolver_test.go
```diff
@@ -549,19 +524,13 @@ func TestModuleResolver(t *testing.T) {
 			t.Fatal(err)
 		}
 		if json.Files != nil {
-			if currentTestCase.name != "" {
-				t.Run(currentTestCase.name, func(t *testing.T) {
-					t.Parallel()
-					if slices.Contains(skip, currentTestCase.name) {
-						t.Skip("Test skipped")
```
andrewbranch, line 512: `t.Skip` on hundreds of tests was making the `hereby test` output unpleasant for everyone else. (Nobody saw this, because a separate test code bug fixed here prevented all these tests from doing anything 😅)

internal/compiler/module/util.go
```diff
@@ -31,3 +34,46 @@ func ParsePackageName(moduleName string) (packageName, rest string) {
...
+			panic("cannot get effective type roots without a config file path or current directory")
+		}
+	}
+
+	typeRoots := make([]string, 0, strings.Count(baseDir, "/"))
+	tspath.ForEachAncestorDirectory(baseDir, func(dir string) (any, bool) {
+		typeRoots = append(typeRoots, tspath.CombinePaths(dir, "node_modules", "@types"))
+		return nil, false
+	})
+	return typeRoots, false
```
andrewbranch: The dependency on tspath means this can’t be a method on CompilerOptions the way packages are organized now, which is a shame

internal/compiler/module/util.go
```diff
@@ -31,3 +34,46 @@ func ParsePackageName(moduleName string) (packageName, rest string) {
 	}
 	return moduleName[:idx], moduleName[idx+1:]
 }
+
+func GetEffectiveTypeRoots(options *core.CompilerOptions, currentDirectory string) (result []string, fromConfig bool) {
```
andrewbranch: `typeRoots` actually have completely different resolution behavior depending on whether they’re defined in config or inferred from the config directory location 😕

internal/compiler/module/util.go
```diff
@@ -3,9 +3,12 @@ package module
 import (
 	"strings"
 
+	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
+var InferredTypesContainingFile = "__inferred type names__.ts"
```
jakebailey:
```suggestion
const InferredTypesContainingFile = "__inferred type names__.ts"
```

(It's never modified)

internal/compiler/module/util.go
jakebailey: `tspath` depends on `core`? Seems wrong...

internal/compiler/module/resolver.go
```diff
@@ -194,6 +249,89 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
...
+
+func (r *resolutionState) resolveTypeReferenceDirective(typeRoots []string, fromConfig bool, fromInferredTypesContainingFile bool) *ResolvedTypeReferenceDirectiveWithFailedLookupLocations {
+	// Primary lookup
+	if len(typeRoots) > 0 {
+		if r.resolver.traceEnabled() {
+			r.resolver.host.Trace(diagnostics.Resolving_with_primary_search_path_0.Format(strings.Join(typeRoots, ", ")))
+		}
+		for _, typeRoot := range typeRoots {
+			candidate := r.getCandidateFromTypeRoot(typeRoot)
+			directoryExists := r.resolver.host.DirectoryExists(candidate)
```
jakebailey:
```suggestion
			directoryExists := r.resolver.host.FS().DirectoryExists(candidate)
```

internal/compiler/module/util.go
andrewbranch: - `core.MakeRegexp` (completely unnecessary)
- `core.EquateStringCaseInsensitive`
- `core.GetStringEqualityComparer`
- `core.GetStringComparer`

jakebailey: I expect `MakeRegexp` to go away if https://github.com/microsoft/typescript-go/pull/65 is uncontroversial, but it does seem like this string stuff could just go over to `stringsx` or `stringutil` or something tbh. But, not a blocker at all

andrewbranch: I would love to keep the compiler options methods with `CompilerOptions`, so I’ll make a `stringutil` package.

internal/ast/diagnostic.go
```diff
@@ -2,6 +2,7 @@ package ast
 
 import (
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+	"github.com/microsoft/typescript-go/internal/compiler/stringutil"
```
jakebailey: I think this is the wrong dir? Shouldn't it have been in internal?

jakebailey: Oh. It moved in later commit, got it

andrewbranch: There was already a `compiler/stringutil`, which depended on `compiler/core` 😠 I had to untangle and combine.

PR #99, opened 2024-11-22
Forbid OptionalType outside tuples

sandersn:
This uses up the last NodeFlags!
Very likely we should break out NodeFlags into a ParserNodeFlags that is only used during parsing, but not saved on the node. But I'm not sure, so I wanted to open this PR to start discussion.

Follow up to #96

### DIFF
--- a/internal/ast/nodeflags.go
+++ b/internal/ast/nodeflags.go
@@ -21,11 +21,12 @@ const (
 	NodeFlagsYieldContext                    NodeFlags = 1 << 14 // If node was parsed in the 'yield' context created when parsing a generator
 	NodeFlagsDecoratorContext                NodeFlags = 1 << 15 // If node was parsed as part of a decorator
 	NodeFlagsAwaitContext                    NodeFlags = 1 << 16 // If node was parsed in the 'await' context created when parsing an async function
-	NodeFlagsDisallowConditionalTypesContext NodeFlags = 1 << 17 // If node was parsed in a context where conditional types are not allowed
-	NodeFlagsThisNodeHasError                NodeFlags = 1 << 18 // If the parser encountered an error when parsing the code that created this node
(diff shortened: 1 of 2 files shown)


internal/compiler/parser.go
```diff
@@ -3151,7 +3152,7 @@ func (p *Parser) parseTupleElementType() *ast.TypeNode {
 		p.finishNode(result, pos)
 		return result
 	}
-	return p.parseType()
+	return doInContext(p, ast.NodeFlagsTupleContext, true, (*Parser).parseType)
```
jakebailey, line 3155: When do we _leave_ the tuple context? What happens if I nest something like `[foo: { prop?: number }]`?

rbuckton:
Do we need a NodeFlags entry? This is what I did in the printer PR: https://github.com/microsoft/typescript-go/pull/97/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-1afa58ec565e11d5c7ab2d09a239aafbdeeeb0923b3d0de8cbea33c06c137247R3155-R3161

internal/ast/nodeflags.go
```diff
@@ -21,11 +21,12 @@ const (
 	NodeFlagsYieldContext                    NodeFlags = 1 << 14 // If node was parsed in the 'yield' context created when parsing a generator
 	NodeFlagsDecoratorContext                NodeFlags = 1 << 15 // If node was parsed as part of a decorator
 	NodeFlagsAwaitContext                    NodeFlags = 1 << 16 // If node was parsed in the 'await' context created when parsing an async function
-	NodeFlagsDisallowConditionalTypesContext NodeFlags = 1 << 17 // If node was parsed in a context where conditional types are not allowed
```
DanielRosenwasser, line 24: Can we just switch these all to use iota?

jakebailey: I'm not sure we're going to merge this PR when #102 also works without a flag, fwiw

PR #101, opened 2024-11-23
Run benchmarks in CI

jakebailey:
This adds a simple workflow which benchmarks a change and its parent, displaying / uploading the `benchstat` results. These benchmarks run on a larger runner.

We'll see how noisy this is given it's on a VM (and we aren't using dedicated hardware for this), but Go benchmarks are a lot less noisy especially given how many runs we can get. (Compared to fully-fledged benchmarks of `tsc` in JS).


jakebailey:
The results here show that the workers are not isolated enough from VM noise to be usable, though I will have to try again using dedicated VMs, or see if we can just make VMs large enough to not actually be used by multiple people.

PR #103, opened 2024-11-25
Port most of CFA and add more accessors

ahejlsberg:
This PR

* ports most of the control flow analyzer,
* refactors `FlowNode` to always use `*Node` for additional data, and
* adds `Type()`, `Initializer()`, `TypeArguments()` and `TypeParameters()` accessors to `Node`.

Note that fields previously named `TypeNode` and `ReturnType` are now just `Type`. This aligns with the original codebase.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -84,6 +84,7 @@ const (
 	CachedTypeKindEquivalentBaseType
 	CachedTypeKindApparentType
 	CachedTypeKindAwaitedType
+	CachedTypeKindEvolvingArrayType
 )
 
 // CachedTypeKey
@@ -123,6 +124,13 @@ type CachedSignatureKey struct {
 type StringMappingKey struct {
 	s *ast.Symbol
 	t *Type
+}
+
+// AssignmentReducedKey
+
+type AssignmentReducedKey struct {
+	id1 TypeId
+	id2 TypeId
 }
 
 // InferenceContext
@@ -328,6 +336,7 @@ type Checker struct {
 	cachedTypes                        map[CachedTypeKey]*Type
 	cachedSignatures                   map[CachedSignatureKey]*Signature
 	narrowedTypes                      map[NarrowedTypeKey]*Type
+	assignmentReducedTypes             map[AssignmentReducedKey]*Type
 	markerTypes                        core.Set[*Type]
 	identifierSymbols                  map[*ast.Node]*ast.Symbol
 	undefinedSymbol                    *ast.Symbol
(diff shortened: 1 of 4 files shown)


internal/ast/flow.go
```diff
@@ -41,17 +41,40 @@ type FlowLabel = FlowNode
 var UnreachableFlow = &FlowNode{Flags: FlowFlagsUnreachable}
 var ReportedUnreachableFlow = &FlowNode{Flags: FlowFlagsUnreachable}
 
-// FlowSwitchClauseData
+// FlowSwitchClauseData (synthetic AST node for FlowFlagsSwitchClause)
 
 type FlowSwitchClauseData struct {
+	NodeBase
```
jakebailey, line 47: I feel at least a little awkward with these being "nodes", in that they don't have their own syntax kinds or anything... Does it have to be a Node?

internal/ast/flow.go
```diff
@@ -41,17 +41,40 @@ type FlowLabel = FlowNode
...
 	SwitchStatement *SwitchStatement
 	ClauseStart     int32 // Start index of case/default clause range
 	ClauseEnd       int32 // End index of case/default clause range
 }
 
-// FlowReduceLabelData
+func NewFlowSwitchClauseData(switchStatement *SwitchStatement, clauseStart int, clauseEnd int) *Node {
+	node := &FlowSwitchClauseData{}
+	node.data = node
```
jakebailey, line 55: Should we just use `newNode` here with `KindUnknown` for consistency? (And in `FlowReduceLabelData`)

internal/ast/flow.go
ahejlsberg: It allows `FlowNode` to store all additional data as a `*Node` instead of an `any`. This reduces the size of `FlowNode` by 20% and saves unnecessary wrapping and unwrapping of nodes. In 99% of cases the payload is just a regular AST node. The `FlowSwitchClause` and `FlowReduceLabel` flow node kinds are quite rare.

internal/ast/flow.go
ahejlsberg: That would require threading through a `NodeFactory` parameter and there's really no need to.

internal/ast/flow.go
DanielRosenwasser: The reason is that it just piggy-backs on `data` for polymorphism, rather than creating a separate wrapper type that adds its own header, or doubling the size of every reference.

internal/ast/flow.go
jakebailey: Ah, that doesn't even need to be a method on the factory, now that I look. I forgot it was one...

PR #104, opened 2024-11-25
Normalize paths, trim trailing slashes in VFS

jakebailey:
The underlying `io/fs.FS` does not expect these to be passed through.

### DIFF
--- a/internal/vfs/vfs_test.go
+++ b/internal/vfs/vfs_test.go
@@ -67,6 +67,7 @@ func TestIOFS(t *testing.T) {
 		assert.Assert(t, fs.DirectoryExists("/"))
 		assert.Assert(t, fs.DirectoryExists("/dir1"))
 		assert.Assert(t, fs.DirectoryExists("/dir1/"))
+		assert.Assert(t, fs.DirectoryExists("/dir1/./"))
 		assert.Assert(t, !fs.DirectoryExists("/bar"))
 	})
 
--- a/internal/vfs/vfs.go
+++ b/internal/vfs/vfs.go
@@ -161,7 +161,8 @@ func rootLength(p string) int {
 	return l
 }
 
-func splitRoot(p string) (rootName, rest string) {
+func splitPath(p string) (rootName, rest string) {
+	p = tspath.NormalizePath(p)
 	l := rootLength(p)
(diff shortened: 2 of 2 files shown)


internal/vfs/vfs_test.go
```diff
@@ -66,6 +66,7 @@ func TestIOFS(t *testing.T) {
 
 		assert.Assert(t, fs.DirectoryExists("/"))
 		assert.Assert(t, fs.DirectoryExists("/dir1"))
+		assert.Assert(t, fs.DirectoryExists("/dir1/"))
 		assert.Assert(t, !fs.DirectoryExists("/bar"))
```
DanielRosenwasser, line 71: Do these all have to be absolute normalized paths? In other words, something like `/dir1/./` won't return true?

Is it worth testing that behavior of the FS?

jakebailey: The FS does not at all clean paths. It's possible that a better solution is to instead document the invariant that all paths must be normalized, clean paths and not merge this PR.

Otherwise, the FS will be cleaning every single path repeatedly which I think feels wrong...

jakebailey: But yes, it's intended that all paths are absolute normalized paths.

jakebailey: Curious what people think the FS should do here. I'm happy to implement whatever, I'm not entirely certain myself.

andrewbranch: My concern is that Node’s `fs` works with non-normalized paths, so we might be inadvertently sending non-normalized paths to FS methods in the TS codebase, and miss normalizing them during the port. This is the kind of thing that can easily escape test coverage. I wonder how much of a perf hit we’re talking about for our FS to clean all incoming paths?

DanielRosenwasser: An alternative is to assert at the FS level when we pass in an uncleaned path. In some cases we could optimistically try to complete certain operations, and when they fail we could do some sort of validation check on debug builds. But that feels less like a port.

jakebailey: Honestly, it's probably fine for me to call the func to renormalize the paths and we can perf bench it later. Not really worth the pain right now.

PR #107, opened 2024-11-25
Concurrent program with module resolution

EricCornelson:
This is mostly the previous PR for lighting up references/imports, but with concurrent parsing, and leaving the current behavior for parsing all files in the directory passed to tsgo (with the one change that if rootPath is a file, we will just parse that file)

Summary of changes:

- Adds parser/scanner types and logic for parsing tripleslash references and adding them to SourceFile
- recursively parsing imports/references for root files
- wiring up module resolution for imports
- add listfiles option

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -160,7 +160,7 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
 			filesToParse = append(filesToParse, resolvedPath)
 		}
 
-		filesToParse = append(filesToParse, p.getImportsToParse(file)...)
+		filesToParse = append(filesToParse, p.resolveImportsAndModuleAugmentations(file)...)
 
 		p.fileProcessingMutex.Lock()
 		defer p.fileProcessingMutex.Unlock()
@@ -229,7 +229,7 @@ func (p *Program) resolveModuleNames(entries []*ast.Node, file *ast.SourceFile) 
 	return resolvedModules
 }
 
-func (p *Program) getImportsToParse(file *ast.SourceFile) []string {
+func (p *Program) resolveImportsAndModuleAugmentations(file *ast.SourceFile) []string {
 	toParse := make([]string, 0, len(file.Imports))
 	if len(file.Imports) > 0 || len(file.ModuleAugmentations) > 0 {
 		moduleNames := getModuleNames(file)


internal/compiler/program.go
```diff
@@ -421,3 +648,9 @@ func (p *Program) collectModuleReferences(file *ast.SourceFile, node *ast.Statem
 		}
 	}
 }
+
+func (p *Program) resolveTripleslashReference(moduleName string, containingFile string) string {
```
andrewbranch: I just merged #98 so you can use the real `resolver.ResolveTypeReferenceDirective` now!

internal/compiler/host.go
```diff
@@ -62,3 +66,29 @@ func (h *compilerHost) WaitForTasks() {
...
+	return d.FS().Realpath(path)
+}
+
+func (d *compilerHost) GetDirectories(path string) []string {
+	return d.FS().GetDirectories(path)
+}
+
+func (d *compilerHost) UseCaseSensitiveFileNames() bool {
+	return d.FS().UseCaseSensitiveFileNames()
+}
```
jakebailey: I think most of these are redundant given the resolver just takes an `FS()` method?

internal/compiler/module/resolver.go
```diff
@@ -164,7 +164,7 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
 
 		switch moduleResolution {
 		case core.ModuleResolutionKindNode16, core.ModuleResolutionKindNodeNext, core.ModuleResolutionKindBundler:
-			state := newResolutionState(moduleName, containingFile, resolutionMode, compilerOptions, r)
+			state := newResolutionState(moduleName, containingDirectory, resolutionMode, compilerOptions, r)
```
jakebailey: This seems spurious but maybe it does fix something.

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
 	}
 	return diagnostics
 }
+
+func getCommentPragmas(sourceText string) (pragmaMap map[string][]ast.Pragma) {
+	for commentRange := range scanner.GetLeadingCommentRanges(sourceText, 0) {
+		comment := sourceText[commentRange.Pos():commentRange.End()]
+		newPragmas, ok := extractPragmas(commentRange, comment)
+		if ok {
+			pragmaMap = make(map[string][]ast.Pragma)
```
jakebailey: I have a suspicion that this is order dependent and may need to be `collections.OrderedMap`, especially given it is looped over below.

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
...
+		return nil, false
+	}
+}
+
+type NamedArgRegEx struct {
+	regex *regexp.Regexp
+	once  sync.Once
+}
+
+var namedArgRegExCache sync.Map
```
jakebailey: The lifetime of these parsed pragmas will be infinite; I was worried about it but it turns out we already let these live forever on TS, wow.

Really, they should be in some sort of TTL cache but in JS that's sort of not a thing (in Go it totally can be).

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
...
+	namedArgRegex.once.Do(func() {
+		namedArgRegex.regex = regexp.MustCompile(fmt.Sprintf(`(?im)(\s%s\s*=\s*)(?:(?:'([^']*)')|(?:"([^"]*)"))`, name))
+	})
+	return namedArgRegex.regex
+}
+
+var tripleSlashXMLCommentStartRegEx = regexp.MustCompile(`(?m)^\/\/\/\s*<(\S+)\s.*?\/>`)
+var singleLinePragmaRegEx = regexp.MustCompile(`(?m)^\/\/\/?\s*@([^\s:]+)((?:[^\S\r\n]|:).*)?$`)
+
+func extractPragmas(commentRange ast.CommentRange, text string) ([]ast.Pragma, bool) {
```
jakebailey: This function is mega nested, even moreso than the original TS code (which was... medium nested?); can we have some early returns here?

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
...
+	// if (range.kind === SyntaxKind.MultiLineCommentTrivia) {
+	//     const multiLinePragmaRegEx = /@(\S+)(\s+(?:\S.*)?)?$/gm; // Defined inline since it uses the "g" flag, which keeps a persistent index (for iterating)
+	//     let multiLineMatch: RegExpExecArray | null; // eslint-disable-line no-restricted-syntax
+	//     while (multiLineMatch = multiLinePragmaRegEx.exec(text)) {
+	//         addPragmaForMatch(pragmas, range, PragmaKindFlags.MultiLine, multiLineMatch);
+	//     }
+	// }
+}
+
+func processPragmasIntoFields(context *ast.SourceFile /* reportDiagnostic func(*ast.Diagnostic)*/) {
```
jakebailey: Are you planning on plumbing the callback into this PR or is that a TODO that we should note?

(The places where we haven't finished a port are all marked with `// !!!`)

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
...
+				noDefaultLib, noDefaultLibOk := pragma.Args["no-default-lib"]
+
+				if noDefaultLibOk && noDefaultLib.Value == "true" {
+					context.HasNoDefaultLib = true
+				} else if typesOk {
+					var parsed core.ResolutionMode
+					if resolutionModeOk {
+						parsed = parseResolutionMode(resolutionMode.Value, types.Pos(), types.End() /*, reportDiagnostic*/)
+					}
+					context.TypeReferenceDirectives = append(context.TypeReferenceDirectives, ast.FileReference{
```
jakebailey: You may want to make all of these `FileReference` slices be `[]*FileReference`; as it stands now, this will incur a bunch of struct copying that is probably not needed.

internal/compiler/parser.go
```diff
@@ -5952,3 +5960,191 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
...
+	//context.CheckJsDirective = nil
+	context.ReferencedFiles = nil
+	context.TypeReferenceDirectives = nil
+	context.LibReferenceDirectives = nil
+	//context.AmdDependencies = nil
+	context.HasNoDefaultLib = false
+	for key, pragmas := range context.Pragmas {
+		switch key {
+		case "reference":
+			for _, pragma := range pragmas {
```
jakebailey: Note the loops here append to `context.<something>` but the loop order of the maps is going to be randomized.

internal/compiler/host.go
```diff
@@ -62,3 +66,29 @@ func (h *compilerHost) WaitForTasks() {
...
 }
+
+// implement ModuleResolutionHost for compilerHost
+
+func (d *compilerHost) FileExists(fileName string) bool {
+	return d.FS().FileExists(fileName)
+}
+
+func (d *compilerHost) Trace(msg string) {
+	fmt.Println(msg)
```
jakebailey: Presumably this should instead hold onto an `io.Writer` or something instead of actually printing out everything? I think it's gated by a bool elsewhere, though...

internal/compiler/program.go
```diff
@@ -52,7 +83,9 @@ func NewProgram(options ProgramOptions) *Program {
 	slices.SortFunc(fileInfos, func(a FileInfo, b FileInfo) int {
 		return int(b.Size) - int(a.Size)
 	})
-	p.parseSourceFiles(fileInfos)
+
+	p.processRootFilesAsync(fileInfos)
```
jakebailey: Everything in Go is async; unlike C# there's no reason to name functions with an `Async` suffix.

internal/compiler/program.go
```diff
@@ -114,23 +147,217 @@ func (p *Program) bindSourceFiles() {
...
+
+	for _, fileInfo := range rootFiles {
+		absPath := tspath.GetNormalizedAbsolutePath(fileInfo.Name, p.currentDirectory)
+		p.processedFileNames.Add(absPath)
+		p.startParseTask(absPath, resultsChannel)
+	}
+
+	for {
+		select {
+		case result := <-resultsChannel:
```
jakebailey: While the channel thing is fun, I think the code would acutally be cleaner if we instead stuck a mutex and locked it to put things onto `files` and `filesByPath`; I don't want to keep the task system around for much longer and I think this processing doesn't actually need to use any abstraction besides a waitgroup. Or at most a singleflight instance, but even that is probably overkill.

internal/compiler/program.go
```diff
@@ -114,23 +147,217 @@ func (p *Program) bindSourceFiles() {
...
+		p.filesByPath[path] = file
+		p.processReferencedFiles(file)
+		p.processImportedModules(file)
+	}
+	return file
+}
+
+func (p *Program) parseSourceFile(fileName string) *ast.SourceFile {
+	path := tspath.ToPath(fileName, p.currentDirectory, p.host.FS().UseCaseSensitiveFileNames())
+	text, _ := p.host.FS().ReadFile(fileName)
```
jakebailey, line 196: Suspicious `ok` ignore here; did the original code in TS ignore `undefined`?

internal/compiler/types.go
```diff
@@ -1121,3 +1121,28 @@ var LanguageFeatureMinimumTarget = LanguageFeatureMinimumTargetMap{
...
+const (
+	RootFile = iota
+	SourceFromProjectReference
+	OutputFromProjectReference
+	Import
+	ReferenceFile
+	TypeReferenceDirective
+	LibFile
+	LibReferenceDirective
+	AutomaticTypeDirectiveFile
```
jakebailey: These are missing a prefix and type annotation:

```suggestion
	FileIncludeKindRootFile FileIncludeKind = iota
	FileIncludeKindSourceFromProjectReference
	FileIncludeKindOutputFromProjectReference
	FileIncludeKindImport
	FileIncludeKindReferenceFile
	FileIncludeKindTypeReferenceDirective
	FileIncludeKindLibFile
	FileIncludeKindLibReferenceDirective
	FileIncludeKindAutomaticTypeDirectiveFile
```

internal/scanner/scanner.go
```diff
@@ -1874,3 +1878,120 @@ func ComputePositionOfLineAndCharacter(lineStarts []core.TextPos, line int, char
...
+				if hasPendingCommentRange {
+					pendingHasTrailingNewLine = true
+				}
+
+				continue
+			case '\t':
+			case '\v':
+			case '\f':
+			case ' ':
+				pos++
```
jakebailey: Empty cases is likely not what you meant to do here; I assume these all should do `pos++` as they are all whitespace:

```suggestion
			case '\t', '\v', '\f', ' ':
				pos++
```

I really need to write that lint rule.

internal/scanner/scanner.go
```diff
@@ -1874,3 +1878,120 @@ func ComputePositionOfLineAndCharacter(lineStarts []core.TextPos, line int, char
...
+		var collecting = trailing
+		if pos == 0 {
+			collecting = true
+			if isShebangTrivia(text, pos) {
+				pos = scanShebangTrivia(text, pos)
+			}
+		}
+	scan:
+		for pos >= 0 && pos < len(text) {
+			ch := text[pos]
```
jakebailey: This is iterating over the string byte by byte; I suspect this is subtly wrong for UTF-8 characters and should be decoding characters instead (`utf8.DecodeRuneInString(text[pos:])`), or at least only doing `pos++` when the width is known.

The scanner does have some examples of this.

internal/scanner/scanner.go
```diff
@@ -1874,3 +1878,120 @@ func ComputePositionOfLineAndCharacter(lineStarts []core.TextPos, line int, char
...
+						pendingKind = kind
+						pendingHasTrailingNewLine = hasTrailingNewLine
+						hasPendingCommentRange = true
+					}
+
+					continue
+				}
+				break scan
+			default:
+				if ch > maxAsciiCharacter && (stringutil.IsWhiteSpaceLike(rune(ch))) {
```
jakebailey: Yeah, I don't think this is right given this will increment and then we'll have broken utf8 encoding. (Note that a lot of the helpers work on runes instead of bytes; if you're converting anything, consider that a warning)

internal/compiler/program.go
```diff
@@ -114,23 +147,217 @@ func (p *Program) bindSourceFiles() {
...
 	directory := tspath.GetDirectoryPath(currentSourceFile.FileName())
 	if tspath.IsExternalModuleNameRelative(moduleReference) {
-		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference))
+		return p.findSourceFileSync(tspath.CombinePaths(directory, moduleReference), FileIncludeReason{Import, 0})
 	}
 	return p.findNodeModule(moduleReference)
 }
 
-func (p *Program) findSourceFile(candidate string) *ast.SourceFile {
+func (p *Program) findSourceFileSync(candidate string, reason FileIncludeReason) *ast.SourceFile {
```
jakebailey: Ah, I sort of see why you named these methods this way now. Probably the real split is that one is a "worker"?

internal/compiler/program.go
```diff
@@ -114,23 +147,217 @@ func (p *Program) bindSourceFiles() {
...
+			if len(filesToParse) > 0 {
+				for _, fileName := range filesToParse {
+					p.processedFileNames.Add(fileName)
+					p.startParseTask(fileName, resultsChannel)
+				}
+			}
+
+		default:
+			// resultChannel is blocking, wait for all tasks to finish, and see if there are any more results
+			p.host.WaitForTasks()
```
jakebailey: This bit scares me; because this work system is on the host, there's nothing saying that other, unrelated tasks couldn't end up on it. I think we should not be using this, but instead consider writing a plain loop or something.

internal/compiler/program.go
ahejlsberg: I agree with Jake. Much cleaner to just have a mutex and a waitgroup.

internal/compiler/parser.go
EricCornelson: This isn't storing the pragmas, it's just storing the dynamically generated regexes for parsing the pragmas

jakebailey: Yeah, I meant "parsed regexes"; they're still user controlled input that will never be cleaned up. Not a regression, though, but something we can actually fix now that we can have something concurrently wiping this every now and then.

internal/compiler/program.go
EricCornelson: yeah, and original would just catch any errors and set the text to ""
https://github.com/microsoft/TypeScript/blob/52c59dbcbee274e523ef39e6c8be1bd5e110c2f1/src/compiler/program.ts#L410

jakebailey: Ah, but there's an `onError`. Wonder what that calls, and if that means we should make our FS interface actually return some sort of error message.

andrewbranch: It looks like `onError` only exists for public API. I can’t find a place where we do anything but pass it through in our own code.

internal/compiler/program.go
EricCornelson: I'm not sure I agree that a mutex would be cleaner. `files` and `filesByPath` aren't the only state that gets mutated during parsing, there's everything in the module resolver, and its caching layer, a bunch of other little caches and things like `mapFromToProjectReferenceRedirectSource`, `redirectTargetsMap`, `projectReferenceRedirects`, `packageIdToSourceFile`, `sourceFileToPackageName`, most of which aren't implemented yet, but accesses on them are peppered throughout the call chain of `processSourceFile` in various spots, and I have no idea how hard it might be to rewrite the order of these things to make it easier to lock over without changing the behavior of the program.

So it's either we lock over a huge section of code that does a ton of stuff, or we have a ton of little locks everywhere that maybe could deadlock?

jakebailey: I would think we could calculate the changes to make, then lock once over the state to update things. Or just have lots of locks. I don't really know how the channel helps, though...

internal/compiler/parser.go
EricCornelson: I don't think they're user controlled. They seem to be parsing over well-known field names for the pragmas, just doing it very dynamically. I kind of followed the way original ts did it, but it might be simpler just to write a little parser function rather than use regexes (they're basically just looking for a pattern like `someName = "someValue"` with whitespace in between).

jakebailey: Ah, ok, then nevermind.

internal/compiler/module/resolver.go
EricCornelson: yeah, this is intentional

internal/compiler/parser.go
EricCornelson: added a `// !!!` comment

internal/compiler/parser.go
EricCornelson: done!

internal/compiler/types.go
EricCornelson: fixed!

internal/scanner/scanner.go
EricCornelson: fixed!

internal/scanner/scanner.go
EricCornelson: Ok, I think this is good now. Take another look and see if I'm missing anything

internal/compiler/stringutil/util.go
```diff
@@ -112,3 +112,10 @@ func ComputeLineStarts(text string) []core.TextPos {
 	result = append(result, core.TextPos(lineStart))
 	return result
 }
+
+func RuneAt(text string, pos int) (rune, int) {
+	if pos < 0 || pos >= len(text) {
+		return -1, 0
+	}
```
jakebailey: Panicking may be a safer alternative here, given its use doesn't seem to check the output. Or, it'll panic anyway due to being out of bounds of the string...

I'll also note that the scanner already has this as a helper, at least in some form, so there may be some code reuse there which may help.

internal/scanner/scanner.go
jakebailey: Seems better now, thanks.

internal/compiler/program.go
EricCornelson: well, partly that, and partly we do things differently in the async vs. sync methods. Eventually I plan to merge these, but right now I'm not sure of the best way to do that, so I'm leaving the "sync" methods there for the places that use them currently

internal/compiler/program.go
EricCornelson: using a separate work group now

internal/compiler/stringutil/util.go
EricCornelson: I just removed this function since it didn't really do much other than call `utf8.DecodeRuneInString` and the pos check is already done inside `iterateCommentRanges` anyway

internal/compiler/program.go
EricCornelson: changed to remove channel stuff and just use mutex and waitgroup

internal/compiler/parser.go
EricCornelson: added some early returns so it should be flatter now

internal/compiler/parser.go
EricCornelson: ok, for now I just got rid of the map, since the pragmas are only ever accessed via the map in a few spots for jsx pragmas. We can assess whether an orderedmap is worth it later

internal/compiler/host.go
EricCornelson: Just removing the println for now and adding a todo comment, since I'm not sure how tracing should work yet

internal/compiler/program.go
```diff
@@ -114,23 +149,213 @@ func (p *Program) bindSourceFiles() {
...
+	go func() {
+		defer wg.Done()
+		normalizedPath := tspath.NormalizePath(fileName)
+		file := p.parseSourceFile(normalizedPath)
+		p.collectExternalModuleReferences(file)
+
+		p.mutex.Lock()
+		defer p.mutex.Unlock()
+		p.files = append(p.files, file)
+		p.filesByPath[file.Path()] = file
```
jakebailey: This lock is held for the rest of the function; this means binding is no longer parallel, nor is resolving references or anything.

I think the lock is only here to serialize access to files/filesByPath, along with `processedFileNames`.

Can we put the two `p.files`/`p.filesByPath` into a helper which locks the mutex, and then if we need, lock later just for `processedFileNames`?

internal/compiler/program.go
```diff
@@ -114,23 +149,213 @@ func (p *Program) bindSourceFiles() {
...
 	extensionless := tspath.RemoveFileExtension(candidate)
 	for _, ext := range []string{tspath.ExtensionTs, tspath.ExtensionTsx, tspath.ExtensionDts} {
 		path := tspath.ToPath(extensionless+ext, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
 		if result, ok := p.filesByPath[path]; ok {
 			return result
 		}
 	}
-	return nil
+
+	file := p.parseSourceFile(candidate)
```
jakebailey: I am quite confused by all of the different functions here, process, find, etc, which all seem to contain parsing.

Shouldn't we have already parsed all referenced files by the end of `processRootFiles` such that nothing other than the above function should have parsed anything?

internal/compiler/program.go
jakebailey: I'm still not a huge fan of it; I'm finding it difficult to understand where things are used and what the naming implies.

The main naming convention I've noticed in gopls is to have a `Locked` name suffix that implies that all locks are held before the function is called.

internal/compiler/program.go
EricCornelson: Yeah, I'm not sure. The way it's currently written implies that we might have to parse something after the initial pass is done, e.g. `getResolvedModule` is called from the checker in `resolveExternalModule`, and that calls `findSourceFile`, which, in the current TypeScript, will check if the file was already parsed, and if not, parse it.

Can we actually guarantee that we wont have to parse anything after the initial pass?

andrewbranch: `getResolvedModule` only ever returns already parsed files; i.e. checking never triggers additional parsing. `getResolvedModule` in current TS is this:

```ts
function getResolvedModule(file: SourceFile, moduleName: string, mode: ResolutionMode) {
    return resolvedModules?.get(file.path)?.get(moduleName, mode);
}
```

internal/compiler/program.go
EricCornelson: again, though, those aren't the only two things that we touch once we start processing imported files. I think I can move `file.ReferencedFiles` out of the mutex since those don't touch any state, but doing the imports involves the module resolver and touches some program state.

I guess I'm trying to strike a balance here between maintaining the current order of processing and stuff from the old code, while also parallelizing it. I can totally just rewrite stuff to make it easier to parallelize, but I just don't have enough context to know if that's going to just really break something.

internal/compiler/program.go
andrewbranch: Also, `host.getSourceFile` in the checker calls this Program method:

```ts
function getSourceFile(fileName: string): SourceFile | undefined {
    return getSourceFileByPath(toPath(fileName));
}

function getSourceFileByPath(path: Path): SourceFile | undefined {
    return filesByName.get(path) || undefined;
}
```

EricCornelson: I'm more worried about what is coming from `resolveExternalModule`. Like, what does that mean? What's an external module? Is it something we should have already seen, or something that might be a new file to parse?

internal/compiler/program.go
jakebailey: I am not sure there is a need to strike any balances; the thing we had talked about was letting parsing go wild with whatever it wants to do, then doing a traversal of all files to determine the final file order after that all is completed. As long as we have the initial list of files still in order, we should be able to reorder the rest of the files that way.

jakebailey: I'll note that right now, though, it's not really a big deal, because we don't even read a tsconfig and have a list of files to start with, only a temporary recursive FS scan.

internal/compiler/program.go
andrewbranch: “External module” is archaic for “module” (as opposed to an old-school TS namespace, once called an “internal module”). The functions I just pasted are the ones called from `resolveExternalModule`. It is impossible for checking to trigger more parsing. That’s a strong design invariant of the compiler.

EricCornelson: Ok perfect, that's exactly what I needed to know. I can just clean up some of these methods then.

EricCornelson: ok so with that said, if `findSourceFile` doesn't find a source file, what should we do? panic? return nil? parse it?

jakebailey: I assume panic, given that's breaking the invariant.

jakebailey: Er, well, Andrew said it returns undefined above, so that

internal/compiler/program.go
EricCornelson: ok, shrunk the mutex lock section

internal/compiler/program.go
EricCornelson: cleaned up the extraneous functions

internal/compiler/program.go
EricCornelson: got rid of the sync/async

internal/compiler/program.go
```diff
@@ -114,23 +125,154 @@ func (p *Program) bindSourceFiles() {
...
-		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference))
+		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference), FileIncludeReason{FileIncludeKindImport, 0})
 	}
 	return p.findNodeModule(moduleReference)
 }
 
-func (p *Program) findSourceFile(candidate string) *ast.SourceFile {
+func (p *Program) findSourceFile(candidate string, reason FileIncludeReason) *ast.SourceFile {
 	extensionless := tspath.RemoveFileExtension(candidate)
 	for _, ext := range []string{tspath.ExtensionTs, tspath.ExtensionTsx, tspath.ExtensionDts} {
```
andrewbranch: Oh, this should go away

EricCornelson: @ahejlsberg, are you relying on this behavior at all for your testing?

jakebailey: This is something that is explicitly replaced by real module resolution, so I assume not

internal/compiler/program.go
```diff
@@ -114,25 +125,152 @@ func (p *Program) bindSourceFiles() {
...
+
+		p.mutex.Lock()
+		defer p.mutex.Unlock()
+		p.files = append(p.files, file)
+		p.filesByPath[file.Path()] = file
+
+		if len(filesToParse) > 0 {
+			for _, fileName := range filesToParse {
+				p.processedFileNames.Add(fileName)
+				p.startParseTask(fileName, wg)
```
andrewbranch: I think the placement of reads/writes to `processedFileNames` doesn't account for duplication within or between a single file's imports/references.

EricCornelson: good point. Moved the check down below where we kick off the parse tasks.

internal/compiler/program.go
andrewbranch: Reminder that this is still oustanding, should be an easy update

internal/compiler/program.go
```diff
@@ -114,25 +125,146 @@ func (p *Program) bindSourceFiles() {
...
+
+func (p *Program) startParseTask(fileName string, wg *sync.WaitGroup) {
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		normalizedPath := tspath.NormalizePath(fileName)
+		file := p.parseSourceFile(normalizedPath)
+		p.collectExternalModuleReferences(file)
+
+		bindSourceFile(file, p.options)
```
jakebailey: I don't think `bindSourceFile` should be here anymore; that is now done later in a different pass.

internal/compiler/program.go
```diff
@@ -114,25 +125,146 @@ func (p *Program) bindSourceFiles() {
 	p.host.WaitForTasks()
 }
 
+func (p *Program) processRootFiles(rootFiles []FileInfo) {
+	wg := sync.WaitGroup{}
```
jakebailey: One thing I want to keep around is the ability to use `SingleThreaded`; we've been using it to gauge the speedup of parallelism.

The current code doesn't respect it, but I think that this may be a good time to delete the old work queuing system in favor of something else.

What I'm thinking is something like this:

```go
type workGroup struct {
	wg               sync.WaitGroup
	singleThreaded   bool
	singleThreadedMu sync.Mutex
}

func (w *workGroup) run(fn func()) {
	w.wg.Add(1)
	go func() {
		defer w.wg.Done()
		if w.singleThreaded {
			w.singleThreaded.Lock()
			defer w.singleThreaded.Unlock()
		}
		fn()
	}()
}

func (w *workGroup) wait() {
	w.wg.Wait()
}
```

Then, you can create this where you currently create a WaitGroup, and also change binding to do that too.

andrewbranch: Hm, on a semi-related note, did I possibly break this PR by merging #131, since multiple goroutines may perform module resolution at the same time using the same cache?

jakebailey: I don't see you starting any goroutines there, so I assume not? Can you point me to what you mean?

jakebailey: Oh, you don't mean single threaded ness, you mean races. Yeah, you may need to add mutexes to serialize access to those caches.

jakebailey: (this is where some basic tests would be helpful, like one which just tries to load `_submodules/TypeScript/src/compiler` and verifies the list matches an expected set of files)

internal/compiler/program.go
EricCornelson: It is, but I thought in the design meeting a while back Anders said we should do bind right after parsing, since it can also be done in parallel. Maybe in a separate goroutine though, so we can resolve the references and start parsing them as soon as possible also?

jakebailey: See: https://github.com/microsoft/typescript-go/blob/9385c45e2176f06e4d5b79eaf780bd03ffc22e23/internal/compiler/program.go#L106-L115

jakebailey: This has to be done later because we skip binding when there are parse errors, it seems. I'm not sure that's what the old TS does, though. But I think we can very safely _not_ bind right in the same pass. Splitting them apart will mean we can still perf benchmark them like we do `tsc --diagnostics`, which shows the phases.

internal/compiler/program.go
EricCornelson: where do you think is a good spot to put that `workGroup` stuct? Core?

jakebailey: Right now, only `program` needs it, so it would be okay to keep it in this package. But, you could move it to `core` too. It's possible we may need it in a dependency of `core` at some point in which case we can move it again.

internal/compiler/program.go
andrewbranch: Discussed in chat; I was mistaken, this is only for `<reference path`, not `<reference types`, so maybe we should just rename this function something less confusing

internal/compiler/program.go
```diff
@@ -114,25 +125,142 @@ func (p *Program) bindSourceFiles() {
 	p.host.WaitForTasks()
 }
 
+func (p *Program) processRootFiles(rootFiles []FileInfo, singleThreaded bool) {
+	wg := core.NewWorkGroup(singleThreaded)
```
jakebailey:
```suggestion
	wg := core.NewWorkGroup(p.options.SingleThreaded)
```

This should work, rather than using the parameter. This would also be needed for the `bindSourceFiles` func.

EricCornelson: `p.options` is `CompilerOptions`, SingleThreaded is on `ProgramOptions` which we aren't actually storing right now. How do you feel about actually just adding fields (or methods) to access both, and named more clearly, since I've gotten myself confused over which `options` options is referring to a few times.

jakebailey: Oh, you can just put `singleThreaded` on `Program`, in that case, oops.

jakebailey: (whatever works, I just want it attached to the Program and not as a parameter to pass around)

internal/compiler/program.go
EricCornelson: removed

internal/compiler/program.go
EricCornelson: renamed to `resolveTripleslashPathReference`

internal/compiler/program.go
EricCornelson: removed

internal/compiler/program.go
EricCornelson: added in core for now

internal/compiler/program.go
EricCornelson: ok, added the program options to the program, and now we're getting `singleThreaded` from there

internal/compiler/program.go
```diff
@@ -5,44 +5,75 @@ import (
...
+	compilerOptions    *core.CompilerOptions
+	rootPath           string
+	processedFileNames core.Set[string]
+	files              []*ast.SourceFile
+	filesByPath        map[tspath.Path]*ast.SourceFile
+	nodeModules        map[string]*ast.SourceFile
+	checker            *Checker
+	resolver           *module.Resolver
+	currentDirectory   string
+	mutex              sync.Mutex
```
jakebailey: This mutex guards specific things; can we move those props down below the mutex in its own group just to make it clear what this mutex is for?

EricCornelson: done. Also named the mutex something more descriptive

internal/compiler/program.go
```diff
@@ -80,59 +114,157 @@ func readFileInfos(fs vfs.FS, rootPath string, extensions []string) []FileInfo {
...
-		if result, ok := p.filesByPath[path]; ok {
-			return result
-		}
+func (p *Program) findSourceFile(candidate string, reason FileIncludeReason) *ast.SourceFile {
+	path := tspath.ToPath(candidate, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
+	if result, ok := p.filesByPath[path]; ok {
+		return result
 	}
+
 	return nil
```
jakebailey:
```suggestion
	return p.filesByPath[path]
```

internal/compiler/program.go
```diff
@@ -427,6 +559,12 @@ func (p *Program) collectModuleReferences(file *ast.SourceFile, node *ast.Statem
 	}
 }
 
+func (p *Program) resolveTripleslashPathReference(moduleName string, containingFile string) string {
+	basePath := tspath.GetDirectoryPath(containingFile)
+	referencedFileName := core.IfElse(tspath.IsRootedDiskPath(moduleName), moduleName, tspath.CombinePaths(basePath, moduleName))
```
jakebailey:
```suggestion
	referencedFileName := moduleName
	if !tspath.IsRootedDiskPath(moduleName) {
		referencedFileName = tspath.CombinePaths(basePath, moduleName)
	}
```

`core.IfElse` should really only be used if its inputs are cheap, which `CombinePaths. isn't

jakebailey:
I did some more perf testing, and it seems like a lot of the slowdown is actually just additional memory allocations and GCing due to path normalization. During resolution:

image

These both should be optimized; I fixed up `NormalizePaths` for absolute paths, but most paths processed during module resolution are relative, or need to be `ToPath`'d, which is also still using a regex.

jakebailey:
(This is not a blocker, but, something we should immediately follow up on, write a benchmark for program load, etc)

internal/compiler/program.go
EricCornelson: done

internal/compiler/program.go
EricCornelson: and done

jakebailey:
With the latest changes, things are better:

image

I am not sure we can do much better.

jakebailey:
I _suspect_ that's because we aren't finding the ambients because `@types/node` is not being loaded.

internal/compiler/program.go
```diff
@@ -81,57 +115,151 @@ func readFileInfos(fs vfs.FS, rootPath string, extensions []string) []FileInfo {
...
+			continue
+		}
+		resolvedModule := p.resolver.ResolveModuleName(moduleName, file.FileName(), core.ModuleKindNodeNext, nil)
+		resolvedModules = append(resolvedModules, resolvedModule)
+	}
+
+	return resolvedModules
+}
+
+func (p *Program) getImportsToParse(file *ast.SourceFile) []string {
```
andrewbranch: This function name is breaking my brain every time:
- it's not just imports, it's module augmentations too
- we're not going to parse imports; we're going to parse the files the imports resolve to

I think the name should reflect the fact that it’s a small wrapper around module resolution

internal/compiler/program.go
```diff
@@ -81,57 +115,151 @@ func readFileInfos(fs vfs.FS, rootPath string, extensions []string) []FileInfo {
...
+	for _, absPath := range absPaths {
+		p.startParseTask(absPath, wg)
+	}
+
+	wg.Wait()
+}
+
+func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
+	wg.Run(func() {
+		normalizedPath := tspath.NormalizePath(fileName)
```
andrewbranch, line 152: This normalization is redundant in at least _some_ cases, so if we’re trying to optimize path stuff, maybe we could move it out and evaluate whether each caller needs to apply any additional normalization on the argument it passes in.

internal/compiler/program.go
```diff
@@ -81,57 +115,151 @@ func readFileInfos(fs vfs.FS, rootPath string, extensions []string) []FileInfo {
...
+			// - module name comes from the list of imports
+			// - it's not a top level JavaScript module that exceeded the search max
+
+			//const elideImport = isJsFileFromNodeModules && currentNodeModulesDepth > maxNodeModuleJsDepth;
+
+			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
+			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
+
+			shouldAddFile := resolution.IsResolved()
+			// TODO(ercornel): !!!: other checks on whether or not to add the file
```
andrewbranch, line 254: This is why you see JS files being added to the program now; it should depend on `allowJs`. Not sure if that’s something worth adding now or not.

internal/compiler/program.go
EricCornelson: Any suggestions for the name? `getResolvedImportPaths`?

internal/compiler/program.go
jakebailey: I'd say we should, but we don't have a way for someone to control that except to modify the code, so... I think an implicit default of allowJs is sort of okay...

internal/compiler/program.go
jakebailey: It didn't show up in my profiling; I can see it maybe being worse if on Windows due to slashes (but I think that wasn't a top contender), but I can check that

internal/compiler/program.go
andrewbranch: `resolveImportsAndModuleAugmentations`?

EricCornelson: works for me!

PR #108, opened 2024-11-26
Define a CaseSensitivity type instead of using useCaseSensitiveFileNames bool

jakebailey:
This _feels_ less difficult to think about...


internal/tspath/path.go
```diff
@@ -583,16 +596,16 @@ func IsExternalModuleNameRelative(moduleName string) bool {
...
 
 func (o ComparePathsOptions) GetComparer() func(a, b string) int {
-	return stringutil.GetStringComparer(!o.UseCaseSensitiveFileNames)
+	return stringutil.GetStringComparer(!o.CaseSensitivity.IsCaseSensitive())
 }
 
 func (o ComparePathsOptions) getEqualityComparer() func(a, b string) bool {
-	return stringutil.GetStringEqualityComparer(!o.UseCaseSensitiveFileNames)
+	return stringutil.GetStringEqualityComparer(!o.CaseSensitivity.IsCaseSensitive())
 }
```
jakebailey: I did not plumb this further, but maybe case sensitivity should be in `stringutil`.

internal/tspath/path.go
```diff
@@ -10,6 +10,19 @@ import (
...
+//go:generate go run golang.org/x/tools/cmd/stringer -type=CaseSensitivity
+
+type CaseSensitivity uint8
+
+const (
+	CaseSensitive CaseSensitivity = iota
+	CaseInsensitive
+)
+
+func (c CaseSensitivity) IsCaseSensitive() bool {
```
jakebailey: Or, `IsSensitive()`...

internal/tspath/path.go
```diff
@@ -10,6 +10,19 @@ import (
 
 type Path string
 
+//go:generate go run golang.org/x/tools/cmd/stringer -type=CaseSensitivity
+
+type CaseSensitivity uint8
+
+const (
+	CaseSensitive CaseSensitivity = iota
```
jakebailey: This makes the default value "sensitive" which is backwards from before; will flip.

andrewbranch:
This doesn’t really feel necessary to me 🤷 I think `isCaseSensitive` (a better name for the same bool) sounds very understandably like a true/false file system property, personally.

jakebailey:
The reason I'm not a fan is that case insensitivity is the "weird" case, yet it's the one we treat as the default of false. I have a really hard time remembering which sensitivity true or false is just by reading.

andrewbranch:
I don’t think that’s right—Windows and macOS are case-insensitive, making the default of `false` correct for the vast majority of users.

internal/tspath/path_test.go
```diff
@@ -307,7 +307,7 @@ func TestToFileNameLowerCase(t *testing.T) {
 
 func TestToPath(t *testing.T) {
 	t.Parallel()
-	assert.Equal(t, string(ToPath("file.ext", "path/to", false /*useCaseSensitiveFileNames*/)), "path/to/file.ext")
-	assert.Equal(t, string(ToPath("file.ext", "/path/to", true /*useCaseSensitiveFileNames*/)), "/path/to/file.ext")
-	assert.Equal(t, string(ToPath("/path/to/../file.ext", "path/to", true /*useCaseSensitiveFileNames*/)), "/path/file.ext")
+	assert.Equal(t, string(ToPath("file.ext", "path/to", CaseInsensitive)), "path/to/file.ext")
+	assert.Equal(t, string(ToPath("file.ext", "/path/to", CaseSensitive)), "/path/to/file.ext")
```
EricCornelson, line 570: this doesn't actually test the case sensitive case, right? since the path is all lowercase

jakebailey:
> I don’t think that’s right—Windows and macOS are case-insensitive, making the default of `false` correct for the vast majority of users.

It's not about the number of users (obviously Windows/macOS outshadows Linux for dev machines, though not CI), but about how often we need to special case things in our code. If we're doing something with the bool it's almost always a case of `if !useCaseSensitiveFileNames { // do complicated thing }`.

andrewbranch:
How does this PR change that? https://github.com/microsoft/typescript-go/pull/108/files#diff-bc9badf529b541b86d6fda311f6e0b75b5afcd6d1c9c672d7e3a24454c4aca0cR604

jakebailey:
I guess it doesn't, since I didn't take it that far and reverse conditions.

I guess my point isn't super great if you can see from the variable name what it is, but mainly when I was working on the FS stuff, I was having trouble writing tests where I couldn't remember what `true`/`false` meant when passing things in, where a named value was a better argument to the call.

If this seems pointless, I can just drop it.

DanielRosenwasser:
I think the point is not so much just the naming, but more that when you pass the thing in, it's clear what you're passing in. If we have 2 boolean parameters in a row, you'll never have a hazardous refactoring. I personally think it's nice to have the safety of a distinct type.

jakebailey:
I'll reopen, but it's entirely possible that this would be moot as most places which set this flag by hand are already written...

internal/tspath/path_test.go
jakebailey: No, it doesn't seem like these are testing the case part. Probably this should have better tests.

PR #109, opened 2024-11-26
Add a `-pprof` flag for generating pprof profiles.

DanielRosenwasser:
This change adds a flag called `-pprof` which takes a directory to emit paths for CPU and memory profiles.

```
$ ./bin/ts-go -pprof my-profiles ../vscode/src
Files:         4273
Types:         333325
Compile time:  3.067687752s
Memory used:   1206547K
CPU profile: my-profiles/102432-cpuprofile.pb.gz
Memory profile: my-profiles/102432-memprofile.pb.gz
```


cmd/tsgo/main.go
```diff
@@ -86,10 +95,16 @@ func main() {
 		}
 	}
 	compileTime := time.Since(startTime)
+
+	if profileSession != nil {
+		profileSession.stop()
+	}
```
jakebailey: I would think we'd just want to defer this...

cmd/tsgo/main.go
```diff
@@ -116,3 +131,48 @@ func main() {
 	fmt.Printf("Compile time:  %v\n", compileTime)
 	fmt.Printf("Memory used:   %vK\n", memStats.Alloc/1024)
 }
+
+type profileSession struct {
+	cpuFilePath string
+	memFilePath string
+	cpuFile     *os.File
+	memFile     *os.File
```
jakebailey, line 135: I don't think these files are ever `Close`'d.

cmd/tsgo/main.go
```diff
@@ -116,3 +131,48 @@ func main() {
...
+}
+
+func beginProfiling(profileDir string) *profileSession {
+	if err := os.MkdirAll(profileDir, 0755); err != nil {
+		panic(err)
+	}
+
+	cpuProfilePath := tspath.ResolvePath(profileDir, fmt.Sprintf("cpu-%d.prof", os.Getpid()))
+	memProfilePath := tspath.ResolvePath(profileDir, fmt.Sprintf("mem-%d.prof", os.Getpid()))
```
jakebailey: I would probably not use `.prof` as the term, because that to be implies the V8 format, wheras `.out` or `.pb.gz` or something are pprof profiles...

cmd/tsgo/main.go
DanielRosenwasser: The problem is that it will include diagnostic printback time if we're okay with that.

cmd/tsgo/main.go
DanielRosenwasser: Sure, I only used `.prof` because they used it in https://go.dev/blog/pprof

jakebailey: Oh, true. They do have `.prof` in their docs and not `pb.gz`.

cmd/tsgo/main.go
jakebailey: Given it's a profile over a single run that one has to manually look at, i would actually assume we _do_ want to include that time if it's nonzero.

cmd/tsgo/main.go
jakebailey: Confusingly, I see `.pb.gz` in other blog posts, and the pprof tool itself talks about `.pb.gz` https://github.com/google/pprof#generate-a-text-report-of-the-profile-sorted-by-hotness

cmd/tsgo/main.go
```diff
@@ -50,6 +52,7 @@ func main() {
 	flag.BoolVar(&parseAndBindOnly, "p", false, "Parse and bind only")
 	flag.BoolVar(&printTypes, "t", false, "Print types defined in main.ts")
 	flag.BoolVar(&pretty, "pretty", true, "Get prettier errors")
+	flag.StringVar(&profileDir, "pprof", "", "Generate pprof CPU/memory profiles to the given directory")
```
jakebailey: can you call the flag `pprofdir` or similar to indicate that it's a dir name?

cmd/tsgo/main.go
```diff
@@ -71,6 +74,12 @@ func main() {
 
 	programOptions := ts.ProgramOptions{RootPath: normalizedRootPath, Options: compilerOptions, SingleThreaded: singleThreaded, Host: host}
 
+	if pprofDir != "" {
+		pprofDir = tspath.ResolvePath(currentDirectory, pprofDir)
```
jakebailey: This should just be `filepath`; I don't think `ResolvePath` will use the right normalization?

DanielRosenwasser: What do you mean by right normalization? What would differ in a meaningful way?

DanielRosenwasser: Something I specifically take advantage of here is that with `ResolvePath`, an absolute path resets the root of all prior paths.

Also, `rootPath` above also uses `ResolvePath`, so I'm wondering what would be wrong here.

jakebailey: As in, I'm wary of any use of `tspath` on paths that are not already normalized and internal to the compiler; this is a place that should just be `filepath` because we're working in Go's OS package land.

jakebailey: I mean, it's fine, it's just going to give the "wrong" slashes to the pprof package, which in practice is fine

jakebailey: > Something I specifically take advantage of here is that with `ResolvePath`, an absolute path resets the root of all prior paths.

There's no reason to do this, though, if you pass the path through as-is, then the path will automatically be joined with CWD if it's relative on use. The only reason the TS one needs to resolve is that we require absolute, normalized paths at all times within the compiler.

cmd/tsgo/main.go
```diff
@@ -116,3 +127,54 @@ func main() {
...
+
+func beginProfiling(profileDir string) *profileSession {
+	if err := os.MkdirAll(profileDir, 0755); err != nil {
+		panic(err)
+	}
+
+	pid := os.Getpid()
+
+	cpuProfilePath := filepath.Join(profileDir, fmt.Sprintf("cpuprofile-%d.pb.gz", pid))
+	memProfilePath := filepath.Join(profileDir, fmt.Sprintf("memprofile-%d.pb.gz", pid))
```
jakebailey: Did we want to put the PIDs first to make this sort properly?

.gitignore
```diff
@@ -16,6 +16,9 @@ pids
 *.seed
 *.pid.lock
 
+# Profiling files
+*.prof
```
jakebailey, line 20: Can you add `*.pb.gz` here too?

PR #110, opened 2024-11-27
Fix multiple issues in grammar checking

ahejlsberg:
The ported grammar checking code has several issues that causes it to report incorrect errors. On `vscode/src` these incorrect checks more than double the number of reported errors (to over 120K). That's likely the reason for the performance issue we discussed.

This PR fixes some of the issues which brings down the error count and improves performance. There are still grammar errors being reported on `vscode/src` and they're likely incorrect. I would imagine that codebase is free of grammar errors, so probably would make sense to examine each of them.

### DIFF
--- a/internal/compiler/grammarchecks.go
+++ b/internal/compiler/grammarchecks.go
@@ -570,11 +570,6 @@ func isJSDocTypedefTag(_ *ast.Node) bool {
 	return false
 }
 
-/**
- * TSTrue | TSFalse: Early return this value from checkGrammarModifiers.
- * Unknown: Need to do full checking on the modifiers.
- */
-
 func (c *Checker) reportObviousModifierErrors(node *ast.Node) bool {
 	modifier := c.findFirstIllegalModifier(node)
 	if modifier == nil {


internal/compiler/grammarchecks.go
```diff
@@ -200,19 +200,15 @@ func (c *Checker) checkGrammarModuleElementContext(node *ast.Statement, errorMes
 }
 
 func (c *Checker) checkGrammarModifiers(node *ast.Node /*Union[HasModifiers, HasDecorators, HasIllegalModifiers, HasIllegalDecorators]*/) bool {
-	quickResult := c.reportObviousDecoratorErrors(node)
-	if quickResult == core.TSTrue {
-		quickResult = c.reportObviousModifierErrors(node)
+	if node.Modifiers() == nil {
```
DanielRosenwasser, line 203: Since you're fetching `Modifiers()` here, you can technically pass it through to each of the functions which will do the same sort of check. Maybe that's out of scope here though.

ahejlsberg: It's super cheap to fetch them (just a method call), so not worth optimizing for. The important optimization is that we immediately bail out when there are no modifiers.

internal/compiler/grammarchecks.go
```diff
@@ -200,19 +200,15 @@ func (c *Checker) checkGrammarModuleElementContext(node *ast.Statement, errorMes
...
+		return false
 	}
-
-	if quickResult != core.TSUnknown {
-		return quickResult == core.TSTrue
+	if c.reportObviousDecoratorErrors(node) || c.reportObviousModifierErrors(node) {
```
DanielRosenwasser, line 206: There's actually a semantic difference here. The original code does really use a tristate to model this, but it's `true`/`false`/`undefined`:

```ts
    /**
     * true | false: Early return this value from checkGrammarModifiers.
     * undefined: Need to do full checking on the modifiers.
     */
```

ahejlsberg: I know, but that old code was ick! Totally confusing to understand, and obviously lead to errors in the port.

DanielRosenwasser: I didn't like it either, but I think the purpose was just to bail out early to avoid extra work in cases where no error is reported, so if we're okay with the added cost, it's fine for now.

One alternative would just be to have a return type like `(errorReported bool, needsMoreChecks bool)` and just write it like this:

```go
errReported, needsMoreChecks := reportObviousDecoratorErrors(node)
if !needsMoreChecks {
  return errReported
}

errReported, needsMoreChecks := reportObviousModifierErrors(node)
if !needsMoreChecks {
  return errReported
}
```

ahejlsberg: It was totally obtuse. There's no need for the tristate and the code in this PR bails right away when there are no modifiers.

PR #111, opened 2024-11-27
Fix misplaced read semaphore

jakebailey:
I put this on the wrong FS.


internal/vfs/vfs.go
```diff
@@ -69,7 +69,6 @@ var _ FS = (*vfs)(nil)
 // For paths like `c:/foo/bar`, fsys will be used as though it's rooted at `/` and the path is `/c:/foo/bar`.
 func FromIOFS(useCaseSensitiveFileNames bool, fsys fs.FS) FS {
 	return &vfs{
-		readSema: osReadSema,
```
DanielRosenwasser, line 72: Is there ever a case where a user would want a virtual file system with a read semaphore?

jakebailey: I would think not

PR #112, opened 2024-11-27
Add FS wrapper to correctly emulate case insensitivity in testing

jakebailey:
My PR to add a VFS was wrong in that it claimed to support case insensitive FSs, but did not in practice.

To handle this, the new `vfstest` package's `FromMapFS` takes a `fstest.MapFS` and converts it into an `vfs.FS` which "canonicalizes" paths, and queries against that canonical set. But, the files themselves report their names as the original paths, just like a real OS FS. It also has a `Realpath` method which will report that full name.

This revealed that while the module resolution tests said they were case insensitive, they actually only pass with case sensitive paths; the change to that file reflects me flipping that pending a later investigation.

Along the way:

- Flip `useCaseSensitiveFileNames bool, fsys fs.FS` to `fsys fs.FS, useCaseSensitiveFileNames bool`
- Disable `wrapcheck`, which is getting too annoying for these fiddly bits. Maybe I should just put it back.
- Export `tspath.GetCanonicalFileName` for use in `vfstest`. Maybe its name should change.

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,0 +1,10 @@
+# yaml-language-server: $schema=https://golangci-lint.run/jsonschema/custom-gcl.jsonschema.json
+
+version: v1.62.2 # NOTE: this must match golangciLintVersion in Herebyfile.mjs
+
+destination: ./_tools
+
+plugins:
+  - module: 'github.com/microsoft/typescript-go/_tools'
+    import: 'github.com/microsoft/typescript-go/_tools/customlint'
+    path: ./_tools
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: npm ci
 
@@ -60,6 +61,7 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: npm ci
 
@@ -77,6 +79,7 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: npm ci
 
@@ -91,6 +94,7 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: npm ci
 
@@ -106,6 +110,7 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: npm ci
 
@@ -121,8 +126,10 @@ jobs:
       - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
           go-version: '>=1.23.0'
+          cache-dependency-path: '**/*.sum'
 
       - run: go mod tidy
+      - run: go -C ./_tools mod tidy
 
       - run: git add .
       - run: git diff --staged --exit-code --stat
--- a/.gitignore
+++ b/.gitignore
@@ -176,3 +176,5 @@ go.work.sum
 # Local baselines
 testdata/baselines/local
 
+custom-gcl
+custom-gcl.exe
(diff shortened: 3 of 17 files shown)


internal/compiler/module/resolver_test.go
```diff
@@ -311,9 +313,13 @@ func newVFSModuleResolutionHost(files map[string]string) *vfsModuleResolutionHos
 			Data: []byte(content),
 		}
 	}
+
+	const useCaseSensitiveFileNames = true
```
jakebailey, line 317: Note `true` instead of `false`. Flipping it back nets:

```
--- FAIL: TestModuleResolver (0.08s)
    --- FAIL: TestModuleResolver/typingsLookup3.ts (0.00s)
        resolver_test.go:440: assertion failed: resolved.IsResolved() is true
FAIL
FAIL    github.com/microsoft/typescript-go/internal/compiler/module     0.103s
```

So this test appears to have been passing due to an error in my FS casing implementation.

internal/vfs/vfs.go
```diff
@@ -65,12 +65,37 @@ var (
...
+
 // FromIOFS creates a new FS from an [fs.FS].
+//
 // For paths like `c:/foo/bar`, fsys will be used as though it's rooted at `/` and the path is `/c:/foo/bar`.
-func FromIOFS(useCaseSensitiveFileNames bool, fsys fs.FS) FS {
+//
+// If the provided [fs.FS] implements [RealpathFS], it will be used to implement the Realpath method.
+//
+// Deprecated: FromIOFS does not actually handle case-insensitivity; ensure the passed in [fs.FS]
+// respects case-insensitive file names if needed. Consider using [vfstest.FromMapFS] for testing.
```
jakebailey, line 80: This comment causes the editor to strikethrough the name.

internal/vfs/vfs.go
```diff
@@ -275,7 +298,10 @@ func (v *vfs) WalkDir(root string, walkFn WalkDirFunc) error {
 }
 
 func (v *vfs) Realpath(path string) string {
-	_ = rootLength(path) // panic if not absolute
-	path, _ = v.realpath(path)
-	return path
+	root, rest := splitPath(path)
+	realpath, err := v.realpath(root + rest)
```
jakebailey, line 307: splitPath does a bunch of normalization, so I'm using it here and then jamming the two back together on purpose.

internal/vfs/vfs.go
```diff
@@ -65,11 +65,37 @@ var (
...
-func FromIOFS(useCaseSensitiveFileNames bool, fsys fs.FS) FS {
+//
+// If the provided [fs.FS] implements [RealpathFS], it will be used to implement the Realpath method.
+//
+// Deprecated: FromIOFS does not actually handle case-insensitivity; ensure the passed in [fs.FS]
+// respects case-insensitive file names if needed. Consider using [vfstest.FromMapFS] for testing.
+func FromIOFS(fsys fs.FS, useCaseSensitiveFileNames bool) FS {
+	var realpath func(path string) (string, error)
+	if fsys, ok := fsys.(RealpathFS); ok {
+		fsysRealpath := fsys.Realpath
```
DanielRosenwasser: Does doing this cause an extra allocation on `fsysRealpath`? If you just capture `fsys` in `realpath` below, is that slightly better?

internal/vfs/vfs.go
DanielRosenwasser: I would add that comment inline, or just write the normalization logic separately.

internal/vfs/vfs.go
jakebailey: It's sort of pointless, I'll remove it. We don't actually pass it as a callback directly so capturing fsys is fine.

internal/vfs/vfstest/vfstest.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, err
+	}
+
+	info := must(f.Stat())
+
+	newInfo, ok := convertInfo(info)
+	if !ok {
+		// This is a synthesized dir.
+		if name != "." {
+			panic(fmt.Sprintf("unexpected synthesized dir: %q", name))
```
DanielRosenwasser, line 162: does `!ok` really imply a synthesized directory, or just an unrecognized one?

DanielRosenwasser: Is it the case that the built-in FS always has entries for `.` but not `..`?

jakebailey: > does `!ok` really imply a synthesized directory, or just an unrecognized one?

I have forcibly added my own `Sys` field to every file that the FS contains, so the only `!ok` cases are synthesized. Anything else will be an error above.

> Is it the case that the built-in FS always has entries for `.` but not `..`?

In the fstest abstraction, you can't pop out above the root. So `.` can only mean the root.

DanielRosenwasser: Okay, so `"."` is just the root directory and only handled in a few places. There is no `"."` and `".."` synthesized per directory. `""` is not a valid directory path.

jakebailey: Correct all around.

jakebailey:
> Does it make sense to add a few tests for `ReadDir` and the results you get back?

Yeah, let me add some.

internal/vfs/vfstest/vfstest_test.go
```diff
@@ -157,3 +193,11 @@ func TestFromMapFS(t *testing.T) {
 	assert.Assert(t, !ok)
 	assert.Equal(t, content, "")
 }
+
+func dirEntriesToNames(entries []fs.DirEntry) []string {
```
DanielRosenwasser, line 197: Just use `core.Map`?

jakebailey: Can't, that would be an import cycle. Also, I don't wanna

PR #113, opened 2024-11-28
Add custom linting, "emptycase" lint rule

jakebailey:
This uses `golangci-lint`'s "module plugin" system (https://golangci-lint.run/plugins/module-plugins/) to add custom rules.

The way it works is that running `golangci-lint custom` will build a custom binary which includes extra rules, and write a binary out to `_tools/custom-gcl` (or `_tools/custom-gcl.exe`). You then use that binary to lint instead. I've updated `settings.template.json` to show how to configure `vscode-go` to use that binary instead.

With this in place, I've made an `emptycase` lint rule, which errors when cases contain no body. These are all places where we've probably forgotten the difference between Go fallthrough and JS fallthrough. This lint rule catches a couple of places we've already messed up.

If a case is intentionally any, the rule can be disabled by putting _anything_ in the body, including a comment like:

```go
switch foo {
case 1:
	// do nothing
case 2:
	fmt.Println(foo)
}
```

Which is how I typically do this when I write code.


_tools/customlint/emptycase.go
```diff
@@ -0,0 +1,99 @@
...
+	default:
+		return
+	}
+
+	if len(body) == 1 {
+		block, ok := body[0].(*ast.BlockStmt)
+		if !ok || len(block.List) != 0 {
+			return
+		}
+		// Also error on a case statement containing a single empty block.
```
DanielRosenwasser: I would move this to the top of the current block.

_tools/customlint/emptycase.go
```diff
@@ -0,0 +1,99 @@
...
+	} else if len(body) != 0 {
+		return
+	}
+
+	for _, comment := range file.Comments {
+		if comment.Pos() > colon && comment.End() < nextCasePos {
+			return
+		}
+	}
```
DanielRosenwasser: This works, I might use `sort.Find` or `sort.Search`.

Any reason you can't just check whether the comment falls inside of `body` instead of between `colon`/`nextCasePos`?

.vscode/settings.template.json
```diff
@@ -24,6 +24,9 @@
         }
     },
     "go.lintTool": "golangci-lint",
+    "go.alternateTools": {
+        "golangci-lint": "${workspaceFolder}/_tools/custom-gcl${env:GOEXE}"
```
DanielRosenwasser, line 28: what's with GOEXE here? I don't see it in other commands (e.g. in `Herebyfile.mjs`)

_tools/customlint/emptycase.go
jakebailey: I guess I could binary search it.

The body is a slice of statements, so by being empty, there's no info I can poke at.

_tools/customlint/emptycase.go
jakebailey: I was trying to indicate that at this point, it's a case statement with a single empty block, but, ok

.vscode/settings.template.json
jakebailey: `GOEXE` is `.exe` on Windows, but the empty string on every other OS.

internal/compiler/utilities.go
```diff
@@ -470,8 +470,7 @@ func getErrorRangeForNode(sourceFile *ast.SourceFile, node *ast.Node) core.TextR
 		errorNode = getNameOfDeclaration(node)
 	case ast.KindArrowFunction:
 		return getErrorRangeForArrowFunction(sourceFile, node)
-	case ast.KindCaseClause:
-	case ast.KindDefaultClause:
+	case ast.KindCaseClause, ast.KindDefaultClause:
 		start := scanner.SkipTrivia(sourceFile.Text, node.Pos())
 		end := node.End()
 		statements := node.AsCaseOrDefaultClause().Statements.Nodes
```
jakebailey, line 476: Here's one bug; note the `AsCaseOrDefaultClause()`; this was meant to be one case.

internal/core/compileroptions.go
```diff
@@ -101,9 +101,7 @@ func (options *CompilerOptions) GetESModuleInterop() bool {
 		return options.ESModuleInterop == TSTrue
 	}
 	switch options.GetEmitModuleKind() {
-	case ModuleKindNode16:
-	case ModuleKindNodeNext:
-	case ModuleKindPreserve:
+	case ModuleKindNode16, ModuleKindNodeNext, ModuleKindPreserve:
 		return true
 	}
 	return false
```
jakebailey, line 99: The TS code was:

```ts
    esModuleInterop: {
        dependencies: ["module", "target"],
        computeValue: (compilerOptions): boolean => {
            if (compilerOptions.esModuleInterop !== undefined) {
                return compilerOptions.esModuleInterop;
            }
            switch (_computedOptions.module.computeValue(compilerOptions)) {
                case ModuleKind.Node16:
                case ModuleKind.NodeNext:
                case ModuleKind.Preserve:
                    return true;
            }
            return false;
        },
    },
```

So this is a bug too.

_tools/customlint/emptycase.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func runEmptyCase(pass *analysis.Pass) (interface{}, error) {
+	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
+
+	nodeFilter := []ast.Node{
+		(*ast.File)(nil),
+		(*ast.SwitchStmt)(nil),
+		(*ast.SelectStmt)(nil),
+	}
```
jakebailey, line 29: This is a cool trick; `inspect.Analyzer` does one big walk of all ASTs, producing what is effectively a big list of all node types as bits. Then later, `inspect.Preorder` can look at every node in the AST via a bitmask, which is much, much cheaper than a traditional AST walker (which normally have to linearly check every node type before determining that the node is uninteresting).

This analysis is shared between multiple analyzers running in the same process, so this work is effectively free to use (amortized across all analyzers golangci-lint would have run that also need to walk the AST).

.vscode/settings.template.json
jakebailey: For reference, it's not needed in `Herebyfile.mjs` because `execa`/`which` both resolve executables with `PATHEXT`, i.e. `exe;cmd;bat` etc which tell a shell or similar to also look for an exectuable with an extension.

Herebyfile.mjs
```diff
@@ -41,21 +55,15 @@ function assertTypeScriptCloned() {
 }
 
 const tools = new Map([
-    ["github.com/golangci/golangci-lint/cmd/golangci-lint", "v1.62.0"],
+    ["github.com/golangci/golangci-lint/cmd/golangci-lint", "v1.62.2"], // NOTE: this must match the version in .custom-gcl.yml
```
jakebailey: I'm actually considering not bothering to install `golangci-lint` anymore; it should be sufficient to just `go run` it because we never actually need the binary anymore.

_tools/customlint/emptycase.go
```diff
@@ -0,0 +1,112 @@
...
+		body = stmt.Body
+		colon = stmt.Colon
+	case *ast.CommClause:
+		body = stmt.Body
+		colon = stmt.Colon
+	default:
+		return
+	}
+
+	afterColon := colon + 1
```
DanielRosenwasser: I would move this closer to the usages - it's not used anywhere in the bail-out checks.

_tools/customlint/plugin_test.go
```diff
@@ -0,0 +1,194 @@
...
+				squiggle := strings.Repeat("~", max(1, len(line)-startColumn))
+				fmt.Fprintf(&buf, "\t%s\n", squiggle)
+			}
+		}
+	}
+
+	return buf.String()
+}
+
+func toWhitespace(linePrefix []byte) string {
```
DanielRosenwasser, line 183: We should eventually actually use this in the compiler itself - currently TypeScript's error display logic replaces every `\t` with a ` ` to avoid weirdness with inline tabs.

PR #115, opened 2024-11-29
Add compiler test runner

gabritto:
This adds a compiler test runner, capable of running compiler tests with error baselines, which is the only kind of baseline we have so far.

Things missing:
- No tsconfig parsing yet
- No compiler or harness options parsing yet
- No symlinks

Things temporary:
- Program creation is temporary because we don't have a proper `createProgram` yet, one that takes root files etc.

I also have to move the tests to a `compiler_test` package or something, but I don't think the tests are useful now, so maybe I'll do that later.

You can run specific compiler tests like so:
`$ go test -run TestCompilerBaselines/.*/simpleTestSingleFile github.com/microsoft/typescript-go/internal/testutil/runner`

The command for diffing baselines also has to change to something like
`git diff --diff-filter=AM --no-index ./testdata/baselines/reference ./testdata/baselines/local`

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -487,6 +487,8 @@ type Checker struct {
 	host                                    CompilerHost
 	compilerOptions                         *core.CompilerOptions
 	files                                   []*ast.SourceFile
+	fileIndexMap                            map[*ast.SourceFile]int
+	compareSymbols                          func(*ast.Symbol, *ast.Symbol) int
 	typeCount                               uint32
 	symbolCount                             uint32
 	totalInstantiationCount                 uint32
@@ -733,6 +735,8 @@ func NewChecker(program *Program) *Checker {
 	c.host = program.host
 	c.compilerOptions = program.compilerOptions
 	c.files = program.files
+	c.fileIndexMap = createFileIndexMap(c.files)
+	c.compareSymbols = c.compareSymbolsWorker // Closure optimization
 	c.languageVersion = c.compilerOptions.GetEmitScriptTarget()
 	c.moduleKind = c.compilerOptions.GetEmitModuleKind()
 	c.legacyDecorators = c.compilerOptions.ExperimentalDecorators == core.TSTrue
@@ -863,7 +867,7 @@ func NewChecker(program *Program) *Checker {
 	c.emptyStringType = c.getStringLiteralType("")
 	c.zeroType = c.getNumberLiteralType(0)
 	c.zeroBigIntType = c.getBigIntLiteralType(PseudoBigInt{negative: false, base10Value: "0"})
-	c.typeofType = c.getUnionType(core.Map(slices.Collect(maps.Keys(typeofNEFacts)), c.getStringLiteralType))
+	c.typeofType = c.getUnionType(core.Map(slices.Sorted(maps.Keys(typeofNEFacts)), c.getStringLiteralType))
 	c.flowLoopCache = make(map[FlowLoopKey]*Type)
 	c.flowNodeReachable = make(map[*ast.FlowNode]bool)
 	c.flowNodePostSuper = make(map[*ast.FlowNode]bool)
@@ -902,6 +906,14 @@ func NewChecker(program *Program) *Checker {
 	c.initializeIterationResolvers()
 	c.initializeChecker()
 	return c
+}
+
+func createFileIndexMap(files []*ast.SourceFile) map[*ast.SourceFile]int {
+	result := make(map[*ast.SourceFile]int, len(files))
+	for i, file := range files {
+		result[file] = i
+	}
+	return result
 }
 
 func (c *Checker) reportUnreliableWorker(t *Type) *Type {
@@ -15110,7 +15122,7 @@ func (c *Checker) getNamedMembers(members ast.SymbolTable) []*ast.Symbol {
 			result = append(result, symbol)
 		}
 	}
-	sortSymbols(result)
+	c.sortSymbols(result)
 	return result
 }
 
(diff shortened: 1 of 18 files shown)


testdata/lib/lib.d.ts
jakebailey, line 1: I think that checking these in is an okay workaround for working on the PR, but I think I'd kinda like to see a version of this where we instead emulate what `Herebyfile.mjs` used to do with `src/lib` but in memory, avoiding checking these files in for now.

That or we need to come up (in some other PR) with a solution for how we expect to be able to handle `lib.d.ts` in this new system...

jakebailey:
One thing I was envisioning here was that for all existing tests in the TS codebase, we don't _acutally_ check in any test artifacts, but instead diff them against what we have here and check those diffs in; if the diff is missing (or, should be deleted) the test errors out. That way we can slowly pare things down to nothing until we get to parity.

We can add _new_ tests which contribute real baselines, which could be separate, though.

It's not clear that we could do that without all of the other stuff ready, though, e.g. actual program construction. Perhaps that implies that this PR should be restructured to _not_ actually reference `testing.T`/`testing.TB`, and instead return all results in a struct for a future test to do with. Then we can diff for some cases, not diff for others.

jakebailey:
Of course, I say that, but we still need to use `t.Run()` to generate the tests in order to be able to target just one...

testdata/lib/lib.d.ts
gabritto: I'd rather we figure out how we're handling lib files in Go, and then update the test setup.

jakebailey:
Is there a test we could write for this "now" to verify that it works?

jakebailey:
Uh, I think I meant to comment that on the symbol PR.

.golangci.yml
```diff
@@ -39,7 +39,6 @@ linters:
     - unconvert
     - usestdlibvars
     - whitespace
-
```
jakebailey: This space was actually intentional 😄

jakebailey:
I approved #191; I assume this PR contains it, so I'll give this a harder look after #191 is merged.

internal/compiler/checker.go
```diff
@@ -8779,7 +8779,7 @@ func (c *Checker) getTargetOfModuleDefault(moduleSymbol *ast.Symbol, node *ast.N
 	// 		}
 	// 		exportEqualsSymbol := moduleSymbol.exports.get(ast.InternalSymbolNameExportEquals)
 	// 		exportAssignment := exportEqualsSymbol.valueDeclaration
-	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.symbolToString(moduleSymbol), compilerOptionName)
+	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.SymbolToString(moduleSymbol), compilerOptionName)
```
```suggestion
	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.symbolToString(moduleSymbol), compilerOptionName)
```

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -0,0 +1,430 @@
+package harnessutil
```
gabritto, line 1: Couldn't put this in `testutil` because it would cause a cycle with `vfstest`.

jakebailey: `testutil` itself is only meant for helpers for writing tests (e.g. an extension to `assert`) and should be dependency free, so I think what you've done here is good. Potentially we should just move its contents to a package like `assertx` or `xassert` (not an uncommon naming convention to "extend" another package), but I'm not sure.

jakebailey: Honestly to me this means that `testutil` is actually the wrong place for any of this harness/runner stuff; I think I just didn't notice where it was going initially...

gabritto: Should I move `runner`, this one, and `baseline` to some other folder? Maybe "testharness".

jakebailey: I think yes, but I don't think it has to be this PR.

internal/testutil/runner/compiler_runner.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/microsoft/typescript-go/internal/repo"
+	"github.com/microsoft/typescript-go/internal/testutil/baseline"
+	"github.com/microsoft/typescript-go/internal/testutil/harnessutil"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+var (
+	compilerBaselineRegex = regexp.MustCompile(`\.tsx?$`)
+	requireRegex          = regexp.MustCompile(`require\(`)
+	referencesRegex       = regexp.MustCompile(`reference\spath`)
```
jakebailey: At some level I wonder if these can be dropped, given these seem like things that can be pretty easily found with regular string operations...

internal/testutil/runner/compiler_runner.go
```diff
@@ -0,0 +1,289 @@
...
+	testFiles    []string
+	basePath     string
+	testSuitName string
+}
+
+var _ Runner = (*CompilerBaselineRunner)(nil)
+
+func NewCompilerBaselineRunner(testType CompilerTestType) *CompilerBaselineRunner {
+	testSuitName := testType.String()
+	basePath := "tests/cases/" + testSuitName
```
jakebailey, line 60: Does this need to change, or is it a part of some other message we're expecting to keep/

gabritto: I'm not sure I understand, why would it need to change? `basePath` is the path within `testdata` where we get the test files from.

PR #116, opened 2024-11-30
Finish porting control flow analysis

ahejlsberg:
This PR includes

* Finish porting control flow analysis (modulo call expessions).
* Finish porting union type reduction logic.
* Port reachability and assignment analysis.
* Port `checkComputedPropertyName` (to silence erroneous grammar checks).
* Fix two issues that caused non-deterministic type checks.

Some of the above contain computationally expensive logic, so as expected check times are now a bit longer (about 10%).

The issues in the last bullet were sort of interesting. I noticed that successive runs would print out slightly different numbers of allocated types and error outputs would sometimes differ slightly in union type ordering. Turned out to be a combination of `compareSymbols` not providing a stable ordering and `createUnionOrIntersectionProperty` being sensitive to iteration order of maps.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -12924,8 +12924,8 @@ func (c *Checker) removeSubtypes(types []*Type, hasObjectTypes bool) []*Type {
 	hasEmptyObject := hasObjectTypes && core.Some(types, func(t *Type) bool {
 		return t.flags&TypeFlagsObject != 0 && !c.isGenericMappedType(t) && c.isEmptyResolvedType(c.resolveStructuredTypeMembers(t))
 	})
-	len := len(types)
-	i := len
+	length := len(types)
+	i := length
 	count := 0
 	for i > 0 {
 		i--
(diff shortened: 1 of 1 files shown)


internal/compiler/program.go
```diff
@@ -166,16 +166,15 @@ func (p *Program) tryLoadNodeModule(modulePath string) *ast.SourceFile {
...
 
 func (p *Program) GetSyntacticDiagnostics(sourceFile *ast.SourceFile) []*ast.Diagnostic {
-	return p.getDiagnosticsHelper(sourceFile, p.getSyntaticDiagnosticsForFile)
+	return p.getDiagnosticsHelper(sourceFile, false /*ensureBound*/, p.getSyntaticDiagnosticsForFile)
 }
 
 func (p *Program) GetBindDiagnostics(sourceFile *ast.SourceFile) []*ast.Diagnostic {
-	p.bindSourceFiles()
-	return p.getDiagnosticsHelper(sourceFile, p.getBindDiagnosticsForFile)
+	return p.getDiagnosticsHelper(sourceFile, true /*ensureBound*/, p.getBindDiagnosticsForFile)
```
DanielRosenwasser, line 173: Why not just make `getBindDiagnosticsForFile` do the binding instead?

ahejlsberg: Because we want binding to be parallelized if possible and it's too late once we're in the single file only function.

PR #119, opened 2024-12-03
Fix code for parsing `import { type default as ... }`.

DanielRosenwasser:
This change amends errors for the following code sample:

<details>
<summary>Code Sample</summary>

```ts
export { default } from "foo"
export { as } from "foo";
export { as as } from "foo";
export { as as as } from "foo";
export { foo } from "foo";
export { default } from "foo";
export { continue } from "foo";
export { default as } from "foo";
export { continue as } from "foo";
export { default as foo } from "foo";
export { continue as foo } from "foo";
export { default as default } from "foo";
export { continue as continue } from "foo";
export { type } from "foo";
export { type as } from "foo";
export { type as as } from "foo";
export { type as as as } from "foo";
export { type foo } from "foo";
export { type default } from "foo";
export { type continue } from "foo";
export { type default as } from "foo";
export { type continue as } from "foo";
export { type default as foo } from "foo";
export { type continue as foo } from "foo";
export { type default as default } from "foo";
export { type continue as continue } from "foo";

import { default } from "foo"
import { as } from "foo";
import { as as } from "foo";
import { as as as } from "foo";
import { foo } from "foo";
import { default } from "foo";
import { continue } from "foo";
import { default as } from "foo";
import { continue as } from "foo";
import { default as foo } from "foo";
import { continue as foo } from "foo";
import { default as default } from "foo";
import { continue as continue } from "foo";
import { type } from "foo";
import { type as } from "foo";
import { type as as } from "foo";
import { type as as as } from "foo";
import { type foo } from "foo";
import { type default } from "foo";
import { type continue } from "foo";
import { type default as } from "foo";
import { type continue as } from "foo";
import { type default as foo } from "foo";
import { type continue as foo } from "foo";
import { type default as default } from "foo";
import { type continue as continue } from "foo";
```

</details>

The errors produced are now consistent with the current TypeScript compiler, with the following delta:

```diff
 /workspaces/typescript-go/blah/foo.ts(3,16): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(8,21): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(9,22): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(19,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(20,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(21,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(21,26): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(22,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(22,27): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(23,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(24,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(25,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(26,15): error TS1003: Identifier expected.
+/workspaces/typescript-go/blah/foo.ts(28,10): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(30,16): error TS1003: Identifier expected.
+/workspaces/typescript-go/blah/foo.ts(33,10): error TS1003: Identifier expected.
+/workspaces/typescript-go/blah/foo.ts(34,10): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(35,21): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(36,22): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(39,21): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(40,22): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(46,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(47,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(48,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(48,26): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(49,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(49,27): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(50,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(51,15): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(52,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(52,26): error TS1003: Identifier expected.
-/workspaces/typescript-go/blah/foo.ts(53,15): error TS1003: Identifier expected.
 /workspaces/typescript-go/blah/foo.ts(53,27): error TS1003: Identifier expected.
```

This is because the code unconditionally reported an error when encountering a keyword; however, in the original code, subsequent calls to `parseModuleExportName` would call `parseNameWithKeywordCheck`, which resets the state on whether to *later* error.

### DIFF
--- a/internal/compiler/parser.go
+++ b/internal/compiler/parser.go
@@ -1948,7 +1948,7 @@ func (p *Parser) parseImportOrExportSpecifier(kind ast.Kind) (isTypeOnly bool, p
 	// let checkIdentifierEnd = scanner.getTokenEnd();
 	canParseAsKeyword := true
 	disallowKeywords := kind == ast.KindImportSpecifier
-	nameOk := true
+	var nameOk bool
 	name, nameOk = p.parseModuleExportName(disallowKeywords)
 	if name.Kind == ast.KindIdentifier && name.AsIdentifier().Text == "type" {
 		// If the first token of an import specifier is 'type', there are a lot of possibilities,


internal/compiler/parser.go
```diff
@@ -1990,29 +1991,35 @@ func (p *Parser) parseImportOrExportSpecifier(kind ast.Kind) (isTypeOnly bool, p
...
 
-func (p *Parser) parseModuleExportName(disallowKeywords bool) *ast.Node {
+func (p *Parser) parseModuleExportName(disallowKeywords bool) (node *ast.Node, potentialKeywordRange *core.TextRange) {
 	if p.token == ast.KindStringLiteral {
-		return p.parseLiteralExpression()
+		return p.parseLiteralExpression(), nil
 	}
 	if disallowKeywords && isKeyword(p.token) && !p.isIdentifier() {
-		p.parseErrorAtCurrentToken(diagnostics.Identifier_expected)
+		errorRange := p.scanner.TokenRange()
```
DanielRosenwasser: I could also avoid an allocation by just returning the node itself and then skipping trivia. I kind of prefer that.

internal/compiler/parser.go
```diff
@@ -1990,29 +1991,35 @@ func (p *Parser) parseImportOrExportSpecifier(kind ast.Kind) (isTypeOnly bool, p
...
+
 	return isTypeOnly, propertyName, name
 }
 
 func (p *Parser) canParseModuleExportName() bool {
 	return tokenIsIdentifierOrKeyword(p.token) || p.token == ast.KindStringLiteral
 }
 
-func (p *Parser) parseModuleExportName(disallowKeywords bool) *ast.Node {
+func (p *Parser) parseModuleExportName(disallowKeywords bool) (node *ast.Node, potentialKeywordRange *core.TextRange) {
```
sandersn: should it be `invalidKeywordRange` to match the callers? It's either keyword or nil, right?

internal/compiler/parser.go
sandersn: so, using `p.token` instead of `p.scanner.TokenRange()`? Do you have to skip trivia because p.token's range isn't the same as TokenRange?

DanielRosenwasser:
> so, using p.token instead of p.scanner.TokenRange()? Do you have to skip trivia because p.token's range isn't the same as TokenRange?

By the time we know we need to error, we've already consumed the token so `p.token` can't be used. But it's really not a big deal. The current code diverges from the original TS in naming, but I think it is a bit simpler.

PR #121, opened 2024-12-03
Improve CI caching, consolidate setup-go

jakebailey:
Before:

image

After:

image

Windows CI perf continues to be cursed, but everything else gets faster.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,7 +39,6 @@ linters:
     - unconvert
     - usestdlibvars
     - whitespace
-    - wrapcheck
 
     - customlint
 
--- a/internal/compiler/module/resolver_test.go
+++ b/internal/compiler/module/resolver_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/testutil/baseline"
 	"github.com/microsoft/typescript-go/internal/tspath"
 	"github.com/microsoft/typescript-go/internal/vfs"
+	"github.com/microsoft/typescript-go/internal/vfs/vfstest"
 	"gotest.tools/v3/assert"
 )
 
@@ -40,8 +41,6 @@ var skip = []string{
 	"bundlerNodeModules1(module=preserve).ts",
 	"bundlerRelative1(module=esnext).ts",
 	"bundlerRelative1(module=preserve).ts",
-	"checkExportsObjectAssignProperty.ts",
-	"checkObjectDefineProperty.ts",
 	"commonJsExportTypeDeclarationError.ts",
 	"commonSourceDir5.ts",
 	"commonSourceDirectory.ts",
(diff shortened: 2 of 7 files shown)


.github/actions/setup-go/action.yml
```diff
@@ -0,0 +1,31 @@
...
+      id: install-go
+      uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+      with:
+        go-version: ${{ inputs.go-version }}
+        cache: false
+
+    - name: Go cache
+      uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+      with:
+        # There is more code downloaded and built than is covered by '**/go.sum',
```
DanielRosenwasser, line 23: Why?

.github/actions/setup-go/action.yml
```diff
@@ -0,0 +1,31 @@
...
+        # so give each job its own cache to try and not end up sharing the wrong
+        # cache between jobs.
+        key: ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ github.workflow }}-${{ inputs.cache-name }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs') }}
```
DanielRosenwasser, line 26: Does this need to look at `package-lock.json`?

.github/actions/setup-go/action.yml
jakebailey: `go tool` isn't available so the herebyfile installs deps, which aren't in any go.mod. That and custom linting downloads the linter's code and builds it, which is also not in any declared dep. The best we can do is to key on the herebyfile which contains references to deps.

.github/actions/setup-go/action.yml
jakebailey: No, there's no Go deps pinned by a package-lock file.

DanielRosenwasser: Okay, so this is solely about caching Go dependencies.

.github/actions/setup-go/action.yml
DanielRosenwasser: Is this because Go does not have the notion of `devDependencies`?

jakebailey: Correct, `go tool` helps with that but it won't be available until Go 1.24.

jakebailey: That and, `golangci-lint custom` doesn't pin anything at all from within ones own repo.

jakebailey:
It's absurd that installing `gotestsum` takes a full minute on Windows. Even pulling the cache takes a minute. This has been my experience with Windows CI caching ☹️

PR #127, opened 2024-12-03
Fix lockfile after bad dependabot PR

jakebailey:
#120 incorrectly changed the lockfile; relock it to fix.


DanielRosenwasser:
Sorry!

jakebailey:
It's okay, only I am paranoid enough to read lockfile diffs

PR #131, opened 2024-12-04
Implement module resolution cache

andrewbranch:
Adds the (single-program) module resolution cache and tests `FailedLookupLocations`/`AffectingLocations` against baselines.


internal/compiler/module/resolver_test.go
```diff
@@ -49,31 +45,27 @@ var skip = []string{
...
-	"declarationEmitUnnessesaryTypeReferenceNotAdded.ts",
 	"declarationEmitUsingAlternativeContainingModules1.ts",
 	"declarationEmitUsingAlternativeContainingModules2.ts",
 	"declarationEmitUsingTypeAlias1.ts",
 	"declarationEmitUsingTypeAlias2.ts",
-	"declarationEmitWithInvalidPackageJsonTypings.ts",
 	"decoratorMetadataTypeOnlyExport.ts",
 	"decoratorMetadataTypeOnlyImport.ts",
+	"emit(jsx=preserve).ts",
+	"emit(jsx=react).ts",
```
andrewbranch, line 64: Added skips had failed lookup locations in either package.json "imports" or symlinks.

internal/compiler/module/cache.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type ModeAwareCache[T any] map[ModeAwareCacheKey]T
+
+type caches struct {
+	moduleNameCache             *resolutionCache[*ResolvedModuleWithFailedLookupLocations]
+	typeReferenceDirectiveCache *resolutionCache[*ResolvedTypeReferenceDirectiveWithFailedLookupLocations]
+	packageJsonInfoCache        *packagejson.InfoCache
+}
```
andrewbranch, line 17: I’ve made a lot of stuff internal structs for now that _may_ need to become public interfaces later, but I actually think much less of it will need to be public than is today. TS’s public APIs allow consumers to override module resolution and caching, and it’s unclear still how/whether any of that will be allowed in Corsa. In our own usage, `tsc` always runs with this cache that I’ve implemented here, while `tsc --watch` and TS Server wrap this cache in another layer (resolutionCache.ts) that allows the cache to persist between program updates and handles a lot of file watching logic.

internal/compiler/module/cache.go
```diff
@@ -0,0 +1,316 @@
...
+		useCaseSensitiveFileNames: useCaseSensitiveFileNames,
+		options:                   options,
+		getResolvedFileName:       getResolvedFileName,
+		moduleNameToDirectoryMap:  newCacheWithRedirects[ModeAwareCacheKey, *perNonRelativeNameCache[T]](options, optionsToRedirectsKey),
+	}
+}
+
+func (c *nonRelativeNameResolutionCache[T]) getFromNonRelativeNameCache(nameAndMode ModeAwareCacheKey, directoryName string, redirectedReference *ResolvedProjectReference) (T, bool) {
+	if tspath.IsExternalModuleNameRelative(nameAndMode.name) {
+		panic("module name must be non-relative")
```
jakebailey, line 132: For these panics, I would suggest including the bad string in the panic message.

internal/compiler/module/cache.go
```diff
@@ -0,0 +1,316 @@
...
+//
+//	[
+//	  /a/b/c/d, /a/b/c, /a/b
+//	]
+//
+// this means that request for module resolution from file in any of these folder will be immediately found in cache.
+func (c *perNonRelativeNameCache[T]) set(directory tspath.Path, value T, getResolvedFileName func(result T) tspath.Path) {
+	if c == nil {
+		c = &perNonRelativeNameCache[T]{}
+	}
```
jakebailey: This seems suspicious; this isn't going to go set the caller's reference to something, it's just going to use this instance within the function and then throw it away, as though the function had returned early (since it has no side effects).

internal/compiler/module/cache.go
andrewbranch: It’s likely to be user code

jakebailey: Is that a gotcha for something? I thought we already had debug asserts with this kind of thing elsewhere but maybe not for editor telemetry reasons?

andrewbranch: I thought we always avoided putting source file content or filenames in asserts, since a compiler crash can always show up in editor telemetry.

PR #132, opened 2024-12-05
Add CHANGES.md with JSDoc proposal

sandersn:
I'm putting the JSDoc proposal here in typescript-go because it's a private repo where we can review it before making it public after the Corsa announcement. Please review for two things at least:

1. Does this get the right message across when it's made public?
2. Is the list of things to cut good?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,16 +4,16 @@ I propose dropping some features in order to simplify and speed up the compiler 
 
 When we added Javascript support to the Typescript compiler 7-8 years ago, we tried to support 4 kinds of users:
 
-1. JS-only -- people opening a random JS file in VS/VSCode, often by somebody who's never heard of TS.
-2. TS-early-adopter -- somebody who's adding `// @ts-check` to an existing code base in order to convince their team of the benefits of Typescript.
-3. Closure -- an entire team writing for Google's Closure compiler. At the time, Closure was the source most of JSDoc type syntax, and the bulk of typechecked JSDoc code in the world.
+1. JS-only -- people opening a random JS file in VSCode, often by somebody who's never heard of TS.
+2. TS-early-adopter -- somebody who's adding `// @ts-check` to a JSDoc-but-unchecked code base in order to convince their team of the benefits of Typescript.
+3. Closure/other JSDoc processors -- an entire team writing for Google's Closure compiler, which used JSDoc type syntax, or perhaps the original JSDoc documentation generator.
 4. TS-via-JSDoc -- people who wanted to use Typescript but
 	- didn't want a build step.
 	- didn't want to rely on a non-standard JS variant with a relatively short history.
 	- didn't want to restrict their contributor pool to people who knew Typescript.
 	- didn't want to rely on a Microsoft-provided compiler.
 
-That is a wide variety of users, from people who don't even know they're using Typescript to those who are writing complete Typescript-first code, just in a different syntax. Today, however, only the two users on the extremes are left. People don't need to kick the tires on Typescript -- they already love it or hate it. And all the Closure codebases that I know of have migrated to Typescript (although it's still used as an optimiser). Even TS-via-JSDoc projects are getting rarer as Typescript syntax is supported everywhere in the JS ecosystem.
+That is a wide variety of users, from people who don't even know they're using Typescript to those who are writing complete Typescript-first code, just in a different syntax. Today, however, only the two users on the extremes are left. The wild variation in JSDoc is long gone, in part because JS files are now mostly written in an editor running tsserver. And all the Closure codebases that I know of have migrated to Typescript (although it's still used as an optimiser). Even TS-via-JSDoc projects are getting rarer as Typescript syntax is supported everywhere in the JS ecosystem.
 
 Also, Javascript itself has changed a lot since 2017. `class` has replaced constructor functions. ES modules are finally replacing CommonJS, and the syntax for both has long since been replaced for people with a build. In a word, Typescript's JS support needs cuts. Patterns with no Typescript equivalent are less important than in 2017, and there are many fewer to support. Those that remain are easier to analyse statically because of ES standards and because of Typescript's own place as the de facto interpreter of JSDoc semantics.
 
@@ -61,7 +61,7 @@ In addition, most JSDoc-specific type syntax stays:
 ```
 
 ### CommonJS 
-CommonJS keeps `module.exports` and `require` support; most handwritten code no longer uses it, but it's still useful so Typescript can understand semantics of a JS file inside node_modules:
+CommonJS keeps `module.exports` and `require` support; most new handwritten code no longer uses it, but it's still useful so Typescript can understand semantics of a JS file inside node_modules:
  ```js
 const fs = require('fs')
 function f() { }
@@ -164,15 +164,21 @@ function mathematical() { }
 
 ### Types
 
-#### `?`
-`?` is a synonym of `any` - Barely used, trivial semantics.
+#### Standalone `?`
+`?` is a synonym of `any` -- but it's barely used:
+
+```js
+/** @type {?} */
+var unknown
+```
+
 #### Namepaths
 `module:path~Type` - Barely used, no semantics. Displaced entirely by `import("path").Type`
 #### Postfix types
 `T!` and `T?`, but keep `!T` and `?T` - The postfix variants came from Closure. See below. Prefix `?T` is used in Flow, which is useful for TS to understand.
 
 #### `String`, `Number` etc synonyms
-Using `String` for `string` is rare in modern code, so I decided that I'd rather have more `any` in random JS files, but less confusion among people who write TS-via-JSDoc.
+Using `String` for `string` is rare in modern code, so I decided that I'd rather have slightly wrong types in a few JS files, but less confusion among people who write TS-via-JSDoc.
 
 ```ts
 /** @type {Number} */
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+```
+
+In addition, most JSDoc-specific type syntax stays:
+
+```js
+/** @typedef {{ anything: *, indexSignature: Object.<string,string> }} Example */
+```
+
+### CommonJS 
+CommonJS keeps `module.exports` and `require` support; most handwritten code no longer uses it, but it's still useful so Typescript can understand semantics of a JS file inside node_modules:
```
DanielRosenwasser: Most *new* handwritten code.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Tags
+Note that removing support for a tag will still parse it, but as `JSDocUnknownTag`.
+
+#### `@class`
+`@class` was only useful for constructor functions, and Typescript rarely needs help identifying constructor functions. Basically unused in handwritten code. 
+
+**NOTE**: Typescript will continue to *emit* `@class` for ES5 in order to aid other tools.
+```js
+/** @class - is redundant because TS uses `this.x = x` as a marker. */
```
DanielRosenwasser, line 141: I guess the tradeoff here is that there's no way to declare that something as a base class even if it declares no properties/methods.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+ */
+function mathematical() { }
+```
+#### `@enum`
+`@enum` is a Closure tag. See below.
+
+### Types
+
+#### `?`
+`?` is a synonym of `any` - Barely used, trivial semantics.
```
DanielRosenwasser: You mean bare `?`?

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+#### Namepaths
+`module:path~Type` - Barely used, no semantics. Displaced entirely by `import("path").Type`
+#### Postfix types
+`T!` and `T?`, but keep `!T` and `?T` - The postfix variants came from Closure. See below. Prefix `?T` is used in Flow, which is useful for TS to understand.
+
+#### `String`, `Number` etc synonyms
+Using `String` for `string` is rare in modern code, so I decided that I'd rather have more `any` in random JS files, but less confusion among people who write TS-via-JSDoc.
```
DanielRosenwasser: I don't understand what you mean about `any` here. Do you mean the `Object` -> `any` override?

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+/** @type {(x: number) => number} */
+var sub1 = x => x - 1
+```
+And the remaining uses are rare -- about 0.2% of the most-used tag, `@param`:
+```js
+/** @type {FancySignature} */
+function f(x,y,z) {
+}
+```
+The workaround is easy -- switch to a variable declaration with an arrow function.
```
DanielRosenwasser, line 217: Hoisting.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+#### `@type` on functions
+This is a nice feature from the original JSDoc document processor, but around 80% of uses are in places where an arrow would already be contextually typed:
+```js
+/** @type {(x: number) => number} */
+var sub1 = x => x - 1
+```
+And the remaining uses are rare -- about 0.2% of the most-used tag, `@param`:
+```js
+/** @type {FancySignature} */
+function f(x,y,z) {
```
DanielRosenwasser, line 214: I would still really prefer we don't drop this one.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+}
+```
+The workaround is easy -- switch to a variable declaration with an arrow function.
+#### `function(new, string, string): T` 
+This is Closure syntax. See below.
+
+### Constructor functions
+#### Multiple assignments to the same `C.prototype.property` 
+Currently the types are unioned, but will instead take the type of the first declaration. This code is quite rare.
```
DanielRosenwasser: Rephrase to make it more explicit that you're *proposing* to take the type of the first assignment. As it's currently phrased, it almost sounds like there might be a special case where that kicks in.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+this.p = 12
+// inside a module is equivalent to
+module.exports.p = 12
+// outside a module is equivalent to
+global {
+  var p: number
+}
+```
+#### Multiple assignments to the same export
+Currently the types are determined using control flow, but will instead take the type of the first declaration. This code is quite rare, and is used for platform selection, not mutation, so control flow is overkill.
```
DanielRosenwasser: I'm confused with the way this is phrased. As it stands, it looks like this will still use control flow, but possibly won't union all the assignments? Is that right?

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+```js
+if (isWindows) {
+  module.exports = { ... }
+} else {
+  module.exports = { ... }
+}
+```
+#### single-property access `require`
+```js
+var readFileSync = require('fs').readFileSync
```
DanielRosenwasser, line 287: I am actually surprised you didn't see this more often in your investigations.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+```
+#### ignoring preventative empty assignment of `module.exports`:
+```js
+module.exports = {}
+module.exports.f = function() { ... }
+```
+Modern code assumes that `module.exports` is always defined.
+
+
+**NOTE**: I would like to drop CommonJS entirely, since it is rare to write code in CommonJS these days. But lots of emitted code is CommonJS, so it will be a few years before people stop running into it occasionally in node_modules.
```
DanielRosenwasser: I don't mean to nitpick, but the way this is phrased (here and a bit above) still makes it sound like CommonJS usage will vanish in a few years' time, and I don't think this is the right framing. CommonJS is definitely now rarer for web code and as the source module system for most libraries, but it is definitely used for smaller Node apps, libraries, and scripts. Using ESM has often been painful, and for many people, the option of using ESM will really only open up in the next few months with `require(ESM)`.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
+# JSDoc and JS inference support
+JSDoc and JS inference in Typescript supports a huge variety of users, some of which no longer exist. 
+I propose dropping some features in order to simplify and speed up the compiler and language service.
+
+When we added Javascript support to the Typescript compiler 7-8 years ago, we tried to support 4 kinds of users:
+
+1. JS-only -- people opening a random JS file in VS/VSCode, often by somebody who's never heard of TS.
+2. TS-early-adopter -- somebody who's adding `// @ts-check` to an existing code base in order to convince their team of the benefits of Typescript.
+3. Closure -- an entire team writing for Google's Closure compiler. At the time, Closure was the source most of JSDoc type syntax, and the bulk of typechecked JSDoc code in the world.
```
DanielRosenwasser: I think it was more that Closure and the JSDoc tools both had a specified set of constructs, and so there was a desire not to forge a separate path. So if we implemented something, it'd be ideal if it worked across all the tools.

CHANGES.md
```diff
@@ -0,0 +1,391 @@
...
+4. TS-via-JSDoc -- people who wanted to use Typescript but
+	- didn't want a build step.
+	- didn't want to rely on a non-standard JS variant with a relatively short history.
+	- didn't want to restrict their contributor pool to people who knew Typescript.
+	- didn't want to rely on a Microsoft-provided compiler.
+
+That is a wide variety of users, from people who don't even know they're using Typescript to those who are writing complete Typescript-first code, just in a different syntax. Today, however, only the two users on the extremes are left. People don't need to kick the tires on Typescript -- they already love it or hate it. And all the Closure codebases that I know of have migrated to Typescript (although it's still used as an optimiser). Even TS-via-JSDoc projects are getting rarer as Typescript syntax is supported everywhere in the JS ecosystem.
```
DanielRosenwasser: > People don't need to kick the tires on Typescript -- they already love it or hate it.

I dunno if I could make this statement, and I don't necessarily think that most of what you're proposing gets in the way of taking the opposite stance.

CHANGES.md
sandersn: True, but subclassing never worked for constructor functions.

CHANGES.md
sandersn: Yeah, let me give a longer example to make it clear.

CHANGES.md
sandersn: I'm pretty sure that people are writing the destructuring and letting a compiler emit statement+alias for them when targetting ES5.

CHANGES.md
sandersn: Oh, oops. You're right, `String` won't be `any`, it'll be `String`. That's actually quite close to `string`.

CHANGES.md
sandersn: In that case the workaround is clunky -- you either move the function or write individual `@param` tags that use `Parameters`.

CHANGES.md
sandersn: Here are my reasons for dropping it. Should I put them in the proposal too? At first, I didn't want to make this section a lot larger than the others.

1. It's both hard **and** annoying to support -- FancySignature has to be resolved in the checker. And even for looking up signatures which are syntactically available, the current code forces you to use getEffectiveParameterType all over, which has been easy to forget. It's been a bug farm over the years.
2. Usage is low. It's concentrated in a few projects.
3. The workaround is usually simple. Hoisting doesn't happen that often.
4. Usage is going to go lower as people switch to TS even for scripts. You can't write this in TS at all.

To me, this is clearly a feature to wait for people (who aren't us) to request after we release.

CHANGES.md
sandersn: No, control flow will go away and be replaced with use-first. I'll reword this.

CHANGES.md
sandersn: Hmm, I'll drop this paragraph then. It's not low-level detail like the surrounding points, and is speculative.

CHANGES.md
sandersn: Yeah, I should focus on people needing to run TS on legacy JS rather than "kick the tires". That's what's mostly gone away.

PR #133, opened 2024-12-06
Port more type checking

ahejlsberg:
In this PR:

* Control flow analysis for property access and element access expressions.
* Core contextual typing.
* Checking and contextual typing of function expressions.
* Core call expression checking including overload resolution.
* Retire `*ast.MessageChain` and use `*ast.Diagnostic` for everything.

The retirement of `*ast.MessageChain` simplifies logic having to do with construction of diagnostic message chains. Now, instead of having message chains with no file/position information and diagnostic nodes with file/position information (and functions that go between the two representations), we just have diagnostic nodes throughout, and in diagnostic chains all nodes by default have the same file/position information.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -3424,10 +3424,10 @@ func (c *Checker) isSignatureApplicable(node *ast.Node, args []*ast.Node, signat
 		restArgCount := len(args) - argCount
 		var errorNode *ast.Node
 		if reportErrors {
-			switch {
-			case restArgCount == 0:
+			switch restArgCount {
+			case 0:
 				errorNode = node
-			case restArgCount == 1:
+			case 1:
 				errorNode = c.getEffectiveCheckNode(args[argCount])
 			default:
 				errorNode = c.createSyntheticExpression(node, spreadType, false, nil)
@@ -3455,9 +3455,7 @@ func (c *Checker) isSignatureApplicable(node *ast.Node, args []*ast.Node, signat
 	// }
 }
 
-/**
- * Returns the this argument in calls like x.f(...) and x[f](...). Undefined otherwise.
- */
+// Returns the `this` argument node in calls like `x.f(...)` and `x[f](...)`. `nil` otherwise.
 func (c *Checker) getThisArgumentOfCall(node *ast.Node) *ast.Node {
 	if ast.IsBinaryExpression(node) {
 		return node.AsBinaryExpression().Right


internal/compiler/relater.go
```diff
@@ -262,15 +262,15 @@ func (c *Checker) isEnumTypeRelatedTo(source *ast.Symbol, target *ast.Symbol, er
 }
 
 func (c *Checker) checkTypeAssignableTo(source *Type, target *Type, errorNode *ast.Node, headMessage *diagnostics.Message) bool {
-	return c.checkTypeRelatedToEx(source, target, c.assignableRelation, errorNode, headMessage, nil, nil)
+	return c.checkTypeRelatedToEx(source, target, c.assignableRelation, errorNode, headMessage, nil)
 }
 
-func (c *Checker) checkTypeAssignableToEx(source *Type, target *Type, errorNode *ast.Node, headMessage *diagnostics.Message, containingMessageChain func() *ast.MessageChain, errorOutputObject *ErrorOutputContainer) bool {
-	return c.checkTypeRelatedToEx(source, target, c.assignableRelation, errorNode, headMessage, containingMessageChain, errorOutputObject)
+func (c *Checker) checkTypeAssignableToEx(source *Type, target *Type, errorNode *ast.Node, headMessage *diagnostics.Message, diagnosticOutput **ast.Diagnostic) bool {
```
jakebailey, line 268: The double pointer is... fun; I feel like for ergonomic reasons I would have used a callback instead but I guess it's okay.

internal/ast/diagnostic.go
```diff
@@ -70,48 +71,9 @@ func NewDiagnostic(file *SourceFile, loc core.TextRange, message *diagnostics.Me
...
-		code:         messageChain.code,
-		category:     messageChain.category,
-		message:      messageChain.message,
-		messageChain: messageChain.messageChain,
-	}
-}
-
-// MessageChain
-
-type MessageChain struct {
```
jakebailey, line 86: We have message chains in the old TS compiler, and passing messages through to the language server or API is going to be something we have to do; is eliminating this construct going to make it harder to interop?

internal/compiler/checker.go
```diff
@@ -2628,2495 +2663,3952 @@ func (c *Checker) checkImportCallExpression(node *ast.Node) *Type {
...
-		// 		c.checkExternalEmitHelpers(node, ExternalEmitHelpersClassPrivateFieldGet)
-		// 	}
-		// }
-		lexicallyScopedSymbol := c.lookupSymbolForPrivateIdentifierDeclaration(right.Text(), right)
-		if assignmentKind != AssignmentKindNone && lexicallyScopedSymbol != nil && lexicallyScopedSymbol.ValueDeclaration != nil && ast.IsMethodDeclaration(lexicallyScopedSymbol.ValueDeclaration) {
-			c.grammarErrorOnNode(right, diagnostics.Cannot_assign_to_private_method_0_Private_methods_are_not_writable, right.Text())
+	minCount := math.MaxInt64 // smallest parameter count
+	maxCount := math.MinInt64 // largest parameter count
+	maxBelow := math.MinInt64 // largest parameter count that is smaller than the number of arguments
+	minAbove := math.MaxInt64 // smallest parameter count that is larger than the number of arguments
```
```suggestion
	minCount := math.MaxInt // smallest parameter count
	maxCount := math.MinInt // largest parameter count
	maxBelow := math.MinInt // largest parameter count that is smaller than the number of arguments
	minAbove := math.MaxInt // smallest parameter count that is larger than the number of arguments
```

Otherwise this code would fail to compile on 32 bit machines; the math constants are untyped so this was being silently converted to `int` instead of `int64` (hence why you were able to use `strconv.Itoa`).

internal/ast/ast.go
```diff
@@ -316,7 +331,7 @@ func (n *Node) Type() *Node {
 		return n.AsJSDocFunctionType().Type
 	case KindJSDocOptionalType:
 		return n.AsJSDocOptionalType().Type
-	case KindEnumMember, KindBindingElement:
+	case KindEnumMember, KindBindingElement, KindExportAssignment:
```
DanielRosenwasser, line 334: Are these for error recovery?

internal/compiler/relater.go
DanielRosenwasser: The original TypeScript is closer to what's here, and it only uses an object because there's no way to efficiently express an output parameter.

internal/ast/ast.go
```diff
@@ -4335,6 +4361,27 @@ func (node *TemplateLiteralTypeSpan) ForEachChild(v Visitor) bool {
 	return visit(v, node.Type) || visit(v, node.Literal)
 }
 
+// SyntheticExpression
+
+type SyntheticExpression struct {
+	ExpressionBase
+	Type            any
```
DanielRosenwasser, line 4368: Why is this `any`?

DanielRosenwasser: In other words, why not either make this `*Type` or `*Node`?

internal/compiler/flow.go
```diff
@@ -2085,6 +2086,53 @@ func (c *Checker) typeMaybeAssignableTo(source *Type, target *Type) bool {
...
+	// We don't infer a type if assignments are only null or undefined.
+	if everyType(flowType, c.isNullableType) {
+		return nil
+	}
+	return c.convertAutoToAny(flowType)
+}
+
+func (c *Checker) getFlowTypeInStaticBlocks(symbol *ast.Symbol, staticBlocks []*ast.Node) *Type {
+	var accessName *ast.Node
+	if strings.HasPrefix(symbol.Name, InternalSymbolNamePrefix+"#") {
```
DanielRosenwasser, line 2113: Does this construct the expected prefix on every call, or is this optimized away? Either way, I'd recommend creating a separate well-known prefix.

internal/ast/ast.go
ahejlsberg: They're so that `xxx.Name()` is `nil` instead of hitting a panic. I forget exactly which check causes us to ask for `Name` on an export assignment, but seemed fine to just handle it this way (which is what would happen in the old code because export assignments never have a name).

internal/ast/diagnostic.go
ahejlsberg: The only difference is that `Diagnostic` has file/position information. It's easy to discard it when marshaling.

internal/compiler/relater.go
ahejlsberg: The old code has a confusing combination of a callback that optionally provides the containing message chain and an "error output container" which receives the output diagnostic. The new approach here is a lot easier to understand: If you pass a non-nil `diagnosticOutput` pointer, the reported diagnostic is stored there (and not added to the checker's diagnostic collection). Otherwise, the diagnostic is reported as normal. Any diagnostic chaining then happens at the call site, which is easy to do now that there's no distinction between message chains and diagnostics. The code is easier to understand are we're not creating a bunch of useless closures.

internal/compiler/checker.go
ahejlsberg: Not sure how I missed that there is a `math.MaxInt`. Of course that's the right one to use.

internal/ast/ast.go
jakebailey: `Type` is defined in `compiler` right now, so that would be a cycle.

DanielRosenwasser: Okay, I see, I wasn't familiar with `SyntheticExpression`s. I think I do have a slight preference to make these local to only the checker, but it's your call.

internal/compiler/checker.go
```diff
@@ -2628,2495 +2663,3952 @@ func (c *Checker) checkImportCallExpression(node *ast.Node) *Type {
...
+	if restType != nil {
+		spreadType := c.getSpreadArgumentType(args, argCount, len(args), restType, nil /*context*/, checkMode)
+		restArgCount := len(args) - argCount
+		var errorNode *ast.Node
+		if reportErrors {
+			switch {
+			case restArgCount == 0:
+				errorNode = node
+			case restArgCount == 1:
+				errorNode = c.getEffectiveCheckNode(args[argCount])
```
DanielRosenwasser:
```suggestion
			switch restArgCount {
			case 0:
				errorNode = node
			case 1:
				errorNode = c.getEffectiveCheckNode(args[argCount])
```

internal/compiler/checker.go
```diff
@@ -2628,2495 +2663,3952 @@ func (c *Checker) checkImportCallExpression(node *ast.Node) *Type {
...
+	// 		if awaitedTypeOfSource != nil && c.isTypeRelatedTo(awaitedTypeOfSource, target, relation) {
+	// 			addRelatedInfo(errorOutputContainer.errors[0], createDiagnosticForNode(errorNode, Diagnostics.Did_you_forget_to_use_await))
+	// 		}
+	// 	}
+	// }
+}
+
+/**
+ * Returns the this argument in calls like x.f(...) and x[f](...). Undefined otherwise.
+ */
```
DanielRosenwasser:
```suggestion
// Returns the `this` argument node in calls like `x.f(...)` and `x[f](...)`. `nil` otherwise.
```

PR #142, opened 2024-12-06
Use gofumpt for formatting

jakebailey:
This is a stricter variant of `gofmt` which does things like:

- Ensure there's a newline between declarations
- Group vars together
- Use `:=` when the type is inferred (not var)
- Write octal literals like `0o755` instead of `0755`

I didn't do this initially because the generated code for diagnostics was reformatted into a block which would have been bad for diffing / git, but I fixed that by ignoring generated files.


internal/compiler/diagnostics/generate.go
```diff
@@ -74,7 +74,7 @@ func main() {
 
 	var buf bytes.Buffer
 
-	buf.WriteString("// GENERATED BY generate.go; DO NOT EDIT\n")
+	buf.WriteString("// Code generated by generate.go; DO NOT EDIT\n")
```
jakebailey: I mistakenly used the wrong magic string here, oops.

internal/ast/ast.go
```diff
@@ -358,507 +358,675 @@ func (n *Node) Initializer() *Node {
...
 }
+
 func (n *Node) AsQualifiedName() *QualifiedName {
 	return n.data.(*QualifiedName)
 }
+
 func (n *Node) AsSourceFile() *SourceFile {
 	return n.data.(*SourceFile)
 }
+
```
jakebailey, line 418: These extra newlines are technically a downside, as we'd actually prefer they be single line.

But I don't think it matters; we eventually want this file to be generated.

DanielRosenwasser:
> Use `:=` when the type is inferred (not var)

I assume this doesn't fight with the "inconsequential assignment" lint where the value is always immediately overwritten.

internal/ast/ast.go
DanielRosenwasser: To be honest, I think we only want them to be adjacent because they started out that way.

jakebailey:
> > Use `:=` when the type is inferred (not var)
> 
> I assume this doesn't fight with the "inconsequential assignment" lint where the value is always immediately overwritten.

It won't, it's mainly for `var x int = 1`

internal/compiler/grammarchecks.go
```diff
@@ -1157,7 +1162,8 @@ func (c *Checker) checkGrammarJsxElement(node *ast.Node, jsxCommon struct {
 	tagName       *ast.JsxTagNameExpression
 	typeArguments *ast.NodeList
 	attributes    *ast.JsxAttributesNode
-}) bool {
+},
+) bool {
```
jakebailey, line 1167: This is weird, but his is because this is the only place we have used an anonymous type as a parameter, which is weird.

jakebailey: `checkGrammarJsxElement` isn't called anywhere, weird

PR #143, opened 2024-12-06
add parseCommandLineWorker

iisaduan:
Parses a command line into a `map[string]any`. For testing, the parsed result unmarshal well into the `compilerOptions` because none of the options that can passed via commandline makes a circular structure. Not all the options are implemented in `compilerOptions` yet (but also none of the tests use things that _aren't_ yet implemented in `compilerOptions`. 

What is not included in this PR: 
- execute command line
- options map unmarshaling to `compilerOptions`
- better error handling/test baselining (tests are not baselined yet since error generation is not equivalent, and using the existing error writer introduces circular dependencies (since the error writer/formatter is currently in `compiler`)
- build
- watch

### DIFF
--- a/internal/tsoptions/commandlineparser.go
+++ b/internal/tsoptions/commandlineparser.go
@@ -1,7 +1,6 @@
 package tsoptions
 
 import (
-	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -32,20 +31,24 @@ type ParseCommandLineWorkerDiagnostics struct {
 	OptionTypeMismatchDiagnostic *diagnostics.Message
 }
 
-type OptionsBase map[string]any //CompilerOptionsValue|TsConfigSourceFile
+type OptionsBase map[string]any // CompilerOptionsValue|TsConfigSourceFile
 
 func (p *CommandLineParser) AlternateMode() *AlternateModeDiagnostics {
 	return p.workerDiagnostics.didYouMean.alternateMode
 }
+
 func (p *CommandLineParser) OptionsDeclarations() []CommandLineOption {
 	return p.workerDiagnostics.didYouMean.OptionDeclarations
 }
+
 func (p *CommandLineParser) UnknownOptionDiagnostic() *diagnostics.Message {
 	return p.workerDiagnostics.didYouMean.UnknownOptionDiagnostic
 }
+
 func (p *CommandLineParser) UnknownDidYouMeanDiagnostic() *diagnostics.Message {
 	return p.workerDiagnostics.didYouMean.UnknownDidYouMeanDiagnostic
 }
+
 func (p *CommandLineParser) GetOptionsNameMap() *NameMap {
 	p.workerDiagnostics.optionsNameMapOnce.Do(func() {
 		optionsNames := map[string]*CommandLineOption{}
@@ -74,28 +77,6 @@ type CommandLineParser struct {
 	errors    []*ast.Diagnostic
 }
 
-// todo: used in executeCommandLine
-type ParsedCommandLine struct {
-	Options    *core.CompilerOptions
-	ConfigFile *ast.SourceFile // TsConfigSourceFile
-	// TypeAquisition *core.TypeAcquisition
-	FileNames         []string
-	ProjectReferences []string // todo: core.ProjectReference
-	// WatchOptions WatchOptions
-	// Raw any
-	Errors []*ast.Diagnostic
-	// WildcardDirectories map[string]watchDirectoryFlags
-	CompileOnSave bool
-}
-
-func (p *ParsedCommandLine) GetConfigFileParsingDiagnostics() []*ast.Diagnostic {
-	if p.ConfigFile != nil {
-		// todo: !!! should be ConfigFile.ParseDiagnostics, check if they are the same
-		return slices.Concat(p.ConfigFile.Diagnostics(), p.Errors)
-	}
-	return p.Errors
-}
-
 func ParseCommandLine(
 	commandLine []string,
 	fs vfs.FS,
@@ -111,7 +92,7 @@ func parseCommandLineWorker(
 	commandLine []string,
 	fs vfs.FS,
 ) *CommandLineParser {
-	var parser = &CommandLineParser{
+	parser := &CommandLineParser{
 		fs:                fs,
 		workerDiagnostics: parseCommandLineWithDiagnostics,
 		fileNames:         []string{},
@@ -338,7 +319,7 @@ func (p *CommandLineParser) validateJsonOptionValue(
 	opt *CommandLineOption,
 	value string,
 	loc *core.TextRange,
-	sourceFile *ast.SourceFile, //TODO TsConfigSourceFile,
+	sourceFile *ast.SourceFile, // TODO TsConfigSourceFile,
 ) string {
 	if opt.extraValidation != nil {
 		d, args := opt.extraValidation(value)
--- a/internal/tsoptions/commandlineparser_test.go
+++ b/internal/tsoptions/commandlineparser_test.go
@@ -18,7 +18,7 @@ func TestCommandLineParseResult(t *testing.T) {
 	t.Parallel()
 	repo.SkipIfNoTypeScriptSubmodule(t)
 
-	var parseCommandLineSubScenarios = []*subScenarioInput{
+	parseCommandLineSubScenarios := []*subScenarioInput{
 		// --lib es6 0.ts
 		{"Parse single option of library flag", []string{"--lib", "es6", "0.ts"}},
 		{"Handles may only be used with --build flags", []string{"--build", "--clean", "--dry", "--force", "--verbose"}},
@@ -201,7 +201,7 @@ func parseExistingCompilerBaseline(t *testing.T, baseline string) *TestCommandLi
 	e := json.Unmarshal([]byte(compilerOptions), &baselineOptions)
 	assert.NilError(t, e)
 
-	var parser = TestCommandLineParser{
+	parser := TestCommandLineParser{
 		options:   *baselineOptions,
 		fileNames: fileNames,
 		errors:    errors,
--- a/internal/tsoptions/parsedcommandline.go
+++ b/internal/tsoptions/parsedcommandline.go
@@ -1,14 +1,29 @@
 package tsoptions
 
 import (
+	"slices"
+
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
+// todo: used in executeCommandLine
 type ParsedCommandLine struct {
 	Options *core.ParsedOptions
-	// ConfigFile    *ast.SourceFile  // used in Program and ExecuteCommandLine
-	Errors        []*ast.Diagnostic
-	Raw           any
+	// WatchOptions WatchOptions
+
+	ConfigFile *ast.SourceFile // TsConfigSourceFile, used in Program and ExecuteCommandLine
+	Errors     []*ast.Diagnostic
+	Raw        any
+	// WildcardDirectories map[string]watchDirectoryFlags
 	CompileOnSave *bool
+	// TypeAquisition *core.TypeAcquisition
+}
+
+func (p *ParsedCommandLine) GetConfigFileParsingDiagnostics() []*ast.Diagnostic {
+	if p.ConfigFile != nil {
+		// todo: !!! should be ConfigFile.ParseDiagnostics, check if they are the same
+		return slices.Concat(p.ConfigFile.Diagnostics(), p.Errors)
+	}
+	return p.Errors
 }


internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,389 @@
...
+			} else {
+				if optValue == "true" {
+					p.options[opt.Name] = true
+					i++
+				}
+				p.errors = append(p.errors, ast.NewDiagnostic(nil, p.errorLoc, diagnostics.Option_0_can_only_be_specified_in_tsconfig_json_file_or_set_to_false_or_null_on_command_line, opt.Name))
+			}
+		} else {
+			p.errors = append(p.errors, ast.NewDiagnostic(nil, p.errorLoc, diagnostics.Option_0_can_only_be_specified_in_tsconfig_json_file_or_set_to_null_on_command_line, opt.Name))
+			if len(optValue) != 0 && !stringutil.StartsWith(optValue, '-') {
```
jakebailey:
```suggestion
			if len(optValue) != 0 && !strings.HasPrefix(optValue, "-") {
```

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,389 @@
...
+			p.options[opt.Name] = nil
+			i++
+		}
+	}
+	return i
+}
+
+func (p *CommandLineParser) parseListTypeOption(opt *CommandLineOption, value string) []string {
+	value = strings.TrimFunc(value, stringutil.IsWhiteSpaceLike)
+	if stringutil.StartsWith(value, '-') {
```
jakebailey:
```suggestion
	if strings.HasPrefix(value, "-") {
```

internal/stringutil/util.go
```diff
@@ -75,3 +77,8 @@ func IsHexDigit(ch rune) bool {
 func IsASCIILetter(ch rune) bool {
 	return ch >= 'A' && ch <= 'Z' || ch >= 'a' && ch <= 'z'
 }
+
+func StartsWith(text string, ch rune) bool {
+	r, _ := utf8.DecodeRuneInString(text)
+	return r == ch
+}
```
jakebailey: I would just drop this and use `strings.HasPrefix` instead.

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,389 @@
...
+			args = append(args, string(curArg))
+		}
+	}
+	p.parseStrings(args)
+}
+
+func incrementStringByRune(text string, argSoFar []rune) (rest string, r rune, curArg []rune) {
+	r, size := utf8.DecodeRuneInString(text)
+	text = text[size:]
+	return text, r, append(argSoFar, r)
```
jakebailey: Rather than doing this custom append, could you just write `[]rune(text)` somewhere and iterate over it?

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,389 @@
...
+		return strings.ToLower(input[1:])
+	}
+	return strings.ToLower(input)
+}
+
+func (p *CommandLineParser) parseResponseFile(fileName string) {
+	text := ""
+	if p.readFile == nil {
+		text, _ = tryReadFile(fileName, func(fileName string) (string, error) {
+			read, err := os.ReadFile(fileName)
```
jakebailey: All of these direct `os` calls should get moved to explicit `vfs.FS` passing and path normalization so we can test them.

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+}
+
+// removes one '-' from the input string
+func getInputOptionName(input string) string {
+	char, _ := utf8.DecodeRuneInString(input)
+	if char == '-' {
+		return strings.ToLower(input[1:])
+	}
+	return strings.ToLower(input)
```
jakebailey: This could just be:

```suggestion
	return strings.ToLower(strings.TrimPrefix(input, "-"))
```

internal/tsoptions/commandLineParser.go
iisaduan: Ooh thanks, i forgot to update since that preceeded vfs

internal/tsoptions/commandLineParser.go
iisaduan: Iterating over a []rune should be fine to do. I wasn't sure if I should stick with string operations/if there is a convention so far that I should be consistent with

internal/tsoptions/commandLineParser.go
iisaduan: Is this why the tests are failing?

jakebailey: No, the other thing you need to do in tests is check if the submodule is even there, and skip the test if it's not.

iisaduan: https://github.com/microsoft/typescript-go/blob/6538ba4d4630f8310a38952d70bdedf40318ffd1/internal/compiler/parser_test.go#L48-L51 Do I check it like this?

jakebailey: Yeah, or similar. Let me send a PR which should make this a little easier to do generally speaking.

jakebailey: https://github.com/microsoft/typescript-go/pull/144

internal/tsoptions/commandLineParser.go
```diff
@@ -148,13 +148,9 @@ func (p *CommandLineParser) parseStrings(args []string) {
...
 
-// removes one '-' from the input string
+// removes leading '-' from the input string
 func getInputOptionName(input string) string {
-	char, _ := utf8.DecodeRuneInString(input)
-	if char == '-' {
-		return strings.ToLower(input[1:])
-	}
-	return strings.ToLower(input)
+	return strings.ToLower(strings.TrimLeft(input, "-"))
```
jakebailey: I think this will now make `--------blah` legal and so on, which may not be desirable.

internal/tsoptions/commandLineParser.go
jakebailey, line 1: These files are misnamed; no camelCase in Go files.

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+
+	return p.workerDiagnostics.optionsNameMap
+}
+
+type CommandLineParser struct {
+	workerDiagnostics *ParseCommandLineWorkerDiagnostics
+	fs                vfs.FS
+	options           OptionsBase
+	// todo: watchOptions   OptionsBase
+	readFile  *func(string) (string, bool)
```
jakebailey: This callback seems redundant if `fs` is passed, right?

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+
+func (p *CommandLineParser) parseStrings(args []string) {
+	i := 0
+	for i < len(args) {
+		s := args[i]
+		i++
+		if s == "" {
+			continue
+		}
+		char, _ := utf8.DecodeRuneInString(s)
```
jakebailey: Given we only check `@` and `-`, you actually don't need this at all; a single byte read will always be safe

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+		switch char {
```
jakebailey:
```suggestion
		switch s[0] {
```

internal/tsoptions/commandLineParser.go
iisaduan: oof yeah

jakebailey: You could just do:

```go
// Trim off at most two dashes
input = strings.TrimLeft(input, "-")
input = strings.TrimLeft(input, "-")
return strings.ToLower(input)
```

internal/tsoptions/commandLineParser.go
iisaduan: There's places in the ts version where a non-host version is passed in. It feels like it was for convenience reasons and I'm not sure how necessary it is. I think I'll remove readfile for now, and add it back in if there's something that needs a dfferent one for some reason

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+	text := []rune(fileContents)
+	textLength := len(text)
+	start := 0
+	pos := 0
+	for pos < textLength {
+		pos++
+		for pos < textLength && text[pos] < ' ' {
+			pos++
+		}
+		if pos < textLength {
```
DanielRosenwasser: This condition is inverted from https://github.com/microsoft/TypeScript/blob/421f5c5458d09dad601604e6eff8765283aef14e/src/compiler/commandLineParser.ts#L1962

So I think this stops parsing the entire file as soon as it hits any non-whitespace. Are all the tests for command-line parsing ported over? Are there any that specifically use response file parsing?

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+	text = text[size:]
+	return text, r, append(argSoFar, r)
+}
+
+func tryReadFile(fileName string, readFile func(string) (string, bool), errors []*ast.Diagnostic) (string, []*ast.Diagnostic) {
+	text, e := readFile(fileName)
+
+	if !e {
+		// no errror message from the standard error
+		// errors = append(errors, ast.NewDiagnostic(nil, core.NewTextRange(-1,-1), diagnostics.Cannot_read_file_0_Colon_1, *e));
```
DanielRosenwasser: If there's a divergence, 

```suggestion
		// !!! We don't have a useful message from the returned error
		// errors = append(errors, ast.NewDiagnostic(nil, core.NewTextRange(-1,-1), diagnostics.Cannot_read_file_0_Colon_1, *e));
```

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,393 @@
...
+		errors = append(errors, ast.NewDiagnostic(nil, core.NewTextRange(-1, -1), diagnostics.Cannot_read_file_0, fileName))
```
DanielRosenwasser: I think we should port over `createCompilerDiagnostic`.

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,388 @@
...
+			} else {
+				if optValue == "true" {
+					p.options[opt.Name] = true
+					i++
+				}
+				p.errors = append(p.errors, ast.NewDiagnostic(nil, p.errorLoc, diagnostics.Option_0_can_only_be_specified_in_tsconfig_json_file_or_set_to_false_or_null_on_command_line, opt.Name))
+			}
+		} else {
+			p.errors = append(p.errors, ast.NewDiagnostic(nil, p.errorLoc, diagnostics.Option_0_can_only_be_specified_in_tsconfig_json_file_or_set_to_null_on_command_line, opt.Name))
+			if len(optValue) != 0 && !strings.HasPrefix(optValue, "=") {
```
DanielRosenwasser: This must be a typo.

https://github.com/microsoft/TypeScript/blob/421f5c5458d09dad601604e6eff8765283aef14e/src/compiler/commandLineParser.ts#L2010

```suggestion
			if len(optValue) != 0 && !strings.HasPrefix(optValue, "-") {
```

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,388 @@
...
+				p.options[opt.Name] = true
+			}
+			return i
+		}
+		if args[i] != "null" {
+			switch opt.Kind {
+			case "number":
+				num, e := strconv.ParseInt(args[i], 10, 0)
+				if e == nil {
+					p.options[opt.Name] = num
```
DanielRosenwasser: Unrelated, but I'm sure the current compiler probably just rolls with `NaN`. Kind of surprised nobody has ever reported that?

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,388 @@
...
+				// boolean flag has optional value true, false, others
+				optValue := args[i]
+
+				// check next argument as boolean flag value
+				if optValue == "false" {
+					p.options[opt.Name] = false
+				} else {
+					p.options[opt.Name] = true
+				}
+				// do not consume argument if it is not "true" or "false"
```
DanielRosenwasser:
```suggestion
				// Try to consume the next argument as boolean flag value,
				// but do not consume argument if it is not "true" or "false".
```

internal/tsoptions/commandLineParser.go
```diff
@@ -0,0 +1,388 @@
...
+func (p *CommandLineParser) parseListTypeOption(opt *CommandLineOption, value string) []string {
+	value = strings.TrimFunc(value, stringutil.IsWhiteSpaceLike)
+	if strings.HasPrefix(value, "-") {
+		return []string{}
+	}
+	if opt.Kind == "listOrElement" && !strings.ContainsRune(value, ',') {
+		return []string{p.validateJsonOptionValue(opt, value, nil, nil)}
+	}
+	if value == "" {
+		return []string{}
```
DanielRosenwasser: I'm actually not sure if this should be represented as `nil` or a as a `0`-length slice here and above.

internal/tsoptions/commandLineParser.go
iisaduan: Nope, none that use any of the response file parsing :/ I commented out the entire function in the TS repo, and all the ts tests still pass.

There are still command-line related tests that arent ported, but they all need `executeCommandLine` or tsconfig parsing.

internal/tsoptions/commandlineparser_test.go
```diff
@@ -0,0 +1,323 @@
...
+			nonNullValue: "abc,xyz",
+		},
+		createVerifyNullForNonNullIncluded("option of type string", CommandLineOptionTypeString, "hello"),
+		createVerifyNullForNonNullIncluded("option of type number", CommandLineOptionTypeNumber, "10"),
+		// todo: make the following work for tests
+		// createVerifyNullForNonNullIncluded("option of type custom map", CommandLineOptionTypeEnum, "node"),
+	}
+
+	for _, verifyNullCase := range verifyNullSubScenarios {
+		t.Parallel()
```
jakebailey: What has to be marked parallel here is the func passed to Run; this func is already marked parallel above.

iisaduan: Hmmm, its telling me that the "range statement" is missing the call to parallel-- this means I need to put it into every call to `t.Run`, and not in the for loop?

jakebailey: You'd need it in createSubScenario.

iisaduan: `createSubScenario().assertParseResult` is the function that's actually called in `t.Run` and takes in a `t *testing.T`. I put in a `t.Parallel()` (and `t.Helper()`) in `assertParseResult`, but it doesn't seem to have changed it. I also tried to pass in `t` to `createSubSecnario` and the lint error still happens. 

Do I need to wrap the call to `assertParseResult` in an inline function parameter in `t.Run`? That feels unfortunate, since it feels like there should be a way to not have to write the function signature in line :P

internal/tsoptions/commandLineParser.go
iisaduan: https://github.com/microsoft/typescript-go/pull/145

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		return
+	}
+
+	var args []string
+	text := []rune(fileContents)
+	textLength := len(text)
+	start := 0
+	pos := 0
+	for pos < textLength {
+		pos++
```
jakebailey: Won't this skip the first character? The original code doesn't seem to have this line.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		for pos < textLength && text[pos] < ' ' {
+			pos++
+		}
+		if pos >= textLength {
+			break
+		}
+		if text[pos] == '"' {
```
jakebailey, line 169: Should start be defined above this line instead of at the top?

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+			pos++
+			for pos < textLength && text[pos] != '"' {
+				pos++
+			}
+			if pos < textLength {
+				args = append(args, string(text[start:pos]))
```
jakebailey:
```suggestion
				args = append(args, string(text[start+1:pos]))
```

The original code seems to have substring'd after start

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+				pos++
+				start = pos
```
jakebailey: This seems different than the previous code, which didn't set `start` here.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+			} else {
+				p.errors = append(p.errors, ast.NewCompilerDiagnostic(diagnostics.Unterminated_quoted_string_in_response_file_0, fileName))
+			}
+		} else {
+			for text[pos] > ' ' {
+				pos++
+			}
+			args = append(args, string(text[start:pos]))
+			start = pos
```
jakebailey: Same here.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+	}
+	p.parseStrings(args)
+}
+
+func incrementStringByRune(text string, argSoFar []rune) (rest string, r rune, curArg []rune) {
```
jakebailey: I think this function is unused.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+				}
+			} else {
+				p.options[opt.Name] = true
+			}
+			return i
+		}
+		if args[i] != "null" {
+			switch opt.Kind {
+			case "number":
+				num, e := strconv.ParseInt(args[i], 10, 0)
```
jakebailey, line 246: I was concerned at first because I wasn't sure if we did int only here, but the original code does do parseInt.

I would leave a `// !!! TODO: verify matches JS's parseInt` comment or something since Go's integer parsing may not be identical to JS's.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		if optValue == "null" {
+			p.options[opt.Name] = nil
+			i++
+		} else if opt.Kind == "boolean" {
+			if optValue == "false" {
+				p.options[opt.Name] = false
+				i++
+			} else {
+				if optValue == "true" {
+					p.options[opt.Name] = true
```
jakebailey: Hm, the original code doesn't seem to set true here, but just skips past it...

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+			}
+		} else {
+			p.errors = append(p.errors, ast.NewCompilerDiagnostic(diagnostics.Option_0_can_only_be_specified_in_tsconfig_json_file_or_set_to_null_on_command_line, opt.Name))
+			if len(optValue) != 0 && !strings.HasPrefix(optValue, "-") {
+				i++
+			}
+		}
+	} else {
+		// Check to see if no argument was provided (e.g. "--locale" is the last command-line argument).
+		if i >= len(args) {
```
jakebailey, line 231: The original code doesn't have this check; do we only work accidentally in the old code?

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		} else {
+			p.options[opt.Name] = nil
+			i++
+		}
+	}
+	return i
+}
+
+func (p *CommandLineParser) parseListTypeOption(opt *CommandLineOption, value string) []string {
+	value = strings.TrimFunc(value, stringutil.IsWhiteSpaceLike)
```
jakebailey:
```suggestion
	value = strings.TrimSpace(value)
```

This should be equivalent to `value.trim()` in JS.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+		return []string{p.validateJsonOptionValue(opt, value, nil, nil)}
+	}
+	if value == "" {
+		return []string{}
+	}
+	values := strings.Split(value, ",")
+	switch opt.Elements().Kind {
+	case "string":
+		return core.Filter(core.Map(values, func(v string) string { return p.validateJsonOptionValue(opt.Elements(), v, nil, nil) }), isDefined)
+	case "boolean", "object", "number":
```
jakebailey, line 304: The original code supports `number`; should that have a case like string above?

internal/tsoptions/commandlineparser.go
iisaduan: I had just assumed js just handeled it as an `undefined`. I added this (and a lot of other similar checks) because I ran into out of index/nil pointer access errors

jakebailey: Yeah, if the old code was okay with it going out of range, I can imagine that.

iisaduan: The old code was `if args[i] {}` 😄

internal/tsoptions/commandlineparser.go
iisaduan: probably leftover from an incorrect byte/rune/string conversion I tried to make

internal/tsoptions/commandlineparser.go
iisaduan: Moving the start down into the for loop fixes the reason why I put this extra `start = pos` here

internal/tsoptions/commandlineparser.go
iisaduan: we don't actually have any list options of number kind, only enum and string

internal/tsoptions/commandlineparser.go
iisaduan: Makes sense, I'll remove it. I thought about it more, and it doesn't actually matter what the parsed value is since if this error is hit, then we exit

internal/tsoptions/commandlineparser.go
```diff
@@ -88,14 +87,22 @@ type ParsedCommandLine struct {
 	CompileOnSave bool
 }
 
+func (p *ParsedCommandLine) GetConfigFileParsingDiagnostics() []*ast.Diagnostic {
+	if p.ConfigFile != nil {
+		// todo: !!! should be ConfigFile.ParseDiagnostics, check if they are the same
+		return append(p.ConfigFile.Diagnostics(), p.Errors...)
```
jakebailey: `append` modifies its input array; you probably want `slices.Concat`.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+	if fileContents == "" {
+		return
+	}
+
+	var args []string
+	text := []rune(fileContents)
+	textLength := len(text)
+	pos := 0
+	for pos < textLength {
+		for pos < textLength && text[pos] < ' ' {
```
jakebailey:
```suggestion
		for pos < textLength && text[pos] <= ' ' {
```

Original code had `<=`.

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+}
+
+func getInputOptionName(input string) string {
+	// removes at most two leading '-' from the input string
+	return strings.ToLower(strings.TrimLeft(strings.TrimLeft(input, "-"), "-"))
+}
+
+func (p *CommandLineParser) parseResponseFile(fileName string) {
+	fileContents, _ := tryReadFile(fileName, func(fileName string) (string, bool) {
```
jakebailey: This discards the error result; I think it should be assigning it back to `p.errors`.

internal/tsoptions/commandlineparser_test.go
```diff
@@ -0,0 +1,322 @@
+package tsoptions
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	json2 "github.com/go-json-experiment/json"
```
jakebailey: Is it strictly required to use `json2` here? It's entirely possible we won't actually use this package given its third-party-ness and instability.

internal/tsoptions/commandlineparser.go
jakebailey: (Are there tests for this code?)

iisaduan: nope, commenting it out on the ts repo doesn't fail any tests

internal/tsoptions/commandlineparser.go
```diff
@@ -0,0 +1,379 @@
...
+			for text[pos] > ' ' {
+				pos++
+			}
+			args = append(args, string(text[start:pos]))
+		}
+	}
+	p.parseStrings(args)
+}
+
+func TryReadFile(fileName string, readFile func(string) (string, bool), errors []*ast.Diagnostic) (string, []*ast.Diagnostic) {
```
jakebailey, line 190: Why was this exported?

iisaduan: Needed for executeCommandLine

jakebailey: Moderately weird, given tsoptions really has nothing to do with file reading....

iisaduan: I spent a bit of time looking at how `TryReadFile` is actually used, and the function seems to exist to consistently issue the same "file cannot be read" error(s) on failed file readings (there used to be more error handling in the original repo, that didn't translate well to go). I think it would be reasonable to define this elsewhere, like in `System`, so 
```go
func (s *System) ReadFileWithErrors (fileName string) (string, []*ast.Diagnostic) {
    // do what tryReadFile does with readFile = s.Host().FS().ReadFile
}
```

iisaduan: I was going to say that doing that would make the error checking for the command line parsing tests not be equal, but parsing response files isn't checked in those tests anyways 😅

jakebailey: We definitely would not want to have an FS method which also imports the ast; that's a backwardsish import direction. If we didn't have this extra package, it'd be in `compiler` and we'd just have it as a local helper, but I assume that's not possible with the current split.

(The func is short enough that if I were writing things, I would _probably_ just do a little copy and pasting of the diag code where i need it, rather than writing an abstraction)

iisaduan: For execute command line, I've created `System` to be a layer higher than `Host` that would also do things like `exit` and write to stdout (I'm assuming `CompilerHost` isn't supposed to have these, since they are os related). 

`System` isn't just the FS, so the above implementation would look something like this https://github.com/microsoft/TypeScript/blob/f69580f82146bebfb2bee8c7b8666af0e04c7e34/src/compiler/sys.ts#L1790,  so `System.TryReadFile` would call it's `fs.ReadFile`, and report the diagnostics directly to the commandline.

jakebailey: The main thing to note is that this code should not require a concrete implementation, but instead define an interface that has all of the methods being used, which allows callers to provide anything (mocks, real implementations, etc), but that implies that `TryReadFile` may not be defined in this package. It certainly _could_ be and accept an FS, but, it does seem more broadly useful.

PR #144, opened 2024-12-06
Add SkipIfNoTypeScriptSubmodule for easier test skipping without submodule

jakebailey:
Rather than repeating this code over and over.

### DIFF
--- a/internal/compiler/parser_test.go
+++ b/internal/compiler/parser_test.go
@@ -32,6 +32,7 @@ func BenchmarkParse(b *testing.B) {
 
 func TestParseTypeScriptRepo(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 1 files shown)


internal/repo/paths.go
```diff
@@ -31,3 +32,26 @@ func findGoMod(dir string) string {
...
+	if _, err := os.Stat(p); err != nil {
+		if os.IsNotExist(err) {
+			return false
+		}
+		panic(err)
+	}
+	return true
+})
+
+type skippable interface {
```
jakebailey, line 47: It's probably weird for this to be unexported, but one should really never be directly constructing these, only passing in `t`, and I did not want to make this package depend on `testing.TB` or `assert.T`.

PR #146, opened 2024-12-09
Make non-pretty printed diags use relative paths

jakebailey:
For demo purposes.


jakebailey:
Oh, non-pretty, no wonder I was confused

PR #148, opened 2024-12-09
Allow root files, and multiple roots to be passed at the command line.

DanielRosenwasser:
This change makes it possible again to pass an individual file to our entry point.

This change also makes it possible to pass in multiple roots (files or directories) to our entry point, making it possible to write something like `./bin/tsgo ./test1.ts ./test2.ts`.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -13,7 +13,7 @@ import (
 )
 
 type ProgramOptions struct {
-	RootPath       string
+	RootPaths      []string
 	Host           CompilerHost
 	Options        *core.CompilerOptions
 	SingleThreaded bool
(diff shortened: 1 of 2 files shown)


internal/compiler/program.go
```diff
@@ -43,15 +43,38 @@ func NewProgram(options ProgramOptions) *Program {
...
+
+	if len(options.RootPaths) == 0 {
+		panic("root paths required")
+	}
+
+	// !!!
+	// This isn't really a concept, but this is just
+	// to preserve existing functionality around node_modules lookups.
+	if len(options.RootPaths) == 1 {
+		p.rootPath = options.RootPaths[0]
```
DanielRosenwasser, line 55: Probably need to verify that this is actually a directory.

internal/compiler/program.go
```diff
@@ -43,15 +43,38 @@ func NewProgram(options ProgramOptions) *Program {
...
+	} else {
+		p.rootPath = p.host.GetCurrentDirectory()
+	}
```
iisaduan, line 58: Is this the reason why `rootPath` is being changed to `rootPaths` in `ProgramOptions` and not yet in `Program` as well? https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/compiler/program.ts#L1965

internal/compiler/program.go
```diff
@@ -43,15 +43,38 @@ func NewProgram(options ProgramOptions) *Program {
...
 	slices.SortFunc(fileInfos, func(a FileInfo, b FileInfo) int {
-		return int(b.Size) - int(a.Size)
+		sizeDiff := int(b.Size) - int(a.Size)
+		if sizeDiff != 0 {
+			return sizeDiff
+		}
+		return strings.Compare(b.Name, a.Name)
 	})
+	// Make sure that we don't duplicate files.
+	fileInfos = slices.Compact(fileInfos)
```
jakebailey, line 76: Technically, this will be wrong if you're on Windows or macOS and pass in like `tsgo ~/work/typescript ~/work/TYPESCRIPT`, but it's not clear it much matters in this instance

internal/compiler/program.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type ProgramOptions struct {
-	RootPath       string
+	RootPaths      []string
```
jakebailey, line 16: I feel like it'd be smart for us to rename this to indicate (for now) that it has nothing to do with a "root path", but is instead a list of input files/directories.

internal/compiler/program.go
DanielRosenwasser: It's mostly because `rootPath` is kind of a sketchy concept that I think will eventually go away. There really isn't any notion of a single `rootPath` - just the current directory, and the root set of files.

jakebailey:
FWIW this is now conflicting; I do want to get this change in because it'd be useful, though.

iisaduan:
It would be nice to have this in before I open a PR for execute command line. Should we change the name to something like `InputPaths` instead of `RootPaths`?

jakebailey:
I'm going to close this now; my PR that wires up a more tsc-y CLI is now merged.

PR #149, opened 2024-12-09
Add toplevel await

sandersn:
1. No incremental reparsing, although the non-incremental reparsing only reparses the statements that might contain await.
2. No TransformFlags, just an approximation in the parser.

I haven't added all the code to turn off PossibleTopLevelAwait because I'd like some agreement that this approach is good enough. I made enough changes to pass our test suite, in ClassDeclaration and ImportDeclaration parsing, but there are 16-17 more places to save/overwrite the toplevel await tracking bool. 

That means in the current commit, for example, a toplevel FunctionDeclaration that contains `await` somewhere inside will reparse the FunctionDeclaration, even though the old compiler would not. It's possible that reparsing this doesn't change anything anyway, but it's inefficient to reparse it when it's not needed.


internal/compiler/parser.go
```diff
@@ -45,23 +46,40 @@ const (
 type ParsingContexts int
 
 type Parser struct {
-	scanner               *scanner.Scanner
```
jakebailey: We should really break these up into groups or something so we stop moving them by like a couple characters each time we add something

internal/compiler/parser.go
```diff
@@ -220,20 +246,154 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+	statement := p.parseStatement()
+	if p.hasAwaitIdentifier {
+		setPossibleAwait(statement)
+		p.hasPossibleAwaitStatement = true
+	}
+	return statement
+}
+
+func (p *Parser) reparseTopLevelAwait(sourceFile *ast.SourceFile) *ast.Node {
+	statements := []*ast.Statement{}
```
sandersn: this part is a simplified port from tsc. It's simplified because it drops the attempt to use the incremental parser (which doesn't exist in tsgo).

internal/compiler/parser.go
```diff
@@ -45,23 +46,40 @@ const (
...
+	hasAwaitIdentifier        bool
+	hasPossibleAwaitStatement bool
+}
+
+var (
+	possibleAwaitCache   map[*ast.Node]bool = make(map[*ast.Node]bool)
+	possibleAwaitCacheMu sync.Mutex
+)
 
+func setPossibleAwait(node *ast.Node) {
```
jakebailey: I think this will end up OOMing us because these nodes will never be cleaned up, and there isn't a method I can think of that would allow this to be global except weak maps (which Go does not yet have).

Can we store this somewhere that isn't global? It seems like this can be a prop on the parser?

internal/compiler/parser.go
jakebailey: (Not for this PR but a general comment given GitHub does a very bad job with this while other tools are okay)

sandersn: TIL

I shuffled them into 4 categories, roughly
- factories
- public, immutable properties
- dynamic variables
- caches

token could arguably go with factories, but that's what happens any time you start re-arranging things into little piles.

sandersn: (Anybody can send a followup PR to put them in better categories; I just did this quickly.)

internal/compiler/parser.go
```diff
@@ -1318,6 +1479,7 @@ func (p *Parser) parseClassExpression() *ast.Node {
 
 func (p *Parser) parseClassDeclarationOrExpression(pos int, hasJSDoc bool, modifiers *ast.ModifierList, kind ast.Kind) *ast.Node {
 	saveContextFlags := p.contextFlags
+	saveHasAwaitIdentifier := p.statementHasAwaitIdentifier
```
sandersn: these are all the save/restores that approximate TransformFlags propagation in tsc.

internal/compiler/parser.go
```diff
@@ -45,23 +46,40 @@ const (
...
+	parsingContexts               ParsingContexts
+	statementHasAwaitIdentifier   bool
+	fileHasPossibleAwaitStatement bool
+
 	identifiers           core.Set[string]
-	sourceFlags           ast.NodeFlags
 	notParenthesizedArrow core.Set[int]
 	nodeSlicePool         core.Pool[*ast.Node]
+	possibleAwaitCache    map[*ast.Node]bool
+	possibleAwaitCacheMu  sync.Mutex
```
rbuckton: Are we sharing this cache across threads *during* parse? Parsing a file is single-threaded, and the parser cannot be used by multiple threads at once. Locking and unlocking a mutex can introduce unnecessary contention for reads and writes. 

If this data is only used by a single `Parser` instance during parse of a single file to determine whether it is necessary to reparse TLA, there really isn't any reason for it to be shared. If this data needs to be used elsewhere by multiple threads, I suggest we instead maintain a cache unique to the `Parser` instance that is single threaded and then just merge all of the resulting caches when we are finished parsing.

internal/compiler/parser.go
```diff
@@ -220,20 +247,154 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+		if diagnosticStart >= 0 {
+			p.diagnostics = append(p.diagnostics, savedParseDiagnostics[diagnosticStart:]...)
+		}
+	}
+
+	return p.factory.NewSourceFile(sourceFile.Text, sourceFile.FileName(), p.factory.NewNodeList(sourceFile.Statements.Loc, statements))
+}
+
+func (p *Parser) containsPossibleTopLevelAwait(node *ast.Node) bool {
+	return !(node.Flags&ast.NodeFlagsAwaitContext != 0) && p.getPossibleAwait(node)
```
rbuckton:
```suggestion
	return node.Flags&ast.NodeFlagsAwaitContext == 0 && p.getPossibleAwait(node)
```

internal/compiler/parser.go
sandersn: Nope, just used during parsing. It took until a 3rd re-reading after lunch to notice that `notParenthesizedArrow`, for example, works exactly this way. I think possibleAwaitCache should be an unguarded `core.Set[*ast.Node]`.

jakebailey: Parsing itself is synchronous, so I do not think this needs a mutex at all.

internal/compiler/parser.go
```diff
@@ -45,21 +45,27 @@ const (
...
+	sourceFlags                   ast.NodeFlags
+	contextFlags                  ast.NodeFlags
+	parsingContexts               ParsingContexts
+	statementHasAwaitIdentifier   bool
+	fileHasPossibleAwaitStatement bool
+
+	identifiers            core.Set[string]
+	notParenthesizedArrow  core.Set[int]
+	nodeSlicePool          core.Pool[*ast.Node]
+	possibleAwaitStatement core.Set[*ast.Node]
```
jakebailey: Once parsing is done, we don't need this set, right? So can we clear it out / overwrite it with a zero value to regain that memory?

internal/compiler/parser.go
sandersn: Ugh, yes, I saw that as I pasted it and then forgot to go fix it.

internal/compiler/parser.go
```diff
@@ -254,6 +254,7 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 			result.SetDiagnostics(attachFileToDiagnostics(p.diagnostics, result))
 			result.ExternalModuleIndicator = isFileProbablyExternalModule(result)
 			result.IsDeclarationFile = isDeclarationFile
+			p.possibleAwaitStatement = core.Set[*ast.Node]{}
```
jakebailey: Should this be unconditional? Otherwise we'll still have the info left behind when we don't execute this block, I think.

sandersn: Yep, I was so proud that I remembered the init syntax that I confused myself into thinking `Add` was called during the reparse. But it's called during the main parse. I'll move it to the end of the function.

(I fiddled with `defer` but there's no logical place for it and it only works on function calls anyway.)

jakebailey: You would just have to `defer func() { p.possibleAwaitStatement = core.Set[*ast.Node]{} }()`.

internal/parser/parser.go
```diff
@@ -239,9 +253,143 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 	result.SetDiagnostics(attachFileToDiagnostics(p.diagnostics, result))
 	result.ExternalModuleIndicator = isFileProbablyExternalModule(result)
 	result.IsDeclarationFile = isDeclarationFile
+	if !result.IsDeclarationFile && result.ExternalModuleIndicator != nil && p.fileHasPossibleAwaitStatement {
```
DanielRosenwasser: It really does feel like a call with an `await` is actually less common, and we make every top-level `await` user go through an extra reparse. Ideally, we would only reparse these on *non-module* files.

sandersn: Why switch to reparsing non-modules when the old compiler only reparses modules?
Regardless, I want to make this a straight port and not worry about semantic differences. 

If/when we do make changes, I'd start with Anders' idea of "always parse top-level await as keyword". Usage of `await` as an identifier is super low in reality (I counted 4 in my npm corpus) and likely to be a mistake that people will be happy to have pointed out.

rbuckton: > Ideally, we would only reparse these on _non-module_ files.

The problem is that we usually don't know if these are modules or not until we've parsed the whole file, since we base it on the presence of syntax rather than on a goal symbol. We could probably avoid reparse for `.mjs`/`.mts` since we know those *definitely* support TLA, and avoid reparse for `.cjs`/`.mjs` because those *definitely don't* support TLA, but any `.js`/`.ts` is fair game.

DanielRosenwasser: I think it's fine to say "we're going to keep this as a straight port".

internal/parser/parser.go
```diff
@@ -242,9 +256,146 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+	p.diagnostics = []*ast.Diagnostic{}
+
+	pos := 0
+	start := p.findNextStatementWithAwait(sourceFile.Statements, 0)
+	for start != -1 {
+		// append all statements between pos and start
+		prevStatement := sourceFile.Statements.Nodes[pos]
+		nextStatement := sourceFile.Statements.Nodes[start]
+		statements = append(statements, sourceFile.Statements.Nodes[pos:start]...)
+		pos = p.findNextStatementWithoutAwait(sourceFile.Statements, start)
```
DanielRosenwasser: This isn't used until way way lower. Can you move it down?

DanielRosenwasser: Also possibly rename this with something better than `pos`?

internal/parser/parser.go
```diff
@@ -242,9 +256,146 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+	return p.factory.NewSourceFile(sourceFile.Text, sourceFile.FileName(), p.factory.NewNodeList(sourceFile.Statements.Loc, statements))
+}
+
+func (p *Parser) containsPossibleTopLevelAwait(node *ast.Node) bool {
+	return node.Flags&ast.NodeFlagsAwaitContext == 0 && p.possibleAwaitStatement.Has(node)
+}
+
+func (p *Parser) findNextStatementWithAwait(statements *ast.NodeList, start int) int {
+	for i, statement := range statements.Nodes[start:] {
+		if p.containsPossibleTopLevelAwait(statement) {
```
DanielRosenwasser: Just to avoid crazy worst-case scearios: Instead of running across every statement in the file and seeing if it's in the set, I'd rather we iterated over every possible top level statement (probably sorted), and get the index via a binary search. You can even run it on a smaller and smaller slice of the original statement list like you're doing here.

DanielRosenwasser:
> That means in the current commit, for example, a toplevel FunctionDeclaration that contains `await` somewhere inside will reparse the FunctionDeclaration, even though the old compiler would not. It's possible that reparsing this doesn't change anything anyway, but it's inefficient to reparse it when it's not needed.

That might actually have some effects on things. What errors do you issue on the following cases?

```ts
// @filename: a.ts
function foo(x = await(10)) {
}

// @filename: b.ts
async function bar(x = await(10)) {
}

// @filename: c.ts
export {};
export function fooExported(x = await(10)) {
}

// @filename: d.ts
export {};
export async function barExported(x = await(10)) {
}
```

Weirdly, we don't have any test case where `await` in a parameter initializer is syntactically ambiguous as both an *AwaitExpression* and as a *CallExpression* like in these examples. Would you be able to add it?

internal/parser/parser.go
rbuckton: > Just to avoid crazy worst-case scearios: Instead of running across every statement in the file and seeing if it's in the set, I'd rather we iterated over every possible top level statement (probably sorted), and get the index via a binary search. You can even run it on a smaller and smaller slice of the original statement list like you're doing here.

We could keep track of the index when calling `parseTopLevelStatement` and store it in a slice instead of storing the statement itself in a `core.Set`.

sandersn:
The new commit simplifies tracking a lot compared to Strada. Instead of tracking await-possible statements and searching to
reconstruct spans, it builds spans during parsing. Then reparsing can read the spans directly, which simplifies code quite a bit.

The downside is that I need to provide indices in parseList. Because Go isn't Javascript, I decided to copy parseList to parseListIndex. Two other options: change parseList to always provide indices then

- Add ignored indices to parseStatement, parseHeritageClause et al.
- Add a parseList that passes an adapter `func` which ignores indices.

If adapter funcs aren't optimised away, I'd say the current code is best. I tried adding `_ int` parameters to parseStatement, et al and didn't like it at all--normal calls to parseStatement now need to pass `-1` and there are lots of those calls.

I've tested on our current test cases but I need to try Daniel's cases he suggested on this PR.

jakebailey:
> If adapter funcs aren't optimised away, I'd say the current code is best. I tried adding `_ int` parameters to parseStatement, et al and didn't like it at all--normal calls to parseStatement now need to pass `-1` and there are lots of those calls.

You can largely assume that simple one-liners that aren't generic are inlineable. If you wanted to double check, there are flags that you can pass to the compiler that would show that information.

rbuckton:
> The downside is that I need to provide indices in parseList.

Since `parseTopLevelStatement` is only ever called for statements at the top of a `SourceFile`, we could just have a `topLevelStatementIndex` field that we increment in `parseTopLevelStatement`, and reset it to `0` if we reuse a `Parser`

rbuckton:
> Weirdly, we don't have any test case where `await` in a parameter initializer is syntactically ambiguous as both an _AwaitExpression_ and as a _CallExpression_ like in these examples. Would you be able to add it?

Not Weirdly. `await` cannot be an AwaitExpression in a parameter initializer. Per https://tc39.es/ecma262/#sec-async-function-definitions-static-semantics-early-errors, it is a Syntax Error if the parameters contain _AwaitExpression_.

sandersn:
@DanielRosenwasser re your test cases:

a: no (parse) errors, `await` is legal in a non-module (no reparsing either).
b. Same.
c. no (parse) errors, there is a reparse but it doesn't change anything. Not sure if this is legal in a module.
d. Same.

@rbuckton Are (c) and (d) correct? `await(10)` is parseable as a function call even in a module, but should it be parsed that way?

rbuckton:
> @rbuckton Are (c) and (d) correct? `await(10)` is parseable as a function call even in a module, but should it be parsed that way?

`await` will always be parsed in the `Await` context of the function:

(c) should be an error because a module is always parsed in strict mode, and `await` is illegal as an identifier in strict mode, per https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors, though we could choose to parse it as an _Identifier_ and issue a grammar error. 

(d) should be an error because it will be parsed as an _AwaitExpression_, but would be a syntax error per https://tc39.es/ecma262/#sec-async-function-definitions-static-semantics-early-errors, which we could also treat as a grammar error.

There are two other scenarios to consider:

```ts
// @filename: e.ts
function fooStrict(x = await(10)) {
  "use strict"
}

// @filename: f.ts
(function () {
  "use strict"
  function fooStrict(x = await(10)) {
  }
})()
```
These should be grammar errors because `await` is parsed as an _Identifier_, but fails https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors since the function is strict-mode code.

Please note that V8 seems to have a bug here, as it incorrectly allows `await` as an identifier in strict-mode code, in violation of https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors, though it correctly forbids `yield` as an identifier in strict mode per the same rule.

rbuckton:
> These should be grammar errors because `await` is parsed as an _Identifier_, but fails https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors since the function is strict-mode code.

Sorry, this was an incorrect interpretation of 13.1.1. 13.1.1 only disallows `await` if the goal symbol is _Module_.

To clarify:

- (a) should parse as a call expression for the id `await`
- (b) should parse as an AwaitExpression, but error due to [15.8.1](https://tc39.es/ecma262/#sec-async-function-definitions-static-semantics-early-errors)
- (c) should parse as a call expression for the id `await`, but error due to [13.1.1](https://tc39.es/ecma262/#sec-identifiers-static-semantics-early-errors) because we are parsing a _Module_.
- (d) should parse as an AwaitExpression, but error due to [15.8.1](https://tc39.es/ecma262/#sec-async-function-definitions-static-semantics-early-errors)
- (e) and (f) should be ignored as they were based on a faulty interpretation of 13.1.1

sandersn:
Ah, I see the difference. a,b,c,d,e,f all give *grammar* errors after parsing correctly. I think the PR's code is working the same as the old code -- by the end of parsing it produces exactly the same trees and exactly the same errors (none).

rbuckton:
> Ah, I see the difference. a,b,c,d,e,f all give _grammar_ errors after parsing correctly. I think the PR's code is working the same as the old code -- by the end of parsing it produces exactly the same trees and exactly the same errors (none).

I take it the grammar errors for this are not yet ported in the checker, or are they missing/broken in both compilers?

sandersn:
The grammar errors are present and correct in Strada.
Errors for (b) and (d) and two of (e)'s are ported. But this PR ports just the parser code.

internal/parser/parser.go
```diff
@@ -1787,13 +1933,15 @@ func (p *Parser) parseModuleOrNamespaceDeclaration(pos int, hasJSDoc bool, modif
...
 	_ = hasJSDoc
+	p.statementHasAwaitIdentifier = saveHasAwaitIdentifier
 	return result
 }
 
 func (p *Parser) parseImportDeclarationOrImportEqualsDeclaration(pos int, hasJSDoc bool, modifiers *ast.ModifierList) *ast.Statement {
 	p.parseExpected(ast.KindImportKeyword)
 	afterImportPos := p.nodePos()
 	// We don't parse the identifier here in await context, instead we will report a grammar error in the checker.
+	saveHasAwaitIdentifier := p.statementHasAwaitIdentifier
```
jakebailey, line 1944: I wish we didn't have all of this manual saving/restoring in favor of some sort of defer but...

internal/parser/parser.go
```diff
@@ -247,17 +258,136 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+			result = reparse.AsSourceFile()
+			result.SetDiagnostics(attachFileToDiagnostics(p.diagnostics, result))
+			result.ExternalModuleIndicator = isFileProbablyExternalModule(result)
+			result.IsDeclarationFile = isDeclarationFile
+			result.LanguageVersion = p.languageVersion
+			result.LanguageVariant = p.languageVariant
+			result.ScriptKind = p.scriptKind
+		}
+	}
+	p.possibleAwaitSpans = []int{}
```
DanielRosenwasser, line 274: Rather than an `int` array, you could make this a `struct {}` but I guess it's fine.

PR #150, opened 2024-12-10
Fix import attributes parsing

sandersn:
Just a flipped boolean.


sandersn:
Fixed 4 hours ago by #97

PR #151, opened 2024-12-10
Make module.Resolver concurrency safe, test

jakebailey:
The current module resolver races on its caches; in #107 we're going to need it to be concurrency safe.

This PR applies a big mutex on the resolver, and ensures any results it returns are cloned. This is not the most ideal solution as we're basically saying that only one resolution can happen at a time, but making the code more granularly lock caches is a harder change and I'd like #107 to be unblocked.

All tests pass without the fix, even with concurrent checks, but running with `-race` shows the problem (and CI does that).

### DIFF
--- a/internal/compiler/module/resolver.go
+++ b/internal/compiler/module/resolver.go
@@ -199,6 +199,11 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
+	result := r.resolveModuleName(moduleName, containingFile, resolutionMode, redirectedReference)
+	return result.clone()
+}
+
+func (r *Resolver) resolveModuleName(moduleName string, containingFile string, resolutionMode core.ResolutionMode, redirectedReference *ResolvedProjectReference) *ResolvedModuleWithFailedLookupLocations {
 	traceEnabled := r.traceEnabled()
 
 	compilerOptions := r.compilerOptions
@@ -261,7 +266,7 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
 		}
 	}
 
-	return result.clone()
+	return result
 }
 
 func (r *Resolver) traceTypeReferenceDirectiveResult(typeReferenceDirectiveName string, result *ResolvedTypeReferenceDirectiveWithFailedLookupLocations) {


internal/compiler/module/resolver_test.go
```diff
@@ -343,6 +345,66 @@ func sanitizeTraceOutput(trace string) string {
...
+				CommandLine: module.ParsedCommandLine{
+					Options: call.args.RedirectedRef.CommandLine.Options,
+				},
+			}
+		}
+
+		var locations module.WithFailedLookupLocations
+		if call.call == "resolveModuleName" {
+			resolved := resolver.ResolveModuleName(call.args.Name, call.args.ContainingFile, core.ModuleKind(call.args.ResolutionMode), redirectedReference)
+			assert.Check(t, resolved != nil, "ResolveModuleName should not return nil")
```
jakebailey, line 364: An invariant of Go's testing package is that `.FailNow()` and so on are not called from the "wrong" goroutine; it's okay for different goroutines to use things like `Errorf`, which is what `Check` will do.

internal/compiler/module/resolver_test.go
```diff
@@ -343,6 +345,66 @@ func sanitizeTraceOutput(trace string) string {
...
+			if expectedResolvedTypeReferenceDirective, ok := call.returnValue["resolvedTypeReferenceDirective"].(map[string]any); ok {
+				assert.Check(t, resolved.IsResolved())
+				assert.Check(t, cmp.Equal(resolved.ResolvedTypeReferenceDirective.ResolvedFileName, expectedResolvedTypeReferenceDirective["resolvedFileName"].(string)))
+				assert.Check(t, cmp.Equal(resolved.ResolvedTypeReferenceDirective.Primary, expectedResolvedTypeReferenceDirective["primary"].(bool)))
+				assert.Check(t, cmp.Equal(resolved.ResolvedTypeReferenceDirective.IsExternalLibraryImport, expectedResolvedTypeReferenceDirective["isExternalLibraryImport"].(bool)))
+			} else {
+				assert.Check(t, !resolved.IsResolved())
+			}
+		}
+		if skipLocations {
```
jakebailey, line 388: This skips checking the contents of the locations since those are resolution order dependent, but because we read the value above, we still correctly trigger the races.

internal/compiler/module/resolver.go
```diff
@@ -245,7 +256,8 @@ func (r *Resolver) ResolveModuleName(moduleName string, containingFile string, r
 		}
 	}
 
-	return result
+	cloned := *result
+	return &cloned
```
jakebailey: This may end up cloning twice; I'm going to instead see if I can tack on a boolean field to the lookup locations struct to assert that it was in fact a cloned copy or something...

RyanCavanaugh:
Notably in https://github.com/microsoft/TypeScript/issues/60556, it has happened that module resolution order matters, so we're potentially opening up nondeterministic builds

jakebailey:
That effectively implies we can't do any sort of fully concurrent program load; the only way we could be deterministic is to instead make program load do batching over the input files to parse, then do module resolution on them in order synchronously, then repeat.

jakebailey:
I like #156 more

PR #152, opened 2024-12-10
Speed up CI with more jobs

jakebailey:
This should give test results faster, and avoid running race mode on a slower builder like Windows.


.github/workflows/ci.yml
```diff
@@ -39,23 +39,27 @@ jobs:
           - ubuntu-latest
           - windows-latest
           - macos-latest
-        submodules:
-          - true
+        no-submodules:
```
jakebailey, line 42: Flipping this so that the default value used in the `include` block is falsy

PR #153, opened 2024-12-10
Shebang error parsing: compatibility with tsc

sandersn:
In tsc, `#!` anywhere except pos=0 in a file gives an error and increments past `#`. In tsgo, it was incrementing past `#!`. tsc's behaviour seems bad, but not bad enough to differ from, especially since the parse tree will be wrong either way.

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -707,8 +707,8 @@ func (s *Scanner) Scan() ast.Kind {
 			if s.charAt(1) == '!' {
 				if s.pos == 0 {
 					s.pos += 2
-					for s.char() >= 0 && !stringutil.IsLineBreak(s.char()) {
-						s.pos++
+					for ch, size := s.charAndSize(); size > 0 && !stringutil.IsLineBreak(ch); ch, size = s.charAndSize() {
+						s.pos += size
 					}
 					continue
 				}


jakebailey:
The original code was:

```ts
case CharacterCodes.hash:
    if (pos !== 0 && text[pos + 1] === "!") {
        error(Diagnostics.can_only_be_used_at_the_start_of_a_file, pos, 2);
        pos++;
        return token = SyntaxKind.Unknown;
     }
```

What is all of that logic above the code you're modifying?

sandersn:
The original code checked for valid `#!` separately, at the beginning of the function. Since shebang is rare, and `#` is rare, it makes more sense to move and inline the code for the valid case. So in tsgo it sits right next to the invalid case.

(This is speculation, maybe @ahejlsberg had some other reason for inlining it.)

DanielRosenwasser:
It never made sense to check specifically at the beginning of the function if you want to give a good error message anyway. I did the same thing at https://github.com/microsoft/TypeScript/pull/59244

DanielRosenwasser:
Is it just me or does that inlined code not work for a shebang terminated by any newline character other than `\n`?

sandersn:
Shebang only works in Unix doncha know 🙃 
In JS, it's a regex, so whatever a regex considers a newline. I'll switch it to stringutils.IsLineBreak.

internal/scanner/scanner.go
```diff
@@ -703,13 +703,13 @@ func (s *Scanner) Scan() ast.Kind {
 			if s.charAt(1) == '!' {
 				if s.pos == 0 {
 					s.pos += 2
-					for s.char() >= 0 && s.char() != '\n' {
+					for s.char() >= 0 && !stringutil.IsLineBreak(s.char()) {
 						s.pos++
```
jakebailey: `IsLineBreak` returns true for chars of non-1 width, so I think this will break.

jakebailey: Er, well, it's a negated check, so I'm wrong, but can we be sure that `++` is okay here?

sandersn: It's definitely suspicious. Let me try to rework this to account for non-1 linebreaks.

PR #154, opened 2024-12-10
Fix super type argument parsing

sandersn:
Change an inner declaration `:=` to an assignment `=`.

The shortcut syntax's error proneness should have been caught before it shipped. It really sucks!


jakebailey:
> The shortcut syntax's error proneness should have been caught before it shipped. It really sucks!

It's really only worse because `:=` looks like `=`, whereas you can do this exact same bug in JS but that requires a `let`/`var`.

(I am personally used to noticing `:=`...)

PR #156, opened 2024-12-10
Make module.Resolver concurrency safe, and concurrent

andrewbranch:
This is an alternative to #151, which works primarily by moving lookup locations off results, since those are what get mutated after being returned from the cache, but they're only used for file watching, which isn't even implemented yet and doesn't impact normal `tsc` runs. This allows us to eliminate the result cloning. Locking is still needed around cache read/write, but I made it more granular to allow more resolution to happen in parallel.

### DIFF
--- a/internal/compiler/module/cache.go
+++ b/internal/compiler/module/cache.go
@@ -1,8 +1,11 @@
 package module
 
 import (
+	"slices"
 	"strings"
+	"sync"
 
+	"github.com/microsoft/typescript-go/internal/ast"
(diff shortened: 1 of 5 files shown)


jakebailey:
main vs Eric's PR + my PR vs Eric's PR + your PR

image

EricCornelson:
Just wanted to note here that I also see a bit of a slowdown when running with -p, but not otherwise, which is interesting.

image

jakebailey:
Within the margin of error, both seem to have the same absolute time difference, which makes sense given all flags load the program.

Issue #140, opened 2024-12-06
Incorrect object rest syntax parses differently than tsc


sandersn:
`const { ...rest: notAllowed } = x` puts the DotDotDotToken after `rest`, not `before`.

See  _submodules/TypeScript/tests/cases/conformance/salsa/plainJSGrammarErrors.ts

I'm not sure if this is an error in the visitor or the parser.

DanielRosenwasser:
Fixed by #157.

PR #157, opened 2024-12-10
Fix BindingElement visit order

sandersn:
`...` should always be first. This is only visible when both PropertyName and DotDotDotToken are defined, which is a parse error&mdash;but the visit will be incorrect in that case.


DanielRosenwasser:
Fixes #140.

PR #158, opened 2024-12-10
Add `getSymbolAtLocation`

gabritto:


internal/compiler/checker_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+	"testing/fstest"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/vfs/vfstest"
+)
+
+func TestGetSymbolAtLocation(t *testing.T) {
```
jakebailey, line 11: Nice, our first compiler test 😄

internal/compiler/checker.go
```diff
@@ -399,205 +399,209 @@ const (
 // Checker
 
 type Checker struct {
-	program                            *Program
```
jakebailey: We should really go and split these into groups so this entire thing isn't redone... This is the worst thing about having too many fields

internal/compiler/checker.go
```diff
@@ -17984,3 +18022,498 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+		return nil
+	}
+
+	if isExpressionNode(name) {
+		if ast.NodeIsMissing(name) {
+			// Missing entity name.
+			return nil
+		}
+
+		isJSDoc := ast.FindAncestor(name, core.Or(isJSDocLinkLike, ast.IsJSDocNameReference))
```
jakebailey: It seems like this could just be a plain function closure rather than `core.Or`, (which will actually be faster too)

I think we had been avoiding adding that on purpose.

gabritto: For performance reasons? I do feel like `core.Or(...)` is easier to read than `func (...) bool { return ... }`, but if we're avoiding it I can remove it.

jakebailey: Well, for one to be consistent, but a static function literal will be inlined and so on which is usually better.

jakebailey:
There's a lot of stuff here; I think I'd probably like @ahejlsberg to be the main one looking at this to make sure it's not duplicating something he's already done.

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+}
+
+func (c *Checker) getSymbolOfNameOrPropertyAccessExpression(name *ast.Node) *ast.Symbol {
+	if ast.IsDeclarationName(name) {
+		return c.getSymbolOfNode(name.Parent)
+	}
+
+	if ast.IsInJSFile(name) &&
+		name.Parent.Kind == ast.KindPropertyAccessExpression &&
+		name.Parent.Parent.Kind == ast.KindBinaryExpression &&
```
jakebailey: This line doesn't seem to map to the original code; is this a bugfix or something else?

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+		if name.Kind == ast.KindIdentifier {
+			if ast.IsJsxTagName(name) && isJsxIntrinsicTagName(name) {
+				symbol := c.getIntrinsicTagSymbol(name.Parent)
+				return core.IfElse(symbol == c.unknownSymbol, nil, symbol)
+			}
+			result := c.resolveEntityName(
+				name,
+				meaning,
+				true, /*ignoreErrors*/
+				true, /*dontResolveAlias*/
```
```suggestion
				true /*ignoreErrors*/,
				true /*dontResolveAlias*/,
```

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+			} else {
+				c.checkQualifiedName(name, CheckModeNormal)
+			}
+
+			if links.resolvedSymbol == nil && isJSDoc != nil && ast.IsQualifiedName(name) {
+				return c.resolveJSDocMemberName(name, false /*ignoreErrors*/, nil /*container*/)
+			}
+			return links.resolvedSymbol
+		} else if ast.IsJSDocNameReference(name) {
+			return c.resolveJSDocMemberName(name, false /*ignoreErrors*/, nil /*container*/)
```
jakebailey: The original code said `isJSDocMemberName` here; are they different constructs?

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+		return c.getUnresolvedSymbolForEntityName(name)
+	}
+
+	if name.Parent.Kind == ast.KindTypePredicate {
+		return c.resolveEntityName(
+			name,
+			ast.SymbolFlagsFunctionScopedVariable, /*meaning*/
+			false,                                 /*ignoreErrors*/
+			false,                                 /*dontResolveAlias*/
+			nil,                                   /*location*/
```
jakebailey, line 20241:
```suggestion
			ast.SymbolFlagsFunctionScopedVariable /*meaning*/,
			false                                 /*ignoreErrors*/,
			false                                 /*dontResolveAlias*/,
			nil                                   /*location*/,
```

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+func (c *Checker) getApplicableIndexSymbol(t *Type, keyType *Type) *ast.Symbol {
+	infos := c.getApplicableIndexInfos(t, keyType)
+	if len(infos) > 0 && t.AsObjectType().members != nil {
+		symbol := getIndexSymbolFromSymbolTable(c.resolveStructuredTypeMembers(t).members)
+		if core.Same(infos, c.getIndexInfosOfType(t)) {
+			return symbol
+		} else if symbol != nil {
+			indexSymbolLinks := c.indexSymbolLinks.get(symbol)
+			declarationList := core.MapNonNil(infos, func(info *IndexInfo) *ast.Node { return info.declaration })
+			nodeListId := strings.Join(core.Map(declarationList, func(n *ast.Node) string { return strconv.FormatInt(int64(getNodeId(n)), 10) }), ",")
```
jakebailey: Probably this key needs to be changed to use Anders' preexisting key infra that avoids the high CPU/memory cost of this operation.

internal/compiler/checker.go
```diff
@@ -19881,3 +19938,633 @@ func (c *Checker) getActualTypeVariable(t *Type) *Type {
...
+		}
+	}
+	return nil
+}
+
+func (c *Checker) getUnresolvedSymbolForEntityName(name *ast.Node) *ast.Symbol {
+	identifier := core.IfElse(
+		name.Kind == ast.KindQualifiedName,
+		name.AsQualifiedName().Right,
+		core.IfElse(name.Kind == ast.KindPropertyAccessExpression, name.Name(), name))
```
jakebailey: I feel like this'd read better as a plain if statement or something.

internal/compiler/checker.go
gabritto: The formatter rewrites those to be `true, /*ignoreErrors*/` 🙁

internal/compiler/checker.go
gabritto: At some point I wasn't sure if we were gonna have a `JSDocMemberName` or reuse `JSDocNameReference`. Good catch, I'll fix this but it'll be commented out.

internal/compiler/checker.go
gabritto: My interpretation of the original code, which is just `name.parent === (name.parent.parent as BinaryExpression).left`, is that we don't know if `name.parent.parent` is a `BinaryExpression`, and we're accessing a property which may be `undefined` if that node is not a binary expression. In Go, we both still didn't have a `Left()` node method, and also I think existing methods panic on unhandled node kinds instead of just returning `nil` or something. So I figured we had to add this check in Go.

internal/compiler/checker.go
jakebailey: Then, nevermind, I guess. 😅 

I guess that only works inline.

jakebailey:
Can we just plain drop the JS code, since we did not preserve it elsewhere in Anders' code? (As opposed to commenting it out.)

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
 	}
 	return false
 }
+
+func IsInJSFile(node *Node) bool {
```
ahejlsberg, line 846: We shouldn't need this as we're omitting `.js` specific handling for now.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+func IsDeclarationNameOrImportPropertyName(name *Node) bool {
+	switch name.Parent.Kind {
+	case KindImportSpecifier, KindExportSpecifier:
+		return IsIdentifier(name) || name.Kind == KindStringLiteral
+	default:
+		return IsDeclarationName(name)
+	}
+}
+
+func isStringLiteralLike(node *Node) bool {
```
ahejlsberg: We already have `isStringLiteralLike` in the compiler package. We should unify the two.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return node.Kind == KindStringLiteral || node.Kind == KindNoSubstitutionTemplateLiteral
+}
+
+func isStringOrNumericLiteralLike(node *Node) bool {
```
ahejlsberg: Seems like this should be public.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+func IsExternalModuleImportEqualsDeclaration(node *Node) bool {
+	return node.Kind == KindImportEqualsDeclaration && node.AsImportEqualsDeclaration().ModuleReference.Kind == KindExternalModuleReference
+}
+
+func IsLiteralImportTypeNode(node *Node) bool {
+	return IsImportTypeNode(node) && IsLiteralTypeNode(node.AsImportTypeNode().Argument) && IsStringLiteral(node.AsImportTypeNode().Argument.AsLiteralTypeNode().Literal)
+}
+
+// Add an extra underscore to identifiers that start with two underscores to avoid issues with magic names like '__proto__'
+func EscapeLeadingUnderscores(identifier string) string { // !!! Use a new type instead of `string`
```
ahejlsberg: We don't need escaping in the ported code, so this can go away. Special names are now prefixed by `InternalSymbolNamePrefix` which is an invalid utf-8 character that never occurs in normal strings.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	if strings.HasPrefix(identifier, "__") {
+		return "_" + identifier
+	}
+	return identifier
+}
+
+// Remove extra underscore from escaped identifier text content.
+func UnescapeLeadingUnderscores(identifier string) string {
```
ahejlsberg: This needs to go too.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	if strings.HasPrefix(identifier, "___") {
+		return identifier[1:]
+	}
+	return identifier
+}
+
+func IdText(identifierOrPrivateName *Node) string {
```
ahejlsberg: This is not needed in the port, just use the `.Text()` accessor.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+		escapedText = identifierOrPrivateName.AsIdentifier().Text
+	case KindPrivateIdentifier:
+		escapedText = identifierOrPrivateName.AsPrivateIdentifier().Text
+	default:
+		panic("Unhandled case in IdText: " + identifierOrPrivateName.Kind.String())
+	}
+	return UnescapeLeadingUnderscores(escapedText)
+}
+
+func IsBindableObjectDefinePropertyCall(expr *CallExpression) bool {
```
ahejlsberg: This needs to go, it only is used by JSDoc/.js specific code.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return len(expr.AsNode().Arguments()) == 3 &&
+		IsPropertyAccessExpression(expr.Expression) &&
+		IsIdentifier(expr.Expression.AsPropertyAccessExpression().Expression) &&
+		IdText(expr.Expression.AsPropertyAccessExpression().Expression) == "Object" &&
+		IdText(expr.Expression.AsPropertyAccessExpression().Name()) == "defineProperty" &&
+		isStringOrNumericLiteralLike(expr.Arguments.Nodes[1]) &&
+		isBindableStaticNameExpression(expr.Arguments.Nodes[0], true /*excludeThisKeyword*/)
+}
+
+func isBindableStaticNameExpression(node *Node, excludeThisKeyword bool) bool {
```
ahejlsberg: I think this needs to go too, again only used by JSDoc/.js code. You might look in the ported checker code to see what happened in cases where it was used in the old code base.

internal/ast/utilities.go
ahejlsberg: Actually, fine to keep since we already had in the ported code.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+func isPropertyAccessEntityNameExpression(node *Node) bool {
+	if node.Kind == KindPropertyAccessExpression {
+		expr := node.AsPropertyAccessExpression()
+		return expr.Name().Kind == KindIdentifier && IsEntityNameExpression(expr.Expression)
+	}
+	return false
+}
+
+// Any series of property and element accesses.
+func isBindableStaticAccessExpression(node *Node, excludeThisKeyword bool) bool {
```
ahejlsberg: This needs to go, only used by old JSDoc/.js code.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return IsPropertyAccessExpression(node) &&
+		(!excludeThisKeyword && node.Expression().Kind == KindThisKeyword ||
+			IsIdentifier(node.Name()) && isBindableStaticNameExpression(node.Expression(), excludeThisKeyword)) ||
+		isBindableStaticElementAccessExpression(node, excludeThisKeyword)
+}
+
+func isBindableStaticElementAccessExpression(node *Node, excludeThisKeyword bool) bool {
```
ahejlsberg: Needs to go also.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return isLiteralLikeElementAccess(node) &&
+		((!excludeThisKeyword && node.Expression().Kind == KindThisKeyword) ||
+			IsEntityNameExpression(node.Expression()) ||
+			isBindableStaticAccessExpression(node.Expression(), true /*excludeThisKeyword*/))
+}
+
+// x[0] OR x['a'] OR x[Symbol.y]
+func isLiteralLikeElementAccess(node *Node) bool {
```
ahejlsberg: And I think this needs to go too.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return false
+}
+
+func IsImportOrExportSpecifier(node *Node) bool {
+	return IsImportSpecifier(node) || IsExportSpecifier(node)
+}
+
+type AssignmentDeclarationKind int32
+
+const (
```
ahejlsberg: This definitely needs to go, this is the classification code used by the old JSDoc/.js code. The only special case we have now is declarations with an `ast.KindBinaryExpression` node that occur in function property assignments (i.e. assignments of properties to the static side of a function written right after the function declaration). That particular case is now handled by the `isFunctionPropertyAssignment` helper, you can see an example of it in the binder code.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	AssignmentDeclarationKindObjectDefinePropertyValue
+	// Object.defineProperty(exports || module.exports, 'name', ...);
+	AssignmentDeclarationKindObjectDefinePropertyExports
+	// Object.defineProperty(Foo.prototype, 'name', ...);
+	AssignmentDeclarationKindObjectDefinePrototypeProperty
+)
+
+// Given a BinaryExpression, returns SpecialPropertyAssignmentKind for the various kinds of property
+// assignments we treat as special in the binder
+func GetAssignmentDeclarationKind(expr *Node) AssignmentDeclarationKind {
```
ahejlsberg: This needs to go.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	special := getAssignmentDeclarationKindWorker(expr)
+	if special == AssignmentDeclarationKindProperty || IsInJSFile(expr) {
+		return special
+	}
+	return AssignmentDeclarationKindNone
+}
+
+func getAssignmentDeclarationKindWorker(expr *Node) AssignmentDeclarationKind {
```
ahejlsberg: Delete this.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+		GetElementOrPropertyAccessName(expr.AsBinaryExpression().Left) == "prototype" &&
+		IsObjectLiteralExpression(GetInitializerOfBinaryExpression(expr)) {
+		// F.prototype = { ... }
+		return AssignmentDeclarationKindPrototype
+	}
+
+	return GetAssignmentDeclarationPropertyAccessKind(expr.AsBinaryExpression().Left)
+}
+
+func GetAssignmentDeclarationPropertyAccessKind(lhs *Node) AssignmentDeclarationKind {
```
ahejlsberg: Delete this.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+
+func IsSignedNumericLiteral(node *Node) bool {
+	if node.Kind == KindPrefixUnaryExpression {
+		node := node.AsPrefixUnaryExpression()
+		return (node.Operator == KindPlusToken || node.Operator == KindMinusToken) && IsNumericLiteral(node.Operand)
+	}
+	return false
+}
+
+func IsPrototypeAccess(node *Node) bool {
```
ahejlsberg: I think this needs to go, only JSDoc/.js uses it.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return isBindableStaticAccessExpression(node, false /*excludeThisKeyword*/) && GetElementOrPropertyAccessName(node) == "prototype"
+}
+
+func GetInitializerOfBinaryExpression(expr *Node) *Node {
```
ahejlsberg: Delete this, JSDoc/.js specific.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+
+func IsExportsIdentifier(node *Node) bool {
+	return IsIdentifier(node) && node.Text() == "exports"
+}
+
+func IsModuleIdentifier(node *Node) bool {
+	return IsIdentifier(node) && node.Text() == "module"
+}
+
+func IsModuleExportsAccessExpression(node *Node) bool {
```
ahejlsberg: You may want to check if this can go.

internal/ast/utilities.go
```diff
@@ -842,3 +843,358 @@ func IsImportMeta(node *Node) bool {
...
+	return (IsPropertyAccessExpression(node) || isLiteralLikeElementAccess(node)) &&
+		IsModuleIdentifier(node.Expression()) &&
+		GetElementOrPropertyAccessName(node) == "exports"
+}
+
+func GetElementOrPropertyAccessName(node *Node) string {
```
ahejlsberg: This can go. Just replace it with use of `.Text()` accessor.

PR #159, opened 2024-12-10
Eliminate relativePathSegmentRegExp

jakebailey:
This regex took up a significant amount of time in program load. Replace it with some simple string matches, leaving behind the old implementation as a differential fuzz test.

```
goos: linux
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/tspath
                                                     │    old.txt    │               new.txt               │
                                                     │    sec/op     │   sec/op     vs base                │
HasRelativePathSegment/foo/bar/baz-20                   558.95n ± 1%   31.74n ± 2%  -94.32% (p=0.000 n=10)
HasRelativePathSegment/./some/path-20                   106.35n ± 1%   21.87n ± 1%  -79.44% (p=0.000 n=10)
HasRelativePathSegment//foo/./bar/../../.-20            303.60n ± 1%   11.25n ± 1%  -96.30% (p=0.000 n=10)
HasRelativePathSegment/foo/foo/foo/foo/foo/...etc-20   17723.0n ± 3%   226.2n ± 0%  -98.72% (p=0.000 n=10)
geomean                                                  752.0n        36.45n       -95.15%

                                                     │     B/op     │    B/op     vs base                 │
HasRelativePathSegment/foo/bar/baz-20                  0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment/./some/path-20                  0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment//foo/./bar/../../.-20           0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment/foo/foo/foo/foo/foo/...etc-20   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                           ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean

                                                     │  allocs/op   │ allocs/op   vs base                 │
HasRelativePathSegment/foo/bar/baz-20                  0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment/./some/path-20                  0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment//foo/./bar/../../.-20           0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
HasRelativePathSegment/foo/foo/foo/foo/foo/...etc-20   0.000 ± 0%     0.000 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                           ²               +0.00%                ²
¹ all samples are equal
² summaries must be >0 to compute geomean
```

### DIFF
--- a/internal/tspath/path.go
+++ b/internal/tspath/path.go
@@ -333,19 +333,19 @@ func hasRelativePathSegment(p string) bool {
 		return true
 	}
 
-	if strings.Contains(p, "//") {
-		return true
-	}
-
-	if strings.Contains(p, "/./") || strings.Contains(p, "/../") {
-		return true
-	}
-
 	if strings.HasPrefix(p, "./") || strings.HasPrefix(p, "../") {
 		return true
 	}
 
 	if strings.HasSuffix(p, "/.") || strings.HasSuffix(p, "/..") {
+		return true
+	}
+
+	if strings.Contains(p, "//") {
+		return true
+	}
+
+	if strings.Contains(p, "/./") || strings.Contains(p, "/../") {
 		return true
 	}
 


internal/tspath/path.go
```diff
@@ -333,16 +328,40 @@ func GetNormalizedAbsolutePath(fileName string, currentDirectory string) string
 	return getPathFromPathComponents(getNormalizedPathComponents(fileName, currentDirectory))
 }
 
+func hasRelativePathSegment(p string) bool {
```
jakebailey, line 331: This is a super simple and dumb piece of code which works so much better than the regex that I'm not going to attempt to microoptimize it right now.

jakebailey:
Really I should add some benchmarks for `NormalizePath` itself but I know from profiling that this is the main problem.

internal/tspath/path.go
andrewbranch: Fair, but you don’t think the `HasPrefix` and `HasSuffix` conditions would be faster than the `Contains`?

jakebailey: Yeah, I should flip them. I wrote these in order of the cases I thought of via the regex.

PR #160, opened 2024-12-11
Fix:allow type annotation on arrow function with empty parameters

sandersn:
Previously the empty parameter case was confused with the missing parameter case because empty slices are optimised to nil slices. This put the parser in an error mode which skipped the return type annotation.

That is, `(): T => value` was misparsed as `=> value` because `()` was ambiguous with a missing parameter list.

I made NewSlice return an empty slice for zero-length requests *only for parseDelimitedList*, but probably should for other calls to NewSlice as well.

The other option I thought of was to port the isMissingList bool to NodeList, but that changes the struct from 32 to 40 bytes, which I believe would be less efficient than a non-nil slice only for zero-length lists.


internal/compiler/parser.go
```diff
@@ -326,7 +326,7 @@ func (p *Parser) parseBracketedList(kind ParsingContext, parseElement func(p *Pa
 }
 
 func (p *Parser) parseEmptyNodeList() *ast.NodeList {
```
sandersn: I decided to not to inline `parseEmptyNodeList` in case `nil` turns out to be wrong. I'm not opposed to changing it if others care though.

ahejlsberg:
Where in the parser is the code that gets confused currently, i.e. the code that mis-parses the parameter list?

sandersn:
`typeHasArrowFunctionBlockingParseError` and the check that calls it in `parseParenthesizedArrowFunctionExpression`

```go
	if returnType != nil && !allowAmbiguity && typeHasArrowFunctionBlockingParseError(returnType) {
		return nil
	}
```

internal/compiler/parser.go
jakebailey: I'll be inlined by the compiler, so having a named function that indicates what's happening is not a bad idea.

sandersn: After much discussion with Anders, we decided that this wasn't the source of the `nil`. So I reverted it. I'm going to run tsc's test cases to figure what syntax *is* the source of the nil.

PR #161, opened 2024-12-11
Split parser into own subpackage

rbuckton:
This moves `Parser` into its own `parser` subpackage.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -314,8 +314,8 @@ func (p *Parser) parseDelimitedList(kind ParsingContext, parseElement func(p *Pa
 	return p.factory.NewNodeList(core.NewTextRange(pos, p.nodePos()), slice)
 }
 
-// Return a non-nil (but possibly empty) slice if parsing was successful, or nil if opening token wasn't found
-// or parseElement returned nil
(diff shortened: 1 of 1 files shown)


internal/parser/fixtures_test.go
rbuckton, line 1: This is a duplicate of internal/compiler/fixtures_test.go. If necessary we could pull this out into a shared fixtures_test package.

ahejlsberg:
Would love to have the binder move to its own package as well.

internal/parser/fixtures_test.go
jakebailey: It would be nice to stick these in some new package to not duplicate them, if you can find a good place to put them.

Honestly, you could get away with just moving them to the `filefixture` package; I had higher hopes for this little package but I don't think we'll acutally end up using it that much.

jakebailey:
Ah, seems like the stringer code gen needs to be updated.

rbuckton:
> Would love to have the binder move to its own package as well.

I'll look into that in a subsequent PR

rbuckton:
@jakebailey, @ahejlsberg once this is approved and merged I'll have a PR up for binder as I made the changes in a branch off of this branch

PR #163, opened 2024-12-11
Add `ModuleKindNode18`

andrewbranch:
Counterpart to https://github.com/microsoft/TypeScript/pull/60722


jakebailey:
My impressive was that we were porting up to the pinned commit in the submodule; are we wanting to take this now?

iisaduan:
If we are taking this, could you also update `tsoptions/enummaps.go`?

andrewbranch:
When are we moving the pin? I’ve already included a fix in module resolution that I found due to porting.

jakebailey:
I'm not sure... I was assuming we were aiming for full checker completeness before then, so we can hand backport each change.

PR #165, opened 2024-12-12
Fix compile after merge conflict

jakebailey:
#161 and #164 raced.

PR #166, opened 2024-12-12
Fix merge issue

ahejlsberg:
Build is broken without this fix.


jakebailey:
Racing with a race fix #165 😄

PR #167, opened 2024-12-12
Increase lint timeout in CI

jakebailey:
Linting is getting longer and longer as we add more code; we're just at the 1 minute default in CI, so increase that to 5 minutes.

(My own machine is roughly 5x faster than CI, hence the arbitrary choice.)


jakebailey:
A passing run from a few commits ago:

```
Finished lint in 56.5s
```

Versus a failure now in main:

```
Error in lint in 1m 1.4s
```

Just barely on the edge!

jakebailey:
Merging this to get main green again.

PR #168, opened 2024-12-12
Split binder into own subpackage

rbuckton:
This moves `Binder` into its own `binder` subpackage.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -26,6 +26,7 @@ var printTypes = false
 var pretty = true
 var listFiles = false
 var pprofDir = ""
+var outDir = ""
 
 func printDiagnostic(d *ast.Diagnostic, level int, comparePathOptions tspath.ComparePathsOptions) {
 	file := d.File()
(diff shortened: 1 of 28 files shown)


internal/binder/types.go
```diff
@@ -0,0 +1,25 @@
+package binder
```
ahejlsberg: I think all of the declarations in this file should move to `ast/symbol.go` where the `Symbol` type is defined.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
+package binder
```
ahejlsberg: I would probably just fold this file into `binder.go`. I'm not sure the utility functions here are any different from the utility functions that already live in `binder.go`.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	return false
+}
+
+// Atomic ids
+
+var nextNodeId atomic.Uint32
+var nextSymbolId atomic.Uint32
+
+func GetNodeId(node *ast.Node) ast.NodeId {
```
ahejlsberg: I think this should be in the `ast` package. We want one central place where node IDs are generated.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+	if node.Id == 0 {
+		node.Id = ast.NodeId(nextNodeId.Add(1))
+	}
+	return node.Id
+}
+
+func GetSymbolId(symbol *ast.Symbol) ast.SymbolId {
```
ahejlsberg: Also move this to `ast`.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+func isAsyncFunction(node *ast.Node) bool {
+	switch node.Kind {
+	case ast.KindFunctionDeclaration, ast.KindFunctionExpression, ast.KindArrowFunction, ast.KindMethodDeclaration:
+		data := node.BodyData()
+		return data.Body != nil && data.AsteriskToken == nil && ast.HasSyntacticModifier(node, ast.ModifierFlagsAsync)
+	}
+	return false
+}
+
+func SymbolName(symbol *ast.Symbol) string {
```
ahejlsberg: Feels like this should be in `ast` since it is used by the checker and isn't really a binder thing.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func GetSymbolTable(data *ast.SymbolTable) ast.SymbolTable {
+	if *data == nil {
+		*data = make(ast.SymbolTable)
+	}
+	return *data
+}
+
+func GetMembers(symbol *ast.Symbol) ast.SymbolTable {
```
ahejlsberg: Should be in `ast`.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+	return GetSymbolTable(&symbol.Members)
+}
+
+func GetExports(symbol *ast.Symbol) ast.SymbolTable {
```
ahejlsberg: Should be in `ast`.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+	return GetSymbolTable(&symbol.Exports)
+}
+
+func GetLocals(container *ast.Node) ast.SymbolTable {
```
ahejlsberg: Should be in `ast`.

internal/binder/utilities.go
```diff
@@ -0,0 +1,271 @@
...
+		return node.AsPropertySignatureDeclaration().PostfixToken
+	case ast.KindMethodDeclaration:
+		return node.AsMethodDeclaration().PostfixToken
+	case ast.KindMethodSignature:
+		return node.AsMethodSignatureDeclaration().PostfixToken
+	}
+	panic("Unhandled case in getPostfixTokenFromNode")
+}
+
+func ExportAssignmentIsAlias(node *ast.Node) bool {
```
ahejlsberg: Should be in `ast`.

PR #170, opened 2024-12-12
Avoid calculating spread indices multiple times per array literal

EricCornelson:
It looks like `indicesComputed` was added to `ArrayLiteralLinks`, but is never set. This PR sets the flag to true after calculating the indices the first time, which saves a significant amount of time (tested on vscode src with -q).

image

Before:
image

After:
image


internal/compiler/checker.go
```diff
@@ -17751,6 +17751,7 @@ func (c *Checker) createSyntheticExpression(parent *ast.Node, t *Type, isSpread
 func (c *Checker) getSpreadIndices(node *ast.Node) (int, int) {
 	links := c.arrayLiteralLinks.get(node)
 	if !links.indicesComputed {
+		links.indicesComputed = true
```
jakebailey: Nit, but can you set it at the bottom of the block with the other fields?

EricCornelson: Absolutely!

PR #171, opened 2024-12-12
Make getResolvedModule use module resolver

jakebailey:
I am unsure if the `FileIncludeReason` is correct.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -1,7 +1,6 @@
 package compiler
 
 import (
-	"encoding/json"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/compiler/program.go
```diff
@@ -179,11 +178,8 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
 }
 
 func (p *Program) getResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile {
-	directory := tspath.GetDirectoryPath(currentSourceFile.FileName())
-	if tspath.IsExternalModuleNameRelative(moduleReference) {
-		return p.findSourceFile(tspath.CombinePaths(directory, moduleReference), FileIncludeReason{FileIncludeKindImport, 0})
-	}
-	return p.findNodeModule(moduleReference)
+	resolved := p.resolver.ResolveModuleName(moduleReference, currentSourceFile.FileName(), core.ModuleKindNodeNext, nil)
```
jakebailey, line 181: I don't like that this is hardcoding the module kind stuff but the other code does too.

internal/compiler/program.go
```diff
@@ -179,11 +178,8 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
...
+	return p.findSourceFile(resolved.ResolvedFileName, FileIncludeReason{FileIncludeKindImport, 0})
```
jakebailey, line 182: There's a reason here but it doesn't get used at all.

PR #172, opened 2024-12-12
Add emitter scaffolding

rbuckton:
This adds `Program.Emit()` and the basic scaffolding for the emitter. At present, this only performs printback of the input source file as output transformation has not yet been implemented.

This also adds an `--outDir DIR` option to `tsgo.exe` to verify emit. When not specified, `tsgo.exe` will not trigger emit.

In addition, this fixes a bug in `getLinesBetweenPositions` that incorrectly caused the printer to print excess newlines.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -92,6 +92,11 @@ func NewProgram(options ProgramOptions) *Program {
 
 	p.processRootFiles(fileInfos)
 
+	// Sort files by path so we get a stable ordering between runs
+	slices.SortFunc(p.files, func(f1 *ast.SourceFile, f2 *ast.SourceFile) int {
+		return strings.Compare(string(f1.Path()), string(f2.Path()))
+	})
+
 	return p
 }
 
--- a/internal/core/compileroptions.go
+++ b/internal/core/compileroptions.go
@@ -18,6 +18,7 @@ type CompilerOptions struct {
 	CustomConditions                   []string             `json:"customConditions"`
 	EmitDeclarationOnly                Tristate             `json:"emitDeclarationOnly"`
 	EmitBOM                            Tristate             `json:"emitBOM"`
+	DownlevelIteration                 Tristate             `json:"downlevelIteration"`
 	ESModuleInterop                    Tristate             `json:"esModuleInterop"`
 	ExactOptionalPropertyTypes         Tristate             `json:"exactOptionalPropertyTypes"`
 	ExperimentalDecorators             Tristate             `json:"experimentalDecorators"`
--- a/internal/vfs/os.go
+++ b/internal/vfs/os.go
@@ -1,11 +1,11 @@
 package vfs
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
-	"sync"
 	"unicode"
 
 	"github.com/microsoft/typescript-go/internal/tspath"
@@ -26,18 +26,31 @@ type osFS struct {
 	common
 }
 
-var isFileSystemCaseSensitive = sync.OnceValue(func() bool {
+// We do this right at startup to minimize the chance that executable gets moved or deleted.
+var isFileSystemCaseSensitive = func() bool {
 	// win32/win64 are case insensitive platforms
 	if runtime.GOOS == "windows" {
 		return false
 	}
 
+	// As a proxy for case-insensitivity, we check if the current executable exists under a different case.
+	// This is not entirely correct, since different OSs can have differing case sensitivity in different paths,
+	// but this is largely good enough for our purposes (and what sys.ts used to do with __filename).
+	exe, err := os.Executable()
+	if err != nil {
+		panic(fmt.Sprintf("vfs: failed to get executable path: %v", err))
+	}
+
 	// If the current executable exists under a different case, we must be case-insensitve.
-	if _, err := os.Stat(swapCase(os.Args[0])); os.IsNotExist(err) {
-		return false
+	swapped := swapCase(exe)
+	if _, err := os.Stat(swapped); err != nil {
+		if os.IsNotExist(err) {
+			return false
+		}
+		panic(fmt.Sprintf("vfs: failed to stat %q: %v", swapped, err))
 	}
 	return true
-})
+}()
 
 // Convert all lowercase chars to uppercase, and vice-versa
 func swapCase(str string) string {
@@ -52,7 +65,7 @@ func swapCase(str string) string {
 }
 
 func (vfs *osFS) UseCaseSensitiveFileNames() bool {
-	return isFileSystemCaseSensitive()
+	return isFileSystemCaseSensitive
 }
 
 var osReadSema = make(chan struct{}, 128)
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -4531,10 +4531,10 @@ type JsxNamespacedName struct {
 	Namespace *IdentifierNode // IdentifierNode
 }
 
-func (f *NodeFactory) NewJsxNamespacedName(name *IdentifierNode, namespace *IdentifierNode) *Node {
+func (f *NodeFactory) NewJsxNamespacedName(namespace *IdentifierNode, name *IdentifierNode) *Node {
 	data := &JsxNamespacedName{}
-	data.name = name
 	data.Namespace = namespace
+	data.name = name
 	return newNode(KindJsxNamespacedName, data)
 }
 
@@ -4659,6 +4659,10 @@ func (f *NodeFactory) NewJsxAttribute(name *JsxAttributeName, initializer *JsxAt
 	data.name = name
 	data.Initializer = initializer
 	return newNode(KindJsxAttribute, data)
+}
+
+func (node *JsxAttribute) Name() *JsxAttributeName {
+	return node.name
 }
 
 func (node *JsxAttribute) ForEachChild(v Visitor) bool {
--- a/internal/ast/diagnostic.go
+++ b/internal/ast/diagnostic.go
@@ -73,7 +73,7 @@ func NewDiagnostic(file *SourceFile, loc core.TextRange, message *diagnostics.Me
 
 func NewDiagnosticChain(chain *Diagnostic, message *diagnostics.Message, args ...any) *Diagnostic {
 	if chain != nil {
-		return NewDiagnostic(chain.file, chain.loc, message, args...).SetRelatedInfo(chain.relatedInformation)
+		return NewDiagnostic(chain.file, chain.loc, message, args...).AddMessageChain(chain).SetRelatedInfo(chain.relatedInformation)
 	}
 	return NewDiagnostic(nil, core.TextRange{}, message, args...)
 }
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -113,6 +113,10 @@ func IsLiteralKind(kind Kind) bool {
 	return KindFirstLiteralToken <= kind && kind <= KindLastLiteralToken
 }
 
+func IsLiteralExpression(node *Node) bool {
+	return IsLiteralKind(node.Kind)
+}
+
 func IsStringLiteralLike(node *Node) bool {
 	switch node.Kind {
 	case KindStringLiteral, KindNoSubstitutionTemplateLiteral:
--- a/internal/compiler/binder.go
+++ b/internal/compiler/binder.go
@@ -1245,11 +1245,8 @@ func (b *Binder) lookupName(name string, container *ast.Node) *ast.Symbol {
 		}
 	}
 	declaration := container.DeclarationData()
-	if declaration != nil {
-		symbol := declaration.Symbol
-		if symbol != nil {
-			return symbol.Exports[name]
-		}
+	if declaration != nil && declaration.Symbol != nil {
+		return declaration.Symbol.Exports[name]
 	}
 	return nil
 }
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -398,209 +398,220 @@ const (
 	IterationUseAsyncGeneratorReturnType = IterationUseAllowsAsyncIterablesFlag
 )
 
+type IterationTypeKind int32
+
+const (
+	IterationTypeKindYield IterationTypeKind = iota
+	IterationTypeKindReturn
+	IterationTypeKindNext
+)
+
 // Checker
 
 type Checker struct {
-	program                            *Program
-	host                               CompilerHost
-	compilerOptions                    *core.CompilerOptions
-	files                              []*ast.SourceFile
-	typeCount                          uint32
-	symbolCount                        uint32
-	totalInstantiationCount            uint32
-	instantiationCount                 uint32
-	instantiationDepth                 uint32
-	inlineLevel                        int
-	currentNode                        *ast.Node
-	languageVersion                    core.ScriptTarget
-	moduleKind                         core.ModuleKind
-	isInferencePartiallyBlocked        bool
-	legacyDecorators                   bool
-	allowSyntheticDefaultImports       bool
-	strictNullChecks                   bool
-	strictFunctionTypes                bool
-	strictBindCallApply                bool
-	strictPropertyInitialization       bool
-	noImplicitAny                      bool
-	noImplicitThis                     bool
-	useUnknownInCatchVariables         bool
-	exactOptionalPropertyTypes         bool
-	arrayVariances                     []VarianceFlags
-	globals                            ast.SymbolTable
-	evaluate                           Evaluator
-	stringLiteralTypes                 map[string]*Type
-	numberLiteralTypes                 map[float64]*Type
-	bigintLiteralTypes                 map[PseudoBigInt]*Type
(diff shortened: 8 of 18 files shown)


internal/vfs/vfs.go
```diff
@@ -310,3 +312,37 @@ func (v *vfs) Realpath(path string) string {
 	}
 	return realpath
 }
+
+func (v *vfs) writeFile(path string, content string, writeByteOrderMark bool) error {
+	file, err := os.Create(path)
```
jakebailey: FWIW we'll have to split things up here because this will cause all virtual FSs to write files which won't be right. I was going to work on other stuff but if you are blocked for testing I can finagle this library to allow writing to the virtual FS, potentially.

internal/core/compileroptions.go
```diff
@@ -16,6 +16,8 @@ type CompilerOptions struct {
 	AllowUnusedLabels                  Tristate             `json:"allowUnusedLabels"`
 	CheckJs                            Tristate             `json:"checkJs"`
 	CustomConditions                   []string             `json:"customConditions"`
+	EmitDeclarationOnly                Tristate             `json:"emitDeclarationOnly"`
+	EmitBOM                            Tristate             `json:"emitBOM"`
```
jakebailey, line 20: I could have sworn we deprecated this 😢

internal/vfs/vfs.go
jakebailey: For now, though, you could just make this a "hook" sort of thing like `rootFor` is, and then panic if we aren't using it with the OS variant.

rbuckton: Wouldn't a VFS that can't write files just return an `error`?

jakebailey: Maybe, but it's really just a limitation of the current in-memory FS support, so if you try to depend on that, it's more of an error in the programming because it's never going to work.

internal/core/compileroptions.go
rbuckton: Did we? It's not listed in https://github.com/microsoft/TypeScript/issues/54500

jakebailey: I don't think we did, no, I just thought we did (wrongly).

EricCornelson:
The program parts look pretty good to me. The only thing that gave me pause was whether or not `Emit` should live in emitter or program, but I don't think it's a big deal either way

internal/vfs/vfs.go
rbuckton: I've updated it to be hook-based, aligning with `RealpathFS`.

rbuckton:
> The program parts look pretty good to me. The only thing that gave me pause was whether or not `Emit` should live in emitter or program, but I don't think it's a big deal either way

I've moved it to program.go. I kept it separate mostly to avoid any potentially messy merge conflicts as I'd started this before #107 had merged.

internal/vfs/vfs.go
jakebailey: That works for one side, but not for the OS side I don't think; the "FS extension" thing is an `io.FS` method that I don't think will work on the OS side.

I sent https://github.com/microsoft/typescript-go/pull/173 which should rework things in a way that makes it a lot easier to implement writing.

internal/vfs/os.go
```diff
@@ -79,3 +79,38 @@ func (vfs *osFS) Realpath(path string) string {
...
+
+	if _, err := file.WriteString(content); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (vfs *osFS) ensureDirectoryExists(directoryPath string) error {
+	return os.MkdirAll(directoryPath, 0o777)
```
jakebailey, line 117: Should this be `0o755`? I would normally say "yes definitely" but Node's default is `0777` which seems... completely wrong.

internal/vfs/os.go
```diff
@@ -79,3 +79,38 @@ func (vfs *osFS) Realpath(path string) string {
...
+}
+
+func (vfs *osFS) WriteFile(path string, content string, writeByteOrderMark bool) error {
+	_ = rootLength(path) // Assert path is rooted
+	if err := vfs.writeFile(path, content, writeByteOrderMark); err == nil {
+		return nil
+	}
+	if err := vfs.ensureDirectoryExists(tspath.GetDirectoryPath(tspath.NormalizePath(path))); err != nil {
```
jakebailey, line 125: Can we get away with not normalizing again? I think all callers are normalized.

(I would really like to explore using a named `NormalizedPath` or something to avoid extra work...)

internal/vfs/iofs.go
```diff
@@ -13,11 +13,23 @@ type RealpathFS interface {
...
+	MkdirAll(path string, perm fs.FileMode) error
+}
+
 // FromIOFS creates a new FS from an [fs.FS].
 //
 // For paths like `c:/foo/bar`, fsys will be used as though it's rooted at `/` and the path is `/c:/foo/bar`.
 //
 // If the provided [fs.FS] implements [RealpathFS], it will be used to implement the Realpath method.
+// If the provided [fs.FS] implements [WriteFileFS], it will be used to implement the WriteFile method.
+// If the provided [fs.FS] implements [MkdirAllFS], it will be used to implement the WriteFile method.
```
jakebailey, line 32:
```suggestion
// If the provided [fs.FS] implements [MkdirAllFS], it will be used to implement the MkdirAll method.
```

internal/compiler/emitHost.go
```diff
@@ -0,0 +1,67 @@
...
+type WriteFileData struct {
+	SourceMapUrlPos int
+	// BuildInfo BuildInfo
+	Diagnostics      []*ast.Diagnostic
+	DiffersOnlyInMap bool
+	SkippedDtsWrite  bool
+}
+
+// NOTE: EmitHost operations must be thread-safe
+type EmitHost interface {
```
jakebailey, line 17: Are we ever going to have more than one implementation of this?

internal/compiler/emitHost.go
```diff
@@ -0,0 +1,67 @@
...
+	host.commonSourceDirectoryMutex.Lock()
+	defer host.commonSourceDirectoryMutex.Unlock()
+
+	// double-check after lock
+	if len(host.commonSourceDirectory) > 0 {
+		return host.commonSourceDirectory
+	}
+
+	host.commonSourceDirectory = host.program.CommonSourceDirectory()
+	return host.commonSourceDirectory
```
jakebailey: This is still racy at the first if statement because it's an unprotected read; what I would suggest is writing:

```suggestion
	host.commonSourceDirectoryOnce.Do(func() {
		host.commonSourceDirectory = host.program.CommonSourceDirectory()
	})
	return host.commonSourceDirectory
```

And then replacing the mutex with `commonSourceDirectoryOnce sync.Once`.

But, I'm also not sure if the emit host is who should be synchronizing this; it seems like it's the Program which should have this code and be concurrency safe.

internal/vfs/iofs.go
rbuckton: There is no `MkdirAll` method on `vfs.FS`.

internal/compiler/emitHost.go
rbuckton: I'm not 100% certain yet. This was something we allowed users to override when consuming the API, so that all depends on our long term API story (as in, could someone build on top of our go package). I opted to leave it as-is to more faithfully port portions of the emitter.

internal/compiler/emitHost.go
rbuckton: Though I can use a `Once`, the racy read should be fine (?) because once `commonSourceDirectory` is set to a non-empty string it will never change. The first check is a racy fast-path. If the fast path fails, it locks the mutex and double-checks the initial condition. `sync.Once.Do` essentially does the same thing except it uses a function closure.

(unless Go does some unexpected optimization or reordering I'm not familiar with that would break that assumption?)

I'm fine with `Program` synchronizing this instead, we just haven't documented on `Program` how much of it is thread-safe.

internal/vfs/os.go
rbuckton: Node's default is `0o777 & ~process.umask()`. Does Go not incorporate `umask`? From the documentation it seems like it does.

rbuckton: From what I can tell, Go supports `umask` for both `os.Create` and `os.MkdirAll`, so if your `umask` is set to `0022` then you'll get files with `0644` and folders with `0755` even if perms are set to `0666`/`0777` (respectively).

internal/vfs/iofs.go
jakebailey: Oh, hm. I assume we will need it to replace TS's host `createDirectory`?

internal/compiler/emitHost.go
jakebailey: It won't be fine; it'll error in `-race` mode. You can try out `go run -race ./cmd/tsgo ... -outdir=...` and it should complain.

The string is also a multi-word value that can't be written atomically (it's a pointer and len pair), so it isn't safe even if the race is ignored (which you wouldn't do anyway given real atomics exist in `sync/atomic`, which is what `once` is doing).

You can do the double check with an `RWMutex` instead, if you'd like. (Or use an `atomic.Pointer` maybe.) But I would just use a `once`.

internal/vfs/iofs.go
rbuckton: `writeFile` was the only code that ever invoked `createDirectory`.

internal/vfs/os.go
jakebailey: Yeah, I missed "before umask" in the `os` docs.

internal/vfs/iofs.go
jakebailey: Yeah, surprising, I would have thought we'd use it for some other reason. All good, then.

cmd/tsgo/main.go
```diff
@@ -57,10 +58,16 @@ func main() {
 	flag.BoolVar(&pretty, "pretty", true, "Get prettier errors")
 	flag.BoolVar(&listFiles, "listfiles", false, "List files in the program")
 	flag.StringVar(&pprofDir, "pprofdir", "", "Generate pprof CPU/memory profiles to the given directory")
+	flag.StringVar(&outDir, "outDir", "", "Emit to the given directory")
```
```suggestion
	flag.StringVar(&outDir, "outdir", "", "Emit to the given directory")
```

Just a consistency nit with the other flags, until we get our own CLI parser in here and go back to our own naming rules.

internal/compiler/program.go
```diff
@@ -541,3 +546,159 @@ func (p *Program) getEmitModuleFormatOfFile(sourceFile *ast.SourceFile) core.Mod
 	// }
 	return p.compilerOptions.GetEmitModuleKind()
 }
+
+func (p *Program) CommonSourceDirectory() string {
+	if !p.hasComputedCommonSourceDirectory {
```
jakebailey: Ah, yeah, this is definitely the thing to use a `once` on, rather than `EmitHost`. Missed this in my top-to-botton review.

internal/compiler/program.go
```diff
@@ -541,3 +546,159 @@ func (p *Program) getEmitModuleFormatOfFile(sourceFile *ast.SourceFile) core.Mod
...
+			p.host.GetCurrentDirectory(),
+			p.host.FS().UseCaseSensitiveFileNames(),
+		)
+		p.hasComputedCommonSourceDirectory = true
+	}
+	return p.commonSourceDirectory
+}
+
+func computeCommonSourceDirectoryOfFilenames(fileNames []string, currentDirectory string, useCaseSensitiveFileNames bool) string {
+	var commonPathComponents []string = nil
```
jakebailey: The nil here is redundant.

```suggestion
	var commonPathComponents []string
```

internal/compiler/program.go
```diff
@@ -541,3 +546,159 @@ func (p *Program) getEmitModuleFormatOfFile(sourceFile *ast.SourceFile) core.Mod
...
+
+	return tspath.GetPathFromPathComponents(commonPathComponents)
+}
+
+func getCommonSourceDirectory(options *core.CompilerOptions, files []string, currentDirectory string, useCaseSensitiveFileNames bool) string {
+	var commonSourceDirectory string
+	// !!! If a rootDir is specified use it as the commonSourceDirectory
+	// !!! Project compilations never infer their root from the input source paths
+
+	commonSourceDirectory = computeCommonSourceDirectoryOfFilenames(files, currentDirectory, useCaseSensitiveFileNames)
```
jakebailey, line 622: I assume this assignment is this way because we're just missing code above?

rbuckton: Yes, there are as-yet unimplemented branches in this code

internal/compiler/program.go
rbuckton: Won't `Once` allocate a new function closure for every invocation, even if no work is done?

internal/compiler/emitHost.go
rbuckton: Ah, this is a case of "Go slices are weird", since you also can't compare them by reference.

internal/compiler/program.go
jakebailey: You'd think, but IIRC they've made it such that `sync.Once.Do` inlines in the common case such that the allocation (which is on the stack anyway) does not actually happen.

It's different than the `b.bind` thing where we were passing it into code that couldn't inline or be optimized.

jakebailey: See also https://godbolt.org/z/xd6hxPPbf

internal/vfs/os.go
rbuckton: There's no guarantee the path is normalized, just as there's no guarantee the path is rooted (which is why this and other methods validate the path is rooted). We also call `NormalizePath` for other VFS methods, though it's buried in `splitPath`/`rootAndPath`, so it makes sense to remain consistent.

jakebailey:
I think this looks good, but once last thing to check; since our program construction is more or less fake right now, where we just walk a dir and find all ts/tsx files, will this end up emitting files we shouldn't? I assume so, which can be fine, but we're going to start loading `lib.d.ts` files and those shouldn't be emitted for our testing, I would think.

rbuckton:
It won't generate emit for .d.ts input files at present, and only emits if you specify --outdir, so the risk is fairly low.

jakebailey:
Yeah, that seems fine, then. I was just thinking of random `.ts` files that happen to appear within someone's project, but if we're ignoring `.d.ts` files for our basic use right now, I think that's plenty given most projects will have some sort of `src` dir we can run on, like vscode or TS itself.

cmd/tsgo/main.go
```diff
@@ -57,10 +58,16 @@ func main() {
...
 	flag.StringVar(&pprofDir, "pprofdir", "", "Generate pprof CPU/memory profiles to the given directory")
+	flag.StringVar(&outDir, "outdir", "", "Emit to the given directory")
 	flag.Parse()
 
 	rootPath := flag.Arg(0)
-	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext}
+	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext, NoEmit: core.TSTrue}
+	if len(outDir) > 0 {
+		compilerOptions.NoEmit = core.TSFalse
+		compilerOptions.OutDir = outDir
```
jakebailey: I just ran this; this needs to be something like:

```suggestion
		compilerOptions.OutDir = tspath.ResolvePath(currentDirectory, outDir)
```

Otherwise I get:

```
panic: vfs: path "wow/src/vs/platform/utilityProcess/common/utilityProcessWorkerService.js" is not absolute
```

(so, probably move down below)

jakebailey:
Interestingly since we currently include `.js` files for vscode (due to missing node types), I get:

```
├── node_modules
│   ├── events
│   │   └── events.js
│   ├── string_decoder
│   │   └── lib
│   │       └── string_decoder.js
│   └── util
│       └── util.js
└── src
    ├── bootstrap-cli.js
    ├── bootstrap-esm.js
    ├── bootstrap-fork.js
    ├── bootstrap-import.js
    ├── bootstrap-meta.js
```

rbuckton:
Excluding Node_modules is one of the as-yet unimplemented branches

cmd/tsgo/main.go
```diff
@@ -57,11 +58,18 @@ func main() {
...
-	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext}
+	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext, NoEmit: core.TSTrue}
+
 	currentDirectory, err := os.Getwd()
+	if len(outDir) > 0 {
+		compilerOptions.NoEmit = core.TSFalse
+		compilerOptions.OutDir = tspath.ResolvePath(currentDirectory, outDir)
+	}
+
 	if err != nil {
```
jakebailey: I think this error check needs to be above with the `Getwd`.

PR #173, opened 2024-12-12
Split VFS implementations apart for easier platform-specific additions

jakebailey:
This should make #172 a lot easier; we can just add `WriteFile`/`MkdirAll` to the OS side, and panic on the `io.FS` side until we implement file writing into `fstest.MapFS`.

### DIFF
--- a/internal/vfs/vfstest/vfstest.go
+++ b/internal/vfs/vfstest/vfstest.go
@@ -16,6 +16,8 @@ type mapFS struct {
 	useCaseSensitiveFileNames bool
 }
 
+var _ vfs.RealpathFS = (*mapFS)(nil)
+
 type sys struct {
 	original any
 	realpath string


internal/vfs/os.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var isFileSystemCaseSensitive = sync.OnceValue(func() bool {
+	// win32/win64 are case insensitive platforms
+	if runtime.GOOS == "windows" {
+		return false
+	}
+
+	// If the current executable exists under a different case, we must be case-insensitve.
+	if _, err := os.Stat(swapCase(os.Args[0])); os.IsNotExist(err) {
```
jakebailey, line 36: Rereading this code, I copied it from main originally and didn't realize that it's subtly wrong because `os.Args[0]` is not guaranteed to actually point to anywhere; should be `os.Executable` if anything.

jakebailey: #175

PR #174, opened 2024-12-12
More regex error recovery in reScanSlashToken

sandersn:
This still does NOT include the ability to log diagnostics from https://github.com/microsoft/TypeScript/pull/55600. But it does match tsc's smarter regex-end detection for unterminated regexes (which you get a lot when parsing JSX as JS).

Note that `s.pos++` gets pulled back out of the main loop into the post-loop processing, where it was in tsc.

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -881,7 +881,7 @@ func (s *Scanner) ReScanSlashToken() ast.Kind {
 					break
 				}
 			}
-			s.errorAt(diagnostics.Unterminated_regular_expression_literal, s.tokenStart, s.pos - s.tokenStart);
+			s.errorAt(diagnostics.Unterminated_regular_expression_literal, s.tokenStart, s.pos-s.tokenStart)
 		} else {
 			// Consume the slash character
 			s.pos++


internal/scanner/scanner.go
```diff
@@ -835,12 +835,57 @@ func (s *Scanner) ReScanSlashToken() ast.Kind {
...
+							break
+						}
+					}
+				}
+				s.pos += size
+			}
+			// Whitespaces and semicolons at the end are not likely to be part of the regex
+			for stringutil.IsWhiteSpaceLike(s.charAt(-1)) || s.charAt(-1) == ';' {
+				s.pos--
+			}
```
jakebailey: Are there any whitespace characters which are not 1 byte in size? I suspect there are such that this `s.pos--` would be wrong.

ahejlsberg: Yes, there are multiple.

jakebailey: We chatted on Teams about this but the "fix" is probably going to be to use `utf8.DecodeLastRuneInString` on a sliced version the text (which we do the opposite way already for char reading).

We should probably make `charAt` reject negative numbers for this reason; I was surprised when it actually worked.

internal/scanner/scanner.go
```diff
@@ -835,12 +835,62 @@ func (s *Scanner) ReScanSlashToken() ast.Kind {
...
+				ch, size := s.charAndSize()
+				if inEscape {
+					inEscape = false
+				} else if ch == '\\' {
+					inEscape = true
+				} else if ch == '[' {
+					characterClassDepth++
+				} else if ch == ']' && characterClassDepth != 0 {
+					characterClassDepth--
+				} else if !(characterClassDepth != 0) {
```
```suggestion
				} else if characterClassDepth == 0 {
```

internal/scanner/scanner.go
```diff
@@ -835,12 +835,62 @@ func (s *Scanner) ReScanSlashToken() ast.Kind {
...
+			}
+			// Whitespaces and semicolons at the end are not likely to be part of the regex
+			for {
+				ch, size := utf8.DecodeLastRuneInString(s.text[:s.pos])
+				if stringutil.IsWhiteSpaceLike(ch) || ch == ';' {
+					s.pos -= size
+				} else {
+					break
+				}
+			}
```
jakebailey, line 883: The old code had a diagnostic error here; is that missing intentionally?

jakebailey: For reference:

```ts
error(Diagnostics.Unterminated_regular_expression_literal, tokenStart, pos - tokenStart);
```

internal/scanner/scanner.go
```diff
@@ -835,12 +835,62 @@ func (s *Scanner) ReScanSlashToken() ast.Kind {
...
+							groupDepth--
+						} else if ch == ')' || ch == ']' || ch == '}' {
+							// We encountered an unbalanced bracket outside a character class. Treat this position as the end of regex.
+							break
+						}
+					}
+				}
+				s.pos += size
+			}
+			// Whitespaces and semicolons at the end are not likely to be part of the regex
```
DanielRosenwasser, line 875: The old scanner did this?

DanielRosenwasser: Wow, TIL

sandersn: Yep, that's the part that was wrong in the baselines.

internal/scanner/scanner.go
sandersn: I wish ts-to-go fixed these up further.

internal/scanner/scanner.go
sandersn: It was intentional, but wrong. I mis-categorised this as part of microsoft/Typescript#55600 and skipped it. I'll add it back.

jakebailey:
(Approved but this needs a format / `hereby lint --fix`.)

PR #177, opened 2024-12-13
More type checking

ahejlsberg:
In this PR:

* Mapped resolution, instantiation, relations, inference.
* Non-null assertions.
* `new` call expressions.
* More contextual types.
* Late-bound symbols.
* More type constraints.
* This-less types.
* Several small issues fixed.
* Sorting of program files to ensure stable ordering between runs.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -12880,7 +12880,35 @@ func (c *Checker) getApparentType(t *Type) *Type {
 }
 
 func (c *Checker) getApparentTypeOfMappedType(t *Type) *Type {
-	return t // !!!
+	m := t.AsMappedType()
+	if m.resolvedApparentType == nil {
+		m.resolvedApparentType = c.getResolvedApparentTypeOfMappedType(t)
+	}
+	return m.resolvedApparentType
+}
+
+func (c *Checker) getResolvedApparentTypeOfMappedType(t *Type) *Type {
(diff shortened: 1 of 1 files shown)


internal/compiler/program.go
```diff
@@ -87,6 +87,11 @@ func NewProgram(options ProgramOptions) *Program {
 
 	p.processRootFiles(fileInfos)
 
+	// Sort files by path so we get a stable ordering between runs
+	slices.SortFunc(p.files, func(f1 *ast.SourceFile, f2 *ast.SourceFile) int {
+		return strings.Compare(string(f1.Path()), string(f2.Path()))
+	})
```
jakebailey, line 93: Doing the "correct" thing here (the graph walk) is next on @EricCornelson's list, I think.

internal/compiler/checker.go
```diff
@@ -398,209 +398,220 @@ const (
...
+const (
+	IterationTypeKindYield IterationTypeKind = iota
+	IterationTypeKindReturn
+	IterationTypeKindNext
+)
+
 // Checker
 
 type Checker struct {
-	program                            *Program
```
jakebailey, line 404: Is there any chance you can break all of these fields up into groups so we don't keep reformatting them all as one giant thing? 😅

internal/compiler/program.go
EricCornelson: Correct!

PR #182, opened 2024-12-16
Add finishNode when parsing malformed property declaration

sandersn:
Without finishNode, the missing identifier has pos=end=0.


sandersn:
Hah, we do. I thought there was a reason not to use it, but it just looks like an oversight. Maybe there was in the old compiler.

jakebailey:
There are a few other cases where we create an empty identifier which could use `createMissingIdentifier`, I think.

sandersn:
I left the ones in parseImportIdentifier and createIdentifierWithDiagnostic in place, because they don't use `pos=p.nodePos()`

PR #183, opened 2024-12-16
Ensure deterministic file ordering in program

EricCornelson:
Notable changes:
- Add structure in program to ensure that file ordering is deterministic, even when we are parsing in parallel
- Fix a bug where only the first pragma in a file was parsed

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -183,25 +183,25 @@ func (t *parallelParseTask) createSubTasks(count int) {
 }
 
 func (p *Program) processRootFiles(rootFiles []FileInfo) {
-	ctx := newParallelParseContext(p.programOptions.SingleThreaded, len(rootFiles))
+	parseContext := newParallelParseContext(p.programOptions.SingleThreaded, len(rootFiles))
 
 	absPaths := make([]string, 0, len(rootFiles))
 	for _, fileInfo := range rootFiles {
 		absPath := tspath.GetNormalizedAbsolutePath(fileInfo.Name, p.currentDirectory)
-		ctx.processedFileNames.Add(absPath)
+		parseContext.processedFileNames.Add(absPath)
 		absPaths = append(absPaths, absPath)
 	}
 
 	for i, absPath := range absPaths {
-		p.startParseTask(absPath, ctx, &ctx.rootTasks[i])
+		p.startParseTask(absPath, parseContext, &parseContext.rootTasks[i])
 	}
 
-	ctx.wait()
-	p.addAllFilesToProgram(ctx)
+	parseContext.wait()
+	p.addAllFilesToProgram(parseContext)
 }
 
-func (p *Program) startParseTask(fileName string, ctx *parallelParseContext, task *parallelParseTask) {
-	ctx.run(func() {
+func (p *Program) startParseTask(fileName string, parseContext *parallelParseContext, task *parallelParseTask) {
+	parseContext.run(func() {
 		normalizedPath := tspath.NormalizePath(fileName)
 		file := p.parseSourceFile(normalizedPath)
 		p.collectExternalModuleReferences(file)
@@ -215,29 +215,29 @@ func (p *Program) startParseTask(fileName string, ctx *parallelParseContext, tas
 
 		filesToParse = append(filesToParse, p.resolveImportsAndModuleAugmentations(file)...)
 		task.file = file
-		p.processFiles(filesToParse, task, ctx)
+		p.processFiles(filesToParse, task, parseContext)
 	})
 }
 
-func (p *Program) processFiles(files []string, task *parallelParseTask, ctx *parallelParseContext) {
+func (p *Program) processFiles(files []string, task *parallelParseTask, parseContext *parallelParseContext) {
 	if len(files) > 0 {
 		task.createSubTasks(len(files))
 
-		ctx.lock()
-		defer ctx.unlock()
+		parseContext.lock()
+		defer parseContext.unlock()
 		for i, fileName := range files {
-			if ctx.addIfNotProcessed(fileName) {
-				p.startParseTask(fileName, ctx, &task.subTasks[i])
+			if parseContext.addIfNotProcessed(fileName) {
+				p.startParseTask(fileName, parseContext, &task.subTasks[i])
 			}
 		}
 	}
 }
 
-func (p *Program) addAllFilesToProgram(ctx *parallelParseContext) {
-	p.files = make([]*ast.SourceFile, 0, ctx.processedFileCount())
-	p.filesByPath = make(map[tspath.Path]*ast.SourceFile, ctx.processedFileCount())
+func (p *Program) addAllFilesToProgram(parseContext *parallelParseContext) {
+	p.files = make([]*ast.SourceFile, 0, parseContext.processedFileCount())
+	p.filesByPath = make(map[tspath.Path]*ast.SourceFile, parseContext.processedFileCount())
 
-	p.addAllFilesToProgramWorker(ctx.rootTasks)
+	p.addAllFilesToProgramWorker(parseContext.rootTasks)
 }
 
 func (p *Program) addAllFilesToProgramWorker(tasks []parallelParseTask) {


internal/compiler/program.go
```diff
@@ -134,25 +133,40 @@ func (p *Program) bindSourceFiles() {
...
+type ParallelParseContext struct {
+	mu                 sync.Mutex
+	wg                 *core.WorkGroup
+	processedFileNames core.Set[string]
+}
+
+type ParallelParseTask struct {
+	file     *ast.SourceFile
+	subTasks []ParallelParseTask
+}
```
jakebailey: These types can be lowercased.

internal/compiler/program.go
```diff
@@ -134,25 +133,40 @@ func (p *Program) bindSourceFiles() {
...
+
 func (p *Program) processRootFiles(rootFiles []FileInfo) {
-	wg := core.NewWorkGroup(p.programOptions.SingleThreaded)
+	ctx := &ParallelParseContext{}
+	ctx.wg = core.NewWorkGroup(p.programOptions.SingleThreaded)
```
jakebailey: I would create a `func newParallelParseContext(singleThreaded bool) *parallelParseContext` instead.

internal/compiler/program.go
```diff
@@ -134,25 +133,40 @@ func (p *Program) bindSourceFiles() {
...
-	for _, absPath := range absPaths {
-		p.startParseTask(absPath, wg)
+	tasks := make([]ParallelParseTask, len(absPaths))
+
+	for i, absPath := range absPaths {
+		p.startParseTask(absPath, ctx, &tasks[i])
 	}
 
-	wg.Wait()
+	ctx.wg.Wait()
```
jakebailey: I would wrap these up as methods on the structs instead; when it's behavior like this it feels a little dubious to reach inward.

internal/compiler/program.go
```diff
@@ -134,25 +133,40 @@ func (p *Program) bindSourceFiles() {
...
+	p.addAllFilesToProgram(tasks)
 }
 
-func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
-	wg.Run(func() {
+func (p *Program) startParseTask(fileName string, ctx *ParallelParseContext, task *ParallelParseTask) {
+	ctx.wg.Run(func() {
```
jakebailey: Same here; you could probably make this a `run` method.

internal/compiler/program.go
```diff
@@ -165,21 +179,61 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup) {
...
 			}
 		}
-	})
+	}
+}
+
+func (p *Program) addAllFilesToProgram(tasks []ParallelParseTask) {
+	if tasks == nil {
+		return
+	}
```
jakebailey: Why the nil check? Usually distinguishing a nil slice from an empty slice indicates a bug.

jakebailey: Hm, I see you're using `nil` as some sort of sentinel; I'd almost prefer a bool which indicates what it's doing, e.g. is it some sort of "is root" or "is leaf"?

internal/compiler/program_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/microsoft/typescript-go/internal/vfs/vfstest"
+	"gotest.tools/v3/assert"
+)
+
+func TestProgramFileOrdering(t *testing.T) {
+	t.Parallel()
+	fs := fstest.MapFS{}
+
+	files := map[string]string{
+		"c:/dev/src/index.ts":          "/// <reference path='c:/dev/src2/a/5.ts' />\n/// <reference path='c:/dev/src2/a/10.ts' />",
```
jakebailey: Can you do a test which uses imports too?

internal/compiler/program_test.go
```diff
@@ -0,0 +1,69 @@
...
+		"c:/dev/src2/a/5.ts",
+		"c:/dev/src2/a/b/c/d/e/f/6.ts",
+		"c:/dev/src2/a/b/c/d/e/7.ts",
+		"c:/dev/src2/a/b/c/d/e/8.ts",
+		"c:/dev/src2/a/b/c/d/9.ts",
+		"c:/dev/src2/a/10.ts",
+		"c:/dev/src/index.ts",
+	}
+
+	assert.DeepEqual(t, expectedOrder, actualOrder)
```
jakebailey: This is reversed, I think.

internal/compiler/program.go
jakebailey: Also, `subTasks` could probably be `[]*parallelParseTask` to prevent a bunch of copying, especially since you're copying a slice and also potentially modifying it sometimes?

jakebailey:
FWIW it does seem a _little_ like this searching process could instead be defined on its own struct which manages collecting files in a sorted way, then the program calls that and then reads out the files list from that, but maybe that's a little too abstract.

internal/compiler/program_test.go
jakebailey: And, one for cycles? It's not obvious to me just from this whether or not this has a cycle in it or not, but we should definitely allow it.

internal/compiler/program_test.go
```diff
@@ -0,0 +1,69 @@
...
+		"c:/dev/src2/a/10.ts":          `/// <reference path="b/c/d/9.ts" />`,
+		"c:/dev/src2/a/b/c/d/9.ts":     `/// <reference path="e/8.ts" />`,
+		"c:/dev/src2/a/b/c/d/e/8.ts":   `/// <reference path="7.ts" />`,
+		"c:/dev/src2/a/b/c/d/e/7.ts":   `/// <reference path="f/6.ts" />`,
+		"c:/dev/src2/a/b/c/d/e/f/6.ts": `console.log("world!");`,
+	}
+
+	fs["c:/dev/src/index.ts"] = &fstest.MapFile{
+		Data: []byte("/// <reference path=\"a/1.ts\" />"),
+	}
```
jakebailey: Why is this block separated from the previous one?

internal/compiler/program_test.go
```diff
@@ -0,0 +1,69 @@
...
+		SingleThreaded: false,
+	})
+
+	actualOrder := []string{}
+	for _, file := range program.files {
+		actualOrder = append(actualOrder, file.FileName())
+	}
+
+	expectedOrder := []string{
+		"c:/dev/src2/a/b/c/1.ts",
```
jakebailey: This order doesn't seem right. The initial files are just `c:/dev/src/index.ts`, and that file references `c:/dev/src2/a/5.ts` and `c:/dev/src2/a/10.ts`, so I would very much expect the order to start with:

```
c:/dev/src/index.ts
c:/dev/src2/a/5.ts
c:/dev/src2/a/10.ts
...
```

But your list seems to be reversed, and then reversed in order of references in the same file.

jakebailey: The algorithm I was expecting is:

- Save the initial list of file paths (which will eventually come from include/exclude/files)
- Kick off parse tasks for those initial files
- When done parsing each file, make a list of files _it_ references, save that list.
- Repeat until done

Then, at the end, create an ordered set, populate it with the initial file set, then do a DFS walk over the initial file set, adding into the ordered set, skipping if already present.

That would effectively allow the code written in `main` to exist as-is, with only new sorting code later.

EricCornelson:
> FWIW it does seem a _little_ like this searching process could instead be defined on its own struct which manages collecting files in a sorted way, then the program calls that and then reads out the files list from that, but maybe that's a little too abstract.

I was actually thinking the same thing. It's a pretty well defined task that could be encapsulated neatly

internal/compiler/program_test.go
EricCornelson: I actually based it on the ordering of current tsc to be as consistent as possible. It's post-order depth first because we would immediately step into the references, parse them, and add them to the list before continuing with the current file
image

jakebailey: Wow, not what I thought at all

jakebailey: But, the simplified algorithm I have would still work, right? (Just have to move the "add" later)

EricCornelson: I guess I'm not clear on how that algorithm differs from what I'm doing here,

> Save the initial list of file paths (which will eventually come from include/exclude/files)

Done with the initial list of `ParallelParseTask`s, though we're just saving a pointer to the `SourceFile` rather than the file name. Maybe the name `ParallelParseTask` isn't as accurate as it could be. It's really just a container for the result of a parse task, not so much a task itself.
> Kick off parse tasks for those initial files

done in `processRootFiles`
> When done parsing each file, make a list of files it references, save that list.
    
that's what the subtasks are
> Then, at the end, create an ordered set, populate it with the initial file set, then do a DFS walk over the initial file set, adding into the ordered set, skipping if already present.

That's what we're doing, except we're putting the results into `files` and `filesByPath`, which we could just consolidate into an ordered set.

internal/compiler/program_test.go
EricCornelson: I think this is right. Here's the diff it shows if I change something in `expectedOrder`:

image

internal/compiler/program_test.go
EricCornelson: Whoops, left-over junk. Removed.

internal/compiler/program.go
EricCornelson: Where/when would this slice get copied?

internal/compiler/program.go
EricCornelson: Done!

internal/compiler/program.go
EricCornelson: Done!

internal/compiler/program.go
EricCornelson: Also done

internal/compiler/program.go
EricCornelson: Added a bool for `hasSubTasks`

internal/compiler/program_test.go
jakebailey: The difference is in the implementation, not the result; I'm trying to avoid the concept of a "context" or "task", when the relationship is actually just a map from "path" to "list of files they pulled in", which is a couple extra lines on top of the original code, followed by a post-processing step.

EricCornelson: But we still need somewhere to store those intermediate, un-ordered results, right? Right now, we're just plopping them into `files` on `Program` as we parse them, but we want `files` to be in the right order. I don't really like the idea of `Program` having a bunch of fields that are only used for intermediate steps, but then are held on to indefinitely, (which is why I moved the mutex and the `processedFiles` map into the context object as well).

internal/compiler/program.go
```diff
@@ -134,25 +127,81 @@ func (p *Program) bindSourceFiles() {
...
+}
+
+func (t *parallelParseTask) createSubTasks(count int) {
+	t.subTasks = make([]parallelParseTask, count)
+	t.hasSubTasks = true
+}
+
 func (p *Program) processRootFiles(rootFiles []FileInfo) {
-	wg := core.NewWorkGroup(p.programOptions.SingleThreaded)
+	ctx := newParallelParseContext(p.programOptions.SingleThreaded, len(rootFiles))
```
jakebailey: This is a Go thing, but I would avoid calling variables/parameters `ctx` unless those variables are of type `context.Context`.

internal/compiler/program_test.go
jakebailey: They don't need to be held onto indefinitely; they can be nil'd out, or, only be scoped to that one function call.

I'm just having a hard time knowing how this structure would interact with lib sorting (a la #194), which has a different sorting algorithm but must _also_ interact with "other" file loading, since "other" files can also add lib references which must be correctly sorted and not duplicated.

PR #184, opened 2024-12-17
More type checking

ahejlsberg:
In this PR:

* Mapped type index types.
* Object literal contextual type discrimination.
* Conditional types.
* Binding elements.
* Contextual intersected signatures.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -467,6 +467,7 @@ type Checker struct {
 	requireSymbol                           *ast.Symbol
 	unknownSymbol                           *ast.Symbol
 	resolvingSymbol                         *ast.Symbol
+	unresolvedSymbols                       map[string]*ast.Symbol
 	errorTypes                              map[string]*Type
 	globalThisSymbol                        *ast.Symbol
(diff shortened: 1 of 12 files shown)


internal/compiler/checker.go
```diff
@@ -2790,8 +2811,136 @@ func (c *Checker) checkExpressionEx(node *ast.Node, checkMode CheckMode) *Type {
...
+		return tp.symbol.Name == name
+	})
+}
+
+func getUniqueTypeParameterName(typeParameters []*Type, baseName string) string {
+	for len(baseName) > 1 && baseName[len(baseName)-1] >= '0' && baseName[len(baseName)-1] <= '9' {
+		baseName = baseName[:len(baseName)-1]
+	}
+	index := 1
+	for {
```
jakebailey, line 2945: This seems like it could be a regular for loop but maybe the original code was like this.

ahejlsberg: The ported code was originally `for index := 1; true; index++ {...}`, but oddly the Go compiler thinks that the statement following that `for` loop is reachable and complains about a missing return. I suppose it could be `for index := 1;; index++ {...}`,  but I just rewrote it like you see it in the PR.

jakebailey: Ah, I think I just misunderstood the code. Nevermind!

PR #185, opened 2024-12-17
Add tsconfig parsing

navya9singh:
Things that are currently missing:
- Some more error testing from tsconfigParsing.ts 
- Implementing `extends`


internal/tsoptions/parsingHelpers.go
```diff
@@ -0,0 +1,396 @@
...
+		case "force":
+			result = core.ModuleDetectionKindForce
+		default:
+			result = core.ModuleDetectionKindNone
+		}
+	}
+	return result
+}
+
+func parseModuleKind(json any) core.ModuleKind {
```
iisaduan, line 137: The options that are `CommandLineOptionTypeEnum`, could we just use a enummap look up? https://github.com/microsoft/typescript-go/pull/87/files#diff-d500fbab2661fa05cd218d04045ebd08e279cf097661b68d7387c9a8544f40cfR133

internal/compiler/types.go
```diff
@@ -1167,3 +1167,12 @@ type FileIncludeReason struct {
...
 	Index int
 }
+
+// CharacterCodes
+const (
+	EOF               = -1
+	NullCharacter     = 0
+	MaxAsciiCharacter = 0x7F
+	Asterisk          = 0x2A
+	Question          = 0x3F // ?
```
iisaduan, line 1171: It should be fine to just use '*' (or the other characters) like this https://github.com/microsoft/typescript-go/pull/143/files#diff-0edf250bdd38c3167ad16cf58801b6e50d5ae2ee752717f36d835f2e92f8355fR188

internal/compiler/types.go
```diff
@@ -1167,3 +1167,12 @@ type FileIncludeReason struct {
 	Kind  FileIncludeKind
 	Index int
 }
+
+// CharacterCodes
+const (
+	EOF               = -1
+	NullCharacter     = 0
+	MaxAsciiCharacter = 0x7F
```
iisaduan, line 1169: Where/How are EOF, NullCharacter, or MaxAscii used?

internal/tsoptions/declscompiler.go
```diff
@@ -296,7 +296,6 @@ var optionsDeclarations = []CommandLineOption{
 		Name: "lib",
 		Kind: CommandLineOptionTypeList,
 		// elements: &CommandLineOption{
-		// 	name:                    "lib",
```
iisaduan, line 299: Not that it really matters since this is a comment, but these sub options need names (structurally), so the repeated "lib" was intentional

navya9singh:
Closing this pr due to missing files, even after reverting the commits.

PR #188, opened 2024-12-17
Fix Loc of synthetic JSX closing elements

sandersn:
In malformed JSX, some synthetic elements are invented after parsing has finished and inserted earlier in the tree to attempt to close unclosed elements. These elements need to specify both a saved pos and saved end, so I created a new function `finishNodeWithEnd` for that.

I remember also creating this function for parsing JSDoc, which occasionally does the same thing, so I expect it'll be used more later.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -5429,8 +5429,7 @@ func (p *Parser) internIdentifier(text string) {
 }
 
 func (p *Parser) finishNode(node *ast.Node, pos int) {
-	node.Loc = core.NewTextRange(pos, p.nodePos())
-	node.Flags |= p.contextFlags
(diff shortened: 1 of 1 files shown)


internal/parser/parser.go
```diff
@@ -5430,6 +5433,11 @@ func (p *Parser) finishNode(node *ast.Node, pos int) {
 	node.Flags |= p.contextFlags
 }
 
+func (p *Parser) finishNodeWithEnd(node *ast.Node, pos int, end int) {
```
jakebailey, line 5435: It seems like that `finishNode` above should just call this function, rather than duplicating this code.

sandersn: I didn't trust the compiler to inline it, but should have.

PR #191, opened 2024-12-17
Add symbol baseliner

gabritto:


internal/testutil/baseline/error_baseline.go
```diff
@@ -19,12 +19,6 @@ import (
 // IO
 const harnessNewLine = "\r\n"
 
-var (
```
gabritto, line 22: Moved to `internal/testutil/baseline/util.go`

internal/ast/utilities.go
```diff
@@ -1123,7 +1123,7 @@ func IsDeclaration(node *Node) bool {
 
 // True if `name` is the name of a declaration node
 func IsDeclarationName(name *Node) bool {
-	return !IsSourceFile(name) && !IsBindingPattern(name) && IsDeclaration(name.Parent)
+	return !IsSourceFile(name) && !IsBindingPattern(name) && IsDeclaration(name.Parent) && name.Parent.Name() == name
```
gabritto, line 1126: This is a bug fix.

internal/testutil/baseline/baseline.go
```diff
@@ -15,11 +15,11 @@ type Options struct {
 
 const NoContent = "<no content>"
 
-func Run(t testing.TB, fileName string, actual string, opts Options) {
+func Run(t *testing.T, fileName string, actual string, opts Options) {
```
gabritto, line 18: Had to make this change from `testing.TB` to `testing.T` because I wanted to use `t.Run` in one symbol baseline function and so I changed it everywhere else, too.

internal/testutil/baseline/error_baseline.go
```diff
@@ -253,19 +246,6 @@ func iterateErrorBaseline(t testing.TB, inputFiles []*TestFile, inputDiagnostics
 	return result
 }
 
-func checkDuplicatedFileName(resultName string, dupeCase map[string]int) string {
```
gabritto, line 256: I realized this was essentially unused in the old test infra.

internal/compiler/checker.go
```diff
@@ -2366,7 +2366,7 @@ func (c *Checker) checkVarDeclaredNamesNotShadowed(node *ast.Node) {
 				// here we know that function scoped variable is "shadowed" by block scoped one
 				// a var declatation can't hoist past a lexical declaration and it results in a SyntaxError at runtime
 				if !namesShareScope {
-					name := c.symbolToString(localDeclarationSymbol)
+					name := c.SymbolToString(localDeclarationSymbol)
```
jakebailey: I believe there's a difference in the old TS between "public" and "internal" `symbolToString`; perhaps these should all remain as `symbolToString` and then we add a new exported `SymbolToString` which represents the public version? See also a PR like https://github.com/microsoft/TypeScript/pull/60729

jakebailey: The same might go for `getSymbolAtLocation`, actually.

jakebailey:
Is there a test we could write for this "now" to verify that it works?

jakebailey:
Ah, #115 seems to include it.

internal/compiler/checker.go
```diff
@@ -8450,7 +8450,7 @@ func (c *Checker) getTargetOfModuleDefault(moduleSymbol *ast.Symbol, node *ast.N
 	// 		}
 	// 		exportEqualsSymbol := moduleSymbol.exports.get(ast.InternalSymbolNameExportEquals)
 	// 		exportAssignment := exportEqualsSymbol.valueDeclaration
-	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.symbolToString(moduleSymbol), compilerOptionName)
+	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.SymbolToString(moduleSymbol), compilerOptionName)
```
```suggestion
	// 		err := c.error(node.name, Diagnostics.Module_0_can_only_be_default_imported_using_the_1_flag, c.symbolToString(moduleSymbol), compilerOptionName)
```

PR #194, opened 2024-12-18
Add new "bundled" package to provide access to bundled files, add lib.d.ts support

jakebailey:
This PR adds a new "bundled" package, which provides functions that give access to "bundled" files. In our package, we "bundle" lib.d.ts files, but also eventually translations. I've copied the logic from the previous compiler's `Herebyfile.mjs` into a new `go:generate` script, just like we did diagnostic generation.

To ensure that `go test` and `go run github.com/microsoft/typescript-go/cmd/tsgo@latest` work, this package uses `go:embed` to embed the bundled files into the binary by default. A VFS wrapper then detects when these files are accessed, and redirects to those embedded files.

I've also made the change to make `hereby build` (and now, `local`) build to `built/local` instead of `bin`, and then also copy in `lib.d.ts` files like we did in the previous package.

When building to `built/local` (where our final package bits will actually be), we build with `-tags=noembed`. This disables the embedding, instead using the executable's path on disk to determine where files live just like the previous TypeScript package. This will be needed since people can _technically_ patch these files, and most importantly expect to be able to navigate to them.

With bundled files implemented, I've now added `lib`/`noLib` support to `Program`, including the correct sorting of lib files.

Tests run with embedding enabled by default, but we may also consider variants which run with `-tags=noembed`; in this case, the `bundled` package detects that it's running in a test, and then use its own source file location (like the `repo` package) to then find the bundled files. A special function `TestingLibPath` can be used to explicitly ask that dir in tests, in case we want to copy the files into a testfs or something.

This PR depends on #193, since as a part of this PR I discovered that our list of libs was ported from the wrong version of TypeScript.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -27,6 +27,7 @@ var (
 	printTypes       = false
 	pretty           = true
 	listFiles        = false
+	noLib            = false
 	pprofDir         = ""
 	outDir           = ""
 )
@@ -60,12 +61,16 @@ func main() {
 	flag.BoolVar(&printTypes, "t", false, "Print types defined in main.ts")
 	flag.BoolVar(&pretty, "pretty", true, "Get prettier errors")
 	flag.BoolVar(&listFiles, "listfiles", false, "List files in the program")
+	flag.BoolVar(&noLib, "nolib", false, "Do not load lib.d.ts files")
 	flag.StringVar(&pprofDir, "pprofdir", "", "Generate pprof CPU/memory profiles to the given directory")
 	flag.StringVar(&outDir, "outdir", "", "Emit to the given directory")
 	flag.Parse()
 
 	rootPath := flag.Arg(0)
 	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext, NoEmit: core.TSTrue}
+	if noLib {
+		compilerOptions.NoLib = core.TSTrue
+	}
 
 	currentDirectory, err := os.Getwd()
 	if err != nil {


internal/parser/parser.go
```diff
@@ -5975,7 +5975,7 @@ func attachFileToDiagnostics(diagnostics []*ast.Diagnostic, file *ast.SourceFile
 func getCommentPragmas(sourceText string) (pragmas []ast.Pragma) {
 	for commentRange := range scanner.GetLeadingCommentRanges(sourceText, 0) {
 		comment := sourceText[commentRange.Pos():commentRange.End()]
-		pragmas = extractPragmas(commentRange, comment)
+		pragmas = append(pragmas, extractPragmas(commentRange, comment)...)
```
gabritto, line 5978: Is this a bug fix?

internal/bundled/embed.go
```diff
@@ -0,0 +1,95 @@
...
+		return "", "", false
+	}
+	return scheme, rest, true
+}
+
+func libPath() string {
+	return scheme + "libs"
+}
+
+//go:embed libs
```
gabritto, line 29: Kinda surprising that this directive does all the heavy lifting.

internal/parser/parser.go
jakebailey: Yeah, this is already in #183 too. Could be pulled out to another PR early.

internal/bundled/embed.go
jakebailey: In what way?

gabritto: In a nice way that Go has this 😊

jakebailey: Ah, yeah.

C is actually gaining this too, though, just for single files IIRC (since there's no such thing as an FS abstraction that would ever be in the C spec, just "hey it's bytes").

internal/compiler/program.go
```diff
@@ -78,50 +86,94 @@ func NewProgram(options ProgramOptions) *Program {
...
+
+	otherFiles := walkFiles(p.host.FS(), p.rootPath, extensions)
+
+	// Process otherFiles first so breakpoints on user code are more likely to be hit first.
+	p.processRootFiles(otherFiles, false)
+	p.processRootFiles(libs, true)
+
+	slices.SortFunc(p.libFiles, func(f1 *ast.SourceFile, f2 *ast.SourceFile) int {
+		return cmp.Compare(p.getDefaultLibFilePriority(f1), p.getDefaultLibFilePriority(f2))
+	})
```
jakebailey, line 131: @EricCornelson this is part of the reason why I was advocating for something simpler in #183; we need to sort these two things differently, and it'd be more convenient if the sorting there were not a part of the overall "find files and load them" step, since we need to do two different orderings to handle lib files versus other files. The old TS Program implementation did it similar to what I have done here.

(Though, I don't like that `libFiles` and `otherFiles` are fields on `Program` when they could technically be a returned value from `processRootFiles`, but that's just something I did here to emulate one of the ways the old code worked.)

jakebailey:
Hm, I'm writing a test for lib file ordering using this PR, and it breaks in `noembed` mode because without the wrapper FS, the files just don't actually exist within the fake FS. I might need to think a little harder about how we can do testing...

@gabritto you're still copying lib.d.ts files into a `fstest.Map`, right?

jakebailey:
The only remaining work here is specifically just "how will this work in testing when the files aren't embedded and `WithEmbedded` is in use".

It's possible we could simply _not_ bother with this for testing, and say that the embedded files are all we use in testing, otherwise, I'll need to make a helper that is able to copy all of the lib files into a tempfs or something on-demand.

jakebailey:
Added `--noLib` to tsgo, since that's the "current" behavior which may be helpful for testing.

cmd/tsgo/main.go
```diff
@@ -59,12 +61,16 @@ func main() {
 	flag.BoolVar(&printTypes, "t", false, "Print types defined in main.ts")
 	flag.BoolVar(&pretty, "pretty", true, "Get prettier errors")
 	flag.BoolVar(&listFiles, "listfiles", false, "List files in the program")
+	flag.BoolVar(&noLib, "nolib", false, "Do not load lib.d.ts files")
 	flag.StringVar(&pprofDir, "pprofdir", "", "Generate pprof CPU/memory profiles to the given directory")
 	flag.StringVar(&outDir, "outdir", "", "Emit to the given directory")
 	flag.Parse()
 
 	rootPath := flag.Arg(0)
 	compilerOptions := &core.CompilerOptions{Strict: core.TSTrue, Target: core.ScriptTargetESNext, ModuleKind: core.ModuleKindNodeNext, NoEmit: core.TSTrue}
```
jakebailey, line 70: The hard-coded target here of esnext means that we'll load `lib.esnext.full.d.ts`, which includes basically everything (including DOM). I'm not sure that this is _bad_; it's certainly _correct_, but it will mean that our perf testing will imply things are slightly slower where we were previously not loading the DOM at all.

But, measuring on a small and large project, the difference is some 2%, but with error bars much larger than that.

This will eventually not matter when we start loading tsconfigs.

PR #195, opened 2024-12-18
Type erasing transformer prototype

rbuckton:
This is a prototype of a simple type erasing transformer intended to work with the pretty printer as-is. It does not currently transform parameter property initializers (aside from stripping off the modifiers), nor does it transform `enum` or `namespace` as of yet.

This prototype attempts to illustrate the minimal functionality necessary to implement a transforming emitter in the new compiler codebase. At a minimum, node-specific `VisitEachChild` functionality is necessary to properly update nodes in the tree with the correct context without disrupting the existing parse tree. As in the old compiler, `VisitEachChild` differs from `ForEachChild` in the following ways:

- _Non-shortcutting_ - `ForEachChild` is shortcutting, while `VisitEachChild` is not. When `ForEachChild`'s visitor callback returns `true`, visiting stops immediately.
- _Contextual_ - `ForEachChild` maintains no context related to visitor's position in the tree, making it unsuitable for updating specific properties of a parent node. `VisitEachChild` maintains the visitor's position within the tree such that replacing a node can update the property being visited.

Unfortunately, it is not possible to implement `VisitEachChild` in terms of `ForEachChild` given the aforementioned behaviors, though it is possible (but not recommended) to implement `ForEachChild` in terms of `VisitEachChild` as you can ignore context and emulate shortcutting:

```go
func ForEachChild(node *Node, cb Visitor) bool {
    v := NodeVisitor{}
    v.Visit = func (node *Node) *Node {
        if cb(node) {
            v.Visit = nil // setting to nil both emulates shortcutting and indicates result
        }
        return node
    }
    v.VisitEachChild(node)
    return v.Visit == nil
}
```

While this does work, it is far slower than the current version of `ForEachChild` which does perform shortcutting. 

It may be feasible to build both `ForEachChild` and `VisitEachChild` in terms of a third mechanism, but my current experiments have shown that it is difficult to do this efficiently due to limitations in Go's generics. It may be possible to do this through reflection, but that also would be far slower than maintaining two independent implementations. I am still investigating approaches to unify these two mechanisms.

~~I should also note that this prototype avoids introducing the `Update*` methods proposed in #54, however due to the repetition between both `VisitEachChild` and `TypeEraserTransformer.visit`, it may be worthwhile to reintroduce them.~~

Commit ceeb3e904443217f4435938ee2214e85704fb6f0 introduces update methods. For the version of this PR without update methods, see https://github.com/microsoft/typescript-go/pull/195/files/1d311e1e31f88461f8293c9bfd623b38535bc074


jakebailey:
Just in terms of approach, I do think this is sound; it's just like the old code and it seems like things would be very straightforward to port.

I do understand the need for both `Visit` and `ForEachChild`, and I'm not worried about duplicate code; we really do want this all to be generated at some point in the future (my attempt was not so good and I've personally been working on other stuff that's been more critical).

jakebailey:
Are we waiting for some kind of consensus on this approach, or should I treat this as something to approve?

internal/transformers/typeeraser_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+func formatDiagnostics(diagnostics []*ast.Diagnostic) string {
+	var b strings.Builder
+	for _, d := range diagnostics {
+		formatDiagnostic(&b, d, 0)
+	}
+	return b.String()
+}
+
+func formatDiagnostic(b *strings.Builder, d *ast.Diagnostic, level int) {
```
jakebailey, line 111: I think we have these helpers elsewhere now, even in a place that can be imported without a cycle. (though this could be a `transformers_test` package and avoid it)

rbuckton:
> Are we waiting for some kind of consensus on this approach, or should I treat this as something to approve?

I'm still waiting on consensus for this approach.

jakebailey:
Go stacks are resizable, so overflows are much less likely to happen.

I'm not sure channels are a good fit for this; they're pretty expensive and probably shouldn't be used in a hot path (though, they can be "less slow" when there's exactly one reader and writer, as the runtime can directly hand the value off).

I'm also not sure how we would do the above without spawning a bunch of goroutines either, but maybe I'd have to see a more fleshed out example.

A similar construct would be to use go's iterators, since it has a coroutine implementation (`coro`, I think public in 1.24?) in the runtime, but both channels/iterators have downsides for type safety given they mean no more specific type than `*Node` (which I understand is effectively how safe we are now, but it closes the door to analysis based code flow analysis, which I plan to do in the future via `go/ssa` and so on).

internal/ast/visitor.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"slices"
+)
+
+// NodeVisitor
+
+type NodeVisitor struct {
+	Visit   func(node *Node) *Node // The callback used to visit a node
+	Factory NodeFactory            // A NodeFactory used to produce new nodes when passed to VisitEachChild
```
jakebailey, line 11: Does `Factory` need to be exported? Are you expecting someone to set it?

Copying `NodeFactory` is probably a bad idea, so don't expect that we'd do it.

internal/transformers/typeeraser.go
```diff
@@ -0,0 +1,204 @@
+package transformers
```
jakebailey, line 1: Just as a meta comment, but I would expect that we would not pool every single transform into a single package, no? Or will there be a load of shared functionality?

rbuckton: There is sometimes shared functionality, but I don't expect the set of transformers we will have in the future to be as extensive as what we had previously. A package per transform seems like overkill at the moment, but I don't have a strong preference.

jakebailey: Easy to change later; certainly not a blocking thing.

internal/ast/visitor.go
rbuckton: `NodeFactory` needs to be public so that a transformer can use it to make new nodes, and so that the `VisitEachChild` functions can use it to update nodes, and ideally share the same `Pool` for things like identifiers. I am wondering if this should be a `*NodeFactory` instead, though. Making it a `*NodeFactory` means we can share the same pools between successive transforms, but it also means we need to ensure it is allocated prior to use.

jakebailey: Oh, duh, I forgot that it was used outside this package.

If we wanted to be fancy, we could just make `NodeFactory` a method which populates an unexported field if it's nil, and then provide a way to construct a visitor. But, given NodeVisitor is an embeddable struct, that is a little less exiting. But I guess every visitor is _already_ having to assign over the `Visit` function, so an explicit constructor is probably also okay.

internal/ast/ast.go
```diff
@@ -1180,6 +1197,7 @@ type NodeDefault struct {
 
 func (node *NodeDefault) AsNode() *Node                                     { return &node.Node }
 func (node *NodeDefault) ForEachChild(v Visitor) bool                       { return false }
+func (node *NodeDefault) VisitEachChild(v *NodeVisitor) *Node               { return node.AsNode() }
```
jakebailey, line 1203: Should this panic or something such that we for sure implement it for every node kind?

rbuckton: We don't do that for `ForEachChild`, and there are a few nodes that would need to copy this implementation. Panicking also wouldn't catch cases where we failed to change an `Update*` method. Are we are still planning on generating these in the long term?

internal/ast/visitor.go
rbuckton: I'd like to delay changing this until I put up the PR for the enum transform as that may better inform how we want to handle this case.

jakebailey: Oh, yeah, this is not a blocking change either.

internal/ast/ast.go
jakebailey: We are planning on generating them; we can keep it like this for now since it would effectively just lead to "nothing happening", not an infinite loop or something.

PR #198, opened 2024-12-19
Implement file writing in vfstest

jakebailey:
This allows the `vfstest` package's VFS to support file writing to the underlying map. This VFS is safe for concurrent use.

### DIFF
--- a/internal/vfs/vfstest/vfstest_test.go
+++ b/internal/vfs/vfstest/vfstest_test.go
@@ -2,10 +2,15 @@ package vfstest
 
 import (
 	"io/fs"
+	"math/rand/v2"
+	"runtime"
+	"slices"
+	"sync"
 	"testing"
 	"testing/fstest"
 
 	"github.com/microsoft/typescript-go/internal/testutil"
+	"github.com/microsoft/typescript-go/internal/vfs"
 	"gotest.tools/v3/assert"
 )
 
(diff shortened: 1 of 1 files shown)


internal/vfs/iofs.go
```diff
@@ -13,13 +13,9 @@ type RealpathFS interface {
 	Realpath(path string) (string, error)
 }
 
-type WriteFileFS interface {
+type WritableFS interface {
```
jakebailey, line 16: I've opted to merge these two together since they are effectively both "write to the FS in some way", so I expect any implementation to have both or neither.

internal/vfs/vfstest/vfstest.go
```diff
@@ -133,7 +199,10 @@ func (f *readDirFile) ReadDir(n int) ([]fs.DirEntry, error) {
 }
 
 func (m *mapFS) Open(name string) (fs.File, error) {
-	f, err := m.m.Open(tspath.GetCanonicalFileName(name, m.useCaseSensitiveFileNames))
+	m.mu.RLock()
```
rbuckton, line 202: mutexes are a bit of a sledgehammer as they block *all* concurrent readers, even if there are no writers. Wouldn't a `sync.RWMutex` be better? Or instead, we could use a `sync.Map` to support concurrent readers *and* writers.

jakebailey: It is an `RWMutex`; note `RLock` here.

I can't use a `sync.Map` since this piggy backs on top of `fstest.Map`, but there is a world where I could copy all of their code to make that happen.

rbuckton: > It is an `RWMutex`; note `RLock` here.

Ah, sorry. I overlooked that, thanks.

> I can't use a `sync.Map` since this piggy backs on top of `fstest.Map`, but there is a world where I could copy all of their code to make that happen.

In the long run we may want to, as any vfs-based tests that depend on parallel emit will be slower than necessary as all writes become serialized.

jakebailey: Yeah, though it may not matter since we'll have multiple tests running in parallel, so the CPU will already be chugging though `runtime.GOMAXPROCS(0)` tests at once.

PR #199, opened 2024-12-19
More type checking

ahejlsberg:
In this PR:

* Infer return types from bodies
* Infer type predicates from bodies
* Instantiation expressions
* Substitution types
* Type simplification
* Reverse mapped types in inference
* Implied types from binding patterns
* Object literal widening
* Various expression checking
* Intra-expression inference sites
* Fixed a few porting bugs

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -3300,7 +3300,7 @@ func isSpreadIntoCallOrNew(node *ast.Node) bool {
 func (c *Checker) checkQualifiedName(node *ast.Node, checkMode CheckMode) *Type {
 	left := node.AsQualifiedName().Left
 	var leftType *Type
-	if isPartOfTypeQuery(node) && isThisIdentifier(left) {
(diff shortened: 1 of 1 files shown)


internal/compiler/checker.go
```diff
@@ -20614,16 +21990,33 @@ func (c *Checker) convertAutoToAny(t *Type) *Type {
...
- * Promise-like type; otherwise, it is the type of the expression. If the "promised
- * type" is itself a Promise-like, the "promised type" is recursively unwrapped until a
- * non-promise type is found.
- *
- * This is used to reflect the runtime behavior of the `await` keyword.
- */
+// Gets the "awaited type" of a type.
+// @param type The type to await.
+// @param withAlias When `true`, wraps the "awaited type" in `Awaited<T>` if needed.
+// @remarks The "awaited type" of an expression is its "promised type" if the expression is a
```
jakebailey, line 22000: FWIW these sorts of comments will probably just render weirdly in the tooltip and godoc, since these I think are paragraphs.

PR #200, opened 2024-12-27
Total ordering of types

ahejlsberg:
This PR implements a total ordering of types that replaces the previous ordering based on type IDs. The new total ordering depends on type names, declaration order, and/or literal values, but is free of dependencies on type creation order and never requires resolution of constituent types or members. A desirable effect of the new ordering is that in a given program, union type constituents are always ordered (and possibly reduced) the same regardless of when and how the union types are materialized.

I expect there is some performance penalty associated with the new ordering, though it doesn't appear to be significant.

For details on the exact manner in which types are ordered, see the `compareTypes` function [here](https://github.com/microsoft/typescript-go/pull/200/files#diff-cf9f708f1c35e356c380f65b1717857de2defe13aae4da2d003d46b2f9d84c8aR1632).

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -487,7 +487,7 @@ type Checker struct {
 	compilerOptions                         *core.CompilerOptions
 	files                                   []*ast.SourceFile
 	fileIndexMap                            map[*ast.SourceFile]int
-	compareSymbolsFunc                      func(*ast.Symbol, *ast.Symbol) int
+	compareSymbols                          func(*ast.Symbol, *ast.Symbol) int
 	typeCount                               uint32
 	symbolCount                             uint32
 	totalInstantiationCount                 uint32
@@ -735,7 +735,7 @@ func NewChecker(program *Program) *Checker {
 	c.compilerOptions = program.compilerOptions
 	c.files = program.files
 	c.fileIndexMap = createFileIndexMap(c.files)
-	c.compareSymbolsFunc = c.compareSymbols // Closure optimization
+	c.compareSymbols = c.compareSymbolsWorker // Closure optimization
 	c.languageVersion = c.compilerOptions.GetEmitScriptTarget()
 	c.moduleKind = c.compilerOptions.GetEmitModuleKind()
 	c.legacyDecorators = c.compilerOptions.ExperimentalDecorators == core.TSTrue
@@ -908,7 +908,7 @@ func NewChecker(program *Program) *Checker {
 }
 
 func createFileIndexMap(files []*ast.SourceFile) map[*ast.SourceFile]int {
-	result := make(map[*ast.SourceFile]int)
+	result := make(map[*ast.SourceFile]int, len(files))
 	for i, file := range files {
 		result[file] = i
 	}
--- a/internal/compiler/utilities.go
+++ b/internal/compiler/utilities.go
@@ -1580,10 +1580,10 @@ func createSymbolTable(symbols []*ast.Symbol) ast.SymbolTable {
 }
 
 func (c *Checker) sortSymbols(symbols []*ast.Symbol) {
-	slices.SortFunc(symbols, c.compareSymbolsFunc)
+	slices.SortFunc(symbols, c.compareSymbols)
 }
 
-func (c *Checker) compareSymbols(s1, s2 *ast.Symbol) int {
+func (c *Checker) compareSymbolsWorker(s1, s2 *ast.Symbol) int {
 	if s1 == s2 {
 		return 0
 	}


internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		return 1
+	}
+	if n2 == nil {
+		return -1
+	}
+	f1 := ast.GetSourceFileOfNode(n1)
+	f2 := ast.GetSourceFileOfNode(n2)
+	if f1 != f2 {
+		// Compare the full paths (no two files should have the same full path)
+		return strings.Compare(string(f1.Path()), string(f2.Path()))
```
jakebailey: I suspect we'll need to change this to instead check the ordering of files in a program, otherwise this ordering will depend on where TS is installed, where the code was checked out, and maybe put lib.d.ts files out of place.

Passing the program down is probably a no go (given that this function is top level), but maybe we can just stick the file index onto SourceFile as a field? That'd turn it into an integer comparison too, avoiding a full string comparison, which could be significant given how many comparisons are cross-file.

(Then if that index is not present, we could fall back to the Path comparison, maybe)

internal/compiler/types.go
```diff
@@ -375,21 +375,21 @@ const (
...
-	TypeFlagsNumberLiteral   TypeFlags = 1 << 8
-	TypeFlagsBooleanLiteral  TypeFlags = 1 << 9
-	TypeFlagsEnumLiteral     TypeFlags = 1 << 10 // Always combined with StringLiteral, NumberLiteral, or Union
-	TypeFlagsBigIntLiteral   TypeFlags = 1 << 11
-	TypeFlagsESSymbol        TypeFlags = 1 << 12 // Type of symbol primitive introduced in ES6
-	TypeFlagsUniqueESSymbol  TypeFlags = 1 << 13 // unique symbol
-	TypeFlagsVoid            TypeFlags = 1 << 14
-	TypeFlagsUndefined       TypeFlags = 1 << 15
-	TypeFlagsNull            TypeFlags = 1 << 16
+	TypeFlagsUndefined       TypeFlags = 1 << 2
```
jakebailey, line 384: Hopefully this reordering doesn't make interop too hard (but I understand why it's a good idea here)

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
-				// Compare the full paths (no two files should have the same full path)
-				return strings.Compare(string(f1.Path()), string(f2.Path()))
+	if s1 == nil {
+		return 1
+	}
+	if s2 == nil {
+		return -1
+	}
+	if len(s1.Declarations) != 0 && len(s2.Declarations) != 0 {
+		if c := compareNodes(s1.Declarations[0], s2.Declarations[0]); c != 0 {
```
jakebailey: Is there ever a condition where we would need to compare the other declarations?

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		return -1
+	} else if len(s2.Declarations) != 0 {
+		return 1
+	}
+	if c := strings.Compare(s1.Name, s2.Name); c != 0 {
+		return c
+	}
+	// Fall back to symbol IDs. This is a last resort that should happen only when symbols have
+	// no declaration and duplicate names.
+	return int(ast.GetSymbolId(s1)) - int(ast.GetSymbolId(s2))
```
jakebailey, line 1610: This could be:

```suggestion
	return cmp.Compare(ast.GetSymbolId(s1), ast.GetSymbolId(s2))
```

Since these types are ordered.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		return 0
+	}
+	if t1 == nil {
+		return -1
+	}
+	if t2 == nil {
+		return 1
+	}
+	// First sort in order of increasing type flags values.
+	if c := getSortOrderFlags(t1) - getSortOrderFlags(t2); c != 0 {
```
jakebailey, line 1647: This seems like it could be a `cmp.Compare`, but, not really a big deal.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		}
+	}
+	// Fall back to type IDs. This results in type creation order for built-in types.
+	return int(t1.id) - int(t2.id)
+}
+
+func getSortOrderFlags(t *Type) int {
+	// We want enum literal and computed values to be ordered by their declarations, so we merge TypeFlagsEnum into
+	// TypeFlagsEnumLiteral and clear TypeFlagsEnum.
+	return int((t.flags&TypeFlagsEnum)>>1 | t.flags&^TypeFlagsEnum)
```
jakebailey, line 1802: This depends on the ordering of the enum; is there a clearer way to do this? I would honestly expect the Go compiler to figure out some branching here just fine and convert it to nicer ops, but maybe not.

If anything, would be good to have some sort of test somehow that maintains the invariant that this fiddling is legal.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+}
+
+func compareTupleTypes(t1, t2 *TupleType) int {
+	if t1 == t2 {
+		return 0
+	}
+	if t1.readonly != t2.readonly {
+		return core.IfElse(t1.readonly, 1, -1)
+	}
+	if len(t1.elementInfos) != len(t2.elementInfos) {
```
jakebailey, line 1847: I feel like this would be clearer as a `c := cmp.Compare(len(t1.elementInfos), len(t2.elementInfos)); c != 0 { return c }` (which I think I'm just saying for most of these manual checks).

internal/compiler/utilities.go
ahejlsberg: Yeah, we can definitely adjust this. The important thing is to find the most stable ordering possible. Unfortunately we can't just stick an index on `SourceFile` because it is entirely possible for a source file to belong to multiple programs at the same time.

internal/compiler/utilities.go
ahejlsberg: That would matter only if two symbols include the same declaration and also have other differing declarations--which should never happen.

internal/compiler/utilities.go
ahejlsberg: Yeah, but I think that's just a more expensive way.

internal/compiler/utilities.go
ahejlsberg: Again, I think just subtracting the two ints is about as efficient as it can get.

internal/compiler/utilities.go
ahejlsberg: We want all enum members to be ordered together (and appear in declaration order), so we need to remove the distinction between enum literals and computed enum values.

jakebailey: I understand the rationale, I was just trying to make sure we weren't likely to forget this and reorder the constants 😅

internal/compiler/utilities.go
jakebailey: Yeah, if we really are reusing ASTs between programs (we aren't yet), then there's not really something we can stick on that would work (two programs could have two files in reversed orders, even a weak map wouldn't help since we wouldn't know which program is being used), but that means that all of this will have to turn into checker/program(/`*objectSorter`?) methods so that list can be accessed...

(It's not even sufficient to just special case things that look like lib.d.ts files or something, since those can be in `node_modules` via lib replacement...)

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,358 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		return -1
+	}
+	f1 := ast.GetSourceFileOfNode(n1)
+	f2 := ast.GetSourceFileOfNode(n2)
+	if f1 != f2 {
+		// Compare the full paths (no two files should have the same full path)
+		return strings.Compare(string(f1.Path()), string(f2.Path()))
+	}
+	// In the same file, compare source positions
+	return n1.Pos() - n2.Pos()
```
jakebailey, line 1630: Just noticed, but two nodes can share the same starting pos and not be equal. I think this needs to check End too.

But, I think those can _also_ be equal, especially in partially broken trees. (and maybe synthetic ones if we switch JS/JSX to check synthetic or something)

Maybe it needs to be range + kind comparison?

internal/compiler/utilities.go
jakebailey: Certainly, but I was mainly going for clarity.

ahejlsberg:
Latest commit changes the comparison logic to order types first by kind, then by name (when applicable), and finally by declaration order. This makes the ordering less sensitive to source code movement and file renaming.

internal/compiler/utilities.go
ahejlsberg: We may want to specially tag lib.d.ts files and order them as a separate group before or after regular files such that install location doesn't affect ordering. But otherwise I think it is fine to ultimately sort by file paths. With my latest commit, named types are always sorted by name first, so source location really only matters for unnamed and identically named types.

internal/compiler/utilities.go
ahejlsberg: Got it. I already put a comment on `TypeFlagsEnum`, hopefully that's sufficient.

internal/compiler/utilities.go
ahejlsberg: The `compareSymbols` and `compareTypes` functions guarantee a total ordering (ultimately falling back to IDs), but `compareNodes` is just a helper and nothing would be gained by comparing full range and kinds. When node positions are identical, the calling logic moves on to comparing the next level of detail.

internal/compiler/utilities.go
jakebailey: "is lib" is certainly a boolean inherent to #194, so that's certainly doable.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,352 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+	if len(s1) != len(s2) {
+		return len(s1) - len(s2)
+	}
+	for i, t1 := range s1 {
+		if c := strings.Compare(t1, s2[i]); c != 0 {
+			return c
+		}
+	}
+	return 0
+}
```
DanielRosenwasser: This function and the above are *almost* the same as [`slices.Compare`](https://pkg.go.dev/slices#Compare), but they differ in that here, a different length short circuits comparing each individual element.

This does, though, mean that something like

```ts
`foobar${T}baz` | `xyz${T}xyz` | `foo${T}bar${T}baz`
```

is considered "sorted" even though a user might prefer to see

```ts
`foobar${T}baz` | `foo${T}bar${T}baz` | `xyz${T}xyz`
```

I think it's worth reconsidering at least for template string types.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,352 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
+		return c
+	}
+	// Order named types by name and, in the case of aliased types, by alias type arguments.
+	if c := compareTypeNames(t1, t2); c != 0 {
+		return c
+	}
+	// We have unnamed types or types with identical names. Now sort by data specific to the type.
+	switch {
+	case t1.flags&(TypeFlagsAny|TypeFlagsUnknown|TypeFlagsString|TypeFlagsNumber|TypeFlagsBoolean|TypeFlagsBigInt|TypeFlagsESSymbol|TypeFlagsVoid|TypeFlagsUndefined|TypeFlagsNull|TypeFlagsNever|TypeFlagsNonPrimitive) != 0:
+		// Only distinguished by type IDs
```
DanielRosenwasser:
```suggestion
		// Only distinguished by type IDs, handled below.
```

ahejlsberg:
With latest commit we now sort template literal types by their text segments in sequence.

internal/compiler/utilities.go
```diff
@@ -1585,35 +1586,349 @@ func compareSymbols(s1, s2 *ast.Symbol) int {
...
-	panic("Symbols must have unique names to be sorted")
+	for i, t1 := range s1 {
+		if c := compareTypes(t1, s2[i]); c != 0 {
+			return c
+		}
+	}
+	return 0
+}
+
+func compareTexts(s1, s2 []string) int {
```
DanielRosenwasser: Why not just use `slices.Compare`?

DanielRosenwasser: Also, maybe worth noting in the function why you don't short-circuit on length.

ahejlsberg:
With the latest commits, when ordering depends on source positions we compare the indices of the source files in the program's file list instead of comparing file paths. We obtain source file indices using a map (from source file object to index) constructed by the checker. In order to have access to the checker, type objects now carry a reference to the checker that created them. We'll need this reference in the language service anyway, and it allows us to panic on attempts to relate types that didn't originate in the same checker. I verified that the cost of adding this field is less than 0.5% (when compiling VSCode).

internal/compiler/checker.go
```diff
@@ -903,6 +907,14 @@ func NewChecker(program *Program) *Checker {
 	return c
 }
 
+func createFileIndexMap(files []*ast.SourceFile) map[*ast.SourceFile]int {
+	result := make(map[*ast.SourceFile]int)
```
```suggestion
	result := make(map[*ast.SourceFile]int, len(files))
```

internal/compiler/utilities.go
```diff
@@ -1579,11 +1579,11 @@ func createSymbolTable(symbols []*ast.Symbol) ast.SymbolTable {
...
 }
 
-func sortSymbols(symbols []*ast.Symbol) {
-	slices.SortFunc(symbols, compareSymbols)
+func (c *Checker) sortSymbols(symbols []*ast.Symbol) {
+	slices.SortFunc(symbols, c.compareSymbolsFunc)
 }
 
-func compareSymbols(s1, s2 *ast.Symbol) int {
+func (c *Checker) compareSymbols(s1, s2 *ast.Symbol) int {
```
jakebailey: Should this be `compareSymbolsWorker`, and `compareSymbolsFunc` be renamed `compareSymbols`, just to promote the latter's usage?

ahejlsberg: I wanted to only use the closure version when necessary (it's an extra indirection), but probably not a worthwhile optimization.

PR #201, opened 2024-12-30
Set up CodeQL

jakebailey:
Meant to do this when copying stuff over from the main repo.

Uses the config from https://github.com/microsoft/TypeScript/pull/60828 but ignoring the TS submodule and all `testdata` files.

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://golangci-lint.run/jsonschema/custom-gcl.jsonschema.json
 
-version: v1.62.2 # NOTE: this must match golangciLintVersion in Herebyfile.mjs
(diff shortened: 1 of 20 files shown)


jakebailey:
Quite a few things caught by CodeQL above, all to do with conversions in the evaluator where I actually think they're not "wrong" given JS's own truncation, but may mean we need to include some sort of explicit truncation or rounding to be clear about what we're doing for these JS evaluations.

jakebailey:
All of the flagged stuff is valid, but mainly because it's showing that our handling of JS integers is wrong; I'll come up with a different PR with some tests which can better handle this.

jakebailey:
With the new jsnum package, there's only one conversion complaint (the fast path), which I have silenced as it's intentional and its result is checked.

PR #202, opened 2025-01-01
Concurrent type checking

ahejlsberg:
This PR introduces concurrent type checking. By default, we now create four type checkers and divide programs into four equal parts that are type checked concurrently. Diagnostics are then collected from each of the checkers and merged into a single set of diagnostics. The division into equal parts is determined from the index of each particular source file in the program modulo the number of checkers. This means the partitioning is deterministic for the same set of source files, which in turn means the produced set of diagnostics remains stable across compilations. The set of diagnostics may however differ from a single-threaded run of the compiler since the compiler may choose different locations for elaborations.

Concurrent type checking isn't as perfectly parallelizable as parsing and binding because checking often resolves types across source file boundaries, potentially duplicating work between the four checkers. Indeed, concurrent type checking always consumes more memory (because of the duplicated types) and in the worst case is no faster than single-threaded checking. However, in practice, concurrent type checking appears to have a significant positive effect. For example, with single-threaded parsing, binding, and checking, the compile time for VSCode is around 12s, and with concurrent parsing and binding, but single threaded checking (as was the case before this PR), the compile time is 8.5s. With full concurrent parsing, binding, and type checking, the compile time drops to 4s at a cost of consuming 20% more memory. (Measured on an 8-core Intel Core i9.)

Note that four type checkers were experimentally chosen as a good balance between performance and memory consumption. With eight type checkers, VSCode compiles about 10% faster, but memory consumption further increases by 20%.

Also note that #200 is a prerequisite for this PR. Without #200, types are ordered by type IDs, which become non-deterministic with concurrent checking.

### DIFF
--- a/internal/compiler/checker_test.go
+++ b/internal/compiler/checker_test.go
@@ -33,9 +33,8 @@ foo.bar;`
 	}
 	p := NewProgram(opts)
 	p.bindSourceFiles()
-	p.checkSourceFiles()
+	c := p.GetTypeChecker()
 	file := p.filesByPath["/foo.ts"]
-	c := p.getTypeChecker(file)
 	interfaceId := file.Statements.Nodes[0].Name()
 	varId := file.Statements.Nodes[1].AsVariableStatement().DeclarationList.AsVariableDeclarationList().Declarations.Nodes[0].Name()
 	propAccess := file.Statements.Nodes[2].AsExpressionStatement().Expression
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -219,7 +219,20 @@ func (p *Program) createCheckers() {
 	}
 }
 
-func (p *Program) getTypeChecker(file *ast.SourceFile) *Checker {
+// Return the type checker associated with the program.
+func (p *Program) GetTypeChecker() *Checker {
+	p.createCheckers()
+	// Just use the first (and possibly only) checker for checker requests. Such requests are likely
+	// to obtain types through multiple API calls and we want to ensure that those types are created
+	// by the same checker so they can interoperate.
+	return p.checkers[0]
+}
+
+// Return a checker for the given file. We may have multiple checkers in concurrent scenarios and this
+// method returns the checker that was tasked with checking the file. Note that it isn't possible to mix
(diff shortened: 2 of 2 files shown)


internal/compiler/utilities.go
```diff
@@ -42,12 +43,21 @@ func (s *LinkStore[K, V]) has(key K) bool {
 
 // Atomic ids
 
-var nextMergeId atomic.Uint32
+var (
+	symbolMergeMutex sync.Mutex
+	nextMergeId      uint32
+)
```
jakebailey, line 48: With concurrent checkers, are we going to have a problem here? Exhausting int32 globally with parallel checking doesn't seem infeasible at all anymore...

ahejlsberg: It would take 4+ billion merged symbols, and then we'd really only have a problem if some of the symbols from 4+ billion in the past are still around to cause conflicts. I'm not too worried.

jakebailey: Given the old compiler had a max of `2^52`, that's a lot of headroom we had before that could have been insulating us from this problem; I suspect that someone typing in a large project could actually get up to this limit, honestly...

DanielRosenwasser:
Will there be a way to run everything else in parallel, but checking in a single-threaded manner? Or to change the split-up count?

internal/compiler/utilities.go
```diff
@@ -42,12 +42,21 @@ func (s *LinkStore[K, V]) has(key K) bool {
...
 
 func getMergeId(symbol *ast.Symbol) ast.MergeId {
+	if symbol.MergeId != 0 {
+		return symbol.MergeId
+	}
+	symbolMergeMutex.Lock()
 	if symbol.MergeId == 0 {
```
jakebailey, line 55: I was initially confused about this, until I realized that the race is on `symbol.MergeId` itself (and not necessarily the global), since symbols can be shared between checkers.

jakebailey:
> Will there be a way to run everything else in parallel, but checking in a single-threaded manner? Or to change the split-up count?

Not in this PR there isn't; it's possible we could add some sort of flag to do that, since "concurrent checking" is the most dangerous thing comparatively.

PR #203, opened 2025-01-02
Update golangci-lint to v1.63.1, improve management of custom binary

jakebailey:
This PR:

- Consolidates the custom linter version into `.custom-gcl.yml`. I just used a regex to "parse" this.
- Adds a new `custom-gcl.hash` file, which contains the sha256 hash of all of the files used to build the custom binary, including `.custom-gcl.yml` (for config/version), `_tools/go.mod`, and the `customlint` package. If `custom-gcl` doesn't exist or the hash file contents differs, the binary will be rebuilt. This allows for faster iteration of modfiying the linter, and also means pulling in main will make `hereby lint` do the right thing.
- Updates to `golangci-lint` v1.63.1, which introduces support for `SuggestedFixes`.
- Fixes new lints added by that version.


Herebyfile.mjs
```diff
@@ -124,19 +125,60 @@ export const testAll = task({
 });
 
 const customLinterPath = "./_tools/custom-gcl";
-const golangciLintVersion = "v1.62.2"; // NOTE: this must match the version in .custom-gcl.yml
+const customLinterHashPath = customLinterPath + ".hash";
+
+const golangciLintVersion = memoize(() => {
```
DanielRosenwasser, line 183: Might rename this with `get`

jakebailey: The other memoized funcs in the file don't have the `get` prefix, I'd probably prefer not to have one looking at other uses in context I think, but these are a little extra work-y than te others which are mainly just a wrapper around `which`...

PR #204, opened 2025-01-02
move `ParsedCommandLine` to `core`

iisaduan:


jakebailey:
Needs a format but seems okay

internal/tsoptions/parsedcommandline.go
```diff
@@ -0,0 +1,14 @@
+package tsoptions
+
+import (
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+type ParsedCommandLine struct {
+	Options       *core.ParsedOptions
+	ConfigFile    *ast.SourceFile
```
navya9singh: Why do you need `ConfigFile`?

iisaduan: I moved config file out of `CompilerOptions`, because otherwise, we'd have to move `CompilerOptions` out of core

PR #206, opened 2025-01-05
More type checking

ahejlsberg:
In this PR:

* Resolution of iteration types.
* A few issues fixed.

### DIFF
--- a/internal/compiler/checker.go
+++ b/internal/compiler/checker.go
@@ -708,6 +708,16 @@ type Checker struct {
 	getGlobalPromiseLikeType                func() *Type
 	getGlobalPromiseConstructorSymbol       func() *ast.Symbol
 	getGlobalOmitSymbol                     func() *ast.Symbol
+	getGlobalIteratorType                   func() *Type
+	getGlobalIterableType                   func() *Type
+	getGlobalIterableIteratorType           func() *Type
+	getGlobalIteratorObjectType             func() *Type
+	getGlobalGeneratorType                  func() *Type
+	getGlobalAsyncIteratorType              func() *Type
+	getGlobalAsyncIterableType              func() *Type
+	getGlobalAsyncIterableIteratorType      func() *Type
+	getGlobalAsyncIteratorObjectType        func() *Type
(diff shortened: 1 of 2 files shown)


internal/compiler/checker.go
```diff
@@ -2530,8 +2613,180 @@ func (c *Checker) getIterationTypeOfIterable(use IterationUse, typeKind Iteratio
...
+			return c.getIterationTypesOfIteratorWorker(iteratorType, r, errorNode)
+		}
+	}
+	return IterationTypes{}
+}
+
+// Gets the *yield*, *return*, and *next* types from an `Iterator`-like or `AsyncIterator`-like type.
+//
+// If we successfully found the *yield*, *return*, and *next* types, an `IterationTypes`
+// record is returned. Otherwise, `undefined` is returned.
```
DanielRosenwasser:
```suggestion
// record is returned. Otherwise, a default `IterationTypes` record is returned
```

Is that intentional?

internal/compiler/checker.go
```diff
@@ -2530,8 +2613,180 @@ func (c *Checker) getIterationTypeOfIterable(use IterationUse, typeKind Iteratio
...
+		returnType: core.OrElse(c.getAwaitedTypeEx(iterationTypes.returnType, errorNode, nil), c.anyType),
+		nextType:   iterationTypes.nextType,
+	}
+}
+
+// Gets the *yield*, *return*, and *next* types of an `Iterable`-like or `AsyncIterable`-like
+// type from its members.
+//
+// If we successfully found the *yield*, *return*, and *next* types, an `IterationTypes`
+// record is returned. Otherwise, `noIterationTypes` is returned.
```
DanielRosenwasser:
```suggestion
// record is returned. Otherwise, a default `IterationTypes` record is returned
```

internal/compiler/checker.go
ahejlsberg: Yeah, I didn't mess with the comments, but I probably should. As you've probably noticed, we now return a struct with the three types instead of an allocated object, so `nil` isn't an available marker type.

PR #209, opened 2025-01-06
Add jsnum.Number type

jakebailey:
This named type prevents misuse, i.e. direct usage with the `math` package, which has different behaviors than JS.

Pure operators like `+`, `-`, etc, all work correctly, along with direct usage of values, and the named type does not prevent those usages.

I have moved the implementation of number/string conversion into `jsnum`, but it still has not been verified or tested. That'll come later.

After this PR, the only use of the `math` package outside `jsnum` are for accessing `MaxInt` and similar.

### DIFF
--- a/internal/jsnum/jsnum.go
+++ b/internal/jsnum/jsnum.go
@@ -19,6 +19,7 @@ const (
 // not the "math" package and conversions.
 type Number float64
 
+// https://tc39.es/ecma262/2024/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-tostring
 func (n Number) String() string {
 	// !!! verify that this is actually the same as JS.
 	return strconv.FormatFloat(float64(n), 'g', -1, 64)


internal/jsnum/jsnum.go
```diff
@@ -1,20 +1,76 @@
...
+// Number represents a JS-like number.
+//
+// All operations that can be performed directly on this type
+// (e.g., conversion, arithmetic, etc.) behave as they would in JavaScript,
+// but any other operation should use this type's methods,
+// not the "math" package and conversions.
+type Number float64
+
+func (n Number) String() string {
+	// !!! verify that this is actually the same as JS.
```
ahejlsberg, line 24: Regarding being the same as JS, I know for a fact that it isn't. As I recall, Go switches to engineering notation around 1e6, whereas JS switches above 1e20.

internal/jsnum/jsnum.go
```diff
@@ -1,20 +1,76 @@
...
+func Inf(sign int) Number {
+	return Number(math.Inf(sign))
+}
+
+func (n Number) IsInf() bool {
+	return math.IsInf(float64(n), 0)
+}
+
+// https://tc39.es/ecma262/2024/multipage/abstract-operations.html#sec-stringtonumber
+func ToNumber(s string) Number {
```
ahejlsberg: Wouldn't it make more sense to call this `FromString`? Reading `jsnum.ToNumber(x)` I find myself confused about what's happening.

internal/jsnum/jsnum.go
```diff
@@ -1,20 +1,76 @@
...
+	MinSafeInteger Number = -MaxSafeInteger
+)
+
+// Number represents a JS-like number.
+//
+// All operations that can be performed directly on this type
+// (e.g., conversion, arithmetic, etc.) behave as they would in JavaScript,
+// but any other operation should use this type's methods,
+// not the "math" package and conversions.
+type Number float64
```
ahejlsberg, line 20: Did you consider making `Number` a struct? It's less flexible, but would ensure that all conversions go through a defined interface.

internal/jsnum/jsnum.go
jakebailey: Yeah, I'm working on that, but the ECMAScript spec for this is very, very obtuse.

internal/jsnum/jsnum.go
jakebailey: I'm trying to call it the same as what the ECMAScript spec calls it, which is `StringToNumber`, but I could flip it.

internal/jsnum/jsnum.go
jakebailey: I can do that, it was just a little annoying because it meant conversions needed to turn into construction struct construction and then you couldn't use any operators, but, I can still try it.

ahejlsberg:
One other thing... For all of the binary operators, I kind of prefer `jsnum.Operator(x, y)` instead of `x.Operator(y)`. Just feels odd that one of the operands is elevated to receiver status when they're both equally important.

jakebailey:
I honestly don't prefer that; a top level function is less discoverable than a method when all you have is the `Number`, and if you ever need to do more than one, you'd have to write `jsnum.BitwiseOR(a, jsnum.BitwiseOR(b, jsnum.BitwiseOR(c, d)))` or similar, compared to `a.BitwiseOR(b).BitwiseOR(c).BitwiseOR(d)`. But I can do it however.

internal/jsnum/jsnum.go
jakebailey: Renamed to `FromString`.

internal/jsnum/jsnum.go
jakebailey: I've tried this out, and the only thing it seems to do is make simple/safe operations (like arithemetic, comparison) more annoying; I didn't catch any bugs or anything. I think I'd prefer to keep this as a plain named type for now, and I can put up that PR separately to show what it'd be.

jakebailey: Here's what the change would look like: https://github.com/microsoft/typescript-go/pull/212

I personally don't think this is worth it.

PR #210, opened 2025-01-07
Make `packagejson.Exports` and `packagejson.JSONValue` use non-pointer-valued maps and slices so `null` unmarshalls properly

andrewbranch:
Noticed a problem where top-level JSON `null` unmarshals into our custom structs correctly, but a `null` inside an object becomes a `nil` inside the `OrderedMap[string, *Exports]`. We need that composite type to be `OrderedMap[string, Exports]` instead to have a chance to change how we represent `null`.

### DIFF
--- a/internal/compiler/packagejson/exports.go
+++ b/internal/compiler/packagejson/exports.go
@@ -35,11 +35,11 @@ func (e Exports) AsObject() *collections.OrderedMap[string, Exports] {
 	return e.Value.(*collections.OrderedMap[string, Exports])
 }
 
-func (e Exports) AsArray() []*Exports {
+func (e Exports) AsArray() []Exports {
 	if e.Type != JSONValueTypeArray {
 		panic("expected array")
 	}
-	return e.Value.([]*Exports)
+	return e.Value.([]Exports)
 }
 
 func (e Exports) IsSubpaths() bool {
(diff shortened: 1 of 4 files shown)


internal/compiler/packagejson/exports.go
```diff
@@ -28,26 +28,26 @@ func (e *Exports) UnmarshalJSONV2(dec *jsontext.Decoder, opts json2.Options) err
...
+func (e Exports) AsObject() *collections.OrderedMap[string, Exports] {
 	if e.Type != JSONValueTypeObject {
 		panic("expected object")
 	}
-	return e.Value.(*collections.OrderedMap[string, *Exports])
+	return e.Value.(*collections.OrderedMap[string, Exports])
 }
 
-func (e *Exports) AsArray() []*Exports {
+func (e Exports) AsArray() []*Exports {
```
andrewbranch: Should arrays be changed to unmarshal into `[]Exports` instead to match? Merging this fix into my in-progress package.json "exports" work exposes a little awkwardness around this.

andrewbranch: Actually, I don’t think the tests cover this, but I bet the same issue occurs for `null` within arrays (pointless/rare in `"exports"`)

PR #215, opened 2025-01-09
Reimplement number/string conversions per ECMAScript spec

jakebailey:
Lucky for us, Go's `encoding/json` package actually implements float64 marshalling the same way as the ECMAScript spec. In `floatEncoder`:

```go
	// Convert as if by ES6 number to string conversion.
	// This matches most other JSON generators.
	// See golang.org/issue/6384 and golang.org/issue/14135.
	// Like fmt %g, but the exponent cutoffs are different
	// and exponents themselves are not padded to two digits.
```

This is exactly the difference we saw with `%g`.

This PR reimplements our own suspicious conversions using `encoding/json`.

The tests are extensive, including tests pulled (with attribution) from `ryu`, the floating point printing implementation used in Go. V8 uses `Grisu3`, which Go also used to use. Regardless, the behavior is identical, and a load of testing shows it.

There's also a test which sends every test over to `node` and then verifies that things work identically, along with a fuzz test which attempts to find any place that differs with node. I have found no such difference in a bunch of testing. I even tried `bun`; it all checks out.

If this doesn't work in the future, then we'll definitely need to go port ryu/grisu3/dragonbox or something, which is a bad time. Go's json library basically already does this, applying the correct fixups, so I think this is solid.

---

For _parsing_ numbers, the ECMAScript spec is written in terms of a grammar, which is more or less the same as a JS parser, which is a problem.

To tackle this, I've instead written something which effectively "cleans up" the user's input to then provide to Go's existing number parsing functions; extensive fuzz testing found bugs, but the current implementation has been fuzzed for a pretty long time with good success.


jakebailey:
Eek, I forgot to run tests for parsing; it doesn't match. May require more thought.

internal/jsnum/ryu_test.go
```diff
@@ -0,0 +1,164 @@
+package jsnum
+
+// Copyright 2018 Ulf Adams
```
jakebailey, line 3: This file is modified from ryu's tests; this file will not actually ship to anyone, so I don't think we need to worry about licensing besides conforming to specifically the fact that this file needs to preserve the license information and note that it has been modified.

internal/jsnum/string_test.go
```diff
@@ -0,0 +1,347 @@
...
+		}));
+
+		process.stdout.write(JSON.stringify(output));
+	`
+
+	scriptPath := filepath.Join(tmpdir, "script.cjs")
+	err = os.WriteFile(scriptPath, []byte(script), 0o644)
+	assert.NilError(t, err)
+
+	execCmd := exec.Command(exe, scriptPath, jsonInputPath)
```
jakebailey, line 358: This test uses `node`, but I have actually been testing with a local change that uses `v8go` to run things. It's not a good idea to check that in, though, since it introduces a dependency we don't want and especially it means testing our codebase requires the ability to compile V8. It also breaks linting since it's so much code.

jakebailey:
Okay, this should be working now. The code is _not_ going to be fast, but I believe it to be correct enough for now to unblock places where we need conversions to work.

jakebailey:
(Long term I'll probably port Chakra's code or something but that's a lot more work and code)

jakebailey:
This branch really makes CodeQL take forever. main takes 3 minutes, but this branch consistently takes _an hour_.

Seems to be due to:

```
[31/31 eval 52m51s] Evaluation done; writing results to codeql/go-queries/Security/CWE-089/StringBreak.bqrs.
```

Which is some sort of quote sanitization check. This PR does so much string manipulation that I assume this gets very mad. Maybe there's some way to disable it, or otherwise indicate that we don't care about inputs that contain quotes at all.

jakebailey:
Disabling that one query worked.

internal/jsnum/string.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	// Otherwise, the Go json package handles this correctly.
+	b, _ := json.Marshal(float64(n)) //nolint:errchkjson
+	return string(b)
+}
+
+// https://tc39.es/ecma262/2024/multipage/abstract-operations.html#sec-stringtonumber
+func FromString(s string) Number {
+	// Implemeting StringToNumber exactly as written in the spec involves
```
sandersn:
```suggestion
	// Implementing StringToNumber exactly as written in the spec involves
```

internal/jsnum/string.go
```diff
@@ -0,0 +1,347 @@
...
+func stringToFloat64(s string) float64 {
+	if f, err := strconv.ParseFloat(s, 64); err == nil {
+		return f
+	} else {
+		if errors.Is(err, strconv.ErrRange) {
+			return f
+		}
+	}
+
+	i, err := strconv.ParseInt(s, 0, 64)
```
sandersn: when will ParseInt succeed but ParseFloat fail?

jakebailey: Good catch, seemingly never 
image

jakebailey:
> Seems plausible as a medium-term solution.

Yes, long term I plan to port Chakra's code, or reuse Go's Ryu implementation (which we already depend on indirectly).

jakebailey:
There are a couple of uncovered lines; they're lines that make sense but aren't covered due to other prior early bailouts. I'm not really worried about that, especially for a "temporary correct but inefficient, to be replaced" implementation.

PR #216, opened 2025-01-09
Process triple-slash-types directives

ahejlsberg:
Quick PR to actually process `/// <reference types="..."/>` directives during program construction. This gives us a workaround until we have config file parsing and processing of the `types` option working.

~The code works, but I'm not sure if we need additional checks. The code just resolves the reference in the same manner as an import statement which permits both absolute and relative names.~

EDIT: We now resolve type references using the resolver's `ResolveTypeReferenceDirective` method.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -92,7 +92,7 @@ func main() {
 		fmt.Fprintf(os.Stderr, "Error: The directory %v does not exist.\n", normalizedRootPath)
 		os.Exit(1)
 	}
-	compilerOptions.ConfigFilePath = rootPath // This matters for type reference directive resolution
+	compilerOptions.ConfigFilePath = normalizedRootPath // This matters for type reference directive resolution
 
 	programOptions := ts.ProgramOptions{
 		RootPath:           normalizedRootPath,


internal/compiler/program.go
```diff
@@ -275,6 +275,13 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup, isLib bool
 			filesToParse = append(filesToParse, toParse{p: resolvedPath})
 		}
 
+		for _, ref := range file.TypeReferenceDirectives {
+			resolved := p.resolver.ResolveModuleName(ref.FileName, file.FileName(), core.ModuleKindNodeNext, nil)
```
andrewbranch:
```suggestion
			resolved := p.resolver.ResolveTypeReferenceDirective(ref.FileName, file.FileName(), core.ModuleKindNodeNext, nil)
```

jakebailey: I think we actually have `p.resolver.ResolveTypeReferenceDirective` which should do the right thing.

RyanCavanaugh: This should be `ResolveTypeReferenceDirective` instead, otherwise `typeRoots` won't be properly consulted

RyanCavanaugh:
c-c-c-c-combo breaker

internal/compiler/program.go
DanielRosenwasser: Someone mentioned a function named `ResolveTypeReferenceDirective`, I don't know if that's appropriate here but you might want to look into it.

ahejlsberg: Doh! Sure sounds like a better method to use.

ahejlsberg:
With latest commit we use `ResolveTypeReferenceDirective`. This also required passing an appropriate path in `CompilerOptions.ConfigFilePath`, so that's now initialized to the root path.

jakebailey:
Seems good; we'll fix up that config file path thing for tsconfig parsing later.

cmd/tsgo/main.go
```diff
@@ -92,6 +92,7 @@ func main() {
 		fmt.Fprintf(os.Stderr, "Error: The directory %v does not exist.\n", normalizedRootPath)
 		os.Exit(1)
 	}
+	compilerOptions.ConfigFilePath = rootPath // This matters for type reference directive resolution
```
```suggestion
	compilerOptions.ConfigFilePath = normalizedRootPath // This matters for type reference directive resolution
```

ahejlsberg: Argh!

PR #221, opened 2025-01-10
package.json "imports" and "exports"

andrewbranch:
I think a lot of the remaining test failures are due to bugs that I accidentally fixed here in Corsa, then ported into TypeScript, but aren’t reflected in the submodule/baselines yet.


internal/compiler/packagejson/packagejson.go
```diff
@@ -16,7 +16,7 @@ type PathFields struct {
 	Types         Expected[string] `json:"types"`
 	Typings       Expected[string] `json:"typings"`
 	TypesVersions JSONValue        `json:"typesVersions"`
-	Imports       JSONValue        `json:"imports"`
+	Imports       Exports          `json:"imports"`
```
jakebailey: Seems like this type could be renamed to `ImportsOrExports` or something, now that it's serving two purposes?

PR #222, opened 2025-01-11
More type checking

ahejlsberg:
In this PR:

* Proper generator return types.
* Resolution of optional call chain signatures.
* Checking of `super` expressions.
* Checking of `yield` expressions.
* Synthetic defaults for modules.
* Late-bound index signatures.
* Proper resolution of the promised type of a `Promise<T>`.
* More contextual types.
* Template literal type inference.
* Fix for protected property override checks.
* Fix instantiation of mapped types on array types.
* Fix `excludeProperties` function.
* Fix `getIterationTypesOfIterable` function.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -387,7 +387,7 @@ func (p *Program) resolveImportsAndModuleAugmentations(file *ast.SourceFile) []s
 			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
 			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
 
-			shouldAddFile := resolution.IsResolved() && tspath.FileExtensionIsOneOf(resolvedFileName, []string{".ts", ".tsx"})
+			shouldAddFile := resolution.IsResolved() && tspath.FileExtensionIsOneOf(resolvedFileName, []string{".ts", ".tsx", ".mts", ".cts"})
 			// TODO(ercornel): !!!: other checks on whether or not to add the file
 
 			if shouldAddFile {


internal/compiler/checker.go
```diff
@@ -483,250 +493,253 @@ type WideningContext struct {
...
-	getGlobalAsyncIterableIteratorType      func() *Type
-	getGlobalAsyncIteratorObjectType        func() *Type
-	getGlobalAsyncGeneratorType             func() *Type
-	syncIterationTypesResolver              *IterationTypesResolver
-	asyncIterationTypesResolver             *IterationTypesResolver
-	isPrimitiveOrObjectOrEmptyType          func(*Type) bool
-	containsMissingType                     func(*Type) bool
-	couldContainTypeVariables               func(*Type) bool
-	isStringIndexSignatureOnlyType          func(*Type) bool
+	program                                   *Program
```
jakebailey, line 496: We should really consider breaking this up into groups, since we keep modifying it in ways that reformat it. It's frustrating that on GitHub this means no ability to see what changed. (Gerrit, vscode, do better.)

internal/core/compileroptions.go
```diff
@@ -25,7 +25,6 @@ type CompilerOptions struct {
 	IsolatedModules                    Tristate             `json:"isolatedModules"`
 	Jsx                                JsxEmit              `json:"jsx"`
 	Lib                                []string             `json:"lib"`
-	LegacyDecorators                   Tristate             `json:"legacyDecorators"`
```
jakebailey, line 28: Is this being removed because we're not going to attempt to port the feature?

internal/compiler/program.go
```diff
@@ -387,7 +387,7 @@ func (p *Program) resolveImportsAndModuleAugmentations(file *ast.SourceFile) []s
 			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
 			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
 
-			shouldAddFile := resolution.IsResolved()
+			shouldAddFile := resolution.IsResolved() && tspath.FileExtensionIsOneOf(resolvedFileName, []string{".ts", ".tsx"})
```
jakebailey: I don't think this works; `FileExtensionIsOneOf` requires its first argument to be an extension itself.

I would do `tspath.TryExtractTSExtension(resolvedFileName) != ""`.

internal/compiler/checker.go
```diff
@@ -3723,7 +3771,7 @@ func (c *Checker) getConstituentProperty(objectType *Type, propertyName string)
 }
 
 func (c *Checker) checkImportCallExpression(node *ast.Node) *Type {
-	// !!!
+	// !!!!
```
jakebailey, line 3997: Does four `!`s mean something special? 😄

internal/compiler/program.go
jakebailey: Oh, I'm wrong, `FileExtensionIsOneOf` does split the path. But probably best to still make sure cts/mts works.

jakebailey: (I am surprised we don't have a helper which is `FileExtensionIsOneOf(path, supportedTSImplementationExtensions)`...)

internal/core/compileroptions.go
ahejlsberg: This was added by mistake when grammar checks were ported. It should've never been a compiler option (we have no such option), rather the code should've just used the `legacyDecorators` flag on the checker object.

jakebailey: Oh, duh, legacy is not "experimental".

internal/compiler/program.go
ahejlsberg: Sure, this is just a workaround for logic that eventually needs proper alignment with the old codebase (there are already `!!!` comments in the surrounding code).

internal/compiler/checker.go
ahejlsberg: Hehe. I was just using it as a flag to prioritize porting of those functions.

internal/compiler/checker.go
ahejlsberg: I think the longest field name now is long enough that this won't happen again. Grouping would just be arbitrary and likely would drift over time (at least if our current codebase is an indicator).

internal/compiler/program.go
jakebailey: Okay, I guess hardcoding them is fine for now.

PR #223, opened 2025-01-12
Fix program construction

ahejlsberg:
Fixes program construction to ensure files are only parsed and added to the program once, no matter how they were referenced. Before this PR, processing of root files didn't protect against duplicates, so if a source file included a `/// <reference lib="dom"/>`, the DOM lib files would be parsed twice, once because of that reference and again from the processing of lib root files.


jakebailey:
(Would be nice to add a test for this case but my understanding is that this is purely a performance problem and would never have been observed as a correctness problem, so it's a bit hard to test outside asserting the number of parse calls or something.)

PR #225, opened 2025-01-13
Add JSDoc parsing

sandersn:
This PR adds JSDoc parsing. It builds on my previous PRs so it doesn't touch a lot of files for such a large feature. Specifically, the main AST still doesn't have a jsdoc property. Instead, the parser has a `jsdocCache` map that uses node as keys for jsdoc.

Like Strada,
- It's disabled for .ts files unless the file contains `@see` or `@link`.
- It moves diagnostics added during jsdoc parsing to a separate array, `jsdocDiagnostics`, so that JS files without checkJS can ignore them.
- There are dual "nextToken/nextTokenJSDoc" pairs of functions. The JSDoc versions are used in JSDoc proper, but the parser switches back to the normal versions with parsing types in JSDoc.

Some notes:

- Like Strada, it's disabled for .ts files unless the file contains a `@see` or `@link`.
- Because I now have actual tests of JSDoc, there are quite a few bug fixes in ast.go &mdash; incorrect constructor functions or oorder visitor code mostly.
- I removed the features I proposed removing in https://github.com/microsoft/typescript-go/pull/132.
	- These are mostly deletions of entire functions except for `parseJSDocTypeNameWithNamespace`. In Corsa, it always returns a name, whereas in Strada it could return undefined, since `typedef/enum/callback`'s names were optional.
	- Including an embarrassing AI artifact: KindJSDocImmediateTag. I think Copilot added this because of the `@immediate` proposal in 2024.
- I didn't simplify the indentation algorithm, but I would like to later. Why not now?
	- It would be a complex rewrite compared to removing entire features.
	- I don't think test coverage of jsdoc indentation will be good enough until we have fourslash tests.
- JSDoc parsing operates on a slice of the source text so that the parser will reliably emit EOF when it hits the end of the comment.
- getJSDocCommentRanges returns overlapping trailing/leading comments when there are multiple jsdoc beforehand, which is common because of standalone tags like `@typedef`. I copied the Strada behaviour to Corsa even though it's wrong.
- 50% of the lines changed in scanner.go is me fixing up my old style `scanner *Scanner -> s *Scanner` in methods, sorry.

I'll get some performance numbers and post them shortly. I don't expect this to be cheap because the fundamental algorithm hasn't changed from Strada.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -73,6 +73,7 @@ type Parser struct {
 	possibleAwaitSpans      []int
 	jsdocCommentsSpace      []string
 	jsdocCommentRangesSpace []ast.CommentRange
+	jsdocTagCommentsSpace   []string
 }
 
 func NewParser() *Parser {
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -49,6 +49,8 @@ type NodeFactory struct {
 	identifierPool core.Pool[Identifier]
 	tokenPool      core.Pool[Token]
 	nodeListPool   core.Pool[NodeList]
+	jsdocPool      core.Pool[JSDoc]
+	jsdocTextPool  core.Pool[JSDocText]
 }
 
 func newNode(kind Kind, data nodeData) *Node {
@@ -6553,7 +6555,7 @@ type JSDoc struct {
 }
 
 func (f *NodeFactory) NewJSDoc(comment *NodeList, tags *NodeList) *Node {
-	data := &JSDoc{}
+	data := f.jsdocPool.New()
 	data.Comment = comment
 	data.Tags = tags
 	return newNode(KindJSDoc, data)
@@ -6591,7 +6593,7 @@ type JSDocText struct {
 }
 
 func (f *NodeFactory) NewJSDocText(text string) *Node {
-	data := &JSDocText{}
+	data := f.jsdocTextPool.New()
 	data.Text = text
 	return newNode(KindJSDocText, data)
 }
--- a/internal/parser/jsdoc.go
+++ b/internal/parser/jsdoc.go
@@ -166,7 +166,7 @@ func (p *Parser) parseJSDocCommentWorker(start int, indent int) *ast.Node {
 	tagsPos := -1
 	tagsEnd := -1
 	state := jsdocStateSawAsterisk
-	var commentParts []*ast.Node
+	commentParts := p.nodeSlicePool.NewSlice(1)[:0]
 	comments := p.jsdocCommentsSpace
 	commentsPos := -1
 	linkEnd := start
@@ -275,6 +275,7 @@ loop:
 			p.nextTokenJSDoc()
 		}
 	}
+	p.jsdocCommentsSpace = comments[:0] // Reuse this slice for further parses
 	if commentsPos == -1 {
 		commentsPos = p.scanner.TokenFullStart()
 	}
@@ -291,7 +292,6 @@ loop:
 		p.newNodeList(core.NewTextRange(start, commentsPos), commentParts),
 		p.newNodeList(core.NewTextRange(tagsPos, tagsEnd), tags))
 	p.finishNodeWithEnd(jsdocComment, start, len(p.sourceText))
-	p.jsdocCommentsSpace = comments[:0] // Reuse this slice for further parses
 	return jsdocComment
 }
 
@@ -461,7 +461,8 @@ func (p *Parser) parseTrailingTagComments(pos int, end int, margin int, indentTe
 
 func (p *Parser) parseTagComments(indent int, initialMargin *string) *ast.NodeList {
 	commentsPos := p.nodePos()
-	var comments []string
+	comments := p.jsdocTagCommentsSpace
+	p.jsdocTagCommentsSpace = nil // !!! can parseTagComments call itself?
 	var parts []*ast.Node
 	linkEnd := -1
 	state := jsdocStateBeginningOfLine
@@ -526,7 +527,7 @@ loop:
 				p.finishNodeWithEnd(text, commentStart, commentEnd)
 				parts = append(parts, text)
 				parts = append(parts, link)
-				comments = []string{}
+				comments = comments[:0]
 				linkEnd = p.scanner.TokenEnd()
 			} else {
 				pushComment(p.scanner.TokenText())
@@ -566,6 +567,8 @@ loop:
 			tok = p.nextTokenJSDoc()
 		}
 	}
+
+	p.jsdocTagCommentsSpace = comments[:0]
 
 	comments = removeLeadingNewlines(comments)
 	trimmedComments := trimEnd(strings.Join(comments, ""))


sandersn:
Performance:

(numbers below)

In the normal mode, empty.ts is 17.7% slower, checker is the same, dom.generated.d.ts is 1.3% faster and Herebyfile.mjs is 0.4% slower. None of these are actually parsing JSDoc, so I don't know why empty.ts is so much slower. Maybe it's overhead from initialising the jsdocCache.

In always-parse mode, everything is slower. The interesting ones are the large ones: checker.ts and dom.generated.d.ts. checker.ts doesn't have much jsdoc and is only 2% slower. The DOM has JSDoc on nearly every declaration and is 58% slower.

My initial opinion is that this isn't worth optimising now because the language service isn't ready; the common case is that jsdoc isn't parsed at all, and parsing is a small part of take taken to boot.
Edit: Also because when I have more time I want to simplify and change the indent rules, which will allow much chunkier parsing per line of text.

In the normal mode:


```
                                                      │   old.txt   │             newnone.txt             │
Parse/empty.ts-16                                       352.9n ± 0%   415.4n ± 0%  +17.71% (p=0.000 n=10)
Parse/checker.ts-16                                     28.87m ± 2%   28.62m ± 1%        ~ (p=0.436 n=10)
Parse/dom.generated.d.ts-16                             8.919m ± 1%   8.806m ± 0%   -1.27% (p=0.000 n=10)
Parse/Herebyfile.mjs-16                                 396.8µ ± 0%   398.5µ ± 0%   +0.42% (p=0.029 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   149.5µ ± 1%   148.9µ ± 0%        ~ (p=0.218 n=10)
geomean                                                 351.8µ        362.0µ        +2.88%

                                                      │   old.txt    │             newnone.txt             │
Parse/empty.ts-16                                       2.250Ki ± 0%   2.453Ki ± 0%  +9.03% (p=0.000 n=10)
Parse/checker.ts-16                                     24.34Mi ± 0%   24.34Mi ± 0%       ~ (p=0.796 n=10)
Parse/dom.generated.d.ts-16                             8.626Mi ± 0%   8.625Mi ± 0%       ~ (p=0.280 n=10)
Parse/Herebyfile.mjs-16                                 359.2Ki ± 0%   359.3Ki ± 0%  +0.05% (p=0.001 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   157.0Ki ± 0%   157.2Ki ± 0%  +0.14% (p=0.000 n=10)
geomean                                                 488.9Ki        497.6Ki       +1.78%

                                                      │   old.txt   │             newnone.txt             │
                                                      │  allocs/op  │  allocs/op   vs base                │
Parse/empty.ts-16                                        6.000 ± 0%    8.000 ± 0%  +33.33% (p=0.000 n=10)
Parse/checker.ts-16                                     157.9k ± 0%   157.9k ± 0%        ~ (p=0.456 n=10)
Parse/dom.generated.d.ts-16                             72.84k ± 0%   72.84k ± 0%        ~ (p=0.066 n=10)
Parse/Herebyfile.mjs-16                                 2.606k ± 0%   2.608k ± 0%   +0.08% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16    917.0 ± 0%    919.0 ± 0%   +0.22% (p=0.000 n=10)
```

In the old-style, always-parse mode:

```
                                                      │   old.txt   │              newall.txt              │
                                                      │   sec/op    │    sec/op     vs base                │
Parse/empty.ts-16                                       352.9n ± 0%    417.8n ± 0%  +18.39% (p=0.000 n=10)
Parse/checker.ts-16                                     28.87m ± 2%    29.47m ± 1%   +2.07% (p=0.000 n=10)
Parse/dom.generated.d.ts-16                             8.919m ± 1%   14.118m ± 1%  +58.30% (p=0.000 n=10)
Parse/Herebyfile.mjs-16                                 396.8µ ± 0%    453.0µ ± 0%  +14.17% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   149.5µ ± 1%    263.2µ ± 1%  +76.07% (p=0.000 n=10)
geomean                                                 351.8µ         460.6µ       +30.91%

                                                      │   old.txt    │              newall.txt               │
                                                      │     B/op     │     B/op       vs base                │
Parse/empty.ts-16                                       2.250Ki ± 0%    2.453Ki ± 0%   +9.03% (p=0.000 n=10)
Parse/checker.ts-16                                     24.34Mi ± 0%    24.68Mi ± 0%   +1.37% (p=0.000 n=10)
Parse/dom.generated.d.ts-16                             8.626Mi ± 0%   11.598Mi ± 0%  +34.46% (p=0.000 n=10)
Parse/Herebyfile.mjs-16                                 359.2Ki ± 0%    396.6Ki ± 0%  +10.44% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   157.0Ki ± 0%    223.5Ki ± 0%  +42.36% (p=0.000 n=10)
geomean                                                 488.9Ki         579.3Ki       +18.50%

                                                      │   old.txt   │              newall.txt               │
                                                      │  allocs/op  │  allocs/op    vs base                 │
Parse/empty.ts-16                                        6.000 ± 0%     8.000 ± 0%   +33.33% (p=0.000 n=10)
Parse/checker.ts-16                                     157.9k ± 0%    161.1k ± 0%    +2.06% (p=0.000 n=10)
Parse/dom.generated.d.ts-16                             72.84k ± 0%   118.11k ± 0%   +62.14% (p=0.000 n=10)
Parse/Herebyfile.mjs-16                                 2.606k ± 0%    3.040k ± 0%   +16.65% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16    917.0 ± 0%    2213.0 ± 0%  +141.33% (p=0.000 n=10)
geomean                                                 2.776k         4.000k        +44.09%
```

I used the command `go test -run=- -bench=. -benchmem -count=10 ./internal/parser | tee old.txt` to generate the performance numbers.

sandersn:
Indeed making jsdocCache initialisation lazy make empty.ts only slow down by 2.4% and use 4.9% more memory.

```
                                                      │   old.txt   │          newnone-lazy.txt          │
Parse/empty.ts-16                                       352.9n ± 0%   361.5n ± 0%  +2.44% (p=0.000 n=10)
Parse/checker.ts-16                                     28.87m ± 2%   28.77m ± 1%       ~ (p=0.853 n=10)
Parse/dom.generated.d.ts-16                             8.919m ± 1%   8.726m ± 1%  -2.16% (p=0.000 n=10)
Parse/Herebyfile.mjs-16                                 396.8µ ± 0%   397.8µ ± 1%       ~ (p=0.075 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   149.5µ ± 1%   149.0µ ± 1%       ~ (p=0.436 n=10)
geomean                                                 351.8µ        351.7µ       -0.05%

                                                      │   old.txt    │          newnone-lazy.txt           │
Parse/empty.ts-16                                       2.250Ki ± 0%   2.359Ki ± 0%  +4.86% (p=0.000 n=10)
Parse/checker.ts-16                                     24.34Mi ± 0%   24.34Mi ± 0%       ~ (p=0.529 n=10)
Parse/dom.generated.d.ts-16                             8.626Mi ± 0%   8.624Mi ± 0%       ~ (p=0.063 n=10)
Parse/Herebyfile.mjs-16                                 359.2Ki ± 0%   359.1Ki ± 0%       ~ (p=0.289 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16   157.0Ki ± 0%   157.0Ki ± 0%       ~ (p=0.060 n=10)
geomean                                                 488.9Ki        493.6Ki       +0.95%

                                                      │   old.txt   │           newnone-lazy.txt           │
Parse/empty.ts-16                                        6.000 ± 0%    6.000 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-16                                     157.9k ± 0%   157.9k ± 0%       ~ (p=0.135 n=10)
Parse/dom.generated.d.ts-16                             72.84k ± 0%   72.84k ± 0%       ~ (p=0.074 n=10)
Parse/Herebyfile.mjs-16                                 2.606k ± 0%   2.606k ± 0%       ~ (p=1.000 n=10) ¹
Parse/jsxComplexSignatureHasApplicabilityError.tsx-16    917.0 ± 0%    917.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                 2.776k        2.776k       -0.00%
¹ all samples are equal
```

internal/parser/utilities.go
```diff
@@ -59,3 +61,32 @@ func tokenIsIdentifierOrKeyword(token ast.Kind) bool {
...
+		return comment.End() > node.End() || text[comment.Pos()+1] != '*' || text[comment.Pos()+2] != '*' || text[comment.Pos()+3] == '/'
+	})
+}
+
+func isKeywordOrPunctuation(token ast.Kind) bool {
+	return ast.IsKeywordKind(token) || ast.IsPunctuationKind(token)
+}
+
+func isJSDocLikeText(text string) bool {
+	return text[1] == '*' && text[2] == '*' && text[3] != '/'
```
jakebailey: Does this need a bounds check?

internal/parser/parser.go
```diff
@@ -2294,9 +2370,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	JSDocStateSavingBackticks
+)
+
+type PropertyLikeParse int32
+
+const (
+	PropertyLikeParseProperty          PropertyLikeParse = 1 << 0
+	PropertyLikeParseParameter         PropertyLikeParse = 1 << 1
+	PropertyLikeParseCallbackParameter PropertyLikeParse = 1 << 2
+)
```
jakebailey: Are these ever used in this package's exported API? If not, they should probably be unexported.

internal/parser/parser.go
```diff
@@ -2294,9 +2370,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	}
+
+	result := p.factory.NewJSDocNameReference(entityName)
+	ast.SetParentInChildren(result)
+	p.finishNode(result, pos)
+	return result
+}
+
+/** Pass length=-1 to parse the text to the end */
+func (p *Parser) parseJSDocComment(parent *ast.Node, start int /*  = 0 */, end int) *ast.Node {
```
```suggestion
func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int) *ast.Node {
```

internal/parser/parser.go
```diff
@@ -258,6 +268,8 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 	result.LanguageVersion = p.languageVersion
 	result.LanguageVariant = p.languageVariant
 	result.ScriptKind = p.scriptKind
+	result.SetJSDocCache(&p.jsdocCache)
+	p.jsdocCache = nil
```
jakebailey: I don't think this works; you're effectively taking an interior pointer to a field in the parser, then setting it to zero, which means the resulting `SourceFile` will itself contain a pointer to a nil map, but then also hold onto the entire `Parser` forever.

I don't think `SetJSDocCache` should accept a pointer to a map, but rather a map itself. (Pointers to a map are very atypical.)

internal/parser/parser.go
sandersn: nope, the capitals are ported straight over from TS.

internal/parser/utilities.go
sandersn: I don't think it can ever be called on a string less than 4 because the scanner won't set hasJSDoc and getJSDocCommentRanges won't return anything for too-short strings -- but it's a good idea to check just in case we start calling this somewhere else.

internal/scanner/scanner.go
```diff
@@ -370,6 +387,26 @@ func (s *Scanner) charAndSize() (rune, int) {
...
+	if s.JSDocParsingMode == JSDocParsingModeParseForTypeInfo {
+		// If we're in TS, but we don't need to produce reliable errors,
+		// we don't need to parse to find @see or @link.
+		return false
+	}
+	for i, c := range s.text {
+		if c == '@' && (strings.HasPrefix(s.text[i:], "@see") || strings.HasPrefix(s.text[i:], "@link")) {
+			return true
+		}
+	}
```
jakebailey: I've been thinking about this; I think we can actually go faster by doing:

```go
text := s.text
for {
	i := strings.IndexByte(text, '@')
	if i < 0 {
		break
	}
	text = text[i+1:]
	if strings.HasPrefix(text, "see") || strings.HasPrefix(text, "link") {
		return true
	}
}
return false
```

Since `IndexByte` is something the stdlib has optimized.

I've been meaning to write this into a helper somewhere, given we also have to do checks like this for `(import|require)`.

That or, we should really be making https://pkg.go.dev/index/suffixarray for all source texts to amortize the cost of looking for substrings.

internal/parser/parser.go
```diff
@@ -145,6 +149,7 @@ func (p *Parser) initializeState(fileName string, sourceText string, languageVer
 	p.languageVersion = languageVersion
 	p.scriptKind = ensureScriptKind(fileName, scriptKind)
 	p.languageVariant = getLanguageVariant(p.scriptKind)
+	p.scanner.SetJSDocParsingMode(scanner.JSDocParsingModeParseNone)
```
jakebailey: I assume this is temporary until we can plumb something in?

internal/parser/parser.go
```diff
@@ -267,6 +279,8 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 			result.LanguageVersion = p.languageVersion
 			result.LanguageVariant = p.languageVariant
 			result.ScriptKind = p.scriptKind
+			result.SetJSDocCache(p.jsdocCache)
+			p.jsdocCache = nil
```
jakebailey: Wishing we could dedupe this with the code above, since we're setting basically the same stuff onto both...

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+// JSDOC
+
+type jSDocState int32
+
+const (
+	jSDocStateBeginningOfLine jSDocState = iota
+	jSDocStateSawAsterisk
+	jSDocStateSavingComments
+	jSDocStateSavingBackticks
+)
```
```suggestion
type jsDocState int32

const (
	jsDocStateBeginningOfLine jSDocState = iota
	jsDocStateSawAsterisk
	jsDocStateSavingComments
	jsDocStateSavingBackticks
)
```

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+
+type propertyLikeParse int32
+
+const (
+	propertyLikeParseProperty          propertyLikeParse = 1 << 0
+	propertyLikeParseParameter         propertyLikeParse = 1 << 1
+	propertyLikeParseCallbackParameter propertyLikeParse = 1 << 2
```
```suggestion
	propertyLikeParseProperty propertyLikeParse = 1 << iota
	propertyLikeParseParameter
	propertyLikeParseCallbackParameter
```

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	pos := p.nodePos()
+	var hasBrace bool
+	if mayOmitBraces {
+		hasBrace = p.parseOptional(ast.KindOpenBraceToken)
+	} else {
+		p.parseExpected(ast.KindOpenBraceToken)
+	}
+	saveContextFlags := p.contextFlags
+	p.setContextFlags(ast.NodeFlagsJSDoc, true)
+	type_ := p.parseJSDocType()
```
jakebailey: This must predate the change I made to make `ts-go-go` write `t` instead of `type_` where possible...

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	propertyLikeParseCallbackParameter propertyLikeParse = 1 << 2
+)
+
+func (p *Parser) parseJSDocTypeExpression(mayOmitBraces bool) *ast.Node {
+	pos := p.nodePos()
+	var hasBrace bool
+	if mayOmitBraces {
+		hasBrace = p.parseOptional(ast.KindOpenBraceToken)
+	} else {
+		p.parseExpected(ast.KindOpenBraceToken)
```
```suggestion
		hasBrace = p.parseExpected(ast.KindOpenBraceToken)
```

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	return result
+}
+
+// Pass end=-1 to parse the text to the end
+func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int) *ast.Node {
+	if end == -1 {
+		end = len(p.sourceText)
+	}
+	// Check for /** (JSDoc opening part)
+	if !isJSDocLikeText(p.sourceText[start:]) {
```
jakebailey: This was in `parseJSDocCommentWorker` previously; not sure how much it matters.

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	initialIndent := start + 4 - (strings.LastIndex(p.sourceText[:start], "\n") + 1)
+	// -2 for trailing `*/`
+	p.sourceText = p.sourceText[:end-2]
+	p.scanner.SetText(p.sourceText)
+	// +3 for leading `/**`
+	p.scanner.ResetPos(start + 3)
+	p.setContextFlags(ast.NodeFlagsJSDoc, true)
+	p.parsingContexts = p.parsingContexts | ParsingContexts(PCJSDocComment)
+
+	comment := p.parseJSDocCommentWorker(start+3, initialIndent)
```
jakebailey: These indexes are definitely a little different than the old compiler; hopefully it's right since I can't exactly tell just from reading line for line anymore.

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+			if margin > -1 && indent+len(whitespace) > margin {
+				start := margin - indent
+				if start < 0 {
+					start += len(whitespace)
+				}
+				if start < 0 {
+					start = 0
+				}
+				comments = append(comments, whitespace[start:])
+			}
```
jakebailey: The original code was:

```ts
if (margin !== undefined && indent + whitespace.length > margin) {
    comments.push(whitespace.slice(margin - indent));
}
```

Was this relying on some sort of OOB slicing before?

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+		case ast.KindJSDocCommentTextToken:
+			state = jSDocStateSavingComments
+			pushComment(p.scanner.TokenValue())
+		case ast.KindOpenBraceToken:
+			state = jSDocStateSavingComments
+			commentEnd := p.scanner.TokenFullStart()
+			linkStart := p.scanner.TokenEnd() - 1
+			link := p.parseJSDocLink(linkStart)
+			if link != nil {
+				if linkEnd == start {
```
jakebailey: This used to be `!linkEnd`; is this hte same?

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+		if state == jSDocStateSavingComments {
+			p.nextJSDocCommentTextToken(false)
+		} else {
+			p.nextTokenJSDoc()
+		}
+	}
+	if commentsPos == -1 {
+		commentsPos = p.scanner.TokenFullStart()
+	}
+	trimmedComments := trimEnd(strings.Join(comments, ""))
```
jakebailey: The JS did `.trimEnd()`, which I believe is different than what you've written; I believe it should be doing maybe `strings.TrimEndFunc(stringutil.IsWhitespace)`, maybe?

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	if len(trimmedComments) > 0 {
```
jakebailey: The old code said `if (parts.length && trimmedComments.length) {`; so should this check `len(commentParts)` too?

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+				state = jSDocStateSavingComments
+			}
+			indent += len(whitespace)
+		case ast.KindOpenBraceToken:
+			state = jSDocStateSavingComments
+			commentEnd := p.scanner.TokenFullStart()
+			linkStart := p.scanner.TokenEnd() - 1
+			link := p.parseJSDocLink(linkStart)
+			if link != nil {
+				text := p.factory.NewJSDocText(strings.Join(comments, ""))
```
jakebailey: I feel like I'm having deja vu (isn't this identical code to above? I know it's ported so it's also duplicated before too, but, huh)

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+		if state == jSDocStateSavingComments || state == jSDocStateSavingBackticks {
+			tok = p.nextJSDocCommentTextToken(state == jSDocStateSavingBackticks)
+		} else {
+			tok = p.nextTokenJSDoc()
+		}
+	}
+
+	comments = removeLeadingNewlines(comments)
+	trimmedComments := trimEnd(strings.Join(comments, ""))
+	if len(trimmedComments) > 0 {
```
jakebailey: The old code checked `len(parts)` here; is this code semantically identical still?

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+// "NONE" is when the token is not a valid link prefix
+func (p *Parser) parseJSDocLinkPrefix() string {
+	p.skipWhitespaceOrAsterisk()
+	if p.token == ast.KindOpenBraceToken && p.nextTokenJSDoc() == ast.KindAtToken && tokenIsIdentifierOrKeyword(p.nextTokenJSDoc()) {
+		kind := p.scanner.TokenValue()
+		if isJSDocLinkTag(kind) {
+			return kind
+		}
+	}
+	return "NONE"
```
jakebailey: I think a `(string, bool)` return might be clearer here than a magic string.

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	p.finishNodeWithEnd(create, start, p.scanner.TokenEnd())
+	return create
+}
+
+func (p *Parser) parseJSDocLinkName() *ast.Node {
+	if tokenIsIdentifierOrKeyword(p.token) {
+		pos := p.nodePos()
+
+		var name *ast.Node
+		name = p.parseIdentifierName()
```
```suggestion
		name := p.parseIdentifierName()
```

internal/parser/parser.go
jakebailey, line 1: If you wanted, you could split the jsdoc code into a new file in the same package. (I know it was confined to a namespace before.)

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+}
+
+func (p *Parser) parseNestedTypeLiteral(typeExpression *ast.Node, name *ast.EntityName, target propertyLikeParse, indent int) *ast.Node {
+	if typeExpression != nil && isObjectOrObjectArrayTypeReference(typeExpression.Type()) {
+		pos := p.nodePos()
+		var child *ast.Node
+		var children []*ast.Node
+		for {
+			state := p.mark()
+			child = p.parseChildParameterOrPropertyTag(target, indent, name)
```
jakebailey: It seems like this could just be `child :=`, since it is not persisted through iterations.

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+
+func (p *Parser) parseImportTag(start int, tagName *ast.IdentifierNode, margin int, indentText string) *ast.Node {
+	afterImportTagPos := p.scanner.TokenFullStart()
+
+	var identifier *ast.IdentifierNode
+	if p.isIdentifier() {
+		identifier = p.parseIdentifier()
+	}
+
+	importClause := p.tryParseImportClause(identifier, afterImportTagPos /*isTypeOnly*/, true /*skipJsDocLeadingAsterisks*/, true)
```
jakebailey: The inline comments here appear to be misplaced.

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+	//   > type bound they must be declared on separate lines.
+	//
+	// TODO: Determine whether we should enforce this in the checker.
+	// TODO: Consider moving the `constraint` to the first type parameter as we could then remove `getEffectiveConstraintOfTypeParameter`.
+	// TODO: Consider only parsing a single type parameter if there is a constraint.
+	var constraint *ast.Node
+	if p.token == ast.KindOpenBraceToken {
+		constraint = p.parseJSDocTypeExpression(false)
+	} else {
+		constraint = nil
```
```suggestion
```

internal/parser/parser.go
```diff
@@ -2292,9 +2368,1235 @@ func (p *Parser) parseNamedExports() *ast.Node {
...
+func (p *Parser) parseJSDocTypeNameWithNamespace(nested bool) *ast.Node {
+	start := p.scanner.TokenStart()
+	typeNameOrNamespaceName := p.parseJSDocIdentifierName(nil)
+	if p.parseOptional(ast.KindDotToken) {
+		body := p.parseJSDocTypeNameWithNamespace(true)
+		var flags ast.NodeFlags
+		if nested {
+			flags = ast.NodeFlagsNestedNamespace
+		} else {
+			flags = ast.NodeFlagsNone
```
```suggestion
```

internal/parser/parser.go
```diff
@@ -5656,6 +6971,39 @@ func (p *Parser) finishNodeWithEnd(node *ast.Node, pos int, end int) {
...
+		if p.hasDeprecatedTag {
+			p.hasDeprecatedTag = false
+			node.Flags |= ast.NodeFlagsDeprecated
+		}
+		p.jsdocCache[node] = jsDoc
+	}
+}
+
+func (p *Parser) setJSDoc(node *ast.Node, jsDocs []*ast.Node) {
+}
```
jakebailey: Missing?

internal/parser/parser.go
sandersn: Yeah, although I might as well do it now. I was thinking of "tsc only" when I wrote this.

Later: I pushed it out a couple of calls but I don't know how much plumbing there is for this. The mapping from Strada to Corsa isn't particularly 1:1 (at least right now).

sandersn: Specifically, I have None in both fileLoader.parseSourceFile and program.parseSourceFile, but one or both should decide based on the file type eventually.

internal/parser/parser.go
sandersn: I extracted it into a new function finishSourceFile -- and in fact several things have been added to the first piece of code without adding them to the second, so it really does need to be deduped.

internal/parser/parser.go
sandersn: 🙃

internal/parser/parser.go
sandersn: yep, I originally ported this a long time ago, pre typescript-go even

internal/parser/parser.go
sandersn: I think this might change the number of errors compared to Strada, but for the better, since you don't need both "missing {" and missing "}" errors.

internal/parser/parser.go
sandersn: This is an intentional code move: I wanted the early-exit checks first.

internal/parser/parser.go
sandersn: Yes, it's well tested.

Previously it was `"start+3 for /**, length-5 for /***/"`. Now it's `"start + 3 for /**, end - 2 for */"`, which is a switch from start/length to start/end. (End is a few lines above now)

initialIndent's calculation moved earlier, and I also moved the constants around to match the explanation in the comment above.

internal/parser/parser.go
sandersn: haha yes. It was really bad. 

In some cases the new indentation will be different but better. See CHANGES.md

internal/parser/parser.go
sandersn: Yes, we're not using a slice from the start anymore (so that node pos/end are correct), so `linkEnd === 0` has become `linkEnd == start`.

internal/parser/parser.go
sandersn: that sounds right since JS will trim all whitespace but our trimEnd only gets single character whitespace

internal/parser/parser.go
sandersn: No, this is simplified from the old code. Strada tries to return a string for the common case of no `@link` (since that's what it used to be). Corsa always creates a NodeList and at least a single JSDocText node. This means that even the common case treats the comment text more like a node and less like trivia.

It also means that there's one less layer of checks in this PR compared to the old code.

internal/parser/parser.go
sandersn: Yeah, there are two things:
1. Comments for tags are parsed differently; special handling for backticks among others, plus ending at `@` or EOF.
2. Comments for tags follow an offside rule so that you can format your tag comments like so and have the nested text have the right indent for later processing as markdown:

```js
/**
 * @param {LongType} longName This comment
 *                            can wrap here
 *                            if you want.
 */
```

I regret this feature and I don't think anybody uses it, but it's quite complex to remove. And I haven't confirmed that (nearly) nobody is using it.

If there were just difference (2), you could maybe make the parser recursive with difference handling for `@`, but nearly every state is subtly different.

internal/parser/parser.go
sandersn: Same comment here about always returning NodeList and JSDocText.

internal/parser/parser.go
sandersn: I agree, without type checking this is a bad idea.

internal/parser/parser.go
sandersn: weird, I thought the linter caught that now

internal/parser/parser.go
sandersn: That is a great idea!
I'll do it in a separate commit so you can review the diff from this round of PR comments.

internal/parser/parser.go
sandersn: very old code in which I couldn't believe that post-name comments were the standard, and thought that ts-to-go output was wrong.

internal/parser/parser.go
sandersn: Unused! (I inlined it when I realised it was simple and only used in one place.)

internal/compiler/program.go
```diff
@@ -225,7 +225,7 @@ func (p *Program) findSourceFile(candidate string, reason FileIncludeReason) *as
 func (p *Program) parseSourceFile(fileName string) *ast.SourceFile {
 	path := tspath.ToPath(fileName, p.currentDirectory, p.host.FS().UseCaseSensitiveFileNames())
 	text, _ := p.host.FS().ReadFile(fileName)
-	sourceFile := parser.ParseSourceFile(fileName, text, p.compilerOptions.GetEmitScriptTarget())
+	sourceFile := parser.ParseSourceFile(fileName, text, p.compilerOptions.GetEmitScriptTarget(), scanner.JSDocParsingModeParseNone)
```
jakebailey: I don't think `None` is the right mode, since that would do no parsing and tsc would be wrong about `@see`/`@link`.

Probably this needs to be passed down in ProgramOptions, and then in the root be set to: https://github.com/microsoft/TypeScript/blob/7f802bbca25634c6bffe316de75771274f1d9b1d/src/compiler/executeCommandLine.ts#L797

internal/parser/parser.go
```diff
@@ -724,6 +736,18 @@ func (p *Parser) isListTerminator(kind ParsingContext) bool {
 	return false
 }
 
+func (p *Parser) parseExpectedJSDoc(kind ast.Kind) bool {
```
jakebailey, line 746: Should this be in `jsdoc.go`?

internal/parser/parser.go
```diff
@@ -1433,6 +1481,26 @@ func (p *Parser) parseOptionalToken(kind ast.Kind) *ast.Node {
 	return nil
 }
 
+func (p *Parser) parseExpectedTokenJSDoc(kind ast.Kind) *ast.Node {
```
jakebailey: Same here; the other file?

internal/parser/parser.go
```diff
@@ -5683,6 +5772,36 @@ func (p *Parser) finishNodeWithEnd(node *ast.Node, pos int, end int) {
 	node.Flags |= p.contextFlags
 }
 
+func (p *Parser) withJSDoc(node *ast.Node, hasJSDoc bool) {
```
jakebailey: Maybe this one too.

internal/parser/parser_test.go
```diff
@@ -25,7 +26,7 @@ func BenchmarkParse(b *testing.B) {
 			sourceText := f.ReadFile(b)
 
 			for range b.N {
-				ParseSourceFile(fileName, sourceText, core.ScriptTargetESNext)
+				ParseSourceFile(fileName, sourceText, core.ScriptTargetESNext, scanner.JSDocParsingModeParseNone)
```
jakebailey: We'll have to think about how we do benchmarks; what this will effectively do is hide the fact that we're getting slower to parse because of jsdoc that will be in enabled in other modes. Perhaps this needs to be `All` or `ParseForTypeErrors`.

internal/scanner/scanner.go
```diff
@@ -218,6 +218,14 @@ var (
...
 )
 
+type JSDocParsingMode int
+
+const (
+	JSDocParsingModeParseAll JSDocParsingMode = iota
+	JSDocParsingModeParseNone
+	JSDocParsingModeParseForTypeInfo
```
jakebailey, line 227: Where did `ParseForTypeErrors` go?

CHANGES.md
```diff
@@ -11,10 +11,20 @@ CHANGES.md lists intentional changes between the Strada (Typescript) and Corsa (
 3. Malformed string ImportSpecifiers (`import x as "OOPS" from "y"`) now contain the string's text instead of an empty identifier.
 4. Empty binding elements no longer have a separate kind for OmittedExpression. Instead they have Kind=BindingElement with a nil Initialiser, Name and DotDotDotToken.
 5. ShorthandPropertyAssignment no longer includes an EqualsToken as a child when it has an ObjectAssignmentInitializer.
+6. The parser always parses a JSDocText node for comments in JSDoc. `string` is no longer part of the type of `comment`.
+7. In cases where Strada did produce a JSDocText node, Corsa no longer (incorrectly) includes all leading and trailing whitespace/asterisks, as well as initial `/**`.
```
jakebailey, line 15: Hopefully nobody's depending on that.

sandersn: True. It's a bad enough bug that (1) I doubt it (2) I want to fix it and see if it breaks anybody rather than trying to keep the bug.

The switch to always returning a JSDocText array is going to be more disruptive anyway, since naive processors might just handle `string` today.

internal/parser/parser.go
sandersn: No, I'd rather have all the parseExpected* functions collocated. Changes to one should prompt authors to check the others.

internal/parser/parser.go
sandersn: Same (although this is a weird place for parseExpectedToken to begin with)

internal/parser/parser.go
sandersn: I guess it does make more sense to have the entrypoint at the top of the file.

internal/parser/parser_test.go
sandersn: I was copying what we did with Strada, which was ignore jsdoc performance except when specifically investigating it, and changing None to All manually then.
That's not great but I'm not sure what's better. Adding a jsdoc-All run for comparison?

internal/scanner/scanner.go
sandersn: Missing, because it was only implicitly used in the scanner and parser! I totally missed that. Nice catch.

internal/compiler/program.go
sandersn: I don't know enough about this part of the code to fix this:
1. Where is the root?
2. Should jsdocParsingMode be an internal property on CompilerOptions? It's on ProgramHost in executeCommandLine.ts.

Program.parseSourceFile isn't called and has no equivalent in Strada, so I have no idea what it's supposed to do.

internal/parser/parser_test.go
jakebailey: That isn't what we did in the old compiler; the old compiler set "for type errors" in tsc, then "all" (the default zero) everywhere else for compatibility.

internal/compiler/program.go
jakebailey: Oh, is this dead code? Probably leftover from the file loader PR and we should delete it

internal/scanner/scanner.go
jakebailey: Right, but the value should be plumbed all the way down from tsgo (part of the difference here)

jakebailey:
Retesting but setting `JSDocParsingModeParseForTypeErrors` everywhere, I don't see a significant change:

```
goos: linux
goarch: amd64
Parse/empty.ts-20                                       750.7n ± 1%   784.3n ± 1%  +4.48% (p=0.000 n=10)
Parse/checker.ts-20                                     45.49m ± 3%   45.23m ± 4%       ~ (p=0.739 n=10)
Parse/dom.generated.d.ts-20                             14.89m ± 2%   14.98m ± 2%       ~ (p=0.481 n=10)
Parse/Herebyfile.mjs-20                                 626.4µ ± 2%   628.6µ ± 0%       ~ (p=0.631 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   256.5µ ± 2%   257.9µ ± 1%       ~ (p=0.393 n=10)
geomean                                                 606.0µ        612.4µ       +1.06%

Parse/empty.ts-20                                       2.328Ki ± 0%   2.438Ki ± 0%  +4.70% (p=0.000 n=10)
Parse/checker.ts-20                                     24.30Mi ± 0%   24.31Mi ± 0%       ~ (p=0.075 n=10)
Parse/dom.generated.d.ts-20                             8.620Mi ± 0%   8.620Mi ± 0%       ~ (p=0.912 n=10)
Parse/Herebyfile.mjs-20                                 360.0Ki ± 0%   360.0Ki ± 0%       ~ (p=0.912 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   157.6Ki ± 0%   157.7Ki ± 0%       ~ (p=0.079 n=10)
geomean                                                 492.7Ki        497.3Ki       +0.94%

                                                      │   old.txt   │               new.txt                │
Parse/empty.ts-20                                        6.000 ± 0%    6.000 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     156.0k ± 0%   156.0k ± 0%       ~ (p=0.115 n=10)
Parse/dom.generated.d.ts-20                             72.32k ± 0%   72.32k ± 0%       ~ (p=0.317 n=10)
Parse/Herebyfile.mjs-20                                 2.599k ± 0%   2.599k ± 0%       ~ (p=1.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20    917.0 ± 0%    917.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                 2.764k        2.764k       +0.00%
¹ all samples are equal
```

But, `poop` doesn't seem to like it:

```
Benchmark 1 (11 runs): ./built/local-old/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           965ms ± 23.9ms     937ms … 1.01s           0 ( 0%)        0%
  peak_rss            454MB ± 4.47MB     447MB …  460MB          0 ( 0%)        0%
  cpu_cycles         17.6G  ±  327M     17.0G  … 18.2G           0 ( 0%)        0%
  instructions       19.9G  ±  200M     19.7G  … 20.3G           0 ( 0%)        0%
  cache_references    458M  ± 4.78M      450M  …  466M           0 ( 0%)        0%
  cache_misses       88.1M  ± 2.80M     82.2M  … 91.2M           0 ( 0%)        0%
  branch_misses      64.2M  ±  490K     63.6M  … 64.8M           0 ( 0%)        0%
Benchmark 2 (11 runs): ./built/local/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           985ms ± 28.6ms     946ms … 1.02s           0 ( 0%)          +  2.1% ±  2.4%
  peak_rss            462MB ± 3.56MB     456MB …  467MB          0 ( 0%)        💩+  1.8% ±  0.8%
  cpu_cycles         18.4G  ±  292M     17.9G  … 18.8G           0 ( 0%)        💩+  4.3% ±  1.6%
  instructions       20.6G  ±  200M     20.4G  … 21.1G           0 ( 0%)        💩+  3.3% ±  0.9%
  cache_references    477M  ± 7.65M      465M  …  489M           0 ( 0%)        💩+  4.0% ±  1.2%
  cache_misses       91.4M  ± 3.45M     85.2M  … 96.0M           0 ( 0%)          +  3.7% ±  3.2%
  branch_misses      66.1M  ±  605K     65.2M  … 66.9M           0 ( 0%)        💩+  3.0% ±  0.8%
```

internal/parser/parser_test.go
sandersn: OK, they're all All now except that main.go sets ForTypeErrors.

sandersn:
I think 4% cost is reasonable for JSDoc parsing -- at least initially. I have some non-backward-compatible ideas for how to make it faster, but it's already cheaper than JSDoc parsing in Strada.

internal/parser/parser_test.go
```diff
@@ -25,7 +26,7 @@ func BenchmarkParse(b *testing.B) {
 			sourceText := f.ReadFile(b)
 
 			for range b.N {
-				ParseSourceFile(fileName, sourceText, core.ScriptTargetESNext)
+				ParseSourceFile(fileName, sourceText, core.ScriptTargetESNext, scanner.JSDocParsingModeParseAll)
```
jakebailey, line 29: At some level I'm wondering if we should be benchmarking in `JSDocParsingModeParseForTypeErrors`, but maybe that's just not worth it since we want to make the editor case fast too.

internal/parser/parser.go
```diff
@@ -264,33 +274,35 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+	processPragmasIntoFields(result)
+	result.SetDiagnostics(attachFileToDiagnostics(p.diagnostics, result))
+	result.ExternalModuleIndicator = isFileProbablyExternalModule(result)
+	result.IsDeclarationFile = isDeclarationFile
+	result.LanguageVersion = p.languageVersion
+	result.LanguageVariant = p.languageVariant
+	result.ScriptKind = p.scriptKind
+	result.Flags |= p.sourceFlags
+	result.SetJSDocCache(p.jsdocCache)
+	p.jsdocCache = nil
```
jakebailey, line 307: I would remove this; duplicated above and not related to the source file itself, but the memory retention of the Parser.

internal/parser/parser.go
```diff
@@ -264,33 +274,35 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
+	result.CommentDirectives = p.scanner.CommentDirectives()
+	result.Pragmas = getCommentPragmas(&p.factory, p.sourceText)
+	processPragmasIntoFields(result)
+	result.SetDiagnostics(attachFileToDiagnostics(p.diagnostics, result))
+	result.ExternalModuleIndicator = isFileProbablyExternalModule(result)
+	result.IsDeclarationFile = isDeclarationFile
+	result.LanguageVersion = p.languageVersion
+	result.LanguageVariant = p.languageVariant
+	result.ScriptKind = p.scriptKind
+	result.Flags |= p.sourceFlags
```
jakebailey, line 305: Is there a chance that the source flags differ when we reparse for await?

jakebailey:
Nathan and I looked at this on a call; there were a couple of bugs (not caused by this PR) that prevented things from working, notably that we never set the ScriptKind on the scanner. That plus some performance improvements leave us at:

```
Benchmark 1 (11 runs): ./built/local-old/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           939ms ± 25.4ms     915ms …  994ms          0 ( 0%)        0%
  peak_rss            454MB ± 3.78MB     445MB …  458MB          2 (18%)        0%
  cpu_cycles         17.5G  ±  242M     17.1G  … 17.8G           0 ( 0%)        0%
  instructions       20.0G  ±  231M     19.8G  … 20.5G           0 ( 0%)        0%
  cache_references    471M  ± 8.49M      458M  …  486M           0 ( 0%)        0%
  cache_misses       81.5M  ± 2.79M     79.5M  … 89.3M           1 ( 9%)        0%
  branch_misses      64.7M  ±  581K     64.0M  … 65.8M           0 ( 0%)        0%
Benchmark 2 (11 runs): ./built/local/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           942ms ± 48.9ms     897ms … 1.08s           1 ( 9%)          +  0.4% ±  3.7%
  peak_rss            455MB ± 2.95MB     452MB …  458MB          0 ( 0%)          +  0.4% ±  0.7%
  cpu_cycles         17.3G  ±  335M     16.5G  … 17.8G           0 ( 0%)          -  0.9% ±  1.5%
  instructions       20.1G  ±  171M     19.8G  … 20.3G           0 ( 0%)          +  0.5% ±  0.9%
  cache_references    470M  ± 8.03M      460M  …  485M           0 ( 0%)          -  0.1% ±  1.6%
  cache_misses       82.2M  ± 1.90M     79.5M  … 85.0M           0 ( 0%)          +  0.8% ±  2.6%
  branch_misses      64.4M  ±  523K     63.3M  … 65.0M           1 ( 9%)          -  0.4% ±  0.8%
```

aka no difference for a tsc run, so that's great.

I think there is sill some performance to be gained when parsing JSDoc but that's not strictly required now that we got over the hurdle of a plain tsc run.

internal/parser/utilities.go
```diff
@@ -59,3 +61,31 @@ func tokenIsIdentifierOrKeyword(token ast.Kind) bool {
 func tokenIsIdentifierOrKeywordOrGreaterThan(token ast.Kind) bool {
 	return token == ast.KindGreaterThanToken || tokenIsIdentifierOrKeyword(token)
 }
+
+func getJSDocCommentRanges(f *ast.NodeFactory, commentRanges []ast.CommentRange, node *ast.Node, text string) []ast.CommentRange {
```
jakebailey, line 65: Note: this would be better as an iterator, but it's not performant enough until Go 1.24 is released.

sandersn:
A quick run of TS 5.7.3 on dom.generated.ts shows a 30% slowdown between ParseNone and ParseAll. I don't think we measured that number after the last round of performance fixes, but that's an initial target at least.

That's in a file where 49% of characters are in JSDoc. I didn't measure the increase in number of nodes.

I'm going to merge this now and re-run my AST tests afterwards -- the overhead of branch management is way lower that way.

sandersn:
AST tests are still good.

PR #226, opened 2025-01-14
Fix embedded file paths

jakebailey:
This one stray `path.Clean` messed with the source file path and confused me when I was working with the bundled PR. Fixing that means that I can use a more "normal" rooted URI, like `bundled:///` for the paths, which then fixes an issue @EricCornelson is having.

### DIFF
--- a/internal/bundled/embed.go
+++ b/internal/bundled/embed.go
@@ -12,7 +12,7 @@ import (
 
 const embedded = true
 
-const scheme = "bundled:"
+const scheme = "bundled:///"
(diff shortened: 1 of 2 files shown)


jakebailey:
I will note that these scheme paths still don't behave super well under our tspath functions; any URI is treated as unrooted (even though they are surely _always_ rooted??) and so attempting to do things like get a relative path out of a URI will fail. Thankfully, this does not matter for the lib files (or eventually, the translations).

PR #227, opened 2025-01-14
Move checker to checker package

jakebailey:
This moves all of the checker code into its own package, mainly to speed up Program compile time.


internal/checker/checker.go
```diff
@@ -490,16 +490,27 @@ type WideningContext struct {
...
+type Program interface {
+	Options() *core.CompilerOptions
+	Files() []*ast.SourceFile
+
+	BindSourceFiles()
+	GetEmitModuleFormatOfFile(sourceFile *ast.SourceFile) core.ModuleKind
+	GetResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile
+}
+
+type Host interface{}
```
jakebailey, line 510: This is empty because it's unused; will fixish.

rbuckton:
I'll probably need to refactor #229 if this merges first.

jakebailey:
I'm going to wait for Anders to okay this since he probably has something in flight.

ahejlsberg:
Quick question: Are we following up with PRs to also create `program` and `emitter` packages?

jakebailey:
The latter is already in https://github.com/microsoft/typescript-go/tree/main/internal/printer; the emit code in the compiler package is just the stuff that does the looping, more or less.

We could move Program, though that would leave the compiler package empty. Maybe no compiler package is good?

jakebailey:
Yeah, after this PR, the `compiler` package is quite literally only `Program` related code.

jakebailey:
https://github.com/microsoft/typescript-go/pull/232 shows what the rest of that movement would look like.

jakebailey:
Merge with main worked, since git does understand this is a code move, very nice.

PR #228, opened 2025-01-15
Process dynamic imports, fix program `getResolvedModule`

andrewbranch:
This fixes the issue @ahejlsberg saw with `undici-types` not being found. @types/node imports undici-types by `import("undici-types")` which wasn't being processed.

### DIFF
--- a/internal/compiler/fileloader.go
+++ b/internal/compiler/fileloader.go
@@ -46,7 +46,6 @@ func processAllProgramFiles(
 		programOptions:     programOptions,
 		compilerOptions:    compilerOptions,
 		resolver:           resolver,
-		resolvedModules:    resolvedModules,
 		defaultLibraryPath: programOptions.DefaultLibraryPath,
 		comparePathsOptions: tspath.ComparePathsOptions{
 			UseCaseSensitiveFileNames: host.FS().UseCaseSensitiveFileNames(),


internal/compiler/program.go
```diff
@@ -285,7 +288,7 @@ func (p *Program) startParseTask(fileName string, wg *core.WorkGroup, isLib bool
 		}
 
 		for _, ref := range file.TypeReferenceDirectives {
-			resolved := p.resolver.ResolveTypeReferenceDirective(ref.FileName, file.FileName(), core.ModuleKindNodeNext, nil)
+			resolved := p.resolver.ResolveTypeReferenceDirective(ref.FileName, file.FileName(), core.ModuleKindCommonJS /* !!! */, nil)
```
jakebailey: Why is this CommonJS?

jakebailey:
(though it conflicts, and probably conflicts again with Eric's PR)

jakebailey:
I took the time to merge main into this, which involved fixing up things for `checker`'s move, and then the new `fileloader.go`.

I'm still sort of confused about the resolution mode, but this + Navya's branch + my flag PR + my tsconfig types change all combined mean we can actually load the old TypeScript compiler, vscode, etc, without any modification.

andrewbranch:
Resolution mode is hardcoded to the one that allows omitting file extensions / loading directories for now. ModuleKindNodeNext is not a valid resolution mode value, so I don't know what it was doing before. The code that actually determines the resolution mode is pending.

PR #229, opened 2025-01-15
Extract `Evaluator` and `NameResolver` out of checker

rbuckton:
To better support enum emit without relying on the type checker, this moves `NameResolver` and related functionality from the `compiler` package to the `binder` package, and extracts `Evaluator` and related functionality into a new `evaluator` package.


internal/compiler/printer.go
```diff
@@ -31,17 +31,9 @@ func (c *Checker) SymbolToString(s *ast.Symbol) string {
 	return c.symbolToString(s)
 }
 
+// Deprecated: use scanner.SymbolToString(s)
```
jakebailey: How does this work long term? Isn't this actually some really complicated thing with node building and so on?

rbuckton: I've moved it back and made it a callback.

jakebailey: I guess I'm confused; the evaluator I'm aware of in the old Ts code is `createEvaluator`, which doesn't accept that; is this a different thing such that this function is required?

rbuckton: `createEvaluator` in the old compiler is only part of what needs to be extracted to support enum emit. I also needed `computeEnumMemberValues` and `getEnumMemberValue` which are intrinsically linked to `createEvaluator`, so I pulled all of the related logic out. `symbolToString` is used in diagnostics reporting, which will be defined for `checker` but won't be used for the enum emit.

internal/stringutil/util.go
```diff
@@ -1,6 +1,8 @@
 // Package stringutil Exports common rune utilities for parsing and emitting javascript
 package stringutil
 
+import "github.com/microsoft/typescript-go/internal/jsnum"
```
jakebailey: Isn't this an import cycle?

internal/stringutil/util.go
```diff
@@ -75,3 +77,28 @@ func IsHexDigit(ch rune) bool {
...
+	// This is motivated by ECMA-262 sections 9.3.1, 9.8.1, 11.1.5, and 11.2.1.
+	//
+	// Here, we test whether 'ToString(ToNumber(name))' is exactly equal to 'name'.
+	// The '+' prefix operator is equivalent here to applying the abstract ToNumber operation.
+	// Applying the 'toString()' method on a number gives us the abstract ToString operation on a number.
+	//
+	// Note that this accepts the values 'Infinity', '-Infinity', and 'NaN', and that this is intentional.
+	// This is desired behavior, because when indexing with them as numeric entities, you are indexing
+	// with the strings '"Infinity"', '"-Infinity"', and '"NaN"' respectively.
+	return jsnum.FromString(name).String() == name
```
jakebailey: Should this be in the jsnum package? This package really only has tiny string manipulation helpers...

internal/stringutil/util.go
jakebailey: Oh, this PR predates my jsnum PR which took a dependency on this package.

internal/binder/nameresolver.go
```diff
@@ -0,0 +1,579 @@
...
+			return r.CompilerOptions.GetEmitScriptTarget() < core.ScriptTargetES2017
+		}
+		if ast.IsTypeNode(node) {
+			return false
+		}
+		return node.ForEachChild(r.requiresScopeChangeWorker)
+	}
+}
+
+func (r *NameResolver) ResolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ignoreErrors bool, dontResolveAlias bool, location *ast.Node) *ast.Symbol {
```
jakebailey, line 408: Just skimming this again; this is something that isn't present on the NameResolver in the old compiler; is there some difference here that makes this something that should move? I see it referenced in the constant evaluator, but the evaluator in the old compiler doesn't seem to require it.

(More or less, it seems like there's more moving than we previously had which I'm trying to understand)

rbuckton: As I mentioned in the PR description, I'm extracting related functionality that `enum` emit depends on from the checker, `resolveEntityName` is used by `resolveEntity`, which is critical to the constant evaluator. Rather than duplicate `resolveEntity` in the checker and the enum transform, I opted to extract `resolveEntity` out of checker and centralize it within the new `ConstantEvaluator` type.

I've opted to do that in this PR instead of the `enum` emit PR to make both PRs easier to reason over.

rbuckton:
Closing in favor of the more limited `NameResolver` in #284

PR #230, opened 2025-01-15
Fix some concurrent checking race conditions

jakebailey:
It turns out that a load of stuff is now racy due to concurrent checking; none of the IDs are guarded, nor any of the other props.

I'm not entirely certain how to fix most of it yet outside the IDs, kinda.

### DIFF
--- a/internal/checker/flow.go
+++ b/internal/checker/flow.go
@@ -2586,7 +2586,7 @@ func (c *Checker) isPastLastAssignment(symbol *ast.Symbol, location *ast.Node) b
 }
 
 func (c *Checker) ensureAssignmentsMarked(symbol *ast.Symbol) {
-	if c.markedAssignmentSymbolLinks.has(symbol) && c.markedAssignmentSymbolLinks.get(symbol).lastAssignmentPos != 0 {
+	if c.markedAssignmentSymbolLinks.get(symbol).lastAssignmentPos != 0 {
 		return
 	}
 	parent := ast.FindAncestor(symbol.ValueDeclaration, ast.IsFunctionOrSourceFile)


internal/ast/utilities.go
```diff
@@ -13,17 +13,27 @@ var (
 )
 
 func GetNodeId(node *Node) NodeId {
```
jakebailey, line 17: I feel like these IDs could be unconditionally assigned at creation.

If not, I think we should drop these top level functions in favor of methods.

jakebailey:
I've been able to fix some races, but there are other races which I don't know how I can fix easily. An example:

```go
		case flags&ast.FlowFlagsReduceLabel != 0:
			c.lastFlowNode = nil
			data := flow.Node.AsFlowReduceLabelData()
			saveAntecedents := data.Target.Antecedents
			data.Target.Antecedents = data.Antecedents
			result := c.isReachableFlowNodeWorker(flow.Antecedent, false /*noCacheCheck*/)
			data.Target.Antecedents = saveAntecedents
			return result
```

Code flow modifies flow nodes, but we can't do that.

There are other complicated ones, and they don't always reproduce in testing.

internal/ast/utilities.go
ahejlsberg: Yeah, I think it would be fine to just assign the IDs at creation.

ahejlsberg:
Yes, I know about the `FlowFlagsReduceLabel` issues, I've been thinking about the best way to fix it.

internal/compiler/checker.go
```diff
@@ -3014,10 +3014,8 @@ func (c *Checker) getTypeOfExpression(node *ast.Node) *Type {
 		return quickType
 	}
 	// If a type has been cached for the node, return it.
-	if node.Flags&ast.NodeFlagsTypeCached != 0 {
```
ahejlsberg: This flag is just a hint that _someone_ at _some point_ cached a type, and thus the cache lookup can't be skipped. I don't think there is actually any race risk here.

jakebailey: While it's theoretically possible that it runs "okay", it's still a real race reported by the race detector (try `hereby build --race` and use the binary); it's possible for it to break in super weird ways because the compiler is allowed to assume certain things.

jakebailey:
I did an iffy thing and was able to get rid of the flow node race; now the flow node has a reenterable mutex that's keyed on the checker doing the work; if two checkers attempt to update the flow node, they'll wait on each other.

The problem is that I suspect there could be a place where there's an interleaving and it could deadlock, so maybe it should be reverted entirely and we figure out some other method and just accept the rest?

internal/ast/utilities.go
jakebailey: I've opted not to do that for now, but can if that's preferred.

jakebailey:
Switched to links for the assignment stuff. Not all races are fixed, but @ahejlsberg noted that the flow node problem can be fixed by making the checker keep its own list of what things have been updated to remap them.

internal/checker/flow.go
```diff
@@ -2568,24 +2568,25 @@ func (c *Checker) isPostSuperFlowNode(flow *ast.FlowNode, noCacheCheck bool) boo
...
 func (c *Checker) isPastLastAssignment(symbol *ast.Symbol, location *ast.Node) bool {
 	c.ensureAssignmentsMarked(symbol)
-	return symbol.LastAssignmentPos == 0 || location != nil && int(symbol.LastAssignmentPos) < location.Pos()
+	lastAssignmentPos := c.markedAssignmentSymbolLinks.get(symbol).lastAssignmentPos
+	return lastAssignmentPos == 0 || location != nil && int(lastAssignmentPos) < location.Pos()
 }
 
 func (c *Checker) ensureAssignmentsMarked(symbol *ast.Symbol) {
-	if symbol.LastAssignmentPos != 0 {
+	if c.markedAssignmentSymbolLinks.has(symbol) && c.markedAssignmentSymbolLinks.get(symbol).lastAssignmentPos != 0 {
```
ahejlsberg: No need for the `.has` check here, all code paths that call `ensureAssignmentsMarked` also call `.get` to get the node link, so it will be created anyways.

ahejlsberg:
Fix for the flow node race is in #235.

internal/checker/flow.go
jakebailey: Aha, yes, right.

jakebailey:
The other thing I believe is racy but I haven't tried to fix yet is the use of `GetSymbolTable`, which tacks symbol tables onto `Symbol` and other bits. I think it's "okay" because the binder happens to call this for most things, but I suspect there are still paths to racy behavior that I just haven't caught in recent runs.

PR #231, opened 2025-01-15
Ensure deterministic file ordering in program

EricCornelson:
Break out the parallel file loading to a separate file, ensure file ordering and add tests.


testdata/fixtures/program/program_test_cases.json
```diff
@@ -0,0 +1,326 @@
...
+            { "name": "c:/dev/src2/a/b/2.ts", "contents": "/// <reference path='c/1.ts' />"},
+            { "name": "c:/dev/src2/a/b/c/1.ts", "contents": "console.log('hello');"},
+            { "name": "c:/dev/src2/a/10.ts", "contents": "/// <reference path='b/c/d/9.ts' />"},
+            { "name": "c:/dev/src2/a/b/c/d/9.ts", "contents": "/// <reference path='e/8.ts' />"},
+            { "name": "c:/dev/src2/a/b/c/d/e/8.ts", "contents": "/// <reference path='7.ts' />"},
+            { "name": "c:/dev/src2/a/b/c/d/e/7.ts", "contents": "/// <reference path='f/6.ts' />"},
+            { "name": "c:/dev/src2/a/b/c/d/e/f/6.ts", "contents": "console.log('world!');"}
+        ],
+        "expectedFiles":[
+            "bundled:///libs/lib.es5.d.ts",
```
jakebailey: We can't have these paths in a test like this unless we skip the test outside of the embedded mode, which I don't think we should do. At best we could replace the prefix of the lib dir with a string like `<libs>` or something.

Does this need to be in a JSON artifact like this?

EricCornelson: I can probably do something similar to what your test did where we just check the prefix and/or the lib path. We could also just keep the expected lib list per target somewhere else since it's the same for each test

jakebailey: A prefix check / replacement would be sufficient, I think. This is kind of a temporary test.

jakebailey: But, I would honestly prefer this all to be inlined into the Go test or something; it doesn't seem worth it when the Go code for the above here would be about the same length?

EricCornelson: Ok, I removed the json fixture file and just inlined the cases

internal/compiler/program_test.go
```diff
@@ -11,7 +11,193 @@ import (
...
 )
 
-func TestLibOrdering(t *testing.T) {
+type testFile struct {
+	FileName string `json:"name"`
+	Contents string `json:"contents"`
+}
+
+type programTest struct {
+	TestName      string            `json:"name"`
```
jakebailey: The json tags can now be removed, as well as the uppercase field names.

internal/compiler/program_test.go
```diff
@@ -11,7 +11,193 @@ import (
...
+			{FileName: "c:/dev/src2/a/b/3.ts", Contents: "/// <reference path='2.ts' />"},
+			{FileName: "c:/dev/src2/a/b/2.ts", Contents: "/// <reference path='c/1.ts' />"},
+			{FileName: "c:/dev/src2/a/b/c/1.ts", Contents: "console.log('hello');"},
+			{FileName: "c:/dev/src2/a/10.ts", Contents: "/// <reference path='b/c/d/9.ts' />"},
+			{FileName: "c:/dev/src2/a/b/c/d/9.ts", Contents: "/// <reference path='e/8.ts' />"},
+			{FileName: "c:/dev/src2/a/b/c/d/e/8.ts", Contents: "/// <reference path='7.ts' />"},
+			{FileName: "c:/dev/src2/a/b/c/d/e/7.ts", Contents: "/// <reference path='f/6.ts' />"},
+			{FileName: "c:/dev/src2/a/b/c/d/e/f/6.ts", Contents: "console.log('world!');"},
+		},
+		ExpectedFiles: append(esnextLibs,
```
jakebailey: `append` will modify `esnextLibs`; you should use `slices.Concat` for all of these.

EricCornelson: done

internal/compiler/program_test.go
EricCornelson: done

internal/compiler/fileloader.go
```diff
@@ -0,0 +1,370 @@
...
+		for _, task := range tasks {
+			if !p.processedFileNames.Has(task.normalizedFilePath) {
+				p.processedFileNames.Add(task.normalizedFilePath)
+				task.start(p)
+			}
+		}
+	}
+}
+
+func collectFiles(tasks []*parseTask) (files []*ast.SourceFile, libFiles []*ast.SourceFile) {
```
jakebailey, line 81: Did using an iterator for this to avoid allocating extra slices not work out in the end?

EricCornelson:
> I think this is good, but just to double check, it's definitely the case that these results are the same as TS? I tried it out on VS Code and the results are different, but we don't have globbing yet so it's probable that it's not a good test.

Yeah, it's really hard to tell at the moment due to the way we are recursively grabbing everything from a root dir, so almost everything ends up in `rootFiles` anyway, regardless of import structure, so the file order is definitely going to be different from tsc at least until we can actually build from a tsconfig. That's why I set up all the tests to have the entry file in a separate directory from the imports.

internal/compiler/fileloader.go
EricCornelson: Just forgot. I'll be sure to follow up though

PR #232, opened 2025-01-15
Move unrelated subpackages out of compiler dir

jakebailey:
These being here is an artifact of the pre-split compiler.


jakebailey:
FWIW I personally find this package structure "too flat"; I feel like there are bits which could still be nested further into `internal`.

jakebailey:
Of course, things are racing due to the symbol ID stuff :|

internal/program/program.go
```diff
@@ -1,4 +1,4 @@
-package compiler
+package program
```
jakebailey: Downside: we often use the name `program` as a variable name.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -260,14 +260,14 @@ func compileFilesWithHost(
...
-	programOptions := compiler.ProgramOptions{
+func createProgram(host program.CompilerHost, options *core.CompilerOptions, rootFiles []string) *program.Program {
+	programOptions := program.ProgramOptions{
 		RootFiles:          rootFiles,
 		Host:               host,
 		Options:            options,
 		DefaultLibraryPath: bundled.LibPath(),
 	}
-	program := compiler.NewProgram(programOptions)
+	program := program.NewProgram(programOptions)
```
jakebailey: Note that one can shadow here and that's "okay", but once you do this, you can't touch the package anymore in scope.

gabritto (approved):

PR #235, opened 2025-01-16
Fix race conditions involving flow nodes

ahejlsberg:
This PR fixes race conditions involving `FlowFlagsReduceLabel` flow nodes by maintaining a stack of active flow reductions for flow labels instead of modifying the flow labels themselves.

### DIFF
--- a/internal/checker/flow.go
+++ b/internal/checker/flow.go
@@ -42,6 +42,7 @@ type FlowState struct {
 	refKey          string
 	depth           int
 	sharedFlowStart int
+	reduceLabels    []*ast.FlowReduceLabelData
 }
 
 func getFlowNodeOfNode(node *ast.Node) *ast.FlowNode {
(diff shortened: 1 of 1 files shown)


internal/checker/flow.go
```diff
@@ -2434,6 +2448,7 @@ func (c *Checker) isReachableFlowNode(flow *ast.FlowNode) bool {
 }
 
 func (c *Checker) isReachableFlowNodeWorker(flow *ast.FlowNode, noCacheCheck bool) bool {
+	var reduceLabels []*ast.FlowReduceLabelData
```
jakebailey, line 2451: Double checking, but `isReachableFlowNodeWorker` and `isPostSuperFlowNode` can't call each other, correct? So these separate stacks are okay?

jakebailey: And `getTypeAtFlowNode`, since I guess that's a third stack?

internal/checker/flow.go
```diff
@@ -2493,11 +2508,9 @@ func (c *Checker) isReachableFlowNodeWorker(flow *ast.FlowNode, noCacheCheck boo
 		case flags&ast.FlowFlagsReduceLabel != 0:
 			// Cache is unreliable once we start adjusting labels
 			c.lastFlowNode = nil
-			data := flow.Node.AsFlowReduceLabelData()
-			saveAntecedents := data.Target.Antecedents
-			data.Target.Antecedents = data.Antecedents
+			reduceLabels = append(reduceLabels, flow.Node.AsFlowReduceLabelData())
```
jakebailey, line 2511: Can entries ever end up duplicated?

jakebailey: (Maybe that doesn't matter, since the original code would override it _again_, which is what appending does, so nevermind)

internal/checker/flow.go
ahejlsberg: It is possible for `getTypeAtFlowNode` to recursively invoke itself (e.g. when evaluating an operand expression in a condition flow node), but when that happens, we actually *do* want a separate stack because any recursive invocation of `getFlowTypeOfReference` should be independent of outer ones. So, your question really reveals a potential issue in the existing codebase, but apparently one that doesn't arise.

It's also possible for recursion to occur between `getFlowTypeOfReference` and `isReachableFlowNode`, but again, we want the separate local stacks for those.

PR #238, opened 2025-01-16
Use iterator for collecting task results

EricCornelson:
Use an iterator to avoid unnecessary allocations when collecting parse task results.

### DIFF
--- a/internal/compiler/fileloader.go
+++ b/internal/compiler/fileloader.go
@@ -87,24 +87,26 @@ func (p *fileLoader) startTasks(tasks []*parseTask) {
 }
 
 func iterTasks(tasks []*parseTask) iter.Seq[*parseTask] {
-	var iterTasksWorker func(tasks []*parseTask, yield func(*parseTask) bool)
-	iterTasksWorker = func(tasks []*parseTask, yield func(*parseTask) bool) {
-		for _, task := range tasks {
-			if len(task.subTasks) > 0 {
-				iterTasksWorker(task.subTasks, yield)
-			}
-
-			if task.file != nil {
-				if !yield(task) {
-					return
-				}
-			}
-		}
-	}
-
 	return func(yield func(*parseTask) bool) {
 		iterTasksWorker(tasks, yield)
 	}
+}
+
+func iterTasksWorker(tasks []*parseTask, yield func(*parseTask) bool) bool {
+	for _, task := range tasks {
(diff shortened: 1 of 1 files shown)


internal/compiler/fileloader.go
```diff
@@ -78,27 +86,24 @@ func (p *fileLoader) startTasks(tasks []*parseTask) {
...
-func collectFiles(tasks []*parseTask) (files []*ast.SourceFile, libFiles []*ast.SourceFile) {
-	files = make([]*ast.SourceFile, 0)
-	libFiles = make([]*ast.SourceFile, 0)
+func iterTasks(tasks []*parseTask) iter.Seq[*parseTask] {
+	return func(yield func(*parseTask) bool) {
+		iterTasksWorker(tasks, yield)
+	}
+}
+
+func iterTasksWorker(tasks []*parseTask, yield func(*parseTask) bool) {
```
jakebailey: You can probably just lift this into the above func.

jakebailey: Oh, shoot, it's recursive. That makes it annoying, sorry, I see why it was a top level func.

It's also possible that it could be...:

```go
func iterTasks(tasks []*parseTask) iter.Seq[*parseTask] {
	return func(yield func(*parseTask) bool) {
        for _, task := range tasks {
			for subtask := range iterTasks(task.subTasks) {
				if !yield(subtask) {
					return
				}
            }
			if task.file != nil && !yield(task) {
				return
			}
		}
    }
}
```

EricCornelson: Yeah I thought about that, I just wondered if the extra yield calls would have any effect on perf (e.g. for an item n levels deep in the tree, in this code we will call yield n times on it) or if the go compiler will just turn this into a state machine under the covers

jakebailey: The Go transformation is not _that_ smart; I think in this instance it probably wouldn't work because it's recursive. The optimizations they're adding are for inlining code that's simple loops (think, the code in `(*collections.OrderedMap).Keys` or something), but I don't think this would be inlinable.

So, you probably could just undo my suggestion...

Though, my implementation above has revealed a bug in the current PR (both before/after the worker func); `iterTasksWorker` should return false if a yield returned false and then return early at the recursive call, otherwise it will continue to execute past the first yield. Using iterators the whole time avoided that since I effectively had to handle the breakout.

PR #241, opened 2025-01-21
execute command line initial implementation

iisaduan:
`tsc` can be invoked by writing `tsc` as the second position, for example 
``` 
tsgo tsc [rest of args to be passed into tsc]
```
Without `tsc` in the second position, there is no change to `tsgo` flag parsing/execution 

ToDos for after this PR:
- Things that are not yet implemented: 
  - Watch, build, incremental, and display functions like `showConfig`, `help`, `version`, are not implemented
  - Not all error handling/outputs/printing are implemented

- Possible places for refactors:
  - `System` vs `Host` vs `Program` layering: [discussion](https://github.com/microsoft/typescript-go/pull/241#discussion_r1934761478), [comment](https://github.com/microsoft/typescript-go/pull/241#discussion_r1934857100) https://github.com/microsoft/typescript-go/issues/272
  - May need to revisit path handling for tsconfig/project references: [discussion](https://github.com/microsoft/typescript-go/pull/241#discussion_r1934805652)
 
- Testing
  - most tests are not yet ported 
  - environment variables do not do anything
  - refactor testsys to get rid of `System.EndWrite()": [discussion 1](https://github.com/microsoft/typescript-go/pull/241#discussion_r1926144177 ), [discussion 2](https://github.com/microsoft/typescript-go/pull/241#discussion_r1925880734)
  - nilpointer crash: https://github.com/microsoft/typescript-go/issues/271

### DIFF
--- a/cmd/tsgo/sys.go
+++ b/cmd/tsgo/sys.go
@@ -15,7 +15,7 @@ import (
 type osSys struct {
 	writer  io.Writer
 	fs      vfs.FS
-	newline string
+	newLine string
 	cwd     string
 }
 
@@ -28,7 +28,7 @@ func (s *osSys) GetCurrentDirectory() string {
 }
 
 func (s *osSys) NewLine() string {
-	return s.newline
+	return s.newLine
 }
 
 func (s *osSys) Writer() io.Writer {
@@ -51,6 +51,6 @@ func newSystem() *osSys {
 		cwd:     tspath.NormalizePath(cwd),
 		fs:      vfs.FromOS(),
 		writer:  os.Stdout,
-		newline: core.IfElse(runtime.GOOS == "windows", "\r\n", "\n"),
+		newLine: core.IfElse(runtime.GOOS == "windows", "\r\n", "\n"),
 	}
 }


internal/execute/system.go
```diff
@@ -0,0 +1,88 @@
+package execute
+
+import (
+	"fmt"
+	"os"
+
+	ts "github.com/microsoft/typescript-go/internal/compiler"
+	dw "github.com/microsoft/typescript-go/internal/diagnosticwriter"
```
jakebailey: Naming imports like this is generally frowned upon these days in Go, except for cases where the package path doesn't match the package name, or when two names conflict. The names make it hard to copy code between packages or even files since the names could differ, and auto-imports will not suggest them.

I would just write `compiler` and `diagnosticwriter` out.

cmd/tsgo/main.go
```diff
@@ -85,6 +86,10 @@ func parseArgs() *cliOptions {
 }
 
 func main() {
+	if os.Args[1] == "tsc" {
+		sys := execute.NewSystem()
+		sys.Exit(execute.CommandLine(sys, nil, os.Args[2:]))
```
jakebailey: Generally speaking, I think we should try our best to avoid referencing `os` in a shared package; we don't want to accidentally use the "real" OS-based system in a test or browser or something, so we'll likely want to have some sort of specific function name, package, or something, that makes it clear that an OS system is being constructed.

(I already don't particularly like how I have both a virtual and an OS backed FS in the same package, along with the type definitions, such that modifying the OS implementation causes every package to recompile since they only import the package for the interface)

internal/execute/system.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/microsoft/typescript-go/internal/vfs"
+)
+
+// todo: implement system? figure out where it is implemented?
+// is it the same as fs?
+// from looking at go and main, it looks like they are different, host extends vfs sorta.
+// compilerhost in compiler/host.go does not have a sys.exit or sys.write
+type System interface {
+	Exit(status ExitStatus) ExitStatus
+	Write(p []byte) (n int, err error)
```
jakebailey: This is sort of an odd thing; can we not just have an `io.Writer` returned from a method?

internal/execute/system.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (s *osSys) Host() ts.CompilerHost {
+	return s.host
+}
+
+func (s *osSys) Exit(e ExitStatus) ExitStatus {
+	s.exit(int(e))
+	return e
+}
```
jakebailey: It'd be a divergence from the old codebase, but I think it'd be very beneficial if we didn't put `Exit` on host at all, and instead had our main return an exit code, letting the caller do what it wants. That makes it a lot more testable.

Reading `main.go`, I think you already have that, so it's really just that this method could be deleted.

internal/execute/system.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (s *osSys) Write(p []byte) (n int, err error) {
+	return s.write(p)
+}
+
+func (s *osSys) EndWrite() {
+	// do nothing, this is needed in the interface for testing
```
jakebailey: I'm curious, when is this called? Is it a flushing operation? Something else?

internal/execute/system.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (s *osSys) GetFormatOpts() *dw.FormattingOptions {
+	return &s.formatOpts
+}
+
+func NewSystem() *osSys {
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting current directory: %v\n", err)
```
jakebailey: I think this is copy/pasted, but generally speaking we don't want to have anything but main be trying to print to stdout/stderr directly.

internal/execute/verifytsc/testsys.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (s *testSys) Write(p []byte) (n int, err error) {
+	// todo: check accuracy with original
+	return fmt.Fprint(s.currentWrite, string(p))
+}
+
+func (s *testSys) EndWrite() {
+	// todo: revisit if improving tsc/build/watch unittest baselines
+	s.output = append(s.output, s.currentWrite.String())
```
jakebailey: Couldn't we just write all output to a `bytes.Buffer` or something, and then check it there? Checking line-based output seems like a lot of work.

cmd/tsgo/main.go
iisaduan: I was going to ask about that-- should i move `execute/system.go` to `main` or somewhere else out of `execute`? (maybe even its own package)?

internal/execute/system.go
iisaduan: Every occurrence of "write" is tracked in the original code for several reasons, and the tracking is most used probably with screen clears (which seem mostly relevant in watch?) The reason why this is needed is because the old code implemented `write` to be a separate entry each time we called `write`. However, because we are using stringwriters in diagnostic outputting, and we could be passing the writers through `fmt.fprint..`, several times in each diagnostic writing function, without an `endwrite()`, writing in the test system generates different output based on if it's implemented as 

```
fmt.fprint(sys, "Output::")
fmt.fprint(sys, errors)
fmt.fprint(sys, "\n")
```
vs
```
fmt.fprint(sys, "Output::" + errors + "\n")
```
and i don't see an easier way to make them the same atm. I hope to be able to simplify it eventually, but I don't want to make too many changes when watch isn't implemented yet, since that's the part where "each write" is tracked.

cmd/tsgo/main.go
jakebailey: Its own package or main, since main is "providing" the implementation because it's running on the OS.

internal/execute/system.go
jakebailey: That makes sense, though it seems like the assertion doesn't have to be per write, but it could be the entire output string buffered, right?

But, I can understand not wanting to mess with it for a port.

iisaduan: I'll continue to take a look at your buffer suggestion as I port/see what testing looks like after i finish adding tsconfig parsing to tests right now

internal/execute/verifytsc/verifytsc.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/execute"
+	"github.com/microsoft/typescript-go/internal/testutil/baseline"
+)
+
+type testTscCompile struct {
```
jakebailey: Can't this all live in `execute`? I don't see a need for this to be a separate package when all of this could live in `execute` in `_test.go` files.

iisaduan: It makes sense to me to just test the interface of `execute`, especially since the actual `sys` is in `main` now. There's a lot of implementation just for tests and there's a lot of tests, so being able to collapse the testing folder/files made it possible to just see the implementation files in one place. There's more to come with the unimplemented parts, like watch and incremental, so having the testing implementation files mixed in there probably won't be fun in the future

jakebailey: You can restrict yourself to the public interface of a package by using a `_test` package directive, i.e. in a `execute_test.go`, you write `package execute_test`, and then are required to import the package instead of gaining access to internals.

I would really like to avoid having extra packages that test other packages; it prevents the usage of the Go extension's code coverage tool, and makes `go test -cover` less accurate without other flags to force testing of seemingly unrelated packages.

cmd/tsgo/main.go
```diff
@@ -113,6 +114,10 @@ func parseArgs() *cliOptions {
 }
 
 func main() {
+	if os.Args[1] == "tsc" {
+		sys := NewSystem()
+		sys.Exit(execute.CommandLine(sys, nil, os.Args[2:]))
```
jakebailey: I feel like `sys.Exit` shouldn't exist anymore now, and CommandLine should really just stick to a return.

I see `Exit` being used as a way to print baselines, but I think it'd be better to not have that and instead record the exit code at the caller in the tests.

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,59 @@
...
+	return s.formatOpts
+}
+
+func NewSystem() *osSys {
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting current directory: %v\n", err)
+		os.Exit(int(execute.ExitStatusInvalidProject_OutputsSkipped))
+	}
+	newHost := compiler.NewCompilerHost(nil, cwd, vfs.FromOS())
```
jakebailey: cwd is probably not normalized at this point.

internal/compiler/host.go
```diff
@@ -11,6 +11,7 @@ import (
 
 type CompilerHost interface {
 	FS() vfs.FS
+	SetOptions(options *core.CompilerOptions)
```
jakebailey: I suspect this isn't required; seems odd for the host to have a setter when that's the opposite direction of the flow of data...

(I think you have a comment of the sort in another file now that I read)

jakebailey: Hm, but the host does acutally have it for other things. Makes me think the existing code isn't quite right.

internal/compiler/program.go
```diff
@@ -20,25 +21,27 @@ import (
...
-	DefaultLibraryPath string
-	JSDocParsingMode   scanner.JSDocParsingMode
+	ConfigFilePath           string
+	RootFiles                []string
+	Host                     CompilerHost
+	Options                  *core.CompilerOptions
+	SingleThreaded           bool
+	ProjectReference         []core.ProjectReference
+	DefaultLibraryPath       string
+	ConfigParsingDiagnostics []*ast.Diagnostic
```
jakebailey: This field probably conflicts / is redundatn with code I added for the "options diagnostics" (which I think is what the old compiler called them)?

Might be good to merge main.

internal/compiler/program.go
```diff
@@ -168,14 +171,31 @@ func (p *Program) Files() []*ast.SourceFile {
...
 
-func (p *Program) SourceFiles() []*ast.SourceFile { return p.files }
-func (p *Program) Options() *core.CompilerOptions { return p.compilerOptions }
-func (p *Program) Host() CompilerHost             { return p.host }
-
-func (p *Program) GetOptionsDiagnostics() []*ast.Diagnostic {
-	return p.optionsDiagnostics
+func NewProgramFromParsedCommandLine(config *tsoptions.ParsedCommandLine, host CompilerHost) *Program {
+	// todo: remove/edit setOptions call if compilerHost no longer has CompilerOptions. Since compilerHost was created before options were parsed, compiler host would
+	//     not have options yet
```
jakebailey: Yeah, this sort of split feels weird.

internal/compiler/program.go
iisaduan: Which code? The PR should be up to date with yesterday (monday) evening

internal/execute/verifytsc/verifytsc.go
iisaduan: I didnt realize that `_test` after packages would do something special! Should I keep it in the same `execute` folder?

jakebailey: Yes, you can just move the files from this package into `execute`, rename them to end in `_test.go`, then update their package directives to `execute_test`, and they should "just work".

internal/compiler/program.go
jakebailey: Hm, I guess you already replaced `GetOptionsDiagnostics` (though it's still there as a TODO that can be removed?) with `getOptionsDiagnosticsOfConfigFile`. Nevermind.

iisaduan: The todo is related with updateprogram/watch, so I will leave it for now  🤔

internal/execute/system.go
iisaduan: I would like this to be a TODO for a later PR since I would like testing to not be the priority for this PR

internal/execute/verifytsc/testsys.go
iisaduan: https://github.com/microsoft/typescript-go/pull/241#discussion_r1934655512

internal/compiler/host.go
iisaduan: > Makes me think that the existing code isn't quite right

Which existing code? I often had a hard time deciding which things should belong to which interface, so it's pretty possible that I'm just confused, and _also_ possible that the implementer of existing code also wasn't sure how things would shake out in the future/also was confused.

jakebailey: I mean the existing code implements hosts with methods that depends on options, which implies the host can answer questions about a config that hasn't even been set yet.

iisaduan: oh yeah, it hits a lot of `nil`

iisaduan: The methods on `host` so far at least do checks if `options` are needed, so I think that it'll be fine to assume that `options` have been set by the time that other code uses `host`.

iisaduan: I looked through the flow again, basically the `Host` is somewhat needed for tsoptions parsing (first command line parsing, and then tsconfig parsing). The output of tsoptions parsing will then be set in `CompilerHost`.

Altogether this sounds like we should initialize `sys` with a new type of `Host` that is just used for parsing?

jakebailey: I would sort of expect to start out with an interface of "just what the host system knows about", which is effectively just the FS/working dir/etc, then we give tsconfig parsing something based on that with fixed options, it gives back a config, then we construct the host to use for the rest of the compilation.

It seems odd to be back propagating things up to the host, but, I'm not sure.

iisaduan: Eventually for things like watch, the options would need to propagate back up to the host. It seems to make enough sense to me?

jakebailey: That feels a bit like a recipe for weird races / behavior, or confusing behavior if we want to share the host between multiple `Program`s... If we're saying this thing is created _by_ `Program` for it to use, then I'm less afraid of it, but the way I'm reading it right now is that we are constructing one with `NewCompilerHost` to _give_ to a `Program`.

jakebailey: For watch mode, I would expect us to take one program, then use it to construct another program, which would involve giving it the last program's info, not reusing the same host object with mutable state...

iisaduan: I see what you mean. I think the confusion comes from strada having `compilerOptions` be a one-and-all object to hold everything, and at each level, not every option is used. For example, `incremental` is only used in `executeCommandLine` and `emitter`, and I'm _pretty_ sure that the separate `watchOptions` object gets combined into the `compilerOptions` object at some points, but because the watchoptions aren't used in places like checker, we don't really see that.

So the `CompilerHost` would only need a subset of `compilerOptions`, and `Program` would need some overlapping subset. Since my implementation here has the parsed options being passed into the Program creation, that part for Program should be fine? 🤔

iisaduan: On the other hand, it might be wrong to reset `CompilerHost.options` during program creation 🤔, but I'm not sure that at this state we could tell, since we're only doing the config parsing once

(As far as I can tell, if I comment that line out, there's no difference in behavior atm)

internal/core/compileroptions.go
```diff
@@ -133,6 +133,16 @@ type CompilerOptions struct {
...
+
+	// Internal fields
+	// ConfigFile      func()   `json:"configFile"` // todo: TsConfigSourceFile,
+
+	// // cli
+	// ShowConfig    bool `json:"showConfig"`
+	// Watch         bool `json:"watch"`
+	// Incremental   bool `json:"incremental"`
+	// ListFilesOnly bool `json:"listFilesOnly"`
+	// Pretty bool `json:"pretty"`
```
iisaduan: note for self: these should be removed

internal/execute/tsc.go
```diff
@@ -0,0 +1,309 @@
...
+			sys.ReportDiagnostic(ast.NewCompilerDiagnostic(diagnostics.Cannot_find_a_tsconfig_json_file_at_the_current_directory_Colon_0, tspath.NormalizePath(sys.Host().GetCurrentDirectory())))
+		} else {
+			// print version
+			// print help
+		}
+		return ExitStatusDiagnosticsPresent_OutputsSkipped
+	}
+
+	// !!! convert to options with absolute paths is usualy done here, but for ease of implementation, it's done in `tsoptions.ParseCommandLine`
+	compilerOptionsFromCommandLine := commandLine.CompilerOptions()
```
iisaduan, line 74: @navya9singh

internal/execute/tsc.go
```diff
@@ -0,0 +1,309 @@
...
+
+	if configFileName != "" {
+		extendedConfigCache := map[string]*tsoptions.ExtendedConfigCacheEntry{}
+		configParseResult, exitStatus := getParsedCommandLineOfConfigFile(configFileName, compilerOptionsFromCommandLine, sys, extendedConfigCache)
```
iisaduan: @navya9singh

internal/execute/tsc.go
```diff
@@ -0,0 +1,309 @@
...
+				cb,
+				commandLine,
+			)
+		}
+	}
+
+	return ExitStatusSuccess
+}
+
+func getParsedCommandLineOfConfigFile(configFileName string, options *core.CompilerOptions, sys System, extendedConfigCache map[string]*tsoptions.ExtendedConfigCacheEntry) (*tsoptions.ParsedCommandLine, ExitStatus) {
```
iisaduan: @navya9singh (this whole function) I cut out `parseConfigFileWithSystem` because it was a layer that we did not need with the new implementation of hosts

internal/tsoptions/commandlineparser.go
```diff
@@ -50,20 +48,32 @@ func (p *CommandLineParser) GetOptionsNameMap() *NameMap {
...
-	// todo: return correct type (waiting on shared tsconfig parsing utilities)
-	// parseCommandLineWorker()
-	return &ParsedCommandLine{}
+	if commandLine == nil {
+		commandLine = []string{}
+	}
+	parser := parseCommandLineWorker(CompilerOptionsDidYouMeanDiagnostics, commandLine, host.FS())
+	optionsWithAbsolutePaths := convertToOptionsWithAbsolutePaths(parser.options, commandLineCompilerOptionsMap, host.GetCurrentDirectory())
+	o, d := convertOptionsFromJson(commandLineCompilerOptionsMap, optionsWithAbsolutePaths, host.GetCurrentDirectory(), &core.CompilerOptions{})
+	return &ParsedCommandLine{
```
iisaduan, line 67: @navya9singh (the whole function). Mainly I needed tsconfig utilities to transform `optionsBase` (a map) to `compilerOptions`, not sure if this was the best/most correct way to do it

internal/tsoptions/parsinghelpers.go
```diff
@@ -386,3 +387,27 @@ func MergeCompilerOptions(targetOptions, sourceOptions *core.CompilerOptions) *c
 	}
 	return targetOptions
 }
+
+func convertToOptionsWithAbsolutePaths(optionsBase map[string]any, optionMap map[string]*CommandLineOption, cwd string) map[string]any {
```
iisaduan, line 391: @navya9singh

internal/tsoptions/tsconfigparsing.go
```diff
@@ -208,6 +208,13 @@ func tsconfigToSourceFile(tsconfigSourceFile *TsConfigSourceFile) *ast.SourceFil
 	return tsconfigSourceFile.SourceFile
 }
 
+func NewTsconfigSourceFileFromFilePath(configFileName string, configSourceText string) *TsConfigSourceFile {
```
iisaduan, line 211: @navya9singh Should the other fields on `TsConfigSourceFile` be initialized to something?

internal/tsoptions/tsconfigparsing.go
```diff
@@ -103,7 +103,7 @@ type fileExtensionInfo struct {
 	isMixedContent bool
 	scriptKind     core.ScriptKind
 }
-type extendedConfigCacheEntry struct {
+type ExtendedConfigCacheEntry struct {
```
iisaduan, line 106: @navya9singh  exported for `execute`

internal/execute/tsc.go
```diff
@@ -0,0 +1,309 @@
...
+	tsConfigSourceFile.SourceFile.SetPath(tspath.ToPath(configFileName, cwd, sys.FS().UseCaseSensitiveFileNames()))
+	// tsConfigSourceFile.resolvedPath = tsConfigSourceFile.FileName()
+	// tsConfigSourceFile.originalFileName = tsConfigSourceFile.FileName()
+	return tsoptions.ParseJsonSourceFileConfigFileContent(
+		tsConfigSourceFile,
+		sys.Host(),
+		tspath.GetNormalizedAbsolutePath(tspath.GetDirectoryPath(configFileName), cwd),
+		options,
+		tspath.GetNormalizedAbsolutePath(configFileName, cwd),
+		nil,
```
iisaduan, line 164: @navya9singh are these all supposed to be `nil`?

internal/execute/tsc.go
```diff
@@ -0,0 +1,309 @@
...
+		for _, e := range errors {
+			sys.ReportDiagnostic(e)
+		}
+		return nil, ExitStatusDiagnosticsPresent_OutputsSkipped
+	}
+
+	tsConfigSourceFile := tsoptions.NewTsconfigSourceFileFromFilePath(configFileName, configFileText)
+	cwd := sys.Host().GetCurrentDirectory()
+	tsConfigSourceFile.SourceFile.SetPath(tspath.ToPath(configFileName, cwd, sys.FS().UseCaseSensitiveFileNames()))
+	// tsConfigSourceFile.resolvedPath = tsConfigSourceFile.FileName()
```
iisaduan, line 156: @navya9singh  are these fields are needed?

internal/execute/system.go
```diff
@@ -0,0 +1,33 @@
...
+	Host() compiler.CompilerHost
+	SetReportDiagnostics(r DiagnosticReporter)
+	ReportDiagnostic(d *ast.Diagnostic)
+	GetFormatOpts() *diagnosticwriter.FormattingOptions // todo: should this be part of Host?
+}
+
+type ExitStatus int
+
+const (
+	ExitStatusSuccess                              ExitStatus = iota
```
gabritto:
```suggestion
	ExitStatusSuccess                              ExitStatus = 0
```
Tiny nit, maybe since the exit statuses have to be fixed to match the original ones don't use iota? Or maybe just remove the numbers.

internal/execute/tsc_test.go
```diff
@@ -0,0 +1,220 @@
...
+	}),
+},`,
+		},
+			"/home/src/workspaces/solution",
+		),
+		commandLineArgs: []string{"--p", "project"},
+	}).verify(t)
+}
+
+// func TestExtends(t *testing.T) {
```
gabritto, line 134: Might be good to add a todo comment here saying we can run those tests when we have support for `extends`.

internal/compiler/program.go
```diff
@@ -164,17 +167,22 @@ func NewProgram(options ProgramOptions) *Program {
...
 }
 
-func (p *Program) SourceFiles() []*ast.SourceFile { return p.files }
-func (p *Program) Options() *core.CompilerOptions { return p.compilerOptions }
-func (p *Program) Host() CompilerHost             { return p.host }
-
-func (p *Program) GetOptionsDiagnostics() []*ast.Diagnostic {
-	return p.optionsDiagnostics
-}
+func (p *Program) SourceFiles() []*ast.SourceFile              { return p.files }
```
gabritto: Not about this PR, but: why does compiler host have compiler options in the first place?

internal/compiler/program.go
```diff
@@ -20,25 +21,27 @@ import (
...
-	currentDirectory string
+	host                     CompilerHost
+	programOptions           ProgramOptions
+	compilerOptions          *core.CompilerOptions
+	configFilePath           string
+	nodeModules              map[string]*ast.SourceFile
+	checkers                 []*checker.Checker
+	checkersByFile           map[*ast.SourceFile]*checker.Checker
+	currentDirectory         string
+	configParsingDiagnostics []*ast.Diagnostic
```
gabritto: Should `optionsDiagnostics` just be replaced by `configParsingDiagnostics`?

internal/compiler/program.go
gabritto: I see there's a discussion on this elsewhere. I also agree it's weird.

internal/execute/testsys_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (s *testSys) serializeOutput(baseline io.Writer) {
+	fmt.Fprintln(baseline, "\nOutput::")
+	// todo screen clears
+	s.baselineOutputs(baseline, 0, len(s.output))
+}
+
+func (s *testSys) diff(baseline io.Writer) {
+	// todo: watch isnt implemented
+	// todo: doesn't actually do anything rn, but don't really care atm because we aren't passing edits into the test, so we don't care abt diffs
```
gabritto: To check my understanding: this whole function will be used in the future when we have `watch` implemented, for diffing edits?

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,72 @@
...
+func (s *osSys) Writer() io.Writer {
+	return s.writer
+}
+
+func (s *osSys) EndWrite() {
+	// do nothing, this is needed in the interface for testing
+	// todo: revisit if improving tsc/build/watch unittest baselines
+}
+
+func NewSystem() *osSys {
```
jakebailey:
```suggestion
func newSystem() *osSys {
```

This isn't exported so can just be here.

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,72 @@
...
+func (s *osSys) SetReportDiagnostics(r execute.DiagnosticReporter) {
+	s.reporter = r
+}
+
+func (s *osSys) ReportDiagnostic(d *ast.Diagnostic) {
+	s.reporter(d)
+}
+
+func (s *osSys) GetFormatOpts() *diagnosticwriter.FormattingOptions {
+	return s.formatOpts
```
jakebailey: These options normally depend on flags like `--project` and so on, no? I would have thought that things like diags are printed relative to the tsconfig, but maybe not...

When does this get used?

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (s *osSys) FS() vfs.FS {
+	return s.fs
+}
+
+func (s *osSys) GetCurrentDirectory() string {
+	return s.cwd
+}
+
+func (s *osSys) SetReportDiagnostics(r execute.DiagnosticReporter) {
```
jakebailey: I'm suspicious of a setter like this, for the same reason I was suspicious of the `SetCompilerOptions`. Is this actually needed all the way at the top?

internal/execute/tsc_test.go
jakebailey: I thought we already supported extends? vscode and the compiler are working.

internal/execute/testsys_test.go
iisaduan: Yup! 

That comment should say
```suggestion
	// todo: not sure if this actually runs a diff correctly, but don't really care atm because we aren't passing edits into the test, so we don't care abt diffs
```

internal/execute/tsc_test.go
iisaduan: These tests run into nilpointer errors. I think it may be an issue with path handling that wasn't evident in the `tsconfigparsing` tests for `extends`

cmd/tsgo/sys.go
iisaduan: It's needed when formatting things that written as output (`stdout`), since the diagnosticwriter needs them. Your comment probably means each `project` should have their own `GetFormatOpts` method. This also feels like a weirdness that is similar to the location of `compilerOptions`

cmd/tsgo/sys.go
iisaduan: I think this makes sense since in `executeCommandLine`, calling this reporter means writing to the output. It also feels easiest to keep the `reportDiagnostic` function with `System` instead of passing a diagnostic reporter through every single `executeCommandLine` function, and there points (with watch) where the `reportDiagnostic` function is updated, so I think some sort of setter makes sense (original code was called `updateReportDiagnostic`)

internal/execute/tsc_test.go
jakebailey: Right, but extends doesn't have the exception when run via `tsgo`, so it feels odd for it to fail...

cmd/tsgo/sys.go
jakebailey: I don't really agree; the "system" should be providing the FS, working dir, stdout, and other system specific information, but the reporter is purely derived from those OS methods, plus information that may depend on the tsconfig and other info (the format ops, pretty). I don't think pushing the value up the chain is a good idea; it's basically a global var in disguise when the scope of the reporter is tied to the executeCommandLine code.

iisaduan: I'll refactor so that `reportDiagnostic` is just a single function in `execute` that takes in some options and `sys`. I think I probably could've done that in the first place

internal/execute/tsc.go
navya9singh: This should be good. For consistency, we could separate out the path handling from `parsedCommandLine` but I don't think this should be an issue since we're only using it to check for compilerOptions

internal/execute/tsc.go
```diff
@@ -0,0 +1,308 @@
...
+				sys.ReportDiagnostic(ast.NewCompilerDiagnostic(diagnostics.The_specified_path_does_not_exist_Colon_0, fileOrDirectory))
+				return ExitStatusDiagnosticsPresent_OutputsSkipped
+			}
+		}
+	} else if len(commandLine.FileNames()) == 0 {
+		searchPath := tspath.NormalizePath(sys.GetCurrentDirectory())
+		configFileName = findConfigFile(searchPath, sys.FS().FileExists, "tsconfig.json")
+	}
+
+	if configFileName != "" && len(commandLine.FileNames()) > 0 {
```
navya9singh: The original has `commandLine.fileNames.length === 0 && !configFileName`, so the check needs to change?

internal/execute/tsc_test.go
iisaduan: On the same test cases?

jakebailey: No, just on vscode and src/compiler; obviously not representative but I was just hoping this could run with some skips on specific tests or something. It's not a big deal, when I left my comment first I thought this was just accidentally commented out from before we supported it.

internal/tsoptions/commandlineparser.go
navya9singh: This isn't wrong, but might be a little confusing because we don't really need to do `convertToOptionsWithAbsolutePaths`. From my understanding, you really just need compiler options for `ParsedCommandLine`, so you can do that by calling `convertOptionsFromJson` like you have, directly with `parser.options` and call `convertToOptionsWithAbsolutePaths` only when needed.

internal/tsoptions/commandlineparser.go
```diff
@@ -50,20 +48,32 @@ func (p *CommandLineParser) GetOptionsNameMap() *NameMap {
...
+		ParsedConfig: &core.ParsedOptions{
+			CompilerOptions: o,
+			FileNames:       parser.fileNames,
+		},
+		ConfigFile:    nil,
+		Errors:        append(parser.errors, d...),
+		Raw:           parser, // todo: ask navya to review what this field is for
```
navya9singh: This was included in tsconfig parsing to represent the raw json that the user provided as objects.

internal/execute/tsc.go
navya9singh: Looks good! Add the comment to make it easier to understand (as given in original).

internal/execute/tsc.go
```diff
@@ -0,0 +1,308 @@
...
+				sys,
+				cb,
+				commandLine,
+			)
+		}
+	}
+
+	return ExitStatusSuccess
+}
+
```
navya9singh, line 129:
```suggestion
// Reads the config file, reports errors if any and exits if the config file cannot be found
```

internal/execute/tsc.go
navya9singh: I believe they are when files are included through project references. @jakebailey ?

internal/execute/tsc.go
navya9singh: According to the original, yes.

internal/execute/tsc.go
navya9singh: Looks good!

internal/tsoptions/parsinghelpers.go
navya9singh: Not sure I understand this. What is the benefit of passing in `optionsBase` as `map[string]any` instead of compiler options?

internal/tsoptions/tsconfigparsing.go
navya9singh: No but https://github.com/microsoft/typescript-go/pull/241#discussion_r1936356940.

internal/execute/tsc.go
jakebailey: I don't really know off the top of my head, but that sounds plausible.

internal/execute/tsc.go
iisaduan: `ParsedCommandLine` should ever only be run once, so it feels ok to just convert the paths in the first place?

internal/tsoptions/commandlineparser.go
iisaduan: In what occasions is `convertToOptionsWithAbsolutePaths` needed? I feel my port is fairly faithful regarding when the conversion happens, (it converts the options parsed from the commandline _before_ the tsconfig is parsed)

internal/tsoptions/parsinghelpers.go
iisaduan: My understanding is that it is slower and more difficult to loop through a struct, and we should avoid using`reflect` if possible?

internal/execute/tsc.go
iisaduan: dunno how i messed up the logic so badly haha

cmd/tsgo/main.go
```diff
@@ -113,6 +114,10 @@ func parseArgs() *cliOptions {
 }
 
 func main() {
+	if os.Args[1] == "tsc" {
```
```suggestion
	if args := os.Args[1:]; len(args) > 0 && args[0] == "tsc" {
```

cmd/tsgo/main.go
```diff
@@ -113,6 +114,10 @@ func parseArgs() *cliOptions {
...
+		exitCode := execute.CommandLine(newSystem(), nil, os.Args[2:])
```
iisaduan:
```suggestion
		exitCode := execute.CommandLine(newSystem(), nil, args[1:])
```

internal/tsoptions/commandlineparser.go
navya9singh: In the original it's used at a couple of more places, but since we're not using this anywhere else right now, this should work. Like I said, it's not wrong 👍

internal/tsoptions/parsinghelpers.go
navya9singh: Right. It would've mattered if we were using this outside of `ParsedCommandLine` but since we're not, this should be good!

internal/execute/tsc.go
```diff
@@ -0,0 +1,297 @@
...
+func CommandLine(sys System, cb cbType, commandLineArgs []string) ExitStatus {
+	_, exitStatus := TestCommandLine(sys, cb, commandLineArgs)
+	return exitStatus
+}
+
+func TestCommandLine(sys System, cb cbType, commandLineArgs []string) (*tsoptions.ParsedCommandLine, ExitStatus) {
+	parsedCommandLine := tsoptions.ParseCommandLine(commandLineArgs, sys)
+	return parsedCommandLine, executeCommandLineWorker(sys, cb, parsedCommandLine)
+}
```
jakebailey:
```suggestion
func TestCommandLine(sys System, cb cbType, commandLineArgs []string) (*tsoptions.ParsedCommandLine, ExitStatus) {
	parsedCommandLine := tsoptions.ParseCommandLine(commandLineArgs, sys)
	return parsedCommandLine, executeCommandLineWorker(sys, cb, parsedCommandLine)
}
```

I assume this is here only for testing?

If so, what you can do is move this declaration to a file named `export_test.go`:

```
package execute

func TestCommandLine(sys System, cb cbType, commandLineArgs []string) (*tsoptions.ParsedCommandLine, ExitStatus) {
	parsedCommandLine := tsoptions.ParseCommandLine(commandLineArgs, sys)
	return parsedCommandLine, executeCommandLineWorker(sys, cb, parsedCommandLine)
}
```

Which will then provide this function as an exported function for packages testing this package e.g. `execute_test`.

(The name `export_test.go` is not critical, but is a convention).

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/execute"
+	"github.com/microsoft/typescript-go/internal/tspath"
+	"github.com/microsoft/typescript-go/internal/vfs"
+)
+
+type osSys struct {
+	writer  io.Writer
+	fs      vfs.FS
+	newline string
```
jakebailey:
```suggestion
	newLine string
```

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,56 @@
...
+func (s *osSys) FS() vfs.FS {
+	return s.fs
+}
+
+func (s *osSys) GetCurrentDirectory() string {
+	return s.cwd
+}
+
+func (s *osSys) NewLine() string {
+	return s.newline
```
jakebailey:
```suggestion
	return s.newLine
```

cmd/tsgo/sys.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting current directory: %v\n", err)
+		os.Exit(int(execute.ExitStatusInvalidProject_OutputsSkipped))
+	}
+
+	return &osSys{
+		cwd:     tspath.NormalizePath(cwd),
+		fs:      vfs.FromOS(),
+		writer:  os.Stdout,
+		newline: core.IfElse(runtime.GOOS == "windows", "\r\n", "\n"),
```
jakebailey:
```suggestion
		newLine: core.IfElse(runtime.GOOS == "windows", "\r\n", "\n"),
```

internal/execute/outputs.go
```diff
@@ -0,0 +1,81 @@
...
+		sys.EndWrite()
+	}
+}
+
+func shouldBePretty(sys System, options *core.CompilerOptions) bool {
+	if options == nil || options.Pretty.IsTrueOrUnknown() {
+		// todo: return defaultIsPretty(sys);
+		return true
+	}
+	return options.Pretty.IsTrue()
```
jakebailey, line 45:
```suggestion
    return options == nil || options.Pretty.IsTrueOrUnknown()
```

Right?

iisaduan: Are we getting rid of `defaultIsPretty` altogether? I'm not sure if there is/is a need for the go equivalent

jakebailey: Oh, this is like some weird TTY detection thing; nevermind, leave that 😄 

We'll have to port that into the system interface.

PR #244, opened 2025-01-21
More type checking

ahejlsberg:
With this PR we now type check the VSCode project with no errors (other than ones related to `@ts-expect-error` for errors that aren't yet being reported). This means we're successfully resolving all types and symbols in the project. There's still plenty more checker code left to port, but only related to checking and not resolution. In the PR:

* Several flipped boolean conditions fixed.
* `getTypeOfPrototypeProperty` port
* `extractRedundantTemplateLiterals` port
* `getSpreadArgumentType` fixed.
* Allow `--pretty=false` on command-line.
* Avoid repeated allocations for `FlowState` and `Relater` structs.
* Port `checkImportCallExpression`.
* Port `getIterationTypesOfIteratorSlow`.
* Improved inference from return types (old code base is inconsistent here).

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -2,7 +2,7 @@
 package main
 
 import (
-	"errors"
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jessevdk/go-flags"
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/bundled"
 	ts "github.com/microsoft/typescript-go/internal/compiler"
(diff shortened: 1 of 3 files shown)


internal/checker/inference.go
```diff
@@ -1280,14 +1278,25 @@ func (c *Checker) getInferredType(n *InferenceContext, index int) *Type {
...
+				if n.compareTypes(inferredType, constraintWithThis, false) == TernaryFalse {
+					var filteredByConstraint *Type
+					if inference.priority == InferencePriorityReturnType {
+						// If we have a pure return type inference, we may succeed by removing constituents of the inferred type
+						// that aren't assignable to the constraint type (pure return type inferences are speculation anyway).
+						filteredByConstraint = c.mapType(inferredType, func(t *Type) *Type {
+							return core.IfElse(n.compareTypes(t, constraintWithThis, false) != TernaryFalse, t, c.neverType)
+						})
+					}
+					inferredType = core.IfElse(filteredByConstraint != nil && filteredByConstraint.flags&TypeFlagsNever == 0, filteredByConstraint, nil)
```
jakebailey, line 1292: Not `len(filteredByConstraint) != 0`?

internal/checker/flow.go
```diff
@@ -67,15 +68,18 @@ func (c *Checker) getFlowTypeOfReferenceEx(reference *ast.Node, declaredType *Ty
 			return declaredType
 		}
 	}
-	var f FlowState
+	c.flowStates = append(c.flowStates, FlowState{})
```
jakebailey: A gotcha with this method is that we'll never release any of the memory allocated by `flowStates`; if avoiding the allocation of this (and relaters) is helpful, we might want to consider a `sync.Pool`.

(But that isn't super critical right now)

internal/checker/flow.go
```diff
@@ -67,15 +68,18 @@ func (c *Checker) getFlowTypeOfReferenceEx(reference *ast.Node, declaredType *Ty
...
 	f.declaredType = declaredType
 	f.initialType = initialType
 	f.flowContainer = flowContainer
 	f.sharedFlowStart = len(c.sharedFlows)
+	f.reduceLabels = f.reduceLabelsBuffer[:0]
 	c.flowInvocationCount++
-	evolvedType := c.getTypeAtFlowNode(&f, flowNode).t
+	evolvedType := c.getTypeAtFlowNode(f, flowNode).t
 	c.sharedFlows = c.sharedFlows[:f.sharedFlowStart]
+	c.flowStates = c.flowStates[:len(c.flowStates)-1]
```
jakebailey: Before slicing this off, we may want to consider zeroing out the last element, otherwise it will keep holding onto everything it references in memory because it still exists in the cap.

internal/checker/relater.go
```diff
@@ -401,6 +402,7 @@ func (c *Checker) checkTypeRelatedToEx(
 			}
 		}
 	}
+	c.relaters = c.relaters[:len(c.relaters)-1]
```
jakebailey: Same gotcha here; we probably want to zero the last element to release any intermediate state held by the relater.

internal/checker/relater.go
```diff
@@ -356,7 +356,8 @@ func (c *Checker) checkTypeRelatedToEx(
 	headMessage *diagnostics.Message,
 	diagnosticOutput **ast.Diagnostic,
 ) bool {
-	r := Relater{}
+	c.relaters = append(c.relaters, Relater{})
```
jakebailey: Did the Relater begin escaping?

ahejlsberg: It was escaping all along. I just didn't realize.

internal/checker/inference.go
ahejlsberg: Will fix.

internal/checker/flow.go
ahejlsberg: The intent is for the memory allocated in the `flowStates` backing array to survive. It'll go when the containing checker goes. `sync.Pool` isn't as good of an option as it just amortizes the allocations. With the current scheme we avoid them altogether once the `flowStates` array reaches a steady state (which is likely immediately since these things never nest deeply).

internal/checker/flow.go
ahejlsberg: I think everything in the struct is already being kept alive anyways.

internal/checker/relater.go
ahejlsberg: Probably same as for the other stack, but I guess it couldn't hurt to zero them out.

internal/checker/flow.go
jakebailey: Yeah, if they aren't going to be nested deeply, then that's alright.

internal/checker/inference.go
ahejlsberg: Wait, `filteredByConstraint` is a `*Type`. Code is fine.

jakebailey: Ah, oops, the name threw me.

jakebailey:
LGTM but needs a merge from main; I merged the flag revert so you can just undo that bit now.

PR #245, opened 2025-01-21
Switch back to go flags lib

jakebailey:
The flag parser I introduced was overly strict and mistakenly handled `/` as a flag starter (which can only be disabled with a build tag, which is not helpful).

Just restore us back to using Go's flag parser, and we can replace it all once `executeCommandLine` is a thing.

I'll do a fancier thing with Go's flag parser later for the functionality I need in the "it works" PR.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -2,7 +2,7 @@
 package main
 
 import (
-	"errors"
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jessevdk/go-flags"
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/bundled"
 	ts "github.com/microsoft/typescript-go/internal/compiler"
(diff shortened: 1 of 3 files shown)


cmd/tsgo/main.go
```diff
@@ -46,40 +45,40 @@ func printMessageChain(messageChain []*ast.Diagnostic, level int) {
 
 type cliOptions struct {
 	Tsc struct {
-		Pretty    bool   `long:"pretty" description:"Get prettier errors (default: true)"`
-		ListFiles bool   `long:"listFiles" description:"List files in the program"`
-		NoLib     bool   `long:"noLib" description:"Do not load lib.d.ts files"`
-		OutDir    string `long:"outDir" description:"Emit to the given directory"`
-	} `group:"tsc options"`
+		Pretty    bool
```
jakebailey, line 48: I am keeping the grouping since it makes it a lot easier to see what I've forgotten to plumb through to `CompilerOptions`.

cmd/tsgo/main.go
```diff
@@ -46,40 +45,40 @@ func printMessageChain(messageChain []*ast.Diagnostic, level int) {
...
+	flag.StringVar(&opts.Tsc.OutDir, "outDir", "", "Emit to the given directory")
+
+	flag.BoolVar(&opts.Devel.Quiet, "q", false, "Do not print diagnostics")
+	flag.BoolVar(&opts.Devel.SingleThreaded, "singleThreaded", false, "Run in single threaded mode")
+	flag.BoolVar(&opts.Devel.ParseAndBindOnly, "parseAndBindOnly", false, "Parse and bind only")
+	flag.BoolVar(&opts.Devel.PrintTypes, "printTypes", false, "Print types defined in main.ts")
+	flag.StringVar(&opts.Devel.PprofDir, "pprofDir", "", "Generate pprof CPU/memory profiles to the given directory")
+	flag.Parse()
+
+	opts.Args.Root = flag.Arg(0)
```
jakebailey, line 81: `Arg` returns the empty string for OOB, so this is safe.

PR #249, opened 2025-01-22
Adds `EmitContext` side table and name generation

rbuckton:
This adds `EmitContext`, a side table for emit-related data associated with Nodes, as well as an implementation of the name generation logic from the old compiler. While not used directly in this PR, this will be used by the `enum` and `namespace` transforms that will follow later.

This also addresses a typo in `addToContainerChain` in binder.go that was preventing `isUniqueLocalName` from working correctly.


internal/parser/parser.go
```diff
@@ -5564,9 +5566,12 @@ func (p *Parser) parseKeywordExpression() *ast.Node {
 	return result
 }
 
-func (p *Parser) parseLiteralExpression() *ast.Node {
+func (p *Parser) parseLiteralExpression(intern bool) *ast.Node {
```
jakebailey, line 5661: When would we not want to intern?

internal/printer/generatedidentifierflags.go
```diff
@@ -0,0 +1,21 @@
...
+
+const (
+	// Kind
+
+	GeneratedIdentifierFlagsNone     = iota // Not automatically generated.
+	GeneratedIdentifierFlagsAuto            // Automatically generated identifier.
+	GeneratedIdentifierFlagsLoop            // Automatically generated identifier with a preference for '_i'.
+	GeneratedIdentifierFlagsUnique          // Unique name based on the 'text' property.
+	GeneratedIdentifierFlagsNode            // Unique name based on the node in the 'original' property.
+	GeneratedIdentifierFlagsKindMask = 7    // Mask to extract the kind of identifier from its flags.
```
jakebailey, line 13: If you want to avoid the hardcoded 7, which I assume is `0b1111`, you could do:

```suggestion
	generatedIdentifierFlagsEnd
	GeneratedIdentifierFlagsKindMask = generatedIdentifierFlagsEnd-1    // Mask to extract the kind of identifier from its flags.
```

internal/printer/namegenerator.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+}
+
+func (g *nameGenerator) generateNameForImportOrExportDeclaration(node *ast.Node /* ImportDeclaration | ExportDeclaration */) string {
+	expr := ast.GetExternalModuleName(node)
+	baseName := "module"
+	if ast.IsStringLiteral(expr) {
+		baseName = makeIdentifierFromModuleName(expr.Text())
+	}
+	return g.makeUniqueName(baseName, nil /*checkFn*/, false /*optimistic*/, false /*scoped*/, false /*privateName*/, "" /*prefix*/, "" /*suffix*/)
```
jakebailey, line 241: I feel like this is an indication that an options struct would work better; it shouldn't be any less optimal.

internal/parser/parser.go
rbuckton: We generally don't intern strings used in expressions, only in property names. This matches the current behavior in the old parser.

jakebailey: I'm surprised, we're going to compare strings a lot, especially within the same file...

Is this change just because the identifiers map is being used for something other than interning in the old compiler? I assume for the whole name generation? (Maybe we shouldn't be dual purposing it if it matters?)

internal/printer/generatedidentifierflags.go
rbuckton: Wouldn't that give it the value `4`? The mask is essentially `Auto | Loop | Unique | Node`.

internal/printer/namegenerator.go
rbuckton: Perhaps, but this is a direct port from the old compiler. It didn't make sense to change it for no reason as it would make taking fixes from the old compiler more difficult. Since there are no external callers, this code would only ever change if it was related to a bug in both this and the old compiler.

internal/printer/generatedidentifierflags.go
jakebailey: Oh, you're right, but only because the iota isn't `1 << iota`. Is this flags or not flags?

jakebailey: Eh, I still don't get it, nevermind this comment then. The old code was the same way.

rbuckton: Did you mean `(generatedIdentifierFlagsEnd << 1) - 1`? That seems like overkill for something that hasn't changed in years.

internal/parser/parser.go
```diff
@@ -5658,8 +5662,16 @@ func (p *Parser) createIdentifierWithDiagnostic(isIdentifier bool, diagnosticMes
 	return p.createMissingIdentifier()
 }
 
-func (p *Parser) internIdentifier(text string) {
-	p.identifiers.Add(text)
+func (p *Parser) internIdentifier(text string) string {
+	if identifier, ok := p.identifiers[text]; ok {
+		return identifier
+	}
+	identifier := strings.Clone(text)
```
rbuckton: NOTE: This matches the suggested behavior in https://go.dev/blog/unique regarding string interning.

jakebailey: It was sort of intentional to slice out of the original string, since I don't think they can really escape anywhere that the SourceFile itself isn't also being held onto; would be curious how much this PR affects parser perf.

rbuckton: Exact string matches are more efficient when both sides can be compared by pointer rather than by contents, per the blog post linked above, but I concur that we should compare performance.

jakebailey: What I'm seeing is this:

```
goos: linux
goarch: amd64
Parse/empty.ts-20                                       742.4n ± 1%   737.5n ± 0%  -0.65% (p=0.035 n=10)
Parse/checker.ts-20                                     43.92m ± 2%   45.12m ± 1%  +2.73% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             14.46m ± 1%   14.59m ± 1%  +0.86% (p=0.023 n=10)
Parse/Herebyfile.mjs-20                                 614.8µ ± 1%   636.7µ ± 1%  +3.58% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   249.3µ ± 0%   262.3µ ± 1%  +5.19% (p=0.000 n=10)
geomean                                                 591.3µ        605.0µ       +2.32%

Parse/empty.ts-20                                       2.328Ki ± 0%   2.328Ki ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     24.31Mi ± 0%   25.11Mi ± 0%  +3.31% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             8.620Mi ± 0%   9.027Mi ± 0%  +4.72% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 359.6Ki ± 0%   381.7Ki ± 0%  +6.13% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   157.4Ki ± 0%   169.8Ki ± 0%  +7.87% (p=0.000 n=10)
geomean                                                 492.5Ki        514.0Ki       +4.37%
¹ all samples are equal

                                                      │   old.txt   │                new.txt                │
                                                      │  allocs/op  │  allocs/op   vs base                  │
Parse/empty.ts-20                                        6.000 ± 0%    6.000 ± 0%        ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     156.0k ± 0%   164.8k ± 0%   +5.66% (p=0.000 n=10)
Parse/dom.generated.d.ts-20                             72.32k ± 0%   78.81k ± 0%   +8.98% (p=0.000 n=10)
Parse/Herebyfile.mjs-20                                 2.599k ± 0%   2.869k ± 0%  +10.39% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20    917.0 ± 0%   1109.0 ± 0%  +20.94% (p=0.000 n=10)
geomean                                                 2.764k        3.012k        +8.98%
¹ all samples are equal
```

Quite a few more allocs in parse, with a time regression. This is what I expected since we are cloning more strings, and the parse benchmark can't show any potential improvement in further stages that do string comparisons. But I do think this may matter.

Overall running on `_submodules/src/compiler` end to end:

```
Benchmark 1 (11 runs): ./built/local-old/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           974ms ± 19.2ms     944ms … 1.00s           0 ( 0%)        0%
  peak_rss            454MB ± 4.15MB     445MB …  458MB          2 (18%)        0%
  cpu_cycles         17.8G  ±  379M     17.2G  … 18.5G           0 ( 0%)        0%
  instructions       19.9G  ±  243M     19.6G  … 20.4G           0 ( 0%)        0%
  cache_references    471M  ± 8.57M      460M  …  492M           0 ( 0%)        0%
  cache_misses       90.6M  ± 1.76M     87.8M  … 93.2M           0 ( 0%)        0%
  branch_misses      64.6M  ±  553K     63.7M  … 65.4M           0 ( 0%)        0%
Benchmark 2 (11 runs): ./built/local/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           977ms ± 57.2ms     917ms … 1.14s           1 ( 9%)          +  0.3% ±  3.9%
  peak_rss            463MB ± 9.81MB     452MB …  490MB          1 ( 9%)          +  1.9% ±  1.5%
  cpu_cycles         17.8G  ±  468M     17.2G  … 18.9G           1 ( 9%)          -  0.4% ±  2.1%
  instructions       20.2G  ±  417M     19.5G  … 21.2G           0 ( 0%)          +  1.2% ±  1.5%
  cache_references    471M  ± 10.1M      453M  …  491M           0 ( 0%)          -  0.0% ±  1.8%
  cache_misses       88.0M  ± 3.70M     83.2M  … 94.9M           0 ( 0%)          -  2.8% ±  2.8%
  branch_misses      65.1M  ±  767K     63.6M  … 66.7M           1 ( 9%)          +  0.8% ±  0.9%
```

Which is more mixed, but the peak RSS does go up.

rbuckton: Is this comparing this branch to `main`, or this branch with `strings.Clone` to this branch without `strings.Clone`?

jakebailey: This branch to main; I guess what I should really be testing is with and without clone (like you said) and also one that still interns the other strings (just in a way that doesn't break `isFileLevelUniqueName`)

rbuckton: Is there a reason to intern all strings? Interning is primarily an optimization for comparison performance, especially since non-interned strings are most likely pointers to the original source file text which will hang around (making interning for memory optimization moot). The majority of comparisons are for identifiers, and we currently intern some cases of string literals used as property keys since they interact with the symbol table in the same way.

Even then the benefits are limited to comparisons within a single file since we don't do interning across the whole program. Interning every string literal could theoretically benefit something like event overloads in  `EventTarget`/`EventEmitter` subtypes, except that such definitions are predominantly found in declaration files and their consumers are often in a different file, which means pointer-based comparisons are likely to miss.

Whole program interning is likely a performance footgun as it would require frequent synchronization across parallel parse, so I doubt we would want to go that route unless we opt to be lossy by using a lock-free map and be willing to allow for cache misses. Plus, a lossy intern table could result in unstable performance characteristics as randomness and CPU pressure could result in an infrequently used source of a string entering the cache while the more frequently used source of the string misses the cache and disrupts later comparisons.

jakebailey: IIRC we do comparisons of strings other than identifiers, like in literal types, but interning may not actually be that helpful at all since it won't make map lookups any faster...

jakebailey: I did have a branch where I made every name that could be a `Symbol.Name` into a `unique.Handle[string]`, which would help comparison in many many places, but it was slower at the time since at that point we only had parsing, nothing else.

jakebailey: This is a little hard to read (sorry), but these are the results of:

- main
- this PR + no clone
- this PR + no clone + no interning (always write to the map)
- this PR + `identifiers core.Set[string]` (so no interning)0

```
goos: linux
goarch: amd64
                                                      │   old.txt   │            no-clone.txt            │           no-intern.txt            │         identifier-set.txt         │
                                                      │   sec/op    │   sec/op     vs base               │   sec/op     vs base               │   sec/op     vs base               │
Parse/empty.ts-20                                       747.5n ± 1%   744.8n ± 0%       ~ (p=0.078 n=10)   750.2n ± 1%       ~ (p=0.796 n=10)   748.2n ± 0%       ~ (p=0.838 n=10)
Parse/checker.ts-20                                     44.45m ± 1%   45.45m ± 1%  +2.25% (p=0.000 n=10)   45.54m ± 2%  +2.45% (p=0.003 n=10)   44.91m ± 1%  +1.04% (p=0.015 n=10)
Parse/dom.generated.d.ts-20                             14.67m ± 2%   14.88m ± 0%  +1.48% (p=0.043 n=10)   14.59m ± 1%       ~ (p=0.579 n=10)   14.55m ± 1%       ~ (p=0.393 n=10)
Parse/Herebyfile.mjs-20                                 623.9µ ± 1%   634.7µ ± 1%  +1.74% (p=0.002 n=10)   630.5µ ± 0%  +1.06% (p=0.011 n=10)   613.5µ ± 1%  -1.67% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   252.2µ ± 1%   262.2µ ± 1%  +3.94% (p=0.000 n=10)   258.6µ ± 1%  +2.52% (p=0.000 n=10)   253.2µ ± 1%       ~ (p=0.971 n=10)
geomean                                                 598.3µ        609.1µ       +1.80%                  605.3µ       +1.17%                  597.2µ       -0.19%

                                                      │   old.txt    │             no-clone.txt              │             no-intern.txt             │          identifier-set.txt           │
                                                      │     B/op     │     B/op      vs base                 │     B/op      vs base                 │     B/op      vs base                 │
Parse/empty.ts-20                                       2.328Ki ± 0%   2.328Ki ± 0%       ~ (p=1.000 n=10) ¹   2.328Ki ± 0%       ~ (p=1.000 n=10) ¹   2.328Ki ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     24.31Mi ± 0%   24.88Mi ± 0%  +2.34% (p=0.000 n=10)     24.88Mi ± 0%  +2.34% (p=0.000 n=10)     24.31Mi ± 0%       ~ (p=0.315 n=10)
Parse/dom.generated.d.ts-20                             8.620Mi ± 0%   8.924Mi ± 0%  +3.52% (p=0.000 n=10)     8.923Mi ± 0%  +3.51% (p=0.000 n=10)     8.621Mi ± 0%       ~ (p=0.971 n=10)
Parse/Herebyfile.mjs-20                                 359.8Ki ± 0%   378.4Ki ± 0%  +5.14% (p=0.000 n=10)     378.4Ki ± 0%  +5.17% (p=0.000 n=10)     359.7Ki ± 0%       ~ (p=0.105 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20   157.5Ki ± 0%   167.1Ki ± 0%  +6.08% (p=0.000 n=10)     167.1Ki ± 0%  +6.05% (p=0.000 n=10)     157.5Ki ± 0%       ~ (p=0.159 n=10)
geomean                                                 492.6Ki        509.3Ki       +3.39%                    509.3Ki       +3.39%                    492.5Ki       -0.01%
¹ all samples are equal

                                                      │   old.txt   │             no-clone.txt             │            no-intern.txt             │          identifier-set.txt          │
                                                      │  allocs/op  │  allocs/op   vs base                 │  allocs/op   vs base                 │  allocs/op   vs base                 │
Parse/empty.ts-20                                        6.000 ± 0%    6.000 ± 0%       ~ (p=1.000 n=10) ¹    6.000 ± 0%       ~ (p=1.000 n=10) ¹    6.000 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts-20                                     156.0k ± 0%   156.0k ± 0%  -0.00% (p=0.032 n=10)     156.0k ± 0%       ~ (p=0.211 n=10)     156.0k ± 0%       ~ (p=0.899 n=10)
Parse/dom.generated.d.ts-20                             72.32k ± 0%   72.32k ± 0%  -0.00% (p=0.000 n=10)     72.32k ± 0%  -0.00% (p=0.008 n=10)     72.32k ± 0%       ~ (p=0.129 n=10)
Parse/Herebyfile.mjs-20                                 2.599k ± 0%   2.602k ± 0%  +0.12% (p=0.000 n=10)     2.602k ± 0%  +0.12% (p=0.000 n=10)     2.599k ± 0%       ~ (p=0.087 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx-20    917.0 ± 0%    920.0 ± 0%  +0.33% (p=0.000 n=10)      920.0 ± 0%  +0.33% (p=0.000 n=10)      917.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                 2.764k        2.766k       +0.09%                    2.766k       +0.09%                    2.764k       -0.00%
¹ all samples are equal
```

What's surprising is that the one with even perf with main is the one where we stop doing any interning at all; sort of suggests that this PR is just doing a little bit more

Maybe it's just all fine; we totally need to keep an accurate identifiers set for the name generation (unless we calculate that as a second pass, which seems worse), and we may need interning for other perf, which we just aren't measuring in a benchmark yet. 

What's curious is that the "just use a set, don't intern anything" PR does not appear to super negatively affect a full tsc run:

```
Benchmark 1 (11 runs): ./built/local-old/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           938ms ± 50.4ms     904ms … 1.08s           1 ( 9%)        0%
  peak_rss            456MB ± 1.14MB     453MB …  458MB          1 ( 9%)        0%
  cpu_cycles         17.2G  ±  275M     16.7G  … 17.7G           0 ( 0%)        0%
  instructions       19.9G  ± 45.5M     19.8G  … 19.9G           0 ( 0%)        0%
  cache_references    468M  ± 7.84M      454M  …  479M           0 ( 0%)        0%
  cache_misses       80.7M  ± 1.75M     78.3M  … 84.5M           0 ( 0%)        0%
  branch_misses      64.4M  ±  396K     63.9M  … 65.1M           0 ( 0%)        0%
Benchmark 2 (11 runs): ./built/local/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           919ms ± 25.4ms     893ms …  977ms          0 ( 0%)          -  2.1% ±  3.8%
  peak_rss            456MB ± 4.68MB     448MB …  460MB          0 ( 0%)          +  0.1% ±  0.7%
  cpu_cycles         17.3G  ±  176M     16.9G  … 17.6G           1 ( 9%)          +  0.7% ±  1.2%
  instructions       20.1G  ±  296M     19.9G  … 20.7G           0 ( 0%)          +  1.5% ±  0.9%
  cache_references    473M  ± 8.82M      459M  …  492M           2 (18%)          +  1.1% ±  1.6%
  cache_misses       81.3M  ± 3.15M     77.7M  … 87.4M           0 ( 0%)          +  0.7% ±  2.8%
  branch_misses      64.6M  ±  351K     63.9M  … 65.2M           0 ( 0%)          +  0.3% ±  0.5%
```

jakebailey: Going back to this PR but removing the clone:

```
Benchmark 1 (11 runs): ./built/local-old/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           922ms ± 25.2ms     889ms …  970ms          0 ( 0%)        0%
  peak_rss            452MB ± 4.40MB     445MB …  457MB          0 ( 0%)        0%
  cpu_cycles         17.2G  ±  274M     16.7G  … 17.7G           0 ( 0%)        0%
  instructions       20.1G  ±  246M     19.8G  … 20.5G           0 ( 0%)        0%
  cache_references    468M  ± 5.76M      460M  …  482M           0 ( 0%)        0%
  cache_misses       82.6M  ± 2.44M     78.7M  … 87.0M           0 ( 0%)        0%
  branch_misses      64.5M  ±  396K     63.8M  … 65.0M           0 ( 0%)        0%
Benchmark 2 (11 runs): ./built/local/tsgo -p _submodules/TypeScript/src/compiler
  wall_time           926ms ± 18.0ms     899ms …  961ms          0 ( 0%)          +  0.5% ±  2.1%
  peak_rss            459MB ± 2.25MB     453MB …  462MB          1 ( 9%)          +  1.6% ±  0.7%
  cpu_cycles         17.4G  ±  195M     17.0G  … 17.6G           0 ( 0%)          +  0.8% ±  1.2%
  instructions       20.1G  ± 53.3M     20.0G  … 20.1G           0 ( 0%)          -  0.3% ±  0.8%
  cache_references    477M  ± 6.09M      469M  …  486M           0 ( 0%)          +  1.9% ±  1.1%
  cache_misses       82.0M  ± 2.08M     79.6M  … 87.1M           0 ( 0%)          -  0.8% ±  2.4%
  branch_misses      64.6M  ±  386K     63.8M  … 65.1M           0 ( 0%)          +  0.2% ±  0.5%
```

This all seems well within the noise; I think that may be the right move for now.

The main reason the `unique` package suggests cloning strings is because it's global, and oddly I think that's not even required anymore after https://go-review.googlesource.com/c/go/+/610738 when using the unique package.

jakebailey: So, sorry for waffling around; my feeling is just that we should probably skip the clone, retaining the old behavior while still being usable for the emit step.

internal/printer/namegenerator.go
```diff
@@ -0,0 +1,422 @@
...
+	GetTextOfNode                      func(*ast.Node) string    // callback for Printer.getTextOfNode
+	nodeIdToGeneratedName              map[ast.NodeId]string     // Map of generated names for specific nodes
+	nodeIdToGeneratedPrivateName       map[ast.NodeId]string     // Map of generated private names for specific nodes
+	autoGeneratedIdToGeneratedName     map[autoGenerateId]string // Map of generated names for temp and loop variables
+	nameGenerationScope                *nameGenerationScope
+	privateNameGenerationScope         *nameGenerationScope
+}
+
+type nameGenerationScope struct {
+	next                   *nameGenerationScope // The next nameGenerationScope in the stack
```
jakebailey, line 31: I'm sort of curious if the linked list versus stack approach affect things, but, I'll definitely defer to you on what's more convenient here on porting / not porting. I would have expected an implementation closer to the original code to have been more performant due to the ability to reuse entries in a `[]nameGenerationScope` slice (where push is append, pop is slicing off the last item).

internal/printer/namegenerator_test.go
```diff
@@ -0,0 +1,637 @@
...
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/binder"
+	"github.com/microsoft/typescript-go/internal/core"
+	"gotest.tools/v3/assert"
+)
+
+func TestTempVariable1(t *testing.T) {
+	t.Parallel()
+
+	ctx := &EmitContext{}
```
jakebailey, line 15: Minor nit, but generally we're going to want to avoid the names `ctx` and `context`, since those are terms typically reserved for `ctx context.Context`, the thing we'll have to pass around eventually for cancellation.

internal/printer/namegenerator.go
rbuckton: The nameGenerationScope-related arrays in the old compiler were necessitated by poor performance in JS for this approach, but the implementation was never very convenient as I needed to maintain an array per field to avoid allocating an object for each scope and needed to double that to differentiate between variable scope and private scope. Struct allocation in Go is far more performant than object allocation in JS, so if using a slice in Go would be more performant as well, I can make that change.

internal/printer/namegenerator_test.go
rbuckton: Given that this is in a test, will that matter here?

internal/printer/namegenerator.go
jakebailey: I think it depends how many you expect to be chained, and how often. If it can be deep, and once done the array is no longer needed, it's cheapest to just do something like:

```go
stack := make([]scope, 0, 4) // random cap
// ...
stack = append(stack, item) // push
// ...
stack[len(stack)-1] = nil // optional clearing of ref; if stack is kept alive later
stack = stack[:len(stack)-1] // pop
```

(The above is sort of psuedocode-y; probably this should involve taking the address of the last element of the array if a pointer is needed)

jakebailey: That being said it's not a strict requirement for this PR; we can perf test it separately, I was only suggesting it since it'd technically be closer to the original code.

internal/printer/namegenerator_test.go
jakebailey: It might, since we could pass in `t.Context()`, but in this particular test file? Probably not, since this is just the printer or something. So, also not required.

PR #252, opened 2025-01-22
Fix up JSON file handling, skip checking more files, fix Text for regexp

jakebailey:
This unbreaks checking `pyright`.

~We're still including the JSON files in https://github.com/microsoft/pyright/tree/main/packages/pyright-internal/src/tests/samples, which seems wrong.~

### DIFF
--- a/internal/compiler/fileloader.go
+++ b/internal/compiler/fileloader.go
@@ -218,10 +218,12 @@ func (t *parseTask) start(loader *fileLoader) {
 func (p *fileLoader) parseSourceFile(fileName string) *ast.SourceFile {
 	path := tspath.ToPath(fileName, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
 	text, _ := p.host.FS().ReadFile(fileName)
+	var sourceFile *ast.SourceFile
 	if tspath.FileExtensionIs(fileName, tspath.ExtensionJson) {
-		return parser.ParseJSONText(fileName, text)
+		sourceFile = parser.ParseJSONText(fileName, text)
+	} else {
+		sourceFile = parser.ParseSourceFile(fileName, text, p.compilerOptions.GetEmitScriptTarget())
 	}
-	sourceFile := parser.ParseSourceFile(fileName, text, p.compilerOptions.GetEmitScriptTarget())
 	sourceFile.SetPath(path)
 	return sourceFile
 }


internal/ast/ast.go
```diff
@@ -160,8 +161,10 @@ func (n *Node) Text() string {
 		return n.AsTemplateTail().Text
 	case KindJsxNamespacedName:
 		return n.AsJsxNamespacedName().Namespace.Text() + ":" + n.AsJsxNamespacedName().Name().Text()
+	case KindRegularExpressionLiteral:
+		return n.AsRegularExpressionLiteral().Text
```
jakebailey, line 165: This crashed in emit.

internal/ast/ast.go
```diff
@@ -160,8 +161,10 @@ func (n *Node) Text() string {
...
 	}
-	panic("Unhandled case in Node.Text")
+	panic(fmt.Sprintf("Unhandled case in Node.Text: %T", n.data))
```
jakebailey, line 167: We should really add these to all of the other cases.

internal/compiler/fileloader.go
```diff
@@ -218,6 +218,9 @@ func (t *parseTask) start(loader *fileLoader) {
 func (p *fileLoader) parseSourceFile(fileName string) *ast.SourceFile {
 	path := tspath.ToPath(fileName, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
 	text, _ := p.host.FS().ReadFile(fileName)
+	if tspath.FileExtensionIs(fileName, tspath.ExtensionJson) {
```
jakebailey, line 222: There's another bug here where we are loading a load of files in the program, a separate bug that needs to be fixed.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -1461,15 +1461,15 @@ func getFileNamesFromConfigSpecs(
...
 						jsonOnlyIncludeRegexes = nil
 					}
-					includeIndex := core.FindIndex(jsonOnlyIncludeRegexes, func(re *regexp2.Regexp) bool { return core.Must(re.MatchString(file)) })
-					if includeIndex != -1 {
-						key := keyMappper(file)
-						if !literalFileMap.Has(key) && !wildCardJsonFileMap.Has(key) {
-							wildCardJsonFileMap.Set(key, file)
-						}
+				}
+				includeIndex := core.FindIndex(jsonOnlyIncludeRegexes, func(re *regexp2.Regexp) bool { return core.Must(re.MatchString(file)) })
```
jakebailey, line 1465: This was just in the wrong block.

PR #207, opened 2025-01-06
Tsconfig parsing implementation

navya9singh:

PR #255, opened 2025-01-23
clean up error handling and add baselines for command line parsing

iisaduan:
Some stuff has been exported/moved for executeCommandLine

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -177,8 +177,9 @@ func main() {
 	var bindTime, checkTime time.Duration
 
 	diagnostics := program.GetOptionsDiagnostics()
+	formatOpts := getFormatOpts(host)
 	if len(diagnostics) != 0 {
-		printDiagnostics(diagnostics, host, compilerOptions)
+		printDiagnostics(diagnostics, formatOpts, compilerOptions)
 		os.Exit(1)
 	}
 
@@ -216,7 +217,7 @@ func main() {
 	runtime.ReadMemStats(&memStats)
 
 	if !opts.devel.quiet && len(diagnostics) != 0 {
-		printDiagnostics(diagnostics, host, compilerOptions)
+		printDiagnostics(diagnostics, formatOpts, compilerOptions)
 	}
 
 	if compilerOptions.ListFiles.IsTrue() {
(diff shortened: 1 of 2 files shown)


internal/tsoptions/commandlineparser_test.go
```diff
@@ -159,38 +161,36 @@ func (f commandLineSubScenario) assertParseResult(t *testing.T) {
 	t.Run(f.testName, func(t *testing.T) {
 		t.Parallel()
 		originalBaseline := f.baseline.ReadFile(t)
-		existing := parseExistingCompilerBaseline(t, originalBaseline)
+		tsBaseline := parseExistingCompilerBaseline(t, originalBaseline)
```
gabritto, line 164: For my own future reference: why can't we just use the existing baseline infra and compare the old and new baselines with e.g. `baseline.RunAgainstSubmodule`?

internal/tsoptions/commandlineparser.go
```diff
@@ -357,10 +336,12 @@ func convertJsonOptionOfEnumType(
 		return nil, nil
 	}
 	val, b := typeMap.Get(key)
-	if (val != nil) && (val != "" || b) {
+	if b {
```
gabritto: tiny nit: maybe `b` should be `ok`? I was confused at first about what `b` was.

internal/tsoptions/commandlineparser_test.go
iisaduan: The baselines wouldn't be identical [#207 (diff)](https://github.com/microsoft/typescript-go/pull/255/files#diff-a0766c16d65a7d887c1cb23b4c83e47ec56ebdf15ae69a5d4838ddd5548eba2bR183), so `RunAgainstSubmodule` will fail (it also may fail because the structure of `compilerOptions`/`watchOptions`/etc. I've also taken advantage of the baselines going to be different, and I added some extra info to the new baselines that could be helpful during porting. I've also put the baselines in a folder which matches the file structure of the go repo

internal/tsoptions/commandlineparser.go
iisaduan: I think at some point during the original port, I ran out of letters (and then the original `ok` was removed)

PR #261, opened 2025-01-24
Basic project system layers

andrewbranch:
## What currently works

- Searching up the directory spine for tsconfig.json and loading projects
- Putting files without a tsconfig and untitled buffers into an inferred project
- Typing updates the program and only reparses the changed file
- Hover shows `checker.TypeToString`
- Go-to-def returns symbol declaration locations
- Lots of log messages appear in the output panes
- Command palette has a "TypeScript Go: Restart Language Server" command

## What doesn't work

- Literally everything else

## On deck

- Closing files / general cleanup
- Caching synthetic node children and/or finding tokens/nodes without creating them
- Diagnostics?

This work should unblock someone else (or multiple other people) to flesh out some basic language service features, e.g. replacing the TypeToString hover with more realistic quick info.

## Simplifications I hope to keep

- All filenames/paths are assumed to be normalized, since they come from URIs. This removes a ton of `normalizePath` calls. Paths in the project service are either strings (always normalized) or `tspath.Path` for map keys. Strada had additional types that represented case-transformed-but-not-normalized and normalized-but-not-case-transformed paths that we can ignore. Down the line, we can hopefully adopt URIs everywhere.
- No incremental parsing simplifies lots of things outside the actual parser:
  - ScriptVersionCache is deleted
  - TextStorage is merged into ScriptInfo
  - DocumentRegistry is very simplified
- I've split things into two packages, `project` and `ls`, similar to `server` and `services`, respectively, but I'm aiming to keep the language service layer much lighter by moving most of the update complexity into `Project` (its host). The Strada codebase has this complexity awkwardly split (see `host.updateFromProject` and `host.updateFromProjectInProgress` in services.ts), with the LanguageService object directly dealing with the DocumentRegistry. In contrast, this `ls` package should be mostly just a container for editor-supporting functionality.
- I've made some simplifications around how inferred projects get created, which at some point will need to be tested with multi-root workspaces, but I feel confident multi-root workspaces won't work at all yet.
- In general, I've been ignoring anything that looks highly indirected, plugin-related, or external-project-related.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -305,8 +305,7 @@ func (s *Server) handleDefinition(req *lsproto.RequestMessage) error {
 		}
 	}
 
-	s.sendResult(req.ID, &lsproto.Definition{Locations: &lspLocations})
-	return nil
+	return s.sendResult(req.ID, &lsproto.Definition{Locations: &lspLocations})
 }
 
 func (s *Server) getFileAndProject(uri lsproto.DocumentUri) (*project.ScriptInfo, *project.Project) {
--- a/internal/project/logger.go
+++ b/internal/project/logger.go
@@ -63,14 +63,14 @@ func (l *Logger) HasLevel(level LogLevel) bool {
 func (l *Logger) msg(s string, messageType string) {
 	for _, output := range l.outputs {
 		header := fmt.Sprintf("%s %d", messageType, l.seq)
-		output.WriteString(header)
-		output.WriteString(strings.Repeat(" ", max(0, 10-len(header))))
-		output.WriteRune('[')
-		output.WriteString(time.Now().Format("15:04:05.000"))
-		output.WriteString("] ")
-		output.WriteString(s)
-		output.WriteRune('\n')
-		output.Flush() //nolint: errcheck
+		output.WriteString(header)                                      //nolint: errcheck
+		output.WriteString(strings.Repeat(" ", max(0, 10-len(header)))) //nolint: errcheck
+		output.WriteRune('[')                                           //nolint: errcheck
+		output.WriteString(time.Now().Format("15:04:05.000"))           //nolint: errcheck
+		output.WriteString("] ")                                        //nolint: errcheck
+		output.WriteString(s)                                           //nolint: errcheck
+		output.WriteRune('\n')                                          //nolint: errcheck
+		output.Flush()
 	}
 	if !l.inGroup {
 		l.seq++


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -65,6 +65,10 @@ func (r *RequestMessage) UnmarshalJSON(data []byte) error {
 
 	r.ID = raw.ID
 	r.Method = raw.Method
+	if r.Method == "shutdown" || r.Method == "exit" {
```
jakebailey: Can you use the constants here?

internal/lsp/server.go
```diff
@@ -125,14 +151,9 @@ func (s *Server) sendError(id *lsproto.ID, err error) error {
 }
 
 func (s *Server) sendResponse(resp *lsproto.ResponseMessage) error {
-	// TODO(jakebailey): temporary debug logging
-	if _, err := s.stderr.Write([]byte("RESPONSE ")); err != nil {
-		return err
-	}
-	if err := s.stderrJSON.Encode(resp); err != nil {
-		return err
+	if !s.requestTime.IsZero() {
```
jakebailey, line 154: In the future I'll probably want to change the way this package works to be less "single thread of messages" and more "a queue of events" (since that will be required to make requests back to the client, cancellation, etc), so this won't work, but I think it's all good for now.

internal/lsp/server.go
```diff
@@ -162,25 +183,112 @@ func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
...
+			FileName: documentUriToFileName(params.TextDocument.Uri),
+			Changes:  changes,
+		}},
+		nil, /*closedFiles*/
+	)
+
+	return s.sendResult(req.ID, nil)
+}
+
 func (s *Server) handleMessage(req *lsproto.RequestMessage) error {
```
jakebailey: Nit but, I would probably have put this above the individual handlers. Maybe that's a personal choice, though.

internal/lsp/server.go
```diff
@@ -162,25 +183,112 @@ func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
...
 	case *lsproto.InitializeParams:
 		return s.sendError(req.ID, lsproto.ErrInvalidRequest)
+	case *lsproto.InitializedParams:
+		return s.handleInitialized(req)
+	case *lsproto.DidOpenTextDocumentParams:
+		return s.handleDidOpen(req)
+	case *lsproto.DidChangeTextDocumentParams:
+		return s.handleDidChange(req)
 	case *lsproto.HoverParams:
+		file, project := s.GetFileAndProject(params.TextDocument.Uri)
```
jakebailey: The other ones have handler funcs, so this could probably all be hoisted out if you wanted.

internal/lsp/server.go
```diff
@@ -162,25 +183,112 @@ func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
...
-		if req.ID != nil {
-			return s.sendError(req.ID, lsproto.ErrInvalidRequest)
+		switch req.Method {
+		case lsproto.MethodShutdown:
+			s.projectService.Close()
+			s.Log("shutdown")
+			return s.sendResult(req.ID, nil)
+		case lsproto.MethodExit:
+			s.Log("exit")
+			return nil
```
jakebailey: Does this only end up working because the client kills the process and so the stdin pipe closes?

internal/project/documentregistry.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func (r *documentRegistry) getDocumentWorker(
+	scriptInfo *ScriptInfo,
+	compilerOptions *core.CompilerOptions,
+	key registryKey,
+) *ast.SourceFile {
+	scriptTarget := core.IfElse(scriptInfo.scriptKind == core.ScriptKindJSON, core.ScriptTargetJSON, compilerOptions.GetEmitScriptTarget())
+	r.documentsMu.RLock()
```
jakebailey: I think there's a logical race in this function; we check under `RLock`, and then release it if the doc isn't found, but in the time before we get to `Lock`, another goroutine may have come in and done this too.

I think this needs the two-step thing like we do elsewhere, or some other method (like a `sync.Map.LoadOrStore`?)

internal/project/logger.go
```diff
@@ -0,0 +1,77 @@
...
+	builder.WriteString(strings.Repeat(" ", max(0, 10-builder.Len())))
+	builder.WriteRune('[')
+	builder.WriteString(time.Now().Format("15:04:05.000"))
+	builder.WriteString("] ")
+	builder.WriteString(s)
+	builder.WriteRune('\n')
+	if !l.inGroup {
+		l.seq++
+	}
+	l.write([]byte(builder.String()))
```
jakebailey: Why build the string instead of writing each byte to the output?

(If you want to buffer it, I think you could also use a `bufio.Writer` somewhere at the top of the stack like in the lsp package)

internal/project/documentregistry.go
jakebailey: (That or, just use a regular mutex)

internal/project/project.go
```diff
@@ -0,0 +1,297 @@
...
+	rootFileNames   *collections.OrderedMap[tspath.Path, string]
+	compilerOptions *core.CompilerOptions
+	languageService *ls.LanguageService
+	program         *compiler.Program
+}
+
+func NewConfiguredProject(configFileName string, configFilePath tspath.Path, projectService *ProjectService) *Project {
+	project := NewProject(configFileName, ProjectKindConfigured, tspath.GetDirectoryPath(configFileName), projectService)
+	project.configFileName = configFileName
+	project.configFilePath = configFilePath
```
jakebailey, line 62: NewProject accepts `configFileName` and `configFilePath`, but then they're assigned outside, which is maybe indicitive that the new func should just do it, maybe?

internal/project/project.go
```diff
@@ -0,0 +1,297 @@
...
+	return core.GetScriptKindFromFileName(fileName)
+}
+
+func (p *Project) markFileAsDirty(path tspath.Path) {
+	p.markAsDirty()
+}
+
+func (p *Project) markAsDirty() {
+	if !p.dirty {
+		p.dirty = true
```
jakebailey, line 188: How concurrent-aware is Project needing to be? I assume, not? Since ProjectService "owns" it?

internal/project/projectservice.go
```diff
@@ -0,0 +1,686 @@
...
+		s.log(project.print(false /*writeFileNames*/, false /*writeFileExpanation*/, false /*writeFileVersionAndText*/))
+	}
+	for _, project := range s.inferredProjects {
+		s.log(project.print(false /*writeFileNames*/, false /*writeFileExpanation*/, false /*writeFileVersionAndText*/))
+	}
+
+	s.log("Open files: ")
+	for path, projectRootPath := range s.openFiles {
+		info := s.getScriptInfo(path)
+		s.log(fmt.Sprintf("\tFileName: %s ProjectRootPath: %s", info.fileName, projectRootPath))
```
jakebailey: I feel like we should (here, elsewhere) consider adding `logf` variants or something, since we seem to format strings into logs fairly often...

internal/lsp/server.go
andrewbranch: I initially had this has a map, and then I noticed it was one message at a time. I also noticed afterwards that the built-in tracing already does this for the round trip time, so maybe we could remove it altogether.

internal/lsp/server.go
andrewbranch: Yeah, but I think this is basically the way it’s supposed to work. We're guaranteed not to receive any other messages after this. https://github.com/golang/tools/blob/8baeceabcef6fd04778372724cca217b8f5a9b93/gopls/internal/server/general.go#L657-L659

internal/project/documentregistry.go
andrewbranch: Just to check that I'm understanding what you mean, IIUC, the consequence of this happening is that two goroutines could duplicate the work of parsing a source file, and then last-in to the lock would win, but the results would be identical. So the issue isn't crashing or data corruption but just unnecessary duplication.

What do you mean by the two-step thing? The document registry is expected to be heavily read-concurrent with rare writes in most cases. (I.e., during program construction, almost every call to this function will result in a parse and a map write, but after every keystroke, there will be a single write and N-1 reads for N=len(program.files).) That's why I wanted to avoid a regular mutex.

internal/project/project.go
andrewbranch: No, that first param is a name for the logger - inferred projects don't have config files.

internal/project/project.go
andrewbranch: As of now, the concurrency just comes from Program's source file parsing, which Project ultimately supplies. I don't think it's possible for this function to be called from multiple goroutines, but it's also not important that the version increments by exactly one anyway.

internal/project/documentregistry.go
jakebailey: The logical problem I'm thinking of here is two different goroutines reaching the `else` branch below, then both individually creating source files, both attempting to write into the document map (with the last one winning), then both returning their own source files instead of the winner. This is then a problem because two callers of this function got two different source files, one of which is not the one that any other caller is ever going to get. That and the ref count will be wrong, since they'd both say "there's one ref" but two things wrote to the map.

An algorithm that avoids this race would attempt to store into the map, then read back out again after to get the "real" result after races. This is what `(*sync.Map).LoadOrStore` is for; you can still do the speculative load at the start, but when you go to store, you can say "load or store" and then only one goroutine will actually store, but all will get back the same object.

In your code, using it would look maybe like:

```go
func (r *documentRegistry) getDocumentWorker(...) {
	if entryAny, ok := r.documents.Load(key); ok {
		entry := entryAny.(*registryEntry) // should be pointer; mutable
		entry.mu.Lock()
		defer entry.mu.Unlock()
		// do something

		return entry
	}
	// parse
	entry := &registryEntry{sourceFile: sourceFile} // note: refcount 0!
	newEntry := r.documents.LoadOrStore(key).(*registryEntry)
	newEntry.mu.Lock() // Take a lock; someone else may get this elsewhere
	defer newEntry.mu.Unlock()
	newEntry.refCount++
	return newEntry.sourceFile
}
```

But even this I'm not sure about; there is also the case where one `getDocumentWorker` reaches one branch where it's trying to increment the ref, but then another one is attempting to replace the entry entirely with another one...

> What do you mean by the two-step thing?

An example of the two-step thing is something like `(*SourceFile).LineMap`, which tries to check under `RLock`, then if it fails, falls back to a real lock (but has to check again). This means there's only full lock contention when the entry doesn't exist yet, and one still wins correctly. Now, I can see why that would be less good here, since you want to avoid being under the lock for all documents while parsing,

jakebailey: (Note I also noticed when writing this that the entries right now are values, so the refcount stuff isn't actually working at the moment since they will updating copies on the stack, not the actual document map)

PR #263, opened 2025-01-29
Add tsconfig and options parsing to compiler test runner

gabritto:
Add tsconfig and options parsing to compiler test runner, plus some refactoring and cleanup that I discovered as I added the new options handling code.

### DIFF
--- a/internal/testutil/runner/test_case_parser.go
+++ b/internal/testutil/runner/test_case_parser.go
@@ -9,6 +9,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 	"github.com/microsoft/typescript-go/internal/testutil/harnessutil"
 	"github.com/microsoft/typescript-go/internal/tsoptions"
+	"github.com/microsoft/typescript-go/internal/tsoptions/tsoptionstest"
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
@@ -119,7 +120,7 @@ func makeUnitsFromTest(code string, fileName string) testCaseContent {
 	for _, data := range testUnits {
 		allFiles[tspath.GetNormalizedAbsolutePath(data.name, currentDirectory)] = data.content
 	}
-	parseConfigHost := tsoptions.NewVFSParseConfigHost(allFiles, currentDirectory)
+	parseConfigHost := tsoptionstest.NewVFSParseConfigHost(allFiles, currentDirectory)
 
 	// check if project has tsconfig.json in the list of files
 	var tsConfig *tsoptions.ParsedCommandLine
--- a/internal/tsoptions/tsconfigparsing_test.go
+++ b/internal/tsoptions/tsconfigparsing_test.go
@@ -1,4 +1,4 @@
-package tsoptions
+package tsoptions_test
 
 import (
 	"bytes"
@@ -15,6 +15,8 @@ import (
 	"github.com/microsoft/typescript-go/internal/parser"
 	"github.com/microsoft/typescript-go/internal/repo"
 	"github.com/microsoft/typescript-go/internal/testutil/baseline"
+	"github.com/microsoft/typescript-go/internal/tsoptions"
+	"github.com/microsoft/typescript-go/internal/tsoptions/tsoptionstest"
 	"github.com/microsoft/typescript-go/internal/tspath"
 	"github.com/microsoft/typescript-go/internal/vfs"
 	"gotest.tools/v3/assert"
@@ -131,7 +133,7 @@ func TestParseConfigFileTextToJson(t *testing.T) {
 			for i, jsonText := range rec.input {
 				baselineContent.WriteString("Input::\n")
 				baselineContent.WriteString(jsonText + "\n")
-				parsed, errors := ParseConfigFileTextToJson("/apath/tsconfig.json", "/apath", jsonText)
+				parsed, errors := tsoptions.ParseConfigFileTextToJson("/apath/tsconfig.json", "/apath", jsonText)
 				if configText, err := jsonToReadableText(parsed); err != nil {
 					t.Fatal(err)
 				} else {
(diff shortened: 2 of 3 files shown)


internal/testutil/harnessutil/harnessutil.go
```diff
@@ -7,73 +7,99 @@ import (
...
+	"github.com/microsoft/typescript-go/internal/tsoptions"
 	"github.com/microsoft/typescript-go/internal/tspath"
 	"github.com/microsoft/typescript-go/internal/vfs"
 	"github.com/microsoft/typescript-go/internal/vfs/vfstest"
 )
 
 type TestFile struct {
-	UnitName    string
-	Content     string
-	FileOptions map[string]string
```
gabritto, line 26: We had two separate mechanisms for passing in the compiler options for a test. I've unified that and got rid of this property as a result.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -196,35 +381,24 @@ func compileFilesWithHost(
 	// 	delete compilerOptions.project;
 	// }
 
-	// establish defaults (aligns with old harness)
-	if options.NewLine == core.NewLineKindNone {
```
gabritto, line 200: These defaults are already being set in `CompileFiles` above.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -307,43 +497,66 @@ func listFilesWorker(spec *regexp.Regexp, recursive bool, folder string) ([]stri
...
 		}
-		fmt.Fprintf(&output, "@%s: %s", key, config[key])
+		fmt.Fprintf(&output, "%s=%s", key, config[key])
 	}
 	return output.String()
 }
 
-func GetFileBasedTestConfigurations(settings map[string]string, option []string) []TestConfiguration {
-	var optionEntries [][]string
+func GetFileBasedTestConfigurations(t *testing.T, settings *collections.OrderedMap[string, string], varyByOptions map[string]struct{}) []*NamedTestConfiguration {
```
gabritto: Some significant changes here:
1. This function now computes the test description for when you have varying compiler options provided via @options in test files, e.g. if you have `@strict: true, false`, the description will include `strict=true` and will become part of the baseline file/subtest name.
2. This function is also now the only one that produces test configurations.

internal/testutil/runner/compiler_runner.go
```diff
@@ -219,15 +259,12 @@ func newCompilerTest(filename string, testContent *testCaseContent, configuratio
 		}
 	}
 
-	if tsConfigOptions != nil && tsConfigOptions.ConfigFilePath != "" {
```
gabritto, line 222: In Corsa, `CompilerOptions` doesn't have a `configFile` property.
I also checked, and in Strada all our tests pass if we delete this statement, so I got rid of it here.

internal/testutil/runner/test_case_parser.go
```diff
@@ -59,8 +65,10 @@ func makeUnitsFromTest(code string, fileName string) testCaseContent {
 			// Comment line, check for global/file @options and record them
 			metaDataName := strings.ToLower(testMetaData[1])
 			metaDataValue := strings.TrimSpace(testMetaData[2])
+			if metaDataName == "currentdirectory" {
```
gabritto, line 67: In Strada, we were doing something that looked sketchy to me:
We always used `srcFolder` (i.e. `/.src`) as the current directory when parsing the tsconfig file below. I changed this so that we use the current directory provided by the test instead, and only default to `srcFolder` if no current directory was specified.

internal/tsoptions/commandlineparser.go
```diff
@@ -274,26 +274,34 @@ func (p *CommandLineParser) parseOptionValue(
 }
 
 func (p *CommandLineParser) parseListTypeOption(opt *CommandLineOption, value string) []string {
+	elements, errors := ParseListTypeOption(opt, value)
```
gabritto, line 277: Refactored the code here into non-method `ParseListTypeOption` to be able to call it from the test infra code.

internal/tsoptions/tsconfigparsing_test.go
```diff
@@ -29,45 +27,6 @@ type testConfig struct {
 	allFileList    map[string]string
 }
 
-func fixRoot(path string) string {
```
gabritto, line 32: Moved this to its own file `vfsparseconfighost` to be able to use it in test infra code.

internal/testutil/runner/test_case_parser.go
jakebailey: Does this affect any existing tests?

internal/tsoptions/vfsparseconfighost.go
```diff
@@ -0,0 +1,48 @@
+package tsoptions
+
+import (
+	"testing/fstest"
```
jakebailey: This puts test code in the public API for the package; it's possible this is "okay" and the code will get dropped out of the binary, but it is suspect; that's why I have the `vfs`/`vfstest` split just like Go has `fs` and `testing/fstest` split.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -130,46 +155,196 @@ func CompileFiles(
...
+		options.FullEmitPaths = value.(bool)
+	case "noCheck":
+		options.NoCheck = value.(bool)
+	case "reportDiagnostics":
+		options.ReportDiagnostics = value.(bool)
+	case "captureSuggestions":
+		options.CaptureSuggestions = value.(bool)
+	case "typescriptVersion":
+		options.TypescriptVersion = value.(string)
+	}
```
jakebailey, line 313: Does this need a default panic?

internal/testutil/runner/test_case_parser.go
gabritto: It doesn't, no.

internal/testutil/harnessutil/harnessutil.go
gabritto: It shouldn't need to, because its caller already validates that `key` is a valid harness option name, but doesn't hurt to add in case this goes out of sync.

gabritto: This caught a bug. Thanks. 😅

internal/tsoptions/vfsparseconfighost.go
gabritto: I see. Should I move this file to a `tsoptionstest` package?

jakebailey: Maybe, though I suspect it may cause a cycle ☹️

gabritto: I had to move a few things around, but I got it to work, I think.

internal/tsoptions/tsconfigparsing_test.go
```diff
@@ -1,4 +1,4 @@
-package tsoptions
+package tsoptions_test
```
jakebailey, line 1: 👏

PR #265, opened 2025-01-30
Do not allow per-file JSDoc parsing mode customization

andrewbranch:
Or even per-Program customization—that would make the source file caches in #261 needlessly more complicated. We should be fine to let each compiler host implementation set its own JSDoc parsing mode for all files it encounters. N.B. `compilerHost` is really only for `tsc` and should probably move out of the `compiler` package eventually—it’s not used by #261 at all.

### DIFF
--- a/internal/compiler/host.go
+++ b/internal/compiler/host.go
@@ -14,7 +14,7 @@ type CompilerHost interface {
 	GetCurrentDirectory() string
 	NewLine() string
 	Trace(msg string)
-	GetSourceFile(fileName string, languageVersion core.ScriptTarget, jsdocParsingMode scanner.JSDocParsingMode) *ast.SourceFile
+	GetSourceFile(fileName string, languageVersion core.ScriptTarget) *ast.SourceFile
 }
 
 type FileInfo struct {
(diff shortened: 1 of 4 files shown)


internal/compiler/host.go
```diff
@@ -57,10 +57,10 @@ func (h *compilerHost) Trace(msg string) {
...
 }
 
-func (h *compilerHost) GetSourceFile(fileName string, languageVersion core.ScriptTarget, jsdocParsingMode scanner.JSDocParsingMode) *ast.SourceFile {
+func (h *compilerHost) GetSourceFile(fileName string, languageVersion core.ScriptTarget) *ast.SourceFile {
 	text, _ := h.FS().ReadFile(fileName)
 	if tspath.FileExtensionIs(fileName, tspath.ExtensionJson) {
 		return parser.ParseJSONText(fileName, text)
 	}
-	return parser.ParseSourceFile(fileName, text, languageVersion, jsdocParsingMode)
+	return parser.ParseSourceFile(fileName, text, languageVersion, scanner.JSDocParsingModeParseForTypeErrors)
```
jakebailey, line 65: This won't be used by anything other than tsc, right? That isn't the impression I got from the API... I would think we need to be plumbing this down from main.

jakebailey: See also: https://github.com/microsoft/TypeScript/blob/66e1b24831bb8b44a83fb98987cb9a6caec9a98c/src/compiler/executeCommandLine.ts#L797

(which is only in tsc)

jakebailey: > . N.B. compilerHost is really only for tsc and should probably move out of the compiler package eventually—it’s not used by https://github.com/microsoft/typescript-go/pull/261 at all.

Hm, did not read this properly. I guess if it's going to move, then that's okay. We just have to remember to do it.

PR #267, opened 2025-02-03
Use correct ResolutionMode for files when resolving/loading files

EricCornelson:
Pass the correct ResolutionMode to the resolver when resolving imports/type directives, and for resolvedModules on the file.


internal/ast/ast.go
```diff
@@ -7491,6 +7543,71 @@ func (node *SourceFile) LineMap() []core.TextPos {
 	return lineMap
 }
 
+func (file *SourceFile) CollectDynamicImportOrRequireOrJsDocImportCalls() {
```
jakebailey, line 7546: This method modifies the source file; I suspect we'll need some additional protection later if we start sharing these ASTs between programs so that two programs can't possibly race to be the first to update this info (note `lineMap` above). But I assume it doesn't matter at the moment.

jakebailey: Though, I do wonder if we can come up with some method by which we don't do this here, but instead as a part of parsing or binding themselves...

internal/compiler/fileloader.go
```diff
@@ -396,54 +413,194 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile)
...
+/*
+Use `program.getModeForUsageLocation`, which retrieves the correct `compilerOptions`, instead of this function whenever possible.
+Calculates the final resolution mode for a given module reference node. This function only returns a result when module resolution
+settings allow differing resolution between ESM imports and CJS requires, or when a mode is explicitly provided via import attributes,
+which cause an `import` or `require` condition to be used during resolution regardless of module resolution settings. In absence of
+overriding attributes, and in modes that support differing resolution, the result indicates the syntax the usage would emit to JavaScript.
+Some examples:
+
+```ts
+// tsc foo.mts --module nodenext
```
andrewbranch: I don't know how much of this comment is relevant now, but this format isn't going to show correctly in gopls. I think 4-space indentation works for code snippets. JSDoc-style `@param` tags also don't do anything.

internal/compiler/fileloader.go
```diff
@@ -396,54 +413,194 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile)
...
+func getModeForTypeReferenceDirectiveInFile(ref *ast.FileReference, item *sourceFileMetadata, options *core.CompilerOptions) core.ResolutionMode {
+	if ref.ResolutionMode != core.ResolutionModeNone {
+		return ref.ResolutionMode
+	} else {
+		return getDefaultResolutionModeForFile(item, options)
+	}
+}
+
+func getDefaultResolutionModeForFile(item *sourceFileMetadata, options *core.CompilerOptions) core.ResolutionMode {
+	if importSyntaxAffectsModuleResolution(options) {
```
andrewbranch, line 454: Do these functions belong in fileloader.go?

EricCornelson: Yeah, I'm not sure. I thought about making all of these methods on `core.CompilerOptions` but I'm not sure if that's the right place either

jakebailey: If they accept `sourceFileMetadata`, that doesn't sounds like it'd be a method on the compiler options; I feel like this is definitely a feature of the Program and not `CompilerOptions`.

internal/ast/ast.go
EricCornelson: It looks to me like this could be done at parse time, but we are specifically just setting the `NodeFlagsPossiblyContainsDynamicImport` flag in `parseImportType()`, so I guess I'd presume there's some important reason to defer it. Maybe to do with incremental parsing (which might not be valid anymore)?

rbuckton: It looks like this is already being handled in Parser.

internal/compiler/fileloader.go
```diff
@@ -176,19 +183,21 @@ type parseTask struct {
 func (t *parseTask) start(loader *fileLoader) {
 	loader.wg.Run(func() {
 		file := loader.parseSourceFile(t.normalizedFilePath)
+		metadata := loader.buildFileMetadata(file)
```
rbuckton, line 186: The emitter needs to know the implied node format to emit the correct output for `.ts` files. In  Strada, this was available via `SourceFile.impliedNodeFormat` and was passed as an option to `createSourceFile`. Is there any way to access this information from the emitter?

jakebailey:
Closing in favor of #851 and the previous PRs which implemented most of the loader related stuff.

PR #268, opened 2025-02-03
Add starter LSP implementation, test extension

jakebailey:
This PR starts the implementation of LSP. It's not complete by any means, but is enough to talk between VS Code and the server, and provide a hover that says "It works!".

I've added a test extension, along with build/launch tasks which make it easy to start. Ensure `launch.json` is copied to get that.

There is no watch mode for the Go code yet; I have a WIP change for that but it doesn't behave well with VS Code so I need to work on that. For now, you have to close the launched editor and restart the launch task to rebuild.

The text extension host is launched with options that disables VS Code's own TS language feature extensions, so hitting F5 should be all one needs.

In the future, I would like to generate the LSP code from https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/metaModel.json (which is what gopls does), but getting that working is another story.


internal/lsp/lsproto/baseproto.go
```diff
@@ -0,0 +1,140 @@
...
+	Code Integer
+}
+
+func (e *ErrorCode) Error() string {
+	return e.Name
+}
+
+var (
+	// Defined by JSON-RPC
+	ErrParseError     = &ErrorCode{"ParseError", -32700}
```
jakebailey, line 105: We'll see if these predefined errors help much; it's possible the server layer will have to convert our own problems into these instead.

internal/lsp/server.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
+	s.initializeParams = req.Params.(*lsproto.InitializeParams)
+	return s.sendResult(req.ID, &lsproto.InitializeResult{
+		ServerInfo: &lsproto.ServerInfo{
+			Name:    "typescript-go",
+			Version: core.Version,
+		},
+		Capabilities: map[string]any{
+			"textDocumentSync": lsproto.TextDocumentSyncKindIncremental,
```
jakebailey, line 151: I don't know if we've implemented diffing yet but if not, then this should be changed to `Full`.

andrewbranch: Yes, incremental changes are implemented

Herebyfile.mjs
```diff
@@ -71,20 +71,26 @@ function isInstalled(tool) {
     return !!which.sync(tool, { nothrow: true });
 }
 
-export const generateLibs = task({
-    name: "lib",
-    run: async () => {
-        await fs.promises.mkdir("./built/local", { recursive: true });
+const libsDir = "./internal/bundled/libs";
```
jakebailey, line 77: I'm going to revert these unrelated changes; I missed them in a revert of my WIP code to do watch mode for the build.

PR #270, opened 2025-02-03
Add `NodeVisitor` hooks

rbuckton:
This adds a set of hooks to override certain aspects of `NodeVisitor` behavior, much like how the old compiler allowed you to pass the `nodesVisitor`, `tokenVisitor`, and `nodeVisitor` callbacks to `visitEachChild`. This was mostly used in the language service, but will also be used to control the behavior of `SetOriginal` given how that information will be kept in a side table.

In the old compiler you might have invoked `visitEachChild` as follows:
```js
visitEachChild(node, visitor, context, nodesVisitor, tokenVisitor, nodeVisitor);
```

In Go, you will instead define these hooks on the `NodeVisitor` you create:

```go
v := &ast.NodeFactory{
    Hooks: ast.NodeVisitorHooks{
        VisitNode: func(node *ast.Node, v *ast.NodeVisitor) *ast.Node {
            if node != nil {
                // handle a non-token node
            }
            return v.VisitNode(node) // fallback to base behavior
        },
        VisitToken: func(node *ast.Node, v *ast.NodeVisitor) *ast.Node {
            if node != nil {
                // handle a token node
            }
            return v.VisitNode(node) // fallback to base behavior
        },
        VisitNodes: func(nodes *ast.NodeList, v *ast.NodeVisitor) *ast.NodeList {
            if nodes != nil {
                // handle a NodeList
            }
            return v.VisitNodes(nodes) // fallbakc to base behavior
        },
        VisitModifiers: func(nodes *ast.ModifierList, v *ast.NodeVisitor) *ast.ModifierList {
            if nodes != nil {
                // handle a ModifierList
            }
            return v.VisitModifiers(nodes) // fallback to base behavior
        },
    }
}
v.Visit = func(node *ast.Node) *ast.Node {
    // for a shallow visitor, simply return `node`
    //return v;
    
    // for a recursive visitor, return `v.VisitEachChild(node)`
    return v.VisitEachChild(node)
}

// kick off the visitor:
result := v.VisitNode(node) // NOTE: bypasses hook for `node`
// or
result := v.VisitEachChild(node) // NOTE: bypasses `node` but visits children
```

Please note that there are four callback hooks in the Go implementation rather than the three in JS due to the distinction between `NodeList` and `ModifierList`.

### DIFF
--- a/internal/ast/visitor.go
+++ b/internal/ast/visitor.go
@@ -9,16 +9,30 @@ import (
 type NodeVisitor struct {
 	Visit   func(node *Node) *Node // The callback used to visit a node
 	Factory NodeFactory            // A NodeFactory used to produce new nodes when passed to VisitEachChild
+	Hooks   NodeVisitorHooks       // Hooks to be invoked when visiting a node
+}
 
-	// !!! Callbacks for LanguageService purposes (override VisitNodes, VisitModifiers, VisitToken)
+// These hooks are used to intercept the default behavior of the visitor in certain situations such as
+// in tests or in the language service.
+type NodeVisitorHooks struct {
+	VisitNode      func(node *Node, v *NodeVisitor) *Node                  // Overrides visiting a Node. Only invoked by the Update* methods on NodeFactory.
+	VisitToken     func(node *TokenNode, v *NodeVisitor) *Node             // Overrides visiting a TokenNode. Only invoked by the Update* methods on NodeFactory.
+	VisitNodes     func(nodes *NodeList, v *NodeVisitor) *NodeList         // Overrides visiting a NodeList. Only invoked by the Update* methods on NodeFactory.
+	VisitModifiers func(nodes *ModifierList, v *NodeVisitor) *ModifierList // Overrides visiting a ModifierList. Only invoked by the Update* methods on NodeFactory.
+	SetOriginal    func(node *Node, original *Node)                        // Overrides setting the original node. Only invoked by VisitEachChild on NodeVisitor.
 }
 
 func (v *NodeVisitor) VisitSourceFile(node *SourceFile) *SourceFile {
 	return v.VisitNode(node.AsNode()).AsSourceFile()
 }
 
+// Visits a Node, possibly returning a new Node in its place.
+//
+//   - If the input node is nil, then the output is nil.
+//   - If v.Visit is nil, then the output is the input.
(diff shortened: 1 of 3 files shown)


internal/ast/visitor.go
```diff
@@ -124,17 +147,47 @@ func (v *NodeVisitor) visitSlice(nodes []*Node) (result []*Node, changed bool) {
...
-	if updated != original {
-		updated.Loc = original.Loc
+// Invokes v.Hooks.VisitNode if provided, otherwise invokes v.VisitNode.
+func (v *NodeVisitor) visitNode(node *Node) *Node {
+	if v.Hooks.VisitNode != nil {
+		return v.Hooks.VisitNode(node, v)
 	}
-	return updated
+	return v.VisitNode(node)
+}
```
jakebailey, line 169: Hopefully this doesn't negatively impact perf (due to loss of inlining) but we sort of have to do it...

internal/transformers/typeeraser.go
```diff
@@ -144,35 +144,35 @@ func (v *TypeEraserTransformer) visit(node *ast.Node) *ast.Node {
...
+		return v.Factory.UpdateCallExpression(n, v.VisitNode(n.Expression), n.QuestionDotToken, nil, v.VisitNodes(n.Arguments))
 
 	case ast.KindNewExpression:
 		n := node.AsNewExpression()
 		return v.Factory.UpdateNewExpression(n, v.VisitNode(n.Expression), nil, v.VisitNodes(n.Arguments))
 
 	case ast.KindTaggedTemplateExpression:
 		n := node.AsTaggedTemplateExpression()
-		return v.Factory.UpdateTaggedTemplateExpression(n, v.VisitNode(n.Tag), v.VisitToken(n.QuestionDotToken), nil, v.VisitNode(n.Template))
+		return v.Factory.UpdateTaggedTemplateExpression(n, v.VisitNode(n.Tag), n.QuestionDotToken, nil, v.VisitNode(n.Template))
```
jakebailey, line 175: Where these VisitToken calls just redundant?

rbuckton: For these cases, yes. There is no longer a public `VisitToken` method, as it was only intended to invoke the hook and return the result. This matches the behavior in `visitEachChild`, which didn't visit tokens unless a `tokenVisitor` callback was provided.

PR #273, opened 2025-02-04
Fix test name generation

gabritto:
Don't include option with a single variation in the name of the test

### DIFF
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -516,11 +516,11 @@ func GetFileBasedTestConfigurations(t *testing.T, settings map[string]string, va
 					t.Fatal("Provided test options exceeded the maximum number of variations")
 				}
 				optionEntries = append(optionEntries, append([]string{option}, entries...))
+				continue
 			}
-		} else {
-			// Variation is not supported or not needed for the option
-			nonVariyingOptions[option] = value
 		}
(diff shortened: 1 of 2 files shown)


internal/testutil/harnessutil/harnessutil.go
```diff
@@ -510,15 +510,15 @@ func GetFileBasedTestConfigurations(t *testing.T, settings map[string]string, va
 	for option, value := range settings {
 		if _, ok := varyByOptions[option]; ok {
 			entries := splitOptionValues(t, value, option)
-			if len(entries) > 0 {
+			if len(entries) > 1 {
```
jakebailey, line 513: The original code had `0`; is this different in behavior somehow?

iisaduan: If there's only one option, then we don't need the name to reflect which option it is

gabritto: This code path now also generates the name of the test. The original one had a separate scheme, so it was different. The original code sort of does this filtering inside `splitVaryBySettingValue` instead:
```ts
// do nothing if the setting has no variations
  if (includes.length <= 1 && !star && excludes.length === 0) {
      return undefined;
  }
```

jakebailey: Yeah, makes sense now

PR #275, opened 2025-02-04
Compiler tests diff runner

gabritto:


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 )
 
 require (
+	github.com/pkg/diff v0.0.0-20241224192749-4e6772a4315c
```
jakebailey: You're probably going to want to revert and then re-run `go mod tidy` to get this in the right block.

I'm curious what put it here; did gopls do it?

internal/testutil/tsbaseline/util.go
```diff
@@ -22,7 +22,8 @@ var (
 )
 
 func removeTestPathPrefixes(text string, retainTrailingDirectorySeparator bool) string {
-	testPathPrefix.ReplaceAllStringFunc(text, func(scheme string) string {
+	return testPathPrefix.ReplaceAllStringFunc(text, func(match string) string {
```
gabritto: Drive-by bug fix.

internal/testutil/runner/compiler_runner.go
```diff
@@ -156,16 +201,6 @@ func getCompilerFileBasedTest(t *testing.T, filename string) *compilerFileBasedT
 	}
 }
 
-var localBasePath = filepath.Join(repo.TestDataPath, "baselines", "local")
```
gabritto, line 159: Moved to `cleanUpLocal` method above.

internal/testutil/runner/compiler_runner.go
```diff
@@ -77,12 +84,38 @@ func (r *CompilerBaselineRunner) EnumerateTestFiles() []string {
...
+	crashingTests := []string{
+		"typeGuardNarrowsIndexedAccessOfKnownProperty10.ts",
+	}
 	for _, filename := range files {
+		if slices.Contains(crashingTests, tspath.GetBaseFileName(filename)) {
+			t.Skip(fmt.Sprintf("Skipping crashing test %s", filename))
+		}
+		extension := tspath.GetAnyExtensionFromPath(filename, nil, true)
+		if extension == tspath.ExtensionTsx {
+			t.Skip(fmt.Sprintf("Skipping jsx test %s", filename))
```
gabritto: We currently crash on many (all?) jsx files because most of that isn't implemented yet.

testdata/baselines/reference/diff/compiler/2dArrays.symbols.diff
```diff
@@ -0,0 +1,45 @@
+--- 2dArrays.symbols
++++ 2dArrays.symbols
```
jakebailey: Can you include something in here (a prefix maybe) which indicates which one is strada and which is corsa (though maybe not those terms?)

internal/testutil/tsbaseline/type_symbol_baseline.go
```diff
@@ -52,10 +52,17 @@ func DoTypeAndSymbolBaseline(
 
 	fullWalker := newTypeWriterWalker(program, hasErrorBaseline)
 
-	t.Run("type", func(t *testing.T) {
-		checkBaselines(t, baselinePath, allFiles, fullWalker, header, opts, false /*isSymbolBaseline*/)
-	})
+	if !opts.IsDiff {
+		t.Run("type", func(t *testing.T) {
```
gabritto, line 57: I'll test and enable type baselines for diff in a later PR.

testdata/baselines/reference/diff/compiler/2dArrays.symbols.diff
jakebailey: maybe just `old.symbols`, `new.symbols`, or `before.symbols`, `after.symbols`

internal/testutil/tsbaseline/type_symbol_baseline.go
```diff
@@ -52,10 +52,17 @@ func DoTypeAndSymbolBaseline(
...
+			checkBaselines(t, baselinePath, allFiles, fullWalker, header, opts, false /*isSymbolBaseline*/)
+		})
+	}
 	t.Run("symbol", func(t *testing.T) {
+		defer func() {
+			if r := recover(); r != nil {
+				t.Fatalf("Panic on compiling test for baseline %s:\n%v", header, r)
+			}
+		}()
```
jakebailey: Is this still needed?

gabritto: It is if we want to know what/all tests caused a panic without doing `-v` and `parallel=1` and whatnot, I think.

jakebailey: I think it might be the case that the single threaded change means panics actually get sent to the test that failed, but, I might be wrong, yeah.

jakebailey: No, that doesn't make sense, panics would still come through here, nevermind.

testdata/baselines/reference/diff/compiler/2dArrays.symbols.diff
```diff
@@ -1,5 +1,5 @@
---- 2dArrays.symbols
-+++ 2dArrays.symbols
+--- new.2dArrays.symbols
++++ old.2dArrays.symbols
```
jakebailey: this seems to imply that it's backwards; I'd expect the `+++`/green to be the new corsa code.

gabritto: yeah, I was just wondering about that.

internal/race/norace.go
```diff
@@ -0,0 +1,8 @@
+//go:build !race
+// +build !race
+
+// Package israce reports if the Go race detector is enabled.
+package race
```
jakebailey: Can you just put this into the testutil package?

internal/race/norace.go
```diff
@@ -0,0 +1,8 @@
+//go:build !race
+// +build !race
```
jakebailey: No need for the + style, I'm surprised the formatter didn't remove it

internal/testutil/tsbaseline/type_symbol_baseline.go
```diff
@@ -70,9 +77,9 @@ func checkBaselines(
 	isSymbolBaseline bool,
 ) {
 	fullExtension := core.IfElse(isSymbolBaseline, ".symbols", ".types")
-	outputFileName := tspath.RemoveFileExtension(baselinePath)
+	outputFileName := tsExtension.ReplaceAllString(baselinePath, fullExtension)
```
gabritto, line 78: Bug fix: we should preserve the `.d` from a file name.

gabritto:
Tests will keep failing until I commit the baselines, which I'm planning on doing after people review.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -192,70 +193,70 @@ var harnessCommandLineOptions = []*tsoptions.CommandLineOption{
 	},
 	{
 		Name: "useCaseSensitiveFileNames",
-		Kind: "boolean",
+		Kind: tsoptions.CommandLineOptionTypeBoolean,
```
gabritto, line 199: Copilot generated that in a previous PR and for some reason it didn't use the consts.

internal/tsoptions/commandlineoption.go
```diff
@@ -157,6 +157,11 @@ var commandLineOptionElements = map[string]*CommandLineOption{
 		isFilePath:      true,
 		extraValidation: true,
 	},
+	// Test infra options
```
gabritto, line 160: Does that seem ok? This is needed for a call to `option.Elements()` to work, and `option.Elements()` is called in `ParseListTypeOption(...)`, which I'm using to parse e.g. `@libFiles`.

internal/testutil/testutil.go
```diff
@@ -21,3 +22,10 @@ func AssertPanics(tb testing.TB, fn func(), expected any, msgAndArgs ...interfac
 	assert.Assert(tb, got != nil, msgAndArgs...)
 	assert.Equal(tb, got, expected, msgAndArgs...)
 }
+
+func RecoverAndFail(t *testing.T, msg string) {
```
jakebailey, line 30: Does this actually work? I thought you had to recover directly in the function for this to work

gabritto: It works. Example (with a panic in `checkSourceFile`):
```
--- FAIL: TestCompilerDiffBaselines/conformance/typingsLookupAmd.ts (0.08s)
            testutil.go:29: Panic on compiling test C:/Users/gabrielaa/code/typescript-go/_submodules/TypeScript/tests/cases/conformance/typings/typingsLookupAmd.ts:
                Some panic
                goroutine 12840 [running]:
                runtime/debug.Stack()
                	C:/Users/gabrielaa/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.3.windows-amd64/src/runtime/debug/stack.go:26 +0x5e
                github.com/microsoft/typescript-go/internal/testutil.RecoverAndFail(0xc009b12000, {0xc0090c2630, 0x88})
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/testutil.go:28 +0x58
                panic({0xd6c100?, 0x12de660?})
                	C:/Users/gabrielaa/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.3.windows-amd64/src/runtime/panic.go:785 +0x132
                github.com/microsoft/typescript-go/internal/checker.(*Checker).CheckSourceFile(...)
                	C:/Users/gabrielaa/code/typescript-go/internal/checker/checker.go:1703
                github.com/microsoft/typescript-go/internal/compiler.(*Program).CheckSourceFiles.func1()
                	C:/Users/gabrielaa/code/typescript-go/internal/compiler/program.go:195 +0x3a
                github.com/microsoft/typescript-go/internal/core.(*singleThreadedWorkGroup).RunAndWait(0xc0220463f0)
                	C:/Users/gabrielaa/code/typescript-go/internal/core/workgroup.go:83 +0x4b
                github.com/microsoft/typescript-go/internal/compiler.(*Program).CheckSourceFiles(0xc015ecd200)
                	C:/Users/gabrielaa/code/typescript-go/internal/compiler/program.go:199 +0x12b
                github.com/microsoft/typescript-go/internal/compiler.(*Program).getDiagnosticsHelper(0xc015ecd200, 0x0?, 0xc0?, 0x1?, 0xc0073ad700)
                	C:/Users/gabrielaa/code/typescript-go/internal/compiler/program.go:368 +0x54
                github.com/microsoft/typescript-go/internal/compiler.(*Program).GetSemanticDiagnostics(...)
                	C:/Users/gabrielaa/code/typescript-go/internal/compiler/program.go:267
                github.com/microsoft/typescript-go/internal/testutil/harnessutil.compileFilesWithHost({0x12e8610?, 0xc018d8e480?}, {0xc00010a1c0?, 0x23?, 0xd6c100?}, 0xc027f7b8c8, 0xc00e51abe0)
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/harnessutil/harnessutil.go:418 +0x2a7
                github.com/microsoft/typescript-go/internal/testutil/harnessutil.CompileFiles(0xc009b12000, {0xc009f9c940, 0x1, 0x1f19580?}, {0xc00010a150, 0x2, 0xc00a0b5290?}, 0xc009b0e240, 0xc009aa7aa0, {0xef2ee5, ...}, ...)
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/harnessutil/harnessutil.go:158 +0x71c
                github.com/microsoft/typescript-go/internal/testutil/runner.newCompilerTest(0xc009b12000, {0xc0002b6fc0, 0x70}, 0xc009aa7e90, 0xc0097fab58)
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/runner/compiler_runner.go:289 +0x830
                github.com/microsoft/typescript-go/internal/testutil/runner.(*CompilerBaselineRunner).runSingleConfigTest(0xc000290000, 0xc009b12000, 0xc009926900, 0xc0097fab58)
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/runner/compiler_runner.go:169 +0x16b
                github.com/microsoft/typescript-go/internal/testutil/runner.(*CompilerBaselineRunner).runTest.func1(0xc009b12000?)
                	C:/Users/gabrielaa/code/typescript-go/internal/testutil/runner/compiler_runner.go:157 +0x25
                testing.tRunner(0xc009b12000, 0xc0097cdd20)
                	C:/Users/gabrielaa/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.3.windows-amd64/src/testing/testing.go:1690 +0xcb
                created by testing.(*T).Run in goroutine 19
                	C:/Users/gabrielaa/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.3.windows-amd64/src/testing/testing.go:1743 +0x377
```

internal/tsoptions/commandlineoption.go
jakebailey: I think this will cause us to allow this flag in places we don't expect.

Can we parse `@libFiles` separately? Or export the funcs needed to do so?

gabritto: I don't think it will. This slice is only ever used by the `Elements()` method. It's not used to determine which options are valid in the first place (those are different slices, like the ones in `declscompiler.go`, which are used in e.g. tsconfig parsing). But I can either write my own copy of `ParseListTypeOption` that parses just `@libFiles`, or refactor the existing functions somehow to avoid the access to this elements slices, although I'm not that familiar with the code.

jakebailey: I think it's fine if it's not harmful.

FWIW you can probably just get away with splitting the value on `,` and then trimming the elements...

iisaduan: This looks good to me

jakebailey:
Somewhat concerning race failure, which implies that the multi-threaded mode (which we enable in race mode) has some non-determinism that is breaking things.

That and the race mode already takes 48 minutes???? Clearly we need to run that on a souped up machine or something.

jakebailey:
Okay, this is actually pretty interesting; the failure is happening because the test runner is not using the checker per source file. Switching to use the right checker for each source file fixes the problem.

The only difference between them would be whether or not the given checker had checked their source files. I went back into the old compiler, and then disabled everything but the symbols tests, including getting any errors or anything (which is done eagerly), and go the same exact diffs!

This to me indicates that we can't actually do types/symbol baselines properly without fully checking a file, which implies that there's actually some underlying bug between both compilers. It all seems to have to do with `this`.

I'm going to update this PR to:

- Use the correct checker, the one that diagnostics were produced from
- Add a new non-race test which enables multi-threading (since race is too slow)
- Maybe disable race mode for these tests for now, or get a pool setup with some giant machines.

jakebailey:
For reference, here are the previously failing tests and baseline differences. https://gist.github.com/jakebailey/1d6590403a0e58bbfcc635ea06dfe748

We should probably investigate this at some point given it affects other users, but, this same problem is present in the old compiler too, so...

jakebailey:
Wow, okay, so I implemented an extremely rudimentary (and wrong for various reasons) AST cache for testing, and this brought the runtime of the current tests down to 8 seconds.

```
✖  internal/testutil/runner (7.918s)

=== Failed
=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/conformance/exportAsNamespace3.ts_module=commonjs/symbol (0.00s)
    baseline.go:119: the baseline file exportAsNamespace3(module=commonjs).symbols.diff has changed. (Run `hereby baseline-accept` if the new baseline is correct.)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/conformance/exportAsNamespace3.ts_module=commonjs (0.01s)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/conformance (0.63s)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/compiler/importHelpersWithImportOrExportDefault.ts_esmoduleinterop=true,module=system/symbol (0.00s)
    baseline.go:119: the baseline file importHelpersWithImportOrExportDefault(esmoduleinterop=true,module=system).symbols.diff has changed. (Run `hereby baseline-accept` if the new baseline is correct.)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/compiler/importHelpersWithImportOrExportDefault.ts_esmoduleinterop=true,module=system (0.01s)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule/compiler (0.65s)

=== FAIL: internal/testutil/runner TestCompilerBaselinesSubmodule (0.00s)

DONE 25854 tests, 7 failures in 8.659s
```

I think most of the time is actually really spent parsing/binding files that should actually be shared (and were technically shared in the old codebase, ish).

The above failure is a case where the cache key I'm using just isn't good enough.

jakebailey:
Unfortunately there's a bunch of state we store on the SourceFiles that make this not work, `Path` is set after the fact (easy to fix and integrate into a cache key), but `Imports` and so on are not.

That and `collectModuleReferences` modifies the AST nodes by setting their parent pointers, which is actually a bug elsewhere too we need to fix.

But, I think that we can actually still do it in that 8 seconds if we fix all of those.

jakebailey:
Comparing CI test time of the runner package before/after AST caching:

- Regular run: 6m19.137s -> 52.906s
- With concurrent programs: 6m30.342s -> 1m8.742s
- In race mode: 2670.664s -> 391.295s

Roughly 6-7x faster.

I still want to get our CI running on better builders for race mode, though. Race mode has about a 5x cost, it seems (race mode does concurrent programs too, of course, so the last two numbers can be directly compared).

jakebailey:
Switching to a new pool greatly improved the performance. I mistakenly used an 8 core machine instead of a 16 core, and yet it ran race mode in 86 seconds, which is already like 4x faster even though it's only twice the core count of the default GHA runner. I'll rerun with the 16 core I intended later.

jakebailey:
16 core machine nets 76.860s, not much of an improvement, funnily.

jakebailey:
Unfortunately after #373, this PR is now crashing on a few tests.

jakebailey:
There's one failing test: https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/externalModules/typeOnly/chained2.ts

Seemingly in concurrent programs, not all checkers have a symbol at the location for `A` or `B` in `d.ts`.

gabritto:
Submodule test `mappedTypeRecursiveInference.ts` relies on the no truncation truncation limit of the type printer (`noTruncationMaximumTruncationLength` in Strada), otherwise it attempts to print forever. Since we don't have the actual type printer yet, I'm disabling it for now since that's easiest.

internal/testutil/runner/compiler_runner.go
```diff
@@ -89,7 +89,8 @@ func (r *CompilerBaselineRunner) RunTests(t *testing.T) {
 	r.cleanUpLocal(t)
 	files := r.EnumerateTestFiles()
 	skippedTests := []string{
-		"mappedTypeRecursiveInference.ts",
+		"chained2.ts",                     // Corsa bug related to multi/single-threaded mode.
```
jakebailey, line 92: Probably need to delete the baselines for these files; I don't think we have the check in place that looks for unused ones yet.

That or maybe we skip at least chained2 only outside SingleThreaded mode?

PR #276, opened 2025-02-04
add watch options

iisaduan:
Adds `WatchOptions` to `core`, `commandLineParser`, and `tsconfigParsing`, and cleans up how the `optionsNameMap`s are implemented/exported. 
- changed `NameMap` to use an OrderedMap in order to make retrieving keys/value consistent for future uses
- bugfix/refactor to handle list options more consistently in `commandLineParser`

### DIFF
--- a/internal/tsoptions/declsbuild.go
+++ b/internal/tsoptions/declsbuild.go
@@ -1,8 +1,12 @@
 package tsoptions
 
-import "github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+import (
+	"slices"
 
-var BuildOpts = append(commonOptionsWithBuild, optionsForBuild...)
+	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
+)
+
+var BuildOpts = slices.Concat(commonOptionsWithBuild, optionsForBuild)
 
 var tscBuildOption = CommandLineOption{
 	Name:                     "build",
--- a/internal/tsoptions/declscompiler.go
+++ b/internal/tsoptions/declscompiler.go
@@ -1,11 +1,13 @@
 package tsoptions
 
 import (
+	"slices"
+
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
-var OptionsDeclarations = append(commonOptionsWithBuild, optionsForCompiler...)
+var OptionsDeclarations = slices.Concat(commonOptionsWithBuild, optionsForCompiler)
 
 var optionsForCompiler = []*CommandLineOption{
 	//******* commandOptionsWithoutBuild *******
--- a/internal/tsoptions/enummaps.go
+++ b/internal/tsoptions/enummaps.go
@@ -138,6 +138,7 @@ var targetOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[s
 })
 
 var moduleOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[string, any]{
+	{Key: "none", Value: core.ModuleKindNone},
 	{Key: "commonjs", Value: core.ModuleKindCommonJS},
 	{Key: "amd", Value: core.ModuleKindAMD},
 	{Key: "system", Value: core.ModuleKindSystem},
--- a/internal/tsoptions/commandlineparser_test.go
+++ b/internal/tsoptions/commandlineparser_test.go
@@ -3,6 +3,7 @@ package tsoptions_test
 import (
 	"encoding/json"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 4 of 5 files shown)


internal/core/watchoptions.go
```diff
@@ -0,0 +1,42 @@
...
+type WatchFileKind int32
+
+const (
+	WatchFileKind_None                         WatchFileKind = 0
+	WatchFileKind_FixedPollingInterval         WatchFileKind = 1
+	WatchFileKind_PriorityPollingInterval      WatchFileKind = 2
+	WatchFileKind_DynamicPriorityPolling       WatchFileKind = 3
+	WatchFileKind_FixedChunkSizePolling        WatchFileKind = 4
+	WatchFileKind_UseFsEvents                  WatchFileKind = 5
+	WatchFileKind_UseFsEventsOnParentDirectory WatchFileKind = 6
```
jakebailey: This naming scheme doesn't match our other enums; should not have the `_`.

internal/core/watchoptions.go
```diff
@@ -0,0 +1,42 @@
+package core
+
+type WatchOptions struct {
+	FileKind        WatchFileKind      `json:"watchFile"`
+	DirectoryKind   WatchDirectoryKind `json:"watchDirectory"`
+	FallbackPolling PollingKind        `json:"fallbackPolling"`
+	SyncWatchDir    Tristate           `json:"synchronousWatchDirectory"`
```
jakebailey, line 7: Honestly I'm a bit concerned about porting any of these at all; the watch options here are very much guided by the watching strategy we have in Node, and almost certainly won't actually be used in Go (in favor of more options).

Porting them is probably fine as we can simply ignore them, but I do wonder if we can just accept any values for these and then ignore them entirely.

internal/tsoptions/declsbuild.go
```diff
@@ -2,54 +2,56 @@ package tsoptions
 
 import "github.com/microsoft/typescript-go/internal/compiler/diagnostics"
 
+var BuildOpts = append(commonOptionsWithBuild, optionsForBuild...)
+
 var tscBuildOption = CommandLineOption{
 	Name:                     "build",
 	Kind:                     "boolean",
 	shortName:                "b",
 	showInSimplifiedHelpView: true,
```
jakebailey, line 15: While you're here, I'm sort of thinking all of this should be exported too, right? Surely we'll need these flags for help messages and so on.

internal/tsoptions/errors.go
```diff
@@ -36,7 +36,7 @@ func getCompilerOptionValueTypeString(option *CommandLineOption) string {
 	}
 }
 
-func (parser *CommandLineParser) createUnknownOptionError(
+func (parser *commandLineParser) createUnknownOptionError(
```
jakebailey, line 39: Unexported 🎉

testdata/baselines/reference/tsc/commandLine/help-all.js
```diff
@@ -133,3 +133,5 @@ CompilerOptions::{
 Output::
 error TS6266: Option 'help' can only be specified on command line.
 
+error TS6266: Option 'help' can only be specified on command line.
```
jakebailey: Duplicated?

internal/tsoptions/declsbuild.go
iisaduan: I was thinking to just use an `.Entries` method when we get to it. Do you think it matters?

internal/tsoptions/errors.go
iisaduan: this was on purpose, because it is an internal structure (the parser is only used by the worker)

jakebailey: Yes, I'm celebrating 😄

internal/tsoptions/declsbuild.go
jakebailey: Not until we need it; it's probably okay.

testdata/baselines/reference/tsc/commandLine/help-all.js
iisaduan: This is a bug in tsconfig parsing, related to parsing `help` specifically. I'll see if I can fix it

internal/core/watchoptions.go
iisaduan: I'll keep this in mind as I port/implement watch here, and see what I can get rid of/change

internal/core/watchoptions.go
iisaduan: oof I got frustrated because they got hard to read😅

internal/core/watchoptions.go
jakebailey: Yeah, porting the parsing I think is fine, we can't error on people's configs with these options, but I doubt we'll use any of them. I'm not even sure how we're going to implement file watching at all since there is no Go API for that.

internal/core/watchoptions.go
jakebailey: We sort of decided long ago not to have them, but it's a decision I am not sure we'll keep long term...

internal/tsoptions/declscompiler.go
```diff
@@ -5,117 +5,119 @@ import (
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
-var OptionsDeclarations = []*CommandLineOption{
+var OptionsDeclarations = append(commonOptionsWithBuild, optionsForCompiler...)
```
jakebailey: This is spooky; append can modify its input slice. Do you just want `slices.Concat`?

internal/tsoptions/enummaps.go
```diff
@@ -138,7 +138,6 @@ var targetOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[s
 })
 
 var moduleOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[string, any]{
-	{Key: "none", Value: core.ModuleKindNone},
```
jakebailey: Even if we don't support it, shouldn't we parse it?

iisaduan: for consistency it makes sense

iisaduan: (to parse it, i'll re-add)

PR #277, opened 2025-02-04
More type checking

ahejlsberg:
In this PR:

* Optimizations based on memory allocation profiling.
* Complete port of class, interface, and type parameter declaration checking.
* Port `getNarrowedTypeOfSymbol`.
* Remove iterator logic from `getUnmatchedProperties`.
* Compute variance information without expensive `outofbandVarianceMarkerHandler` callback logic.
* Port various other checks.
* Port type relation error elaboration.
* Fix formatting of pretty-printed diagnostics.

### DIFF
--- a/testdata/baselines/reference/compiler/allowSyntheticDefaultImports9.symbols
+++ b/testdata/baselines/reference/compiler/allowSyntheticDefaultImports9.symbols
@@ -9,7 +9,7 @@ export function bar();
 
 === /a.ts ===
 import { default as Foo } from "./b";
->default : Symbol("/b.d.ts", Decl(b.d.ts, 0, 0))
+>default : Symbol(""/b.d.ts"", Decl(b.d.ts, 0, 0))
 >Foo : Symbol(Foo, Decl(a.ts, 0, 8))
 
 Foo.bar();


internal/checker/printer.go
```diff
@@ -32,16 +32,19 @@ func (c *Checker) SymbolToString(s *ast.Symbol) string {
...
 	if s.ValueDeclaration != nil {
 		name := ast.GetNameOfDeclaration(s.ValueDeclaration)
 		if name != nil {
 			return scanner.GetTextOfNode(name)
 		}
 	}
-	if len(s.Name) >= 1 && s.Name[0] != '\xFE' {
-		return s.Name
+	if len(s.Name) == 0 || s.Name[0] != '\xFE' {
+		return "\"" + s.Name + "\"" // !!! Implement escaping
```
jakebailey, line 45: This seems to have affected the baselines.

ahejlsberg: Yeah, just accepted the new baselines. In the particular test, we have a symbol named `"/b.d.ts"` where the double-quotes are actually part of the name. Such names now get double-quotes around them, hence the double-double-quotes!

PR #278, opened 2025-02-05
Generate LSP types from metaModel.json

jakebailey:
Rather than hand writing the types, generate them from https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/metaModel.json.

gopls has a version of this which probably works better, but is very much integrated into their codebase in ways that I don't want to think about right now, so I just did this all from scratch. But, they have a lot more experience getting things like `nil`/`null`/etc correct.

Unions are represented as structs of pointers. Unmarshalling into them just attempts to unmarshal each in order and chooses the first to succeed (which is certainly not fast). But there are totally unions in LSP which overlap like:

```ts
interface Location {
	uri: DocumentUri;
	range: Range;
}

// ...

	/**
	 * The location of this symbol. Whether a server is allowed to
	 * return a location without a range depends on the client
	 * capability `workspace.symbol.resolveSupport`.
	 *
	 * See also `SymbolInformation.location`.
	 */
	location: Location | { uri: DocumentUri };
```

Yay. So I'm hoping the spec orders them enough.

### DIFF
--- a/internal/lsp/lsproto/_generate/generate.mjs
+++ b/internal/lsp/lsproto/_generate/generate.mjs
@@ -104,6 +104,12 @@ function titleCase(s) {
  */
 void 0;
 
+/**
+ * Map from literal value to type name.
+ * @type {Map<string | number | boolean, string>}
+ */
+const literalTypes = new Map();
+
 /** @type {Map<string, UnionMember[]>} */
 const unionTypes = new Map();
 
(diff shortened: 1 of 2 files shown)


jakebailey:
There are still some bits I'd like to do here, notably:

- Eliminate type declarations unreachable from requests/notifications (there are confusingly a lot of them?)
- Come up with better names for long unions (though some are just unused)
- Maybe drop the union struct idea for marker interfaces (may not work out since some types are not named types)

But theoretically it's fine as it is right now?

internal/lsp/server.go
```diff
@@ -140,16 +140,24 @@ func (s *Server) sendResponse(resp *lsproto.ResponseMessage) error {
...
+	return &v
+}
+
 func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
 	s.initializeParams = req.Params.(*lsproto.InitializeParams)
 	return s.sendResult(req.ID, &lsproto.InitializeResult{
 		ServerInfo: &lsproto.ServerInfo{
 			Name:    "typescript-go",
-			Version: core.Version,
+			Version: ptrTo(core.Version),
```
andrewbranch, line 152: What is this for and why not just `&core.Version`?

jakebailey: `core.Version` is a `const` so can't have its address taken.

jakebailey:
The other thing I'm considering here is _not_ making this a `go:generate` line, and then also not committing the JSON files, which would make this PR a bit smaller.

PR #279, opened 2025-02-05
Return unknownSymbol from resolveEntityName

sandersn:
The refactor that created resolveQualifiedName broke the immediate return of unknownSymbol when the left side of a qualified name doesn't resolve. Now resolveEntityName has to check for it and return it.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -12040,7 +12040,7 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 	default:
 		panic("Unknown entity name kind")
 	}
-	if symbol != nil {
+	if symbol != nil && symbol != c.unknownSymbol {
 		if !ast.NodeIsSynthesized(name) && ast.IsEntityName(name) && (symbol.Flags&ast.SymbolFlagsAlias != 0 || name.Parent.Kind == ast.KindExportAssignment) {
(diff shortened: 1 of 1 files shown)


internal/checker/checker.go
```diff
@@ -12040,7 +12040,7 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 	default:
 		panic("Unknown entity name kind")
 	}
-	if symbol != nil {
+	if symbol != nil && symbol != c.unknownSymbol {
```
jakebailey, line 12043: The original code was:

```ts
        if (!nodeIsSynthesized(name) && isEntityName(name) && (symbol.flags & SymbolFlags.Alias || name.parent.kind === SyntaxKind.ExportAssignment)) {
            markSymbolOfAliasDeclarationIfTypeOnly(getAliasDeclarationFromName(name), symbol, /*finalTarget*/ undefined, /*overwriteEmpty*/ true);
        }
```

Should the fix be elsewhere?

sandersn: The original code was in the QualifiedName branch inline in resolveEntityName:

```ts
            else if (namespace === unknownSymbol) {
                return namespace;
            }
```

That code is now in resolveQualifiedName:

```go
	if namespace == c.unknownSymbol {
		return namespace
	}
```

For exact compatibility the QualifiedName and PropertyAccessExp branches could copy that code, but I think it's correct to run the check on the result of resolveName on Identifiers as well; we should never try to resolve unknownSymbol as an alias.

PR #246, opened 2025-01-22
add back node to moduleresolution (for now)

iisaduan:
Allows `node` to parse as an alias for `bundler`. https://github.com/microsoft/typescript-go/pull/207#issuecomment-2606145871-- allowing this makes it easier for us to test the port.

PR #282, opened 2025-02-05
Ignore `moduleResolution` options `classic` and `node10`

gabritto:
Those are deprecated and were not added to Corsa. Ignoring them for now to avoid crashing, at some point we should either remove the usage of those options in the tests, or remove the test entirely if they really mean to test those options.

### DIFF
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -314,9 +314,6 @@ func parseHarnessOption(t *testing.T, key string, value any, options *HarnessOpt
 }
 
 func getOptionValue(t *testing.T, option *tsoptions.CommandLineOption, value string) tsoptions.CompilerOptionsValue {
-	if option.Name == "moduleResolution" && (value == "classic" || value == "node10") { // Deprecated, ignore it for now to avoid errors
-		return core.ModuleResolutionKindUnknown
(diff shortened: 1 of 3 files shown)


internal/testutil/harnessutil/harnessutil.go
```diff
@@ -314,6 +314,9 @@ func parseHarnessOption(t *testing.T, key string, value any, options *HarnessOpt
 }
 
 func getOptionValue(t *testing.T, option *tsoptions.CommandLineOption, value string) tsoptions.CompilerOptionsValue {
+	if option.Name == "moduleResolution" && (value == "classic" || value == "node10") { // Deprecated, ignore it for now to avoid errors
```
jakebailey: I had thought we configured these elsewhere to effectively turn into `bundler`; is that not the case?

gabritto: I can't find we doing it in Corsa. It's possible we'll do that later as part of the deprecation?

jakebailey: I was thinking of https://github.com/microsoft/typescript-go/pull/246.

gabritto: Ah, ok, I'll do that then.

PR #283, opened 2025-02-05
Lint rule to forbid shadowing

sandersn:
I think it's pretty slow but I haven't compared a before/after run because I think I need to make my edits outside vscode, which automatically runs the linter.

I also made all the changes needed to satisfy the rule. The rule is definitely annoying, especially in the context of code we've written where we commonly redeclare instead of assigning. But I think it'll be worth it to prevent silly errors. Even if it's not, I'd like to have the rule around so I can turn it on temporarily when sweeping for bugs.

- I didn't make an exhaustive effort to push and pop scopes. I just did enough to satisfy our current code base. There are probably other nodes that introduce a scope.
- I exempted `err` `ok` `ch` `size` -- the latter two show up a fair bit in the scanner, although maybe it would be better to change the scanner.

### DIFF
--- a/_tools/customlint/shadow.go
+++ b/_tools/customlint/shadow.go
@@ -21,7 +21,6 @@ var shadow = &analysis.Analyzer{
 
 func runShadow(pass *analysis.Pass) (interface{}, error) {
 	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
-
 	nodeFilter := []ast.Node{
 		(*ast.File)(nil),
 		(*ast.FuncDecl)(nil),
@@ -32,11 +31,6 @@ func runShadow(pass *analysis.Pass) (interface{}, error) {
 		(*ast.ForStmt)(nil),
 		(*ast.BlockStmt)(nil),
 	}
-
-	// The inspect package doesn't tell us up front which file is being used,
-	// so keep track of it as part of the traversal. The file is the first node
-	// so will be set before any other nodes are visited.
-	var file *ast.File
 	scopes := []map[string]bool{}
 
 	inspect.Nodes(nodeFilter, func(n ast.Node, push bool) bool {
@@ -48,14 +42,11 @@ func runShadow(pass *analysis.Pass) (interface{}, error) {
 			return false
 		}
 		switch n := n.(type) {
-		case *ast.File:
-			file = n
-			scopes = append(scopes, make(map[string]bool))
-		case *ast.FuncDecl, *ast.CaseClause, *ast.FuncLit, *ast.BlockStmt, *ast.IfStmt, *ast.ForStmt:
+		case *ast.File, *ast.FuncDecl, *ast.CaseClause, *ast.FuncLit, *ast.BlockStmt, *ast.IfStmt, *ast.ForStmt:
 			scopes = append(scopes, make(map[string]bool))
 		case *ast.AssignStmt:
 			if n.Tok == token.DEFINE {
-				checkShadowingAssignment(pass, file, n, scopes)
+				checkShadowingAssignment(pass, n, scopes)
 			}
 		}
 		return true
@@ -64,7 +55,7 @@ func runShadow(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
-func checkShadowingAssignment(pass *analysis.Pass, file *ast.File, stmt *ast.AssignStmt, scopes []map[string]bool) {
+func checkShadowingAssignment(pass *analysis.Pass, stmt *ast.AssignStmt, scopes []map[string]bool) {
 	for _, name := range stmt.Lhs {
 		switch name := name.(type) {
 		case *ast.Ident:


internal/tsoptions/commandlineparser_test.go
```diff
@@ -199,10 +199,10 @@ func (f *commandLineSubScenario) getBaselineName() (baseline.Options, string) {
...
 
 func parseExistingCompilerBaseline(t *testing.T, baseline string) *TestCommandLineParser {
-	_, rest, _ := strings.Cut(baseline, "CompilerOptions::\n")
-	compilerOptions, rest, watchFound := strings.Cut(rest, "\nWatchOptions::\n")
-	watchOptions, rest, _ := strings.Cut(rest, "\nFileNames::\n")
-	fileNames, errors, _ := strings.Cut(rest, "\nErrors::\n")
+	_, rest1, _ := strings.Cut(baseline, "CompilerOptions::\n")
+	compilerOptions, rest2, watchFound := strings.Cut(rest1, "\nWatchOptions::\n")
+	watchOptions, rest3, _ := strings.Cut(rest2, "\nFileNames::\n")
+	fileNames, errors, _ := strings.Cut(rest3, "\nErrors::\n")
```
jakebailey, line 205: This sort of thing is just unfortunate in the case where the compiler actually has to allocate stack space for these, but I think a recent change to the compiler actually started reusing stack space...

internal/compiler/module/resolver_test.go
```diff
@@ -312,8 +312,8 @@ func runTraceBaseline(t *testing.T, test traceTestCase) {
 		}
 
 		t.Run("concurrent", func(t *testing.T) {
-			host := newVFSModuleResolutionHost(test.files, test.currentDirectory)
-			resolver := module.NewResolver(host, test.compilerOptions)
+			host = newVFSModuleResolutionHost(test.files, test.currentDirectory)
+			resolver = module.NewResolver(host, test.compilerOptions)
```
jakebailey: This is definitely causing a new bug; now this will cause races with the code below.

internal/tsoptions/commandlineparser_test.go
sandersn: I would expect any compiler to treat a sequence of redeclarations with the same name vs different names basically the same. It's going to end up with the same thing after dataflow analysis, right?

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,92 @@
...
+var shadow = &analysis.Analyzer{
+	Name: "shadow",
+	Doc:  "bans shadowing",
+	Run:  runNonEmptyCase,
+	Requires: []*analysis.Analyzer{
+		inspect.Analyzer,
+	},
+}
+
+func runNonEmptyCase(pass *analysis.Pass) (interface{}, error) {
```
jakebailey: Did you mean to call this `runShadow`?

internal/checker/checker.go
```diff
@@ -12040,7 +12040,7 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 	default:
 		panic("Unknown entity name kind")
 	}
-	if symbol != nil {
+	if symbol != nil && symbol != c.unknownSymbol {
```
jakebailey: Was your main old for this change to not be on the branch? (Just noticing the diff here unrelated to shadowing)

internal/compiler/module/resolver_test.go
```diff
@@ -312,15 +312,15 @@ func runTraceBaseline(t *testing.T, test traceTestCase) {
 		}
 
 		t.Run("concurrent", func(t *testing.T) {
-			host := newVFSModuleResolutionHost(test.files, test.currentDirectory)
-			resolver := module.NewResolver(host, test.compilerOptions)
+			concurrenthost := newVFSModuleResolutionHost(test.files, test.currentDirectory)
+			concurrentresolver := module.NewResolver(concurrenthost, test.compilerOptions)
```
jakebailey: Nit, but the naming here should be camelCase.

internal/tsoptions/commandlineparser_test.go
jakebailey: Yeah, it should be fine. https://godbolt.org/z/qa9f57698 shows both having the same stack size (40).

jakebailey:
I feel like we can do this, but I would like to see a version of this which allows shadowing as long as all paths from the shadowed var lead to a return such that the external variable can never be used.

_tools/customlint/shadow.go
sandersn: Yep!

internal/checker/checker.go
sandersn: Yep!

sandersn:
This version with a simplified binder only needs a little over 40 changes for our code base. Would a smarter rule really save us that much? I don't think so.

The best simplification of control flow I can think of is: 
1. track all shadowing errors by name but don't issue them till the end.
2. check all identifiers: if they're in the shadowed map and refer to the shadowed variable, keep the error. Otherwise discard the error. (But: how do you know when you've seen all references?)

But checking all identifiers sounds much more expensive than checking the scope-creating nodes+assignments.

jakebailey:
Looking at all of the changes in this PR, all of them are cases where the variable being shadowed is never actually used after, and so are false positives.

The one correct lint that this rule found (before it was merged with main where you fixed it earlier) was:

```go
		result := c.unresolvedSymbols[path]
		if result == nil {
			result := c.newSymbolEx(ast.SymbolFlagsTypeAlias, text, ast.CheckFlagsUnresolved)
			c.unresolvedSymbols[path] = result
			result.Parent = parentSymbol
			c.declaredTypeLinks.get(result).declaredType = c.unresolvedType
		}
		return result
```

Which would be caught by a rule that detects "variable shadowed, then same name used after in the CFG".

I'm certainly not saying to reinvent CFA; there are analysis passes available to do this in Go already (https://pkg.go.dev/golang.org/x/tools@v0.29.0/go/analysis/passes/ctrlflow and https://pkg.go.dev/golang.org/x/tools@v0.29.0/go/ssa).

PR #284, opened 2025-02-06
Enum and namespace transformer

rbuckton:
This introduces transformer to handle `enum` and `namespace` emit to JavaScript. This transformer does not currently support `const enum` inlining, nor does it support unqualified `enum` member name resolution in merged enums (i.e., `enum E { A }; enum E { B = A }`), nor unqualified `namespace` export references in merged namespaces (i.e., `namespace N { export var x = 1; } namespace N { x; }`.

In order to reduce the dependence on the type checker, this transformer pushes some of the logic for auto-incrementing `enum` members to the runtime. When possible, the transformer uses a number of heuristics to simplify the `enum` emit.

The following are some examples of the new emit:

### Auto-numbered `enum`
(matches existing emit)
```ts
// input
enum E {
  A,
  B,
  C,
}

// output
var E;
(function (E) {
  E[E["A"] = 0] = "A";
  E[E["B"] = 1] = "B";
  E[E["C"] = 2] = "C";
})(E || (E = {}));
```

### Auto-numbered `enum` with explicit start
(matches existing emit)
```ts
// input
enum E {
  A = 2,
  B,
  C,
}

// output
var E;
(function (E) {
  E[E["A"] = 2] = "A";
  E[E["B"] = 3] = "B";
  E[E["C"] = 4] = "C";
})(E || (E = {}));
```

### String `enum`
(matches existing emit)
```ts
// input
enum E {
  A = "x",
  B = "y",
  C = "z",
}

// output
var E;
(function (E) {
  E["A"] = "x";
  E["B"] = "y";
  E["C"] = "z";
})(E || (E = {}));
```

### Auto-numbered `enum` with self reference
(changed emit)
```ts
// input
enum E {
  A,
  B = A,
  C,
}

// output
var E;
(function (E) {
  var A, auto
  E[E["A"] = A = 0] = "A";
  E[E["B"] = auto = A] = "B";
  E[E["C"] = ++auto] = "C";
})(E || (E = {}));
```

In this example, we introduce a local `A` binding that is used when evalutating `E.B`. From that point on, we use the varaible `auto` to track further auto-incrementing at runtime.

### Reference to external variable
(changed emit)
```ts
// input
enum E {
  A = x,
  B = y,
}

// output
var E;
(function (E) {
  var A, B
  E["A"] = A = x;
  if (typeof A !== "string") E[A] = "A";
  E["B"] = B = y;
  if (typeof B !== "string") E[B] = "B";
})(E || (E = {}));
```

This emulates at runtime what we validate at check time. In the old compiler, if the checker determined that `x` or `y` were `string` types, then the reverse mapping would not be applied.

### Auto-incrementing after external variable reference
(changed emit)

The old compiler would report an error if you continued using auto-numbered enum members after an enum member initializer that did not produce a numeric literal type. In the new emitter we assume that this check succeeded and resume auto-numbering, as an error would signify that the results were unreliable.

```ts
// input
declare var x: any;
enum E {
  A = x,
  B // Error: Enum member must have initializer.
}

// output (old)
var E;
(function (E) {
    E[E["A"] = x] = "A";
    E[E["B"] = void 0] = "B"; // type-directed emit, produces `void 0`
})(E || (E = {}));

// output (new)
var E;
(function (E) {
    var auto;
    E[E["A"] = auto = x] = "A";
    E[E["B"] = ++auto] = "B";
})(E || (E = {}));
```

In the old compiler, the above emit is type-directed as it changes based on the type of `x`:

```ts
// input
declare const x = 1;
enum E {
  A = x,
  B // Error: Enum member must have initializer.
}

// output (old)
var E;
(function (E) {
    E[E["A"] = 1] = "A"; // type-directed emit, inlines 'x'
    E[E["B"] = 2] = "B"; // type-directed emit
})(E || (E = {}));

// output (new)
var E;
(function (E) {
    var auto;
    E[E["A"] = auto = x] = "A";
    E[E["B"] = ++auto] = "B";
})(E || (E = {}));
```

However, in the new emit we produce the same output in both cases, which produces the correct result in the non-error case.


internal/testutil/parseutil/parseutil.go
```diff
@@ -0,0 +1,85 @@
...
+		diagnosticwriter.WriteFormatDiagnostics(&b, file.Diagnostics(), &diagnosticwriter.FormattingOptions{
+			NewLine: "\n",
+		})
+		t.Error(message + b.String())
+	}
+}
+
+var syntheticRecursiveVisitor = sync.OnceValue(func() *ast.NodeVisitor {
+	v := ast.NodeVisitor{
+		Factory: &ast.NodeFactory{},
```
jakebailey: This variable being shared can technically cause resources to live longer than they need to; but maybe this is testing so that won't matter so much?

rbuckton: It's only used for testing, but testing runs in parallel and `core.Pool` doesn't appear to be threadsafe so I've changed this to not be shared.

internal/ast/visitor.go
```diff
@@ -7,8 +7,8 @@ import (
 // NodeVisitor
 
 type NodeVisitor struct {
-	Visit   func(node *Node) *Node // The callback used to visit a node
-	Factory NodeFactory            // A NodeFactory used to produce new nodes when passed to VisitEachChild
+	Visit   func(node *Node) *Node // Required. The callback used to visit a node
+	Factory *NodeFactory           // Required. The NodeFactory used to produce new nodes when passed to VisitEachChild
```
jakebailey, line 11: Just an idea, but if it's common to not provide this (like before since it was defaulted), I'm wondering if that means we should add a method that initializes this for you if nil and then use it below.

rbuckton: Do you mean changing references to use `Factory()` where `Factory()` checks and initializes a `factory` field? That seems like an awful lot of overhead since `Factory` is used heavily. I'd rather it be initialized once at construction time to avoid the overhead.

jakebailey: I mean:

```go
type Visitor struct {
    Factory *NodeFactory
}

func (v *Visitor) factory() *NodeFactory {
    if v.Factory != nil {
		v.Factory = &NodeFactory{}
	}
	return v.Factory
}

// ...

func (v *Visitor) DoSomething() *Node {
	return v.factory().NewIdentifier()
}
```

That func will be inlined, and shouldn't be any more expensive since the generated code would have had a nil check anyway.

jakebailey: It's super annoying to see but https://godbolt.org/z/Y718fee9T I believe shows that the code generated for the two methods are roughly the same (after the nil check anyway; one has the code for allocating the factory inlined).

It's fine of course to leave as-is, but I'd probably think we'd want a `New` func? (But, still not required at all, I'm just brainstorming)

rbuckton: `Factory` needs to be used by other callers, though so this doesn't quite work.

rbuckton: > I believe shows that the code generated for the two methods are roughly the same

They don't seem the same, and the nil check is the cost I'm concerned about. There's no sense performing a nil check repeatedly for something that is necessary to be present for the component to function.

> I'd probably think we'd want a `New` func?

I considered this, but Go makes this somewhat awkward if I want to use a `*NodeVisitor` on its own in some places, but as a base to a composite struct in other places.

rbuckton:
I've finished most of the `namespace` transformation. As it depends heavily on the work in this PR, rather than put up a separate PR I will amend this PR to include the `namespace` transformation as well.

internal/printer/emitflags.go
```diff
@@ -0,0 +1,33 @@
+package printer
+
+type EmitFlags uint32
+
+const (
+	EFSingleLine                EmitFlags = 1 << iota // The contents of this node should be emitted on a single line.
```
jakebailey, line 6: I think we only have one other case where we do this naming scheme, but, I can't really complain 😄

internal/testutil/emitutil/emitutil.go
jakebailey, line 1: A nit, but I think these testutil packages might be better named like `emittest` or similar; I was initially confused about them until I realized later that they were test utilities.

internal/binder/binder.go
```diff
@@ -767,7 +767,7 @@ func (b *Binder) bindModuleDeclaration(node *ast.Node) {
 }
 
 func (b *Binder) declareModuleSymbol(node *ast.Node) ModuleInstanceState {
-	state := getModuleInstanceState(node, nil /*visited*/)
+	state := GetModuleInstanceState(node, nil /*visited*/)
```
jakebailey: https://github.com/microsoft/typescript-go/pull/296/files#diff-9969a067bafd18e3dafe06418204ef65b894c09827a5c98ca142d7cb2e7be4e3R1930 actually moves this all to the AST package, which might be good to pull out into a PR before these two?

internal/ast/ast.go
```diff
@@ -162,7 +167,11 @@ func (n *Node) TemplateLiteralLikeData() *TemplateLiteralLikeBase {
 }
 
 func (n *Node) Symbol() *Symbol {
-	return n.DeclarationData().Symbol
+	data := n.DeclarationData()
```
jakebailey: I feel like it was intended that this crash if the node wasn't supposed to have a symbol; does this fix something?

rbuckton: I'd intended this to address a crash in `getSymbolExportContainer` (runtimesyntax.go) which is a port of `getReferencedExportContainer`, as the change aligned with related changes @ahejlsberg made to `Locals()` and `LocalSymbol()` following a discussion a few weeks back. However, I've just noticed that the crash was due to a different issue, so this change may not be necessary.

PR #285, opened 2025-02-06
Fix FS case sensitivity detection outside Windows

jakebailey:
This got flipped, not sure when (I blame myself).

### DIFF
--- a/internal/vfs/os.go
+++ b/internal/vfs/os.go
@@ -45,11 +45,11 @@ var isFileSystemCaseSensitive = func() bool {
 	swapped := swapCase(exe)
 	if _, err := os.Stat(swapped); err != nil {
 		if os.IsNotExist(err) {
-			return false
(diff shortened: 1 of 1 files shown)


internal/vfs/os_test.go
```diff
@@ -45,8 +45,11 @@ func TestOS(t *testing.T) {
 		// Just check that it works.
 		fs.UseCaseSensitiveFileNames()
 
-		if runtime.GOOS == "windows" {
+		switch runtime.GOOS {
+		case "windows":
 			assert.Assert(t, !fs.UseCaseSensitiveFileNames())
+		case "linux":
+			assert.Assert(t, fs.UseCaseSensitiveFileNames())
```
jakebailey, line 52: Technically this can be wrong, but no more wrong than Windows. I think macos is the only plausible place where either could happen.

PR #286, opened 2025-02-06
Compare Strada baselines with Corsa ones

sandersn:
Very very drafty right now. 
It points the baselines to _submodules/TypeScript and tweaks symbol baseline output to match Strada's. If we prefer the new output, I should put the tweaks behind a flag.

It also currently disables parallelism to (1) make debugging easier (2) avoid panics. I'll remove them when the panics are gone, since debugging is easy enough by altering the test regex in compiler_runner.go.

Some things that need doing:
1. Put Strada tweaks behind a flag.
2. Flatten output to a single directory to match Strada.
3. Reinstate parallelism.

### DIFF
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -514,7 +514,7 @@ func listFiles(path string, spec *regexp.Regexp, recursive bool) ([]string, erro
 }
 
 func listFilesWorker(spec *regexp.Regexp, recursive bool, folder string) ([]string, error) {
-	folder = tspath.GetNormalizedAbsolutePath(folder, repo.TypeScriptSubmodulePath)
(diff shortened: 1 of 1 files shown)


internal/testutil/tsbaseline/util.go
```diff
@@ -22,16 +22,7 @@ var (
 )
 
 func removeTestPathPrefixes(text string, retainTrailingDirectorySeparator bool) string {
-	testPathPrefix.ReplaceAllStringFunc(text, func(scheme string) string {
```
gabritto: Should be fixed here: https://github.com/microsoft/typescript-go/pull/287

sandersn:
This is basically obsolete now; I'm still using it to debug one test at a time/single-threaded, but I'll close the PR since main can now run Strada baselines.

PR #290, opened 2025-02-07
Make compilerOptions merging align with old compiler

rbuckton:
This aligns how we merge `CompilerOptions` from tsconfig.json with options from the command line to match `tsc` (i.e., command line options override tsconfig.json options) by passing the command line compiler options to `ParseJsonSourceFileConfigFileContent` and then overwriting the program options with the result.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -116,7 +116,7 @@ func NewProgram(options ProgramOptions) *Program {
 			tsConfigSourceFile,
 			p.host,
 			p.host.GetCurrentDirectory(),
-			nil,
+			p.compilerOptions,
 			p.configFilePath,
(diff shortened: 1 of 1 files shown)


rbuckton:
Ah, looks like I'm late to this fix, though I'm not sure why we need to merge them again?

jakebailey:
If passing it into the parameter works then that's good too.

rbuckton:
The assignment alone should suffice as the options from the config file will have already been merged. Merging them again is redundant.

jakebailey:
Yeah, I probably didn't realize that when I was plugging this into the original code. Would be good to fix.

jakebailey:
I sent https://github.com/microsoft/typescript-go/pull/292

PR #292, opened 2025-02-07
Pass CLI options into ParseJsonSourceFileConfigFileContent

jakebailey:
Per #290

Cleans up a bit, since we do accept nil options, and parsing always gives something back (so we should set it before error).


internal/compiler/program.go
```diff
@@ -77,7 +77,6 @@ func NewProgram(options ProgramOptions) *Program {
 	if p.compilerOptions == nil {
 		p.compilerOptions = &core.CompilerOptions{}
 	}
```
jakebailey:
```suggestion
	}

```

PR #293, opened 2025-02-07
Address issues found by tests

ahejlsberg:
In this PR:

* Port `isBlockScopedNameDeclaredBeforeUse`. Fixes stack overflow in test.
* Port logic for symbols with deferred types.
* Fix relation key generation.
* Fix issue in `isMatchingReference`. Fixes issue compiling Playwright.
* Account for possible `errorType` as type parameter in inference. Fixes issue compiling Vite.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -598,6 +598,7 @@ type Checker struct {
 	symbolReferenceLinks                      LinkStore[*ast.Symbol, SymbolReferenceLinks]
 	valueSymbolLinks                          LinkStore[*ast.Symbol, ValueSymbolLinks]
 	mappedSymbolLinks                         LinkStore[*ast.Symbol, MappedSymbolLinks]
+	deferredSymbolLinks                       LinkStore[*ast.Symbol, DeferredSymbolLinks]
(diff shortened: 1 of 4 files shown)


internal/checker/utilities.go
```diff
@@ -2178,13 +2178,10 @@ func isVariableLike(node *ast.Node) bool {
...
-	for node != nil {
-		if node.Kind == kind {
-			return node
-		}
+	for node != nil && node.Kind != kind {
 		node = node.Parent
 	}
-	return nil
+	return node
 }
```
jakebailey, line 2185: Isn't this a behavior change? The previous version matches the old compiler, but I think this new version can now return a node of the wrong kind because the loop breaks out when that happens.

jakebailey: Hm, maybe it is equivalent, thinking harder...

Was this just an efficiency change, then?

jakebailey: Yeah, I did check the output assembly for this, and there is a difference (but a positive one); writing this one down as a issue to file on the Go compiler later as a missed optimization...

ahejlsberg:
Latest PR fixes relation key generation and recursion flags propagation. Those two issues were causing excessively long compile times for `immutable.d.ts`, which is used by the Vite project.

PR #296, opened 2025-02-07
Add types baselines

gabritto:
Add types baselines, temporarily using `checker.TypeToString` while we don't have `typeToTypeNode` etc.
Still no underlines in the type baselines, for the same reason.


internal/ast/utilities.go
```diff
@@ -1813,3 +1813,103 @@ func GetNewTargetContainer(node *Node) *Node {
...
+	case KindEnumMember, KindClassDeclaration:
+		return SemanticMeaningValue | SemanticMeaningType
+
+	case KindModuleDeclaration:
+		if IsAmbientModule(node) {
+			return SemanticMeaningNamespace | SemanticMeaningValue
+		} else {
+			return SemanticMeaningNamespace
+		}
+		// !!! Needs binder function
```
gabritto: In Strada, `GetMeaningFromDeclaration` is in services. We don't have services yet, so I put this here, but turns out this function needs `getModuleInstanteState` that lives in the binder, so I'm not exactly sure where I should move put this function.

jakebailey: I think `getModuleInstanteState` can actually be moved into the ast package; it doesn't appear to have any external dependencies?

internal/ast/utilities.go
```diff
@@ -1913,3 +1911,161 @@ func IsJumpStatementTarget(node *Node) bool {
...
+	ModuleInstanceStateConstEnumOnly
+)
+
+func SetParent(child *Node, parent *Node) {
+	if child != nil {
+		child.Parent = parent
+	}
+}
+
+func GetModuleInstanceState(node *Node, visited map[NodeId]ModuleInstanceState) ModuleInstanceState {
```
jakebailey: If you want to go above and beyond, I would suggest making this unexported, then making a new `GetModuleInstanceState` which explicitly passes in nil, since all callers outside don't seem to pass anything in.

internal/ast/utilities.go
```diff
@@ -1813,3 +1813,263 @@ func GetNewTargetContainer(node *Node) *Node {
...
+		return ModuleInstanceStateInstantiated
+	}
+	for p := node.Parent; p != nil; p = p.Parent {
+		if IsBlock(p) || IsModuleBlock(p) || IsSourceFile(p) {
+			statements := GetStatementsOfBlock(p)
+			found := ModuleInstanceStateUnknown
+			for _, statement := range statements.Nodes {
+				if NodeHasName(statement, name) {
+					if statement.Parent == nil {
+						SetParent(statement, p)
```
jakebailey: Eek, just noticed that this function modifies the AST. Is this load-bearing? Do we have to do it here? This may cause races when ASTs are shared between, maybe... But like I would expect that everything is already bound enough to have parents?

gabritto: Ah yes, that may have been why I thought it was weird to move this out of the binder. The commend on `GetModuleInstanceState` above says "getModuleInstanceStateForAliasTarget needs to walk up the parent chain, so parent pointers must be set on this tree already" and proceeds to call `setParent`, so I would guess parents may be missing when this function is called in the binder, but won't be missing when called elsewhere.

gabritto: I couldn't find a better way to do this for now. The Strada callers of `GetModuleInstanteState` that are not the binder should already have parent pointers set (e.g. types baselines, find all references), and so those calls should never actually set parents, only the ones coming from the binder. What I could do is add a boolean for setting the parents or not, and panic if we try to set parents from outside the binder.

rbuckton: Since #284 also needs this, I'm looking into pulling this out into its own PR and just keeping track of parents as we descend.

rbuckton: See #340

internal/checker/utilities.go
```diff
@@ -1094,7 +1094,7 @@ func hasDotDotDotToken(node *ast.Node) bool {
 	return false
 }
 
-func isTypeAny(t *Type) bool {
+func IsTypeAny(t *Type) bool {
```
jakebailey, line 1097: I'm actually surprised this isn't just a method but maybe I shouldn't be this early on 😄

PR #320, opened 2025-02-10
Fixing nil pointer dereference crash

navya9singh:
This pr adds es2024 lib and target options which fix the crash. 
Another reason for the crash was the empty include, exclude and files being added as []string instead of []any which caused a similar crash as https://github.com/microsoft/typescript-go/pull/295.

### DIFF
--- a/internal/tsoptions/enummaps.go
+++ b/internal/tsoptions/enummaps.go
@@ -135,6 +135,7 @@ var targetOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[s
 	{Key: "es2021", Value: core.ScriptTargetES2021},
 	{Key: "es2022", Value: core.ScriptTargetES2022},
 	{Key: "es2023", Value: core.ScriptTargetES2023},
+	// {"es2024", string(core.ScriptTargetES2024},
 	{Key: "esnext", Value: core.ScriptTargetESNext},
 })
 
--- a/internal/tsoptions/tsconfigparsing.go
+++ b/internal/tsoptions/tsconfigparsing.go
@@ -274,7 +274,7 @@ func isCompilerOptionsValue(option *CommandLineOption, value any) bool {
 		}
 		if option.Kind == "enum" && reflect.TypeOf(value).Kind() == reflect.String {
 			_, ok := option.EnumMap().Get(strings.ToLower(value.(string)))
-			return ok || option.DeprecatedKeys().Has(strings.ToLower(value.(string)))
+			return ok || (option.DeprecatedKeys() != nil && option.DeprecatedKeys().Has(strings.ToLower(value.(string))))
 		}
 	}
 	return false


jakebailey:
The version we're porting doesn't have ES2024 libs at all; was that explicitly part of a crash? If it was, then that implies that someone setting ES2025 will also crash and that would be a bug itself.

I don't mind adding it but we would at least need to alias it to ES2024 for all of the places this PR doesn't add it to (e.g. to `enunmap.go`, see everywhere `es2023` is mentioned).

navya9singh:
> The version we're porting doesn't have ES2024 libs at all; was that explicitly part of a crash? If it was, then that implies that someone setting ES2025 will also crash and that would be a bug itself.
> 
> I don't mind adding it but we would at least need to alias it to ES2024 for all of the places this PR doesn't add it to (e.g. to `enunmap.go`, see everywhere `es2023` is mentioned).

isCompilerOptionsValue was the first crash I saw, but just fixing `[]any` issue solved that problem as well. So adding es2024 is not needed anymore.

navya9singh:
> can you uncomment the tests in `execute/tsc_test.go`?

The tests no longer crash, but they fail. It might be faster if you take a look at it after this pr to figure out why that's happening. Does that work?

jakebailey:
Hm, https://github.com/toeverything/AFFiNE/blob/879157b938632b180099c373d3ebb70f1cac4920/tsconfig.eslint.json is still crashing when I pull in this PR.

internal/tsoptions/enummaps.go
```diff
@@ -135,7 +135,6 @@ var targetOptionMap = collections.NewOrderedMapFromList([]collections.MapEntry[s
 	{Key: "es2021", Value: core.ScriptTargetES2021},
 	{Key: "es2022", Value: core.ScriptTargetES2022},
 	{Key: "es2023", Value: core.ScriptTargetES2023},
-	// {"es2024", string(core.ScriptTargetES2024},
```
iisaduan: can you revert this change? (leave it as commented out)

navya9singh: sure

internal/tsoptions/tsconfigparsing.go
```diff
@@ -274,7 +274,7 @@ func isCompilerOptionsValue(option *CommandLineOption, value any) bool {
 		}
 		if option.Kind == "enum" && reflect.TypeOf(value).Kind() == reflect.String {
 			_, ok := option.EnumMap().Get(strings.ToLower(value.(string)))
-			return ok || option.DeprecatedKeys().Has(strings.ToLower(value.(string)))
+			return ok || (option.DeprecatedKeys() != nil && option.DeprecatedKeys().Has(strings.ToLower(value.(string))))
```
jakebailey, line 277:
```suggestion
			return ok || option.DeprecatedKeys() == nil || option.DeprecatedKeys().Has(strings.ToLower(value.(string)))
```

Maybe? As in there aren't any deprecated keys so it's okay to not have one?

navya9singh: but in that case, it should've already been cleared in the `option.EnumMap()` check, no?

iisaduan: If it has no deprecated keys _and_ we did not find it in the enummap (aka `ok==false`), then the function should return false, because it shoudn't be a recognized value

jakebailey: Sure, I'm probably just wrong 😄

PR #321, opened 2025-02-10
Little polish for hover and go-to-def prototypes

andrewbranch:
This gives hover syntax highlighting and resolves imports in go-to-definition:

https://github.com/user-attachments/assets/24e936e2-bbb8-441a-b6bd-f913ee5d918b

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -332,14 +332,14 @@ func codeFence(lang string, code string) string {
 	}
 	var result strings.Builder
 	result.Grow(len(code) + len(lang) + 2*ticks + 2)
-	for i := 0; i < ticks; i++ {
+	for range ticks {
 		result.WriteByte('`')
 	}
 	result.WriteString(lang)
 	result.WriteByte('\n')
 	result.WriteString(code)
 	result.WriteByte('\n')
-	for i := 0; i < ticks; i++ {
+	for range ticks {
 		result.WriteByte('`')
 	}
 	return result.String()


internal/lsp/server.go
```diff
@@ -320,3 +321,14 @@ func (s *Server) Log(msg ...any) {
...
+
+func codeFence(lang string, code string) string {
+	if code == "" {
+		return ""
+	}
+	ticks := 3
+	for strings.Contains(code, strings.Repeat("`", ticks)) {
+		ticks++
+	}
+	return strings.Repeat("`", ticks) + "\n" + code + "\n" + strings.Repeat("`", ticks)
```
jakebailey: I don't know if it particularly matters at this point, but this would be faster as a `string.Builder` since the length of the output is statically known (`2*ticks+len(code)+2`)

andrewbranch: Interesting that I forgot to include `lang` but it works anyway. I guess it’s assumed to be the source language, but I won’t count on that being true in all editors.

jakebailey: Yes, omitting the lang means "use the lang of the current file".

internal/lsp/server.go
```diff
@@ -320,3 +321,19 @@ func (s *Server) Log(msg ...any) {
...
+	if code == "" {
+		return ""
+	}
+	ticks := 3
+	for strings.Contains(code, strings.Repeat("`", ticks)) {
+		ticks++
+	}
+	var result strings.Builder
+	result.Grow(len(code) + len(lang) + 2*ticks + 2)
+	result.WriteString(strings.Repeat("`", ticks) + lang + "\n")
```
jakebailey: Writing out a string concat defeats the perf benefit of the exact sized builder; you'll want to write these steps out explicitly and loop for `ticks`.

PR #322, opened 2025-02-10
Enable some more linters, fix noteworthy bugs

jakebailey:
This enables some more lint rules, fixing some notable bugs.

The most interesting ones are those caught by `staticcheck`, which unfortunately makes linting a good 20% or more slower, mainly due to a perf bug in `x/exp` (which I think I can fix, but may not be taken upstream...)


internal/checker/flow.go
```diff
@@ -2510,8 +2512,7 @@ func (c *Checker) isReachableFlowNodeWorker(flow *ast.FlowNode, noCacheCheck boo
 			// Cache is unreliable once we start adjusting labels
 			c.lastFlowNode = nil
 			reduceLabels = append(reduceLabels, flow.Node.AsFlowReduceLabelData())
-			result := c.isReachableFlowNodeWorker(flow.Antecedent, false /*noCacheCheck*/)
-			reduceLabels = reduceLabels[:len(reduceLabels)-1]
+			result := c.isReachableFlowNodeWorker(flow.Antecedent, false /*noCacheCheck*/, reduceLabels)
 			return result
```
jakebailey: @ahejlsberg This is a bug found by `staticcheck`; `reduceLabels` is local to this function (and the other place where this pattern is repeated), so pushing and popping here is actually ineffectual. I didn't notice this during the initial review.

I think the intent is to pass this slice downward? You may have a clearer idea of a fix here.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -1446,7 +1446,6 @@ func getFileNamesFromConfigSpecs(
 	host vfs.FS,
 	extraFileExtensions []fileExtensionInfo,
 ) []string {
-	extraFileExtensions = []fileExtensionInfo{}
```
jakebailey, line 1430: This was probably a `extraFileExtensions = []` parameter in the original TS code. @navya9singh

internal/checker/checker.go
```diff
@@ -7466,7 +7466,7 @@ func (c *Checker) isPromiseResolveArityError(node *ast.Node) bool {
 
 func getErrorNodeForCallNode(node *ast.Node) *ast.Node {
 	if ast.IsCallExpression(node) {
-		node := node.Expression()
+		node = node.Expression()
```
jakebailey: This is a fixed bug; the old code is a little more complicated to read comparatively (but I believe I did the right thing)

jakebailey:
Drafting for now but I would like it if @ahejlsberg could take a look at the flow node thing and fix it better than my bad attempt.

internal/checker/flow.go
jakebailey: @ahejlsberg The new baselines actually show that I fixed a bug here!

jakebailey:
Going to split this up in to different PRs.

PR #326, opened 2025-02-10
Fix emit crash

jakebailey:
Emit was failing because the transformer deletes the `in` keyword, which is part of type parameters, but is also a real keyword in runtime code. A debug assert that asserted that the operator can't be nil would have caught it, but we don't have those quite yet.

Add a smoke test which tries to compile the old compiler, in race mode, both with and without multithreading.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,6 +181,7 @@ jobs:
       - format
       - generate
       - tidy
+      - smoke
 
     steps:
       - name: Check required jobs


jakebailey:
Well, that was supposed to show that emit fails, so that's confusing

jakebailey:
I'm dumb and left the code I was testing in the PR

internal/transformers/typeeraser.go
```diff
@@ -32,8 +32,6 @@ func (v *TypeEraserTransformer) visit(node *ast.Node) *ast.Node {
 		ast.KindConstKeyword,
 		ast.KindDeclareKeyword,
 		ast.KindReadonlyKeyword,
-		ast.KindInKeyword,
-		ast.KindOutKeyword,
```
jakebailey, line 36: I think it's safe to remove these for now because they should only be elided in type parameters, but those will have been deleted higher up in the tree anyway.

rbuckton:
Out doesn't matter as much, and I've already fixed `in` in #284.

jakebailey:
Ah, what I did is the same as your PR then; this is blocking other testing, so I would like to try and get this one in ASAP

PR #328, opened 2025-02-11
fixing: interface {} is nil, not string

navya9singh:
Fix for https://github.com/refinedev/refine/blob/351ec87cb0bbc513cb0a5bb9d782aae10964ad45/packages/react-router-v6/tsconfig.declarations.json 
```
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/tsoptions.normalizeNonListOptionValue(0xde0d00?, {0xc000150540?, 0xe09e80?}, {0x0?, 0x0?})

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -154,15 +154,14 @@ func main() {
 
 	currentDirectory = tspath.GetDirectoryPath(configFilePath)
 	// !!! is the working directory actually the config path?
-	host := ts.NewCompilerHost(compilerOptions, currentDirectory, fs)
+	host := ts.NewCompilerHost(compilerOptions, currentDirectory, fs, defaultLibraryPath)
 
 	parseStart := time.Now()
 	program := ts.NewProgram(ts.ProgramOptions{
-		ConfigFilePath:     configFilePath,
-		Options:            compilerOptions,
-		SingleThreaded:     opts.devel.singleThreaded,
-		Host:               host,
-		DefaultLibraryPath: defaultLibraryPath,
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: testdata/baselines/reference/config/tsconfigParsing/parses tsconfig with extends, files, include and other options with json api.js [hunk 1]
```diff
@@ -48,7 +49,7 @@ CompilerOptions::
   "allowUnusedLabels": null,
   "assumeChangesOnlyAffectDirectDependencies": null,
   "alwaysStrict": null,
-  "baseUrl": "",
+  "baseUrl": "/",
```
jakebailey, line 52: Is `/` definitely right here? I would have thought this is preserved but maybe not.

jakebailey: Yeah, when I set `baseUrl=""` in a tsconfig, that seems to come out the other side as `"./"`...

jakebailey:
Hm, but on `refinedev/refine/packages/react-router-v6/tsconfig.json`, `--showConfig` says:

```
"baseUrl": "./",
```

But tsgo says:

```
"baseUrl": "/home/jabaile/work/top-tests/repos/refinedev/refine/packages/react-router-v6",
```

So something must be short circuiting before `normalizeNonListOptionValue`...

jakebailey:
Oh, it really is absolute, it's just that `--showConfig` itself converts paths to relative paths when possible. That's fun...

PR #329, opened 2025-02-11
Ensure tsgo is built on initial watch event

jakebailey:
This explains why things weren't working; the initial event doesn't have any paths. Explicitly denote the initial run with an undefined path set.


Herebyfile.mjs
```diff
@@ -126,20 +126,27 @@ export const buildWatch = task({
     name: "build:watch",
     run: async () => {
         await watchDebounced("build:watch", async (paths, abortSignal) => {
+            console.log(paths);
```
andrewbranch: Did you mean to leave this in?

jakebailey: Definitely not.

PR #330, opened 2025-02-11
Add debug flag to herebyfile, consolidate tag handling

jakebailey:
Adds `--debug`, which is useful for getting a more debug attachable build, and then moves some tag handling around so that in the future we can add some extra tags for stuff like conditional debug asserting builds / tests or something.


jakebailey:
It might be better for this to be called `noopt` or something since that's what it actually does, but, idk.

I have not gone and added this to launch configs or anything.

PR #331, opened 2025-02-11
Skip trivia on go-to-definition.

DanielRosenwasser:
Before:
image

After:
image

Happy to hoist the logic into a helper since we'll need it almost everywhere anyway (though I'm not sure what we should call it. `get____LocationForRange`?).

### DIFF
--- a/internal/ls/definition.go
+++ b/internal/ls/definition.go
@@ -25,7 +25,7 @@ func (l *LanguageService) ProvideDefinitions(fileName string, position int) []Lo
 		for _, decl := range symbol.Declarations {
 			file := ast.GetSourceFileOfNode(decl)
 			loc := decl.Loc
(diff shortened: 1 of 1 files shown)


internal/ls/definition.go
```diff
@@ -19,9 +23,13 @@ func (l *LanguageService) ProvideDefinitions(fileName string, position int) []Lo
 
 		locations := make([]Location, 0, len(symbol.Declarations))
 		for _, decl := range symbol.Declarations {
+			file := ast.GetSourceFileOfNode(decl)
+			loc := decl.Loc
+			pos := scanner.SkipTrivia(file.Text, loc.Pos())
```
andrewbranch: This is equivalent to `node.getStart(sourceFile)`:

```suggestion
			pos := scanner.GetTokenPosOfNode(decl, file, false /*includeJsDoc*/)
```

PR #336, opened 2025-02-11
add simple watch mode

iisaduan:
Adds a very simple watch mode- it recompiles and emits the whole project if it detects any of the files have been modified (or deleted) since the last "snapshot" of the files.

Avg times for a watch cycle that detects no changes on my devbox (recorded using `tsgo watchTime -p [path/to/tsconfig]`. this extra mode is not included in the PR, but I can set this up for you if anyone else would like to test it) 
| `path/to/tsconfig`  | tsconfig parsing time | program creation time | time to check all files for changes | total time |
|:----:|:----:|:----:|:----:|:----:|
| `typescript/src` | `.9ms` | `48ms` | `4.5ms` | `55ms`
| `vscode/src` | `280ms` | `1750ms` | `145ms`| `2100ms`

- future improvements/additional features for file watcher
  - watch options currently do nothing (what do we want to allow specifying in corsa?)
  - ability to watch individual files/dirs
  - ability to watch buffers? (for lsp)
  - watch events channel
  - make watcher not dependent on program creation (this is currently the thing that takes the longest in the cycle, and most of the things happening in program creation are probably unnecessary)
  - make `tsconfigParsing` faster
- tests are not fully ported yet

### DIFF
--- a/internal/execute/verifytscwatch_test.go
+++ b/internal/execute/verifytscwatch_test.go
@@ -45,14 +45,14 @@ func verifyWatch(t *testing.T, test *tscInput, scenario string, edits []*testTsc
 }
 
 func listToTsconfig(base string, tsconfigOpts ...string) (string, string) {
-	optionString := strings.Join(tsconfigOpts, ",\n        ")
+	optionString := strings.Join(tsconfigOpts, ",\n            ")
 	tsconfigText := `{
-		"compilerOptions": {
+	"compilerOptions": {
 `
-	after := "        "
+	after := "            "
 	if base != "" {
-		tsconfigText += "        " + base
-		after = ",\n        "
+		tsconfigText += "            " + base
+		after = ",\n            "
 	}
 	if len(tsconfigOpts) != 0 {
 		tsconfigText += after + optionString
--- a/testdata/baselines/reference/tscWatch/noEmit/dts-errors-without-dts-enabled.js
+++ b/testdata/baselines/reference/tscWatch/noEmit/dts-errors-without-dts-enabled.js
@@ -6,9 +6,9 @@ Input::-w
 const a = class { private p = 10; };
 //// [/home/src/workspaces/project/tsconfig.json] new file
 {
-		"compilerOptions": {
-        "noEmit": true,
-        "outFile": "../outFile.js"
+	"compilerOptions": {
+            "noEmit": true,
+            "outFile": "../outFile.js"
 	}
 }
 
@@ -164,8 +164,8 @@ const a = "hello";
 //// [/home/src/workspaces/project/a.ts] no change
 //// [/home/src/workspaces/project/tsconfig.json] modified. new content:
 {
-		"compilerOptions": {
-        "outFile": "../outFile.js"
+	"compilerOptions": {
+            "outFile": "../outFile.js"
 	}
 }
 
@@ -178,9 +178,9 @@ Output::
 //// [/home/src/workspaces/project/a.ts] no change
 //// [/home/src/workspaces/project/tsconfig.json] modified. new content:
 {
-		"compilerOptions": {
-        "noEmit": true,
-        "outFile": "../outFile.js"
+	"compilerOptions": {
+            "noEmit": true,
+            "outFile": "../outFile.js"
 	}
 }
 
@@ -207,8 +207,8 @@ const a = class {
 //// [/home/src/workspaces/project/a.ts] no change
 //// [/home/src/workspaces/project/tsconfig.json] modified. new content:
 {
-		"compilerOptions": {
-        "outFile": "../outFile.js"
+	"compilerOptions": {
+            "outFile": "../outFile.js"
 	}
 }
 
@@ -221,9 +221,9 @@ Output::
 //// [/home/src/workspaces/project/a.ts] no change
 //// [/home/src/workspaces/project/tsconfig.json] modified. new content:
 {
-		"compilerOptions": {
-        "noEmit": true,
-        "outFile": "../outFile.js"
+	"compilerOptions": {
+            "noEmit": true,
+            "outFile": "../outFile.js"
 	}
 }
 
--- a/testdata/baselines/reference/tscWatch/noEmit/dts-errors.js
+++ b/testdata/baselines/reference/tscWatch/noEmit/dts-errors.js
@@ -6,10 +6,10 @@ Input::-w
 const a = class { private p = 10; };
 //// [/home/src/workspaces/project/tsconfig.json] new file
 {
-		"compilerOptions": {
-        "noEmit": true,
-        "outFile": "../outFile.js",
-        "declaration": true
+	"compilerOptions": {
+            "noEmit": true,
+            "outFile": "../outFile.js",
+            "declaration": true
 	}
 }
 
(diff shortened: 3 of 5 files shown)


cmd/tsgo/main.go
```diff
@@ -121,6 +121,9 @@ func main() {
 			os.Exit(int(execute.CommandLine(newSystem(), nil, args[1:])))
 		case "lsp":
 			os.Exit(runLSP(args[1:]))
+		case "watchTime":
+			watchTime(newSystem(), args[1:])
+			os.Exit(0)
```
jakebailey: Is this just for local debugging?

internal/execute/system.go
```diff
@@ -2,13 +2,16 @@ package execute
...
 	"io"
+	"time"
 
 	"github.com/microsoft/typescript-go/internal/vfs"
 )
 
 type System interface {
 	Writer() io.Writer
 	EndWrite() // needed for testing
+	IsTestDone() bool
```
jakebailey: I assume this is a way to make the program exit; I think we should probably make sure to replace this with something that uses `context.Context`.

cmd/tsgo/main.go
iisaduan: Its a testing mode to see how long certain parts of the watch cycle take

internal/bundled/embed.go
```diff
@@ -85,10 +85,9 @@ func (vfs *wrappedFS) Stat(path string) vfs.FileInfo {
 		if rest == "" || rest == "libs" {
 			return &fileInfo{name: rest, mode: fs.ModeDir}
 		}
-		if libName, ok := strings.CutPrefix(rest, "libs/"); ok {
-			if lib, ok := embeddedContents[libName]; ok {
-				return &fileInfo{name: libName, size: int64(len(lib))}
-			}
+		if lib, ok := embeddedContents[rest]; ok {
+			libName, _ := strings.CutPrefix(rest, "libs/")
+			return &fileInfo{name: libName, size: int64(len(lib))}
```
jakebailey, line 90: Oh, whoops, I must have did this wrong the first time around.

internal/execute/watch.go
```diff
@@ -0,0 +1,79 @@
...
+	// updateProgram()
+	w.program = compiler.NewProgramFromParsedCommandLine(w.options, w.host)
+	if hasBeenModified(w, w.program) {
+		fmt.Fprint(w.sys.Writer(), "build starting at ", w.sys.Now(), w.sys.NewLine())
+		timeStart := w.sys.Now()
+		w.compileAndEmit()
+		fmt.Fprint(w.sys.Writer(), "build finished in", w.sys.Now().Sub(timeStart), w.sys.NewLine())
+	} else {
+		// print something???
+		fmt.Fprint(w.sys.Writer(), "no changes detected at ", w.sys.Now(), w.sys.NewLine())
```
jakebailey: This is going to constantly print something, right? Do we actually want this here?

internal/execute/watch.go
```diff
@@ -0,0 +1,79 @@
...
+		w.host = compiler.NewCompilerHost(w.options.CompilerOptions(), w.sys.GetCurrentDirectory(), w.sys.FS(), w.sys.DefaultLibraryPath())
+	}
+	watchInterval := 1000 * time.Millisecond
+	for {
+		watchCycle(w)
+		time.Sleep(watchInterval)
+	}
+}
+
+func watchCycle(w *watcher) {
```
jakebailey: It seems like all of these funcs could just be methods on `watcher` and then put into `watcher.go`.

internal/execute/watch.go
```diff
@@ -0,0 +1,79 @@
...
+	if errorsInTsConfig(w) {
+		// these are unrecoverable errors--report them and do not build
+		return
+	}
+	// updateProgram()
+	w.program = compiler.NewProgramFromParsedCommandLine(w.options, w.host)
```
jakebailey, line 27: Should see if we can reuse something from the LSP here, since IIRC @andrewbranch's current code is able to reuse ASTs for stuff that doesn't change. Of course, reloading every time is reliable.

internal/execute/watch.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+			return true
+		}
+		w.options = configParseResult
+		w.host = compiler.NewCompilerHost(w.options.CompilerOptions(), w.sys.GetCurrentDirectory(), w.sys.FS(), w.sys.DefaultLibraryPath())
+	}
+	return false
+}
+
+func hasBeenModified(w *watcher, program *compiler.Program) bool {
```
jakebailey: Do we have a strategy for this that handles new/deleted files?

internal/execute/watch.go
```diff
@@ -0,0 +1,79 @@
...
+	// checks watcher's snapshot against program file modified times
+	currState := map[string]time.Time{}
+	filesModified := false
+	for _, sourceFile := range program.SourceFiles() {
+		fileName := sourceFile.FileName()
+		currState[fileName] = w.sys.FS().Stat(fileName).ModTime()
```
jakebailey: Stat returns nil if a file doesn't exist; does this crash when that happens?

internal/execute/watch.go
iisaduan: this is a todo after merging a basic watch mode. I'd like to have a watch mode up and working before adding optimizations

internal/execute/watch.go
iisaduan: The only case where that would happen is if the file was deleted between creating the list `program.SourceFiles()` and when `.Stat()` is called on that file (and its a pretty small time-- this entire function takes 150-200ms on vscode/src, which about 4% of the whole watch cycle). I'll add a statement for this case so that we won't get a nilpointer crash

internal/execute/watch.go
iisaduan: they should be handled: 
- New files should show up as `prevModified[fileName]` has an empty `time.Time` value, so we would set `filesModified = true`. 
- Deleted files means that after looping through the current program files, `prevModified` still has entries in it

jakebailey: Oh, we're just reloading the program over and over in a loop with a delay? I was assuming it didn't do that, but instead used the last program to make a list of places to recheck later.

DanielRosenwasser:
Is it intentional that failed lookup locations aren't watched? e.g. if you have the import

```
import * as foo from "./foo";
```

and `foo.ts` doesn't exist but is then created (or alternatively, if you're running with `--checkJs` and `foo.js` exists, but `foo.ts` is then created which has higher lookup precedence), then `--watch` should probably try to reconstruct the program with that in mind.

jakebailey:
Yeah, that's part of what I'm wondering; I don't know if looping on program itself is "the right thing", but rather some sort of combo of "all files and dirs that match the included paths" + "failed lookup locations".

andrewbranch:
The PR description says doing that is on deck. This version recreates a program every cycle, so failed lookup locations are moot.

jakebailey:
Yeah, as a first step, I think it might be okay, but I just imagine someone pulling our reop to try it out, running watch mode, and then we just use CPU at 100%... Which, this is WIP, so they are getting what we advertise...

iisaduan:
> Is it intentional that failed lookup locations aren't watched? 

> The PR description says doing that is on deck

Yup, recreating the program accounts for "watching" failed lookup locations. The next step (which I'm started on while waiting for reviews here) after this PR is to remove program creation out of the cycle, in which case we will specifically need to account for the failed lookup locations. 

It's taken me a while to get to that because I wanted to make sure we had some baseline testing capabilities for watch mode, so that when we do add optimizations in that vein, we have a way to catch bugs earlier

PR #337, opened 2025-02-11
Runner: Strip BOM and drop lines outside @filenames

### LINKED ISSUE #303: runner: Corsa baselines have (more) leading trivia
but rarely: declareModifierOnTypeAlias

### LINKED ISSUE #335: runner: lines from before first `@filename` are included with first file
declarationEmitAnyComputedPropertyInClass

sandersn:
1. Switch to VFS.ReadFile to strip BOM
2. Add missing strings.Builder.Reset() when reading test files.

Fixes #335
Fixes #303

### DIFF
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -474,7 +474,7 @@ func listFilesWorker(spec *regexp.Regexp, recursive bool, folder string) ([]stri
 	}
 	var paths []string
 	for _, entry := range entries {
-		path := filepath.Join(folder, entry.Name())
+		path := tspath.NormalizePath(filepath.Join(folder, entry.Name()))
 		if !entry.IsDir() {
 			if spec == nil || spec.MatchString(path) {
 				paths = append(paths, path)


internal/testutil/runner/compiler_runner.go
```diff
@@ -142,11 +143,10 @@ type compilerFileBasedTest struct {
 }
 
 func getCompilerFileBasedTest(t *testing.T, filename string) *compilerFileBasedTest {
-	bytes, err := os.ReadFile(filename)
-	if err != nil {
-		panic("Could not read test file: " + err.Error())
+	content, ok := vfs.FromOS().ReadFile(filename)
```
jakebailey, line 146: Is the filename here normalized?

sandersn: Here's the relevant snippet from listFilesRecursiveWorker:

```go
	entries, err := os.ReadDir(folder)
	if err != nil {
		return nil, err
	}
	var paths []string
	for _, entry := range entries {
		path := filepath.Join(folder, entry.Name())
```

Does vfs expect its filenames to come from vfs.ReadDir? In that case listFilesRecursiveWorker should also switch to vfs.

jakebailey: That won't be normalized then, no, so will break on Windows.

You don't need to use the VFS to find things but you should ensure to call NormalizePath.

PR #338, opened 2025-02-11
Add some project service tests

andrewbranch:
Using the public API for tests also helps discover some API cleanup that can be done.

### DIFF
--- a/internal/project/service.go
+++ b/internal/project/service.go
@@ -293,20 +293,7 @@ func (s *Service) delayUpdateProjectGraph(project *Project) {
 }
 
 func (s *Service) getOrCreateScriptInfoNotOpenedByClient(fileName string, path tspath.Path, scriptKind core.ScriptKind) *ScriptInfo {
-	if tspath.IsRootedDiskPath(fileName) || isDynamicFileName(fileName) || isBundledUri(fileName) {
-		return s.getOrCreateScriptInfoWorker(fileName, path, scriptKind, false /*openedByClient*/, "" /*fileContent*/, false /*deferredDeleteOk*/)
-	}
-	// !!!
-	// This is non rooted path with different current directory than project service current directory
-	// Only paths recognized are open relative file paths
-	// const info = this.openFilesWithNonRootedDiskPath.get(this.toCanonicalFileName(fileName))
-	// if info {
-	// 	return info
-	// }
-
-	// This means triple slash references wont be resolved in dynamic and unsaved files
-	// which is intentional since we dont know what it means to be relative to non disk files
-	return nil
+	return s.getOrCreateScriptInfoWorker(fileName, path, scriptKind, false /*openedByClient*/, "" /*fileContent*/, false /*deferredDeleteOk*/)
 }
 
 func (s *Service) getOrCreateOpenScriptInfo(fileName string, path tspath.Path, fileContent string, scriptKind core.ScriptKind, projectRootPath string) *ScriptInfo {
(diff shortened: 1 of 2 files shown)


internal/project/service_test.go
```diff
@@ -0,0 +1,220 @@
...
+	output             strings.Builder
+	logger             *project.Logger
+}
+
+func newProjectServiceHost(files map[string]string) *projectServiceHost {
+	var mapFs = make(fstest.MapFS, len(files))
+	for name, content := range files {
+		mapFs[name] = &fstest.MapFile{Data: []byte(content)}
+	}
+	fs := bundled.WrapFS(vfstest.FromMapFS(mapFs, false /*useCaseSensitiveFileNames*/))
```
andrewbranch: I know this is only a couple lines, but a vfstest helper for this might be nice, along with something that checks/normalizes the file name roots... it took me a while to figure out what was wrong when I made my MapFS file names start with `"/"` 😕

internal/project/service.go
```diff
@@ -331,20 +280,20 @@ func (s *ProjectService) delayUpdateProjectGraphs(projects []*Project, clearSour
...
-	if project.kind == ProjectKindAutoImportProvider || project.kind == ProjectKindAuxiliary {
+	if project.kind == KindAutoImportProvider || project.kind == KindAuxiliary {
 		return
 	}
 	// !!! throttle
 	project.updateIfDirty()
 }
 
-func (s *ProjectService) getOrCreateScriptInfoNotOpenedByClient(fileName string, path tspath.Path, scriptKind core.ScriptKind) *ScriptInfo {
-	if tspath.IsRootedDiskPath(fileName) || isDynamicFileName(fileName) {
```
jakebailey, line 347: I can't say I like having this bundled path check, given it has to copy that detection; where does this end up needed?

andrewbranch: Basically, the assumption is that non-disk paths are in-memory, so we won’t ask the file system for them.

internal/project/service_test.go
jakebailey: Yes, multiple people have expressed this; I will probably send one later I've just forgotten about it over and over...

internal/project/service.go
jakebailey: I feel confused; `bundled://` is a disk path, at least from our point of view. Is this just an optimization to not go to the doc registry to find them?

andrewbranch: It’s not; we have no idea how to handle URIs anywhere:

image

andrewbranch: Even if we handled URIs across the board, how are we supposed to know to let `bundled://` hit the file system like `file://`? In that world, we would treat `untitled://` as in-memory, but you could imagine a future where we have some other kind of handling for `https://`, so I’m not sure what the rule is supposed to be for knowing which schemes are to be looked up in the file system.

jakebailey: In a URI world, the only ones that shouldn't hit the disk are those that have been openFile'd, everything else needs to hit the disk, and it's up to the FS to say "I don't handle this scheme"

In a pure VS Code extension, that's how it works, the FS API can serve requests for `untitled://`, `vscode://`, `file://`, `github://`, `zipfs://`, etc.

I think that's why I'm confused, because I thought this should be as simple as "it's either opened or it's not".

jakebailey: (If we do switch everything to URIs, everything would be a URI, the bundled wrapper would handle the `bundled://` paths and let the next FS layer do the rest, and then it it hits `OS`, it panics if the scheme isn't `file://`)

andrewbranch: I think the `else` condition of that check can’t happen in LSP, so I removed the check altogether.

PR #339, opened 2025-02-11
Update to Go 1.24

jakebailey:
Updates dependencies, and switches to `go tool` away from the infamous `tools.go` hack.

Benchstat results: https://gist.github.com/jakebailey/25343ef0fd5c4a1b35e70d0bb4fa0da9

Most things are better, ~but json v2 is a lot worse all of a sudden!~


jakebailey:
Oh, v2 got slower because the API changed to match the official proposal.

jakebailey:
Go 1.24.1 is out; I think Go 1.24 should be pretty stable.

jakebailey:
(Selfishly, I want to wait for 1.24.1 to appear in Arch Linux, though... But `GOTOOLCHAIN` does work.)

andrewbranch:
I’m at the point where I need `weak` to move on with the API work, so it would be nice to merge whenever we can.

jakebailey:
Alright, I'm going to back off of 1.24.1 and just set the minimum to 1.24.0; setting that won't mean we don't use a newer one when possible.

PR #341, opened 2025-02-12
Omnibus cleanup from o1-mini's syntactic code review

RyanCavanaugh:
Most of this is spelling or clarification, but a few commits are actual bugfixes or might need more follow-up. I'll call them out.

I have some other review comments that *seem* correct or plausible, but need area owners to weigh in

Is calling `finishNode` sometimes optional?
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-137

Intentional backward treatment of `ok` here?
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-173

I don't know the intended caching semantics here
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-183

Seems like a naming error?
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-277

⚠️ This is definitely a logic error but I don't know what the right fix is!
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-285

I don't know if this case is actually possible or not
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-321

⚠️ Definitely a property write that is unconditionally overwritten!
https://gist.github.com/RyanCavanaugh/44184d18e0b51d1ed1f211ff43a5e887#review-329


internal/jsnum/jsnum_test.go
```diff
@@ -114,7 +114,6 @@ func TestBitwiseNOT(t *testing.T) {
 		input, want Number
 	}{
 		{-2147483649, Number(2147483647).BitwiseNOT()},
-		{-4294967296, Number(0).BitwiseNOT()},
```
RyanCavanaugh, line 117: Not clear if this was supposed to be some other test (it's duplicated on L118)

internal/scanner/scanner.go
```diff
@@ -1450,7 +1450,7 @@ func (s *Scanner) scanTemplateAndSetTokenValue(shouldEmitInvalidEscapeError bool
 		if ch == '\r' {
 			contents += s.text[start:s.pos]
 			s.pos++
-			if ch == '\n' {
+			if s.char() == '\n' {
```
RyanCavanaugh, line 1453: I didn't construct a repro here but `ch` is definitely not `\n` here, please validate

internal/semver/version_range_test.go
```diff
@@ -120,7 +120,7 @@ func TestVersionRanges(t *testing.T) {
 	lotsaOnes := strings.Repeat("1", 320)
 	assertRangesGoodBad(t, ">=1.2.3-1"+lotsaOnes, testGoodBad{
 		good: []string{"1.2.3-1" + lotsaOnes, "1.2.3-11" + lotsaOnes + ".1", "1.2.3-1" + lotsaOnes + ".1+build"},
-		bad:  []string{"1.2.3-" + lotsaOnes + ".1+build", "1.2.3-" + lotsaOnes + ".1+build"},
+		bad:  []string{"1.2.3-" + lotsaOnes + ".1+build"},
```
RyanCavanaugh, line 123: Not clear if this was supposed to be some other test case, but it's a duplicate

internal/semver/version_range_test.go
```diff
@@ -858,8 +858,6 @@ func TestTildesOfVersionRanges(t *testing.T) {
 		{"~1", "1.0.0", true},
 		{"~1", "1.2.0", true},
 		{"~1", "1.2.3", true},
-		{"~1", "1.2.0", true},
```
RyanCavanaugh, line 861: More duplicate tests

internal/tsoptions/tsconfigparsing.go
```diff
@@ -796,7 +796,7 @@ func parseOwnConfigOfJson(
 	// watchOptions := convertWatchOptionsFromJsonWorker(json.watchOptions, basePath, errors)
 	// json.compileOnSave = convertCompileOnSaveOptionFromJson(json, basePath, errors)
 	var extendedConfigPath []string
-	if json["extends"] != nil || json["extends"] == "" {
+	if json["extends"] != nil && json["extends"] != "" {
```
RyanCavanaugh, line 799: Seemed like this was the intent; @navya9singh

internal/checker/checker.go
```diff
@@ -7466,7 +7466,7 @@ func (c *Checker) isPromiseResolveArityError(node *ast.Node) bool {
 
 func getErrorNodeForCallNode(node *ast.Node) *ast.Node {
 	if ast.IsCallExpression(node) {
-		node := node.Expression()
+		node = node.Expression()
```
jakebailey, line 7469: staticcheck found this one too, which is good

internal/ast/ast.go
```diff
@@ -7534,7 +7534,7 @@ type Pragma struct {
 type PragmaKindFlags = uint8
 
 const (
-	PragmaKindFlagsNone PragmaKindFlags = iota
+	PragmaKindFlagsNone PragmaKindFlags = 1 << iota
 	PragmaKindTripleSlashXML
```
jakebailey: This isn't quite right; needs to be:

```suggestion
	PragmaKindFlagsNone PragmaKindFlags = 0
	PragmaKindTripleSlashXML PragmaKindFlags = 1 << iota
```

In order to keep the `None` value as zero.

internal/compiler/diagnostics/generate.go
```diff
@@ -86,8 +86,8 @@ func main() {
 		if m.ReportsUnnecessary {
 			buf.WriteString(`, reportsUnnecessary: true`)
 		}
-		if m.ElidedInCompatabilityPyramid {
-			buf.WriteString(`, elidedInCompatabilityPyramid: true`)
+		if m.ElidedInCompatibilityPyramid {
+			buf.WriteString(`, elidedInCompatibilityPyramid: true`)
```
jakebailey, line 91: I'm good at spelling

internal/jsnum/jsnum_test.go
jakebailey: Was probably intended to be a different value; note the other two are off by one from each other.

internal/printer/printer.go
```diff
@@ -3261,6 +3261,7 @@ func (p *Printer) emitModuleDeclaration(node *ast.ModuleDeclaration) {
 	}
 	if body == nil {
 		p.writeTrailingSemicolon()
+		p.exitNode(node.AsNode())
```
jakebailey: Not correct; strada has:

```ts
        if (!body) return writeTrailingSemicolon();
```

(Unless strada is itself broken)

internal/scanner/scanner.go
jakebailey: Yeah, this is a misport.

internal/jsnum/jsnum_test.go
jakebailey: Yeah, this is meant to match https://github.com/tc39/test262/blob/b0319e488dbea1dc8dad613559551f0ea3d47620/test/language/expressions/bitwise-not/S9.5_A2.1_T2.js, which I seem to have done poorly

internal/printer/printer.go
RyanCavanaugh: @rbuckton 🛎️

internal/jsnum/jsnum_test.go
RyanCavanaugh: Will fix

internal/tsoptions/tsconfigparsing.go
navya9singh: Strada had `|| json["extends"] == ""` but it makes sense to have it this way.

internal/printer/printer.go
rbuckton: No, this is a difference from Strada. This is correct, though I think I would rather have:

```go
if body == nil {
  p.writeTrailingSemicolon()
} else {
  p.writeSpace()
  p.emitModuleBlock(body.AsModuleBlock())
}
p.exitNode(node.AsNode())
```
So as to keep the enter/exit calls 1:1 in the function body.

rbuckton: As for the difference from Strada, Strada uses emit pipelining which was a complex beast arising from using stack variables to save emit state and handle things like JIT substitution from transforms. As Corsa's emitter is far simpler, it uses `enterNode`/`exitNode` instead which significantly cuts down on stack depth.

internal/ast/ast.go
rbuckton: Given how `iota` works, won't that give `PragmaKindTripleSlashXML` the value `2`? I've found I have to move `None` or any other explicitly set entries to the end of a `const ()` to avoid skipping bits.

jakebailey: https://go.dev/play/p/XGPy3ZvoCF2

jakebailey: Yeah, wow, I always thought iota reset on each mention. We'll need to vet everything now...

jakebailey: > Within a [constant declaration](https://go.dev/ref/spec#Constant_declarations), the predeclared identifier iota represents successive untyped integer [constants](https://go.dev/ref/spec#Constants). Its value is the index of the respective [ConstSpec](https://go.dev/ref/spec#ConstSpec) in that constant declaration, starting at zero

Wow.................

RyanCavanaugh: Well, that's certainly a design choice

internal/compiler/diagnostics/diagnostics_generated.go
```diff
@@ -462,10 +462,10 @@ var Importing_a_JSON_file_into_an_ECMAScript_module_requires_a_type_Colon_json_i
 var Named_imports_from_a_JSON_file_into_an_ECMAScript_module_are_not_allowed_when_module_is_set_to_0 = &Message{code: 1544, category: CategoryError, key: "Named_imports_from_a_JSON_file_into_an_ECMAScript_module_are_not_allowed_when_module_is_set_to_0_1544", text: "Named imports from a JSON file into an ECMAScript module are not allowed when 'module' is set to '{0}'."}
 var The_types_of_0_are_incompatible_between_these_types = &Message{code: 2200, category: CategoryError, key: "The_types_of_0_are_incompatible_between_these_types_2200", text: "The types of '{0}' are incompatible between these types."}
 var The_types_returned_by_0_are_incompatible_between_these_types = &Message{code: 2201, category: CategoryError, key: "The_types_returned_by_0_are_incompatible_between_these_types_2201", text: "The types returned by '{0}' are incompatible between these types."}
-var Call_signature_return_types_0_and_1_are_incompatible = &Message{code: 2202, category: CategoryError, key: "Call_signature_return_types_0_and_1_are_incompatible_2202", text: "Call signature return types '{0}' and '{1}' are incompatible.", elidedInCompatabilityPyramid: true}
```
jakebailey, line 465: Something went wrong here.

internal/compiler/diagnostics/generate.go
```diff
@@ -27,7 +27,7 @@ type diagnosticMessage struct {
 	Code                         int    `json:"code"`
 	ReportsUnnecessary           bool   `json:"reportsUnnecessary"`
 	ReportsDeprecated            bool   `json:"reportsDeprecated"`
-	ElidedInCompatabilityPyramid bool   `json:"elidedInCompatabilityPyramid"`
```
jakebailey, line 30: No joke, the original code is what is in the JSON file https://github.com/microsoft/TypeScript/blob/6af21a4a3df87e1f9d6218de63a3c36bcee870b0/src/compiler/diagnosticMessages.json#L1852

PR #344, opened 2025-02-12
Export nested namespaces

### LINKED ISSUE #309: binder: nested namespaces Y and Z in `namespace X.Y.Z` are not exported
declarationEmitNameConflicts2

could be related to #308

sandersn:
These are implicitly exported and should be marked during GetCombinedModifierFlags.

Fixes #309

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -1087,8 +1087,18 @@ func getCombinedFlags[T ~uint32](node *Node, getFlags func(*Node) T) T {
 	return flags
 }
 
+func getEffectiveModifierFlags(node *Node) ModifierFlags {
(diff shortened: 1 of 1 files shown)


sandersn:
I left a `!!!` for JSDoc but if post-parser jsdoc transforms work out I shouldn't need that--I'll just generate `export type` for `@typedef`.

sandersn:
After some discussion, I think it would be better to follow the existing (weird) parsing strategy of treating `namespace X.Y.Z` as `namespace X namespace Y namespace Z` and create synthetic `export` modifiers for nested namespaces. We'll need this ability for JS tags later, eg `@typedef`, so it would be good to have a simple place to make sure they work first.

PR #345, opened 2025-02-12
Add GetTokenAtPosition fuzz tests

andrewbranch:
Strada doesn’t have any unit tests for this; instead, it’s exercised a lot via fourslash tests, which we don’t have yet. I added fuzz testing because I’m considering trying to implement a version of this function that avoids creating the `children` array for all parents of the located node, and caught a couple bugs this way.

### DIFF
--- a/internal/core/binarysearch.go
+++ b/internal/core/binarysearch.go
@@ -10,13 +10,13 @@ func BinarySearchUniqueFunc[S ~[]E, E, T any](x S, target T, cmp func(int, E) in
 	low, high := 0, n-1
 	for low <= high {
 		middle := low + ((high - low) >> 1)
-		switch cmp(middle, x[middle]) {
-		case -1:
+		value := cmp(middle, x[middle])
+		if value < 0 {
 			low = middle + 1
-		case 0:
-			return middle, true
-		case 1:
+		} else if value > 0 {
 			high = middle - 1
+		} else {
+			return middle, true
 		}
 	}
 	return low, false


andrewbranch:
Different version of Node 😛

internal/testutil/jstest/node.go
```diff
@@ -0,0 +1,84 @@
...
+	return result, nil
+}
+
+func getNodeExe(t testing.TB) string {
+	t.Helper()
+
+	const exeName = "node"
+	exe, err := exec.LookPath(exeName)
+	if err != nil {
+		t.Skipf("%s not found: %v", exeName, err)
```
jakebailey: Can you lift this up into a `sync.Once` and then export a function which can do the Skip like we do in the `repo` package? In `jsnum`, I tried to ensure this skip happened as early as possible, since what'll happen is that we create a load of skipped subtests or fuzz tests, rather than skipping the top level thing.

internal/testutil/jstest/node.go
```diff
@@ -0,0 +1,84 @@
...
+
+// EvalNodeScriptWithTS is like EvalNodeScript, but provides the TypeScript
+// library to the script as the first argument.
+func EvalNodeScriptWithTS[T any](t testing.TB, script string, dir string, args ...string) (result T, err error) {
+	if dir == "" {
+		dir = t.TempDir()
+	}
+	tsDest := filepath.Join(dir, "typescript.js")
+	tsSrc := filepath.Join(repo.RootPath, "node_modules/typescript/lib/typescript.js")
+	tsText, err := os.ReadFile(tsSrc)
```
jakebailey: This probably won't work (some of the time) because it won't have copied any of TS's supporting files; can we just give node `file:///` URI to the lib in `node_modules`?

(or await import `url.fromFilePath` or whatever)

internal/jsnum/string_test.go
```diff
@@ -200,27 +199,14 @@ func TestStringRoundtrip(t *testing.T) {
...
-	if err != nil {
-		t.Skipf("%s not found: %v", exeName, err)
-	}
-	return exe
-}
-
 func TestStringJS(t *testing.T) {
 	t.Parallel()
 
-	exe := getNodeExe(t)
```
jakebailey, line 217: Just to note it explicitly here, but these calls are here at the top to make it so that the top level test is skipped, and not the subtests.

internal/ls/tokens_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/parser"
+	"github.com/microsoft/typescript-go/internal/repo"
+	"github.com/microsoft/typescript-go/internal/scanner"
+	"github.com/microsoft/typescript-go/internal/testutil/jstest"
+	"gotest.tools/v3/assert"
+)
+
+func FuzzTokens(f *testing.F) {
+	repo.SkipIfNoTypeScriptSubmodule(f)
```
jakebailey, line 22: In the jsnum tests, I have a goofy "skip if not fuzzing" check; I would also do that here. The main rationale is that running `go test` will actually run all known fuzz corpus tests (which are in some temp dir?) and that takes a really long time for a single test run.

I sort of want to take a dep on v8go (which makes this a lot faster; I have a branch locally for that), but maybe that doesn't work 100% of the time for a TS based test unless the test never uses node APIs.

jakebailey: Sorry, I meant that it runs the explicit corpus (`f.Add` calls), but that's just as bad if we spawn one process per invocation.

internal/testutil/jstest/node.go
```diff
@@ -0,0 +1,89 @@
...
+// EvalNodeScriptWithTS is like EvalNodeScript, but provides the TypeScript
+// library to the script as the first argument.
+func EvalNodeScriptWithTS[T any](t testing.TB, script string, dir string, args ...string) (result T, err error) {
+	if dir == "" {
+		dir = t.TempDir()
+	}
+	tsSrc := filepath.Join(repo.RootPath, "node_modules/typescript/lib/typescript.js")
+	tsLoaderScript := fmt.Sprintf(`import script from "./script.mjs";
+import * as ts from "%s";
+process.stdout.write(JSON.stringify(await script(ts, ...process.argv.slice(2))));`, tsSrc)
```
jakebailey, line 49: This (`"%s"` and `tsSrc`) won't work on Windows because `c:/` is a URI; this needs to be converted to a `file://` URI first, or, just use `createRequire`.

internal/core/binarysearch.go
```diff
@@ -0,0 +1,23 @@
...
+	for low <= high {
+		middle := low + ((high - low) >> 1)
+		switch cmp(middle, x[middle]) {
+		case -1:
+			low = middle + 1
+		case 0:
+			return middle, true
+		case 1:
+			high = middle - 1
+		}
```
jakebailey: For consistency, I think these should use comparison against zero; `slices.BinarySearchFunc` allows any value to be returned from cmp.

PR #346, opened 2025-02-12
Fix diagnostic handling, unbreaking bind timing

jakebailey:
After #241, GetOptionsDiagnostics calls GetGlobalDiagnostics which then initializes checkers which ensures files are bound, which then means we aren't tracking bind time correctly.

Shift stuff around to avoid that, and also fix a few other bugs I noticed along the way.

### DIFF
--- a/internal/execute/tsc.go
+++ b/internal/execute/tsc.go
@@ -172,6 +172,7 @@ func performCompilation(sys System, cb cbType, config *tsoptions.ParsedCommandLi
 	// todo: cache, statistics, tracing
 	program := compiler.NewProgramFromParsedCommandLine(config, host)
 	options := program.Options()
+	allDiagnostics := program.GetConfigFileParsingDiagnostics()
 
 	// todo: early exit logic and append diagnostics
 	diagnostics := program.GetSyntacticDiagnostics(nil)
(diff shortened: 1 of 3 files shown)


internal/execute/tsc.go
```diff
@@ -172,7 +172,6 @@ func performCompilation(sys System, cb cbType, config *tsoptions.ParsedCommandLi
 	// todo: cache, statistics, tracing
 	program := compiler.NewProgramFromParsedCommandLine(config, host)
 	options := program.Options()
-	allDiagnostics := program.GetOptionsDiagnostics()
```
iisaduan, line 175: This line should stay and `GetConfigParsingDiagnostics` should be added to `program`. It's actually the next call to `program.GetOptionsDiagnostics` that is wrong (line 180). `ConfigParsingDiagnostics` and `OptionsDiagnostics` currently do the same thing in corsa (because project references are not yet implemented). We should probably also add a !!! to `GetOptionsDiagnostics` because it's not correct ATM.
```suggestion
	allDiagnostics := program.GetConfigParsingDiagnostics()
```

jakebailey: `program.GetConfigParsingDiagnostics` doesn't exist; is there something else you mean?

PR #350, opened 2025-02-13
Accept some gopls suggested changes

jakebailey:
Ran:

```sh
parallel gopls check ::: internal/**/*.go | grep -vE 'unused|deprecated' | rg '^(/[^ ]+):.*' --replace '$1' | sort
```

And then hand picked some ones I thought were good.


taigrr:
these changes look great but for the spelling mistake

gabritto (approved):

PR #354, opened 2025-02-13
Erase syntactic type-only imports and exports

rbuckton:
This amends the type erasing transformer to elide syntactic type-only imports and exports, i.e. `import type ... from "foo"`, `import type x = ...`, `export type ... from "foo"`, and `export { type foo }`. 

Syntactic type-only import elision has no dependency on the checker. Semantic import elision (i.e., normal import elision in absence of `--verbatimModuleSyntax`) depends on the checker and will follow in a separate PR.

### DIFF
--- a/internal/transformers/typeeraser_test.go
+++ b/internal/transformers/typeeraser_test.go
@@ -16,7 +16,7 @@ func TestTypeEraser(t *testing.T) {
 		input  string
 		output string
 		jsx    bool
-		vsm    bool
+		vms    bool
 	}{
 		{title: "Modifiers", input: "class C { public x; private y }", output: "class C {\n    x;\n    y;\n}"},
 		{title: "InterfaceDeclaration", input: "interface I { }", output: ""},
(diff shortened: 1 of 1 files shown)


internal/transformers/typeeraser_test.go
```diff
@@ -62,6 +62,25 @@ func TestTypeEraser(t *testing.T) {
...
+		{title: "ImportEqualsDeclaration#3", input: "import x = y;", output: "import x = y;"},
+		{title: "ImportEqualsDeclaration#4", input: "import type x = y;", output: ""},
+		{title: "ImportDeclaration#1", input: "import \"m\";", output: "import \"m\";"},
+		{title: "ImportDeclaration#2", input: "import * as x from \"m\"; x;", output: "import * as x from \"m\";\nx;"},
+		{title: "ImportDeclaration#3", input: "import x from \"m\"; x;", output: "import x from \"m\";\nx;"},
+		{title: "ImportDeclaration#4", input: "import { x } from \"m\"; x;", output: "import { x } from \"m\";\nx;"},
+		{title: "ImportDeclaration#5", input: "import type * as x from \"m\";", output: ""},
+		{title: "ImportDeclaration#6", input: "import type x from \"m\";", output: ""},
+		{title: "ImportDeclaration#7", input: "import type { x } from \"m\";", output: ""},
+		{title: "ImportDeclaration#8", input: "import { type x } from \"m\";", output: ""},
```
andrewbranch, line 77: Oh wait, this isn’t right. This should result in `import {} from "m"`.

rbuckton: This test doesn't have `verbatimModuleSyntax` set, so it is actually correct. I'll add another test that sets that flag for the output you are expecting.

andrewbranch: Ah, gotcha

internal/transformers/typeeraser_test.go
```diff
@@ -16,6 +16,7 @@ func TestTypeEraser(t *testing.T) {
 		input  string
 		output string
 		jsx    bool
+		vsm    bool
```
andrewbranch: Nit:
```suggestion
		vms    bool
```

PR #357, opened 2025-02-13
Lint rule to forbid shadowing: use builtin govet shadow

sandersn:
Use govet's built-in rule instead of the small one I wrote. Its errors are a lot better, though still not perfect. 
Most of the errors are on code that is either confusing, brittle to changes, or could benefit from slightly more thought in naming.

The biggest single change is in relater.go, where I pushed down a shared-but-not-really declaration to individual branches. compiler_runner.go has a messy port of the CompilerRunner constructor, so there I eliminated some redundancy and got rid of the shadowing in the process.

The scanner has a bonus bug fix that would treat the extended unicode newline codes as end of comment for `/**` comments, but I'm going to package that separately as well.

Obsoletes #283


internal/jsnum/string.go
```diff
@@ -225,17 +225,17 @@ func parseFloatString(s string) float64 {
 	if hasExp {
 		sb.WriteString("e")
 
-		c, negative := strings.CutPrefix(c, "-")
+		c2, negative := strings.CutPrefix(c, "-")
```
jakebailey: This might as well be like:

```
var negative bool
c, negative = ...
```

internal/lsp/server.go
```diff
@@ -94,7 +94,7 @@ func (s *Server) Run() error {
 		req, err := s.read()
 		if err != nil {
 			if errors.Is(err, lsproto.ErrInvalidRequest) {
-				if err := s.sendError(nil, err); err != nil {
+				if err = s.sendError(nil, err); err != nil {
```
jakebailey: Really dislike these kinds of changes; I do think we can copy their linter and fix this, so would honestly prefer we want to accept this PR for that reason so we can flip these back

Specifically this change is only okay because there's a `continue`, otherwise it'd fall through to the return and have been modified.

internal/printer/namegenerator.go
```diff
@@ -285,37 +285,37 @@ func (g *NameGenerator) makeName(name *ast.Node) string {
 // TempFlags._i may be used to express a preference for that dedicated name.
 // Note that names generated by makeTempVariableName and makeUniqueName will never conflict.
 func (g *NameGenerator) makeTempVariableName(flags tempFlags, reservedInNestedScopes bool, privateName bool, prefix string, suffix string) string {
-	var tempFlags tempFlags
+	var tempFlagsValue tempFlags
```
jakebailey: Maybe just `temp`? Note `flags tempFlags` above; seems odd to have `flags` and `tempFlagsValue`...

internal/tsoptions/tsconfigparsing.go
```diff
@@ -527,8 +527,8 @@ func convertOptionsFromJson[O optionParser](optionsNameMap map[string]*CommandLi
 			opt, ok := optionsNameMap[key]
 			commandLineOptionEnumMapVal := opt.EnumMap()
 			if commandLineOptionEnumMapVal != nil {
-				val, ok := commandLineOptionEnumMapVal.Get(strings.ToLower(value.(string)))
-				if ok {
+				val, ok2 := commandLineOptionEnumMapVal.Get(strings.ToLower(value.(string)))
+				if ok2 {
```
jakebailey: This should really have been written:

```
if val, ok := commandLineOptionEnumMapVal.Get(strings.ToLower(value.(string))); ok {
```

And then with improved scoping roles, not complain.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -1136,7 +1136,7 @@ func parseJsonConfigFileContentWorker(
 		}
 		substituteStringArrayWithConfigDirTemplate(validatedFilesSpec, basePathForFileNames)
 	}
-	configFileSpecs := configFileSpecs{
+	configFileSpecsValue := configFileSpecs{
```
jakebailey, line 1138: I'd love to carve out this restriction to not complain about shadowing a type name, which can never result in a shadow-based error. The only way you can shadow and do something weird is if you do like:


```go
type foo struct {
    a, b int
}

func blah(a, b int) any {
	if a > 100 {
		type foo struct {
			b, a int
		}
		return foo{a, b}
    }
	return foo{a, b}
}
```

internal/jsnum/string.go
sandersn: Yes, that's much better, I just couldn't see it.

internal/tsoptions/tsconfigparsing.go
sandersn: It's trivial to run into a compiler error:

```
type foo struct {
  a, b int
}
func test() {
  var foo foo
  var bar foo // error here
}
```

I would like the lint rule to stop me writing `var foo foo` on its own because it makes the code annoyingly fragile -- `foo` doesn't always mean the type `foo` in a (potentially large) function.

internal/lsp/server.go
sandersn: I flipped it back to `err2` which is ugly but safe.

internal/tsoptions/tsconfigparsing.go
sandersn: I rewrote it because I like the form better, but the rule still complains.

jakebailey: Darn :(

PR #358, opened 2025-02-13
Implement import/export elision

rbuckton:
This implements checker-dependent, semantic import/export elision by reintroducing `EmitResolver`. Unlike the old compiler, the intent is to limit the use of `EmitResolver` to only the specific scenarios that require it, such as when the `verbatimModuleSyntax` compiler option is `false`.

To properly support import elision, this PR also implements `markLinkedReferences` and related functions in the checker.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -20,7 +20,7 @@ import (
 )
 
 type ProgramOptions struct {
-	ConfigFilePath               string
+	ConfigFileName               string
 	RootFiles                    []string
 	Host                         CompilerHost
 	Options                      *core.CompilerOptions
@@ -33,7 +33,7 @@ type Program struct {
 	host                         CompilerHost
 	programOptions               ProgramOptions
 	compilerOptions              *core.CompilerOptions
-	configFilePath               string
+	configFileName               string
 	nodeModules                  map[string]*ast.SourceFile
 	checkers                     []*checker.Checker
 	checkersOnce                 sync.Once
@@ -89,13 +89,14 @@ func NewProgram(options ProgramOptions) *Program {
 
 	rootFiles := options.RootFiles
 
-	p.configFilePath = options.ConfigFilePath
-	if p.configFilePath != "" {
-		jsonText, ok := p.host.FS().ReadFile(p.configFilePath)
+	p.configFileName = options.ConfigFileName
+	if p.configFileName != "" {
+		jsonText, ok := p.host.FS().ReadFile(p.configFileName)
 		if !ok {
 			panic("config file not found")
 		}
-		parsedConfig := parser.ParseJSONText(p.configFilePath, jsonText)
+		configFilePath := tspath.ToPath(p.configFileName, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
+		parsedConfig := parser.ParseJSONText(p.configFileName, configFilePath, jsonText)
 		if len(parsedConfig.Diagnostics()) > 0 {
 			p.configFileParsingDiagnostics = append(p.configFileParsingDiagnostics, parsedConfig.Diagnostics()...)
 			return p
@@ -110,7 +111,7 @@ func NewProgram(options ProgramOptions) *Program {
 			p.host,
 			p.host.GetCurrentDirectory(),
 			options.Options,
-			p.configFilePath,
+			p.configFileName,
 			/*resolutionStack*/ nil,
 			/*extraFileExtensions*/ nil,
 			/*extendedConfigCache*/ nil,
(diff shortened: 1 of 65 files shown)


internal/compiler/emitHost.go
```diff
@@ -22,6 +23,7 @@ type EmitHost interface {
 	CommonSourceDirectory() string
 	IsEmitBlocked(file string) bool
 	WriteFile(fileName string, text string, writeByteOrderMark bool, relatedSourceFiles []*ast.SourceFile, data *WriteFileData) error
+	getEmitResolver(file *ast.SourceFile) checker.EmitResolver
```
jakebailey: Moderately funky to have an unexported interface method; it implies that this package is the only one that could provide an EmitHost, and therefore the implementation could just be concrete if there's really one implementation.

internal/compiler/emitter.go
```diff
@@ -21,6 +22,7 @@ const (
 type emitter struct {
 	host               EmitHost
 	emitOnly           emitOnly
+	forceDtsEmit       bool
```
jakebailey: This looks to be unused, maybe leftover from something?

internal/compiler/emitHost.go
rbuckton: `EmitResolver` was never public in the old compiler, though that may not be as important here since none of this is part of an actual public API as of yet. It's something we'll need to keep in mind if that changes, though.

jakebailey: Nothing we have is public, since we're in the internal dir, so mainly it's just about it indicative that things could be structured differently.

rbuckton:
@jakebailey is the race detector detecting an actual race, or just that the a `map` created in the main thread gets reused (read and written to) in two different goroutines, even if those goroutines cannot overlap? I'm attempting to reuse the per-file checker created at check time as part of emit, and those operate in two different `WorkGroup`s and so shouldn't be able to overlap.

jakebailey:
No, it's detecting a real race; it doesn't speculatively determine that.

jakebailey:
Reading this, there are definitely real races; the WorkGroup doesn't guard anything, it's just there to manage the goroutines. The emit resolver needs to guard its use of a checker with a mutex since all goroutines managed by a WorkGroup will run concurrently. Or checker needs to have locks on its public API. That or this state needs to not be calculated by the checker (which I think is not an option).

internal/checker/emitresolver.go
```diff
@@ -0,0 +1,150 @@
...
+
+type emitResolver struct {
+	checker                 *Checker
+	isValueAliasDeclaration func(node *ast.Node) bool
+}
+
+func (c *Checker) NewEmitResolver(file *ast.SourceFile) EmitResolver {
+	// Ensure we have all the type information in place for this file so that all the
+	// emitter questions of this resolver will return the right information.
+	c.checkSourceFile(file)
```
jakebailey: Yeah, so the Checker itself is not at all concurrent safe; I believe `GetEmitResolver` is called concurrently, so this will definitely break by causing concurrent checks using the same checker.

internal/checker/emitresolver.go
```diff
@@ -0,0 +1,150 @@
...
+	return &emitResolver{checker: c}
+}
+
+func isConstEnumOrConstEnumOnlyModule(s *ast.Symbol) bool {
+	return isConstEnumSymbol(s) || s.Flags&ast.SymbolFlagsConstEnumOnlyModule != 0
+}
+
+func (r *emitResolver) IsReferencedAliasDeclaration(node *ast.Node) bool {
+	c := r.checker
```
jakebailey, line 30: A mutex lock here on `r` would help, though the `NewEmitResolver` call is another problem because now you need to coordinate one caller trying to `NewEmitResolver` and `c.checkSourceFile`, but then this other caller also using the checker.

jakebailey: It sorta implies that we should be defining the checker's "public and safe" API somewhere else, put a mutex on checker, then lock on every public API call. (So long as the checker is not reentrant via some weird host thing, which I believe it is not since `checker.Host` is the empty interface!)

internal/checker/emitresolver.go
rbuckton: All concurrent check operations should have been completed before emit starts due to `wg.RunAndWait()`. I thought it might be an issue with loop variable capturing in `Program.Emit`, but fixing that has not addressed the problem.

internal/checker/emitresolver.go
rbuckton: How would that help? An EmitResolver is bound to a single checker which is bound to a single file, and EmitResolvers are not reused between files.

internal/checker/emitresolver.go
jakebailey: I definitely do not want to rely on that; that sounds like a very real source of bugs to be calling the checker concurrently assuming it has already set everything up.

(If that were true, then we wouldn't need the call at all)

internal/checker/emitresolver.go
jakebailey: Program can have either 1 or 4 checkers; the checkers are shared between all files.

If EmitResolver is not unique to each checker, then no, I guess it doesn't help, but we really need to guard Checker somehow.

internal/compiler/program.go
```diff
@@ -580,26 +584,80 @@ func (p *Program) Emit(options *EmitOptions) *EmitResult {
...
 		}
-		emitters = append(emitters, emitter)
+		emitters = append(emitters, e)
+	}
+
+	// queue checker-dependent source files
+	if len(checkerDependentEmitters) > 0 {
+		for index, checker := range p.checkers {
+			if perCheckerEmitters := checkerDependentEmitters[index]; len(perCheckerEmitters) > 0 {
+				wg.Queue(func() {
```
jakebailey: This will limit emit parallelism to the number of checkers; can we not just throw a lock on the checker so that we can still be fully parallel when the checker is not being called, which should be most of the time?

internal/checker/emitresolver.go
rbuckton: I've amended the PR to partition source files to emit by checker, when their emit actually depends on a checker. This should align emit with `Program.CheckSourceFiles` such that all files bound to a specific `Checker` are emit sequentially, while each partition is emit concurrently. Files not bound to a `Checker` remain concurrent.

internal/compiler/program.go
rbuckton: I can look into that. In the end it may not matter once `.d.ts` emit is supported as they will make heavy use of each checker, so there is likely to be a lot of contention.

internal/checker/emitresolver.go
jakebailey: Most files we process won't have VMS set, so I don't think partitioning them will be as fast as we can get; I really think a checker lock (or, a lock in shared checker wrappers) is going to be the way to go.

internal/compiler/program.go
jakebailey: From what I can tell, EmitResolver is actually only dependent on the checker; can we just create one emit resolver per checker, then lock its operations? Then we don't need to fix the checker quite yet.

jakebailey:
I'm definitely happier with the state of this, though I do think we will want to pretty shortly figure out a better layering/locking scheme for the checker, since we are very likely in the future going to end up in a situation where the LSP is both doing say, a hover and a completion, which both want to touch the checker at the same time but would not be sharing this EmitResolver related lock.

The previous PR that does elision in VMS does have some comment that I think need to be address before this PR though, since this one includes that one?

rbuckton:
> The previous PR that does elision in VMS does have some comment that I think need to be address before this PR though, since this one includes that one?

That amounted to a missing test, which has since been addressed. After merging from `main` this should now be limited to just checker-dependent import elision.

internal/checker/emitresolver.go
```diff
@@ -0,0 +1,166 @@
+package checker
+
+import (
+	"sync"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+type EmitResolver interface {
```
jakebailey, line 10: Is there a reason to have this interface if there's just one concrete implementation?

rbuckton: I separated this out so it could eventually be moved to avoid circularities when the checker eventually needs to depend on the printer for type-to-string emit.

jakebailey: Hm, I guess I assumed that wouldn't be cyclic, but yeah we of course have to print types in diags, duh 😅

PR #361, opened 2025-02-14
Enable type baseline in diff tests

gabritto:


jakebailey:
I ended up doing this in #606.

PR #365, opened 2025-02-18
Add lint rule to disallow shadowing when it looks buggy

jakebailey:
This is a version of #357 but with custom lint rule which uses a control flow graph to determine when variable shadowing could never cause a negative side effect (the shadowing cannot flow into any code that uses the shadowed/outer variable). This reduces the number of false positives, and in one instance actually found an extra shadow that the lint rule in #357 didn't catch due to this PR being stricter about closing over outer variables.

Perf testing shows that this lint rule does not have any negative performance impact to linting (I have found other places that cause it to be slow).

### DIFF
--- a/_tools/customlint/plugin_test.go
+++ b/_tools/customlint/plugin_test.go
@@ -121,11 +121,11 @@ func TestPlugin(t *testing.T) {
 				return 0
 			})
 
-			fileContents, err := os.ReadFile(p)
-			assert.NilError(t, err)
+			fileContents, readErr := os.ReadFile(p)
+			assert.NilError(t, readErr)
 
-			goldenPath, err := filepath.Rel(testdataDir, p+".golden")
-			assert.NilError(t, err)
+			goldenPath, relErr := filepath.Rel(testdataDir, p+".golden")
+			assert.NilError(t, relErr)
 
 			expected := toGolden(fileContents, diags)
 


_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/ctrlflow"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/ast/inspector"
+	"golang.org/x/tools/go/cfg"
+)
+
+var shadowAnalyzer = &analysis.Analyzer{
+	Name:     "shadow",
+	Doc:      "check for possible unintended shadowing of variables",
```
sandersn:
```suggestion
	Doc:      "check for unintended shadowing of variables",
```

I'd also drop 'unintended' but I hate reading words.

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+			return
+		}
+		for _, spec := range n.Specs {
+			valueSpec, ok := spec.(*ast.ValueSpec)
+			if !ok {
+				continue
+			}
+			for _, ident := range valueSpec.Names {
+				idents = append(idents, ident)
+			}
```
sandersn: is it possible to copy valueSpec.Names to idents or even copy the pointer?

internal/compiler/module/resolver_test.go
```diff
@@ -338,6 +322,22 @@ func runTraceBaseline(t *testing.T, test traceTestCase) {
 				)
 			})
 		}
+
```
sandersn: this is a much nicer solution than renaming

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	for _, ident := range idents {
+		if ident.Name == "_" {
+			// Can't shadow the blank identifier.
+			continue
+		}
+		obj := s.pass.TypesInfo.Defs[ident]
+		if obj == nil {
+			continue
```
sandersn, line 122: this should never happen, right? Should it be a panic?

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+		_, shadowed := obj.Parent().Parent().LookupParent(obj.Name(), obj.Pos())
+		if shadowed == nil {
+			continue
+		}
+		shadowedScope := shadowed.Parent()
+		// Don't complain if it's shadowing a universe-declared identifier; that's fine.
+		if shadowedScope == types.Universe {
+			continue
+		}
+		// Ignore shadowing a type name, which can never result in a logic error.
```
sandersn, line 135: I think this is a bad idea as a matter of style, not correctness. I thought the examples in our code base were very confusing.

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+		if isTypeName(obj) || isTypeName(shadowed) {
+			continue
+		}
+		// Don't complain if the types differ: that implies the programmer really wants two different things.
```
sandersn, line 139: This is also suspect to me. It could imply that the programmer wants different things, but it seems almost as likely that the programmer didn't notice their type error.

Two errors in the same place is less likely, though.

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+			break
+		}
+	}
+	if start == nil {
+		return true
+	}
+
+	seen := make(map[*cfg.Block]struct{})
+	var posReachable func(b *cfg.Block) (found bool)
+	posReachable = func(b *cfg.Block) (found bool) {
```
sandersn: why is this declaration+init on two lines?

_tools/customlint/shadow.go
```diff
@@ -0,0 +1,257 @@
...
+			return scope
+		}
+	}
+	return nil
+}
+
+func (s *shadowPass) cfgFor(n ast.Node) *cfg.CFG {
+	switch n := n.(type) {
+	case *ast.FuncDecl:
+		return s.cfgs.FuncDecl(n)
```
sandersn, line 250: can this return `nil`?

_tools/customlint/testdata/shadow/shadow.go.golden
```diff
@@ -0,0 +1,333 @@
...
+			value := 2 // OK
+			println(value)
+		}
+		return 1234
+	}
+	
+	func F2() int {
+		value := 1
+		if value == 1 {
+			value := 2 // OK
```
sandersn, line 26: This is technically legal but disagrees with my understanding of banning shadowing as a lint rule, where entire styles of code are forbidden, whether or not they're specifically in error.

But I think it's fine to act like a compiler extension instead and only ban code that is (almost) certainly a bug.

internal/compiler/module/resolver_test.go
jakebailey: Unfortunately it only works because we ignore uses of the shadowed variable below, but this code is actually reachable! You can imagine this kind of setup:

```go
func foo() {
	value := 1

	print := func() {
		println(value)
	}

	setAndPrint := func(v int) {
		value := v
		println(value)
	}

	setAndPrint(1234)
	print()
}
```

Is this a bug?

_tools/customlint/shadow.go
jakebailey: This kind of thing is in the lint rule from the Go team; the rationale is that Go doesn't have subtyping, so it'd be impossible to misuse it, unless you're plumbing it through an `any` or other interface type.

_tools/customlint/shadow.go
jakebailey: I guess I just disagree; I didn't find any of the type name shadows to be confusing.

_tools/customlint/shadow.go
jakebailey: Go doesn't support recursive function literals.

_tools/customlint/shadow.go
jakebailey: No, the pass will have constructed CFGs for all functions. (Which if you think about it, sounds absurd, but is in fact not expensive.)

_tools/customlint/testdata/shadow/shadow.go.golden
jakebailey: Your PR to enable the rule isn't that either; not without the `strict` setting, which would ban all shadowing, and that'd be extremely overbearing. The Go team's WIP rule for this tries to strike a middle ground, but only a syntactic one that would error on this code because `value` is used lexically after the shadow, but my rule refines this by allowing shadowing where later shadowed variable uses are not reachable from the shadowing declaration.

_tools/customlint/shadow.go
jakebailey: No (there can be more than one), but I can use `append` better here.

_tools/customlint/shadow.go
jakebailey: It can happen:

```
// Defs maps identifiers to the objects they define (including
// package names, dots "." of dot-imports, and blank "_" identifiers).
// For identifiers that do not denote objects (e.g., the package name
// in package clauses, or symbolic variables t in t := x.(type) of
// type switch headers), the corresponding objects are nil.
//
// For an embedded field, Defs returns the field *Var it defines.
//
// Invariant: Defs[id] == nil || Defs[id].Pos() == id.Pos()
```

internal/scanner/scanner.go
```diff
@@ -593,9 +593,9 @@ func (s *Scanner) Scan() ast.Kind {
 						}
 						s.pos++
 					} else {
-						ch, size := s.charAndSize()
-						if stringutil.IsLineBreak(ch) {
-							break
+						commentCh, size := s.charAndSize()
+						if stringutil.IsLineBreak(commentCh) {
+							s.tokenFlags |= ast.TokenFlagsPrecedingLineBreak
```
jakebailey, line 597: @sandersn I copied this from your PR, but I'm not sure I understand this change; seems like a driveby bugfix?

internal/compiler/module/resolver_test.go
sandersn: You're right, I didn't notice that it was all in nested functions. It seems like this should be an error because control flow doesn't prove that the variable stops being used before the possibly-shadowed reference. Is it possible to always error on preceding references if they're nested in a function?

jakebailey: > Is it possible to always error on preceding references if they're nested in a function?

That's what this PR does, actually; my example isn't correct. Let me think up a better one.

jakebailey: Eh, no, the current rule doesn't error on the above (my editor was just running a WIP version I have been trying out).

jakebailey:
Updated with some CFA changes; I actually really like how it is now.

internal/compiler/module/resolver_test.go
jakebailey: Updated, and now this case is caught.

internal/checker/relater.go
```diff
@@ -3193,7 +3193,7 @@ func (r *Relater) structuredTypeRelatedToWorker(source *Type, target *Type, repo
 	var originalErrorChain *ErrorChain
 	saveErrorState := r.getErrorState()
 	relateVariances := func(sourceTypeArguments []*Type, targetTypeArguments []*Type, variances []VarianceFlags, intersectionState IntersectionState) (Ternary, bool) {
-		if result := r.typeArgumentsRelatedTo(sourceTypeArguments, targetTypeArguments, variances, reportErrors, intersectionState); result != TernaryFalse {
+		if result = r.typeArgumentsRelatedTo(sourceTypeArguments, targetTypeArguments, variances, reportErrors, intersectionState); result != TernaryFalse {
```
jakebailey, line 3196: This one is legitimately a bug.

internal/scanner/scanner.go
sandersn: Yes, that's right.

PR #370, opened 2025-02-19
Use ordered maps for options and users

### LINKED ISSUE #369: `paths` does not consistently work
```json5
// tsconfig.json
{
    "compilerOptions": {
        "target": "esnext",
        "module": "nodenext",
        "strict": true,
        "noEmit": true,
        "baseUrl": ".",
        "paths": {
            "@oops/*": ["./a.js"],
            "@oops/*/y": ["./b.js"],
            "@oops/*/y/x": ["./c.js"],
        }
    },
}
```

```ts
// @filename: main.ts
import * as Oops from "@oops/z/y/x";

let value: number = Oops.value;
```

```ts
// @filename: a.ts

export let value = 1;
```

```ts
// @filename: b.ts

export let value = true;
```

```ts
// @filename: c.ts

export let value = "hello!";
```

If you compile this, it will not always end up with a consistent result. Sometimes it succeeds, and sometimes it fails:

```
main.ts:4:5 - error TS2322: Type 'string' is not assignable to type 'number'.

4 let value: number = Oops.value;
      ~~~~~


Found 1 error in main.ts:4
```

jakebailey:
This swaps out the unordered maps like `map[string]any`, `map[string][]string` for `OrderedMap`s. This should make `paths` handling no longer nondeterministic.

This required some changes to `OrderedMap` to handle nil.

Fixes #369

### DIFF
--- a/_tools/customlint/plugin.go
+++ b/_tools/customlint/plugin.go
@@ -16,6 +16,7 @@ type plugin struct{}
 func (f *plugin) BuildAnalyzers() ([]*analysis.Analyzer, error) {
 	return []*analysis.Analyzer{
 		emptyCaseAnalyzer,
+		shadowAnalyzer,
 	}, nil
 }
 
--- a/_tools/customlint/plugin_test.go
+++ b/_tools/customlint/plugin_test.go
@@ -121,11 +121,11 @@ func TestPlugin(t *testing.T) {
 				return 0
 			})
 
-			fileContents, err := os.ReadFile(p)
-			assert.NilError(t, err)
+			fileContents, readErr := os.ReadFile(p)
+			assert.NilError(t, readErr)
 
-			goldenPath, err := filepath.Rel(testdataDir, p+".golden")
-			assert.NilError(t, err)
+			goldenPath, relErr := filepath.Rel(testdataDir, p+".golden")
+			assert.NilError(t, relErr)
 
 			expected := toGolden(fileContents, diags)
 
--- a/_tools/customlint/shadow.go
+++ b/_tools/customlint/shadow.go
@@ -1,0 +1,280 @@
+package customlint
+
+import (
+	"cmp"
+	"go/ast"
+	"go/token"
+	"go/types"
+	"slices"
+
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/ctrlflow"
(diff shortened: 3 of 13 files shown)


testdata/baselines/reference/tsoptions/commandLineParsing/Parse multiple compiler flags with input files in the middle.js
```diff
@@ -2,7 +2,7 @@ Args::
 ["--module", "commonjs", "--target", "es5", "0.ts", "--lib", "es5,es2015.symbol.wellknown"]
 
 CompilerOptions::
-{"lib":["lib.es5.d.ts","lib.es2015.symbol.wellknown.d.ts"],"module":1,"target":1}
+{"module":1,"target":1,"lib":["lib.es5.d.ts","lib.es2015.symbol.wellknown.d.ts"]}
```
jakebailey, line 5: Note that this is now in flag order, so that shows that this is working at least.

internal/collections/ordered_map.go
```diff
@@ -169,6 +189,34 @@ func (m *OrderedMap[K, V]) clone() OrderedMap[K, V] {
...
+	buf.WriteByte('{')
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
+
+	for i, k := range m.keys {
+		if i > 0 {
+			buf.WriteByte(',')
+		}
+
+		if err := enc.Encode(k); err != nil {
```
DanielRosenwasser: When you encode here, do you need to make sure that `k` is a `string` so that you actually end up with well-formed JSON?

DanielRosenwasser: Or: can you make the signature the following?

```diff
- func (m OrderedMap[K, V]) MarshalJSON() ([]byte, error) 
+ func (m OrderedMap[string, V]) MarshalJSON() ([]byte, error)

jakebailey: I do need to check, yeah.

No, you can't make a method on a specific instantiation; Go doesn't have specialization like this.

I can however make it immediately error when `K` is not `string`.

jakebailey: Excitingly Go doesn't mind silently converting numbers to string keys: https://go.dev/play/p/xyF2iHLDEHs

Which is what JS does, kinda.

PR #371, opened 2025-02-19
Pre-allocate `parsedPatterns` and cache some calculations

DanielRosenwasser:
For `paths` of `CompilerOptions`, we currently split the entries into a set of strings to do a quick lookup, and a (shorter) list of patterns to walk on. Later on, exact matches or any match with the longest prefix (breaking any ties by ordering in `paths`, see https://github.com/microsoft/typescript-go/issues/369 and https://github.com/microsoft/typescript-go/pull/370 for fixes around that behavior).

A lot of this work was done in https://github.com/microsoft/TypeScript/pull/59048, but that PR also cached the creation of these `parsedPatterns`. tsgo currently does not cache these at all, and on a fairly large internal codebase, this results in a significant amount of user time (13.7 seconds below) spent in `tryParsePatterns`:

`tryParsePatterns` spends 13.7 seconds of user time

This PR reintroduces the cache for `parsedPatterns` for `paths` in `CompilerOptions`, and places it on the `Resolver`'s caches.

This PR also does an extra walk of paths to provide a size hint. When uncached, this actually *added* time at the cost of more allocation. It may be worth experimenting with removing the extra walk.

### DIFF
--- a/internal/compiler/module/cache.go
+++ b/internal/compiler/module/cache.go
@@ -6,7 +6,6 @@ import (
 	"sync"
 
 	"github.com/microsoft/typescript-go/internal/ast"
-	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/compiler/packagejson"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/tspath"
@@ -19,7 +18,11 @@ type caches struct {
 	typeReferenceDirectiveCache                   *resolutionCache[*ResolvedTypeReferenceDirective]
 	packageJsonInfoCache                          *packagejson.InfoCache
 	resolvedTypeReferenceDirectiveLookupLocations map[*ResolvedTypeReferenceDirective]*LookupLocations
-	parsedPatternsCache                           map[*collections.OrderedMap[string, []string]]parsedPatterns
+
+	// Cached representation for `core.CompilerOptions.paths`.
+	// Doesn't handle other path patterns like in `typesVersions`.
+	parsedPatternsForPathsOnce sync.Once
+	parsedPatternsForPaths     *parsedPatterns
 }
 
 func newCaches(
--- a/internal/compiler/module/resolver.go
+++ b/internal/compiler/module/resolver.go
@@ -1030,7 +1030,7 @@ func (r *resolutionState) tryLoadModuleUsingPathsIfEligible() *resolved {
 		return continueSearching()
 	}
 	baseDirectory := getPathsBasePath(r.compilerOptions, r.resolver.host.GetCurrentDirectory())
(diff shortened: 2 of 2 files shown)


internal/compiler/module/resolver.go
```diff
@@ -1698,16 +1698,45 @@ func getPathsBasePath(options *core.CompilerOptions, currentDirectory string) st
...
 	patterns           []core.Pattern
 }
 
-func tryParsePatterns(paths *collections.OrderedMap[string, []string]) parsedPatterns {
-	// !!! TS has a weakmap cache
-	// We could store a cache on Resolver, but maybe we can wait and profile
-	matchableStringSet := collections.OrderedSet[string]{}
-	patterns := make([]core.Pattern, 0, paths.Size())
-	for path := range paths.Keys() {
+func tryParsePatternsCached(r *Resolver, paths *collections.OrderedMap[string, []string]) parsedPatterns {
```
jakebailey:
```suggestion
func (r *Resolver) tryParsePatternsCached(paths *collections.OrderedMap[string, []string]) parsedPatterns {
```

Seems like this doesn't need to be a top-level function.

internal/compiler/module/cache.go
```diff
@@ -18,6 +19,7 @@ type caches struct {
 	typeReferenceDirectiveCache                   *resolutionCache[*ResolvedTypeReferenceDirective]
 	packageJsonInfoCache                          *packagejson.InfoCache
 	resolvedTypeReferenceDirectiveLookupLocations map[*ResolvedTypeReferenceDirective]*LookupLocations
+	parsedPatternsCache                           map[*collections.OrderedMap[string, []string]]parsedPatterns
```
jakebailey: I can't say I super like holding onto these maps like this, but I guess we'll probably throw these objects away periodically.

jakebailey: Wait, we don't even need this to be a map; isn't tryParsePatternsCached only ever called on `r.compilerOptions.Paths`, so we could just sync.Once it.

andrewbranch: I haven’t exactly worked out how I want to do this, but eventually there will be a module resolution cache that persists across program updates and/or is shared between different programs.

internal/core/set.go
```diff
@@ -4,6 +4,13 @@ type Set[T comparable] struct {
 	M map[T]struct{}
 }
 
+// NewSetWithSizeHint creates a new Set with a hint for the number of elements it will contain.
+func NewSetWithSizeHint[T comparable](hint int) *Set[T] {
```
jakebailey, line 8: Probably just do `Set[T]` here, since it's always dereferenced at its use.

DanielRosenwasser: But everyone else is doing `*core.Set` (everyone else is one function)

PR #372, opened 2025-02-20
Cache ASTs in test runner (with supporting changes)

jakebailey:
This PR changes a few things to make it more safe to share ASTs.

- `SourceFileAffectingCompilerOptions` has been moved next to `CompilerOptions`, so that functionality can be shared. A new method `(*core.CompilerOptions).SourceFileAffecting()` constructs them.
- `Path` is now required on construction of a `SourceFile`, like `fileName`. This prevents us from racing when attempting to write that back to the file after parse.
- Do the external module collection up-front on parse. Since this process requires parent nodes, it previously raced with binding. Thankfully, this code always runs when files are parsed via a `Program` (so, would only not happen if manually constructing the AST). By moving this earlier, the work is completed before anyone can ever attempt to bind. The only cost is a slight increase in allocations, which again are already happening (just elsewhere), and those are actually due to the closure allocation in `SetParentInChildren`, not anything else.
- Document in SourceFile which fields are owned by which passes.

With these in place, the compiler test runner can now use a CompilerHost which caches the source files. This cache is indefinite, but only for the duration of test runs. If it comes to it, I could purge the cache every now and then.

Merging this into #275 brings the full suite of tests down from 43 seconds to about 8 seconds on my machine.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -7783,6 +7783,7 @@ type SourceFile struct {
 	Statements *NodeList // NodeList[*Statement]
 
 	// Fields set by parser
+
 	diagnostics                 []*Diagnostic
 	jsdocDiagnostics            []*Diagnostic
 	LanguageVersion             core.ScriptTarget


internal/ast/ast.go
```diff
@@ -7774,52 +7774,69 @@ type SourceFile struct {
...
+	Version int
+
+	// !!!
+
+	ImpliedNodeFormat       core.ModuleKind
+	CommonJsModuleIndicator *Node
+	ExternalModuleIndicator *Node
+	JsGlobalAugmentations   SymbolTable
+	tokenCacheMu            sync.Mutex
+	tokenCache              map[*Node][]*Node
```
jakebailey, line 7833: None of these are set by anyone, so, not sure where to put them.

internal/ast/ast.go
```diff
@@ -7875,8 +7889,7 @@ func (node *SourceFile) VisitEachChild(v *NodeVisitor) *Node {
 
 func (f *NodeFactory) UpdateSourceFile(node *SourceFile, statements *StatementList) *Node {
 	if statements != node.Statements {
-		updated := f.NewSourceFile(node.Text, node.fileName, statements).AsSourceFile()
-		updated.path = node.path
+		updated := f.NewSourceFile(node.Text, node.fileName, node.path, statements).AsSourceFile()
 		updated.LanguageVersion = node.LanguageVersion
 		updated.LanguageVariant = node.LanguageVariant
 		updated.ScriptKind = node.ScriptKind
```
jakebailey, line 7895: I feel like this implies that `NewSourceFile` should really also accept these rather than letting them be set directly on the source file, but, not really doing that yet.

internal/ast/ast.go
andrewbranch: I don’t think I actually meant to commit tokenCache yet, but it’s on the way

internal/parser/references.go
```diff
@@ -0,0 +1,220 @@
...
+		moduleNameExpr := ast.GetExternalModuleName(node)
+		// TypeScript 1.0 spec (April 2014): 12.1.6
+		// An ExternalImportDeclaration in an AmbientExternalModuleDeclaration may reference other external modules
+		// only through top - level external module names. Relative external module names are not permitted.
+		if moduleNameExpr != nil && ast.IsStringLiteral(moduleNameExpr) {
+			moduleName := moduleNameExpr.AsStringLiteral().Text
+			if moduleName != "" && (!inAmbientModule || !tspath.IsExternalModuleNameRelative(moduleName)) {
+				ast.SetParentInChildren(node) // we need parent data on imports before the program is fully bound, so we ensure it's set here
+				file.Imports = append(file.Imports, moduleNameExpr)
+				// !!! removed `&& p.currentNodeModulesDepth == 0`
```
jakebailey, line 81: @andrewbranch This is the one modification I made; I can't tell if this means that our old code (old or new codebase) were actually "wrong" and dependent on parse ordering or something. I just chose to omit this check since I think we're okay to just ignore it somewhere later (or, store in the Program as some sort of sidetable)?

andrewbranch: It was probably wrong upon being ported to Corsa. Shouldn’t be hard to implement another way once needed.

PR #374, opened 2025-02-20
Improve pos for JSDoc

sandersn:
Nodes with JSDoc now have coverage for their entire span. Previously there were holes because JSDoc nodes didn't include leading whitespace.

I'm not sure this is correct for children of the node with associated JSDoc. For example, a FunctionDeclaration may span (0, 273] and have a JSDoc child (0, 38], but its `function` keyword span was parsed as if JSDoc was whitespace--so it will have a span starting a 0 too, eg (0, 110].

Update: I talked to @andrewbranch about the language service and it sounds like he's dropping binary search, so it should work fine there. I need to follow up and find out about emit as well.

### DIFF
--- a/internal/parser/jsdoc.go
+++ b/internal/parser/jsdoc.go
@@ -108,7 +108,7 @@ func (p *Parser) parseJSDocNameReference() *ast.Node {
 }
 
 // Pass end=-1 to parse the text to the end
-func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int, pos int) *ast.Node {
+func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int, fullStart int) *ast.Node {
 	if end == -1 {
 		end = len(p.sourceText)
 	}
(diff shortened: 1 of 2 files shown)


internal/parser/jsdoc.go
```diff
@@ -106,7 +108,7 @@ func (p *Parser) parseJSDocNameReference() *ast.Node {
 }
 
 // Pass end=-1 to parse the text to the end
-func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int) *ast.Node {
+func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int, pos int) *ast.Node {
```
jakebailey: Sort of funky to have `start, end, pos`; is it actually more like pos+end and an offset?

sandersn: `pos` is the parser-style name, but the scanner calls it `fullStart`, so I'll go with that.

PR #378, opened 2025-02-24
Reimplement diagnosticPullOptions match function for unopened files

jakebailey:
Tabs that haven't been clicked on are not present in `vscode.workspace.textDocuments`, but diagnostics can still be requested for them thanks to `onTabs: true`. We can't ask VS Code for a `TextDocument` to then run `vscode.languages.match` since that's an async operation and would be slow (not to mention impossible because the matcher is synchronous).

Reading the docs for `match`, this function is _exclusively_ run for files that aren't in `vscode.workspace.textDocuments` where we only have URIs, so the current code seemingly will never succeed.

Instead, reimplement language selector matching using only the URI itself, which fixes the problem; upon reloading a window, diagnostics appear in tabs that haven't been focused yet.

### DIFF
--- a/_extension/src/extension.ts
+++ b/_extension/src/extension.ts
@@ -4,7 +4,9 @@ import * as vscode from "vscode";
 import {
     LanguageClient,
     LanguageClientOptions,
+    NotebookDocumentFilter,
     ServerOptions,
+    TextDocumentFilter,
     TransportKind,
 } from "vscode-languageclient/node";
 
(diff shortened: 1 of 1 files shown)


jakebailey:
A simpler version of this would just check the scheme and path only, since that's all we put into our document selector input; it seems like other people implementing this don't even look at `documentSelector` and assume it's the one they passed in. That'd certainly be simpler, but, sort of future proofing this...

PR #379, opened 2025-02-25
Implement faster GetTokenAtPosition with fewer allocations

andrewbranch:
Reworked the getTokenAtPosition algorithm to use a visitor and binary search over node lists instead of allocating children arrays for every ancestor of the target token. Before I deleted the old code, I had side-by-side benchmarks and this was running about 10x faster and more memory efficient. This also removes the language service's use of SyntaxLists altogether.

Differences between this implementation and Strada are baselined, with the exception that where Strada returns a SyntaxList, we substitute in its parent and compare the Corsa result against that.

### DIFF
--- a/internal/astnav/tokens_test.go
+++ b/internal/astnav/tokens_test.go
@@ -150,10 +150,14 @@ func tsGetTokensAtPositions(t testing.TB, fileText string, positions []int) []to
 	dir := t.TempDir()
 	err := os.WriteFile(filepath.Join(dir, "file.ts"), []byte(fileText), 0o644)
 	assert.NilError(t, err)
+
+	err = os.WriteFile(filepath.Join(dir, "positions.json"), []byte(core.Must(core.StringifyJson(positions, "", ""))), 0o644)
+	assert.NilError(t, err)
+
 	script := `
 		import fs from "fs";
-		export default (ts, positions) => {
-			positions = JSON.parse(positions);
+		export default (ts) => {
+			const positions = JSON.parse(fs.readFileSync("positions.json", "utf8"));
 			const fileText = fs.readFileSync("file.ts", "utf8");
 			const file = ts.createSourceFile(
 				"file.ts",
@@ -174,7 +178,7 @@ func tsGetTokensAtPositions(t testing.TB, fileText string, positions []int) []to
 			});
 		};`
 
-	info, err := jstest.EvalNodeScriptWithTS[[]tokenInfo](t, script, dir, core.Must(core.StringifyJson(positions, "", "")))
+	info, err := jstest.EvalNodeScriptWithTS[[]tokenInfo](t, script, dir, "")
 	assert.NilError(t, err)
 	return info
 }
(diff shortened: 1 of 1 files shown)


testdata/baselines/reference/astnav/GetTouchingPropertyName.mapCode.ts.baseline.txt
```diff
@@ -0,0 +1,86 @@
...
+45 │    contents: string[],
+46 │    focusLocations: TextSpan[][] | undefined,
+47 │    host: LanguageServiceHost,
+   │........ 14 lines omitted ........
+61 │                    flattenedLocations,
+62 │                );
+63 │            }
+64 │        },
+65 │    );
+66 │}》
```
andrewbranch: This should arguably return the SourceFile, not the function declaration, since the position falls between the JSDoc and the `export` keyword, and `allowPositionInLeadingTrivia` is false. I could make that work, but it’s unlikely to matter, and it would require either tracking a grandparent variable or ensuring the source file has parent pointers—as of now, these functions work with unbound files (though I don’t think they really need to).

internal/astnav/tokens.go
```diff
@@ -0,0 +1,257 @@
...
+	"sync"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/scanner"
+)
+
+var factoryPool = sync.Pool{
+	New: func() any {
+		return &ast.NodeFactory{}
```
jakebailey: This to me feels like a hazard; if we share node factories between different trees, we may end up leaving way more nodes in memory over time than we are expecting.

Is this just to amortize the cost of allocating the factories themselves? If so, we can just write `*factory = ast.NodeFactory{}` in `putNodeFactory` to clear it out.

andrewbranch: That seems to defeat the purpose of having pooled allocations, doesn’t it? We want to reuse a NodeFactory _over time_ in order to benefit from the pools. A single run of any of these functions is going to allocate exactly one node; isn’t that the equivalent of making a new factory in the scope of the function on each call?

jakebailey: Yes, that is unfortunately the effect; it's just that the factory was intended to do a _lot_ of work, like a whole parse or transformation pass, where the lifetime of the nodes is roughly the same.

I'm not sure what to do otherwise. Maybe we should be sticking node factories onto SourceFiles, so that later passes can use them instead?

jakebailey: I'll also note that by using a `sync.Pool`, it's likely the factories will get tossed away when they aren't used often enough to not be collected by GC, so depending on how quickly we call things, they may get collected anyway.

jakebailey: Looking at how this is uses, in one call it's only ever going to allocate one node; it makes me think that maybe it isn't worth it for sparse queries?

andrewbranch: I just realized that with my move of `GetOrCreateToken` to a SourceFile method in ast, I don’t need to rely on the factory anymore, which probably makes the most sense.

andrewbranch:
Any idea on the Windows error?

```
tokens_test.go:211: assertion failed: error is not nil: failed to run node: fork/exec C:\Program Files\nodejs\node.exe: The filename or extension is too long.
```

doesn't seem that long...

internal/astnav/tokens_test.go
```diff
@@ -0,0 +1,286 @@
...
+				}
+				return {
+					kind: ts.Debug.formatSyntaxKind(token.kind),
+					pos: token.pos,
+					end: token.end,
+				};
+			});
+		};`
+
+	info, err := jstest.EvalNodeScriptWithTS[[]tokenInfo](t, script, dir, core.Must(core.StringifyJson(positions, "", "")))
```
jakebailey: The arg list here is too big for Windows; will need to be trafficked in via a tempfile.

internal/astnav/tokens_test.go
```diff
@@ -0,0 +1,294 @@
...
+				}
+				return {
+					kind: ts.Debug.formatSyntaxKind(token.kind),
+					pos: token.pos,
+					end: token.end,
+				};
+			});
+		};`
+
+	info, err := jstest.EvalNodeScriptWithTS[[]tokenInfo](t, script, dir, "")
```
jakebailey, line 181: All of these script evals aren't too fast on Windows; something like 2 seconds to run this package's tests (>2x slower than Linux).

I think it's okay for now but we may want to revisit all of this JS-based testing some day.

andrewbranch: I guess we could save the expected results from the TS compiler to a JSON fixture and commit them if we ever need to.

PR #382, opened 2025-02-25
Add some symlink support

jakebailey:
This unbreaks `@types` auto loading when `@types/*` are symlinks to dirs, which is what pnpm does.

I think this probably completes symlink support outside of testing; `getAccessibleFileSystemEntries` appears to be the only thing that actually checked in the old codebase.

I am not 100% happy with the callback-y nature of this, but I can't yet think of a better way to handle this code without repetition.

### DIFF
--- a/internal/vfs/internal/internal.go
+++ b/internal/vfs/internal/internal.go
@@ -94,6 +94,7 @@ func (vfs *Common) GetAccessibleEntries(path string) (result vfs.Entries) {
 
 		if entryType&fs.ModeIrregular != 0 && vfs.Realpath != nil {
 			// Could be a Windows junction. Try Realpath.
+			// TODO(jakebailey): use syscall.Win32FileAttributeData instead
 			fullPath := path + "/" + entry.Name()
 			if realpath := vfs.Realpath(fullPath); fullPath != realpath {
 				if stat := vfs.stat(realpath); stat != nil {


internal/vfs/osvfs/realpath_test.go
```diff
@@ -28,12 +28,7 @@ func TestSymlinkRealpath(t *testing.T) {
 	assert.NilError(t, os.MkdirAll(target, 0o777))
 	assert.NilError(t, os.WriteFile(targetFile, []byte(expectedContents), 0o666))
 
-	if runtime.GOOS == "windows" {
-		// Don't use os.Symlink on Windows, as it creates a "real" symlink, not a junction.
-		assert.NilError(t, exec.Command("cmd", "/c", "mklink", "/J", link, target).Run())
-	} else {
-		assert.NilError(t, os.Symlink(target, link))
-	}
+	mklink(t, targetFile, linkFile)
```
jakebailey:
```suggestion
	mklink(t, target, link)
```

internal/vfs/internal/internal.go
```diff
@@ -60,16 +62,49 @@ func (vfs *Common) DirectoryExists(path string) bool {
...
+			if stat := vfs.stat(path + "/" + entry.Name()); stat != nil {
+				addToResult(entry.Name(), stat.Mode())
+			}
+			continue
+		}
+
+		if entryType&fs.ModeIrregular != 0 && vfs.Realpath != nil {
+			// Could be a Windows junction. Try Realpath.
+			fullPath := path + "/" + entry.Name()
+			if realpath := vfs.Realpath(fullPath); fullPath != realpath {
```
jakebailey, line 99: Note that this is _not_ the fastest way to do this at all; but the code to do this correctly on Windows is pretty funky (see the intermediate state in #186, which I will bring back). Something I want to address in a follow-up PR.

internal/vfs/internal/internal.go
```diff
@@ -60,16 +62,49 @@ func (vfs *Common) DirectoryExists(path string) bool {
...
+		if entryType&fs.ModeSymlink != 0 {
+			// Easy case; UNIX-like system will clearly mark symlinks.
+			if stat := vfs.stat(path + "/" + entry.Name()); stat != nil {
+				addToResult(entry.Name(), stat.Mode())
+			}
+			continue
+		}
+
+		if entryType&fs.ModeIrregular != 0 && vfs.Realpath != nil {
+			// Could be a Windows junction. Try Realpath.
```
jakebailey, line 96:
```suggestion
			// Could be a Windows junction. Try Realpath.
			// TODO(jakebailey): use syscall.Win32FileAttributeData instead
```

internal/vfs/internal/internal.go
jakebailey: For reference: https://github.com/microsoft/typescript-go/pull/186/commits/8998046c2383c986b8912611d95333b981398ba9

jakebailey: (realpath however, is very very reliable!)

PR #385, opened 2025-02-26
[lsp] Handle closed and deleted files

andrewbranch:
These changes aren’t particularly noticeable in most cases, mostly just releasing unused resources from memory.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -250,13 +250,13 @@ func (s *Server) handleInitialized(req *lsproto.RequestMessage) error {
 		Logger:             s.logger,
 	})
 	s.converters = &converters{projectService: s.projectService}
-	return s.sendResult(req.ID, nil)
+	return nil
 }
 
 func (s *Server) handleDidOpen(req *lsproto.RequestMessage) error {
 	params := req.Params.(*lsproto.DidOpenTextDocumentParams)
 	s.projectService.OpenFile(documentUriToFileName(params.TextDocument.Uri), params.TextDocument.Text, languageKindToScriptKind(params.TextDocument.LanguageId), "")
-	return s.sendResult(req.ID, nil)
+	return nil
 }
 
 func (s *Server) handleDidChange(req *lsproto.RequestMessage) error {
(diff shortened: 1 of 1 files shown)


internal/lsp/server.go
```diff
@@ -182,6 +182,10 @@ func (s *Server) handleMessage(req *lsproto.RequestMessage) error {
 		return s.handleDidOpen(req)
 	case *lsproto.DidChangeTextDocumentParams:
 		return s.handleDidChange(req)
+	case *lsproto.DidSaveTextDocumentParams:
+		return s.handleDidSave(req)
```
andrewbranch, line 186: This isn’t wired all the way through yet, but it should let us avoid some FS reads in the future.

internal/lsp/server.go
```diff
@@ -276,6 +288,18 @@ func (s *Server) handleDidChange(req *lsproto.RequestMessage) error {
 	return s.sendResult(req.ID, nil)
 }
 
+func (s *Server) handleDidSave(req *lsproto.RequestMessage) error {
+	params := req.Params.(*lsproto.DidSaveTextDocumentParams)
+	s.projectService.MarkFileSaved(documentUriToFileName(params.TextDocument.Uri), *params.Text)
+	return s.sendResult(req.ID, nil)
```
jakebailey: I didn't notice this before, but I thought these were notifications, not requests; do we need to reply with anything?

internal/project/service.go
```diff
@@ -307,21 +295,26 @@ func (s *Service) getOrCreateScriptInfoWorker(fileName string, path tspath.Path,
...
-		info.setTextFromDisk(fileContent)
+		if fromDisk {
+			info.setTextFromDisk(fileContent)
+		}
+
 		s.scriptInfosMu.Lock()
 		defer s.scriptInfosMu.Unlock()
 		if prevVersion, ok := s.filenameToScriptInfoVersion[path]; ok {
-			info.version = prevVersion
+			info.version = prevVersion + 1
```
jakebailey, line 317: Just to check, this is an internal version number, and not the version number sent by the client? It'd be nice to use the latter where possible (since that will enable us to send responses with that info for better behavior).

andrewbranch: Yeah, this is still ported from Strada where VS Code doesn’t send us a client version. I agree it would be good to migrate to have them in sync at some point.

internal/lsp/server.go
andrewbranch: Apparently not. I assumed we were supposed to acknowledge somehow.

jakebailey: Nope, we don't have to reply. I'm shocked it didn't crash. We should make sendResult panic if req.ID is missing just to check ourselves.

internal/tsoptions/parsinghelpers.go
```diff
@@ -280,6 +280,8 @@ func ParseCompilerOptions(key string, value any, allOptions *core.CompilerOption
 		allOptions.NoImplicitAny = parseTristate(value)
 	case "noImplicitThis":
 		allOptions.NoImplicitThis = parseTristate(value)
+	case "noLib":
```
jakebailey, line 283: This change appears to have broken main as the PR wasn't up to date.

jakebailey: Fixed in #390

PR #388, opened 2025-02-26
add `Stat()` to vfs and `ModTime` to `FileInfo` when filewriting to vfs 

iisaduan:
add `Stat()` to vfs and `ModTime` to `FileInfo `when filewriting to vfs.

For consistency, also cleans up some locations where `fs` is used instead of `vfs`

### DIFF
--- a/internal/bundled/embed.go
+++ b/internal/bundled/embed.go
@@ -96,11 +96,8 @@ func (vfs *wrappedFS) GetEntries(path string) []vfs.DirEntry {
 
 func (vfs *wrappedFS) Stat(path string) vfs.FileInfo {
 	if rest, ok := splitPath(path); ok {
-		switch rest {
-		case "":
-			return &fileInfo{name: "", mode: fs.ModeDir}
-		case "libs":
-			return &fileInfo{name: "libs", mode: fs.ModeDir}
+		if rest == "" || rest == "libs" {
+			return &fileInfo{name: rest, mode: fs.ModeDir}
 		}
 		if libName, ok := strings.CutPrefix(rest, "libs/"); ok {
 			if lib, ok := embeddedContents[libName]; ok {


internal/vfs/vfstest/vfstest.go
```diff
@@ -333,8 +334,9 @@ func (m *mapFS) WriteFile(path string, data []byte, perm fs.FileMode) error {
 	}
 
 	m.setEntry(path, m.getCanonicalPath(path), fstest.MapFile{
-		Data: data,
-		Mode: perm &^ umask,
+		Data:    data,
+		ModTime: time.Now(),
```
jakebailey, line 338: Maybe one day we'll want to fake this time but probably all good for now, so long as we never expose these times anywhere.

internal/bundled/embed.go
```diff
@@ -94,6 +94,13 @@ func (vfs *wrappedFS) GetEntries(path string) []fs.DirEntry {
 	return vfs.fs.GetEntries(path)
 }
 
+func (vfs *wrappedFS) Stat(path string) vfs.FileInfo {
+	if rest, ok := splitPath(path); ok {
+		return &fileInfo{name: rest, mode: fs.ModeDir, size: int64(len(embeddedContents[rest]))}
```
jakebailey: Er, sorry, no this isn't quite right.

jakebailey: Basically, this has to be written in such a way that it answers correctly for each of the different paths; that's a little annoying here, but what I would do is like:

- Check "", return a dir, no size
- Check "libs", return a dir, no size
- If starts with `libs/`, cut that prefix, then check if there's a file in `embeddedContents`, if there is, return what you have here (name is the cut off version, the basename)
- Otherwise, return nil

PR #393, opened 2025-02-27
Reduce ComputeLineStarts allocations

jakebailey:
Minor perf benefit; we can guess the size of the array by just counting newlines, eliminating any append allocations.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -313,7 +313,7 @@ func Coalesce[T *U, U any](a T, b T) T {
 }
 
 func ComputeLineStarts(text string) []TextPos {
-	var result []TextPos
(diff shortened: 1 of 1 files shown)


andrewbranch:
Does this show an improvement in any benchmarks?

jakebailey:
Not in the existing benchmarks, but it does appear in plain tsgo runs; I think it's just that we don't have emit benchmarks yet and emit recalculates this sometimes.

Comparing profiles, on before this function is 1.36% of the total runtime, after it's 0.89%. Before it's 12.9% of allocations (the most of any function), after it's 7.61%, going from 4277567 allocs to 2234419 allocs (~50% fewer).

image

image

PR #395, opened 2025-02-28
Add functions implementing a Clone method for all Nodes

weswigham:
This is pretty different from how we implement `factory.cloneNode` in the TS codebase, which is pretty JS-idiomatic, iterating over all the enumerable properties and copying them between the old and new nodes.

The declaration emitter pipeline does a *lot* of node cloning in scenarios where input nodes can't be reused as-is (due to context constraints) or where the same cached node tree needs to be reused in multiple places, so needs our cloning functions. I imagine it'll also be useful for quickfixes, eventually.


internal/ast/ast.go
```diff
@@ -1655,6 +1672,10 @@ func (f *NodeFactory) NewToken(kind Kind) *Node {
 	return newNode(kind, f.tokenPool.New())
 }
 
+func (node *Token) Clone(f *NodeFactory) *Node {
```
jakebailey, line 1694: If a factory is required, it seems like it might be better to put the clone method on the factory (which is similar to `New` and `Update`), but I guess it could go both ways.

internal/ast/deepclone.go
```diff
@@ -0,0 +1,42 @@
+package ast
+
+// Ideally, this would get cached on the node factory so there's only ever one set of closures made per factory
+func getDeepCloneVisitor(f *NodeFactory) *NodeVisitor {
+	visitor := NodeVisitor{
```
jakebailey: I would make this `&NodeVistor`; just out of a matter of style. Thankfully it's not the case that `visitor.Visit` is closing over a copy of `visitor` or something.

internal/ast/deepclone.go
```diff
@@ -0,0 +1,42 @@
...
+		Factory: f,
+		Hooks: NodeVisitorHooks{
+			VisitNodes: func(nodes *NodeList, v *NodeVisitor) *NodeList {
+				if nodes == nil {
+					return nil
+				}
+				// force update empty lists
+				if len(nodes.Nodes) == 0 {
```
jakebailey, line 13: I seem to remember having to fiddle with the logic for cloning NodeLists in the old compiler to the point where it's not this simple, so hopefully this turns out to be sufficient 😄

internal/ast/ast.go
weswigham: This way around is a bit nicer to use, and avoids needing to write either `f.CloneFunctionDeclaration(node)` or a massive switch inside a single `f.Clone(node)` and is less error-prone besides - you can't pass the wrong node kind to this like you can a hypothetical `f.CloneFunctionDeclaration`, and the implementation still lives next to the constructor/visitor implementations for each node kind, so we're less likely to forget to add it.

jakebailey: That's true, though you do have to write `node.AsFunctionDeclaration().Clone(f)` if you don't have the right pointer already, but that doesn't change with flipping it, really.

jakebailey:
Besides the visitor `&` thing, I think this seems good.

rbuckton:
I've actually hoped we could do away with `cloneNode` since in Strada we almost always use it to clone an `Identifier` or `StringLiteral`. Cloning `StringLiteral` can be replaced with `emitContext.NewStringLiteralFromNode()` which will create a string literal that sources its text from the original literal. And while cloning identifiers (especially generated identifiers) has its uses, cloning more complex nodes is fairly rare and often better served by either using `New*`+`emitContext.SetOriginal` or `Update*`.

weswigham:
>I've actually hoped we could do away with cloneNode since in Strada we almost always use it to clone an Identifier or StringLiteral

In normal emit where it's always a transform doing an explicit pass over all the nodes it cares about, sure, but in declaration emit, we serialize a type to an AST, cache it, then reuse it in potentially many places (and for some API consumers may even need to set parents, like for quickfixes) to ameliorate the extreme cost associated with symbol chain lookup within those types - we have to copy it for each location it goes so those emit nodes aren't spuriously reused for each location. If I could just use `update` I would, but this (the deep clone) is basically update-an-arbitrary-unknown-node-with-new-leaf-nodes as-is. (Realistically, the deep clone is moreso an example for testing/for quickfixes - declaration emit only ever actually uses a shallow clone on nodes it knows it needs a new emit node for and a custom visitor that tries to minimize the cloning done)

PR #396, opened 2025-02-28
Add symlink support to vfstest package

jakebailey:
#382 added symlink support to the code we use outside of testing, but I didn't add it to our test FS since that's a bigger change. This PR is the bigger change.

This is all thankfully possible without [CL 385534](https://go.dev/cl/385534). But, I noticed some problems with that change that I'll probably want to report upstream.

### DIFF
--- a/internal/vfs/vfstest/vfstest.go
+++ b/internal/vfs/vfstest/vfstest.go
@@ -234,8 +234,8 @@ func (m *mapFS) setEntry(realpath string, canonical canonicalPath, file fstest.M
 	}
 }
 
-func splitPath(s string, sep byte, offset int) (before, after string) {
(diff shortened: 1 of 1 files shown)


internal/execute/testsys_test.go
```diff
@@ -90,21 +92,25 @@ func (s *testSys) baselineFSwithDiff(baseline io.Writer) {
...
+		}
+
+		if !d.Type().IsRegular() {
 			return nil
 		}
 
 		newContents, ok := s.FS().ReadFile(path)
 		if !ok {
-			return e
+			return nil
```
jakebailey, line 105: Returning `e` here was never quite right; it should always just return the error at the top, or return different errors.

PR #397, opened 2025-02-28
Use tspath.GetBaseFileName instead of path.Base

jakebailey:
This matches the original code, but mainly no package (outside of those implementing `fs.FS`) should be importing anything from `path`. Probably leftover and I missed it when making similar changes previously.


jakebailey:
Also changed `GetDeclarationFileExtension`.

PR #398, opened 2025-03-01
More type checking

ahejlsberg:
In this PR:

* Decorator type checking.
* Unused identifier checking.
* Minimal JSDoc checking (process `@link` directives for unused identifier checks).

With this PR, the VS Code project now compiles with no errors or warnings (and the fact that no unused identifiers are reported is indication that everything is being checked). Beyond JSDoc and JSX, very little remains to be ported in the type checker and performance should now be representative of what we'll see in the completed port.

### DIFF
--- a/internal/binder/binder.go
+++ b/internal/binder/binder.go
@@ -40,7 +40,7 @@ type Binder struct {
 	file                    *ast.SourceFile
 	options                 *core.CompilerOptions
 	languageVersion         core.ScriptTarget
-	bind                    func(*ast.Node) bool
+	bindFunc                func(*ast.Node) bool
 	unreachableFlow         ast.FlowNode
 	reportedUnreachableFlow ast.FlowNode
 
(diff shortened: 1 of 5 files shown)


internal/checker/checker.go
```diff
@@ -2193,6 +2226,61 @@ func (c *Checker) checkDeferredNode(node *ast.Node) {
...
+
+func (c *Checker) checkJSDocComments(node *ast.Node, location *ast.Node) {
+	for _, comment := range node.Comments() {
+		c.checkJSDocComment(comment, location)
+	}
+}
+
+func (c *Checker) checkJSDocComment(node *ast.Node, location *ast.Node) {
+	switch node.Kind {
+	case ast.KindJSDocLink, ast.KindJSDocLinkCode, ast.KindJSDocLinkPlain:
```
jakebailey, line 2255: Agh, missed that this doesn't check `@see`, so won't catch everything.

jakebailey:
> With this PR, the VS Code project now compiles with no errors or warnings

I went to retest vscode and got a couple of errors still; maybe just that my version is newer? (Or older?)

internal/checker/checker.go
ahejlsberg: `@see` is a tag, so it is indeed checked (to link to a symbol in a `@see` tag you actually have to write `@see {@link Foo}`).

jakebailey: `@see variable` is valid without any `@link`: https://github.com/microsoft/TypeScript/issues/35524#:~:text=compatible%20doc%20widget.-,Examples,-import%20%7B%20Blah

https://www.typescriptlang.org/play/?ts=4.1.0-beta#example/see-in-jsdoc

ahejlsberg: Right, but the old compiler doesn't actually mark `Foo` as referenced because of an `@see Foo` tag. Maybe it should, but neither the old nor the new compiler has logic for that right now.

jakebailey: Wow, weird.

PR #403, opened 2025-03-03
Speed up SetParentInChildren by reusing closures

jakebailey:
If we shift things around, we can reuse `SetParentInChildren` functions more than once, by not closing over the first parent and instead using the first call as the parent. Putting these in a pool now makes this call allocation-free.

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -2,6 +2,7 @@ package ast
 
 import (
 	"slices"
+	"sync"
 	"sync/atomic"
 
 	"github.com/microsoft/typescript-go/internal/core"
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
I keep wondering if we'd just be better off with `ForEachChildWith` that passes around some context.

jakebailey:
> I keep wondering if we'd just be better off with `ForEachChildWith` that passes around some context.

The problem is that methods can't be generic, so it'd have to be `any`. Which, maybe that's fine, but it is an ergonomic problem.

internal/ast/utilities.go
```diff
@@ -842,17 +843,38 @@ func GetSourceFileOfNode(node *Node) *SourceFile {
...
-		parent = saveParent
+var setParentInChildrenPool = sync.Pool{
+	New: func() any {
+		return newParentInChildrenSetter()
+	},
+}
+
+func newParentInChildrenSetter() func(node *Node) bool {
+	// Consolidate state into one allocation.
+	// Similar to https://go.dev/cl/552375.
```
jakebailey, line 854: Removing this trick for the more obvious code nets:

```
goos: linux
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/binder
                                                     │   new.txt   │              new2.txt              │
Bind/empty.ts-20                                       413.7n ± 5%   406.4n ± 4%       ~ (p=0.159 n=10)
Bind/checker.ts-20                                     17.88m ± 1%   18.93m ± 5%  +5.91% (p=0.004 n=10)
Bind/dom.generated.d.ts-20                             8.021m ± 3%   8.377m ± 4%  +4.44% (p=0.009 n=10)
Bind/Herebyfile.mjs-20                                 321.5µ ± 3%   323.6µ ± 5%       ~ (p=0.853 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20   159.0µ ± 3%   165.3µ ± 4%  +3.95% (p=0.015 n=10)
geomean                                                313.6µ        321.8µ       +2.60%

                                                     │   new.txt    │               new2.txt                │
Bind/empty.ts-20                                         432.0 ± 0%     432.0 ± 0%       ~ (p=1.000 n=10) ¹
Bind/checker.ts-20                                     7.206Mi ± 0%   7.206Mi ± 0%       ~ (p=0.220 n=10)
Bind/dom.generated.d.ts-20                             4.873Mi ± 0%   4.873Mi ± 0%       ~ (p=0.551 n=10)
Bind/Herebyfile.mjs-20                                 174.2Ki ± 0%   174.2Ki ± 0%       ~ (p=1.000 n=10) ¹
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20   114.8Ki ± 0%   114.8Ki ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                198.8Ki        198.8Ki       -0.00%
¹ all samples are equal

                                                     │   new.txt   │               new2.txt               │
Bind/empty.ts-20                                        3.000 ± 0%    3.000 ± 0%       ~ (p=1.000 n=10) ¹
Bind/checker.ts-20                                     13.69k ± 0%   13.69k ± 0%       ~ (p=1.000 n=10) ¹
Bind/dom.generated.d.ts-20                             14.68k ± 0%   14.68k ± 0%       ~ (p=1.000 n=10) ¹
Bind/Herebyfile.mjs-20                                  344.0 ± 0%    344.0 ± 0%       ~ (p=1.000 n=10) ¹
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20    280.0 ± 0%    280.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                 565.9         565.9       +0.00%
¹ all samples are equal
```

So this trick does work, seemingly.

PR #404, opened 2025-03-03
Reduce allocations in various parts of the checker

jakebailey:
The first change removes 4% of the allocations in a test run; we know ahead of time roughly how many globals there will be.

The other three changes overall reduce the allocations during a vscode compile by 8%.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -18744,13 +18744,7 @@ func (c *Checker) instantiateSymbolTable(symbols ast.SymbolTable, m *TypeMapper,
 	if len(symbols) == 0 {
 		return nil
 	}
-	resultCount := 0
-	for id, symbol := range symbols {
-		if c.isNamedMember(symbol, id) {
-			resultCount++
-		}
-	}
-	result := make(ast.SymbolTable, resultCount)
+	result := make(ast.SymbolTable, len(symbols))
 	for id, symbol := range symbols {
 		if c.isNamedMember(symbol, id) {
 			if mappingThisOnly && isThisless(symbol) {


cmd/tsgo/main.go
```diff
@@ -248,6 +249,7 @@ func main() {
 	}
 	stats.add("Total time", totalTime)
 	stats.add("Memory used", fmt.Sprintf("%vK", memStats.Alloc/1024))
+	stats.add("Memory allocs", strconv.FormatUint(memStats.Mallocs, 10))
```
jakebailey, line 252: I can drop this, but it was useful when testing via real projects.

internal/checker/checker.go
```diff
@@ -1012,6 +1012,16 @@ func createFileIndexMap(files []*ast.SourceFile) map[*ast.SourceFile]int {
 	return result
 }
 
+func countLibDtsSymbols(files []*ast.SourceFile) int {
+	count := 0
+	for _, file := range files {
+		if file.IsDeclarationFile && strings.HasPrefix(tspath.GetBaseFileName(file.FileName()), "lib.") {
```
jakebailey, line 1018: This is not exact, but it really doesn't have to be; mainly this is about making it faster to run tests where we're constructing tens of thousands of checkers.

ahejlsberg: Wouldn't it make more sense to count the number of locals in script files (i.e. files for which `ast.isExternalOrCommonJsModule` returns false)? Those are the ones that will actually be merged into the global symbol table.

internal/checker/checker.go
```diff
@@ -18734,7 +18744,13 @@ func (c *Checker) instantiateSymbolTable(symbols ast.SymbolTable, m *TypeMapper,
...
 		return nil
 	}
-	result := make(ast.SymbolTable)
+	resultCount := 0
+	for id, symbol := range symbols {
+		if c.isNamedMember(symbol, id) {
+			resultCount++
+		}
+	}
+	result := make(ast.SymbolTable, resultCount)
```
ahejlsberg: Maybe just use `len(symbols)` as the initial size and forego the counting loop? That may slightly overestimate the size, but at most by a few since all we eliminate here are `InternalSymbolNameCall`, `InternalSymbolNameNew`, `InternalSymbolNameIndex`, and the like.

internal/checker/checker.go
jakebailey: Tried this out, and it does seem to have a similar number of allocations, so that seems like a good change. I didn't realize that info had been fully ported.

internal/checker/checker.go
jakebailey: Yeah, I did that but wasn't sure how much of an overestimate it'd be. Testing, it doesn't seem to change much.

internal/checker/checker.go
```diff
@@ -23367,7 +23377,7 @@ func (c *Checker) mapTypeEx(t *Type, f func(*Type) *Type, noReductions bool) *Ty
 		}
 	}
 	if changed {
-		return c.getUnionTypeEx(mappedTypes, core.IfElse(noReductions, UnionReductionNone, UnionReductionLiteral), nil /*alias*/, nil /*origin*/)
+		return c.getUnionTypeEx(slices.Clone(mappedTypes), core.IfElse(noReductions, UnionReductionNone, UnionReductionLiteral), nil /*alias*/, nil /*origin*/)
```
jakebailey, line 23380: FWIW I tried using a pool here to alloc from, and it was very much worse than just using `slices.Clone`.

PR #405, opened 2025-03-03
Reduce runtime/allocations during test run

jakebailey:
These are some of the top offenders for memory allocation and runtime. The collection of these brings the runtime of the runner from 6.5s to 6s on my machine, a 7.7% improvement.

### DIFF
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -672,7 +672,7 @@ func getValueOfOptionString(t *testing.T, option string, value string) tsoptions
 
 func getCommandLineOption(option string) *tsoptions.CommandLineOption {
 	return core.Find(tsoptions.OptionsDeclarations, func(optionDecl *tsoptions.CommandLineOption) bool {
-		return strings.ToLower(optionDecl.Name) == strings.ToLower(option)
+		return strings.EqualFold(optionDecl.Name, option)
 	})
 }
 


internal/testutil/harnessutil/harnessutil.go
```diff
@@ -672,7 +672,7 @@ func getValueOfOptionString(t *testing.T, option string, value string) tsoptions
 
 func getCommandLineOption(option string) *tsoptions.CommandLineOption {
 	return core.Find(tsoptions.OptionsDeclarations, func(optionDecl *tsoptions.CommandLineOption) bool {
-		return strings.ToLower(optionDecl.Name) == strings.ToLower(option)
+		return strings.EqualFold(optionDecl.Name, option)
```
DanielRosenwasser, line 675: Can you please add tests for command line options with ß and İ? (kidding)

internal/testutil/tsbaseline/util.go
```diff
@@ -11,7 +11,6 @@ var (
 	lineDelimiter      = regexp.MustCompile("\r?\n")
 	nonWhitespace      = regexp.MustCompile(`\S`)
 	tsExtension        = regexp.MustCompile(`\.tsx?$`)
-	testPathPrefix     = regexp.MustCompile(`(?:(file:\/{3})|\/)\.(?:ts|lib|src)\/`)
```
DanielRosenwasser, line 14: It's a little bit suspicious that this regex was never rooted, but I guess it was never an issue.

internal/testutil/tsbaseline/util.go
```diff
@@ -21,17 +20,30 @@ var (
...
+var (
+	testPathPrefixReplacer = strings.NewReplacer(
+		"/.ts/", "",
+		"/.lib/", "",
+		"/.src/", "",
+		"file:///./ts/", "file:///",
+		"file:///./lib/", "file:///",
+		"file:///./src/", "file:///",
+	)
+	testPathTrailingReplacerTrailingSeparator = strings.NewReplacer(
```
DanielRosenwasser, line 32: I would probably keep the old logic of asking "if `retainTrailingDirectorySeparator` and this thing returned an empty string, return `"/"`".

jakebailey: That's not what the old code did, though; ignoring the scheme thing (where it always returns `file:///`), it either replaces with `"/"` or `""`.

jakebailey: Oh, you mean as a simplification to have one fewer replacer, oops, yes, absolutely.

jakebailey: Oh, no, that doesn't work; the replacer does multiple replacements, so the result won't be an empty string, but rather another string that now "contains" the empty string, so we have no way to hook into this process.

PR #407, opened 2025-03-04
Use iterator for line starts during emit

jakebailey:
The `ComputeLineStarts` calls in the emitter are the source of 50% of its allocations; but it only ever cares about how many lines were scanned and the position of the last one. If we switch the implementation to an iterator, we don't need to allocate an array for every element.

Running `tsgo` on vscode, this reduces the overall allocations from 39063016 to 29706541, which is 24% of all allocations.

Best viewed without whitespace: https://github.com/microsoft/typescript-go/pull/407/files?w=1

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -314,33 +314,42 @@ func Coalesce[T *U, U any](a T, b T) T {
 
 func ComputeLineStarts(text string) []TextPos {
 	result := make([]TextPos, 0, strings.Count(text, "\n")+1)
-	pos := 0
-	lineStart := 0
-	for pos < len(text) {
-		b := text[pos]
-		if b < 0x7F {
-			pos++
-			switch b {
-			case '\r':
-				if pos < len(text) && text[pos] == '\n' {
-					pos++
+	return slices.AppendSeq(result, ComputeLineStartsSeq(text))
+}
+
+func ComputeLineStartsSeq(text string) iter.Seq[TextPos] {
+	return func(yield func(TextPos) bool) {
+		pos := 0
+		lineStart := 0
+		for pos < len(text) {
+			b := text[pos]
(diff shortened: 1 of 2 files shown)


internal/core/core.go
```diff
@@ -314,33 +314,42 @@ func Coalesce[T *U, U any](a T, b T) T {
...
-	lineStart := 0
-	for pos < len(text) {
-		b := text[pos]
-		if b < 0x7F {
-			pos++
-			switch b {
-			case '\r':
-				if pos < len(text) && text[pos] == '\n' {
-					pos++
+	return slices.AppendSeq(result, ComputeLineStartsSeq(text))
```
jakebailey, line 317: Preallocating and using `AppendSeq` is the performant way to collect an iterator of known length.

internal/printer/textwriter.go
```diff
@@ -86,11 +86,18 @@ func (w *textWriter) RawWrite(s string) {
...
+	for lineStart := range core.ComputeLineStartsSeq(s) {
+		count++
+		lastLineStart = lineStart
+	}
+
+	if count > 1 {
+		w.lineCount += count - 1
 		curLen := w.builder.Len()
-		w.linePos = curLen - len(s) + int(lineStartsOfS[len(lineStartsOfS)-1])
+		w.linePos = curLen - len(s) + int(lastLineStart)
```
DanielRosenwasser, line 100: Why doesn't this code walk backwards looking for a `\r` or `\n` and just stop there? This seems like it was always overkill.

DanielRosenwasser: Oh I guess you need the line count.

DanielRosenwasser: FWIW I don't see anyone who uses the `GetLine()` method.

PR #408, opened 2025-03-04
Remove code for JSDocFunctionType

sandersn:
It's not supported in Corsa so doesn't need to be handled in the printer. Also removed future-looking comments in the binder and ast which reference it.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -1491,7 +1491,7 @@ type (
 	LeftHandSideExpression      = Node // subset of Expression
 	JSDocComment                = Node // JSDocText | JSDocLink | JSDocLinkCode | JSDocLinkPlain;
 	JSDocTag                    = Node // Node with JSDocTagBase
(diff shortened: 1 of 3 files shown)


jakebailey:
Remind me again what `JSDocFunctionType` looked like? Hopefully we're still not talking about removing `/** @type {(x: number) => number)} */` on function declarations?

sandersn:
It is `function (string, string): string` syntax.

But, yes, unrelated to this, `@type` on functions is out because so few usages exist in the wild, and there's no (simple) translation to Typescript. The bulk of usages are on arrow functions that would be contextually-typed without the support.

jakebailey:
Not having `@type` on functions would break basically all JSDoc scripts I've written for our team, so, I would really like that to not be removed. But that's separate from this PR.

sandersn:
I only counted 2 usages in scripts/*.mjs that wouldn't be contextually typed by the type tag. And I figured we could change scripts to *.ts any time now that experimental-strip-types is unflagged in node.

jakebailey:
That's not the only place, and we can't do that given we still support back to Node 14 and the scripts are used by the build.

PR #409, opened 2025-03-04
Remove unused GetEntries FS method

jakebailey:
After #382, this function is no longer needed.


andrewbranch:
Can we do the same for WalkDir?

jakebailey:
WalkDir is still used elsewhere, but only in testing since it's convenient. I think you can for now just ignore / make an implementation of it panic.

PR #412, opened 2025-03-05
Synthetic export for nested namespace

### LINKED ISSUE #309: binder: nested namespaces Y and Z in `namespace X.Y.Z` are not exported

sandersn:
Add a synthetic modifier list and `export` modifier when parsing nested namespaces. This is very similar to the way I'm going to add support for `@type`, `@param` and `@return` next, so I think this is a good test case.

When adding the synthetic node, the parser also adds it to a set of synthetic nodes that is stored on the source file so that getTokenAtPosition can filter only node lists containing synthetic nodes. Unfortunately, the visitor passes in `[]*ast.Node`, which isn't comparable, so I use the first node in the list as the marker for whether the containing NodeList contains any synthetic nodes.

I also added code in the type eraser to skip synthetic nodes -- as marked by NodeFlagsSynthesized, not Pos=-1,-1. That's because I think having a real position will make error reporting easier for JSDoc tags, although it doesn't help for this PR.
 
Fixes #309


weswigham:
> which isn't comparable, so I use the first node in the list as the marker for whether the containing NodeList contains any synthetic nodes.

Shouldn't we add flags back to `NodeList`s? The old compiler had them, and it seems useful to be able to mark an empty list as synthetic.

sandersn:
It would be useful but there's no space. Adding a single bool bumps the size from 32 to 40.

jakebailey:
Adding the bool and bumping to 40B:

```
goos: linux
goarch: amd64
Parse/empty.ts/tsc-20                                          477.9n ± 0%   487.8n ± 1%  +2.07% (p=0.000 n=10)
Parse/empty.ts/server-20                                       479.4n ± 0%   488.8n ± 1%  +1.97% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        41.97m ± 2%   41.91m ± 1%       ~ (p=0.631 n=10)
Parse/checker.ts/server-20                                     43.15m ± 2%   43.41m ± 2%       ~ (p=0.579 n=10)
Parse/dom.generated.d.ts/tsc-20                                13.17m ± 0%   13.04m ± 1%  -0.94% (p=0.043 n=10)
Parse/dom.generated.d.ts/server-20                             20.47m ± 1%   20.37m ± 1%       ~ (p=0.063 n=10)
Parse/Herebyfile.mjs/tsc-20                                    710.8µ ± 1%   700.1µ ± 1%  -1.50% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 710.7µ ± 1%   700.6µ ± 1%  -1.42% (p=0.007 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      251.2µ ± 1%   247.5µ ± 1%  -1.46% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   404.2µ ± 1%   400.1µ ± 1%  -1.03% (p=0.019 n=10)
geomean                                                        596.8µ        595.4µ       -0.24%
```
```
Parse/empty.ts/tsc-20                                            816.0 ± 0%     832.0 ± 0%  +1.96% (p=0.000 n=10)
Parse/empty.ts/server-20                                         816.0 ± 0%     832.0 ± 0%  +1.96% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        24.21Mi ± 0%   24.54Mi ± 0%  +1.36% (p=0.000 n=10)
Parse/checker.ts/server-20                                     24.39Mi ± 0%   24.73Mi ± 0%  +1.38% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                8.639Mi ± 0%   8.765Mi ± 0%  +1.46% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             11.00Mi ± 0%   11.21Mi ± 0%  +1.87% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    427.4Ki ± 0%   433.2Ki ± 0%  +1.36% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 427.6Ki ± 0%   433.2Ki ± 0%  +1.30% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      159.8Ki ± 0%   162.9Ki ± 0%  +1.95% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   236.9Ki ± 0%   241.4Ki ± 0%  +1.92% (p=0.000 n=10)
geomean                                                        439.9Ki        447.1Ki       +1.65%
```
```
                                                             │   old.txt   │               new.txt                │
Parse/empty.ts/tsc-20                                           4.000 ± 0%    4.000 ± 0%       ~ (p=1.000 n=10) ¹
Parse/empty.ts/server-20                                        4.000 ± 0%    4.000 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts/tsc-20                                        28.80k ± 0%   28.82k ± 0%  +0.04% (p=0.000 n=10)
Parse/checker.ts/server-20                                     29.42k ± 0%   29.43k ± 0%  +0.04% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                21.05k ± 0%   21.05k ± 0%  +0.01% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             27.98k ± 0%   27.99k ± 0%  +0.02% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    1.208k ± 0%   1.207k ± 0%       ~ (p=0.650 n=10)
Parse/Herebyfile.mjs/server-20                                 1.208k ± 0%   1.207k ± 0%  -0.08% (p=0.020 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20       346.0 ± 0%    346.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20    612.0 ± 0%    612.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                        1.094k        1.094k       -0.00%
¹ all samples are equal
```

So, it's not tragic.

jakebailey:
Can you update the PR description to match what is in the PR, now that you've changed how it works?

internal/transformers/typeeraser.go
```diff
@@ -18,7 +18,8 @@ func NewTypeEraserTransformer(emitContext *printer.EmitContext, compilerOptions
 
 func (tx *TypeEraserTransformer) visit(node *ast.Node) *ast.Node {
 	// !!! TransformFlags were traditionally used here to skip over subtrees that contain no TypeScript syntax
-	if ast.IsStatement(node) && ast.HasSyntacticModifier(node, ast.ModifierFlagsAmbient) {
+	if ast.IsStatement(node) && ast.HasSyntacticModifier(node, ast.ModifierFlagsAmbient) ||
+		node.Flags&ast.NodeFlagsSynthesized != 0 {
```
rbuckton: This should be more specific. Any `Node` produced by a `NodeFactory` used in a transform will be marked `Synthesized`, so this only works so long as the type eraser always comes first. If we ever need to change the order of transforms then this will elide every changed node.

internal/astnav/tokens.go
```diff
@@ -81,29 +81,31 @@ func getTokenAtPosition(
...
+						return node.Flags&ast.NodeFlagsSynthesized == 0
+					})
+				}
+				index, match := core.BinarySearchUniqueFunc(nodes, position, func(middle int, node *ast.Node) int {
+					cmp := testNode(node)
+					if cmp < 0 {
+						left = node.End()
+					}
+					return cmp
+				})
```
rbuckton: Given that this is the only place `NodeList.Flags` is read, we could drop `NodeList.Flags` if we exit early (e.g., `return 0`) in the callback to `BinarySearchUniqueFunc` when the node is synthesized and then re-run the binary search against a filtered list if the result was synthetic.

internal/transformers/typeeraser.go
sandersn: I think I'll bring back a lookaside table of jsdoc-inserted synthetic nodes. Since this would be the only place it would be checked, it would still be fewer places to update than setting positions to -1,-1.

internal/astnav/tokens.go
sandersn: 👍🏼

internal/transformers/typeeraser.go
sandersn: I pushed a commit with a hasty version of this. I wasn't sure of the best way to let the transformer track the current sourcefile, so I added a property to the type eraser transformer. I'll talk with you later today to figure out how to improve that.

sandersn: Well, the commit didn't make it this morning, and now the fork is messed up because of private-to-public switch. I may need to recreate this PR from a new branch or something.

sandersn:
Continues at #521 now that my fork is recreated as public.

PR #416, opened 2025-03-06
Modify setup-go cache key to work better with merge queues

jakebailey:
Merge queues work by creating a new branch on the repo to run CI on. Unfortunately, the caches are not shared between branches, so CI is always slower there than needed.

But, GHA _can_ restore caches that come from `main`; this normally works, but the cache key I set up for our `setup-go` action was scoped per workflow, so we'd never grab it from say, the main CI build.

Reconfigure this to try and restore caches a bit more.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -28,7 +28,6 @@ runs:
         key: ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs', '**/.custom-gcl.yml') }}-${{ github.workflow }}-${{ inputs.cache-name }}
         restore-keys: |
           ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs', '**/.custom-gcl.yml') }}
-          ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}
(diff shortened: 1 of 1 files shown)


.github/actions/setup-go/action.yml
```diff
@@ -25,7 +25,9 @@ runs:
         # There is more code downloaded and built than is covered by '**/go.sum',
         # so give each job its own cache to try and not end up sharing the wrong
         # cache between jobs.
-        key: ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ github.workflow }}-${{ inputs.cache-name }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs', '**/.custom-gcl.yml') }}
+        key: ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs', '**/.custom-gcl.yml') }}-${{ github.workflow }}-${{ inputs.cache-name }}
+        restore-keys: |
+          ts-setup-go-${{ runner.os }}-${{ steps.install-go.outputs.go-version }}-${{ hashFiles('**/go.sum', '**/Herebyfile.mjs', '**/.custom-gcl.yml') }}
```
jakebailey, line 30: The difference between these two now is that the step is allowed to restore the cache from another workflow.

This is technically "bad" because the space used may increase over time, but hopefully the cache will be invalidated by us changing the hashed files over time. Easy to change later, though, or only enable this sort of cache reuse in the merge queue.

PR #417, opened 2025-03-06
Add CommonJS module transform

rbuckton:
This adds the CommonJS module transform, along with extracting reference resolution from the checker to reduce our dependency on the checker during emit. This also contains some cleanup of `EmitContext` and improves the interaction between `EmitContext`, `NodeVisitor`, and `NodeFactory` so as to ensure we properly and reliably track "original" node pointers.

- `NodeFactory` now accepts a `NodeFactoryHooks` object that provides hookable callbacks for `New*`, `Update*`, and `Clone`. These hooks are used by `EmitContext` to set the `NodeFlagsSynthesized` flag, invoke `SetOriginal`, and clone generated name information, respectively.
- `NodeVisitor` has additional hooks in `NodeVisitorHooks` that can intercept specific edges of a node, such as the embedded `Statement` of an `if` or `with` statement, the per-iteration body of an iteration statement, and the parameter list and body of a function or method. These hooks are used by `EmitContext` to properly emit temporary variables generated during transformation of these edges.
- `EmitContext` now supports requesting emit helpers as part of a transformation.
- `Printer` now supports writing emit helpers.
- `RuntimeSyntaxTransformer` now uses `ReferenceResolver` along with the new `CommonJSTransformer`

### DIFF
--- a/internal/printer/emitcontext.go
+++ b/internal/printer/emitcontext.go
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 
 	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
@@ -22,7 +23,7 @@ type EmitContext struct {
 	emitNodes     core.LinkStore[*ast.Node, emitNode]
 	varScopeStack core.Stack[*varScope]
 	letScopeStack core.Stack[*varScope]
-	emitHelpers   core.Set[*EmitHelper]
+	emitHelpers   collections.OrderedSet[*EmitHelper]
 
 	isCustomPrologue           func(node *ast.Statement) bool
 	isHoistedFunction          func(node *ast.Statement) bool
@@ -789,7 +790,7 @@ func (c *EmitContext) RequestEmitHelper(helper *EmitHelper) {
 }
 
 func (c *EmitContext) ReadEmitHelpers() []*EmitHelper {
-	helpers := slices.Collect(maps.Keys(c.emitHelpers.Keys()))
+	helpers := slices.Collect(c.emitHelpers.Values())
 	c.emitHelpers.Clear()
 	return helpers
 }


internal/core/set.go
```diff
@@ -24,8 +29,8 @@ func (s *Set[T]) Len() int {
 	return len(s.M)
 }
 
-func (s *Set[T]) Keys() map[T]struct{} {
-	return s.M
+func (s *Set[T]) Keys() iter.Seq[T] {
+	return maps.Keys(s.M)
```
jakebailey: This was sort of intentional on Anders' part, I believe, just because it was faster. In Go 1.24, it may be the case that this code is not slow at all, though.

internal/printer/emitresolver.go
```diff
@@ -0,0 +1,15 @@
+package printer
+
+import (
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/binder"
+)
+
+type EmitResolver interface {
```
jakebailey, line 8: Very happy to see this move to `printer`.

internal/stringutil/util.go
```diff
@@ -75,3 +77,60 @@ func IsHexDigit(ch rune) bool {
...
+		}
+		pos++
+	}
+	if start < len(text) {
+		lines = append(lines, text[start:])
+	}
+	return lines
+}
+
+const MAX_SMI_X86 int = 0x3fff_ffff
```
jakebailey: If this is unused elsewhere (which I believe to be true), I would suggest just sticking it as a const local of `GuessIndentation`.

internal/tspath/extension.go
```diff
@@ -80,6 +83,10 @@ func TryExtractTSExtension(fileName string) string {
 	return ""
 }
 
+func HasTSFileExtension(path string) bool {
+	return FileExtensionIsOneOf(path, SupportedTSExtensionsFlat)
```
jakebailey, line 87: IIRC Andrew had mentioned changing all of our extension lists and flattening them anyway (since apparently we always flatten when we use them?)

internal/core/multimap.go
```diff
@@ -0,0 +1,57 @@
+package core
+
+import (
+	"iter"
+	"maps"
+	"slices"
+)
+
+type MultiMap[K comparable, V comparable] struct {
```
jakebailey, line 9: I would have probably stuck this in `collections` but we haven't really been consistent, given other stuff getting added to `core`. Would be nice to break things apart eventually (so, don't worry about leaving it here).

internal/transformers/commonjstransformer.go
jakebailey, line 1: File naming nit, but the package is already "transformers"; this file could probably just be called `commonjs.go`.

internal/compiler/emitter.go
```diff
@@ -44,26 +44,42 @@ func (e *emitter) emitJsFile(sourceFile *ast.SourceFile, jsFilePath string, sour
...
 	}
-	sourceFile = transformers.NewRuntimeSyntaxTransformer(emitContext, options).TransformSourceFile(sourceFile)
+
+	// transform `enum`, `namespace`, and parameter properties
+	sourceFile = transformers.NewRuntimeSyntaxTransformer(emitContext, options, referenceResolver).TransformSourceFile(sourceFile)
+
+	// transform module syntax
+	switch options.ModuleKind {
+	case core.ModuleKindCommonJS:
+		sourceFile = transformers.NewCommonJsTransformer(emitContext, options, referenceResolver).TransformSourceFile(sourceFile)
```
jakebailey, line 81: Is there something else we need to do here to have modules transformed when in other modes? This is looking like it would only work if `module=commonjs`, but not nodenext/node16, etc, based on the detected file kind, right?

rbuckton: Yes, there is. I'm currently working on the ES module and implied module format transformers and will amend this as part of that PR when it is ready.

jakebailey: Great, just verifying that this is just WIP and not unintentional.

jakebailey:
That test failure is interesting; it implies that there's some nondeterminism as the ordering of some import helpers appears to have been swapped, but only in one of the runs.

rbuckton:
The issue is that `core.Set.Keys()` order is nondeterministic and those two helpers have equal priorities. I've switched to using `colletions.OrderedSet` to avoid that.

PR #418, opened 2025-03-06
Error in tsgo on positional args

jakebailey:
Our CLI doesn't do anything with positional args; only the subcommands do.

Eventually, this current debugging CLI will go away and we'll make `tsc` the entrypoint, but for now, error on extra args so people who try out the CLI aren't confused.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -113,7 +113,7 @@ func parseArgs() *cliOptions {
 	flag.Parse()
 
 	if len(flag.Args()) > 0 {
-		fmt.Fprintf(os.Stderr, "Unknown positional arguments %v; try tsgo tsc <args>\n", flag.Args())
+		fmt.Fprintf(os.Stderr, "Unknown positional arguments %v. Current compiler is not identical to tsc but can be partially emulated by running:\n\ntsgo tsc <args>\n", flag.Args())
 		os.Exit(1)
 	}
 


cmd/tsgo/main.go
```diff
@@ -112,6 +112,11 @@ func parseArgs() *cliOptions {
 	flag.StringVar(&opts.devel.pprofDir, "pprofDir", "", "Generate pprof CPU/memory profiles to the given directory.")
 	flag.Parse()
 
+	if len(flag.Args()) > 0 {
+		fmt.Fprintf(os.Stderr, "Unknown arguments: %v\n", flag.Args())
```
iisaduan: Do we want to add an instruction in the error message like "did you mean to use `tsgo tsc <args>`"

jakebailey: How ready is that subcommand? We can do that, potentially, though that would then turn off our extra stats, profiling, etc. Most of the time people are just going to try and add args that aren't supported, or accidentally try and pass paths.

jakebailey: Okay, I updated it. Note that this code is not called when flag parsing fails for other reasons, e.g. unknown flags. So I'm not sure if mentioning it is actually the best move still.

cmd/tsgo/main.go
```diff
@@ -112,6 +112,11 @@ func parseArgs() *cliOptions {
 	flag.StringVar(&opts.devel.pprofDir, "pprofDir", "", "Generate pprof CPU/memory profiles to the given directory.")
 	flag.Parse()
 
+	if len(flag.Args()) > 0 {
+		fmt.Fprintf(os.Stderr, "Unknown positional arguments %v; try tsgo tsc <args>\n", flag.Args())
```
DanielRosenwasser: Nit - the old compiler never uses standard error this way because erroneous input is meant to be a normally handled condition. But everything else in this file uses it, and while I'd argue that should be fixed, it's fine for this PR.

DanielRosenwasser: Or maybe

```suggestion
		fmt.Fprintf(os.Stderr, "Unknown positional arguments %v. Current compiler is not identical to tsc. Instead, try:\n\ntsgo tsc <args>\n", flag.Args())
```

DanielRosenwasser: Or

```suggestion
		fmt.Fprintf(os.Stderr, "Unknown positional arguments %v. Current compiler is not identical to tsc but can be partially emulated by running:\n\ntsgo tsc <args>\n", flag.Args())
```

jakebailey: Unfortunately that doesn't mesh well with any of the standard Go stuff; this is all temporary anyway though.

PR #419, opened 2025-03-06
Fix symlink following behavior for test FS Open

jakebailey:
A misunderstanding of how `Open` and `Stat` worked led to me making them behave incorrectly; `Open`, and therefore `Stat` should follow symlinks; it's `Lstat` which wouldn't. This fixes `FileExists` and `DirectoryExists`.

Fixing this lets me reenable the tests I disabled in #396. I should have listened to them.

### DIFF
--- a/internal/vfs/vfstest/vfstest_test.go
+++ b/internal/vfs/vfstest/vfstest_test.go
@@ -74,6 +74,8 @@ func TestInsensitive(t *testing.T) {
 	_, err = fs.Stat(vfs, "Does/Not/Exist")
 	assert.ErrorContains(t, err, "file does not exist")
 
+	// TODO: TestFS doesn't understand case-insensitive file systems.
+	// This same thing would happen with an os.Dir on Windows.
 	// assert.NilError(t, fstest.TestFS(vfs, "Foo/Bar/Baz"))
 }
 


internal/vfs/vfstest/vfstest_test.go
```diff
@@ -75,7 +74,6 @@ func TestInsensitive(t *testing.T) {
 	_, err = fs.Stat(vfs, "Does/Not/Exist")
 	assert.ErrorContains(t, err, "file does not exist")
 
-	// TODO: reenable in Go 1.25 when TestFS understands symlinks.
 	// assert.NilError(t, fstest.TestFS(vfs, "Foo/Bar/Baz"))
```
gabritto, line 79: Why are some of those asserts still commented out?

jakebailey: This one I mistakenly added a comment onto:  https://github.com/microsoft/typescript-go/pull/396/files#diff-9e8a46fd8ee4009af032a5ceb601a5eefb079b222e3b3a527702675cced403a5R78

This test utility doesn't really understand case insensitive paths.

jakebailey: Replaced the comment with a correct one (which I should have left originally).

PR #423, opened 2025-03-07
Add support for `nodenext`/implied module format

rbuckton:
This adds the ES Module transformer that supports `--module nodenext`, along with support for implied module format, as well as a number of fixes to the CommonJS transformer and `EmitContext`.

This also ports support for extension rewriting for relative paths.

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -1,8 +1,8 @@
 package ast
 
 import (
-	"regexp"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -2416,13 +2416,31 @@ func nodeContainsPosition(node *Node, position int) bool {
 	return node.Kind >= KindFirstNode && node.Pos() <= position && (position < node.End() || position == node.End() && node.Kind == KindEndOfFile)
 }
 
-var importOrRequireRegExp = regexp.MustCompile(`import|require`)
+func findImportOrRequire(text string, start int) (index int, size int) {
+	index = max(start, 0)
+	n := len(text)
+	for index < n {
+		next := strings.IndexAny(text[index:], "ir")
+		if next < 0 {
+			break
+		}
+		index += next
 
-func updateLastIndex(lastIndex int, match []int) int {
-	if match == nil {
-		return -1
+		var expected string
+		if text[index] == 'i' {
+			size = 6
+			expected = "import"
+		} else {
+			size = 7
+			expected = "require"
+		}
+		if index+size <= n && text[index:index+size] == expected {
+			return
+		}
+		index++
 	}
-	return lastIndex + match[0]
+
+	return -1, 0
 }
 
 func ForEachDynamicImportOrRequireCall(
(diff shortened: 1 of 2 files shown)


internal/ast/utilities.go
```diff
@@ -2349,3 +2382,95 @@ func IsAliasSymbolDeclaration(node *Node) bool {
...
+		}
+		current = child
+	}
+}
+
+func nodeContainsPosition(node *Node, position int) bool {
+	return node.Kind >= KindFirstNode && node.Pos() <= position && (position < node.End() || position == node.End() && node.Kind == KindEndOfFile)
+}
+
+var importOrRequireRegExp = regexp.MustCompile(`import|require`)
```
jakebailey: I had thought we had this somewhere else in a new way; regex for this is super inefficient in Go.

jakebailey: I vaguely remember @sandersn doing something like this...

jakebailey: Ah, yeah, `collectDynamicImportOrRequireOrJsDocImportCalls` is one place we already optimized this. (And is getting reverted in this PR.)

I believe we had previously talked about an algorithm that basically just looks for the next `r` or `i`, then checks if the prefix is `require` or `import`.

(Depending on how often we plan to do this sort of thing, we could use `suffixarray`, but that'll incur an unfortunate copy.)

internal/printer/emitcontext.go
```diff
@@ -603,6 +603,24 @@ func (c *EmitContext) NewExportStarHelper(moduleExpression *ast.Expression, expo
 	)
 }
 
+// Allocates a new Call expression to the `__rewriteRelativeImportExtension` helper.
+func (c *EmitContext) NewRewriteRelativeImportExtensionsHelper(firstArgument *ast.Node, preserveJsx bool) *ast.Expression {
```
jakebailey, line 607: Note that the submodule predates this feature, so this will be unused.

I'm not sure why our CompilerOptions struct has the prop listed.

testdata/baselines/reference/submodule/conformance/commonJsImportBindingElementNarrowType.symbols.diff
```diff
@@ -9,11 +9,3 @@
  
  if (a) {
  >a : Symbol(a, Decl(bar.js, 0, 7))
-@@= skipped -10, +8 lines =@@
```
jakebailey, line 12: Good change; I wonder which change fixed this!

internal/ast/utilities.go
rbuckton: Yeah, I needed the same functionality but `collectDynamicImportOrRequireOrJsDocImportCalls` was only a partial port of `ForEachDynamicImportOrRequireCall`. I can update this to not use a regex.

testdata/baselines/reference/submodule/conformance/commonJsImportBindingElementNarrowType.symbols.diff
rbuckton: I imagine it's the switch from `collectDynamicImportOrRequireOrJsDocImportCalls` to `ForEachDynamicImportOrRequireCall` since the former didn't yet implement support for `require("...")`

jakebailey: Yeah, super likely.

internal/ast/utilities.go
jakebailey: Looks great, thanks.

PR #425, opened 2025-03-07
Enable error baselines

jakebailey:
Includes #424, but still has other problems, such as:

- ~`TestCompilerBaselinesSubmodule/compiler/unreachableDeclarations.ts_preserveconstenums=false/error` is nondeterministic.~
  - Our `SourceFileAffecting` list did not consider the options the binder uses.
- In `hereby tests --concurrentTestPrograms`, concurrent checking causes problems for any diagnostics that works cross-file, like duplicate identifiers, mismatching type params.
  - `TestCompilerBaselinesSubmodule/compiler/jsFileCompilationDuplicateFunctionImplementation.ts/error`
  - `TestCompilerBaselinesSubmodule/conformance/duplicateNumericIndexers.ts/error`
- Similarly in that mode, some tests appear to be missing exports from other files.
  - `TestCompilerBaselinesSubmodule/conformance/nodeModulesSynchronousCallErrors.ts_module=node16/error`

### DIFF
--- a/internal/testutil/runner/compiler_runner.go
+++ b/internal/testutil/runner/compiler_runner.go
@@ -159,19 +159,19 @@ func (r *CompilerBaselineRunner) runTest(t *testing.T, filename string) {
 			if config.Name != "" {
 				testName += " " + config.Name
 			}
-			t.Run(testName, func(t *testing.T) { r.runSingleConfigTest(t, test, config) })
+			t.Run(testName, func(t *testing.T) { r.runSingleConfigTest(t, testName, test, config) })
 		}
 	} else {
-		t.Run(basename, func(t *testing.T) { r.runSingleConfigTest(t, test, nil) })
+		t.Run(basename, func(t *testing.T) { r.runSingleConfigTest(t, basename, test, nil) })
 	}
 }
 
-func (r *CompilerBaselineRunner) runSingleConfigTest(t *testing.T, test *compilerFileBasedTest, config *harnessutil.NamedTestConfiguration) {
+func (r *CompilerBaselineRunner) runSingleConfigTest(t *testing.T, testName string, test *compilerFileBasedTest, config *harnessutil.NamedTestConfiguration) {
 	t.Parallel()
 	defer testutil.RecoverAndFail(t, "Panic on compiling test "+test.filename)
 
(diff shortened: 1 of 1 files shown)


jakebailey:
Part of the problem is that when we check files, we can issue errors on files _other_ than those being checked. When we are using multiple checkers, we have to make sure to ask all other checkers for their errors too, that fixed about half of the failures. The rest are some other problem.

jakebailey:
- `recursiveExportAssignmentAndFindAliasedType1` - This one seems like it'll be hard to fix; this is a circularity and so by asking all checkers for their diags, we get multiple locations for the same circularity.
  - Same for `recursiveExportAssignmentAndFindAliasedType2`, `recursiveExportAssignmentAndFindAliasedType3`, `circular1`, `circular2`, 
- `unusedTypeParameters7` - this test merges two declarations with type parameters, with multiple checkers, the checkers don't share info about whether or not the type parameters are used, so in concurrent mode we get an unused type declaration error.
- `superInStaticMembers1(target=es2015)` - I _think_ this is a case where there's a stronger error from having shadowed reserved names, but when we check another referencing file, we instead issue a different relationship error instead, whose text span is wider than the "original" error, and so we omit the one inside it...
- `typeOnlyMerge2`, `typeOnlyMerge3` - No clue; it's like we are not detecting that something is a type import. This test is really odd because somehow there's a type error without a location, even in Strada... So something here is cursed both ways.

jakebailey:
Other than those failures, though, everything does actually seem to be working with concurrent checking.

jakebailey:
After talking with @ahejlsberg, it seems like the first one is inherent and mostly unsolvable; we probably will have to just have a concurrent variant or just skip the test in concurrent mode.

For `unusedTypeParameters7`, the only actual solution would be to share usage checks between checkers, but that seems bad, so we might just disable usage checks for declarations that cross file boundaries.

The others, who knows 😄

jakebailey:
Daniel's 3 PRs reduced the number of differences by some 60k lines. Awesome.

jakebailey:
With main merged in, down another 20k lines of diffs.

jakebailey:
I've updated the PR to skip the tests which fail in concurrent mode for now. This should be ready to go in.

PR #426, opened 2025-03-07
Fix SourceFileAffecting for options used in binder

jakebailey:
The binder uses these options to issue errors and construct flow nodes, so we can't reuse the AST when these change.

This fixes more non-determinism in #425.


andrewbranch:
We should really refactor out `--preserveConstEnums` and `--isolatedModules`. Super frustrating to have those affect source file sharing.

jakebailey:
Yes, absolutely, and I kinda even want to make control flow not depend on the unreachable flags and instead move those errors to the checker / perform conditional analysis there.

PR #427, opened 2025-03-07
Print warning in tsgo on currently unsupported file types

jakebailey:
This is a temporary measure pre-launch; printing this message should help people realize that their react projects may only be compiling "super fast" because we're not actually checking any of the JSX.

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -232,7 +232,7 @@ func main() {
 	var unsupportedExtensions []string
 	for _, file := range program.SourceFiles() {
 		extension := tspath.TryGetExtensionFromPath(file.FileName())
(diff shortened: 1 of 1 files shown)


cmd/tsgo/main.go
```diff
@@ -229,6 +229,17 @@ func main() {
 		printDiagnostics(ts.SortAndDeduplicateDiagnostics(diagnostics), host, compilerOptions)
 	}
 
+	var unsupportedExtensions []string
+	for _, file := range program.SourceFiles() {
+		extension := tspath.TryGetExtensionFromPath(file.FileName())
+		if extension == ".js" || extension == ".jsx" || extension == ".tsx" {
```
gabritto: Should we add `.cjs` and `.mjs` too?

jakebailey: Yes, oops.

PR #428, opened 2025-03-07
set watch interval via command line

iisaduan:
Adds option so you can adjust how long watch will wait between cycles

### DIFF
--- a/testdata/baselines/reference/tsc/commandLine/Parse-watch-interval-option.js
+++ b/testdata/baselines/reference/tsc/commandLine/Parse-watch-interval-option.js
@@ -132,7 +132,7 @@ ParsedCommandLine::{
             "tscBuild": null
         },
         "watchOptions": {
-            "watchInterval": null,
+            "watchInterval": 1000,
             "watchFile": 0,
             "watchDirectory": 0,
             "fallbackPolling": 0,


internal/tsoptions/commandlineparser.go
```diff
@@ -224,8 +224,8 @@ func (p *commandLineParser) parseOptionValue(
 		if args[i] != "null" {
 			switch opt.Kind {
 			case "number":
-				// todo: Make sure this parseInt matches JS parseInt
-				num, e := strconv.ParseInt(args[i], 10, 0)
```
jakebailey, line 228: What was wrong with `ParseInt` here? Probably actually this should use `jsnum` now that it's available.

internal/tsoptions/parsedcommandline.go
```diff
@@ -8,13 +8,13 @@ import (
 )
 
 type ParsedCommandLine struct {
-	ParsedConfig *core.ParsedOptions
+	ParsedConfig *core.ParsedOptions `json:"ParsedConfig"`
```
jakebailey: I would recommend having all of the json prop names be `camelCase` as opposed to `PascalCase`.

internal/tsoptions/commandlineparser.go
iisaduan: `ParseInt` returned an `int64`, and that was causing issues later in tsconfig parsing

internal/core/parsedoptions.go
```diff
@@ -1,9 +1,9 @@
 package core
 
 type ParsedOptions struct {
-	CompilerOptions *CompilerOptions
-	WatchOptions    *WatchOptions
+	CompilerOptions *CompilerOptions `json:"CompilerOptions"`
```
jakebailey: Similarly here on the casing.

internal/tsoptions/commandlineparser.go
iisaduan: Whats the best way to use `jsnum` to get an `int`?

iisaduan: Or should I change the types to `jsnum.Number`?

jakebailey: It might need to be the latter, but I wouldn't bother it in this PR.

PR #432, opened 2025-03-08
Restore error handling/suggestions for keywords in place of missing semicolons

DanielRosenwasser:
Addresses some differences found in #425.

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -1879,6 +1879,14 @@ func GetIdentifierToken(str string) ast.Kind {
 	return ast.KindIdentifier
 }
 
+func AllKeywords() []string {
+	keywords := make([]string, 0, len(textToKeyword))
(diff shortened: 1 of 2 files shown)


internal/scanner/scanner.go
```diff
@@ -1879,6 +1879,14 @@ func GetIdentifierToken(str string) ast.Kind {
 	return ast.KindIdentifier
 }
 
+func AllKeywords() []string {
+	keywords := make([]string, 0, len(textToKeyword))
```
jakebailey, line 1883: This func could just be maps.Keys, or better, a global rather than recalculating each call.

jakebailey: Or, `var AllKeywords = sync.OnceFunc(func() []string { ... })` or something.

jakebailey:
Was done in #641

PR #433, opened 2025-03-08
Adopt Strada-style object type printing.

DanielRosenwasser:
While I actually like the comma-style type printing for error messages, this reduces a ton of the differences found in #245. So in the future, I would like us to consider swapping back to the old behavior, but for now this should help us work through the differences in each codebase.


jakebailey:
Comma-style isn't even what any formatter does by default 😄

PR #434, opened 2025-03-10
More type checking + fixes

ahejlsberg:
In this PR:

* Deprecation warnings.
* Check `meta.*` constructs.
* Check unused renamed binding elements.
* Check regular expression literals.
* Check for `require` and `exports` name collisions.
* Exclude classes/interfaces with parts in multiple source files from unused type parameter checks.
* Fixes uncovered by submodule tests.

### DIFF
--- a/internal/checker/printer.go
+++ b/internal/checker/printer.go
@@ -364,18 +364,18 @@ func (p *Printer) printAnonymousType(t *Type) {
 		}
 	}
 	p.print("{")
-	var tail bool
+	hasMembers := false
 	for _, sig := range callSignatures {
 		p.print(" ")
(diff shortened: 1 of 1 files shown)


jakebailey:
Some of this I think is similar/the same as some of Daniel's open PRs, too.

ahejlsberg:
> Some of this I think is similar/the same as some of Daniel's open PRs, too.

Yes, this has the same fixes to object type printing as #433, but also includes several others.

jakebailey:
Do we actually put a trailing semicolon for objects?

jakebailey:
Ah, it's the same as Daniel's PR so, yeah.

PR #435, opened 2025-03-10
Skip most CI jobs when PR is unbuildable

jakebailey:
Saves some resources when the PR doesn't build (which is especially helpful to not waste our custom runners).


.github/workflows/ci.yml
```diff
@@ -22,16 +22,16 @@ jobs:
...
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
       - uses: ./.github/actions/setup-go
         with:
           cache-name: build
 
-      - run: npm ci
-
-      - run: npx hereby build
+      - run: go test -c -o=/dev/null ./...
```
DanielRosenwasser, line 29: Do we still have something that actually tests `hereby` builds?

jakebailey: No, but this is actually "stronger" since it checks tests. But it won't catch if the herebyfile breaks.

This PR still takes longer to build than I am expecting, probably due to the lack of build caching, so I'm not sure I even want to do this PR anymore.

PR #437, opened 2025-03-10
fileloader skips files w/unsupported extensions

### LINKED ISSUE #363: program: js files are present in program even when `allowJs` is `false`
If we consider test `checkJsFiles6.ts`:
```ts
// @allowJs: false
// @checkJs: true
// @noEmit: true

// @fileName: a.js
var x;
```

In Strada, the resulting program created from this test does not have the file `a.js`, but it does in Corsa.
Found by #275

sandersn:
This makes allowJs compile the right set of files. Also skips other files, like *.map -- probably even *.txt.

I'm not 100% sure this should be in fileloader, but it needs to work for the command line, tsconfig and the harness. It is also close to its Strada location.

Fixes #363


testdata/baselines/reference/submodule/compiler/checkJsFiles6.symbols.diff
```diff
@@ -2,10 +2,3 @@
 +++ new.checkJsFiles6.symbols
 @@= skipped -0, +-1 lines =@@
 -<no content>
-@@= skipped --1, +1 lines =@@
```
jakebailey, line 5: These baseline changes imply that things are not working correctly, unfortunately.

sandersn: I read this as saying "new.checkJsFiles6.symbols is empty, but old.checkJsFiles6.symbols has "`<no content>`". That's a problem with the baseliner as far as I can see.

sandersn: Ah, but it could be that the compiler is doing the wrong thing when there are no files to compile. @andrewbranch @iisaduan do you know how the compiler should behave in this case? Specifically, is fileloader supposed to handle this? Or some earlier component?

If it's command-line parsing or tsconfig parsing, then probably the harness needs to handle it itself.

sandersn: On this branch, if you have allowjs off and nothing but .js files in the files list, there's no indication that you aren't compiling anything. Strada gives an error for each skipped JS file in this case. In Strada, the diagnostic is added just after the code I ported in this PR. I don't see any other errors in fileloader.go, so I don't know how hard it would be add it there.

jakebailey: The test says:

```ts
// @allowJs: false
// @checkJs: true
// @noEmit: true

// @fileName: a.js
var x;
```

allowJS is off, yes, but with our silly test runner rules, `a.js` _is_ in the program by default (all files in the test are).

What the diff is saying is that "the old file didn't exist at all, the new one has these contents".

So, I think the problem with file inclusion has to be somewhere other than fileloader.

sandersn: The lines starting with `+` are preceded by `-`, so I'm pretty sure that the new one no longer has the contents of a.js. In fact it doesn't look to me like it has anything in it. If I copy all the lines that aren't prefixed with `-` by Github, I get:

```diff
--- old.checkJsFiles6.symbols
+++ new.checkJsFiles6.symbols
@@= skipped -0, +-1 lines =@@
-<no content>
```

And that looks to me like the old file has "no content" while the new file has ""
I'm not sure what `skipped -0, +-1 lines` means though.

jakebailey: Oh, uh, I misread the diff. There's definitely some bug here; the symbols file is empty and the diff is empty, but both should just not exist. I guess maybe we don't treat "no files" as a case where we don't generate symbol baselines? But instead write empty?

internal/testutil/tsbaseline/type_symbol_baseline.go
```diff
@@ -128,7 +128,12 @@ func generateBaseline(
 	if result.Len() > 0 {
 		return fmt.Sprintf("//// [%s] ////\r\n\r\n%s%s", header, strings.Join(perfLines, "\n"), result.String())
 	}
-	return result.String()
+	s := result.String()
```
jakebailey: `result.Len() > 0` is checked above, so this could just be `return NoContent`.

internal/compiler/fileloader.go
```diff
@@ -78,8 +81,13 @@ func processAllProgramFiles(
 
 func (p *fileLoader) addRootTasks(files []string, isLib bool) {
 	for _, fileName := range files {
-		absPath := tspath.GetNormalizedAbsolutePath(fileName, p.host.GetCurrentDirectory())
-		p.rootTasks = append(p.rootTasks, &parseTask{normalizedFilePath: absPath, isLib: isLib})
+		if core.Tristate.IsTrue(p.compilerOptions.AllowNonTsExtensions) ||
+			core.Some(core.Flatten(p.supportedExtensions), func(extension string) bool {
+				return strings.HasSuffix(fileName, extension)
+			}) {
```
jakebailey: This is going to repeatedly flatten the extension array; we should just do this up front.

internal/compiler/fileloader.go
```diff
@@ -78,8 +81,13 @@ func processAllProgramFiles(
 
 func (p *fileLoader) addRootTasks(files []string, isLib bool) {
 	for _, fileName := range files {
-		absPath := tspath.GetNormalizedAbsolutePath(fileName, p.host.GetCurrentDirectory())
-		p.rootTasks = append(p.rootTasks, &parseTask{normalizedFilePath: absPath, isLib: isLib})
+		if core.Tristate.IsTrue(p.compilerOptions.AllowNonTsExtensions) ||
+			core.Some(p.supportedExtensions, func(extension string) bool {
+				return strings.HasSuffix(fileName, extension)
+			}) {
```
jakebailey: Sorry, one more modification; this can just be:

```go
slices.Contains(p.supportedExtensions, tspath.TryGetExtensionFromPath(fileName))
```

jakebailey: Probably should do that on `absPath` (not `fileName`), though, to ensure normalization.

testdata/baselines/reference/submodule/compiler/checkJsFiles6.symbols.diff
sandersn: yes, that was it. I changed the baselines to return NoContent where it was empty string before.

jakebailey:
PR looks good but needs a baseline update.

PR #439, opened 2025-03-10
Improve reverse mapping and auto numbering emit for more cases

rbuckton:
This improves the `constantValue` utility function to cover more numeric initializer cases and thus simplify enum emit for initializers like `1 << 0`.

### DIFF
--- a/internal/transformers/utilities.go
+++ b/internal/transformers/utilities.go
@@ -206,10 +206,14 @@ func isIdentifierReference(name *ast.IdentifierNode, parent *ast.Node) bool {
 
 func constantValue(node *ast.Expression) any {
 	node = ast.SkipOuterExpressions(node, ast.OEKAll)
-	if ast.IsStringLiteralLike(node) {
+	switch {
+	case ast.IsStringLiteralLike(node):
 		return node.Text()
-	}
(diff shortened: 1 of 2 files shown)


internal/transformers/utilities.go
```diff
@@ -206,10 +206,14 @@ func isIdentifierReference(name *ast.IdentifierNode, parent *ast.Node) bool {
 
 func constantValue(node *ast.Expression) any {
```
jakebailey, line 207: What differentiates this function to `createEvaluator` (which is the same as the one in strada)? Or is there an equivalent to `constantValue` in strada under another name?

rbuckton: I believe I mentioned this in the original PR, but `createEvaluator` relies on type information that the emitter/printer doesn't have access to, and as a result covers far more cases that result in type directed emit. I initially looked into pulling `createEvaluator` out of checker, but that resulted in a fair amount of unusable code paths for the emitter so it was simpler just to pull out the small bit of functionality that was necessary for the printer.

jakebailey: Ah, right, gotcha.

PR #449, opened 2025-03-10
Fix default es6 lib file

jakebailey:
This was just a silly typo.

Removes literally 113k diffs.


jakebailey:
The diffs here are not working properly when they get deleted; I am working on fixing that in a separate PR but this PR is so huge there's no way it would be reviewable.

PR #450, opened 2025-03-10
Fix trailing comma detection after transformation

rbuckton:
It turns out the logic to detect trailing commas in `NodeList.HasTrailingComma()` is not sufficient when printing transformed nodes as some nodes may have been elided. For example, the expression `[a!]` will be transformed to `[a]`, eliding the `NonNullExpression` node. As a result `HasTrailingComma()` will detect a one-character gap from the missing `!` and believe the `NodeList` has a trailing comma.

Our options to address this are to either add a property to `NodeList` indicating whether the list contains a trailing comma, similar to Strada, or to attempt to find the original `NodeList` prior to transformation and use that.

Given previous reticence to inflate the size of every `NodeList` with an extra property, this PR opts to use the latter approach. Rather than trusting `children.HasTrailingComma()`, we instead compute the original node of the list's parent, find the matching property on the original parent, and use the result of that list's `HasTrailingComma()`.

### DIFF
--- a/internal/printer/printer_test.go
+++ b/internal/printer/printer_test.go
@@ -2389,7 +2389,7 @@ func TestNoTrailingCommaAfterTransform(t *testing.T) {
 	emitContext := printer.NewEmitContext()
 
 	var visitor *ast.NodeVisitor
-	visitor = emitContext.NewNodeVisitor(func (node *ast.Node) *ast.Node {
+	visitor = emitContext.NewNodeVisitor(func(node *ast.Node) *ast.Node {
 		switch node.Kind {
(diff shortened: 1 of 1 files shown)


jakebailey:
> Given previous reticence to inflate the size of every `NodeList` with an extra property

FWIW I'm pretty sure we have the space now, or will after Nathan's PR?

rbuckton:
> I'm pretty sure we have the space now, or will after Nathan's PR?

There are currently no bits left in `NodeFlags` for another flag.

jakebailey:
Right, I think I was thinking of a new field. But this PR is good if you're happy with it.

PR #452, opened 2025-03-10
Fix deletion of empty diff baselines

jakebailey:
The changes I suggested for #437 were not complete; this PR fixes diffs when they are supposed to become empty and deleted.

### DIFF
--- a/testdata/baselines/reference/submodule/compiler/arrayConcatMap.errors.txt
+++ b/testdata/baselines/reference/submodule/compiler/arrayConcatMap.errors.txt
@@ -1,1 +1,0 @@
-<no content>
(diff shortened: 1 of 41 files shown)


internal/testutil/baseline/baseline.go
```diff
@@ -119,19 +104,25 @@ func writeComparison(t *testing.T, actual string, relativeFileName string, useSu
...
-		}
+
+	expected := NoContent
+	foundExpected := false
+	if content, err := os.ReadFile(referenceFileName); err == nil {
+		expected = string(content)
+		foundExpected = true
+	}
+
+	if expected != actual || actual == NoContent && foundExpected {
```
jakebailey, line 115: This was all modified to be more like the original `writeComparison`.

jakebailey:
~eek this is wrong for other reasons, will figure it out~

Sorry, I tested this on the wrong branch. PR is working 😅

Issue #468, opened 2025-03-11
Double Call Garbage Collector


fuldaxxx:
What is the reason for using a double garbage collection call? As for me, it is simply redundant and brings no benefit. Or am I missing something?

cmd/tsgo/main.go:230
```go
	var memStats runtime.MemStats
	runtime.GC()
	runtime.GC()
	runtime.ReadMemStats(&memStats)
```

jakebailey:
See: https://github.com/microsoft/typescript-go/pull/470#issuecomment-2715366540

PR #469, opened 2025-03-11
removed the unnecessary garbage collection call

fuldaxxx:
I noticed an unnecessary garbage collection call. In this pull request, I am removing it.

https://github.com/microsoft/typescript-go/issues/468


jakebailey:
See: https://github.com/microsoft/typescript-go/pull/470#issuecomment-2715366540

PR #470, opened 2025-03-11
fix: remove double runtime.GC call

fernandoonrails:
I realized reading the code that in `cmd/tsgo/main.go` on line 230, that the function `runtime.GC()` is called twice, this commit will remove that redundancy

if it's intended to be called twice, would be glad with the reasoning for it to be done

### DIFF
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -227,7 +227,6 @@ func main() {
 
 	var memStats runtime.MemStats
 	runtime.GC()
-	runtime.GC()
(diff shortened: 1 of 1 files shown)


jakebailey:
This is intentional; `runtime.GC` does not do a full GC on the call. Two calls are required. I've seen this pretty often in Go's own tests: https://github.com/golang/go/blob/598df45fceb6e10d643ed0c07a3df80cffd507f1/test/heapsampling.go#L228

PR #479, opened 2025-03-11
Support emit for parameter property initializers

rbuckton:
This ports emit for parameter property initializers from Strada. Since parameter properties are handled by `RuntimeSyntaxTransformer`, we no longer elide visibility modifiers on constructor parameters in `TypeEraserTransformer`.


jakebailey (approved):

jakebailey:
Ah, oops, tests are still failing, forgot to check before reviewing.

PR #485, opened 2025-03-11
Remove $ from README.md getting started commands to fix copy/paste

jonstelly:
Minor annoyance - the copy/paste of the clone commands were copying the `$ ` so they didn't work


jonstelly:

jakebailey:
This is intentional; the `console` language is intended for this look and it's pretty standard for console command examples.

I don't think this is an improvement, and contains other random formatting errors. I don't think we should make these changes, sorry.

Issue #476, opened 2025-03-11
`tsgo tsc` treats `tsconfig.json` path with `-p`/`--project` flag as a directory


MichaelMitchell-at:
```
$ tsgo tsc --project tsconfig.json
error TS5081: Cannot find a tsconfig.json file at the current directory: /Users/michael.mitchell/tsgo_repro/tsconfig.json/tsconfig.json
```

mjnk9xw:
I have the same error

jakebailey:
To be clear, you're saying your `tsconfig.json` is a _directory_? What is the full repro?

MichaelMitchell-at:
> To be clear, you're saying your `tsconfig.json` is a _directory_? What is the full repro?

No, it's a file. But the error message indicates that it thinks the path is to a directory `[...]/tsconfig.json/tsconfig.json`.

Repro steps:
```sh
touch tsconfig.json
tsgo tsc --project tsconfig.json
```

jakebailey:
Gotcha; I understand then, we're just not checking if it's not a directory first.

mjnk9xw:
![Image](https://github.com/user-attachments/assets/d369a372-b927-4e98-9d91-254787e799c1)

I think it's broken at line 52 should be fixed to fileOrDirectory != "" && sys.FS().DirectoryExists(fileOrDirectory)

jakebailey:
Does plain `tsgo` also do this?

MichaelMitchell-at:
> Does plain `tsgo` also do this?

No, just `tsgo tsc`

PR #486, opened 2025-03-11
fix `-p` erroring when directly passed a config file

iisaduan:
fixes https://github.com/microsoft/typescript-go/issues/476

This line was ported incorrectly 
https://github.com/microsoft/TypeScript/blob/0aac72020ee8414218273f654eb7ce1dc2dd0d6b/src/compiler/executeCommandLine.ts#L603


internal/execute/tsc.go
```diff
@@ -49,7 +49,7 @@ func executeCommandLineWorker(sys System, cb cbType, commandLine *tsoptions.Pars
 		}
 
 		fileOrDirectory := tspath.NormalizePath(commandLine.CompilerOptions().Project)
-		if fileOrDirectory != "" || sys.FS().DirectoryExists(fileOrDirectory) {
+		if sys.FS().DirectoryExists(fileOrDirectory) {
```
iisaduan, line 52: This line was ported incorrectly 
https://github.com/microsoft/TypeScript/blob/0aac72020ee8414218273f654eb7ce1dc2dd0d6b/src/compiler/executeCommandLine.ts#L603

jakebailey (approved):

PR #496, opened 2025-03-12
chore: nits

laciferin2024:


laciferin2024:

jakebailey:
I'm not sure I understand the change; the change is mostly naming changes, and a change to stderr output which should realistically always succeed (and we don't check it elsewhere).

laciferin2024:
@jakebailey what if stderr is closed or unavailable?

jakebailey:
I guess, but, the intent is that this error eventually be removed in favor of supporting the other LSP flags, we're just early in the development of LSP.

05TIME:
https://x.com/time65515358782?s=21

jakebailey:
Thanks for this, but I don't think we're looking for this kind of change at this time; this code is temporary.

PR #497, opened 2025-03-12
Improve error handling and logging for profiling in main.go

limichange:


jakebailey:
I can totally understand this, but this is just a debugging command and a failure here is unlikely to happen in reality; I think we don't really need to do this at the moment but when this code is moved to its final place, we can handle it there (which will be quite different).

PR #510, opened 2025-03-12
Process directive in multi-line comments

### LINKED ISSUE #495: block-style directive comments don't work
This errors with `Type 'number' is not assignable to type 'string'`

```ts
/* @ts-ignore */
const s: string = 5
```

haoqixu:
Fixes #495


haoqixu:

testdata/baselines/reference/submodule/conformance/multiline.errors.txt.diff
```diff
@@ -1,48 +1,40 @@
...
      
          {/*
         @ts-expect-error */}
--    ~~~~~~~~~~~~~~~~~~~~~~
--!!! error TS2578: Unused '@ts-expect-error' directive.
-         <MyComponent foo={"hooray"} />
-       </div>
-     );
++    ~~~~~~~~~~~~~~~~~~~~~~
++!!! error TS2578: Unused '@ts-expect-error' directive.
```
jakebailey, line 28: This diff indicates that this behavior doesn't match the old compiler, unfortunately.

haoqixu: I think this is because `tsgo` cannot find out the semantic error in the following `foo={100}` in `jsx` yet. :thinking:

jakebailey: Oh that's very possible, yes.

jakebailey (approved):
Thank you!

PR #511, opened 2025-03-12
Enable virtual terminal processing on Windows

### LINKED ISSUE #480: Error messages don't seem coloured as expected on Windows
image

Windows would require the console to be put into virtual terminal mode so it processes ANSI escape sequences.

Sainan:
Before:
image

After:
image

Fixes #480

(I've never used Go before and I didn't verify this works on MacOS or Linux.)

### DIFF
--- a/cmd/tsgo/enablevtprocessing_windows.go
+++ b/cmd/tsgo/enablevtprocessing_windows.go
@@ -1,5 +1,3 @@
-//go:build windows
-
 package main
 
 import (
--- a/cmd/tsgo/enablevtprocessing_windows.go
+++ b/cmd/tsgo/enablevtprocessing_windows.go
@@ -1,0 +1,18 @@
+//go:build windows
+
+package main
+
+import (
+	"golang.org/x/sys/windows"
+)
+
+func enableVirtualTerminalProcessing() {
+	hStdout, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
+	if err == nil && hStdout != windows.InvalidHandle {
+		var mode uint32
+		err = windows.GetConsoleMode(windows.Handle(hStdout), &mode)
+		if err == nil {
+			windows.SetConsoleMode(windows.Handle(hStdout), mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+		}
+	}
+}
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -135,7 +135,7 @@ func enableVirtualTerminalProcessing() {
 
 func main() {
 	if runtime.GOOS == "windows" {
-		// TypeScript uses ANSI escape sequences which cmd.exe won't parse without enabling virtual terminal mode.
+		// TypeScript uses ANSI escape sequences which cmd.exe won't parse without enabling virtual terminal processing.
 		enableVirtualTerminalProcessing()
 	}
 
--- a/cmd/tsgo/main.go
+++ b/cmd/tsgo/main.go
@@ -24,8 +24,6 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 	"github.com/microsoft/typescript-go/internal/tspath"
 	"github.com/microsoft/typescript-go/internal/vfs/osvfs"
-
-	"golang.org/x/sys/windows"
 )
 
 func printDiagnostic(d *ast.Diagnostic, level int, comparePathOptions tspath.ComparePathsOptions) {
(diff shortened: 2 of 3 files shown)


cmd/tsgo/main.go
```diff
@@ -120,7 +123,23 @@ func parseArgs() *cliOptions {
 	return opts
 }
 
+func enableVirtualTerminalProcessing() {
+	hStdout, err := syscall.GetStdHandle(syscall.STD_OUTPUT_HANDLE)
```
Kamandlou: As of Go 1.17+, the syscall package is considered deprecated for many use cases in favor of `golang.org/x/sys/windows`. Your code already imports `golang.org/x/sys/windows`, so you can use this instead:
```go
hStdout, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
```

jakebailey:
I suspect that this code needs to be locked behind a build tag. It's illegal to use the windows syscall package on the wrong platform, even if conditional.

jakebailey:
For an example of how to do it, try looking at `realpath_windows.go` and `realpath_other.go`.

Sainan:
I was trying to avoid doing that because it would introduce new files, but done that now. Let me know if I should move or rename them.

cmd/tsgo/enablevtprocessing_windows.go
```diff
@@ -0,0 +1,18 @@
+//go:build windows
```
jakebailey: No need for this line; it's implied by the filename.

cmd/tsgo/enablevtprocessing_windows.go
```diff
@@ -0,0 +1,18 @@
...
+	"golang.org/x/sys/windows"
+)
+
+func enableVirtualTerminalProcessing() {
+	hStdout, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
+	if err == nil && hStdout != windows.InvalidHandle {
+		var mode uint32
+		err = windows.GetConsoleMode(windows.Handle(hStdout), &mode)
+		if err == nil {
+			windows.SetConsoleMode(windows.Handle(hStdout), mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
```
jakebailey, line 13: Does Node do this by default? How was this working before?

Sainan: Yeah, executing `console.log("\x1B[38;2;255;0;0mTesting");` via Node shows that it works, so I assume Node just does that by default.

jakebailey: I don't see them doing in their code at all; the constant is defined but seeming unused...

Sainan: It seems to be used here: https://github.com/nodejs/node/blob/59f00d713299bfe7d18d15c4504ac14bd5b428a8/deps/uv/src/win/tty.c#L2312

jakebailey: Aha, thanks!

jakebailey: Alright, so checking this, `uv__determine_vterm_state` is only called when `uv_tty_init` is called; I get the feeling that Node only calls this when it knows the output is a TTY, via https://github.com/nodejs/node/blob/d615a3cfcbc357ccebd07a37fa0002fee162d9ce/src/tty_wrap.cc#L152, But I don't see this used either!

I wonder if there's just a plain package somewhere that sets `ENABLE_VIRTUAL_TERMINAL_PROCESSING`...

Sainan: Well, as you've already seen with trying to locate `ENABLE_VIRTUAL_TERMINAL_PROCESSING`, Github's search results are not always conclusive.

I don't really know Node internals, but I can guarantee you that `ENABLE_VIRTUAL_TERMINAL_PROCESSING` does need to be set for ANSI escape sequences to work on Windows. (Although apparently not on Windows 11)

jakebailey: To be clear, I think this PR is right, at least how to enable it, but I'm just trying to verify we won't break something like LSP or piping where the output is not a TTY.

Given https://github.com/jedib0t/go-pretty/blob/b14745cbcf545d2704cdd6f97163344b0fe683a3/text/ansi_windows.go#L22 looks similar and does not appear to be conditional at all, so I very much suspect this PR is good as-is.

jakebailey: I checked, the LSP is totally fine. I'm going to approve this 😄

jakebailey (approved):

jakebailey:
Once again, thanks for sending this. I'm happy to see that your first time using Go was successful.

Sainan:
It's easier than ever to pick up a new language in the age of AI. ^^

PR #515, opened 2025-03-12
Improve jsnum.Remainder Function for ECMAScript Compliance

parthokr:
This pull request addresses incompleteness and inaccuracies in the `jsnum.Remainder` function, bringing it closer to full compliance with the ECMAScript specification (ECMA-262). The existing implementation shows incorrect behavior in certain edge cases, particularly those involving negative numbers as dividend.


parthokr:

internal/jsnum/jsnum.go
```diff
@@ -138,9 +142,15 @@ func (n Number) Remainder(d Number) Number {
...
-	if r == 0 || n < 0 {
-		return negativeZero
+	// Calculate the quotient of the division.
+	quotient := math.Trunc(float64(n / d))
+
+	// Calculate the remainder using the quotient.
+	// This directly implements the JS remainder definition:  r = n - (d * q)
+	r := n - Number(float64(d)*quotient)
+
+	if r == 0 {
```
jakebailey: I don't understand, isn't this new code semantically identical, just rephrased?

internal/jsnum/jsnum_test.go
```diff
@@ -286,12 +316,27 @@ func TestRemainder(t *testing.T) {
...
+		// edge cases with small numbers
+		{1e-10, 1, 1e-10},
+		{1, 1e-10, 0}, // should be 0, or very close to 0
 	}
 
 	for _, test := range tests {
 		t.Run(fmt.Sprintf("%v %% %v", test.x, test.y), func(t *testing.T) {
 			t.Parallel()
-			assertEqualNumber(t, test.x.Remainder(test.y), test.want)
+			actual := test.x.Remainder(test.y)
```
jakebailey: This is an unnecessary refactor. I don't really get what the point of some of these changes is...

jakebailey:
This PR doesn't seem right; tests don't pass, and the code has been rephrased in places.

Did you just get an LLM to write and send this?

internal/jsnum/jsnum_test.go
parthokr: You are right. I was getting overly finicky.

internal/jsnum/jsnum.go
parthokr: Not exactly rephrased.
This was about **precision**.
A dry run may help? Say n is 1 and d is something like 1e-10, the current code spits out a result with not the precision as we would expect from Javascript.

What's interesting is that my first attempt at fixing this while improving precision and getting closer to the spec (hence the tolerance constant I added) – still isn't perfect.

I'm now thinking that using Go's built-in math.Mod function might be appropriate to achieve same behavior to that of ECMA.

parthokr:
> This PR doesn't seem right; tests don't pass, and the code has been rephrased in places.
> 
> Did you just get an LLM to write and send this?

@jakebailey 
I was dumb enough to test only for jsnum_test and had completely overlooked the fact that `assertEqualNumber` has been used in string_test too.
Though it seemed helpful to distinguish between signed and unsigned zero for the sole purpose of testing remainder operation only, it really made no sense to make tests fail by asserting Number("0") == Number("-0").

RyanCavanaugh:
> the current code spits out a result with not the precision as we would expect from Javascript.

Not saying we wouldn't accept a PR to improve correctness, but this functionality is only used in the constant evaluator (e.g. when we see an enum of the form `enum Foo { a = 6 / 2 }`; there are likely zero codebases on the planet that use this code in a way that exposes precision beyond a handful of bits, so the priority of fixing edge cases is extremely low.

parthokr:
> > the current code spits out a result with not the precision as we would expect from Javascript.
> 
> 
> 
> Not saying we wouldn't accept a PR to improve correctness, but this functionality is only used in the constant evaluator (e.g. when we see an enum of the form `enum Foo { a = 6 / 2 }`; there are likely zero codebases on the planet that use this code in a way that exposes precision beyond a handful of bits, so the priority of fixing edge cases is extremely low.

That makes sense.

Mind checking the latest commits?
Replacing the remainder calculation with math.Mod was quite trivial and feels natural in a way.

Moreover
if I am not mistaken; when dividend is less than zero the remainder shouldn't be negativeZero in no way?

That was also core concern of this PR.

internal/jsnum/jsnum.go
```diff
@@ -117,6 +117,10 @@ func (x Number) Floor() Number {
 	return Number(math.Floor(float64(x)))
 }
 
+func (x Number) Ceil() Number {
+	return Number(math.Ceil(float64(x)))
+}
+
```
jakebailey: This is unused and could be removed.

internal/jsnum/jsnum.go
jakebailey: I see. Are you sure removing `n < 0` is correct? The spec says:

> 10. If r = 0 and n < -0𝔽, return -0𝔽.

parthokr: Well
`r == 0` can be removed also if we use math.Mod.
Why?
Let's be a bit more elaborative.

> If r = 0 and n < -0𝔽, return -0𝔽.

It explicitly mentions to work under a conjunction of constraints where the existing setup is [disjunctive](https://github.com/microsoft/typescript-go/blob/fc9d55b36eadf39b4a1a3387c1a8560693c9d765/internal/jsnum/jsnum.go#L142C12-L142C14)

and returns 0 whenever the `n < 0` is satisfied.

May be cross checking [this](https://github.com/parthokr/typescript-go/commit/7ea5363d29160139aea31e2f7b3f3ee2d1033ebf#diff-460e8cc2631e5334543c66629a96f311266ef178f59e03d07716dcd7f0d6a72aR319) would help with existing setup?

---
I think this switch case branching is not really needed. Go's math.Mod already satisfies what ECMA spec states.

 We can reference [this](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/math/mod.go;l=22#:~:text=//%09Mod(%C2%B1Inf%2C%20y,x%2C%20NaN)%20%3D%20NaN)
```
//	Mod(±Inf, y) = NaN  <--- 2
//	Mod(NaN, y) = NaN <--- 1
//	Mod(x, 0) = NaN  <--- 4
//	Mod(x, ±Inf) = x <--- 3
//	Mod(x, NaN) = NaN. <--- 1
```
Now for `5`

> 5. If n is either +0𝔽 or -0𝔽, return n.

We don't have options for d to be `NaN`, `0`, `±Inf`. So when n is `0`, we are left with rest of the floating point numbers and as per [this](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/math/mod.go;l=22#:~:text=its,that%20of%20x) `0%d` and `-0%d` would fall back here with correct sign.

Similarly `10` would also fall back there.
Why?
For `10` 
if `r` gotta be 0 and we have a valid range of n = (-inf, -0F) (yeah not half open because at n=-inf has been covered in `2`) then it makes sign of `r` to that of `n` implicitly implementing `10`.

So do you recommend replacing the function with math.Mod or just replace `||` with `&&`?
Latter one doesn't guarantee exact precision though.

## PS: [This](https://go.dev/play/p/O7iyhyTHmHt) may help.

internal/jsnum/jsnum.go
parthokr: Okay reverted.

jakebailey:
Er, did you intentionally close this?

05TIME:
no, tell me what to do

PR #517, opened 2025-03-12
Align gutter width with old compiler

haoqixu:
tsc 5.8.2:

tsc

Before:

before

After:

after

### DIFF
--- a/internal/diagnosticwriter/diagnosticwriter.go
+++ b/internal/diagnosticwriter/diagnosticwriter.go
@@ -90,7 +90,7 @@ func writeCodeSnippet(writer io.Writer, sourceFile *ast.SourceFile, start int, l
 	lastLineOfFile, _ := scanner.GetLineAndCharacterOfPosition(sourceFile, len(sourceFile.Text))
 
 	hasMoreThanFiveLines := lastLine-firstLine >= 4
-	gutterWidth := len(strconv.Itoa(lastLineOfFile + 1 + len("")))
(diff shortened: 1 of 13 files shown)


haoqixu:
ref: https://github.com/microsoft/TypeScript/blob/release-3.9/src/compiler/program.ts#L390

internal/diagnosticwriter/diagnosticwriter.go
```diff
@@ -90,7 +90,7 @@ func writeCodeSnippet(writer io.Writer, sourceFile *ast.SourceFile, start int, l
 	lastLineOfFile, _ := scanner.GetLineAndCharacterOfPosition(sourceFile, len(sourceFile.Text))
 
 	hasMoreThanFiveLines := lastLine-firstLine >= 4
-	gutterWidth := len(strconv.Itoa(lastLineOfFile + 1 + len("")))
+	gutterWidth := len(strconv.Itoa(lastLine + 1))
```
jakebailey, line 93: For reference, the original code is:

```ts
    const hasMoreThanFiveLines = (lastLine - firstLine) >= 4;
    let gutterWidth = (lastLine + 1 + "").length;
    if (hasMoreThanFiveLines) {
        gutterWidth = Math.max(ellipsis.length, gutterWidth);
    }
```

So your fix is definitely correct; the `+ ""` was a conversion.

jakebailey (approved):

PR #521, opened 2025-03-12
Synthetic export for nested namespace, redux

sandersn:
Redo of #412 since I had to remake my fork as public.
 
Add a synthetic modifier list and export modifier when parsing nested namespaces. This is very similar to the way I'm going to add support for @type, @param and @return next, so I think this is a good test case.

When adding the synthetic node, I mark it with NodeFlagsSynthesized but provided a realistic position -- not -1,-1 I think having a real position will make error reporting easier for JSDoc tags, although it doesn't help for export keywords on nested namespaces. 

In astnav, if the binary search encounters a synthetic node, it filters out all synthetics and tries the binary search again. Thanks to @rbuckton for that idea.

I also erased the synthetic exports in the type eraser. I'm not 100% sure this is the right transform, but it is a TS-only construct, and it's where I'll erase all the JS-specific nodes as well. (Those which aren't types themselves, and therefore *already* erased by the type eraser.)
 
Fixes https://github.com/microsoft/typescript-go/issues/309


internal/ast/utilities.go
```diff
@@ -80,6 +80,17 @@ func PositionIsSynthesized(pos int) bool {
 	return pos < 0
 }
 
+func FindLastVisibleNode(nodes []*Node) *Node {
+	fromEnd := 1
+	for fromEnd <= len(nodes) && nodes[len(nodes)-fromEnd].Flags&NodeFlagsSynthesized != 0 {
```
rbuckton: I wonder if `Synthesized` is the wrong flag for this and we need something else like `Virtual` instead. We seem to be running out of space in `NodeFlags`, but it might be worth seeing if there are any flags we can reclaim or move. For example, `NodeFlagsNamespace`, `NodeFlagsNestedNamespace`, and `NodeFlagsGlobalAugmentation` only ever apply to a `ModuleDeclaration`, so it might make more sense to move those to a separate enum and field on `ModuleDeclaration` instead. We may not even need `NodeFlagsGlobalAugmentation` as that can be trivially inferred from the name of the module declaration (`global`).

internal/transformers/typeeraser.go
```diff
@@ -9,10 +9,11 @@ import (
 type TypeEraserTransformer struct {
 	Transformer
 	compilerOptions *core.CompilerOptions
+	sourceFile      *ast.SourceFile
 }
 
-func NewTypeEraserTransformer(emitContext *printer.EmitContext, compilerOptions *core.CompilerOptions) *Transformer {
-	tx := &TypeEraserTransformer{compilerOptions: compilerOptions}
+func NewTypeEraserTransformer(emitContext *printer.EmitContext, sourceFile *ast.SourceFile, compilerOptions *core.CompilerOptions) *Transformer {
+	tx := &TypeEraserTransformer{compilerOptions: compilerOptions, sourceFile: sourceFile}
```
rbuckton: This is normally picked up by capturing the source file in `visit` (or `visitSourceFile`) as it's likely that I will change transformers to be reused like they are in Strada (though changed to one per thread rather than one for all emit). See `runtimesyntax.go` or `commonjsmodule.go` for examples.

internal/parser/parser.go
```diff
@@ -2037,7 +2037,13 @@ func (p *Parser) parseModuleOrNamespaceDeclaration(pos int, hasJSDoc bool, modif
...
 	var body *ast.Node
 	if p.parseOptional(ast.KindDotToken) {
-		body = p.parseModuleOrNamespaceDeclaration(p.nodePos(), false /*hasJSDoc*/, nil /*modifiers*/, ast.NodeFlagsNestedNamespace|namespaceFlag)
+		syntheticExport := p.factory.NewModifier(ast.KindExportKeyword)
+		syntheticExport.Loc = core.NewTextRange(p.nodePos(), p.nodePos())
+		syntheticExport.Flags = ast.NodeFlagsSynthesized
+		nodes := p.nodeSlicePool.NewSlice(1)
+		nodes[0] = syntheticExport
+		syntheticModifiers := p.newModifierList(syntheticExport.Loc, nodes)
+		body = p.parseModuleOrNamespaceDeclaration(p.nodePos(), false /*hasJSDoc*/, syntheticModifiers, ast.NodeFlagsNestedNamespace|namespaceFlag)
```
rbuckton: If we add a synthetic `export` keyword, can we do away with `NodeFlagsNestedNamespace`?

rbuckton (changes requested):

internal/ast/utilities.go
ahejlsberg: A number of these flags were previously computed by the checker so emit could use them. I think a lot of them can be eliminated now.

rbuckton: We've discussed deprecating `module id {}` in favor of `namespace id {}` for a while, especially in light of the [Module Declarations Proposal](https://github.com/tc39/proposal-module-declarations). Once we've finished the port, we should give some thought to splitting `namespace id {}` out to `NamespaceDeclaration` and only supporting `module "string" {}` and `global {}` for `ModuleDeclaration`.

sandersn:
I've got the 3 namespace flags down to just FlagsNamespace, which can just be a bool. But I'm not sure how to reduce it further since the keyword token isn't saved during parsing.

rbuckton:
I think a `bool` on `ModuleDeclaration` is probably good enough. A `Token` might be better as there's less work necessary to recover position information for source maps/comments.

sandersn:
ah, didn't see this until I pushed a commit that switches to Kind, not Token. Turns out that `namespace global {` and `global {` are only distinguished by the flag, which means that I need a 3-way distinction that I encoded in the keyword token's kind.

Actually, `global { }` *also* uses "global" as its Name, parsing it as an identifier. I think it's safer to save only the kind for the keyword instead of having two nodes for the same span.

Edit: Also it complicates nested namespaces, which end up reusing the same `namespace` token for all the generated namespaces.

internal/parser/parser.go
sandersn: It's even easier -- just check whether a namespace's parent is also a namespace (and that the parent's body is the namespace).

internal/ast/ast.go
```diff
@@ -3401,22 +3401,23 @@ type ModuleDeclaration struct {
 	ModifiersBase
 	LocalsContainerBase
 	BodyBase
-	name *ModuleName // ModuleName
+	name    *ModuleName // ModuleName
+	Keyword Kind        // KindModuleKeyword (deprecated), KindNamespaceKeyword, KindGlobalKeyword (global augmentation)
```
rbuckton: `module` isn't what's deprecated as we are still going to use `module "string"` for ambient module declarations. Only `module id`/`module a.b` is deprecated.

I'm surprised `KindGlobalKeyword` is needed given we parse `global` as an identifier when parsing `declare global {}`. The information is somewhat redundant, but I suppose we might as well since we need to store `Keyword` anyway. Ideally, we wouldn't bother to parse `global` as an identifier and would just leave `name` as `nil`, though I'm sure that would break some code somewhere since we expect `ModuleDeclaration` to always have some kind of name.

internal/ast/nodeflags.go
```diff
@@ -7,25 +7,23 @@ const (
...
-	NodeFlagsJavaScriptFile                  NodeFlags = 1 << 19 // If node was parsed in a JavaScript
-	NodeFlagsThisNodeOrAnySubNodesHasError   NodeFlags = 1 << 20 // If this node or any of its children had an error
-	NodeFlagsHasAggregatedChildData          NodeFlags = 1 << 21 // If we've computed data from children and cached it in this node
+	NodeFlagsOptionalChain                   NodeFlags = 1 << 5  // Chained MemberExpression rooted to a pseudo-OptionalExpression
+	NodeFlagsExportContext                   NodeFlags = 1 << 6  // Export context (initialized by binding)
+	NodeFlagsContainsThis                    NodeFlags = 1 << 7  // Interface contains references to "this"
+	NodeFlagsHasImplicitReturn               NodeFlags = 1 << 8  // If function implicitly returns on one of codepaths (initialized by binding)
+	NodeFlagsHasExplicitReturn               NodeFlags = 1 << 9  // If function has explicit reachable return on one of codepaths (initialized by binding)
+	NodeFlagsHasAsyncFunctions               NodeFlags = 1 << 10 // If the file has async functions (initialized by binding)
```
rbuckton: This flag isn't used anymore except as an alias to `NodeFlagsIdentifierIsInJSDocNamespace`, which itself is only used in JSDoc namespaces in the same way `NestedNamespace` was used. It's possible we could get rid of both to save more space in `NodeFlags`.

internal/transformers/typeeraser.go
```diff
@@ -100,6 +100,10 @@ func (tx *TypeEraserTransformer) visit(node *ast.Node) *ast.Node {
 			// TypeScript module declarations are elided if they are not instantiated or have no body
 			return nil
 		}
+		if ast.IsModuleDeclaration(tx.parentNode) && tx.parentNode.AsModuleDeclaration().Body == node {
```
rbuckton: Is this necessary? I would think this would be handled in runtimesyntax.go.

internal/transformers/runtimesyntax.go
```diff
@@ -145,7 +145,9 @@ func (tx *RuntimeSyntaxTransformer) isFirstDeclarationInScope(node *ast.Node) bo
 }
 
 func (tx *RuntimeSyntaxTransformer) isExportOfNamespace(node *ast.Node) bool {
-	return tx.currentNamespace != nil && (node.ModifierFlags()&ast.ModifierFlagsExport != 0 || node.Flags&ast.NodeFlagsNestedNamespace != 0)
+	return tx.currentNamespace != nil &&
+		(node.ModifierFlags()&ast.ModifierFlagsExport != 0 ||
+			ast.IsModuleDeclaration(node) && ast.IsModuleDeclaration(tx.parentNode) && tx.parentNode.AsModuleDeclaration().Body == node)
```
rbuckton: Is this line necessary assuming we don't drop the synthetic `export` in typeeraser.go?

rbuckton (changes requested):

internal/ast/ast.go
sandersn: OK, I'll remove the deprecated note.

internal/ast/nodeflags.go
sandersn: When I merged from main this morning, I see it used 3 times as itself in the binder to mark async function expressions/declarations and method declarations. But the emitter never checks the flag. Both the binder and emitter are complete, right? If so, it's safe to remove them.

Actually, `typedef` and `callback` are no longer going to support nested names, since that was a hardly-used Closure feature. So all of the flags can be removed.

internal/transformers/runtimesyntax.go
sandersn: not for semantics. I thought that the emit might be different if we don't transform `namespace X { Y { Z { } }` back into `namespace X.Y.Z { }`, but that would be type emit -- and it might not matter if we changed that.

internal/transformers/typeeraser.go
sandersn: I wasn't sure whether nested namespaces were "type erased" syntax or runtime syntax. Putting it in the type eraser would, I guessed, make sure that it was back to the Strada-expected form both for .js emit and .d.ts emit.

Like I said in the runtimesyntax comment, maybe it's better to expect the .d.ts emit to change (or the .d.ts emitter to change). What do you think?

rbuckton: for `export namespace a.b ...`? I think I would expect our .d.ts to continue emit `export namespace a.b ...`.

sandersn: hmm. OK. I can remove the changes in both typeeraser and runtimesyntax then.

rbuckton (approved):

PR #523, opened 2025-03-12
Add Star history graph

melroyvandenberg:
Just for the fun, lets show off the star history. Engaging users to also star the project.

If you really do not like it, that is fine just close the PR

image


RyanCavanaugh:
Fun to see but I don't think we need this in the readme. Thanks!

PR #543, opened 2025-03-12
Update Windows ANSI code code based on feedback

jakebailey:
cc @qmuntal

### DIFF
--- a/cmd/tsgo/enablevtprocessing_windows.go
+++ b/cmd/tsgo/enablevtprocessing_windows.go
@@ -5,12 +5,18 @@ import (
 )
 
 func enableVirtualTerminalProcessing() {
-	hStdout, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
-	if err == nil && hStdout != windows.InvalidHandle {
+	h, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
+	if err != nil || h == windows.InvalidHandle {
+		return
+	}
+	fileType, err := windows.GetFileType(h)
+	if err != nil || fileType == windows.FILE_TYPE_CHAR {
 		var mode uint32
-		err = windows.GetConsoleMode(windows.Handle(hStdout), &mode)
-		if err == nil {
-			windows.SetConsoleMode(windows.Handle(hStdout), mode|windows.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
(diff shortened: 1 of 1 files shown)


cmd/tsgo/enablevtprocessing_windows.go
```diff
@@ -5,12 +5,18 @@ import (
...
 
 func enableVirtualTerminalProcessing() {
-	hStdout, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
-	if err == nil && hStdout != windows.InvalidHandle {
+	h, err := windows.GetStdHandle(windows.STD_OUTPUT_HANDLE)
+	if err != nil || h == windows.InvalidHandle {
+		return
+	}
+	fileType, err := windows.GetFileType(h)
+	if err != nil || fileType == windows.FILE_TYPE_CHAR {
```
DanielRosenwasser, line 13: I think it's worth noting that you need to do this to only enable virtual terminal processing if stdout hasn't been redirected. Otherwise, leave it as-is.

DanielRosenwasser: Or it's possible you don't even "need" to do it? It's unclear to me what the purpose is since libuv doesn't do this.

DanielRosenwasser (approved):

cmd/tsgo/enablevtprocessing_windows.go
qmuntal: > I think it's worth noting that you need to do this to only enable virtual terminal processing if stdout hasn't been redirected. Otherwise, leave it as-is.

The `STD_OUTPUT_HANDLE` is not redirected if it is a character device and `GetConsoleMode` succeeds. See for example [dotnet's `IsHandleRedirected` function](https://github.com/dotnet/runtime/blob/5ff417fd7fbff9bbd3c66f35a1721ddaa37840eb/src/libraries/System.Console/src/System/ConsolePal.Windows.cs#L163-L171).

> Or it's possible you don't even "need" to do it? It's unclear to me what the purpose is since libuv doesn't do this.

It is optional. I suggested this change to @jakebailey because `GetConsoleMode` can be slow, and `GetFileType` (which is much faster), helps doing the `GetConsoleMode` only when it is really necessary. Having said this, as `enableVirtualTerminalProcessing` is a one-time call, it won't make any difference.

PR #544, opened 2025-03-12
FIX: Removed grammar issues in comments

KeoFoxy:
Fixed grammar issues in files lib.dom.d.ts and lib.es2015.core.d.ts


KeoFoxy:

jakebailey:
These are generated files based on the TypeScript submodule that we cannot change; this kind of PR would need to be made in the main TypeScript repo.

PR #546, opened 2025-03-12
Add missing whitespace in construct signature prints

Andarist:
We can notice that this is closer to Strada given some of the `.diff` files got removed here.


jakebailey (approved):
A lot of stuff doesn't improve but that's because type printing isn't naming types still, so this would look a lot more positive if fixed later (but it's still right).

PR #547, opened 2025-03-12
Fix detection of super statement in visiting constructor

### LINKED ISSUE #540: Parameter properties in subclass constructor are assigned before the `super()` call
**TS Code:**
```ts
class Foo {
	constructor() {
		this.foo = "foo"	
	}
	public foo: string;
}

class Bar extends Foo {
	constructor(
		public bar: string
	) {
		super();
	}
}
```

**typescript output**:
```js
// ...
class Bar extends Foo {
	constructor(bar) {
		super();
		this.bar = bar;
	}
}
```

**typescript-go output**:
```js
// ...
class Bar extends Foo {
	constructor(bar) {
		this.bar = bar;
		super();
	}
}
```

frodi-karlsson:
Fixes #540


jakebailey:
This would probably benefit from an emit test (see the neighboring file for this transform),

frodi-karlsson:

internal/transformers/runtimesyntax_test.go
```diff
@@ -420,9 +420,17 @@ func TestParameterPropertyTransformer(t *testing.T) {
...
+}
+class C extends B {
+    x;
+    constructor(x) {
+        super();
+        this.x = x;
+    }
 }`},
 	}
-
```
jakebailey, line 436: Nit but, I'd suggest restoring this extra change.

frodi-karlsson:
For some more context:

- If there are no parameters, we return early on L804.
- The nodes before pushing the constructor are `{ current: Parameter, parent: ClassDeclaration }`
- Pushing the constructor yields the class declaration, which we should then check for classness and all that

testdata/baselines/reference/submodule/conformance/defineProperty(target=esnext).js.diff
jakebailey, line 1: This to me seems like the evidence that this change is right.

jakebailey (changes requested):
Checked with @rbuckton and there's apparently another cause to this.

jakebailey (changes requested):
Checked with @rbuckton and there's apparently another cause to this.

rbuckton:
There are two issues at play here. One is that `visit` has an early `return` when visiting modifiers on parameters and fails to call `popNode`, the other is that we don't need to call `pushNode(constructor)` in `transformConstructorBody`

rbuckton:
I've put up the fix for these issues in #679

PR #565, opened 2025-03-13
Support json module resolution in file loader

### LINKED ISSUE #534: resolveJsonModule doesn't work
I've set up a simple example project here:
> git clone https://gist.github.com/Sainan/cffd2511bcedfedd2e4ae921d9c9731c

This compiles via `tsc` but when running `tsgo`, it seems to fail with "Cannot find module './tsconfig.json' or its corresponding type declarations."

Note that this can be any JSON file, I just decided to reuse tsconfig.json to keep it minimal.

frodi-karlsson:
I'm sorry if I'm handling the baseline tests incorrectly, they are new to me. I also hope this isn't too naive a solution...

Fixes #534


internal/compiler/fileloader.go
```diff
@@ -274,8 +274,11 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile)
 
 			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
 			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
-
-			shouldAddFile := resolution.IsResolved() && tspath.FileExtensionIsOneOf(resolvedFileName, []string{".ts", ".tsx", ".mts", ".cts"})
+			allowedExtensions := []string{".ts", ".tsx", ".mts", ".cts"}
+			if core.Tristate.IsTrue(p.compilerOptions.ResolveJsonModule) {
+				allowedExtensions = append(allowedExtensions, ".json")
```
jakebailey: This conditional appending is definitely going to be slower than just conditionally choosing between two slices; do we already have two available slices via the tspath package, perhaps?

jakebailey:
Great diff reduction in any case!

Sainan:
If merged, this closes #534.

internal/compiler/fileloader.go
```diff
@@ -274,8 +274,13 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile)
 
 			// Don't add the file if it has a bad extension (e.g. 'tsx' if we don't have '--allowJs')
 			// This may still end up being an untyped module -- the file won't be included but imports will be allowed.
-
-			shouldAddFile := resolution.IsResolved() && tspath.FileExtensionIsOneOf(resolvedFileName, []string{".ts", ".tsx", ".mts", ".cts"})
+			hasAllowedExtension := false
+			if core.Tristate.IsTrue(p.compilerOptions.ResolveJsonModule) {
```
jakebailey:
```suggestion
			if p.compilerOptions.ResolveJsonModule.IsTrue() {
```

jakebailey (approved):

PR #568, opened 2025-03-13
Add a FAQ note for simplicity

Zaid-maker:
Summary:
- Added a "NOTE" section with a brief introduction to FAQs related to TypeScript-go.
- Included a link to the FAQs on TypeScript-go discussions.
- Adjusted line numbers accordingly for the changes.


jakebailey:
If we add a message like this (and I'm not sure we should), I don't think it should be the first text after the header 😅

Zaid-maker:
> If we add a message like this (and I'm not sure we should), I don't think it should be the first text after the header 😅

Its OK we can adjust the message where do you want it to be?

This message will grab eyes immediately that's why I added it there but we adjust it

RyanCavanaugh:
I think we should link to the FAQ, not recite the FAQ

Zaid-maker:
Will come back to this PR tomorrow!

Zaid-maker:
@jakebailey i have updated the message that displays i hope it get's the job done, let me know if any adjustment needed
image

jakebailey:
I just don't think we need this. The existing link already has a link to the FAQ and other info, and what you have here is very uncannily LLM-written text in a style that doesn't match any other text we have here.

PR #572, opened 2025-03-13
Fix name resolution issue (porting bug)

### LINKED ISSUE #545: Parameter with object initializer not working with imported function
Given this `dep.ts` file:

```ts
// dep.ts
export function dep(str: string) {
  return `dep: ${str}`;
}
````

And this main.ts file:

```ts
// main.ts
import { dep } from "./dep.ts";

function run(context = { dep }) {
  const { dep } = context;
  console.log(dep("Hello, world!"));
}
run();
```

It does not compile with typescript-go, but does with TypeScript. As far as I know it should be supported with the current compiler.

<details>
<summary>
It provides these errors:
</summary>

```
main.ts:3:14 - error TS7022: 'context' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

3 function run(context = { dep }) {
               ~~~~~~~~~~~~~~~~~

main.ts:3:26 - error TS2373: Parameter 'context' cannot reference identifier 'dep' declared after it.

3 function run(context = { dep }) {
                           ~~~

main.ts:3:26 - error TS2448: Block-scoped variable 'dep' used before its declaration.

3 function run(context = { dep }) {
                           ~~~

  main.ts:4:11 - 'dep' is declared here.
    4   const { dep } = context;
                ~~~

main.ts:4:11 - error TS7022: 'dep' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

4   const { dep } = context;
            ~~~


Found 4 errors in the same file, starting at: main.ts:3
```

</details>

ahejlsberg:
Fixes #545.


internal/binder/nameresolver.go
```diff
@@ -358,7 +358,7 @@ func (r *NameResolver) useOuterVariableScopeInParameter(result *ast.Symbol, loca
 						r.SetRequiresScopeChangeCache(functionLocation, declarationRequiresScopeChange)
 					}
 				}
-				return declarationRequiresScopeChange == core.TSTrue
+				return declarationRequiresScopeChange != core.TSTrue
```
jakebailey, line 361: This is why I added the methods, I couldn't keep the equalities straight 😅

jakebailey (approved):

internal/binder/nameresolver.go
griffintarzan: Just out of curiosity, how do you guys find these out? It's a huge project to debug, and a minor mistake. Do you guys have the structure memorized so well that you can quickly pinpoint potential causes and dive into the right files?

PR #577, opened 2025-03-13
Update stack.go

Kaawasaaki:
the new code can handle off by one errors and and handles the zero value by returning a default value


jakebailey:
Thanks for the PR, but I don't think we need this (panicking on programmer error is okay), and you haven't updated any of the referencing code.

PR #579, opened 2025-03-13
Don't explicitly ask for bind diags for error baselines

jakebailey:
`GetSemanticDiagnostics` already does this.


ahejlsberg (approved):
Wow, looks like that got rid of a lot of errors. Enough that GitHub stuggles a bit. 😄

PR #583, opened 2025-03-13
Assign normalized decimal value representations to non-decimal number tokens

Andarist:
Related topics might still require further work but this is already small step into the right direction


internal/scanner/scanner.go
```diff
@@ -1863,14 +1863,6 @@ func (s *Scanner) scanBigIntSuffix() ast.Kind {
 		// !!! Convert all bigint tokens to their normalized decimal representation
 		return ast.KindBigIntLiteral
```
jakebailey, line 1864: Oh dang, are we going to have to need bigint processing after all?

jakebailey (approved):

PR #584, opened 2025-03-13
Partially handle named anonymous types in type printer, other fixes

jakebailey:
This is in no way perfect logic, but it does mean we'll print out names for a bunch of named anonymous types in our temporary type printer, e.g. `typeof globalThis` instead of a monstrous string.

This removes 2500 lines of baselines, but more importantly, nearly _10MB_ of baselines on disk.


jakebailey:
~Hm, this isn't working in hover:~

I switched to main to check the file sizes and forgot to switch back, sigh

It works:

image

image

internal/checker/printer.go
```diff
@@ -348,7 +348,27 @@ func (p *Printer) printTupleType(t *Type) {
 	p.print("]")
 }
 
+func (c *Checker) isClassInstanceSide(t *Type) bool {
+	return t.symbol != nil && t.symbol.Flags&ast.SymbolFlagsClass != 0 && (t == c.getDeclaredTypeOfClassOrInterface(t.symbol) || (t.flags&TypeFlagsObject != 0 && t.objectFlags&ObjectFlagsIsClassInstanceClone != 0))
+}
+
 func (p *Printer) printAnonymousType(t *Type) {
+	// !!! logic partially taken from createAnonymousTypeNode
```
ahejlsberg: I think this all can be simplified to the following (which also handles `typeof import`):

```Go
	if t.symbol != nil && len(t.symbol.Name) != 0 {
		if t.symbol.Flags&(ast.SymbolFlagsClass|ast.SymbolFlagsEnum|ast.SymbolFlagsValueModule) != 0 {
			if t == p.c.getTypeOfSymbol(t.symbol) {
				p.print("typeof ")
				if t.symbol.Flags&ast.SymbolFlagsValueModule != 0 && t.symbol.Name[0] == '"' {
					p.print("import(")
					p.print(t.symbol.Name)
					p.print(")")
				} else {
					p.printName(t.symbol)
				}
				return
			}
		}
	}
```

ahejlsberg: I tried it locally and the changes look pretty good, but not sure if they square with yours.

jakebailey: Yeah, that seems to work better. Will do that.

jakebailey: Wow, that brought the PR from 2.5k lines removed to 5k lines removed. Very nice.

internal/checker/printer.go
```diff
@@ -348,7 +348,27 @@ func (p *Printer) printTupleType(t *Type) {
 	p.print("]")
 }
 
+func (c *Checker) isClassInstanceSide(t *Type) bool {
```
ahejlsberg: Don't think this is needed anymore.

jakebailey: Whoops, thanks.

ahejlsberg:
Also, I think you can remove some more differences by stripping the file extension here:

https://github.com/microsoft/typescript-go/blob/main/internal/binder/binder.go#L760

jakebailey:
Ah, great. Will do that too quick.

jakebailey:
Now up to 6.2k lines removed!

ahejlsberg:
Did we have all those double-double-quotes in module names in the old compiler symbol and error baselines as well?

ahejlsberg (approved):

jakebailey:
No, that's something else I've wanted to fix. I can take a look again later.

jakebailey:
Removing the extra quoting (came from #277 it seems), now removing 8k lines.

PR #586, opened 2025-03-13
Run gofumpt via go tool in dprint

jakebailey:
This is a lot slower; `go tool` is faster than `go run` from 1.23, but still a lot slower than a plain binary. But, this doesn't require any global installs of `gofumpt` to run `dprint`. Given the recommendation is to configure `gopls` to run `gofumpt` instead, and `dprint` does cache things, it seems like this isn't such a bad tradeoff to avoid `dprint fmt` failing unexpectedly for those using that CLI.

EDIT: see https://github.com/microsoft/typescript-go/pull/586#issuecomment-2902268019 for perf info; most of the time it's fine.


jakebailey:
I think I want to merge this to make things easier.

For formatting the whole codebase:

```
Benchmark 1: dprint fmt --incremental=false
  Time (mean ± σ):     261.0 ms ±  17.4 ms    [User: 141.6 ms, System: 91.4 ms]
  Range (min … max):   241.3 ms … 294.9 ms    10 runs
 
Benchmark 2: dprint fmt --incremental=false
  Time (mean ± σ):      4.473 s ±  0.114 s    [User: 0.179 s, System: 0.069 s]
  Range (min … max):    4.302 s …  4.669 s    10 runs
 
Summary
  dprint fmt --incremental=false ran
   17.14 ± 1.22 times faster than dprint fmt --incremental=false
```

But, for our largest file:

```
Benchmark 1: dprint fmt --incremental=false ./internal/checker/checker.go
  Time (mean ± σ):     146.6 ms ±   2.7 ms    [User: 15.1 ms, System: 11.6 ms]
  Range (min … max):   144.1 ms … 153.3 ms    10 runs
 
Benchmark 2: dprint fmt --incremental=false ./internal/checker/checker.go
  Time (mean ± σ):     295.8 ms ±  34.2 ms    [User: 14.0 ms, System: 11.9 ms]
  Range (min … max):   232.7 ms … 338.9 ms    10 runs
 
Summary
  dprint fmt --incremental=false ./internal/checker/checker.go ran
    2.02 ± 0.24 times faster than dprint fmt --incremental=false ./internal/checker/checker.go
```

Given incremental mode won't format files that haven't changed, most of the time dprint will be quick. So I think the perf is actually totally fine. Here's what happens if I have it touch `checker.go` each time but with incremental enabled:

```
Benchmark 1: dprint fmt
  Time (mean ± σ):     295.2 ms ±   8.1 ms    [User: 21.2 ms, System: 62.0 ms]
  Range (min … max):   281.0 ms … 306.8 ms    10 runs
 
Benchmark 2: dprint fmt
  Time (mean ± σ):     550.1 ms ±  56.3 ms    [User: 15.0 ms, System: 76.1 ms]
  Range (min … max):   477.0 ms … 635.7 ms    10 runs
 
Summary
  dprint fmt ran
    1.86 ± 0.20 times faster than dprint fmt
```

andrewbranch:
> Given the recommendation is to configure gopls to run gofumpt instead

How do you do this?

jakebailey:
> > Given the recommendation is to configure gopls to run gofumpt instead
> 
> How do you do this?

This is just in: https://github.com/microsoft/typescript-go/blob/f6f7e665c9960411b19391d6c0ba7ac37e65c195/.vscode/settings.template.json#L22

sandersn (approved):
I run `hereby format` 2-3x per PR, so some slowdown is worth it to me. (I also have `"formatting.gofumpt": true` in workspace settings.)

jakebailey:
I'm confused, those two are contradictory; this PR makes it a little slower (but only when the formatter is called many times)

PR #587, opened 2025-03-13
Enum sort order and unused identifier error spans

ahejlsberg:
In this PR:

* Enum members are now consistently sorted by declaration order.
* Improved error spans for unused identifiers (was inconsistent before and in old compiler).


testdata/baselines/reference/submodule/compiler/noUnusedLocals_typeParameterMergedWithParameter.errors.txt.diff
```diff
@@ -24,8 +22,9 @@
 -!!! error TS6133: 'T' is declared but its value is never read.
      
      function useParam<T>(T: number) {
-                      ~~~
+-                     ~~~
 -!!! error TS6133: 'T' is declared but its value is never read.
++                      ~
```
jakebailey, line 27: The new range here appears to be unexpected, though I can't say I like where it used to be (including the angle brackets) either.

testdata/baselines/reference/submodule/compiler/unusedImports1.errors.txt.diff
```diff
@@ -0,0 +1,15 @@
...
+-file2.ts(1,1): error TS6133: 'Calculator' is declared but its value is never read.
++file2.ts(1,9): error TS6133: 'Calculator' is declared but its value is never read.
+ 
+ 
+ ==== file1.ts (0 errors) ====
+@@= skipped -7, +7 lines =@@
+     
+ ==== file2.ts (1 errors) ====
+     import {Calculator} from "./file1"
+-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jakebailey, line 13: Hm, the old range here was the entire import too... Feels like we should backport this one as well honestly.

testdata/baselines/reference/submodule/compiler/noUnusedLocals_typeParameterMergedWithParameter.errors.txt.diff
ahejlsberg: Yes, that's intended.

testdata/baselines/reference/submodule/compiler/unusedImports1.errors.txt.diff
ahejlsberg: Yeah, the old compiler's logic is just kinda wonky and incomplete. Doesn't make sense to talk about "all" when there is only one, and doesn't make sense to highlight a whole statement when you can just highlight the actual identifier.

testdata/baselines/reference/submodule/compiler/noUnusedLocals_typeParameterMergedWithParameter.errors.txt.diff
ahejlsberg: Also, the old compiler talks about "reading" type parameters, which is nonsense for non-values.

testdata/baselines/reference/submodule/compiler/unusedImports1.errors.txt.diff
ahejlsberg: That said, I think backporting is a waste of time. Better to just accept the changes and move forward.

jakebailey: I don't entirely agree; it'd be nice to get these changes out to people and fix up our baselines, if it's easy to do so.

PR #588, opened 2025-03-14
Move tests around to shorter paths

jakebailey:
- `internal/testutil/runner` -> `internal/testrunner`
- `TestCompilerBaselinesLocal` -> `TestLocal`
- `TestCompilerBaselinesSubmodule` -> `TestSubmodule`
- Tests like `TestCompilerBaselinesSubmodule/conformance/globalThisUnknownNoImplicitAny/error` are now `TestSubmodule/globalThisUnknownNoImplicitAny/error`.
  - Our test names never have duplicates, so this is safe.

These should make it simpler to make a generic launch task like we do in Strada.


jakebailey:
Notably this does _not_ move baselines around in `testdata`, at least not yet. I could tack that onto this PR but I think it might be prudent to do that in separate massive PR.

gabritto:
> Tests like `TestCompilerBaselinesSubmodule/conformance/globalThisUnknownNoImplicitAny/error` are now `TestSubmodule/globalThisUnknownNoImplicitAny/error`.
> * Our test names never have duplicates, so this is safe.

This may be true now, but if we want to keep that invariant in the future, we also need to port the code that checks for duplicates, which I intentionally did not port before.
Also, I copied a couple tests from Strada to the local Corsa tests, so those will be duplicated in the local and submodule tests, and need to be removed.

jakebailey:
> This may be true now, but if we want to keep that invariant in the future, we also need to port the code that checks for duplicates, which I intentionally did not port before.

Ah, yes, we should make a test which verifies this explicitly.

> Also, I copied a couple tests from Strada to the local Corsa tests, so those will be duplicated in the local and submodule tests, and need to be removed.

Those are still separate; it's `TestLocal` and `TestSubmodule`, they're not merged together or anything.

gabritto (commented):
I don't know if it will in fact matter, but the idea is that we'll have more test runners (e.g. emit runner) when we port more of the test infra. Then deleting that `Runner` interface might interfere with running all the runners in a single Go test.

jakebailey:
Are those tests also in the same namespace (in terms of test name)? I can revert that change for sure.

gabritto (approved):

jakebailey:
Actually, I just added to the PR a launch task example. You just have to open a test in the submodule dir, and then it works.

PR #589, opened 2025-03-14
Remove $ from README command blocks for easier copy-paste

odairrosario:
* Removed unnecessary `$` prefix from command blocks in README.
* This avoids errors when copying commands, especially on Windows.
* Aligns with the standard used in Microsoft's VS Code documentation.


jakebailey:
This breaks the look of the commands. Before:

image

After:

image

Text that isn't prefixed with `$` is treated as command output and formatted differently in the `console` language.

odairrosario:

odairrosario:
> This breaks the look of the commands. Before:
> 
> image
> 
> After:
> 
> image
> 
> Text that isn't prefixed with `$` is treated as command output and formatted differently in the `console` language.

Thanks for the feedback! I noticed that the TypeScript and VS Code documentation don't use the $ prefix in command blocks. I was aiming for consistency across Microsoft's documentation. Would it make sense to follow the same pattern here?

monoira:
I agree with OP. `$` thing should be deleted. there is no `look` of the commands. It just makes copy-paste harder.

If you want different look, put it in ```bash or something. just get rid of the `$` sign.

README.md
```diff
@@ -12,25 +12,25 @@ This repo uses [Go 1.24 or higher](https://go.dev/dl/), [Node.js with npm](https
 For tests and code generation, this repo contains a git submodule to the main TypeScript repo pointing to the commit being ported.
 When cloning, you'll want to clone with submodules:
 
-```console
-$ git clone --recurse-submodules https://github.com/microsoft/typescript-go.git
+```
```
jakebailey: If you're going to remove these, at least make them `sh` blocks.

jakebailey (changes requested):

PR #590, opened 2025-03-14
Update README.md

surfmuggle:
This intro should make it clear what this repo is about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # TypeScript 7
+`typescript-go` is a transpiler written in Go that transpiles TypeScript to JavaScript.
 
 [Not sure what this is? Read the announcement post!](https://devblogs.microsoft.com/typescript/typescript-native-port/)
 


README.md
```diff
@@ -1,4 +1,5 @@
 # TypeScript 7
+`typescript-go` is a transpiler written in Go that transpiles TypeScript to JavaScript.
 
 [Not sure what this is? Read the announcement post!](https://devblogs.microsoft.com/typescript/typescript-native-port/)
```
jakebailey, line 4: Doesn't this line say what this is?

surfmuggle: This Line 

> _`typescript-go` is a transpiler written in Go that transpiles TypeScript to JavaScript._ 

is a suggestion from me to be added.

Currently it is not included.

image

jakebailey: Well yes, but the text you're adding isn't accurate; this project is certainly a lot more than a transpiler and I believe requires a lot more context (like what is in the blog post) to accurately describe.

RyanCavanaugh:
I don't think this adds any clarity over the existing text. If someone's not reading the first line of the post, they're not reading the first line of the post.

surfmuggle:
Currently anyone who wants to understand what this repo is about has to read the long post and possibly has also to watch the video since to my understanding the blogpost does not mention everything. 
The post states that this is a 
> native port of the TypeScript compiler and tools.

At another place in the text it is listed

> time to load ... drops down ... with the native language service

and

> We’ll also be moving to the Language Server Protocol (LSP),

What other tools are included is not explicitly stated. 

If a brief list of the relevant features would be included there would be no need to read the post and watch the video.

README.md
surfmuggle: If the line only covers a part of what this repo is about then it should be improved.

PR #593, opened 2025-03-14
Align error for invalid enum option with strada

haoqixu:
If we provide a invalid value to an `enum` option, `tsc` will report a `TS6046` error and `tsgo` will report a `TS5024` error. This PR aligns the `tsgo` with `tsc`.

tsc Version 5.8.2:

tsc

Before:

before

After:

after

### DIFF
--- a/internal/tsoptions/tsconfigparsing.go
+++ b/internal/tsoptions/tsconfigparsing.go
@@ -394,6 +394,9 @@ func convertJsonOption(
 			return normalizeNonListOptionValue(opt, basePath, validatedValue), errors
 		}
 	} else {
+		if opt.Kind == CommandLineOptionTypeEnum {
+			return nil, []*ast.Diagnostic{createDiagnosticForInvalidEnumType(opt, sourceFile, valueExpression)}
+		}
 		return nil, []*ast.Diagnostic{createDiagnosticForNodeInSourceFileOrCompilerDiagnostic(sourceFile, valueExpression, diagnostics.Compiler_option_0_requires_a_value_of_type_1, opt.Name, getCompilerOptionValueTypeString(opt))}
 	}
 }


internal/tsoptions/tsconfigparsing_test.go
```diff
@@ -706,7 +721,7 @@ func TestParseSrcCompiler(t *testing.T) {
 
 	opts := parseConfigFileContent.CompilerOptions()
 	assert.DeepEqual(t, opts, &core.CompilerOptions{
-		Lib:                        []string{"es2020"},
+		Lib:                        []string{"lib.es2020.d.ts"},
```
jakebailey, line 720: This doesn't seem right; that's not what's in the config file and I don't think what Strada parses out.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -376,18 +367,19 @@ func convertJsonOption(
...
+		case CommandLineOptionTypeList:
 			return convertJsonOptionOfListType(opt, value, basePath, propertyAssignment, valueExpression, sourceFile) // as ArrayLiteralExpression | undefined
-		} else if optType == "listOrElement" {
+		case CommandLineOptionTypeListOrElement:
 			if reflect.TypeOf(value).Kind() == reflect.Slice {
 				return convertJsonOptionOfListType(opt, value, basePath, propertyAssignment, valueExpression, sourceFile)
 			} else {
 				return convertJsonOption(opt.Elements(), value, basePath, propertyAssignment, valueExpression, sourceFile)
 			}
-		} else if !(reflect.TypeOf(optType).Kind() == reflect.String) {
```
haoqixu, line 388: `Strada` uses a Map as type of enum option and use `!isString(opt.type)` as condition of this branch. However `tsgo` use `CommandLineOptionKind` for `CommandLineOption.Kind` whose underlying type is string.

internal/tsoptions/tsconfigparsing_test.go
haoqixu: I think `tsoptions.GetLibFileName` has already handled this, and both "es2020" and "lib.es2020.d.ts" are valid for the function.
But I wonder which one the parser should use to set as the value for an enum option? It uses the value of map for top level enum option like `target` and uses the key for elements of a list instead. 🤔

haoqixu: I think Strada returns the value of map in `convertJsonOptionOfCustomType`

jakebailey: Whatever the Strada code did, this code should do.

haoqixu: I dumped the options from Strada's [`parseOwnConfigOfJsonSourceFile`](https://github.com/microsoft/TypeScript/blob/release-5.8/src/compiler/commandLineParser.ts#L3553) and have confirmed that Strada uses the value as option value:

ts

gabrielluizsf (approved):

internal/tsoptions/tsconfigparsing.go
```diff
@@ -276,8 +268,7 @@ func isCompilerOptionsValue(option *CommandLineOption, value any) bool {
 			return reflect.TypeOf(value) == orderedMapType
 		}
 		if option.Kind == "enum" && reflect.TypeOf(value).Kind() == reflect.String {
-			_, ok := option.EnumMap().Get(strings.ToLower(value.(string)))
-			return ok || (option.DeprecatedKeys() != nil && option.DeprecatedKeys().Has(strings.ToLower(value.(string))))
+			return true
```
jakebailey, line 271: I am vaugely surprised that removing this "works"; I could have sworn we just added this to fix a bug somewhere.

haoqixu: this is because `convertJsonOption` failed to parse out an enum value and it returned the map key for an enum option. 🤔 
The removed patch here converts key to value so that the top-level enum options “works” while the value of an enum list option parsed by `convertJsonOptionOfListType` is a list of the map keys (like `lib`)

haoqixu: > this is because `convertJsonOption` failed to parse out an enum value and it returned the map key for an enum option. 🤔 The removed patch here converts key to value so that the top-level enum options “works” while the value of an enum list option parsed by `convertJsonOptionOfListType` is a list of the map keys (like `lib`)

This PR fixes `convertJsonOption` to return the map value for an enum option so that we don’t need this patch here and `convertJsonOptionOfListType` returns a list of map value for options like `lib` aligned with strada now

internal/tsoptions/tsconfigparsing.go
haoqixu: I think this is the bug you referred to in another comment. 🤔 @jakebailey

internal/tsoptions/tsconfigparsing.go
haoqixu: I think this is because `convertJsonOption` failed to call `convertJsonOptionOfEnumType` to parse enum options. The validation here should be done inside `convertJsonOptionOfEnumType`.

haoqixu: I deleted some of the previous comments, which I replied at around 6am, and I misread the line of code for the comments, so I misunderstood it 😅

testdata/baselines/reference/config/tsconfigParsing/reports errors for wrong type option and invalid enum value with json api.js
```diff
@@ -0,0 +1,145 @@
...
+				"removeComments": "should be a boolean",
+				"moduleResolution": "invalid value"
+			    }
+			}
+
+
+configFileName:: tsconfig.json
+CompilerOptions::
+{
+  "allowJs": null,
```
jakebailey: Please merge main; this all should drop away.

iisaduan (approved):

PR #595, opened 2025-03-14
Fix typos

honkkki:
Fix typos in:
- internal/ast/utilities.go
- internal/binder/binder.go
- internal/binder/binder_test.go


honkkki:

PR #596, opened 2025-03-14
Fix error type for unkonwn option

### LINKED ISSUE #491: erasableSyntaxOnly not working
Using `tsgo` in a project with `erasableSyntaxOnly` enabled leads to some very confusing output:

Image

I spent about an hour feeling like an idiot who didn't know how to use a CLI and wondering how Anders' command worked in the demo video, theorizing about .bashrc aliases with extra flags.

As soon as I removed `erasableSyntaxOnly` from my tsconfig, everything worked perfectly (save a few minor inference issues I'll follow up on) 🎉

I do wonder if, in an addition to supporting `erasableSyntaxOnly`, the CLI parser might fail more clearly if that's where things went awry here.

haoqixu:
`tsgo` incorrectly reports a `TS6369` error for an unknown option in tsconfig.json without location information.
This PR corrects the error type and adds location information about the option.

run with this `tsconfig.json`:
```json
{
  "compilerOptions": {
      "lemoveComments": true
  },
  "files": ["/tmp/1.ts"]
}
```

tsc 5.8.2 prints:
```
tsconfig.json:3:7 - error TS5025: Unknown compiler option 'lemoveComments'. Did you mean 'removeComments'?

3       "lemoveComments": true
        ~~~~~~~~~~~~~~~~


Found 1 error in tsconfig.json:3
```

`tsgo` on `main`:
```
error TS6369: Option '--build' must be the first command line argument.

Found 1 error.

exit status 1
```

`tsgo` with this PR:
```
tsconfig.json:3:7 - error TS5023: Unknown compiler option 'lemoveComments'.

3       "lemoveComments": true
        ~~~~~~~~~~~~~~~~


Found 1 error in tsconfig.json:3

exit status 1
```

Fixes #491


jakebailey:
This is probably correct and I believe fixes #491, but can you add a test for it? There should be a way to add one in a neighboring test file.

haoqixu:
> This is probably correct and I believe fixes #491, but can you add a test for it? There should be a way to add one in a neighboring test file.

Sure, I'll check it out and add a test.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -516,21 +517,24 @@ func convertOptionsFromJson[O optionParser](optionsNameMap map[string]*CommandLi
 	var errors []*ast.Diagnostic
 	for key, value := range jsonMap.Entries() {
 		opt, ok := optionsNameMap[key]
+		if !ok {
+			// !!! TODO?: support suggestion
+			errors = append(errors, ast.NewCompilerDiagnostic(diagnostics.Unknown_compiler_option_0, key))
+			continue
+		}
+
 		commandLineOptionEnumMapVal := opt.EnumMap()
```
haoqixu, line 526: This will panic with the new test case. Fixed.

internal/tsoptions/tsconfigparsing_test.go
```diff
@@ -471,6 +471,19 @@ var parseJsonConfigFileTests = []struct {
 			allFileList:    map[string]string{"/tsconfig.base.json": tsconfigWithExtendsAndConfigDir, "/src/index.ts": "", "/src/app.ts": "", "/node_modules/module.ts": "", "/dist/output.js": ""},
 		}},
 	},
+	{
```
haoqixu, line 474: I am not really sure if I understand how the test suite works correctly and if I added the test case to the right place. :thinking:

jakebailey: I think this is right...

PR #598, opened 2025-03-14
Remove unused parameters

honkkki:
Remove unused parameters: enclosingDeclaration


jakebailey (changes requested):
I don't think we should remove this right now; this parameter is needed for future code as our type printing is not complete yet.

PR #600, opened 2025-03-14
feature: Port emitter functions

KeoFoxy:
Ported:
- IsEmitBlocked
- getSourceMapFilePath
- getDeclarationEmitOutputFilePath
- getDeclarationEmitOutputFilePathWorker
- GetDeclarationEmitExtensionForPath

Added
- SetEmitBlocked
- BuildInfo struct for future usage

### DIFF
--- a/internal/compiler/buildInfo.go
+++ b/internal/compiler/buildInfo.go
@@ -1,0 +1,26 @@
+package compiler
+
+import (
+	"encoding/json"
+	"errors"
+)
+
+type BuildInfo struct {
+	Version string `json:"version"`
+}
+
+func GetBuildInfo(buildInfoFile string, buildInfoText string) (*BuildInfo, error) {
+	if buildInfoText == "" {
+		return nil, errors.New("empty buildInfoText")
+	}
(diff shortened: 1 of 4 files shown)


internal/compiler/buildInfo.go
KeoFoxy, line 1: Can I keep it for future use and scaling, or should I delete it?

jakebailey:
I'm not entirely convinced it's a good idea to port these functions now, when they aren't be used or tested, and may not actually be useful for future code.

KeoFoxy:
They were marked as TODOs so I decided to port them. So should I keep only used ones? What about BuildInfo? Remove it as well?

jakebailey:
@weswigham If you're nearing the completion of some dts emit, are the functions in this PR useful to you, or are you going to just do something from scratch?

weswigham:
Nah, only `getDeclarationEmitOutputFilePath` which I have a slightly different version of already.  Nothing I'm working on needs any of the rest of these

jakebailey:
Yeah, I'm not sure we need this PR right now, then.

PR #602, opened 2025-03-14
Improve octal literal diagnostic

frodi-karlsson:
<s>The [typescript code](https://github.com/microsoft/TypeScript/blob/main/src/compiler/scanner.ts#L1257-L1262) parses the digits as an octal int, but I thought it might be more straight forward to diverge and parse the strings here. </s>

Octal escape characters still have some diff, but the issue seems to happen in a different place so I thought it wouldn't fit well in the same PR.

Just some cheap baseline diff wins to get familiar with the project :)


internal/scanner/scanner.go
```diff
@@ -1662,7 +1662,20 @@ func (s *Scanner) scanNumber() ast.Kind {
 			} else {
 				s.tokenFlags |= ast.TokenFlagsOctal
 				s.tokenValue = "0o" + digits
-				s.errorAt(diagnostics.Octal_literals_are_not_allowed_Use_the_syntax_0, start, s.pos-start, digits)
+				withMinus := s.token == ast.KindMinusToken
```
jakebailey, line 1665: Now that we have the `jsnum` package, I'd rather we more directly port the original TS code:

```ts
tokenValue = "" + parseInt(tokenValue, 8);
tokenFlags |= TokenFlags.Octal;
const literal = (withMinus ? "-" : "") + "0o" + (+tokenValue).toString(8);
if (withMinus) start--;
error(Diagnostics.Octal_literals_are_not_allowed_Use_the_syntax_0, start, pos - start, literal);
return SyntaxKind.NumericLiteral;
```

We can now parse the int, though we don't support `toString(8)`, but that could be done with just `strconv` until jsnum gains support for specific bases.

jakebailey (changes requested):

internal/scanner/scanner.go
frodi-karlsson: Will do, thank you ✅

internal/jsnum/string.go
```diff
@@ -131,7 +131,7 @@ func tryParseInt(s string) (Number, bool) {
 			}
 			i, err = strconv.ParseInt(rest, 2, 64)
 			hasIntResult = true
-		case "0o", "0O":
+		case "0o", "0O", "0":
```
jakebailey: I don't think this works; `prefix` is guaranteed to be 2 characters.

frodi-karlsson: Sorry, will revert this. `Number("012")` is 12, not 10

internal/scanner/scanner.go
```diff
@@ -1660,9 +1660,14 @@ func (s *Scanner) scanNumber() ast.Kind {
 				s.tokenFlags |= ast.TokenFlagsContainsLeadingZero
 				fixedPart = digits
 			} else {
+				s.tokenValue = jsnum.FromString(digits).String()
 				s.tokenFlags |= ast.TokenFlagsOctal
-				s.tokenValue = "0o" + digits
-				s.errorAt(diagnostics.Octal_literals_are_not_allowed_Use_the_syntax_0, start, s.pos-start, digits)
+				withMinus := s.token == ast.KindMinusToken
+				literal := core.IfElse(withMinus, "-", "") + "0o" + s.tokenValue
```
jakebailey, line 1666: I am not sure this is exactly right; this will not print the number as octal, so only works for 0-8.

What I would do here is to refactor such that the result of `jsnum.FromString` is extracted out into a variable, convert it to an int64, then pass it into `strconv.FormatInt` asking for base 8.

frodi-karlsson: I'm not sure I'm with you... If I'm getting you right, I don't think that'll work:

```go
digits // Octal, but 0[octal representation] instead of 0o[octal representaion]
asNum := jsnum.FromString(digits)  // strconv.ParseInt(s, 10, 64), so it's "cast" as a base 10 number
literal := ... strconv.FormatInt(int64(asNum), 8) // Re-octalizes it, spitting out the wrong number
```
or in other words
```ts
"0o" + parseInt(Number("012"), 8) // Expected 0o12, got 0o10
```

jakebailey: Oh, no, I think I'm just wrong...

jakebailey (approved):

internal/scanner/scanner.go
frodi-karlsson: That's what Fridays are for! Thanks for taking the time to review

PR #604, opened 2025-03-14
docs: Add missing @returns tag to goBuildTags JSDoc

AriajSarkar:
This PR improves the JSDoc comment for the `goBuildTags()` function by adding a missing `@returns {string[]}` tag.

### Why this matters:
- Helps IDEs provide better autocomplete and type inference.
- Improves developer experience, especially when using `// @ts-check`.
- Enhances documentation clarity for future contributors.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -44,7 +44,7 @@ const defaultGoBuildTags = [
 
 /**
  * @param  {...string} extra
- * @returns
+ * @returns {string[]}
  */
 function goBuildTags(...extra) {
     const tags = new Set(defaultGoBuildTags.concat(extra));


Herebyfile.mjs
```diff
@@ -44,7 +44,7 @@ const defaultGoBuildTags = [
 
 /**
  * @param  {...string} extra
- * @returns
+ * @returns {string[]}
```
jakebailey, line 47: Is this not inferred already?

AriajSarkar: You're right — I thought it wasn’t being inferred in this context, but I’ll double-check and get back with more clarity.

AriajSarkar: I confirmed that while TypeScript can infer the return type, this tag helps with JS tooling and improves intellisense when using // @ts-check. Thought it might be worth adding for clarity, but happy to adjust if you prefer keeping it minimal.

jakebailey (approved):

PR #605, opened 2025-03-14
Fix more error baseline differences

ahejlsberg:
In this PR:

* Print conditional types, infer types, type parameter constraints.
* Fix error spans for unused parameters.
* Change name of errorType to `any`.


internal/checker/checker.go
```diff
@@ -878,7 +878,7 @@ func NewChecker(program Program) *Checker {
 	c.autoType = c.newIntrinsicTypeEx(TypeFlagsAny, "any", ObjectFlagsNonInferrableType)
 	c.wildcardType = c.newIntrinsicType(TypeFlagsAny, "any")
 	c.blockedStringType = c.newIntrinsicType(TypeFlagsAny, "any")
-	c.errorType = c.newIntrinsicType(TypeFlagsAny, "error")
+	c.errorType = c.newIntrinsicType(TypeFlagsAny, "any")
```
jakebailey, line 881: Are we going to have to extend the intrinsic type to be able to store both? Or maybe just special case `error` for certain baselines later? We don't always print error types as any.

ahejlsberg: We distinguish `errorType` and `anyType` by object identity and both always had the intrinsic name `any` in the old compiler. I used the name `error` just to make it easier to recognize the type in the early phases of porting the checker.

jakebailey: I'm specifically talking about this part of Strada's type baselines: https://github.com/microsoft/TypeScript/blob/0aac72020ee8414218273f654eb7ce1dc2dd0d6b/src/harness/typeWriter.ts#L278

ahejlsberg: I'm not understanding what you're asking.

jakebailey: The intrinsic name is now set to `any`. What I'm saying is that when we add type baselines, the intrinsic name not being `error` means that we can't do the same thing the above code says.

It's not really a problem right now; we'll fix it up once we get type baselines. Nothing to do right now, sorry if that wasn't clear!

ahejlsberg: Ah, shoot, just realizing that the intrinsic name actually _was_ `error` in the old compiler. So I guess we flipped from one set of baselines being different to another. We probably should flip the name back to `error`. Sorry for the confusion.

jakebailey: Right, though I think we can get away with making the current type printer just always print it as `"any"` for errors and so on right now.

jakebailey: It's only the type baselines that would have ever shown the word "error", along with our own debugging. Which, is valuable of course.

Eigilak:
Der bliver kodet derudaf!

PR #606, opened 2025-03-14
Enable type baselines

jakebailey:
This enables type baselining. I've modified the diffing to ignore the lines of the type baselines that have to do with node reuse, which we do not yet handle in this codebase.


Zaid-maker:
Team is working really hard! Appreciate that 1Milllion+ additions DAMN

jakebailey:
This is now passing!

jakebailey:
For reference, there are 12464 types baselines, and only 6825 diff files. And that's despite union ordering / member ordering changes.

gabritto (approved):

jakebailey:
`hugeDeclarationOutputGetsTruncated.types` is 22.7MB 😨

jakebailey:
I added in a 1,000,000 truncation limit like we previously had, for now. That limits the length to 4MBish.

RyanCavanaugh (approved):
LGTM for the portions I could read before my ~computer melted~ hard drive overheated

jakebailey:
I'm going to quick see if I can port some code to fix the `T : T` thing, but if it's not quick I'll just merge as-is.

EDIT: with this type printer, I made it worse 😅

PR #608, opened 2025-03-14
More faithful port of duplicate declarations check

Andarist:


testdata/baselines/reference/submodule/compiler/checkTypePredicateForRedundantProperties.errors.txt.diff
```diff
@@ -0,0 +1,15 @@
...
+@@= skipped -0, +0 lines =@@
+-checkTypePredicateForRedundantProperties.ts(1,35): error TS2300: Duplicate identifier 'a'.
+ checkTypePredicateForRedundantProperties.ts(1,46): error TS2300: Duplicate identifier 'a'.
+ 
+ 
+-==== checkTypePredicateForRedundantProperties.ts (2 errors) ====
++==== checkTypePredicateForRedundantProperties.ts (1 errors) ====
+     function addProp2(x: any): x is { a: string; a: string; } {
+-                                      ~
+-!!! error TS2300: Duplicate identifier 'a'.
```
Andarist, line 12: This is a regression - but I think it's acceptable at the current stage. The problem is that the old compiler is inconsistent in this regard: [TS playground](https://www.typescriptlang.org/play/?ts=5.9.0-dev.20250314#code/C4TwDgpgBAKu0F4oG8BQUoEMBcUDOwATgJYB2A5gNzpa4EkXUC+1qAxgDaZ55QCCKGpihIARAEZR1DMLEAmKaiaogA). 

Likely, different functions were responsible for this in the old compiler but now a single function is performing this check so the behavior gets aligned to reporting the error on all non-first declarations

ahejlsberg:
I made some intended changes to the duplicate object type members check in the port. I don't think we want the old behavior, it was inconsistent with other duplicate declaration reporting and it was oddly sensitive to declaration order. In particular, the new code consistently reports errors on _all_ declarations, which is what we do elsewhere.

Andarist:
Superseded by https://github.com/microsoft/typescript-go/pull/682

Issue #609, opened 2025-03-14
Reintroduce Preemptive Caching of CouldContainTypeVariables


zwade:
When playing around with Typescript Go on an implementation of a JavaScript VM in Typescript types ([zwade/jsits](https://github.com/zwade/jsits)) I noticed a major regression. 

Where the native Typescript compiler could compile the project in about 3 seconds, the Golang implementation was hitting the type instantiation too deep / possibly infinite error. 

After a few hours of tracing the bug, I found the issue to be the fact that `couldContainTypeVariablesWorker` was returning `true` in places that the Typescript equivalent was returning false. After further investigation, I realized it was due to missing logic in `getObjectTypeInstantiation` from the Typescript version.

I tried reintroducing it [here](https://github.com/microsoft/typescript-go/compare/main...zwade:typescript-go:feat/fix-caching-bug?expand=1) and found `tsgo` behaved exactly as `tsc`.

I'm happy to issue the above as a pull-request, but I wanted to see first if this truly is a regression or if this code path was removed for some reason (e.g. the caching logic is erroneous). 

Thanks!

RyanCavanaugh:
Do you see the problem in TS 5.6? It may be from a PR that hasn't been ported yet

zwade:
It looks like as of 5.6.3 the logic in `tsc` is the same as 5.7-beta~9. A quick blame of the `checker.ts` shows that this code was introduced in `3cfd6d2b872` on 2023-06-06.

This almost looks like an intentional omission, especially since it's for performance reasons so it shouldn't affect soundness. With that said, I can't think of an obvious reason why it should be in `tsc` but not `tsgo`.

RyanCavanaugh:
Cool, please go ahead with the PR and I'll have Anders take a look. Thanks!

PR #612, opened 2025-03-14
Re-introduce preemptive caching of CouldContainTypeVariables

zwade:
Addresses #609 

## Summary

The base Typescript implementation of `getObjectTypeInstantiation` includes a mechanism to preemptively cache `CouldContainTypeVariable` when it instantiates a type that it knows can have no type variables. 

The golang implementation is currently missing this optimization, which can trigger different control flow and a higher number of calls to `instantiateAliasedType` than `tsc`, which in turn leads to differing behavior with regard to instantiation depth restrictions.

This PR reintroduces this logic into the golang codebase.

## Reference

The original Typescript can be found [here](https://github.com/microsoft/TypeScript/blob/52c59dbcbee274e523ef39e6c8be1bd5e110c2f1/src/compiler/checker.ts#L20375)

For simplicity, the relevant code is this:

```ts
const resultObjectFlags = getObjectFlags(result);
if (result.flags & TypeFlags.ObjectFlagsType && !(resultObjectFlags & ObjectFlags.CouldContainTypeVariablesComputed)) {
    const resultCouldContainTypeVariables = some(typeArguments, couldContainTypeVariables); // one of the input type arguments might be or contain the result
    if (!(getObjectFlags(result) & ObjectFlags.CouldContainTypeVariablesComputed)) {
        // if `result` is one of the object types we tried to make (it may not be, due to how `instantiateMappedType` works), we can carry forward the type variable containment check from the input type arguments
        if (resultObjectFlags & (ObjectFlags.Mapped | ObjectFlags.Anonymous | ObjectFlags.Reference)) {
            (result as ObjectFlagsType).objectFlags |= ObjectFlags.CouldContainTypeVariablesComputed | (resultCouldContainTypeVariables ? ObjectFlags.CouldContainTypeVariables : 0);
        }
        // If none of the type arguments for the outer type parameters contain type variables, it follows
        // that the instantiated type doesn't reference type variables.
        // Intrinsics have `CouldContainTypeVariablesComputed` pre-set, so this should only cover unions and intersections resulting from `instantiateMappedType`
        else {
            (result as ObjectFlagsType).objectFlags |= !resultCouldContainTypeVariables ? ObjectFlags.CouldContainTypeVariablesComputed : 0;
        }
    }
}
```

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -20451,7 +20451,7 @@ func (c *Checker) getObjectTypeInstantiation(t *Type, m *TypeMapper, alias *Type
 			result = c.instantiateAnonymousType(target, newMapper, newAlias)
 		}
 		data.instantiations[key] = result
-		if (result.flags&TypeFlagsObjectFlagsType != 0) && (result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0) {
+		if result.flags&TypeFlagsObjectFlagsType != 0 && result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0 {
 			// if `result` is one of the object types we tried to make (it may not be, due to how `instantiateMappedType` works), we can carry forward the type variable containment check from the input type arguments
 			resultCouldContainObjectFlags := core.Some(typeArguments, c.couldContainTypeVariables)
 			if result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0 {
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -20451,6 +20451,20 @@ func (c *Checker) getObjectTypeInstantiation(t *Type, m *TypeMapper, alias *Type
 			result = c.instantiateAnonymousType(target, newMapper, newAlias)
 		}
 		data.instantiations[key] = result
+		if (result.flags&TypeFlagsObjectFlagsType != 0) && (result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0) {
+			// if `result` is one of the object types we tried to make (it may not be, due to how `instantiateMappedType` works), we can carry forward the type variable containment check from the input type arguments
+			resultCouldContainObjectFlags := core.Some(typeArguments, c.couldContainTypeVariables)
+			if result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0 {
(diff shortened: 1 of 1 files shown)


jakebailey:
It's very concerning that fixing this does not change any test.

jakebailey:
But, maybe this barely came in before the types baselines change...

RyanCavanaugh:
https://github.com/microsoft/TypeScript/pull/10347/commits/a1a87254565b6b2eec334b2a280463ee7139449a

zwade:
Thanks for taking a look at this! I updated the changeset with the appropriate comments, and reintroduced the check on line 20457. 

Just to double check on that, I had originally thought it redundant but since `core.Some(typeArguments, c.couldContainTypeVariables)` could theoretically modify that flag, my understanding is we do need to preserve that check as well.

Also re: @jakebailey's comment, I'm not entirely sure how one would test for this behavior. I'm not as familiar with typescript's test suite, but I think you would need additional instrumentation to check for the number of times `getObjectTypeInstantiation` triggers an `instantiateTypeWorker` on certain adversarial inputs. The only observable symptom is an overeager raise of `Type_instantiation_is_excessively_deep_and_possibly_infinite`

jakebailey:
My concern is more joking than anything; this is absolutely cursed behavior to depend on so I'm not surprised it's not tested 😅

zwade:
> this is absolutely cursed behavior to depend on so I'm not surprised it's not tested 😅

Haha yeah I suspect most people are not writing JavaScript parsers in the type system 😶‍🌫️

internal/checker/checker.go
```diff
@@ -20451,6 +20451,20 @@ func (c *Checker) getObjectTypeInstantiation(t *Type, m *TypeMapper, alias *Type
 			result = c.instantiateAnonymousType(target, newMapper, newAlias)
 		}
 		data.instantiations[key] = result
+		if (result.flags&TypeFlagsObjectFlagsType != 0) && (result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0) {
```
jakebailey, line 20454:
```suggestion
		if result.flags&TypeFlagsObjectFlagsType != 0 && result.objectFlags&ObjectFlagsCouldContainTypeVariablesComputed == 0 {
```

jakebailey (approved):
LGTM besides a style nit.

PR #614, opened 2025-03-14
add -pprofDir profiling to LSP

mjames-c:
add support for getting pprof profiles of LSP.

context: https://discord.com/channels/1298076816732328006/1349171342628618374

### DIFF
--- a/internal/pprof/pprof.go
+++ b/internal/pprof/pprof.go
@@ -1,0 +1,63 @@
+package pprof
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime/pprof"
+)
+
+type profileSession struct {
+	cpuFilePath string
+	memFilePath string
+	cpuFile     *os.File
+	memFile     *os.File
+	logWriter   io.Writer
+}
+
+// BeginProfiling starts CPU and memory profiling, writing the profiles to the specified directory.
+func BeginProfiling(profileDir string, logWriter io.Writer) *profileSession {
+	if err := os.MkdirAll(profileDir, 0o755); err != nil {
+		panic(err)
+	}
+
+	pid := os.Getpid()
+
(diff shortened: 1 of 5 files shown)


internal/pprof/pprof.go
```diff
@@ -0,0 +1,59 @@
...
+	err := pprof.Lookup("allocs").WriteTo(p.memFile, 0)
+	if err != nil {
+		panic(err)
+	}
+
+	p.cpuFile.Close()
+	p.memFile.Close()
+
+	fmt.Printf("CPU profile: %v\n", p.cpuFilePath)
+	fmt.Printf("Memory profile: %v\n", p.memFilePath)
```
jakebailey: These should change to `fmt.Fprintf(os.Stderr, ...)`, otherwise the LS will break due to us writing at the end to the JSON-RPC pipe.

internal/pprof/pprof.go
```diff
@@ -0,0 +1,59 @@
...
+	pprof.StopCPUProfile()
+	err := pprof.Lookup("allocs").WriteTo(p.memFile, 0)
+	if err != nil {
+		panic(err)
+	}
+
+	p.cpuFile.Close()
+	p.memFile.Close()
+
+	fmt.Printf("CPU profile: %v\n", p.cpuFilePath)
```
mjames-c: TODO: log this to VS Code `typescript-go` output channel (stderr) in LSP case

jakebailey: This should be as simple as using `Stderr`; I think we'll need to do that before merging.

jakebailey (approved):

mjames-c:
@microsoft-github-policy-service agree Canva

mjames-c:
@microsoft-github-policy-service agree company="Canva"

jakebailey (approved):

PR #617, opened 2025-03-15
Remove unnecessary configurations from .golangci.yml

karamaru-alpha:
The configuration for wrapcheck was left behind even though it was disabled, so I have removed it.

### Appendix
- [Add FS wrapper to correctly emulate case insensitivity in testing #112](https://github.com/microsoft/typescript-go/pull/112)
  - The PR disabled wrapcheck.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,10 +64,6 @@ linters-settings:
     customlint:
       type: module
 
-  wrapcheck:
-    ignorePackageGlobs:
-      - encoding/*
-      - github.com/go-json-experiment/*
 #   revive:
 #     enable-all-rules: true
 #     rules:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -73,3 +73,8 @@ linters-settings:
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
+
+  exclude:
+    - '^could not import'
+    - '^: #'
+    - 'imported and not used$'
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -77,8 +73,3 @@ linters-settings:
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
-
-  exclude:
```
jakebailey: My intent here was to attempt to silence golangci-lints typecheck errors but it doesn't work :(

I wish they'd let them be silenced. Makes the editor quite frustrating when a package can't be built as random other packages get linted...

karamaru-alpha: Thank you for always reviewing so quickly!

> I wish they'd let them be silenced. Makes the editor quite frustrating when a package can't be built as random other packages get linted...

It seems like there are no typecheck errors in CI at the moment, but are you excluding them because they become annoying when running lint in the IDE during development?
If that's the case, I’m thinking of reverting the following commit.
[refactor: golangci-lint delete unnecessary issues](https://github.com/microsoft/typescript-go/pull/617/commits/deacd90165767a1619ea72cf68ff6934975db03a)

karamaru-alpha: Ah, since typecheck errors are compile errors, I understand that it would be annoying for us in the CI environment if each linter warns about them individually.
cf. https://github.com/golangci/golangci-lint/issues/2912#issuecomment-1149995159

Got it! I’ll revert the  commit. 🙆

jakebailey: Yeah, that's the one. There's probably no point in leaving the ignores if they don't work, though. I've had it work before, but probably it was years ago...

jakebailey (approved):

Issue #536, opened 2025-03-12
Parameter properties are not supported


arnaudNYC:
Code being compiled:
```
class Project {
  constructor(
    public name: string,
    public deadline: Date,
    public budget: number
  ) {}

  extendDeadline(days: number): void {
    this.deadline.setDate(this.deadline.getDate() + days);
  }
}
const p = new Project("new app", new Date(), 1000);
console.log(p);
```

Expected:
```
class Project {
    name;
    deadline;
    budget;
    constructor(name, deadline, budget) {
        this.name = name;
        this.deadline = deadline;
        this.budget = budget;
    }
    extendDeadline(days) {
        this.deadline.setDate(this.deadline.getDate() + days);
    }
}
const p = new Project("new app", new Date(), 1000);
console.log(p);

/* outputs:

Project {
  name: 'new app',
  deadline: 2025-03-12T19:10:10.994Z,
  budget: 1000
}

*/

```

Actual:
```
class Project {
    constructor(name, deadline, budget) { }
    extendDeadline(days) {
        this.deadline.setDate(this.deadline.getDate() + days);
    }
}
const p = new Project("new app", new Date(), 1000);
console.log(p);

/* outputs:
Project {}
*/

```

Ref doc: https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

arnaudNYC:
Fixed!

PR #619, opened 2025-03-15
Fix: Support parameter properties in class constructors

zanuka:
This PR addresses issue #536 where parameter properties in TypeScript class constructors
were not properly supported during transformation. Parameter properties like 
`constructor(public name: string)` are now correctly transformed into class properties
and assignments in the constructor body.

The fix modifies the constructor declaration visitor to properly handle parameter
property modifiers and generate the appropriate class properties.

Tests have been run and verified that the transformation now correctly handles all
parameter property cases.

### DIFF
--- a/internal/transformers/runtimesyntax_test.go
+++ b/internal/transformers/runtimesyntax_test.go
@@ -432,6 +432,29 @@ func TestParameterPropertyTransformer(t *testing.T) {
         this.x = x;
     }
 }`},
+		{title: "multiple parameter properties", input: `class Project {
+  constructor(
+    public name: string,
+    public deadline: Date,
+    public budget: number
+  ) {}
+
+  extendDeadline(days: number): void {
+    this.deadline.setDate(this.deadline.getDate() + days);
+  }
+}`, output: `class Project {
+    name;
+    deadline;
+    budget;
+    constructor(name, deadline, budget) {
+        this.name = name;
+        this.deadline = deadline;
+        this.budget = budget;
+    }
+    extendDeadline(days) {
+        this.deadline.setDate(this.deadline.getDate() + days);
+    }
+}`},
 	}
 
 	for _, rec := range data {
--- a/internal/transformers/runtimesyntax.go
+++ b/internal/transformers/runtimesyntax.go
@@ -810,7 +810,35 @@ func (tx *RuntimeSyntaxTransformer) visitClassExpression(node *ast.ClassExpressi
 
 func (tx *RuntimeSyntaxTransformer) visitConstructorDeclaration(node *ast.ConstructorDeclaration) *ast.Node {
 	modifiers := tx.visitor.VisitModifiers(node.Modifiers())
-	parameters := tx.emitContext.VisitParameters(node.ParameterList(), tx.visitor)
+	
+	var nonParamPropertyParams []*ast.ParameterDeclaration
+	for _, param := range node.ParameterList().Nodes {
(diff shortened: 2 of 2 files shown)


zanuka:
@microsoft-github-policy-service agree [company="Zanuka Labs"]

zanuka:
@microsoft-github-policy-service agree company="Zanuka Labs"

zanuka:
@microsoft-github-policy-service agree company="Zanuka Labs"

BansalShivanshu:
Is this a direct port from TS codebase or new addition to the codebase? Looking at issue #536 seems like constructor body transformation from constructor parameters to class properties hasn't been ported yet.

zanuka:
@BansalShivanshu This is not a direct port of specific code from the TypeScript codebase, but rather a new implementation following the same transformation pattern that TypeScript uses.

Since the TS compiler handles parameter properties through its transformation pipeline, it's my understanding that this core pattern involves:

1. Identifying parameters with property modifiers (`public`, `private`, `protected`, `readonly`)
2. Creating property declarations in the class
3. Generating property assignments in the constructor body

I noticed that issue #536 correctly identified that this functionality wasn't properly implemented, so this PR fills this gap by adding logic to:

- Correctly detect parameter properties using `ast.IsParameterPropertyDeclaration()`
- Strip property modifiers from parameters in the transformed constructor
- Generate the appropriate assignments in the constructor body

It follows TypeScript's pattern conceptually, though the specific code structure differs due to the language differences (Go vs TypeScript) and the architectural differences between the codebases.

I think all edge cases have been handled including:
- Multiple parameter properties in a single constructor
- Mixed regular parameters and parameter properties
- Parameter properties with type annotations
- Constructor super calls and other existing functionality

Happy to make adjustments as needed if anything was missed. I fixed the extra blank lines that had caused the initial check fail as well.

internal/transformers/runtimesyntax_test.go
```diff
@@ -431,6 +431,29 @@ func TestParameterPropertyTransformer(t *testing.T) {
     constructor(x) {
         this.x = x;
     }
+}`},
+		{title: "multiple parameter properties", input: `class Project {
```
frodi-karlsson, line 435: Running this test without the changes on up to date main, it also passes. I'm not so sure this is fixing anything, as there doesn't seem to be an issue at first look. Are you experiencing something different?

Theres of course still an issue with a super call's position as it relates to the parameter initializations, which I think just needs a minimal fix (#547)

frodi-karlsson (commented):
I believe the issue was actually fixed 4 days ago here: https://github.com/microsoft/typescript-go/commit/8b954669a61392a3137904644a5d7e632d0a3c40#diff-602f9f2ee92480e93b3b71e218795f3b947dd7745781272ef927d1f2dd431ec1R718

internal/transformers/runtimesyntax_test.go
zanuka: Thanks for the comments, much appreciated. After further investigation and testing, I've confirmed that this PR is redundant. The parameter property transformation functionality is already properly implemented in the codebase through commit 8b95466 (PR #479).

I checked the following: 
1. All tests pass on the main branch without these changes
2. The existing implementation already handles:
   - Parameter property detection
   - Modifier stripping
   - Property assignments in constructor
   - Super call handling
   - Multiple parameter properties

I'll close this PR since it would be duplicating existing functionality.

Issue #623, opened 2025-03-15
types: [] in tsconfig.json is treated as "all visible types"


DoctorGester:
In tsc:

1. Unspecified `types`: all "visible" types (from node_modules)
2. `types: []`: no extra types

in tsgo (tsgo.exe)

1. Unspecified `types`: all "visible" types (from node_modules)
2. `types: []`: also all visible types, no difference from unspecified

The problem is here:

https://github.com/microsoft/typescript-go/blob/main/internal/compiler/module/resolver.go#L1799

```go
if options.Types != nil {
    return options.Types
}
```

Because here https://github.com/microsoft/typescript-go/blob/main/internal/tsoptions/parsinghelpers.go#L26

```go
func parseStringArray(value any) []string {
	if arr, ok := value.([]any); ok {
		var result []string
		for _, v := range arr {
			if str, ok := v.(string); ok {
				result = append(result, str)
			}
		}
		return result
	}
	return nil
}
```

`var result []string` is a nil slice, which is returned if there are no elements in an array.

I have never programmed in Go so I didn't submit a pull request since I don't know if replacing it with a `result := []string{}` is a good way to do it, hopefully it's an easy fix.

AndrewHarrisSPU:
> I have never programmed in Go so I didn't submit a pull request since I don't know if replacing it with a `result := []string{}` is a good way to do it, hopefully it's an easy fix.

Good catch on `nil` versus empty slice. The length of `arr` seems like will be the length of `result`. We can allocate once with a 3-argument`make(type, length, capacity)`, rather than incur the allocation and relocation costs of dynamic resizing: `result := make([]string, 0, len(arr))`.

PR #624, opened 2025-03-15
fixed empty types array which allowed all visible types - unwanted be…

Manny-S-04:
## This PR attempts to target issue #623 : types: [] in tsconfig.json is treated as "all visible types" 

I believe the issue is that the result being nil was the cause of allowing all types so initializing it like this seems to have fixed it, i have tested it using my own test cases so if anyone can break it that would be amazing

for example i now get an error when the types array is present:

// tsconfig.json:
{
    "compilerOptions": {
      "types": []
    }
}

//foo.ts
const x:string = process.version; // this is from npm @types/node
                           ^ Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.ts(2591)

before this line would not throw an error


Manny-S-04:

internal/tsoptions/parsinghelpers.go
```diff
@@ -24,16 +24,16 @@ func parseTristate(value any) core.Tristate {
 }
 
 func parseStringArray(value any) []string {
+	var result []string
 	if arr, ok := value.([]any); ok {
-		var result []string
```
jakebailey, line 28: It seems like the fix could have just been `result := make([]string, 0, len(arr))` here, leaving the rest of the code the same.

jakebailey: That being said, the new code isn't `nil` preserving, so I am wondering if something else is wrong upstream of this as well.

Manny-S-04: Is it ok if i investigate further? im not sure how to test if this single line change is fixing the empty types array issue though haha

jakebailey: Well, if you revert your change and then replace this one line with my suggestion, the code is semantically the same, if that's what you're concerned about.

But, I'm more confused why a `[]any(nil)` input being remapped to `[]string{}` (zero length but non-nil) isn't actually a bug. But that's required for this test, so it implies whoever is generating the slice isn't generating nil for "not present"...

jakebailey: I spent some time trying to look at this myself, our distinguishing of empty slics, nil slices, nil any, are super super weird.

Manny-S-04: im going through it right now too actually, just finished my shift. i am just debugging with panics everywhere to see why types is being set to an empty array even if its not present in the config

Manny-S-04: Jake since this code is a potential fix and im unable to find anything significant :( im going to put this in my backlog and go work on something else
if you need me to investigate or fix anything please let me know !

jakebailey:
Thanks for sending this; this turned out to be more invasive to fix in total, and I've sent #706 for it.

PR #625, opened 2025-03-15
Fix binder inStrictMode initialization

frodi-karlsson:
`bindSourceFile` was not making use of `alwaysStrict` in its call to `file.BindOnce`. I'm assuming it's not intentional to omit this initialization, as it exists in strada


internal/binder/binder.go
```diff
@@ -117,6 +117,7 @@ func bindSourceFile(file *ast.SourceFile, options *core.CompilerOptions) {
 		b.file = file
 		b.options = options
 		b.languageVersion = options.GetEmitScriptTarget()
+		b.inStrictMode = options.AlwaysStrict.IsTrue()
```
jakebailey, line 120: This will have to be marked as a source affecting option on CompilerOptions, since it's used here.

frodi-karlsson: I'm sorry I might not be understanding you, do you mean add it to that struct? If so, it's already there:
https://github.com/microsoft/typescript-go/blob/main/internal/core/compileroptions.go#L271

jakebailey: Aha, so it is!

jakebailey (approved):

PR #627, opened 2025-03-16
Add VersionMajorMinor and Version tests

DouglasPimentel:
This pull request introduces a unit test to verify that the Version Major Minor and Version constants are correctly defined.

### DIFF
--- a/internal/core/version_test.go
+++ b/internal/core/version_test.go
@@ -1,0 +1,24 @@
+package core
+
+import "testing"
+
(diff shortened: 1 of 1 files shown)


DouglasPimentel:

jakebailey:
Thanks, but I don't think we should do this; the version number is tested by CLI tests via `--version`, and otherwise is just copying the constant to another file and asserting it's hardcoded by both.

PR #629, opened 2025-03-16
chore: change empty interface to using any predefine type

iporsut:
Using predefined `any` for empty interface

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -516,7 +516,7 @@ type Program interface {
 	GetResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile
 }
 
-type Host interface{}
+type Host any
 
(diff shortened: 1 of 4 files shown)


iporsut:
> @microsoft-github-policy-service agree

iporsut:

internal/checker/checker.go
```diff
@@ -516,7 +516,7 @@ type Program interface {
 	GetResolvedModule(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile
 }
 
-type Host interface{}
+type Host any
```
jakebailey: This is intentional; the host will eventually contain methods.

iporsut: OK, I set it back to empty interface.

jakebailey (approved):

PR #633, opened 2025-03-16
improve stringutil

kangkona:


internal/stringutil/util.go
```diff
@@ -112,7 +113,7 @@ func GuessIndentation(lines []string) int {
 	const MAX_SMI_X86 int = 0x3fff_ffff
 	indentation := MAX_SMI_X86
 	for _, line := range lines {
-		if len(line) == 0 {
+		if strings.TrimSpace(line) == "" {
```
jakebailey: Is this actually faster? Won't it also loop the same number of times as the code below, just now one extra time?

kangkona: It's not to be faster, but to deal with more blank situations. I'll modify it.

jakebailey: We don't really want to modify this; this is ported code that should behave like it used to for compatibility.

jakebailey (approved):

PR #635, opened 2025-03-16
feat: nvmrc node version

Chiliseo:
Hi every one, i use nvm for every project i think is awesome set .nvmrc version and only run nvm use or nvm install to set current version for running project


jakebailey:
Thanks, but I don't think this is a good idea; we don't need a fixed version, and a lot of us use 22 or even 23.

PR #636, opened 2025-03-16
Port part of tsc help from strada

haoqixu:
This PR partially ports `tsc -h` from Strada, with unimplemented sections marked as TODOs for future PRs.


internal/execute/tsc.go
```diff
@@ -36,12 +36,15 @@ func executeCommandLineWorker(sys System, cb cbType, commandLine *tsoptions.Pars
 
 	if commandLine.CompilerOptions().Init.IsTrue() ||
 		commandLine.CompilerOptions().Version.IsTrue() ||
-		// commandLine.CompilerOptions().Help != nil ||
-		// commandLine.CompilerOptions().All != nil ||
 		commandLine.CompilerOptions().Watch.IsTrue() && commandLine.CompilerOptions().ListFilesOnly.IsTrue() {
 		return ExitStatusNotImplemented, nil
 	}
 
+	if commandLine.CompilerOptions().Help.IsTrue() || commandLine.CompilerOptions().All.IsTrue() {
```
jakebailey, line 46: Once the other PR is merged, can you split up the "not implemented" cases above into the "correct" order from Strada? That way we don't fill them in but mistakenly check them in the wrong order.

internal/tsoptions/commandlineoption.go
```diff
@@ -65,6 +65,10 @@ type CommandLineOption struct {
 	ElementOptions map[string]*CommandLineOption
 }
 
+func (o *CommandLineOption) ShowInSimplifiedHelpView() bool {
+	return o.showInSimplifiedHelpView
```
jakebailey: Probably I'd just make this field exported.

internal/execute/tsc.go
```diff
@@ -35,9 +35,7 @@ func executeCommandLineWorker(sys System, cb cbType, commandLine *tsoptions.Pars
 	}
 
 	if commandLine.CompilerOptions().Init.IsTrue() ||
-		// commandLine.CompilerOptions().Help != nil ||
-		// commandLine.CompilerOptions().All != nil ||
-		commandLine.CompilerOptions().Watch.IsTrue() && commandLine.CompilerOptions().ListFilesOnly.IsTrue() {
+		commandLine.CompilerOptions().Version.IsTrue() {
```
jakebailey: This seems like an oops; version is true is checked below.

internal/execute/tsc.go
haoqixu: No problem. Rebased and updated.

testdata/baselines/reference/tsc/commandLine/does-not-add-color-when-NO_COLOR-is-set.js
```diff
@@ -151,3 +153,36 @@ ParsedCommandLine::{
 Output::
 Version 7.0.0-dev
```
jakebailey, line 29: This version output seems to be full duplicated.

haoqixu: https://github.com/microsoft/TypeScript/blob/release-5.8/tests/baselines/reference/tsc/commandLine/does-not-add-color-when-NO_COLOR-is-set.js#L21

It seems that tsc prints the same duplicated version in this case.

testdata/baselines/reference/tsc/commandLine/does-not-add-color-when-NO_COLOR-is-set.js
```diff
@@ -151,3 +153,36 @@ ParsedCommandLine::{
...
```
haoqixu, line 30: I've confirmed that this extra newline is added by the call of `sys.EndWrite`.

internal/core/compileroptions.go
```diff
@@ -134,6 +134,8 @@ type CompilerOptions struct {
 	Watch               Tristate `json:"watch,omitzero"`
 	ShowConfig          Tristate `json:"showConfig,omitzero"`
 	TscBuild            Tristate `json:"tscBuild,omitzero"`
+	Help                Tristate `json:"help"`
+	All                 Tristate `json:"all"`
```
jakebailey:
```suggestion
	Help                Tristate `json:"help,omitzero"`
	All                 Tristate `json:"all,omitzero"`
```

jakebailey (approved):
I think this seems like a good start for sure.

PR #637, opened 2025-03-16
Set implied node format during source file creation

frodi-karlsson:
Here I introduce setting ~~`SourceFile.ImpliedNodeFormat`~~ `SourceFileMetaData.ImpliedNodeFormat` as well as a field implied in commented code, `PackageJsonScope`. The location of `ImpliedNodeFormat` is as consistent as I could make it, considering some program flow is different in `parseSourceFile` etc.

In doing this, some things were unlocked like commented synthetic default code, and `packageJson.type` module detection, ~~both of which delivered some nice diff improvements.~~ (omitted for scope)

That being said, I'm not confident in making such a large change, but felt it was better to risk pushing it up and being told to sod off rather than saying "maybe I can fix synthetic defaults once the team implements this property" 😄


frodi-karlsson:
Weird, can't reproduce the failure. Probably this change is not concurrency safe somehow...

jakebailey:
I haven't looked at much of the PR, but the main thing to consider is that ASTs are reused both in tests and in the LS, so we can't store anything on the AST that could differ between different uses. The fact that you have fields set by the compiler host is the red flag; that info needs to be stored out of band since it can differ depending on context.

frodi-karlsson:
I see, thanks a lot for the direction.

I'll leave this drafted for now if that's alright to perhaps give it another shot

jakebailey:
Agh, clearly I should have read what you had beforehand. That's what I get for commenting from my phone early in the morning.

What you had previously was probably actually fine; having it as an argument to the `GetSourceFile` option was probably actually okay, it may just have been that you did not add its info to `sourceFileCacheKey` in testing.

Having it be a part of the file loader unfortunately makes less sense; by the time `GetSourceFile` returns, it's no longer safe to concurrently modify the source file in any way.

jakebailey:
My commentary was actually to do things differently entirely and store the implied format on `Program` per-file, then provide access to it via a host method. That would also work, and probably work better than the initial state as well given it'd allow for more AST reuse.

frodi-karlsson:
Thank you for both the patience and guidance, I'm very grateful to get so much feedback.

It seems quite difficult to decouple `ImpliedNodeFormat` from the `SourceFile` and into `Program`, as it's used in places where that wouldn't be available currently, such as `commonjsmodule.go/shouldTransformImportCallWorker`(vicariously through `ast.GetEmitModuleFormatOfFileWorker`) or in `externalmoduleinfo.go/createExternalHelpersImportDeclarationIfNeeded`.

I fear I may simply be in too deep water here, and that I should leave this task to the pros. Looking at the git blames, it looks like Ron Buckton might be working his way toward this problem, and I might just be making noise where there's a plan conceptualized somewhere anyway

frodi-karlsson:
I'll close this PR, and keep keeping busy with smaller tasks until I have better understanding of the bigger picture, so as not to waste valuable review mana in these hectic times :)

jakebailey:
I don't think you were too far off at all; I might try and resurrect this later, if you don't want to keep hacking on it.

frodi-karlsson:
How motivating! I couldn't find any nice way to store it outside of the source file, but I think I managed to clean up the first "probably okay" version a bit, with the cache key fixed.

The diff is a bit less gratifying because I did reduce scope a bit into just `ImpliedNodeFormat`, as it should've been from the start

.golangci.yml
```diff
@@ -64,6 +64,10 @@ linters-settings:
     customlint:
       type: module
 
+  wrapcheck:
```
jakebailey: Uh oh, seems like you may have had a bad merge with main?

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -355,22 +356,29 @@ type cachedCompilerHost struct {
...
+		fileName             string
+		path                 tspath.Path
+		languageVersion      core.ScriptTarget
+		packageJsonTypeValue string
+		text                 string
+	}
+
+	packageJsonTypeValue := ""
+	if packageJsonScope != nil {
+		packageJsonTypeValue = packageJsonScope.Contents.Type.Value
```
jakebailey: I'm noticing that only `Type.Value` is ever used. Do we need to store a full `InfoCacheEntry` on each file? Would storing just the type as a string be sufficient?

As it stands right now, this is _technically_ unsafe because we're reusing ASTs where the entries may be entirely different objects. But the fact that only this string matters does seem to imply it's possible.

internal/project/project.go
```diff
@@ -140,6 +140,27 @@ func (p *Project) GetSourceFile(fileName string, path tspath.Path, languageVersi
 	return nil
 }
 
+func (p *Project) GetImpliedNodeFormatForFileWorker(path string, packageJsonType string) core.ResolutionMode {
+	moduleResolution := p.GetProgram().Options().GetModuleResolutionKind()
```
jakebailey: This code is identical to the one in compiler host, and only depends on options plus the arguments; it seems like this could be done in the parser when setting up the module, right? Then, this method doesn't need to exist at all.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -355,22 +355,27 @@ type cachedCompilerHost struct {
...
 		core.SourceFileAffectingCompilerOptions
 		fileName        string
 		path            tspath.Path
 		languageVersion core.ScriptTarget
+		packageJsonType string
 		text            string
 	}
 
+	packageJsonTypeValue := ""
+	packageJsonTypeValue = packageJsonType
```
jakebailey: This is now redundant and can be inlined into the struct literal below.

andrewbranch:
Duplicate of #267?

jakebailey:
I thought maybe, but I suspect this one is more complete?

I wasn't sure given that other PR contains `getModeForUsageLocation` stuff which this PR does not need to do?

internal/parser/parser.go
```diff
@@ -176,7 +178,7 @@ func ParseJSONText(fileName string, path tspath.Path, sourceText string) *ast.So
 	return result
 }
 
-func (p *Parser) initializeState(fileName string, path tspath.Path, sourceText string, languageVersion core.ScriptTarget, scriptKind core.ScriptKind, jsdocParsingMode scanner.JSDocParsingMode) {
+func (p *Parser) initializeState(fileName string, path tspath.Path, sourceText string, languageVersion core.ScriptTarget, scriptKind core.ScriptKind, jsdocParsingMode scanner.JSDocParsingMode, moduleResolutionKind core.ModuleResolutionKind, packageJsonType string) {
```
makhov: Have you considered passing some struct with data instead? It's a long argument list and there are a lot of places, where an empty string is passed to `packageJsonType` var. At least a room for the future improvements.

jakebailey: The main risk is that we forget to set an option (since struct literal fields are optional), but we have a tendency to "just add one more param" when we need something that's not so good either.

internal/project/documentregistry.go
```diff
@@ -87,12 +87,16 @@ func (r *documentRegistry) getDocumentWorker(
...
+	if compilerOptions != nil {
+		moduleResolutionKind = compilerOptions.GetModuleResolutionKind()
+	}
 	if entryAny, ok := r.documents.Load(key); ok {
 		// We have an entry for this file. However, it may be for a different version of
 		// the script snapshot. If so, update it appropriately.
 		entry := entryAny.(*registryEntry)
 		if entry.sourceFile.Version != scriptInfo.version {
-			sourceFile := parser.ParseSourceFile(scriptInfo.fileName, scriptInfo.path, scriptInfo.text, scriptTarget, scanner.JSDocParsingModeParseAll)
+			sourceFile := parser.ParseSourceFile(scriptInfo.fileName, scriptInfo.path, scriptInfo.text, scriptTarget, scanner.JSDocParsingModeParseAll, moduleResolutionKind, "")
```
jakebailey: These both being set to `""` seems incorrect; shouldn't these files have their modes detected?

internal/parser/parser.go
frodi-karlsson: Yes I also felt it was a bit hard to read so I felt bad about it, but maybe it's best to leave like this for now so stuff like below comment is easily caught in a review

frodi-karlsson:
> Duplicate of https://github.com/microsoft/typescript-go/pull/267?

I hadn't seen that... I'll push forward with addressing comments until it's decided it's to be closed for duplication, considering this way of implementing it seems to have an advantage over 267 in where `impliedNodeFormat` is available

jakebailey:
We do want to move this info off the source file since it prevents AST reuse between Programs with different compiler options (why I mentioned moving stuff onto Program earlier), I just clearly forgot that the other PR existed when I was reviewing this one...

frodi-karlsson:
Hm what's the move:
- Do a semi-derivative metadata version where it lives on the program but with this reduced scope
- Close as duplicate?

internal/compiler/program.go
```diff
@@ -152,6 +153,8 @@ func NewProgram(options ProgramOptions) *Program {
 	p.files, p.resolvedModules = processAllProgramFiles(p.host, p.programOptions, p.compilerOptions, p.resolver, rootFiles, libs)
 	p.filesByPath = make(map[tspath.Path]*ast.SourceFile, len(p.files))
 	for _, file := range p.files {
+		packageType := p.resolver.GetPackageJsonTypeIfApplicable(string(file.Path()))
```
andrewbranch: In the old compiler, we avoid looking up the package.json if the file has a `.mts`/`.cts`/`.mjs`/`.cjs` extension. And I wonder if this can be done concurrently in the file loader? There’s already a mutex protecting some map writes there.

internal/ast/utilities.go
```diff
@@ -2354,27 +2354,40 @@ func IsDefaultImport(node *Node /*ImportDeclaration | ImportEqualsDeclaration |
...
+func GetEmitModuleFormatOfFileWorker(sourceFile *SourceFile, options *core.CompilerOptions, sourceFileMetaData *SourceFileMetaData) core.ModuleKind {
+	result := GetImpliedNodeFormatForEmitWorker(sourceFile, options, sourceFileMetaData)
 	if result != core.ModuleKindNone {
 		return result
 	}
 	return options.GetEmitModuleKind()
 }
 
-func GetImpliedNodeFormatForEmitWorker(sourceFile *SourceFile, options *core.CompilerOptions) core.ResolutionMode {
+func GetImpliedNodeFormatForEmitWorker(sourceFile *SourceFile, options *core.CompilerOptions, sourceFileMetaData *SourceFileMetaData) core.ModuleKind {
```
andrewbranch:
```suggestion
func GetImpliedNodeFormatForEmitWorker(sourceFile *SourceFile, options *core.CompilerOptions, sourceFileMetaData *SourceFileMetaData) core.ResolutionMode {
```

Also, maybe this function should just take a `fileName` instead of a `sourceFile` now?

internal/compiler/fileloader.go
```diff
@@ -221,9 +224,31 @@ func (t *parseTask) start(loader *fileLoader) {
...
+	if _, ok := p.sourceFileMetaDatas[path]; ok {
+		return
+	}
+
+	if p.sourceFileMetaDatas == nil {
+		p.sourceFileMetaDatas = make(map[string]*ast.SourceFileMetaData)
+	}
+
+	packageJsonType := p.resolver.GetPackageJsonTypeIfApplicable(string(path))
+	impliedNodeFormat := ast.GetImpliedNodeFormatForFile(path, packageJsonType)
```
andrewbranch: Same comment as before the move to `fileLoader`; this does unnecessary package.json lookups when the path has a module-specific file extension.

frodi-karlsson: Hmm I added the early return to address that in `GetPackageJsonTypeIfApplicable`. Did I misunderstand?

iamarpitpatidar (approved):

internal/compiler/module/resolver.go
```diff
@@ -145,6 +145,28 @@ func (r *Resolver) GetPackageScopeForPath(directory string) *packagejson.InfoCac
 	return (&resolutionState{compilerOptions: r.compilerOptions, resolver: r}).getPackageScopeForPath(directory)
 }
 
+func (r *Resolver) GetPackageJsonTypeIfApplicable(path string) string {
+	if tspath.FileExtensionIsOneOf(path, []string{".mts", ".cts", "mjs", "cjs"}) {
```
andrewbranch: Some of these have dots and some don’t, so I don’t think it’s going to work 😁 IIRC we have constants `tspath.ExtensionMts` and so on.

internal/compiler/fileloader.go
andrewbranch: Ah, I missed that. It’s a little unfortunate that the extension matching logic is duplicated between that resolver function and this `GetImpliedNodeFormatForFile` function, but I can live with it. It’s not likely to change.

internal/compiler/module/resolver.go
frodi-karlsson: Do we want to pre-calculate an array of module-specific extensions perhaps? Like `tspath.SupportedTSExtensionsWithJson`. I'll update to just the constants for now

frodi-karlsson:
Sorry about the force push, I messed up the merge (again)

internal/compiler/emitter.go
```diff
@@ -20,15 +20,16 @@ const (
...
+	host                EmitHost
+	emitOnly            emitOnly
+	emittedFilesList    []string
+	emitterDiagnostics  ast.DiagnosticsCollection
+	emitSkipped         bool
+	sourceMapDataList   []*sourceMapEmitResult
+	writer              printer.EmitTextWriter
+	paths               *outputPaths
+	sourceFile          *ast.SourceFile
+	sourcefileMetaDatas map[string]*ast.SourceFileMetaData
```
jakebailey: EmitHost already wraps program; I think that the metadatas could just come through that.

internal/printer/emitcontext.go
```diff
@@ -16,14 +16,15 @@ import (
...
-	emitHelpers   collections.OrderedSet[*EmitHelper]
+	Factory             *ast.NodeFactory // Required. The NodeFactory to use to create new nodes
+	autoGenerate        map[*ast.MemberName]*AutoGenerateInfo
+	textSource          map[*ast.StringLiteralNode]*ast.Node
+	original            map[*ast.Node]*ast.Node
+	emitNodes           core.LinkStore[*ast.Node, emitNode]
+	varScopeStack       core.Stack[*varScope]
+	letScopeStack       core.Stack[*varScope]
+	emitHelpers         collections.OrderedSet[*EmitHelper]
+	sourcefileMetaDatas map[string]*ast.SourceFileMetaData
```
jakebailey: @rbuckton What do you want for this? A map, or some function like others?

frodi-karlsson:
Agh I'm confused why the tests are failing in CI and not locally (`go test -race ./... '--timeout=45m'`)

jakebailey:
Run `hereby test --concurrentTestPrograms` or `TS_TEST_PROGRAM_SINGLE_THREADED=false go test ./...`.

But, maybe you're just out of date with main? Those tests are new emit tests we just merged.

internal/printer/emitcontext.go
```diff
@@ -16,14 +16,15 @@ import (
...
-	emitHelpers   collections.OrderedSet[*EmitHelper]
+	Factory            *ast.NodeFactory // Required. The NodeFactory to use to create new nodes
+	autoGenerate       map[*ast.MemberName]*AutoGenerateInfo
+	textSource         map[*ast.StringLiteralNode]*ast.Node
+	original           map[*ast.Node]*ast.Node
+	emitNodes          core.LinkStore[*ast.Node, emitNode]
+	varScopeStack      core.Stack[*varScope]
+	letScopeStack      core.Stack[*varScope]
+	emitHelpers        collections.OrderedSet[*EmitHelper]
+	sourcefileMetaData *ast.SourceFileMetaData
```
rbuckton: `EmitContext` is not necessarily guaranteed to be per-file, even if that's how it's used at the moment. This should be a `map[*ast.SourceFile]*ast.SourceFileMetaData`.

internal/printer/emitcontext.go
```diff
@@ -625,6 +626,10 @@ func (c *EmitContext) NewRewriteRelativeImportExtensionsHelper(firstArgument *as
 	)
 }
 
+func (c *EmitContext) SetSourcefileMetaData(sourcefileMetaData *ast.SourceFileMetaData) {
```
rbuckton: We should pass in the file as an argument.

internal/printer/emitcontext.go
```diff
@@ -949,3 +954,7 @@ func (c *EmitContext) VisitIterationBody(body *ast.Statement, visitor *ast.NodeV
 
 	return updated
 }
+
+func (c *EmitContext) GetSourceFileMetaData() *ast.SourceFileMetaData {
```
rbuckton: We should pass in the file as an argument, and if it's not found, walk up `Original` pointers until we find an entry, e.g.:

```go
for {
  if meta, ok := c.sourceFileMetaDatas[file]; ok {
    return meta;
  }
  original := c.Original(file.AsNode())
  if original == nil || !ast.IsSourceFile(original) {
    return nil
  }
  file = original.AsSourceFile();
}
```

internal/checker/grammarchecks.go
```diff
@@ -1217,7 +1217,8 @@ func (c *Checker) checkGrammarForInOrForOfStatement(forInOrOfStatement *ast.ForI
 					}
 					switch c.moduleKind {
 					case core.ModuleKindNode16, core.ModuleKindNodeNext:
-						if sourceFile.ImpliedNodeFormat == core.ModuleKindCommonJS {
+						sourceFileMetaData := c.program.GetCachedSourceFileMetaData(string(sourceFile.Path()))
```
rbuckton: Some calls use `string(sourceFile.Path())` while others use `sourceFile.FileName()`. These are not guaranteed to be the same. We should pick one (ideally `Path()`) and only use that.

rbuckton (changes requested):

internal/checker/grammarchecks.go
andrewbranch: All filename cache keys should be of type `tspath.Path`, so it makes sense to use that for the lookup.

rbuckton: In that case, we should have the various `GetCachedSourceFileMetaData` signatures take `tspath.Path`.

internal/printer/emitcontext.go
frodi-karlsson: I think I got the rest of the comments down, but I didn't quite understand this one in the context of switching back to a map instead of a single `SourceFileMetaData`. Do you think it makes sense to store a reference to the map from the program instead in emitContext?

internal/printer/emitcontext.go
```diff
@@ -955,6 +956,15 @@ func (c *EmitContext) VisitIterationBody(body *ast.Statement, visitor *ast.NodeV
...
+	for {
+		if meta, ok := c.sourceFileMetaDatas[file.Path()]; ok {
+			return meta
+		}
+		original := c.Original(file.AsNode())
+		if original == nil || !ast.IsSourceFile(original) {
+			return nil
+		}
+		file = original.AsSourceFile()
+	}
```
rbuckton: Since this is keyed by `tspath.Path` and not by node, walking original pointers isn't necessary since `factory.UpdateSourceFile` always copies `path`.

frodi-karlsson: Ah of course 🤦 Thank you, sorry the emitcontext logic hasn't lodged itself in my head yet, I should've taken more time to understand what I was doing

rbuckton: Since this information is sourced from `Program`, I wonder it might make more sense to just pass an interface to the various module transformers to expose this, much like how we pass in a `ReferenceResolver`, rather than copying it to `EmitContext`.

jakebailey: Yeah, that's what I was thinking about here, given EmitHost already depends on program directly, but I didn't immediately see (looking only at this PR) if that was going to work.

rbuckton: Thinking on this a bit, my suggestion would be to add something like this to `printer`:

```go
type SourceFileMetadataResolver interface {
  GetCachedSourceFileMetadata(path tspath.Path) *ast.SourceFileMetadata
}
```

and then pass something implementing that into the various transformers that need them

internal/transformers/metadataprovider.go
frodi-karlsson, line 1: Name may not be the best, but is this the direction you had in mind?

edit: saw comment above after posting this about name and location

rbuckton: Roughly. I would probably call it `SourceFileMetadataProvider`, and rather than having a `NewMetadataProvider`-like function or an implementation, we could have the method/signature match the prior version you had for `Program` that took a path as an argument and just pass the `Program` itself.

For tests that don't need a full program, we could just implement a mock version of `SourceFileMetadataProvider`.

rbuckton (approved):
This looks good to me, assuming @andrewbranch and @jakebailey don't have any further feedback.

internal/compiler/fileloader.go
```diff
@@ -221,9 +224,31 @@ func (t *parseTask) start(loader *fileLoader) {
 	})
 }
 
+func (p *fileLoader) CacheSourceFileMetaData(path tspath.Path) {
+	p.sourceFileMetaDatasMutex.Lock()
+	defer p.sourceFileMetaDatasMutex.Unlock()
```
jakebailey: This lock is held for the entire call, so I don't think we're gaining any parallelism here, no?

The "fix" would be to only hold the lock for the check at the top, unlock it, then relock it at the bottom (moving the make call down too).

internal/compiler/fileloader.go
```diff
@@ -221,9 +224,31 @@ func (t *parseTask) start(loader *fileLoader) {
...
+		ImpliedNodeFormat: impliedNodeFormat,
+	}
+
+	p.sourceFileMetaDatas[path] = metadata
+}
+
 func (p *fileLoader) parseSourceFile(fileName string) *ast.SourceFile {
 	path := tspath.ToPath(fileName, p.host.GetCurrentDirectory(), p.host.FS().UseCaseSensitiveFileNames())
 	sourceFile := p.host.GetSourceFile(fileName, path, p.compilerOptions.GetEmitScriptTarget())
+	p.CacheSourceFileMetaData(path)
```
jakebailey: Nit, but `CacheSourceFileMetaData` shouldn't be exported, and I would probably rename it to be like `processSourceFileMetaData` or `loadSourceFileMetaData` or something (everything in this file is "caching" information for later, in a way, it's the loader).

internal/compiler/host.go
```diff
@@ -73,5 +73,6 @@ func (h *compilerHost) GetSourceFile(fileName string, path tspath.Path, language
 	if tspath.FileExtensionIs(fileName, tspath.ExtensionJson) {
 		return parser.ParseJSONText(fileName, path, text)
 	}
+
```
jakebailey: Nit but, random leftover empty line diff.

internal/parser/parser.go
```diff
@@ -198,6 +198,7 @@ func (p *Parser) initializeState(fileName string, path tspath.Path, sourceText s
 	default:
 		p.contextFlags = ast.NodeFlagsNone
 	}
+
```
jakebailey: Here too.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -399,6 +399,7 @@ func (h *cachedCompilerHost) GetSourceFile(fileName string, path tspath.Path, la
 		fileName        string
 		path            tspath.Path
 		languageVersion core.ScriptTarget
+		packageJsonType string
```
jakebailey: I think this is now left over.

jakebailey (commented):
Only some small notes. Very happy to see this nearly complete.

internal/compiler/fileloader.go
jakebailey: Which, also indicates that this should be a RWMutex.

internal/compiler/fileloader.go
```diff
@@ -224,12 +224,12 @@ func (t *parseTask) start(loader *fileLoader) {
...
-func (p *fileLoader) CacheSourceFileMetaData(path tspath.Path) {
+func (p *fileLoader) loadSourceFileMetaData(path tspath.Path) {
 	p.sourceFileMetaDatasMutex.Lock()
-	defer p.sourceFileMetaDatasMutex.Unlock()
 	if _, ok := p.sourceFileMetaDatas[path]; ok {
 		return
 	}
+	p.sourceFileMetaDatasMutex.Unlock()
 
 	if p.sourceFileMetaDatas == nil {
```
jakebailey: This needs to be down at the bottom under lock.

internal/compiler/fileloader.go
```diff
@@ -225,15 +225,11 @@ func (t *parseTask) start(loader *fileLoader) {
 }
 
 func (p *fileLoader) loadSourceFileMetaData(path tspath.Path) {
-	p.sourceFileMetaDatasMutex.Lock()
+	p.sourceFileMetaDatasMutex.RLock()
 	if _, ok := p.sourceFileMetaDatas[path]; ok {
 		return
```
jakebailey, line 232:
```suggestion
		p.sourceFileMetaDatasMutex.RUnlock()
		return
```

This is a deadlock otherwise.

frodi-karlsson: Gosh I'm really messing up this mutex, I've gotten unproductively excited with the landing close

jakebailey (approved):
:shipit:

andrewbranch (approved):
Thanks!

frodi-karlsson:
Thank you! I know I was a bit in over my head so it was a bit rocky, but I'll do my best to repay all these domain learnings to the repo with time

Issue #631, opened 2025-03-16
bug/lsp: fails to resolve relative type files in a library


laciferin2024:
## Lsp log

[lsp.log](https://github.com/user-attachments/files/19268991/lsp.log)

john-landgrave:
I can see that it isn't necessarily the same cause, but I ran into this yesterday when I had an "Untitled - 1" kind of file (a "scratchpad" sort of editor that wasn't ever saved) and it caused the LSP to crash enough that it stopped trying to boot itself up. I had set the language of the file to Typescript (for syntax highlighting) so I suppose the LSP was trying to find it and failing to?

jakebailey:
> I can see that it isn't necessarily the same cause, but I ran into this yesterday when I had an "Untitled - 1" kind of file (a "scratchpad" sort of editor that wasn't ever saved) and it caused the LSP to crash enough that it stopped trying to boot itself up. I had set the language of the file to Typescript (for syntax highlighting) so I suppose the LSP was trying to find it and failing to?

I think this is unrelated and may now be fixed?

jakebailey:
Please recheck; module resolution should be much improved.

laciferin2024:
Ok

laciferin2024:
Yes its working now

PR #638, opened 2025-03-16
Return errors instead of panics

makhov:
tsgo heavily overuses panics, that can lead to the situations mentioned in #631:
```
The typescript-go-lsp server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

PR is intended to start moving the codebase towards more idiomatic error handling.
Atm, I've changed 2 panics and we probably need to agree on the approach: I can keep adding changes to this PR, but it'll be big and hard to follow, or I can create a separate PRs to keep them small and simple.

### DIFF
--- a/internal/ls/languageservice.go
+++ b/internal/ls/languageservice.go
@@ -1,6 +1,8 @@
 package ls
 
 import (
+	"fmt"
+
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/compiler"
 	"github.com/microsoft/typescript-go/internal/core"
@@ -55,11 +57,11 @@ func (l *LanguageService) GetProgram() *compiler.Program {
 	return l.host.GetProgram()
 }
 
-func (l *LanguageService) getProgramAndFile(fileName string) (*compiler.Program, *ast.SourceFile) {
+func (l *LanguageService) getProgramAndFile(fileName string) (*compiler.Program, *ast.SourceFile, error) {
 	program := l.GetProgram()
 	file := program.GetSourceFile(fileName)
 	if file == nil {
-		panic("file not found")
+		return nil, nil, fmt.Errorf("file not found")
 	}
-	return program, file
+	return program, file, nil
 }
--- a/internal/ls/definition.go
+++ b/internal/ls/definition.go
@@ -1,17 +1,22 @@
 package ls
 
 import (
+	"fmt"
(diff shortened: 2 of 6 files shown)


makhov:

internal/ls/languageservice.go
```diff
@@ -55,11 +57,11 @@ func (l *LanguageService) GetProgram() *compiler.Program {
 	return l.host.GetProgram()
 }
 
-func (l *LanguageService) getProgramAndFile(fileName string) (*compiler.Program, *ast.SourceFile) {
+func (l *LanguageService) getProgramAndFile(fileName string) (*compiler.Program, *ast.SourceFile, error) {
 	program := l.GetProgram()
 	file := program.GetSourceFile(fileName)
 	if file == nil {
-		panic("file not found")
```
jakebailey, line 62: My understanding is that these are mainly invariants that shouldn't be broken, such that we'd want to know about them. But, maybe we need to surface these differently (like gopls' `bug.Report`), or as a general panic handler (sort of like old TS with exceptions, which we also did not use except in tragic cases).

makhov: Yeah, but it happened in the mentioned issue, and I'm not sure it should crash the LSP server. 

I don't have any strong opinion, because it's always up to the team and the project. But in general, panics are for unexpected errors, this one (even if it should never happen in real life) is kinda expected.  
And I'm not a big fan of global recovery exactly because if some panic happens, we'd want to know about it.

I think the same problem would be reported in issue even without panic anyway, since it's a bug (at least looks like the one).

laciferin2024: I agree, lsp servers shouldn't crash for `file not found` cases. At least the ts lsp , go lsp doesn't crash. I think we need an `error channel` to manage errors async. The Current version #645  seems to be halting at every error.

jakebailey (changes requested):
I think I generally disagree with this, at least for now; operations on the Program should be infalliable, more or less. It's a bug in the Project/ProjectService if it ever attempts to look up a file that doesn't exist.

For the other LSP calls, I can imagine us wanting to do error returns (especially for cancellation), but I think that will probably need to be done at a later point.

sandersn:
I think this PR should be closed. In general, TS panics when something is badly broken, and things are commonly broken in Corsa's preview state. 
In the language server case, I predict we'll continue to use panics, and then handle panics in a central place.

PR #639, opened 2025-03-16
Annotate all CompilerOptions fields with omitzero

jakebailey:
This makes baselines a lot cleaner.


andrewbranch (approved):
This is fine, but it’s also a reminder that we need to overhaul this for `null` to be able to work in `extends`...

jakebailey:
Yes, though that all is thankfully before this display layer, and this is all a very quick hack for printing out configs anyhow...

andrewbranch:
FWIW, it’s not just for display in my API branch; it gets sent to JS through that marshalling too.

jakebailey:
Oh, eek. Okay. Though the bits should actually come out to be the same, at least if one goes from Go to Go, but obviously omitting them on the Go side means they will be undefined on the JS side...

andrewbranch:
Which is more than likely how they came in on the JSON side, which is why this whole thing is sketchy 🙃

PR #641, opened 2025-03-17
Fix multiple issues uncovered by test suites

ahejlsberg:


internal/scanner/scanner.go
```diff
@@ -1962,6 +1962,16 @@ func TokenToString(token ast.Kind) string {
 	return tokenToText[token]
 }
 
+func GetViableKeywordSuggestions() []string {
```
jakebailey, line 1965: I think this the third open PR with this fix 😄

ahejlsberg:
Most of the fixes here are trivial, but there was an interesting issue with the caching of `Relater` objects that in border cases allowed old non-zeroed state to survive. They're now kept on a simple free list.

jakebailey:
> Most of the fixes here are trivial, but there was an interesting issue with the caching of `Relater` objects that in border cases allowed old non-zeroed state to survive. They're now kept on a simple free list.

Huh, which field did that? I had sliced off or cleared everything, I thought...

internal/checker/checker.go
```diff
@@ -11521,7 +11533,10 @@ func (c *Checker) checkBinaryLikeExpression(left *ast.Node, operatorToken *ast.N
 				ast.KindGreaterThanGreaterThanGreaterThanEqualsToken:
 				rhsEval := c.evaluate(right, right)
 				if numValue, ok := rhsEval.Value.(jsnum.Number); ok && numValue.Abs() >= 32 {
-					c.errorOrSuggestion(ast.IsEnumMember(ast.WalkUpParenthesizedExpressions(right.Parent.Parent)), errorNode, diagnostics.This_operation_can_be_simplified_This_shift_is_identical_to_0_1_2, scanner.GetTextOfNode(left), scanner.TokenToString(operator), (numValue / 32).Floor())
+					// Adding 0 removes sign from -0
+					shiftMod32 := math.Mod(float64(numValue), 32) + 0
```
jakebailey, line 11537: Probably just `numValue.Remainder(32)`? I'd rather not rely on the `math` package outside of `jsnum`; I eliminated it everywhere except for the max int consts.

internal/checker/relater.go
```diff
@@ -365,10 +365,12 @@ func (c *Checker) checkTypeRelatedToEx(
 	headMessage *diagnostics.Message,
 	diagnosticOutput *[]*ast.Diagnostic,
 ) bool {
-	relaterCount := len(c.relaters)
-	c.relaters = slices.Grow(c.relaters, 1)[:relaterCount+1]
-	r := &c.relaters[relaterCount]
-	r.c = c
+	r := c.freeRelater
+	if r == nil {
+		r = &Relater{c: c}
```
jakebailey, line 370: Honestly I feel like we should probably clean this up into a plain sync pool, but, not in this PR.

jakebailey (approved):

internal/checker/checker.go
ahejlsberg: Yes, I'll switch to `Number.Remainder`. Which by the way has a bug in it that I'll also fix.

ahejlsberg:
> Huh, which field did that? I had sliced off or cleared everything, I thought...

Here's how it went wrong with the old logic: In the first call to `checkTypeRelatedToEx`, `slices.Grow` allocates a one-element array. In a following recursive call to `checkTypeRelatedToEx`, `slices.Grow` re-allocates a two-element array and _copies the current state_ of the first `Relater` into the first element of the new two-element array. Later, when the first call returns, it clears the `Relater` in the _old_ one-element array, but leaves the first element of the two-element array with the copied state, and this state is never cleared.

Yeah, we should switch to a sync pool, but seems like overkill when sync isn't needed.

PR #642, opened 2025-03-17
remove unused "panicValue"

wjdqhry:
While exploring the project to understand its structure, I noticed that "panicValue" was defined but not used.

While this is a small change, I hope to make more meaningful contributions in the future.

Thank you for your work on this project!


jakebailey (approved):
Oops, I left this behind. Thanks.

PR #644, opened 2025-03-17
Fix reduceLabels slice passing

jakebailey:
This is code that we added in #235 to stop mutating flow nodes.

#322 points out that these assignments are ineffectual. Fixing this by passing slices downward fixes one baseline.

(Extracting this out of #322 for a separate review.)


internal/checker/flow.go
```diff
@@ -2442,15 +2442,17 @@ func (c *Checker) getFlowTypeInStaticBlocks(symbol *ast.Symbol, staticBlocks []*
...
 	return result
 }
 
-func (c *Checker) isReachableFlowNodeWorker(flow *ast.FlowNode, noCacheCheck bool) bool {
+func (c *Checker) isReachableFlowNodeWorker(flow *ast.FlowNode, noCacheCheck bool, reduceLabels []*ast.FlowReduceLabelData) bool {
 	var reduceLabelsBuffer [4]*ast.FlowReduceLabelData
-	reduceLabels := reduceLabelsBuffer[:0]
+	if reduceLabels == nil {
+		reduceLabels = reduceLabelsBuffer[:0]
+	}
```
jakebailey, line 2455: Now that this is passed down, I'm not entirely certain that this stack allocation optimization works.

PR #646, opened 2025-03-17
Fix shorthand property assignment type resolution

auvred:
If you try to call `getTypeOfSymbol` on `foo` (shorthand property assignment), `error` type will be returned.

```typescript
const foo = 1

const res = { foo }
//            ^^^
```

---

On line 11973 in checker.ts, `declaration.name` (not just `declaration`) is passed to `checkExpressionForMutableLocation`

```typescript
    function getTypeOfVariableOrParameterOrPropertyWorker(symbol: Symbol): Type {
        // ...
        else if (isShorthandPropertyAssignment(declaration)) {
            type = tryGetTypeFromEffectiveTypeNode(declaration) || checkExpressionForMutableLocation(declaration.name, CheckMode.Normal);
        }
        // ...
    }
```


auvred:

jakebailey (approved):
Great catch, thank you!

PR #651, opened 2025-03-17
Add baseline/diff generation for JS emit

rbuckton:
This adds baseline and diff support for JS emit, fixes a few crashes, and accepts the current set of baselines.


internal/testrunner/compiler_runner.go
```diff
@@ -89,7 +89,8 @@ func (r *CompilerBaselineRunner) RunTests(t *testing.T) {
 	r.cleanUpLocal(t)
 	files := r.EnumerateTestFiles()
 	skippedTests := []string{
-		"mappedTypeRecursiveInference.ts", // Needed until we have type printer with truncation limit.
+		"mappedTypeRecursiveInference.ts",         // Needed until we have type printer with truncation limit.
+		"jsFileCompilationWithoutJsExtensions.ts", // Needed until we have proper allowJS support (and errors when not enabled.)
```
jakebailey: Does this crash, or just fail?

internal/testrunner/compiler_runner.go
```diff
@@ -341,6 +343,33 @@ func (c *compilerTest) verifyDiagnostics(t *testing.T, suiteName string, isSubmo
 	})
 }
 
+func (c *compilerTest) verifyJavaScriptOutput(t *testing.T, suiteName string, isSubmodule bool) {
```
jakebailey, line 349: This same thing will do dts files too, right? Should we just say it's "output"?

internal/testrunner/compiler_runner.go
rbuckton: It crashes due to a panic in DoJsEmitBaseline that we would want to keep. The issue is that we're not reporting a diagnostic related to `allowJS` that hasn't been implemented in program yet.

internal/testrunner/compiler_runner.go
rbuckton: This is what it was called in Strada. I don't see any sense in changing it while we're still porting/migrating as it makes it easier to find.

internal/testutil/tsbaseline/js_emit_baseline.go
```diff
@@ -0,0 +1,273 @@
...
+	otherFiles []*harnessutil.TestFile,
+	harnessSettings *harnessutil.HarnessOptions,
+	opts baseline.Options,
+) {
+	if !options.NoEmit.IsTrue() && !options.EmitDeclarationOnly.IsTrue() && result.Js.Size() == 0 && len(result.Diagnostics) == 0 {
+		panic("Expected at least one js file to be emitted or at least one error to be created.")
+	}
+
+	// check js output
+	tsCode := ""
```
gabritto: This should probably be a `strings.Builder`.

internal/testutil/tsbaseline/js_emit_baseline.go
```diff
@@ -104,7 +104,11 @@ func DoJsEmitBaseline(
 					expected := original.Content
 					actual := doc.Content
 					var patch strings.Builder
-					diff.Text("Expected\tThe full check baseline", "Actual\twith noCheck set", expected, actual, &patch)
+					err := diff.Text("Expected\tThe full check baseline", "Actual\twith noCheck set", expected, actual, &patch)
```
jakebailey: t.Fatal here, or ignore the error and skip the test (if the file is too big for example)

jakebailey:
Hm, seems like this is causing us to check every file twice:

image

I take it this is because `noCheck` doesn't work, so the "repeat" emit test for that doesn't proceed quickly? Perhaps we should just skip this test for now?

rbuckton:
> I take it this is because noCheck doesn't work

I think `noCheck` works. This is likely due to a typo in the `Repeat` implementation, I'll have a fix up shortly.

jakebailey:
It's improved, though we still do seem to do full checks? This is the thing Anders noted about `--noCheck`, that it doesn't skip as much as Strada.

internal/testrunner/compiler_runner.go
```diff
@@ -88,8 +88,13 @@ func (r *CompilerBaselineRunner) EnumerateTestFiles() []string {
...
 	r.cleanUpLocal(t)
 	files := r.EnumerateTestFiles()
-	skippedTests := []string{
-		"mappedTypeRecursiveInference.ts", // Needed until we have type printer with truncation limit.
+	skippedTests := map[string]string{
+		"mappedTypeRecursiveInference.ts":         "Skipped until we have type printer with truncation limit.",
+		"jsFileCompilationWithoutJsExtensions.ts": "Skipped until we have proper allowJS support (and errors when not enabled.)",
+		"fileReferencesWithNoExtensions.ts":       "Skipped until we support adding missing extensions in subtasks in fileloader.go",
+		"typeOnlyMerge2.ts":                       "Needs investigation",
+		"typeOnlyMerge3.ts":                       "Needs investigation",
```
jakebailey, line 96: These two are also skipped in `internal/testrunner/compiler_runner.go`, so could probably be removed there.

jakebailey:
I don't really know what to do to review the files GitHub won't load, but it passes! 😅

jakebailey:
I did some local futzing to undo all of the baselines and get a diff of just the code. I think things look good, my only commentary would be about `OutputRecorderFS`; is the writable test FS not sufficient for the testing? I had to do some goofy stuff to make it compatible with symlinks so I wasn't sure if the `Realpath` stuff would work like that on non-existent files either (but, hopefully should). I don't think it's a big deal for now, though; I'll approve this if you want to start working with it.

jakebailey (approved):

rbuckton:
> is the writable test FS not sufficient for the testing? 

Is it possible to differentiate between what was present in the FS before compilation vs. after? In Strada we use a fake `CompilerHost` which keeps track of outputs.

jakebailey:
Not directly or conveniently, but @iisaduan's tsc tests track this by walking the FS before/after to show additions, removals, changes, etc, I believe.

rbuckton:
The vfs in Strada (harness/vfsUtil.ts) had that capability via snapshotting, but IMO the most reliable way to do this is to ensure a given component only interacts with a mockable host, e.g., instead of exposing `FS` on `CompilerHost`, relevant FS operations (`UseCaseSensitiveFileNames`, `WriteFile`, etc.) should be accessible via `CompilerHost` itself so that that dependency can be mocked as needed, not unlike what we're doing with `cachedCompilerHost` in harnessutil.

PR #653, opened 2025-03-17
Eliminate hardcoded RuneSelf-1

jakebailey:
Notice this while working on UTF-16 handling; rather than hardcoding this constant, just use the one provided by the Go stdlib.


internal/core/core.go
```diff
@@ -323,7 +323,7 @@ func ComputeLineStartsSeq(text string) iter.Seq[TextPos] {
 		lineStart := 0
 		for pos < len(text) {
 			b := text[pos]
-			if b < 0x7F {
+			if b < utf8.RuneSelf {
```
jakebailey, line 326: This is a behavior change, but a correct one if you read the original code (or think about the logic).

jakebailey:
Going in #656

PR #654, opened 2025-03-17
Delete error/types diffs for tests with known unsupported extensions

jakebailey:
Right now, these are all just noise; it's easy to readd these later or when working on these features.


jakebailey:
Need to do types baselines too, oops. Let me do that quick.

jakebailey:
Alternatively, I can roll this into my "expected diffs" PR such that these diffs just get moved somewhere else.

jakebailey:
Yeah, I'm going to do that.

PR #656, opened 2025-03-18
Support UTF-16 in LSP

jakebailey:
Before:

image

After:

image

Includes #653

Per https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments

The eldritch horror is:

- JS defines 4 different newline sequences
- LSP defines 3 different newlines sequences
- Editors like vim, git, only have one newline sequence
- VS Code follows LSP (or vice versa), but you can only select LF or CRLF


DanielRosenwasser:
> - JS defines 4 different newline sequences
> - LSP defines 3 different newlines sequences
>
> - VS Code follows LSP (or vice versa), but you can only select LF or CRLF

Obligatory link to https://github.com/microsoft/TypeScript/issues/38078

jakebailey:
Yeah, so I am clearly doing the LSP thing here, and I can't find a reason why we shouldn't just do that regardless for everything else, honestly.

internal/lsp/converters.go
```diff
@@ -188,23 +192,60 @@ func fileNameToDocumentUri(fileName string) lsproto.DocumentUri {
...
+	// The current line ranges from lineMap.LineStarts[line] (or 0) to lineMap.LineStarts[line+1] (or len(text)).
+
+	start := lineMap.LineStarts[line]
+
+	var character core.TextPos
+	if lineMap.AsciiOnly {
+		character = position - start
+	} else {
+		// We need to rescan the text as UTF-16 to find the character offset.
+		for _, r := range scriptInfo.Text()[start:position] {
```
DanielRosenwasser, line 244: Didn't realize that `range` does "the right thing" over a `string`.

internal/lsp/converters.go
```diff
@@ -188,23 +192,60 @@ func fileNameToDocumentUri(fileName string) lsproto.DocumentUri {
...
+
+	return start + utf8Char
 }
 
-func positionToLineAndCharacter(position int, lineMap []core.TextPos) lsproto.Position {
-	line, character := core.PositionToLineAndCharacter(position, lineMap)
+func positionToLineAndCharacter(scriptInfo *project.ScriptInfo, position core.TextPos) lsproto.Position {
+	// UTF-8 offset to UTF-8/16 0-indexed line and character
+
+	lineMap := scriptInfo.LineMapLSP()
```
DanielRosenwasser, line 230: Does this ever get cached anywhere? Is the work being done on every call?

internal/lsp/server.go
```diff
@@ -211,12 +213,21 @@ func (s *Server) handleMessage(req *lsproto.RequestMessage) error {
 
 func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
 	s.initializeParams = req.Params.(*lsproto.InitializeParams)
+
+	s.positionEncoding = lsproto.PositionEncodingKindUTF16
+	if s.initializeParams.Capabilities.General != nil && s.initializeParams.Capabilities.General.PositionEncodings != nil {
+		if slices.Contains(*s.initializeParams.Capabilities.General.PositionEncodings, lsproto.PositionEncodingKindUTF8) {
```
DanielRosenwasser:
```suggestion
	genCapabilities := s.initializeParams.Capabilities.General
	if genCapabilities != nil && genCapabilities.PositionEncodings != nil {
		if slices.Contains(*genCapabilities.PositionEncodings, lsproto.PositionEncodingKindUTF8) {
```

internal/lsp/converters.go
jakebailey: Yes, it's actually faster than manually calling the utf8 lib, IIRC.

internal/lsp/converters.go
jakebailey: Yes, scriptInfo caches this.

DanielRosenwasser: Can you update the source file's line map to be the same if the file contains only ASCII? That way the two will be deduplicated.

DanielRosenwasser: Or possibly just track whether a non CR/LF line ending was encountered here and do it then.

jakebailey: By that point we'll have constructed the whole thing, so I'm not totally sure if it's helpful, but I guess we could save a little memory sometimes... If we've even requested the line map for other reasons, which we might not have at all.

andrewbranch:
> Yeah, so I am clearly doing the LSP thing here, and I can't find a reason why we shouldn't just do that regardless for everything else, honestly.

I feel like we can just standardize on the LSP-compatible line map as long as we don’t use it for parsing/grammar considerations around `[no LineTerminator here]`, which I’m pretty sure we don’t...

IOW, we have to respect ECMAScript’s conception of a line terminator, but we don’t have to use it in our own reporting of line numbers, which exists to help humans find their code in an editor.

jakebailey:
The line map is also used for source maps. I do genuinely wonder if source maps actually use JS's definition, though.

DanielRosenwasser:
https://tc39.es/ecma426/2024/#extraction-javascript does split across ECMAScript code points, so it is a bit implied.

jakebailey:
> Let lines be the result of [strictly splitting](https://infra.spec.whatwg.org/#strictly-split) source on [ECMAScript line terminator code points](https://tc39.es/ecma262/#table-line-terminator-code-points).

Yay.

andrewbranch (approved):

PR #658, opened 2025-03-18
feat: moving nil check to call sites of IsFunctionLikeDeclaration

SoulPancake:


SoulPancake:
@rbuckton Can you take a look at this?

SaadiSave:
@SoulPancake Sorry for jumping on here for a slightly tangential issue, but I noticed that nil checks are inconsistent throughout the codebase. For example,

https://github.com/microsoft/typescript-go/blob/main/internal%2Fast%2Fast.go#L15-L31

```go
type Visitor func(*Node) bool

func visit(v Visitor, node *Node) bool {
        if node != nil {
                return v(node)
        }
        return false
}

func visitNodes(v Visitor, nodes []*Node) bool {
        for _, node := range nodes {
                if v(node) {
                        return true
                }
        }
        return false
}
```

`visit` checks for nils, but the for loop doesn't check if `node` is nil. Do you have clear documentation on whether `Visitor`s are supposed to handle nils? Because `visit` assumes they don't, and `visitNodes` assumes they do, which are mutually incompatible assumptions.

SoulPancake:
@SaadiSave Agreed, I will add the nil check in the loop

SaadiSave:
@SoulPancake is there a lint to enforce nil checks?

SoulPancake:
Don't think so @SaadiSave

jakebailey:
This PR has gotten out of date as main changed. Could you update it, or close it if you don't plan on working on this anymore?

SoulPancake:
I will update it asap @jakebailey

SoulPancake:
@jakebailey Done, Can you take a look at this?

_submodules/TypeScript
jakebailey, line 1: You seem to have accidentally moved this during the merge/rebase. Do a `git checkout main -- _submodules/TypeScript` to fix it.

jakebailey: Hm, still not up to date, your main must not be up to date with upstream.

SoulPancake: I have updated the submodule with upstream
Can you take a look now? @jakebailey

jakebailey:
I feel like most of these are redundant; I would think we'd only want ones that correspond to undefined checks in the original code.

SoulPancake:
@jakebailey I guess these aren't needed or helpful as you mentioned, do you suggest I remove the TODO comments and remove the changes?

SoulPancake:
Otherwise I am happy to close this PR @jakebailey

PR #661, opened 2025-03-18
Fix BigInt handling

### LINKED ISSUE #494: bigints in template literal strings not working

jakebailey:
Fixes #494


internal/jsnum/pseudobigint.go
```diff
@@ -25,3 +31,37 @@ func (value PseudoBigInt) Sign() int {
...
+			return "0"
+		}
+		return stringValue
+	}
+
+	bi, ok := new(big.Int).SetString(stringValue, 0)
+	if !ok {
+		panic(fmt.Sprintf("Failed to parse big int: %q", stringValue))
+	}
+	return bi.String() // !!!
```
jakebailey, line 66: My super lazy way of not porting all of `ParsePseudoBigInt` right now. The Go bigint parser works fine for these.

internal/jsnum/pseudobigint_test.go
```diff
@@ -0,0 +1,39 @@
...
+				assert.Equal(
+					t,
+					ParsePseudoBigInt(strings.Repeat("0", leadingZeros)+testNumber.String()+"n"),
+					testNumber.String(),
+				)
+			}
+		}
+	})
+
+	// TODO(jakebailey): tests for other bases
```
jakebailey, line 34: The other test cases here depend on being run in JS where we can convert number to string with multiple bases, something jsnum can't do right now. I might try my hand at that quick, though.

testdata/baselines/reference/submodule/conformance/templateLiteralTypes4.types.diff
```diff
@@ -47,28 +47,23 @@
  
  // infer from bigint
  type TBigInt0 = "100" extends `${infer N extends bigint}` ? N : never; // 100n
-->TBigInt0 : 100n
-+>TBigInt0 : never
+ >TBigInt0 : 100n
 +>N : N
```
jakebailey: Noticed these and then figured out #663, yay!

testdata/baselines/reference/submodule/conformance/templateLiteralTypesPatterns.errors.txt.diff
```diff
@@ -81,27 +77,15 @@
...
-+!!! error TS2345: Argument of type '"0x1"' is not assignable to parameter of type '`${bigint}`'.
-     bigints("0o1");
-+            ~~~~~
-+!!! error TS2345: Argument of type '"0o1"' is not assignable to parameter of type '`${bigint}`'.
+@@= skipped -133, +133 lines =@@
+     
+     // whitespace and comments aren't part of numbers
+     bigints("- 1");
+-            ~~~~~
+-!!! error TS2345: Argument of type '"- 1"' is not assignable to parameter of type '`${bigint}`'.
```
jakebailey, line 85: This is missing; will investigate to figure out where this goes wrong.

jakebailey: It turns out our scanner doesn't respect skipTrivia in all of the places it's supposed to.

jakebailey:
I'm going to undraft this; the scanner fixes are very, very invasive. We've apparently never set skipTrivia=false so all of that code was untested and incorrect.

internal/jsnum/pseudobigint_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	t.Run("can parse large literals", func(t *testing.T) {
+		t.Parallel()
+		assert.Equal(t, ParsePseudoBigInt("123456789012345678901234567890n"), "123456789012345678901234567890")
+		assert.Equal(t, ParsePseudoBigInt("0b1100011101110100100001111111101101100001101110011111000001110111001001110001111110000101011010010n"), "123456789012345678901234567890")
```
sandersn, line 39: what happens if one of the binary-format numbers has a non `01` (and the same for octal and decimal, too)?

Edit: Oh, the Go bigint parser receives the prefix as well, so should fail.

internal/jsnum/pseudobigint.go
```diff
@@ -25,3 +31,37 @@ func (value PseudoBigInt) Sign() int {
...
+	return PseudoBigInt{
+		Negative:    negative,
+		Base10Value: ParsePseudoBigInt(text),
+	}
+}
+
+func ParsePseudoBigInt(stringValue string) string {
+	stringValue = strings.TrimSuffix(stringValue, "n")
+
+	var b1 byte
```
sandersn, line 46: minor, but: having names `b1 bi big` in the same small space is way too hard to read.

Edit: good chance to promote https://www.codingfont.com/

internal/jsnum/pseudobigint.go
sandersn: seems reasonable. wait, does this do base conversion from binary etc to decimal? Or does it just roundtrip whatever was written? The TODO comment in scanner.go implies the former.

Edit: And all the baseline changes.

sandersn (approved):
Looks good with one naming complaint. I left in my other questions that I had as I read, but they were all answered.

internal/jsnum/pseudobigint_test.go
jakebailey: The original parse function always succeeded, so I don't think we are actually needing to deal with this case; the code actually strips leading zeros.

internal/jsnum/pseudobigint.go
jakebailey: Yes, it does base conversion. The original TS code does some fancy stuff with Uint16Buffer that I don't really want to muck around with when Go gets this correct enough to parse out valid bigints, which is all this code works on.

PR #665, opened 2025-03-18
Fix more issues uncovered by tests

ahejlsberg:
In this PR:

* Print `readonly` modifier for read-only tuples
* Fix issue in `combineUnionOrIntersectionMemberSignatures`
* `mapType` should return `nil` when all constituents map to `nil`
* Exact optional property checks + string literal spelling suggestions
* Set `File` property of relation information diagnostics
* Contextual type for expando properties when possible


jakebailey:
This PR might be out of date, I see commits from the previous PR (the flow state stuff).

ahejlsberg:
@jakebailey Should be good now.

jakebailey (approved):
:shipit:

PR #669, opened 2025-03-19
Fix skipTrivia=false in scanner

jakebailey:
We never actually set this to false until #661, and has been broken the whole time. This is really fiddly, especially given the ported code attempts to avoid extra UTF-8 rune decoding. I've had to rewrite some of this to look closer to the original code in order to get skipTrivia to work properly, in any case.


testdata/baselines/reference/submodule/conformance/templateLiteralTypesPatterns.errors.txt.diff
```diff
@@ -77,15 +59,3 @@
...
      // the following should work
-@@= skipped -133, +133 lines =@@
-     
-     // whitespace and comments aren't part of numbers
-     bigints("- 1");
--            ~~~~~
--!!! error TS2345: Argument of type '"- 1"' is not assignable to parameter of type '`${bigint}`'.
-     bigints("-/**/1");
--            ~~~~~~~~
--!!! error TS2345: Argument of type '"-/**/1"' is not assignable to parameter of type '`${bigint}`'.
```
jakebailey, line 88: This is the case we're fixing; skipTrivial=false is required here to make the scanner return a space or comment token before the bigint.

internal/scanner/scanner.go
```diff
@@ -389,13 +389,17 @@ func (s *Scanner) errorAt(diagnostic *diagnostics.Message, pos int, length int,
 	}
 }
 
+// NOTE: even though this returns a rune, it only decodes the next byte.
```
DanielRosenwasser:
```suggestion
// NOTE: even though this returns a rune, it only decodes the current byte.
```

internal/scanner/scanner.go
```diff
@@ -389,13 +389,17 @@ func (s *Scanner) errorAt(diagnostic *diagnostics.Message, pos int, length int,
...
+// It must be checked against utf8.RuneSelf to verify that a call to charAndSize
+// is not needed.
 func (s *Scanner) char() rune {
 	if s.pos < len(s.text) {
 		return rune(s.text[s.pos])
 	}
 	return -1
 }
 
+// NOTE: this returns a rune, but only decodes the next byte.
```
DanielRosenwasser:
```suggestion
// NOTE: this returns a rune, but only decodes the byte at the offset.
```

DanielRosenwasser (approved):

PR #674, opened 2025-03-19
Check JSDoc types

sandersn:
Type checks the JSDoc types

- `* = any`
- `!T = T`
- `?T = T | null`
- `...T = T[]`
- `T= = T | undefined`

Only improves a few TS tests since I don't have `@param` or `@type` contributing type annotations in JS.

The rules for JSVariadicType (`...T`) are simplified: it's now always a synonym for `T[]`. Previously it was only `T[]` when used correctly: as a rest parameter, and was `T | undefined` otherwise.

This makes no sense.

And it turns out that signature help in Strada has always used the simpler `...T = T[]` rule anyway, so that's what JS users are used to seeing as long as they don't have checkJs turned on. What `...` is still used is in random JS code that is going to be mostly viewed rather than edited, so a simplification is overdue.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -20903,8 +20903,21 @@ func (c *Checker) getTypeFromTypeNode(node *ast.Node) *Type {
 
 func (c *Checker) getTypeFromTypeNodeWorker(node *ast.Node) *Type {
 	switch node.Kind {
-	case ast.KindAnyKeyword:
+	case ast.KindAnyKeyword, ast.KindJSDocAllType:
 		return c.anyType
+	case ast.KindJSDocNonNullableType:
+		return c.getTypeFromTypeNode(node.AsJSDocNonNullableType().Type)
+	case ast.KindJSDocNullableType:
+		t := c.getTypeFromTypeNode(node.AsJSDocNullableType().Type)
+		if c.strictNullChecks {
+			return c.getNullableType(t, TypeFlagsNull)
+		} else {
+			return t
+		}
+	case ast.KindJSDocVariadicType:
+		return c.createArrayType(c.getTypeFromTypeNode(node.AsJSDocOptionalType().Type))
+	case ast.KindJSDocOptionalType:
+		return c.addOptionality(c.getTypeFromTypeNode(node.AsJSDocOptionalType().Type))
 	case ast.KindUnknownKeyword:
 		return c.unknownType
 	case ast.KindStringKeyword:
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -20915,7 +20915,7 @@ func (c *Checker) getTypeFromTypeNodeWorker(node *ast.Node) *Type {
 			return t
 		}
 	case ast.KindJSDocVariadicType:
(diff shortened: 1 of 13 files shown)


internal/checker/checker.go
```diff
@@ -20903,8 +20903,21 @@ func (c *Checker) getTypeFromTypeNode(node *ast.Node) *Type {
...
+		return c.getTypeFromTypeNode(node.AsJSDocNonNullableType().Type)
+	case ast.KindJSDocNullableType:
+		t := c.getTypeFromTypeNode(node.AsJSDocNullableType().Type)
+		if c.strictNullChecks {
+			return c.getNullableType(t, TypeFlagsNull)
+		} else {
+			return t
+		}
+	case ast.KindJSDocVariadicType:
+		return c.createArrayType(c.getTypeFromTypeNode(node.AsJSDocOptionalType().Type))
```
gabritto, line 20918:
```suggestion
		return c.createArrayType(c.getTypeFromTypeNode(node.AsJSDocVariadicType().Type))
```

?

jakebailey (dismissed):
I was initially thinking this was very very odd, but this is really just a thing right now because we're missing the grammar errors that complain about this recovery in TS code.

internal/checker/checker.go
```diff
@@ -20903,8 +20903,21 @@ func (c *Checker) getTypeFromTypeNode(node *ast.Node) *Type {
...
 	switch node.Kind {
-	case ast.KindAnyKeyword:
+	case ast.KindAnyKeyword, ast.KindJSDocAllType:
 		return c.anyType
+	case ast.KindJSDocNonNullableType:
+		return c.getTypeFromTypeNode(node.AsJSDocNonNullableType().Type)
+	case ast.KindJSDocNullableType:
+		t := c.getTypeFromTypeNode(node.AsJSDocNullableType().Type)
+		if c.strictNullChecks {
+			return c.getNullableType(t, TypeFlagsNull)
```
jakebailey, line 20913: No undefined?

sandersn: we type `?T` with the Closure-derived type `T | null` but the TS error suggests the Flow-derived `T | null | undefined`. That's because I thought that in a JS file we should give the Closure type, but that in a TS file, we were likely to encounter a recent Flow switcher who didn't know that `?T` was illegal TS.

jakebailey: Weird....

internal/checker/checker.go
jakebailey: Shocked this didn't crash somewhere...

jakebailey (approved):

LukeAbby:
This PR inspired me to check how many you could combine:
```ts
/** @param {...!?number=} args */
function example(...args) {
    args;
}
```

apparently on v5.8.2 `args` will be typed as `number | null | undefined`, causing `...args` to error.

internal/checker/checker.go
sandersn: None of the TS tests use it as far as I know.

PR #676, opened 2025-03-19
Comment in and fix reportNonDefaultExport

frodi-karlsson:
Unlocked by implied node format


testdata/baselines/reference/submodule/compiler/es6ImportDefaultBindingFollowedWithNamespaceBindingWithExport.errors.txt.diff
```diff
@@ -1,18 +1,10 @@
...
-     
--==== client.ts (2 errors) ====
-+==== client.ts (1 errors) ====
+@@= skipped -8, +8 lines =@@
      export import defaultBinding, * as nameSpaceBinding  from "./server";
      ~~~~~~
  !!! error TS1191: An import declaration cannot have modifiers.
 -                  ~~~~~~~~~~~~~~
--!!! error TS1192: Module '"server"' has no default export.
++                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jakebailey: Hm, why's the range here so long?

frodi-karlsson: I passed node when I should've passed node.Name(), will address along with formatting

internal/checker/checker.go
```diff
@@ -13842,7 +13841,7 @@ func (c *Checker) canHaveSyntheticDefault(file *ast.Node, moduleSymbol *ast.Symb
 	}
 
 	// JS files have a synthetic default if they do not contain ES2015+ module syntax (export = is not valid in js) _and_ do not have an __esModule marker
-	return file.AsSourceFile().ExternalModuleIndicator == nil && c.resolveExportByName(moduleSymbol, "__esModule", nil /*sourceNode*/, dontResolveAlias) == nil
+	return file.AsSourceFile().ExternalModuleIndicator == nil /* || file.ExternalModuleIndicator == true) */ && c.resolveExportByName(moduleSymbol, "__esModule", nil /*sourceNode*/, dontResolveAlias) == nil // !!!
```
frodi-karlsson: Found a similar comment for the same conundrum elsewhere, so thought it'd make sense to include it

jakebailey: I would suspect that `ast.IsExternalModule` is the thing to use.

jakebailey: (I'm not sure `true` will ever be a valid value for this)

internal/checker/checker.go
```diff
@@ -13576,7 +13576,27 @@ func (c *Checker) getTargetOfModuleDefault(moduleSymbol *ast.Symbol, node *ast.N
 }
 
 func (c *Checker) reportNonDefaultExport(moduleSymbol *ast.Symbol, node *ast.Node) {
-	// !!!
+	if moduleSymbol.Exports != nil && moduleSymbol.Exports[ast.InternalSymbolNameExportEquals] != nil {
```
jakebailey, line 13579: This condition might be wrong; I'd expect `decoratorMetadataWithImportDeclarationNameCollision4` to pass without a diff with `reportNonDefaultExport` implemented.

testdata/baselines/reference/submodule/conformance/nodeModulesCjsFormatFileAlwaysHasDefault(module=node16).errors.txt
frodi-karlsson, line 1: I might be wrong but I think this regression is due to `getEmitSyntaxForModuleSpecifierExpression` not being implemented, but they're just getting surfaced by these changes. Should I aim to implement that as well to confirm that theory in this PR? I'm torn between scope of changes and want for confidence

jakebailey: If it's straightforward, what you have I think is already an improvement.

internal/checker/checker.go
```diff
@@ -13576,7 +13576,27 @@ func (c *Checker) getTargetOfModuleDefault(moduleSymbol *ast.Symbol, node *ast.N
 }
 
 func (c *Checker) reportNonDefaultExport(moduleSymbol *ast.Symbol, node *ast.Node) {
-	// !!!
+	if moduleSymbol.Exports != nil && moduleSymbol.Exports[node.Symbol().Name] != nil {
```
jakebailey, line 13579:
```suggestion
	if moduleSymbol.Exports[node.Symbol().Name] != nil {
```

A nil map is readable, so this is okay.

internal/checker/checker.go
```diff
@@ -13576,7 +13576,27 @@ func (c *Checker) getTargetOfModuleDefault(moduleSymbol *ast.Symbol, node *ast.N
...
+		c.error(node, diagnostics.Module_0_has_no_default_export_Did_you_mean_to_use_import_1_from_0_instead, c.symbolToString(moduleSymbol), c.symbolToString(node.Symbol()))
+	} else {
+		diagnostic := c.error(node.Name(), diagnostics.Module_0_has_no_default_export, c.symbolToString(moduleSymbol))
+		var exportStar *ast.Symbol
+		if moduleSymbol.Exports != nil {
+			exportStar = moduleSymbol.Exports[ast.InternalSymbolNameExportStar]
+		}
```
jakebailey, line 13586:
```suggestion
		exportStar := moduleSymbol.Exports[ast.InternalSymbolNameExportStar]
```

jakebailey:
What is here seems correct, save for a couple of go-specific nits which I don't think really matter.

jakebailey:
Yeah, to go further we need `getEmitSyntaxForUsageLocation`. #267 contained the port of that but needs to be rebased.

jakebailey (approved):

PR #680, opened 2025-03-20
refactor: use variable for zero value checks to improve clarity

aimuz:
Refactor zero value checks in core functions to use a variable instead of dereferencing new(T) for improved code clarity

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -70,10 +70,11 @@ func MapIndex[T, U any](slice []T, f func(T, int) U) []U {
 }
 
 func MapNonNil[T any, U comparable](slice []T, f func(T) U) []U {
+	var zero U
 	var result []U
 	for _, value := range slice {
 		mapped := f(value)
-		if mapped != *new(U) {
+		if mapped != zero {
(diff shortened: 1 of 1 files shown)


aimuz:

jakebailey:
Are you sure this is actually faster? https://godbolt.org/z/K83vrqos4 shows that they compile to the same thing, and Go doesn't really care where you stick values as escape analysis will treat them the same.

jakebailey:
(I personally think `var zero T` is clearer, but that's not a universally held opinion.)

aimuz:
I apologize for the previous implication of performance improvement. The link you shared on godbolt.org shows that the two approaches compile to the same machine code. The Go compiler's escape analysis and optimization capabilities make these two approaches equivalent in terms of performance.

The main purpose of this refactoring is to improve the readability of the code. The use of `var zero T` is more intuitive and clear than the dereferencing of `new(T)`. I have updated the commit message to accurately reflect this.

jakebailey:
Thanks, but I don't think we're going to take this. This is effectively pure style and I don't think it's worth the churn.

PR #681, opened 2025-03-20
[DO NOT MERGE] Example of what backporting ordering changes does to diffs

jakebailey:
Just for example, this is how the diffs change if we backport the union/symbol/member ordering changes back to Strada and then update the submodule here.

I don't think we're to the point where this is useful yet, but it's promising that so many baselines go away even without full type printing.


testdata/baselines/reference/submodule/compiler/coAndContraVariantInferences2.types.diff
```diff
@@ -42,20 +42,17 @@
  
  function bar(node: Identifier | FunctionDeclaration) {
 ->bar : (node: Identifier | FunctionDeclaration) => void
-->node : Identifier | FunctionDeclaration
 +>bar : (node: FunctionDeclaration | Identifier) => void
```
jakebailey, line 45: Curious difference; the function parameter thing should have changed but I guess Strada is printing unions in a node-preserving order still, which means this PR is not even the best example yet either due to the lack of node reuse in corsa.

jakebailey:
Closing this now that I've proven to myself that this all is workable.

PR #682, opened 2025-03-20
Report duplicate declaration errors on duplicated setters and getters

Andarist:
supersedes https://github.com/microsoft/typescript-go/pull/608

### DIFF
--- a/testdata/baselines/reference/submodule/conformance/symbolDeclarationEmit12.errors.txt.diff
+++ b/testdata/baselines/reference/submodule/conformance/symbolDeclarationEmit12.errors.txt.diff
@@ -1,26 +1,20 @@
 --- old.symbolDeclarationEmit12.errors.txt
 +++ new.symbolDeclarationEmit12.errors.txt
-@@= skipped -0, +-1 lines =@@
--symbolDeclarationEmit12.ts(9,13): error TS2300: Duplicate identifier '[Symbol.toPrimitive]'.
--symbolDeclarationEmit12.ts(10,13): error TS2300: Duplicate identifier '[Symbol.toPrimitive]'.
--
--
+@@= skipped -0, +0 lines =@@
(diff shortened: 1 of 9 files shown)


testdata/baselines/reference/submodule/conformance/symbolDeclarationEmit12.errors.txt.diff
```diff
@@ -1,26 +1,20 @@
 --- old.symbolDeclarationEmit12.errors.txt
 +++ new.symbolDeclarationEmit12.errors.txt
-@@= skipped -0, +-1 lines =@@
--symbolDeclarationEmit12.ts(9,13): error TS2300: Duplicate identifier '[Symbol.toPrimitive]'.
--symbolDeclarationEmit12.ts(10,13): error TS2300: Duplicate identifier '[Symbol.toPrimitive]'.
--
--
+@@= skipped -0, +0 lines =@@
++symbolDeclarationEmit12.ts(5,9): error TS2300: Duplicate identifier '[Symbol.toPrimitive]'.
```
Andarist, line 4: It's because those are reported on *all* duplicate declarations and not just on non-first ones. This is desired as per https://github.com/microsoft/typescript-go/pull/608#issuecomment-2731190882

testdata/tests/cases/compiler/duplicateMembers1.ts
Andarist, line 1: This is an effect of me trying to understand how it should work and testing different scenarios. As @ahejlsberg mentions in [this comment](https://github.com/microsoft/typescript-go/pull/608#issuecomment-2731190882) the Strada's behavior was really order-sensitive and weird at times. The output here is much better/consistent - from what I can tell.

ahejlsberg:
Thanks for pointing out the issue. Which turns out to be multiple issues. The fix in this PR unfortunately isn't the right one, I have put up #693 with what I think we need.

PR #684, opened 2025-03-20
Print `NoInfer` symbol

Andarist:


internal/checker/printer.go
```diff
@@ -188,7 +188,13 @@ func (p *Printer) printTypeNoAlias(t *Type) {
 	case t.flags&TypeFlagsStringMapping != 0:
 		p.printStringMappingType(t)
 	case t.flags&TypeFlagsSubstitution != 0:
-		p.printType(t.AsSubstitutionType().baseType)
+		noInferSymbol := core.IfElse(p.c.isNoInferType(t), p.c.getGlobalNoInferSymbolOrNil(), nil)
+		if noInferSymbol != nil {
```
jakebailey: I wouldn't use an `IfElse` here; an if statement suffices and will be faster since it'll avoid calling `getGlobalNoInferSymbolOrNil` in the common case.

Andarist: done

testdata/baselines/reference/submodule/compiler/noInferUnionExcessPropertyCheck1.types
```diff
@@ -2,20 +2,20 @@
 
 === noInferUnionExcessPropertyCheck1.ts ===
 declare function test1<T extends { x: string }>(
->test1 : <T extends { x: string; }>(a: T, b: () => T | T) => void
+>test1 : <T extends { x: string; }>(a: T, b: () => NoInfer<T> | NoInfer<T>) => void
```
jakebailey, line 5: Nothing you did but the printer definitely mishandles precedence of arrow functions and unions 😄

jakebailey (approved):

testdata/baselines/reference/submodule/compiler/noInferUnionExcessPropertyCheck1.types
Andarist: Yep, I noticed that too - the reason is that the current printer doesn't have any support for parenthesizing rules yet (the old one has a concept of `parenthesizerRule`)

jakebailey: The current printer is just a placeholder, but IIRC the new compiler doesn't use parenthesizer rules at all anymore.

Andarist: Ye, might be - I was taking another look at this a moment ago and it seems that the node builder “transforms” those into ParenthesizedTypes (and then those are simply printed)

PR #687, opened 2025-03-21
Compute type precedence for single signature types to fix their prints in unions

Andarist:


testdata/baselines/reference/submodule/compiler/augmentExportEquals5.types.diff
```diff
@@ -14,6 +14,20 @@
  
          interface IRoute {
              all(...handler: RequestHandler[]): IRoute;
+->all : (...handler: RequestHandler[]) => IRoute
+->handler : RequestHandler[]
++>all : (...handler: (RequestHandler)[]) => IRoute
```
jakebailey: Hm, are we missing unsetting something temporarily such that these extra parens appear?

testdata/baselines/reference/submodule/compiler/classReferencedInContextualParameterWithinItsOwnBaseExpression.types.diff
```diff
@@ -1,6 +1,15 @@
 --- old.classReferencedInContextualParameterWithinItsOwnBaseExpression.types
 +++ new.classReferencedInContextualParameterWithinItsOwnBaseExpression.types
-@@= skipped -15, +15 lines =@@
+@@= skipped -7, +7 lines =@@
+ 
+ interface Schema<A> {
+   readonly pretty?: Pretty<A>;
+->pretty : Pretty<A> | undefined
++>pretty : (Pretty<A>) | undefined
```
jakebailey, line 8: Similar question here.

Andarist: @jakebailey I think I fixed it just now

internal/checker/printer.go
```diff
@@ -22,6 +22,8 @@ func (c *Checker) getTypePrecedence(t *Type) ast.TypePrecedence {
 			return ast.TypePrecedenceTypeOperator
 		case c.isArrayType(t):
 			return ast.TypePrecedencePostfix
+		case t.objectFlags&ObjectFlagsClassOrInterface == 0 && c.getSingleCallOrConstructSignature(t) != nil:
```
jakebailey, line 25: I don't see `getSingleCallOrConstructSignature` being used in this way in Strada; is there a specific reason why this ends up working? Why isn't it `getSingleCallSignature` either?

Andarist: Strada serializes the types into type nodes first and then prints those - so it ends up wrapping `FunctionType` **node** with parenthesis. Currently, Corsa stringifies types *directly* - that's why you won't find this in Strada (it has no concept of `TypePrecedence` either).

> is there a specific reason why this ends up working? 

I just figured out that non-single signature types are serialized differently, they either have non-anonymous symbols  or have to be wrapped with `{ ... }` (if the types have properties or more than a single signature). At that point, they don't have to be parenthesized. 

> Why isn't it getSingleCallSignature either?

Constructor types have to be wrapped with parenthesis too. That's indicated both in the comment above `TypePrecedenceFunction` and it also can be seen on [this playground](https://www.typescriptlang.org/play/?ts=5.9.0-dev.20250321#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVxsAZxgAoBDALjhSQFsAjYKAGjjspWAHc5iBKOALwA+OADcImACb8A3gCg4cKMBhIouALKkYACwB0UUikkQafOAH44pOJToBuOQF85ctDiJLCSADbwBeIQkvHIA9KGKigB6FnJAA)

jakebailey: Yeah, it's all different for sure, but I'm just trying to understand if this will negatively affect things like printing single-function interfaces, or functions with properties, since IIRC `getSingleCallOrConstructSignature` will give something back for those?

Andarist: Interfaces are ignored here based on flags and this function by default wont return a signature if the type has properties (allowMembers is false by default)

jakebailey (approved):

PR #690, opened 2025-03-22
Sort numeric symbols

Andarist:
@jakebailey mentioned this is something the team is considering to do.


internal/checker/utilities.go
```diff
@@ -628,6 +629,17 @@ func (c *Checker) compareSymbolsWorker(s1, s2 *ast.Symbol) int {
 	if s2 == nil {
 		return -1
 	}
+	int1, err1 := strconv.Atoi(s1.Name)
+	int2, err2 := strconv.Atoi(s2.Name)
```
ahejlsberg, line 633: I'm afraid it's not that simple. Consider symbols named `"0"`, `"00"`, and `"000"` and so on! We'd want a check that the symbol *round-trips* as a number. Which unfortunately is a more expensive check.

Andarist: Good point, I can implement that - as long as it's even something you'd like to see getting in. At the moment, the main user-oriented improvement in this is a better error message here: [`testdata/baselines/reference/submodule/conformance/arityAndOrderCompatibility01.errors.txt`](https://github.com/microsoft/typescript-go/pull/690/files#diff-737aac48dc9fc10f82cf63bd08b3b1928b500fb9e6ef1001d4a0537e22067f0e). But at the same time, this test case isn't exercising a particularly popular pattern.

sandersn:
@Andarist it sounds like this change might be too expensive if you put in the work to correct it. Do you want to keep working on this PR? Or should we close it?

PR #692, opened 2025-03-22
Fix strict mode eval or arguments condition in `bindParameter`

frodi-karlsson:
I suspect it's just a 0 -> 9 typo here, I don't see why it would check == 9 on purpose


ahejlsberg (approved):

ahejlsberg:
A really smart compiler or linter would have told us right away that the expression could never be true!

PR #693, opened 2025-03-22
Fix multiple issues related to duplicate identifier checks

ahejlsberg:
We have several inconsistencies related to duplicate identifier checks in the old codebase, including not erroring on all duplicate declarations, erroring in different ways when declarations are re-ordered, or not erroring at all on certain duplicates. The `duplicateIdentifierChecks.ts` compiler test in the PR has examples of all of these issues. Every member declaration in that file is supposed to have a duplicate identifier error on it, but many don't when compiled with the old compiler.

Thanks @Andarist for pointing out the issue.


jakebailey (approved):

Andarist:
It would be good to fix here this outdated comment: https://github.com/microsoft/typescript-go/pull/682/files#diff-8e542d9de5c6b06221168642e2d9dd2df1f2d42d9c79ab60dc28fa6eac85fe39R14555

PR #694, opened 2025-03-22
Implement import meta check unlocked by impliednodeformat

frodi-karlsson:


internal/checker/checker.go
```diff
@@ -9961,10 +9961,10 @@ func (c *Checker) checkNewTargetMetaProperty(node *ast.Node) *Type {
 
 func (c *Checker) checkImportMetaProperty(node *ast.Node) *Type {
 	if c.moduleKind == core.ModuleKindNode16 || c.moduleKind == core.ModuleKindNodeNext {
-		// !!! Enable this once ImpliedNodeFormat is computed at program construction
-		// if ast.GetSourceFileOfNode(node).ImpliedNodeFormat != core.ModuleKindESNext {
-		// 	c.error(node, diagnostics.The_import_meta_meta_property_is_not_allowed_in_files_which_will_build_into_CommonJS_output)
-		// }
+		sourceFileMetaData := c.program.GetSourceFileMetaData(ast.GetSourceFileOfNode(node).Path())
+		if sourceFileMetaData != nil && sourceFileMetaData.ImpliedNodeFormat != core.ModuleKindESNext {
```
jakebailey, line 9965: Should this be:

```suggestion
		if sourceFileMetaData == nil || sourceFileMetaData.ImpliedNodeFormat != core.ModuleKindESNext {
```

frodi-karlsson: I wasn't sure whether we wanted to diagnose a missing `SourceFileMetaData` as having a disallowed import meta property, but I can do that change for sure

frodi-karlsson: Ahh wait my bad I see how that's a fairer port

jakebailey: Yeah. I also may not have noticed this gotcha in the other code that was ported, so it might be worth double checking all uses of `GetSourceFileMetaData` to verify that they were also ported properly.

jakebailey (approved):

internal/checker/checker.go
frodi-karlsson: I see a failed port, I'll open a pr for it shortly

PR #696, opened 2025-03-23
Possibly error for languageVersion < core.ScriptTargetES2015 in checkAsyncFunction

frodi-karlsson:
also cleaned up unnecessary nil map checks from one of my previous prs in `getTargetOfModuleDefault`


jakebailey:
I don't think we are actually looking to port some of this, specifically ES5 itself (and the async stuff).

frodi-karlsson:
> and the async stuff

I'm not sure what's left after the async stuff. Makes sense to close it and leave for someone with clearer project insight? I may have misinterpreted

```Possibly error if c.languageVersion < core.ScriptTargetES2015```

as

```Definitely error if c.languageVersion < core.ScriptTargetES2015 && some other conditions, not implemented yet```

frodi-karlsson:
I see the comment is removed in a to-be-merged pr

PR #698, opened 2025-03-24
Port more of tsc help from strada

haoqixu:
This PR ports more of `tsc help` from `strada`.

### DIFF
--- a/internal/execute/outputs.go
+++ b/internal/execute/outputs.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/compiler"
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
 	"github.com/microsoft/typescript-go/internal/core"
@@ -244,9 +245,9 @@ func generateGroupOptionOutput(sys System, optionsList []*tsoptions.CommandLineO
 	}
 
 	// make sure always a blank line in the end.
-	// !!! if lines[len(lines)-2] != sys.NewLine() {
-	// !!! 	lines = append(lines, sys.NewLine())
-	// !!! }
+	if len(lines) < 2 || lines[len(lines)-2] != sys.NewLine() {
+		lines = append(lines, sys.NewLine())
+	}
 
 	return lines
 }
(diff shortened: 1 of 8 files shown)


internal/execute/outputs.go
```diff
@@ -244,9 +245,9 @@ func generateGroupOptionOutput(sys System, optionsList []*tsoptions.CommandLineO
 	}
 
 	// make sure always a blank line in the end.
-	// !!! if lines[len(lines)-2] != sys.NewLine() {
-	// !!! 	lines = append(lines, sys.NewLine())
-	// !!! }
+	if lines[len(lines)-2] != sys.NewLine() {
```
jakebailey: Can this ever go out of range?

haoqixu: Accessing an array element that's `out of range` in TS/JS returns `undefined`, so to align with `strada`, tsgo needs to insert a newline when the length is less than 2.

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
+			),
+		)
+	}
 
-	// !!!
+	var terminalWidth int
+	// const terminalWidth = sys.getWidthOfTerminal?.() ?? 0;
+
+	// Note: child_process might return `terminalWidth` as undefined.
+	if terminalWidth >= 80 {
```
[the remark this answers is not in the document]
haoqixu: `System` interface is not complete yet, keep these lines as `TODO`.

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
+			text = append(text, sys.NewLine())
+		}
+		text = append(text, sys.NewLine())
+	}
 
 	return text
 }
 
+func formatDefaultValue(defaultValue any, option *tsoptions.CommandLineOption) string {
+	if defaultValue == core.TSUnknown {
```
iisaduan:
```suggestion
	if defaultValue == nil || defaultValue == core.TSUnknown {
```

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
 
 	// name and description
-	// !!! name := getDisplayNameTextOfOption(option)
+	name := getDisplayNameTextOfOption(option)
+
+	// value type and possible value
+	valueCandidates := getValueCandidate(option)
+
+	var defaultValueDescription string
+	if msg, ok := option.DefaultValueDescription.(*diagnostics.Message); ok {
```
iisaduan:
```suggestion
	if msg, ok := option.DefaultValueDescription.(*diagnostics.Message); ok && msg != nil {
```

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
+
+func showAdditionalInfoOutput(valueCandidates *valueCandidate, option *tsoptions.CommandLineOption) bool {
+	if option.Category == diagnostics.Command_line_Options {
+		return false
+	}
+	if valueCandidates.possibleValues == "string" && (option.DefaultValueDescription == nil ||
+		option.DefaultValueDescription == "false" ||
+		option.DefaultValueDescription == "n/a") {
+		return false
+	}
```
iisaduan:
```suggestion
    if valueCandidates != nil && valueCandidates.possibleValues == "string" && (
        option.defaultValueDescription == nil ||
    	option.DefaultValueDescription == "false" ||
		option.DefaultValueDescription == "n/a"
    ) {
		return false
	}
```

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
+	// boolean -- any of: boolean
+	// object -- null
+	// list -- one or more: , content depends on `option.element.type`, the same as others
+	// Map<string, number | string> -- any of: key1, key2, ....
+	if option.Kind == tsoptions.CommandLineOptionTypeObject {
+		return nil
+	}
+
+	res := &valueCandidate{}
+	// !!! Debug.assert(option.type !== "listOrElement");
```
iisaduan: An equivalent of this `debug.assert` can be done with something like

```
if option.Kind == tsoptions.CommandLineOptionTypeListOrElement {
    panic("no value candidate for list or element")
}
```

internal/execute/outputs.go
```diff
@@ -260,13 +261,165 @@ func generateOptionOutput(
...
+		// !!!     if (showAdditionalInfoOutput(valueCandidates, option)) {
+		// !!!         if (valueCandidates) {
+		// !!!             text.push(...getPrettyOutput(valueCandidates.valueType, valueCandidates.possibleValues, rightAlignOfLeft, leftAlignOfRight, terminalWidth, /*colorLeft*/ false), sys.newLine);
+		// !!!         }
+		// !!!         if (defaultValueDescription) {
+		// !!!             text.push(...getPrettyOutput(getDiagnosticText(Diagnostics.default_Colon), defaultValueDescription, rightAlignOfLeft, leftAlignOfRight, terminalWidth, /*colorLeft*/ false), sys.newLine);
+		// !!!         }
+		// !!!     }
+		// !!!     text.push(sys.newLine);
+	} else {
```
iisaduan: It would be nice to leave the `// !!! ` comment near the `terminalWidth` declaration, since that's the part/reason that this is not yet implemented

iisaduan (changes requested):
Thanks for the help! I have some comments about nil/undefined handling and a nit to move a `!!!` comment

internal/execute/outputs.go
haoqixu: I think we can not put a `newline` between `"n/a"` and `)` which results in syntax error in `go`. And the extra `newline` after `(` will be removed by `go fmt`. :thinking:

testdata/baselines/reference/tsc/commandLine/Parse-watch-interval-option.js
iisaduan, line 1: not needed for this PR, but this test shouldn't print help (and the exit status should be a success)

internal/execute/outputs.go
iisaduan: Thats fine! I'm never sure of how `go fmt` ends up formatting statements 😅. Can you add the `valueCandidates != nil` statement?

testdata/baselines/reference/tsc/commandLine/Parse-watch-interval-option.js
haoqixu: I think this is because there is no `tsconfig.json` in the `CWD`. I've updated this test and added another test case.

iisaduan (approved):

jakebailey:
Merge queue doing its job; this needs to be updated for main.

PR #700, opened 2025-03-24
Fix many more issues uncovered by tests

ahejlsberg:


ahejlsberg:
Hmm, looks like there is some code in `checkExternalModuleExports` that is non-deterministic due to map iteration order. That's the reason for the test failures which aren't consistently reproducible. I'll investigate.

internal/checker/checker.go
```diff
@@ -10826,39 +10967,15 @@ func (c *Checker) isPropertyAccessible(node *ast.Node, isSuper bool, isWrite boo
...
-		for _, r := range name {
-			if !stringutil.IsASCIILetter(r) {
-				return false
-			}
-		}
-
-		return true
-	}) && c.isEmptyObjectType(containingType)
+	name := t.symbol.Name
+	return name == "EventTarget" || name == "Node" || name == "Element" || strings.HasPrefix(name, "HTML") && strings.HasSuffix(name, "Element")
```
jakebailey, line 10978: The code that was submitted here was a more faithful port, but I guess it's probably not a problem?

internal/checker/checker.go
```diff
@@ -10826,39 +10967,15 @@ func (c *Checker) isPropertyAccessible(node *ast.Node, isSuper bool, isWrite boo
 }
 
 func (c *Checker) containerSeemsToBeEmptyDomElement(containingType *Type) bool {
-	if c.compilerOptions.Lib == nil || slices.Contains(c.compilerOptions.Lib, "lib.dom.d.ts") {
+	return !slices.Contains(c.compilerOptions.Lib, "dom") && everyContainedType(containingType, hasCommonDomTypeName) && c.isEmptyObjectType(containingType)
```
jakebailey, line 10970: The `dom` thing here is a bugfix; `lib` can never contain `dom`, only `lib.dom.d.ts`.

ahejlsberg: Will fix.

internal/checker/checker.go
ahejlsberg: Yeah, little point in wasting code for the ASCII-only character check.

internal/checker/checker.go
```diff
@@ -1597,17 +1606,14 @@ func (c *Checker) getSuggestionForSymbolNameLookup(symbols ast.SymbolTable, name
...
 	if meaning&ast.SymbolFlagsGlobalLookup != 0 {
-		allSymbols = slices.Concat([]*ast.Symbol{
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "string"),
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "number"),
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "boolean"),
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "object"),
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "bigint"),
-			c.newSymbol(ast.SymbolFlagsTypeAlias, "symbol"),
-		}, allSymbols)
+		for _, s := range []string{"stringString", "numberNumber", "booleanBoolean", "objectObject", "bigintBigInt", "symbolSymbol"} {
```
jakebailey, line 1610: Fancy, I'm not sure I like it 😄

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -340,6 +345,8 @@ func parseHarnessOption(t *testing.T, key string, value any, options *HarnessOpt
 		options.CaptureSuggestions = value.(bool)
 	case "typescriptVersion":
 		options.TypescriptVersion = value.(string)
+	case "strictBuiltinIteratorReturn":
+		options.StrictBuiltinIteratorReturn = value.(bool)
```
jakebailey: This shouldn't be required; all of these ones here are special ones added only for the harness. Are we just missing the parser code for this elsewhere? As in, is the `parsinghelpers.go` change sufficient?

jakebailey: Yeah, checked, and reverting this + the struct change still allows the tests to pass.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -48,24 +48,25 @@ type NamedTestConfiguration struct {
...
+	CurrentDirectory            string
+	Symlink                     string
+	Link                        string
+	NoTypesAndSymbols           bool
+	FullEmitPaths               bool
+	NoCheck                     bool
+	ReportDiagnostics           bool
+	CaptureSuggestions          bool
+	TypescriptVersion           string
+	StrictBuiltinIteratorReturn bool
```
jakebailey: This can be reverted too.

jakebailey (approved):

jakebailey:
LGTM but man is GitHub behind on macOS runners ☹️

PR #701, opened 2025-03-24
add 'Remove()' to vfs

iisaduan:
`vfs.Remove(path)` will remove `path` (and all its contents if `path` is a dir)

### DIFF
--- a/internal/vfs/iovfs/iofs.go
+++ b/internal/vfs/iovfs/iofs.go
@@ -148,6 +148,7 @@ func (vfs *ioFS) WalkDir(root string, walkFn vfs.WalkDirFunc) error {
 }
 
 func (vfs *ioFS) Remove(path string) error {
+	_ = internal.RootLength(path) // Assert path is rooted
 	return vfs.remove(path)
 }
 
--- a/internal/vfs/iovfs/iofs.go
+++ b/internal/vfs/iovfs/iofs.go
@@ -19,6 +19,8 @@ type WritableFS interface {
 	fs.FS
 	WriteFile(path string, data []byte, perm fs.FileMode) error
 	MkdirAll(path string, perm fs.FileMode) error
+	// Removes `path` and all its contents. Will return the first error it encounters.
+	Remove(path string) error
 }
 
 // From creates a new FS from an [fs.FS].
@@ -52,6 +54,7 @@ func From(fsys fs.FS, useCaseSensitiveFileNames bool) vfs.FS {
 
 	var writeFile func(path string, content string, writeByteOrderMark bool) error
 	var mkdirAll func(path string) error
+	var remove func(path string) error
 	if fsys, ok := fsys.(WritableFS); ok {
 		writeFile = func(path string, content string, writeByteOrderMark bool) error {
 			rest, _ := strings.CutPrefix(path, "/")
@@ -64,12 +67,19 @@ func From(fsys fs.FS, useCaseSensitiveFileNames bool) vfs.FS {
 			rest, _ := strings.CutPrefix(path, "/")
 			return fsys.MkdirAll(rest, 0o777)
 		}
+		remove = func(path string) error {
+			rest, _ := strings.CutPrefix(path, "/")
+			return fsys.Remove(rest)
+		}
 	} else {
 		writeFile = func(string, string, bool) error {
 			panic("writeFile not supported")
 		}
 		mkdirAll = func(string) error {
 			panic("mkdirAll not supported")
+		}
+		remove = func(string) error {
+			panic("remove not supported")
 		}
 	}
 
@@ -92,6 +102,7 @@ func From(fsys fs.FS, useCaseSensitiveFileNames bool) vfs.FS {
 		realpath:                  realpath,
 		writeFile:                 writeFile,
 		mkdirAll:                  mkdirAll,
+		remove:                    remove,
 	}
 }
 
@@ -102,6 +113,7 @@ type ioFS struct {
 	realpath                  func(path string) (string, error)
 	writeFile                 func(path string, content string, writeByteOrderMark bool) error
 	mkdirAll                  func(path string) error
+	remove                    func(path string) error
 }
 
 var _ vfs.FS = (*ioFS)(nil)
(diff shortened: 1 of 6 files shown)


internal/vfs/iovfs/iofs.go
```diff
@@ -135,6 +147,10 @@ func (vfs *ioFS) WalkDir(root string, walkFn vfs.WalkDirFunc) error {
 	return vfs.common.WalkDir(root, walkFn)
 }
 
+func (vfs *ioFS) Remove(path string) error {
+	return vfs.remove(path)
```
jakebailey, line 152: This will need a `RootLength` call kinda like `WriteFile`.

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +168,29 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
 	return m.m.Open(string(p))
 }
 
+func (m *mapFS) remove(path string) error {
+	canonical := m.getCanonicalPath(path)
+	file, _, err := m.getFollowingSymlinks(canonical)
```
jakebailey: This shouldn't run `getFollowingSymlinks`; a recursive remove needs to stop at any symlinks.

What might be reasonable here is to:

- Get the canonical path
- If it exists, delete it from both `m.m` and `m.symlinks`.
- If it's a dir, loop over `m.m`, and for any path that starts with `canonical+"/"`, delete it from `m.m` and `m.symlinks`.

jakebailey (changes requested):

internal/vfs/osvfs/os.go
```diff
@@ -165,3 +165,10 @@ func (vfs *osFS) WriteFile(path string, content string, writeByteOrderMark bool)
 	}
 	return vfs.writeFile(path, content, writeByteOrderMark)
 }
+
+func (vfs *osFS) Remove(path string) error {
+	if err := os.Remove(path); err != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
jakebailey: Correct review; RemoveAll would be sufficient.

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +168,29 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
 	return m.m.Open(string(p))
 }
 
+func (m *mapFS) remove(path string) error {
+	canonical := m.getCanonicalPath(path)
+	canonicalString := string(canonical)
+	fileInfo, e := m.m.Stat(canonicalString)
```
jakebailey: The error should be named `err` by convention.

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +168,29 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
...
+
+	if e != nil {
+		return e
```
jakebailey: This should return a nil error if the file isn't found.

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +168,29 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
...
+	}
+	delete(m.m, canonicalString)
+	delete(m.symlinks, canonical)
+
+	if fileInfo.IsDir() {
+		for path := range m.m {
+			canonicalPath := m.getCanonicalPath(path)
```
jakebailey: No need to call `getCanonicalPath` here; `m.m`'s keys are guaranteed to be `canonicalPath`, it's just that due to Go's type system, we can't have `map[canonicalPath]...` and then convert that to `map[string]...`.

internal/vfs/vfstest/vfstest_test.go
```diff
@@ -218,6 +218,22 @@ func TestWritableFS(t *testing.T) {
...
+
+	_ = fs.WriteFile("/foo/bar/test/remove2", "remove2", false)
+	assert.Assert(t, fs.DirectoryExists("/foo/bar/test"))
+	err = fs.Remove("/foo/bar/test")
+	assert.NilError(t, err)
+	assert.Assert(t, !fs.FileExists("/foo/bar/test/remove2"))
+	assert.Assert(t, !fs.DirectoryExists("/foo/bar/test"))
+
+	err = fs.Remove("/foo/bar/test")
+	assert.ErrorContains(t, err, `file does not exist`)
```
jakebailey: I'd add some tests here which use symlinks and verify that the code doesn't follow symlinks.

internal/vfs/vfstest/vfstest.go
jakebailey: We should probably stick a comment on `m` that says it's actually `map[canonicalPath]*fstest.MapFile`

internal/vfs/osvfs/os.go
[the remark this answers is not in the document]
jakebailey: Though, long term, we'll want to add a retry mechanism here. (Even pull in https://pkg.go.dev/github.com/rogpeppe/go-internal/robustio maybe...)

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +170,28 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
 	return m.m.Open(string(p))
 }
 
+func (m *mapFS) remove(path string) error {
+	canonical := m.getCanonicalPath(path)
+	canonicalString := string(canonical)
+	fileInfo, err := m.m.Stat(canonicalString)
+	if err != nil {
+		// file not found
```
jakebailey: I'm not sure this is quite right; the Stat could fail due to a symlink being broken, for example, but removing it should succeed.

I think that this code probably just needs to straight up _only_ use `m.m` and `m.symlinks`.

internal/vfs/vfstest/vfstest.go
```diff
@@ -168,6 +170,28 @@ func (m *mapFS) open(p canonicalPath) (fs.File, error) {
...
+		return nil
+	}
+	delete(m.m, canonicalString)
+	delete(m.symlinks, canonical)
+
+	if fileInfo.IsDir() {
+		for path := range m.m {
+			if strings.HasPrefix(path, canonicalString) {
```
jakebailey, line 187: This needs to be HasPrefix on `canonicalString + "/"`, otherwise removing `/foo/bar` would remove `/foo/barbar`.

jakebailey (approved):

PR #704, opened 2025-03-25
Fixed contextually typed expando members

Andarist:


internal/checker/checker.go
```diff
@@ -15389,8 +15403,33 @@ func (c *Checker) widenTypeInferredFromInitializer(declaration *ast.Node, t *Typ
 
 func (c *Checker) getTypeOfFuncClassEnumModule(symbol *ast.Symbol) *Type {
 	links := c.valueSymbolLinks.Get(symbol)
+	originalLinks := links
 	if links.resolvedType == nil {
+		expando := c.getSymbolOfExpando(symbol.ValueDeclaration)
+		if expando != nil {
```
Andarist: this ports Strada's `mergeJSSymbols` - but since JS is not really involved here right now, I decided to just inline its relevant parts here (at least for the time being)

testdata/baselines/reference/submodule/conformance/typeFromPropertyAssignment30.types.diff
```diff
@@ -0,0 +1,11 @@
+--- old.typeFromPropertyAssignment30.types
++++ new.typeFromPropertyAssignment30.types
+@@= skipped -8, +8 lines =@@
+ }
+ const c: Combo = () => 1
+ >c : Combo
+->() => 1 : { (): number; p: {}; }
++>() => 1 : { (): number; p: { [s: string]: number; }; }
```
Andarist, line 8: I have not investigated why this is different **but** I genuinely feel like this is an improvement. It's not even a user-facing - or I don't know how to position my cursor in this example to get such quick info.

If we think about types alone then we can test how this behaves if we start assigning properties to `p` ([TS playground](https://www.typescriptlang.org/play/?target=99&ts=5.9.0-dev.20250325&ssl=9&ssc=23&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgMIHsC2AjdyDeAUMsgBQCUAXMiAK47QDcxyADgPzX7IDaAztT5gooAOYBdanQZRkAX2ZzCCdCCHIE1DDjwBeMuWS6AfMgCMzBADpWRggsLKbVyEIDyAazsWnrFxCEAUSgodFl9ACIYdHQI5iA)):
```ts
interface Combo {
  (): number;
  p?: { [s: string]: number };
}
const c: Combo = () => 1;
c.p = {};

c.p.testOk = 1; // ok
c.p.testError = "foo"; // error
```

So it's not really correct, in my eyes, to say that `p` has `{}` type. At any stage of the process. Either we know here about everything about expando members or we shouldn't know anything (if this would print a flow type or smth).

If I get the `.types` print for the above in Strada I end up with this:
```
interface Combo {
  (): number;
  p?: { [s: string]: number };
>p : { [s: string]: number; } | undefined
>  : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>s : string
>  : ^^^^^^
}
const c: Combo = () => 1;
>c : Combo
>  : ^^^^^
>() => 1 : { (): number; p: {}; }
>        : ^^^^^^^^^^^^^^^^^^^^^^
>1 : 1
>  : ^

c.p = {};
>c.p = {} : {}
>         : ^^
>c.p : { [s: string]: number; }
>    : ^^^^^^^^^^^^^^^^^^^^^^^^
>c : Combo
>  : ^^^^^
>p : { [s: string]: number; }
>  : ^^^^^^^^^^^^^^^^^^^^^^^^
>{} : {}
>   : ^^

c.p.testOk = 1;
>c.p.testOk = 1 : 1
>               : ^
>c.p.testOk : number
>           : ^^^^^^
>c.p : { [s: string]: number; }
>    : ^^^^^^^^^^^^^^^^^^^^^^^^
>c : Combo
>  : ^^^^^
>p : { [s: string]: number; }
>  : ^^^^^^^^^^^^^^^^^^^^^^^^
>testOk : number
>       : ^^^^^^
>1 : 1
>  : ^

c.p.testError = "foo";
>c.p.testError = "foo" : "foo"
>                      : ^^^^^
>c.p.testError : number
>              : ^^^^^^
>c.p : { [s: string]: number; }
>    : ^^^^^^^^^^^^^^^^^^^^^^^^
>c : Combo
>  : ^^^^^
>p : { [s: string]: number; }
>  : ^^^^^^^^^^^^^^^^^^^^^^^^
>testError : number
>          : ^^^^^^
>"foo" : "foo"
>      : ^^^^^
```

ahejlsberg:
I don't think we need any of this logic. I already have a fix for the errors in the `propertyAssignmentUseParentType1.ts` test and it will be in my next PR. Beyond that, the other test differences in this PR don't look right.

In case you're curious, the fix we need is in the `canGetContextualTypeForAssignmentDeclaration` function. The first line of code should be changed to:

```go
symbol := c.getExportSymbolOfValueSymbolIfExported(c.getResolvedSymbol(node.Expression()))
```

Andarist:
> Beyond that, the other test differences in this PR don't look right.

Hm, I took another look at them right now and both look pretty OK to me. `expandoFunctionContextualTypes` matches Strada here. The other one (`typeFromPropertyAssignment30`) also looks OK to me, I mentioned that in [here](https://github.com/microsoft/typescript-go/pull/704/files#r2011798110). It would be interesting to hear your take on it (or both?) - perhaps my mental model for how this is supposed to work is off. Merged symbols and alikes are always somewhat confusing to me ;p

> In case you're curious, the fix we need is in the canGetContextualTypeForAssignmentDeclaration function

Cool. That makes sense! I was looking at that function but I followed the Strada's compatibility~ (in terms of how some of those involved nodes/structures were connected) instead 😅

ahejlsberg:
Some of the code in Strada related to expando properties is overly complicated and at times incorrect, in particular when it comes to our support for inference in .js files. In general, the implementation of our support for .js and JSDoc is going to differ from Strada. For example, see #610.

Conceptually, in a construct like the following

```ts
const fn: SomeType = () => { ... }
fn.prop = someExpr
```

the type of the expression `() => { ... }` is augmented with the expando properties assigned to `fn`, regardless of any type annotation on `fn`. However, when a type annotation is present on `fn`, that type serves as the contextual type for the expression used to initialize expando properties. So above, `someExpr` is contextually typed by `SomeType["prop"]`.

PR #705, opened 2025-03-25
do not print stats if quiet option is given

S-YOU:


S-YOU:

jakebailey:
Sorry, but I don't think we're going to take this. If we do this then quiet will have silenced all output entirely. The point of the flag is to hide a massive amount of errors specifically to only see the stats at the bottom for perf testing.

S-YOU:
okay, no problem.

By the way, the motivation was is that, in my Next.js project, I fixed all the errors on my side.
The difference with `tsc` and `tsgo` is one warning message and stats, so I was trying to make it similar to `tsc`

```
% tsc
%
```

```
% tsgo
Warning: The project contains unsupported file types (.tsx), which are currently not fully type-checked.
Files:             1929
Lines:           203396
Identifiers:     221460
Symbols:         192842
Types:            31604
Instantiations:   33094
Memory used:    158647K
Memory allocs:  1059926
Parse time:      0.178s
Bind time:       0.062s
Check time:      0.090s
Total time:      0.330s
```

PR #706, opened 2025-03-25
Fix empty array tsconfig parsing, ensure helpers are nil-preserving

### LINKED ISSUE #623: types: [] in tsconfig.json is treated as "all visible types"

jakebailey:
Fixes #623
Replaces #624

The issues here were somewhat wide reaching:

- `core.Map` and `core.MapIndex` were not nil-preserving. This led to parsing turning empty slices into nil slices, which is a problem when `nil` means "was not provided".
- `parseStringArray` was not nil-preserving.
- `setPropertyValue` was not nil-preserving.
- `getPropFromRaw` was not nil-preserving.

Fixing these removes a bunch of extra code from `parseJsonToStringKey`, too.

Annoyingly, `core.Pool` _also_ returns `nil` instead of an empty slice, so we can end up with `nil` lists in the tree rather than empty slices. I've not changed this yet, since we can just say "the parser may produce nil slices for NodeLists", but I suspect this will be a problem for API use given `nil` should really mean `undefined` and that's what the JSON package may do depending on the tags.


internal/tsoptions/parsinghelpers.go
```diff
@@ -25,7 +25,10 @@ func parseTristate(value any) core.Tristate {
 
 func parseStringArray(value any) []string {
 	if arr, ok := value.([]any); ok {
-		var result []string
+		if arr == nil {
```
navya9singh, line 28: Is it possible to have a nil `arr` if it has entered the if block?

jakebailey: Yes, `[]any(nil)` is a valid value.

navya9singh (approved):

PR #707, opened 2025-03-25
Port SubtreeFacts (nee TransformFlags) from Strada

rbuckton:
This ports `TransformFlags`, now called `SubtreeFacts`, from Strada. After some profiling I've opted to make some changes in the design to avoid increasing execution time during parse so as not to impact `noCheck` scenarios. Rather than introducing a `transformFlags` field to `Node` as we have in Strada, I've opted to only introduce a `flags` field to nodes whose `SubtreeFlags` cannot be trivially calculated, which reduces the additional memory footprint necessary to store this information.

For `NodeList`, we do not cache `SubtreeFacts` at all, instead relying on the parent node to cache the results. This does mean that reusing a `NodeList` during update will necessitate a shallow traversal of the list, so we may opt to revisit this in the future.

Notably, this approach does not use a side table on `EmitContext`, as `SubtreeFacts` can conceivably be used elsewhere in the compiler, such as the checker, and may allow us to eventually deprecate overlapping `NodeFlags` (such as `NodeFlagsPossiblyContainsDynamicImport`). My goal would be for `NodeFlags` to primarily indicate top-down information, either about the node itself (e.g., `NodeFlagsLet`, `NodeFlagsConst`, etc.), or about the context in which the node was parsed (e.g., `NodeFlagsYieldContext`, `NodeFlagsAwaitContext`, etc.), while `SubtreeFacts` indicates bottom-up information about the node and its descendants. In the long term, this might mean moving flags like `NodeFlagsHasExplicitReturn` as well.

There have been a few changes to how `SubtreeFacts` are propagated to simplify the design somewhat and align with changes we've made to the emitter (i.e., no more JIT substitution). As a result, this is not a 100% direct port so care should be taken when porting code that previously depended on `TransformFlags`.

A quick note on the implementation of `SubtreeFacts()`: In an effort to avoid numerous conditional checks based on whether a node embeds `CompositeNodeBase`, we instead implement a `subtreeFactsWorker` method on `NodeDefault` that, by default, forwards directly to the most derived implementation of `computeSubtreeFacts` on the argument. `CompositeNodeBase` then overrides `subtreeFactsWorker` to implement caching.


internal/ast/ast.go
```diff
@@ -1749,6 +1774,35 @@ func (node *Node) JSDoc(file *SourceFile) []*Node {
 	return nil
 }
 
+// compositeNodeBase
+
+// A composite node is a node that contains a complex subtree. This struct is intended to be
+// embedded in a node that requires caching for its subtree facts.
+type compositeNodeBase struct {
+	facts SubtreeFacts // caches the SubtreeFacts for this node and its subtree
```
jakebailey: Once set, are these ever updated? If so, the races should be fixable by adding a sync.Once or something here. That or, we don't bother making this lazy and just force the parser to do it to save the memory cost of the extra prop.

testdata/baselines/reference/submodule/conformance/asyncConstructor_es6.js.diff
```diff
@@ -0,0 +1,10 @@
+--- old.asyncConstructor_es6.js
++++ new.asyncConstructor_es6.js
+@@= skipped -7, +7 lines =@@
+ 
+ //// [asyncConstructor_es6.js]
+ class C {
+-    constructor() {
++    async constructor() {
```
jakebailey, line 8: This seems like a bug or something; the constructor is getting the modifiers from another node. The same happens below for `accessor constructor()` too.

internal/ast/ast.go
rbuckton: I initially looked into doing this eagerly:

```
                                                            │    sec/op    │    sec/op      vs base                │
Parse/empty.ts/tsc-8                                          603.7n ±  2%    584.5n ±  5%   -3.18% (p=0.035 n=10)
Parse/empty.ts/server-8                                       607.1n ± 16%    590.1n ±  5%        ~ (p=0.165 n=10)
Parse/checker.ts/tsc-8                                        50.49m ± 17%    51.37m ±  3%        ~ (p=0.315 n=10)
Parse/checker.ts/server-8                                     48.31m ±  9%    51.66m ±  1%   +6.94% (p=0.019 n=10)
Parse/dom.generated.d.ts/tsc-8                                16.59m ± 29%    15.94m ±  2%   -3.92% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-8                             24.77m ±  9%    24.08m ±  5%   -2.81% (p=0.043 n=10)
Parse/Herebyfile.mjs/tsc-8                                    955.7µ ±  4%   1054.7µ ±  3%  +10.35% (p=0.002 n=10)
Parse/Herebyfile.mjs/server-8                                 940.2µ ±  1%   1031.3µ ± 31%   +9.69% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-8      305.8µ ±  5%    307.5µ ± 12%        ~ (p=0.529 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-8   483.0µ ±  4%    478.0µ ±  5%        ~ (p=0.796 n=10)
geomean                                                       739.3µ          749.8µ         +1.42%
```
vs. lazily:
```
                                                            │   old.txt    │              lazy.txt               │
                                                            │    sec/op    │    sec/op     vs base               │
Parse/empty.ts/tsc-8                                          603.7n ±  2%   598.1n ± 16%       ~ (p=0.315 n=10)
Parse/empty.ts/server-8                                       607.1n ± 16%   603.4n ±  3%       ~ (p=0.739 n=10)
Parse/checker.ts/tsc-8                                        50.49m ± 17%   46.10m ±  2%  -8.69% (p=0.001 n=10)
Parse/checker.ts/server-8                                     48.31m ±  9%   47.07m ±  5%       ~ (p=0.165 n=10)
Parse/dom.generated.d.ts/tsc-8                                16.59m ± 29%   15.60m ±  3%  -5.94% (p=0.001 n=10)
Parse/dom.generated.d.ts/server-8                             24.77m ±  9%   23.69m ± 13%  -4.35% (p=0.035 n=10)
Parse/Herebyfile.mjs/tsc-8                                    955.7µ ±  4%   940.1µ ±  1%  -1.63% (p=0.009 n=10)
Parse/Herebyfile.mjs/server-8                                 940.2µ ±  1%   938.9µ ±  1%       ~ (p=0.739 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-8      305.8µ ±  5%   300.3µ ±  1%  -1.78% (p=0.003 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-8   483.0µ ±  4%   468.9µ ±  1%  -2.92% (p=0.002 n=10)
geomean                                                       739.3µ         717.2µ        -2.99%
```

The memory footprint was the same:
```
                                                            │   old.txt    │               lazy.txt                │
Parse/empty.ts/tsc-8                                            816.0 ± 0%     816.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/empty.ts/server-8                                         816.0 ± 0%     816.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts/tsc-8                                        24.20Mi ± 0%   25.24Mi ± 0%  +4.27% (p=0.000 n=10)
Parse/checker.ts/server-8                                     24.38Mi ± 0%   25.41Mi ± 0%  +4.22% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-8                                8.644Mi ± 0%   8.943Mi ± 0%  +3.45% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-8                             11.01Mi ± 0%   11.31Mi ± 0%  +2.70% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-8                                    437.3Ki ± 0%   447.5Ki ± 0%  +2.33% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-8                                 437.1Ki ± 0%   447.5Ki ± 0%  +2.39% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-8      159.6Ki ± 0%   163.6Ki ± 0%  +2.48% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-8   236.6Ki ± 0%   240.5Ki ± 0%  +1.63% (p=0.000 n=10)
geomean                                                       441.8Ki        452.1Ki       +2.34%
¹ all samples are equal
```

However, adding a `sync.Once` would add overhead to the lazy version. I'm looking into whether I can just force `SubtreeFacts` to be precomputed prior to emit, since all synthetic nodes will be unique to that thread.

rbuckton: The Herbyfile.mjs tests seem very swingy (+9-10%), but they're also in the sub-millisecond range so they're likely to be more prone to small changes.

testdata/baselines/reference/submodule/conformance/asyncConstructor_es6.js.diff
rbuckton: Not a bug, per se, just different from Strada. We already report an error that `async` isn't allowed on `constructor`, so this change seemed to be acceptable.

internal/ast/ast.go
jakebailey: We can also stick the `sync.Once` in the SourceFile and then do it once at emit time for all "real" nodes before emit. But that's probably what you're saying.

The other alternative if zero means "unset" is to instead store the flags atomically, then do something like:

```go
func (node *Node) compute() {
	if flags := node.flags.Load(); flags != 0 {
		return flags
	}

	// ... do work
	node.flags.Set(flags)
	return flags
}
```

Which will in the worst case repeat work during a race, but usually not. (This is now Node IDs and Symbol IDs are handed out)

rbuckton: An `atomic.Uint32` might work as it's also only 4 bytes. A `sync.Once` is 12 bytes in addition to the 4 bytes for the field. Since the operation is idempotent, a race would only impact time, not correctness.

rbuckton:
Lazy init with atomic seems fairly consistent to without atomic:

```
                                                            │    old.txt    │           lazy-atomic.txt            │
                                                            │    sec/op     │    sec/op     vs base                │
Parse/empty.ts/tsc-8                                           612.3n ± 24%   624.7n ± 48%        ~ (p=0.481 n=10)
Parse/empty.ts/server-8                                        634.1n ±  3%   601.8n ±  2%   -5.09% (p=0.001 n=10)
Parse/checker.ts/tsc-8                                         46.31m ±  3%   46.56m ± 10%        ~ (p=0.912 n=10)
Parse/checker.ts/server-8                                      46.82m ±  2%   47.29m ±  3%        ~ (p=0.143 n=10)
Parse/dom.generated.d.ts/tsc-8                                 15.94m ±  1%   15.84m ±  2%        ~ (p=0.579 n=10)
Parse/dom.generated.d.ts/server-8                              24.60m ±  4%   23.33m ± 12%   -5.18% (p=0.035 n=10)
Parse/Herebyfile.mjs/tsc-8                                    1076.0µ ± 50%   954.4µ ±  3%  -11.30% (p=0.002 n=10)
Parse/Herebyfile.mjs/server-8                                  972.3µ ± 14%   948.2µ ±  0%   -2.48% (p=0.001 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-8       301.6µ ±  2%   305.2µ ±  1%   +1.20% (p=0.029 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-8    471.9µ ±  1%   478.7µ ±  2%   +1.44% (p=0.007 n=10)
geomean                                                        739.9µ         725.6µ         -1.93%

                                                            │   old.txt    │            lazy-atomic.txt            │
Parse/empty.ts/tsc-8                                            816.0 ± 0%     816.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/empty.ts/server-8                                         816.0 ± 0%     816.0 ± 0%       ~ (p=1.000 n=10) ¹
Parse/checker.ts/tsc-8                                        24.21Mi ± 0%   25.23Mi ± 0%  +4.25% (p=0.000 n=10)
Parse/checker.ts/server-8                                     24.38Mi ± 0%   25.41Mi ± 0%  +4.22% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-8                                8.645Mi ± 0%   8.941Mi ± 0%  +3.43% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-8                             11.01Mi ± 0%   11.31Mi ± 0%  +2.71% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-8                                    437.1Ki ± 0%   447.6Ki ± 0%  +2.41% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-8                                 437.1Ki ± 0%   447.4Ki ± 0%  +2.37% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-8      159.6Ki ± 0%   163.6Ki ± 0%  +2.45% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-8   236.8Ki ± 0%   240.4Ki ± 0%  +1.55% (p=0.000 n=10)
geomean                                                       441.8Ki        452.1Ki       +2.33%
¹ all samples are equal
```

testdata/baselines/reference/submodule/conformance/asyncConstructor_es6.js.diff
jakebailey: Is it a problem for our emitter to technically output invalid code?

internal/ast/ast.go
```diff
@@ -1845,12 +1952,18 @@ func IsQualifiedName(node *Node) bool {
 	return node.Kind == KindQualifiedName
 }
 
+func (node *QualifiedName) computeSubtreeFacts() SubtreeFacts {
```
weswigham, line 1955: Is there a perf benefit to hardcoding the base impl of so many `computeSubtreeFacts` rather than visiting each child in a base impl and or'ing the results? I definitely get the ones that introduce a new fact, though.

rbuckton: We don't have a visitor mechanism that can do that efficiently. `ForEachChild` exits early and thus requires a closure to compute that state, and `VisitEachChild` only returns nodes. The closure would get expensive since it can't be cached. If the concern is code size, we could reduce the size by introducing something like a `UnaryExpressionBase` that is shared by `delete`, `void`, `await`, `typeof`, and prefix/postfix unary. I'm already cutting out a lot of repetition for type nodes and the like by doing something similar.

weswigham: Hm, should we have the more general `ReduceEachChild`, or an actual child iterator, then? I can think of a few other fact-finding walks that work the same way.

testdata/baselines/reference/submodule/conformance/asyncConstructor_es6.js.diff
rbuckton: No, we do it all the time in the presence of errors. We make no guarantees on whether syntactically invalid input code will produce syntactically valid output code.

internal/ast/ast.go
rbuckton: `ForEachChild` *almost* matches the signature for `iter.Seq[T]`. We could implement a `Children` iterator on `Node` as simply as:

```go
func (node *Node) Children() iter.Seq[*Node] {
	return func(yield func(*Node) bool) {
		node.ForEachChild(yield)
	}
}
```

However, we don't currently have any `Benchmark` tests involving emit, so it's difficult to gauge the relative performance costs between the alternatives.

rbuckton: Ah sorry, that's inverted. `ForEachChild` returns `true` when done and `false` to continue iterating, while `yield` expects the opposite. That makes it:

```go
func (node *Node) Children() iter.Seq[*Node] {
	return func(yield func(*Node) bool) {
		node.ForEachChild(func(n *Node) bool {
			return !yield(n)
		})
	}
}
```

Which generates two closures and is likely less efficient.

rbuckton: Definitely less efficient:

```
                                                    │     old.txt     │                new.txt                │
                                                    │     sec/op      │    sec/op      vs base                │
Bind/empty.ts-8                                          272.1n ± 16%    261.6n ± 18%        ~ (p=0.684 n=10)
Bind/checker.ts-8                                        18.38m ±  9%    27.72m ±  3%  +50.88% (p=0.000 n=10)
Bind/dom.generated.d.ts-8                                7.678m ±  5%   10.849m ±  5%  +41.30% (p=0.002 n=10)
Bind/Herebyfile.mjs-8                                    303.4µ ±  8%    476.3µ ±  4%  +56.97% (p=0.000 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-8      139.1µ ± 19%    197.3µ ± 19%  +41.82% (p=0.000 n=10)
geomean                                                  276.7µ          374.8µ        +35.47%

                                                    │     old.txt     │                 new.txt             
    │
                                                    │      B/op       │     B/op       vs base                  │
Bind/empty.ts-8                                            192.0 ± 0%      192.0 ± 0%        ~ (p=1.000 n=10) ¹
Bind/checker.ts-8                                        7.206Mi ± 0%   11.262Mi ± 0%  +56.28% (p=0.000 n=10)
Bind/dom.generated.d.ts-8                                4.873Mi ± 0%    6.559Mi ± 0%  +34.62% (p=0.000 n=10)
Bind/Herebyfile.mjs-8                                    173.8Ki ± 0%    243.4Ki ± 0%  +40.08% (p=0.000 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-8      114.4Ki ± 0%    140.7Ki ± 0%  +22.98% (p=0.000 n=10)
geomean                                                  168.9Ki         218.5Ki       +29.37%
¹ all samples are equal
```

This is after switching `bindEachChild` to use `range node.Children()` instead of `ForEachChild()` directly. Duplicating code for specific purposes is definitely more efficient, so a `ReduceEachChild` might be feasible, but a general purpose `Children()` iterator might not be worthwhile.

rbuckton: Unfortunately, we can't write a general purpose `ReudceEachChild[T]` due to Go's restrictions on generics, so we're left with either a `ReduceEachChildUint32` with a lot of casting, or a `ReduceEachChild` for `any` which incurs boxing and unboxing.

rbuckton: We don't really have any other cases in Corsa that would benefit from `ReduceEachChild` at present. IMO, we should wait on adding `ReduceEachChild` until we need it for more than this one case.

jakebailey (approved):

PR #709, opened 2025-03-25
Check correct token in `checkGrammarForInvalidQuestionMark`

Andarist:

PR #710, opened 2025-03-25
More fixes uncovered by tests

ahejlsberg:


internal/checker/grammarchecks.go
```diff
@@ -1002,7 +1002,7 @@ func (c *Checker) checkGrammarForGenerator(node *ast.Node) bool {
 }
 
 func (c *Checker) checkGrammarForInvalidQuestionMark(postfixToken *ast.TokenNode, message *diagnostics.Message) bool {
-	return postfixToken != nil && postfixToken.Kind == ast.KindQuestionQuestionToken && c.grammarErrorOnNode(postfixToken, message)
+	return postfixToken != nil && postfixToken.Kind == ast.KindQuestionToken && c.grammarErrorOnNode(postfixToken, message)
```
jakebailey, line 1005: FWIW this is also #709 which slow CI isn't merging ☹️

internal/jsnum/pseudobigint.go
```diff
@@ -34,20 +39,15 @@ func (value PseudoBigInt) Sign() int {
...
-	return PseudoBigInt{
-		Negative:    negative,
-		Base10Value: ParsePseudoBigInt(text),
-	}
+	return NewPseudoBigInt(ParsePseudoBigInt(text), negative)
 }
 
 func ParsePseudoBigInt(stringValue string) string {
 	stringValue = strings.TrimSuffix(stringValue, "n")
-
```
jakebailey, line 45: What do you have against my newlines? 😄

ahejlsberg: Just not a fan of blank lines in code blocks. Would rather have comments separating the spans, or a separate function if it's important enough.

jakebailey: I can't say I agree but I guess it's fine...

jakebailey (approved):

PR #712, opened 2025-03-25
Print reduced types

Andarist:


internal/checker/printer.go
```diff
@@ -68,6 +68,9 @@ func (c *Checker) TypeToString(t *Type) string {
 
 func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags) string {
 	p := c.newPrinter(flags)
+	if flags&TypeFormatFlagsNoTypeReduction == 0 {
+		t = c.getReducedType(t)
+	}
```
jakebailey, line 73: Does this need to be deeper in `printType` itself? I'm looking at Strada and it's right at the top of `typeToTypeNodeWorker` which is recursive on itself.

jakebailey: Hm, changing that sometimes helps, sometimes doesn't. Odd.

Andarist: `printType` here is also indirectly recursive

> Hm, changing that sometimes helps, sometimes doesn't. Odd.

Could you elaborate?

jakebailey: The function in this PR is `typeToStringEx`, not `printType`, so is only applying to the topmost type.

Andarist: hm, I see - I think normally this would be an issue so that's a good catch overall. But, if I'm not mistaken, it's actually (accidentally) correct here. This flag is only used by `elaborateNeverIntersection` and that only aims to avoid the topmost reduction in the print. Only toplevel properties can make the intersection reduced to `never` (see `isNeverReducedProperty` and its callers).

jakebailey: I'm confused; are you saying that typeToTypeNodeWorker eventually drops its flags somewhere?

Andarist: No, it uses it throughout the the whole context. I think there is an argument to be made that `nested` in [this error message](https://www.typescriptlang.org/play/?ts=5.9.0-dev.20250326#code/CYUwxgNghgTiAEYD2A7AzgF3kgRgKwC54BvAKHngDMkkiAiADzoG5z4URMRgiyKKcsek1YUAvqzHwAZCTbVa8OgE8WbDlx5z+8QTHqrR8CaROlceAHQLWpUJFgJk6LBYBMvdZwzdPOvcJq4pIy2uzevmECQkqGbCZm7rbulho+wLb20HCIqJjY+ADMflQ0gUZpkXz8AUoi8aHV0fqxQcYhstUKBmqJRdY0rEA) should still print as `never` (in `obj3`) - given that it's of type `never` (reduced, can be seen on `obj2`) and the intersection elaboration is only concerned about the top level properties. But it's certainly easier to just use this flag throughout the whole print. OTOH, `obj2`'s print proves me wrong already :D given intersections are not reduced eagerly in a lot of cases... the nested intersection is just not reduced at all when I hover over `obj2`.

But either way... both solutions are probably just fine and this is a minor thing. I **have expected** `obj3`'s error to be different between this branch and Strada... but it's actually the same. It turns out that `c.TypeToString` ends up calling `p.PrintType` and that won't reduce at all, at any stage (from what I can tell). So it ends up behaving the same way according to my tests.

jakebailey: I feel like we're talking about different things. What I'm saying specifically is:

- In Strada, `typeToString` is effectively `typeToTypeNode` + printing.
- `typeToTypeNode` is a simple wrapper on `typeToTypeNodeWorker`.
- `typeToTypeNodeHelper` calls `typeToTypeNodeHelper` to convert child types.
- The top of `typeToTypeNodeWorker` begins with `if (!(context.flags & NodeBuilderFlags.NoTypeReduction)) { type = getReducedType(type); }`, so this means that the reduction must apply to types lower in the tree.
- The current equivalent to `typeToTypeNode` is `printType`.
- This PR adds the reduction to `typeToStringEx`, which is the equivalent to `typeToString` in Strada. But, I would have expected the reduction to have done in `printType`, since that is the equivalent to `typeToTypeNode` in the new code.

Andarist: I have trouble pushing it into `p.printType` due to lack of node reuse. Some `.types` baselines change, especially when it comes to annotated parameter and return types - the inferred ones are reduced, annotated are not (given they are just reduced in Strada).

IIRC you are still porting the node builder so this code will become way closer to Strada and that it might get fixed then. I'll just close this one

jakebailey: To be clear this PR is still an improvement, I think. I'm just trying to figure out the right balance by looking back at Strada.

Andarist: If you want me to reopen this - let me know 😅 You could try to check out this patch locally to see what happens (the diff isn't overwhelming):
```diff
diff --git a/internal/checker/printer.go b/internal/checker/printer.go
index 1b3901682..dcff9b9ef 100644
--- a/internal/checker/printer.go
+++ b/internal/checker/printer.go
@@ -68,9 +68,6 @@ func (c *Checker) TypeToString(t *Type) string {
 
 func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags) string {
 	p := c.newPrinter(flags)
-	if flags&TypeFormatFlagsNoTypeReduction == 0 {
-		t = c.getReducedType(t)
-	}
 	p.printType(t)
 	return p.string()
 }
@@ -156,7 +153,9 @@ func (p *Printer) printType(t *Type) {
 		p.print("...")
 		return
 	}
-
+	if p.flags&TypeFormatFlagsNoTypeReduction == 0 {
+		t = p.c.getReducedType(t)
+	}
 	if t.alias != nil && (p.flags&TypeFormatFlagsInTypeAlias == 0 || p.depth > 0) {
 		p.printQualifiedName(t.alias.symbol)
 		p.printTypeArguments(t.alias.typeArguments)
```

I think the diff above shouldn't be taken as a patch, given it regresses the `.types`. OTOH, this could have been in there the whole time and that regression could have been there the whole time then but it's hard to say which tradeoff you prefer here :p

jakebailey: That's effectively the change I was suggesting, but I wasn't sure why it wasn't a net improvement on top of this PR, which was what https://github.com/microsoft/typescript-go/pull/712#discussion_r2013054650 was mainly thinking aloud about.

I'd reopen it, just since it does fix a lot of stuff and does not appear to regress anything.

jakebailey: Sorry to be difficult, I'm just doing a lot of thinking aloud trying to ensure things are working.

Andarist: > Sorry to be difficult, I'm just doing a lot of thinking aloud trying to ensure things are working.

That's totally fine. Don't worry - you are asking good questions :)

> I'd reopen it, just since it does fix a lot of stuff and does not appear to regress anything.

Done

jakebailey (approved):

PR #715, opened 2025-03-26
Fixed `getContextuallyTypedParameterType` for IIFEs without arguments

Andarist:


internal/checker/checker.go
```diff
@@ -27110,7 +27110,7 @@ func (c *Checker) getContextuallyTypedParameterType(parameter *ast.Node) *Type {
 		return nil
 	}
 	iife := ast.GetImmediatelyInvokedFunctionExpression(fn)
-	if iife != nil && len(iife.Arguments()) != 0 {
```
jakebailey, line 27113: The original code says:

```
if (iife && iife.arguments) {
```

Should this be `iife != nil && iife.Arguments() != nil`, given `Arguments` is nil-preserving?

jakebailey: Hm, no, I tested and that doesn't work. I'm not sure I get why two behave differently.

Andarist: In Strada `iife.arguments` is an empty `NodeArray` here. In Corsa here, it's just `nil` - I assumed that it was not allocated on purpose.

Andarist: I have not tested that but there is a chance that `iife.arguments` is just always truthy in Strada.

jakebailey: I think that's honestly a bug in the parser. I was already looking at this via #706; let me recheck that.

jakebailey: Oh, yeah.

```
export interface CallExpression extends LeftHandSideExpression, Declaration {
    readonly kind: SyntaxKind.CallExpression;
    readonly expression: LeftHandSideExpression;
    readonly questionDotToken?: QuestionDotToken;
    readonly typeArguments?: NodeArray<TypeNode>;
    readonly arguments: NodeArray<Expression>;
}
```

`arguments` is truthy. I'm double checking this in the tests, though.

jakebailey: Yeah, this passes:

image


Funny!

jakebailey (approved):

PR #719, opened 2025-03-26
Initial support for @ts-nocheck

tmm1:
I'm sure this is incorrect and/or incomplete, but it worked in my limited testing.


internal/compiler/program.go
```diff
@@ -330,6 +330,11 @@ func (p *Program) getSemanticDiagnosticsForFile(sourceFile *ast.SourceFile) []*a
 			diags = append(diags, checker.GetDiagnosticsWithoutCheck(sourceFile)...)
 		}
 	}
+	for _, directive := range sourceFile.CommentDirectives {
+		if directive.Kind == ast.CommentDirectiveKindNoCheck {
+			return nil
+		}
+	}
```
jakebailey: I think this isn't quite right; what we should be doing here is skipping the checker loop above if this condition is true.

Something like:

```go
if !slices.ContainsFunc(sourceFile.CommentDirectives, (d) bool { return d.Kind == ast.CommentDirectiveKindNoCheck } {
	for _, checker := range p.checkers {
		// ...
	}
}
```

(Probably that func is worth extracting)

jakebailey: (Or maybe SourceFile should have a NoCheck field).

tmm1: Thanks for quick feedback. Updated impl based on your idea.

tmm1:
> Unfortunately, this means this PR is trying to do this from the wrong angle.

Okay, I expected as much. Thank you.

PR #723, opened 2025-03-26
Port parameters expansion with rest tuple members

Andarist:


jakebailey:
Seemingly panicking because `getTupleElementLabelFromBindingElement` is being called on a `BindingElement` with a nil `name`, while in Strada this was not possible.

jakebailey:
`parseArrayBindingElement` differs significantly between the two compilers, for some reason.

internal/checker/checker.go
```diff
@@ -25833,6 +25833,73 @@ func (c *Checker) markPropertyAsReferenced(prop *ast.Symbol, nodeForCheckWriteOn
...
 }
 
+func (c *Checker) GetExpandedParameters(signature *Signature) []*ast.Symbol {
+	if signatureHasRestParameter(signature) {
+		restIndex := len(signature.parameters) - 1
+		restSymbol := signature.parameters[restIndex]
+		restType := c.getTypeOfSymbol(restSymbol)
+		if isTupleType(restType) {
+			return c.expandSignatureParametersWithTupleMembers(signature, restType.AsTypeReference(), restIndex, restSymbol)
+		}
```
jakebailey, line 25843: Need to leave a `!!!` for the missing `skipUnionExpanding` param?

internal/checker/checker.go
```diff
@@ -25833,6 +25833,73 @@ func (c *Checker) markPropertyAsReferenced(prop *ast.Symbol, nodeForCheckWriteOn
...
+	}
+	return signature.parameters
+}
+
+func (c *Checker) expandSignatureParametersWithTupleMembers(signature *Signature, restType *TypeReference, restIndex int, restSymbol *ast.Symbol) []*ast.Symbol {
+	elementTypes := c.getTypeArguments(restType.AsType())
+	associatedNames := c.getUniqAssociatedNamesFromTupleType(restType, restSymbol)
+	expanded := make([]*ast.Symbol, len(signature.parameters)-1+len(elementTypes))
+	copy(expanded, signature.parameters[:restIndex])
+	for i, t := range elementTypes {
```
jakebailey: Rather than using offsets to do this, I would suggest just appending the leading parameters onto a slice, and then use `append` at the bottom of the loop to stick things on. It's going to be a lot easier to understand than direct assigning at computed offsets.

Andarist: done

internal/checker/checker.go
```diff
@@ -25833,6 +25833,73 @@ func (c *Checker) markPropertyAsReferenced(prop *ast.Symbol, nodeForCheckWriteOn
 	c.symbolReferenceLinks.Get(target).referenceKinds |= ast.SymbolFlagsAll
 }
 
+func (c *Checker) GetExpandedParameters(signature *Signature /* !!! skipUnionExpanding */) []*ast.Symbol {
```
jakebailey, line 25836: Should this be returning `[][]*ast.Symbol` to match the old code?

Andarist: Not yet - unless u want me to implement `skipUnionExpanding` right away. I skipped it since it’s only required by services and there are no tests for them yet. I could ofc just do this preemptively but it felt weird to do so

jakebailey (approved):

internal/checker/checker.go
jakebailey: I guess it's fine; anyone going to add that param will realize they need to modify it.

PR #724, opened 2025-03-27
Use omitted expressions in array binding patterns instead of binding elements with nil name

Andarist:
I could see the lack of omitted expressions here to be intentional but I'm not 100% sure if it is (although the comment in `parseArrayBindingElement` suggests it was). As we can see, this fixes some baselines. They could likely be fixed by handling the other thing (binding element with nil name) in other places but it just shows that it's a tradeoff between one annoyance and another.


ahejlsberg:
The switch to a binding element with a `nil` name was intentional. In the old baselines we'd just print a bunch of useless `undefined` for omitted elements. I'd prefer to not see those, so if that's the only difference we should just move the `.diff` files to the `submoduleAccepted` bucket.

jakebailey:
It's easier to just filter these specific lines away in the diffs, but I will note that this comes out of #723 where this difference is causing a real crash in newly ported code.

jakebailey:
It's also really easy to modify the type printer to skip these too so they don't show up at all, if that was the whole problem. I think it's worse to make the port harder than anything, personally.

ahejlsberg:
I don't recall the specific details, but as I was porting it was simpler to just always just deal with `BindingElement` nodes instead of having to always first check for `OmittedExpression`.

jakebailey:
Then we can just keep it as nil and close this PR, we'll just have to double check for potentially nil names in places we handled binding elements previously (which is what #723 is currently doing).

jakebailey:
I have a three-line baseline writer change which can omit those lines from all of the outputs; I can apply it after #702.

Diff would be: https://github.com/microsoft/typescript-go/compare/jabaile/hide-omitted-expression-types

PR #725, opened 2025-03-27
Refactor while-style loop to standard reverse for loop

wjdqhry:
refactor to use standard for loop


wjdqhry:

jakebailey:
Thanks, but this code is directly ported from the old codebase and I don't think now is the time for a refactor like this when the current code works, unless there's a compelling compiler performance reason.

wjdqhry:
Of course I understand, then I'll close this PR.

PR #728, opened 2025-03-28
Fix resolution of type of `void` keyword

auvred:
https://github.com/microsoft/TypeScript/blob/29e6d6689dfb422e4f1395546c1917d07e1f664d/src/compiler/utilities.ts#L2738-L2739


jakebailey:
No test changes; what does this fix?

auvred:
Calling `getTypeOfNode` on `void` keyword returns intrinsic `error` type instead of intrinsic `void` type. 

```typescript
type Foo = void
//          ^?
```

jakebailey (approved):
It's certainly correct, I'm just not sure how you found or tested this.

PR #730, opened 2025-03-28
Support @libFiles in tests + fix

gabritto:


testdata/lib/lib.d.ts
jakebailey, line 1: Can we pull these out of the submodule instead of copying them?

gabritto: Sure. I just imagined long term we'll want to have that in `testdata/`.

jakebailey: I'm not sure, at least not until we have a reason to use them in the non-submodule tests.

_submodules/TypeScript
jakebailey, line 1: Something has gone wrong here in your checkout during merge.

jakebailey (approved):
Nice!

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -196,6 +203,20 @@ func CompileFilesEx(
 	return result
 }
 
+// Creates a test file as specified by "@libFiles".
+func createLibFile(libFile string) *TestFile {
+	libPath := filepath.Join(repo.TypeScriptSubmodulePath, "tests", "lib")
```
jakebailey, line 208: Actually, do we need to cache this read so we don't keep doing it?

PR #731, opened 2025-03-29
Unify the way anonymous class symbols are printed in error messages

Andarist:
This is slightly different from Strada in two ways:
1. `GetNameOfDeclaration` is preferred now
2. `(anonymous)` becomes `(Anonymous class)` - which is just more consistent

### DIFF
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -10,7 +10,6 @@ import (
 	"github.com/microsoft/typescript-go/internal/compiler/diagnostics"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/jsnum"
-	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
 type SignatureCheckMode uint32
@@ -4272,8 +4271,8 @@ func (r *Relater) reportUnmatchedProperty(source *Type, target *Type, unmatchedP
 		privateIdentifierDescription := unmatchedProperty.ValueDeclaration.Name().Text()
 		symbolTableKey := binder.GetSymbolNameForPrivateIdentifier(source.symbol, privateIdentifierDescription)
 		if r.c.getPropertyOfType(source, symbolTableKey) != nil {
-			sourceName := scanner.DeclarationNameToString(ast.GetNameOfDeclaration(source.symbol.ValueDeclaration))
-			targetName := scanner.DeclarationNameToString(ast.GetNameOfDeclaration(target.symbol.ValueDeclaration))
+			sourceName := classDeclarationNameToString(source.symbol.ValueDeclaration)
+			targetName := classDeclarationNameToString(target.symbol.ValueDeclaration)
 			r.reportError(diagnostics.Property_0_in_type_1_refers_to_a_different_member_that_cannot_be_accessed_from_within_type_2, privateIdentifierDescription, sourceName, targetName)
 			return
 		}
--- a/testdata/baselines/reference/compiler/privateNamesUnique-6.errors.txt
+++ b/testdata/baselines/reference/compiler/privateNamesUnique-6.errors.txt
@@ -1,0 +1,17 @@
+privateNamesUnique-6.ts(6,7): error TS2322: Type '(Anonymous class)' is not assignable to type 'A'.
+  Property '#foo' in type '(Anonymous class)' refers to a different member that cannot be accessed from within type 'A'.
+
+
+==== privateNamesUnique-6.ts (1 errors) ====
+    class A {
+      #foo: number;
+    }
+    
+    // error
+    const test: A = new (class {
+          ~~~~
+!!! error TS2322: Type '(Anonymous class)' is not assignable to type 'A'.
+!!! error TS2322:   Property '#foo' in type '(Anonymous class)' refers to a different member that cannot be accessed from within type 'A'.
+      #foo: number;
+    })();
+    
--- a/testdata/baselines/reference/submodule/conformance/privateNameMethodClassExpression.errors.txt
+++ b/testdata/baselines/reference/submodule/conformance/privateNameMethodClassExpression.errors.txt
@@ -1,5 +1,5 @@
-privateNameMethodClassExpression.ts(9,17): error TS18013: Property '#method' is not accessible outside class '(Missing)' because it has a private identifier.
-privateNameMethodClassExpression.ts(10,17): error TS18013: Property '#field' is not accessible outside class '(Missing)' because it has a private identifier.
+privateNameMethodClassExpression.ts(9,17): error TS18013: Property '#method' is not accessible outside class 'C' because it has a private identifier.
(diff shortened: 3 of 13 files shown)


testdata/baselines/reference/submodule/conformance/privateNameMethodClassExpression.errors.txt.diff
```diff
@@ -13,10 +13,10 @@
      C.getInstance().#method; // Error
                      ~~~~~~~
 -!!! error TS18013: Property '#method' is not accessible outside class '(anonymous)' because it has a private identifier.
-+!!! error TS18013: Property '#method' is not accessible outside class '(Missing)' because it has a private identifier.
++!!! error TS18013: Property '#method' is not accessible outside class 'C' because it has a private identifier.
      C.getInstance().#field; // Error
                      ~~~~~~
 -!!! error TS18013: Property '#field' is not accessible outside class '(anonymous)' because it has a private identifier.
-+!!! error TS18013: Property '#field' is not accessible outside class '(Missing)' because it has a private identifier.
++!!! error TS18013: Property '#field' is not accessible outside class 'C' because it has a private identifier.
```
Andarist, line 20: I think it would be good to add this to accepted baselines but I'm not sure how to do it.

testdata/baselines/reference/submodule/conformance/privateNameMethodClassExpression.errors.txt
```diff
@@ -13,9 +13,9 @@ privateNameMethodClassExpression.ts(10,17): error TS18013: Property '#field' is
     console.log(C.getInstance().getField());
     C.getInstance().#method; // Error
                     ~~~~~~~
-!!! error TS18013: Property '#method' is not accessible outside class '(Missing)' because it has a private identifier.
+!!! error TS18013: Property '#method' is not accessible outside class 'C' because it has a private identifier.
     C.getInstance().#field; // Error
                     ~~~~~~
-!!! error TS18013: Property '#field' is not accessible outside class '(Missing)' because it has a private identifier.
+!!! error TS18013: Property '#field' is not accessible outside class 'C' because it has a private identifier.
```
Andarist, line 19: Strada would print `(anonymous)` here but this has an inferrable~ name from the declaration and I think it's better to use it. Using it is not a new thing to do.

testdata/baselines/reference/compiler/privateNamesUnique-6.errors.txt
```diff
@@ -0,0 +1,17 @@
...
+==== privateNamesUnique-6.ts (1 errors) ====
+    class A {
+      #foo: number;
+    }
+    
+    // error
+    const test: A = new (class {
+          ~~~~
+!!! error TS2322: Type '(Anonymous class)' is not assignable to type 'A'.
+!!! error TS2322:   Property '#foo' in type '(Anonymous class)' refers to a different member that cannot be accessed from within type 'A'.
```
Andarist: Strada would print `(Anonymous class)` in the first line here and then `(anonymous)` in the second line. So I think this change unifying it is a good thing. 

Note this is a new test case, not one from Strada.

internal/checker/relater.go
```diff
@@ -4272,8 +4271,8 @@ func (r *Relater) reportUnmatchedProperty(source *Type, target *Type, unmatchedP
 		privateIdentifierDescription := unmatchedProperty.ValueDeclaration.Name().Text()
 		symbolTableKey := binder.GetSymbolNameForPrivateIdentifier(source.symbol, privateIdentifierDescription)
 		if r.c.getPropertyOfType(source, symbolTableKey) != nil {
-			sourceName := scanner.DeclarationNameToString(ast.GetNameOfDeclaration(source.symbol.ValueDeclaration))
```
Andarist, line 4272: the previous behavior here was already diverging slightly from Strada for those anonymous classes (IMHO, in a good way)

testdata/baselines/reference/submodule/conformance/privateNameMethodClassExpression.errors.txt.diff
jakebailey: Add this line to submoduleAccepted.txt:

```
privateNameMethodClassExpression.errors.txt.diff
```

Andarist: done

jakebailey:
I'm not 100% clear why `(Anonymous class)` is more consistent. The `(Missing)` change seems correct because it improves a baseline, but otherwise I'm not sure.

Andarist:
This isn't particularly consistent: [TS playground](https://www.typescriptlang.org/play/?target=99&noEmit=true&ts=5.9.0-dev.20250331#code/PTAEAEBcEMCcHMCmkBcpEGcB2iAekAoAYwBtoMNQBBUAbwNFAGIAzAezbSwFcBbAI0SxQAXlABGANwEAvsTZYMkUJEypqo0DgDuoABSlyleo1YcufQcLEAmaTICUeh9ILzFygMKbDFOg2YWAEtEEgATTUgACyCMADomXmQotjDnaVMk6NTnf0ZGWGRuWCxQABY7ALlGJWhIIKJQJEgASQ9oLCJEXJN8wshi0p1QT3SqgOaAMRDwnoCCopKVGPjWGbCM0DkZV0845rbazu6HBKyUjbdhgzI-XsTknIc8vsXSis3tp1OH7I2gA). I am surprised now that the first error mentions `(Missing)` - I could swear I've seen `(anonymous)` there over the weekend :o But even considering that, this isn't really consistent

`(Anonymous class)` seems to be way more common in Strada. In its test suite, we can find that 700+ times whereas `(anonymous)` appears only 4 times (and all of those are within a single test case). `(Missing)` appears 100+ times but none of those refer to a missing class name.

jakebailey:
Old PR, but it would be worth updating this PR I think, at least into a form that reduces baseline diffs. I'd rather that be the outcome.

Andarist:
@jakebailey done

jakebailey:
Er, well sort of, but it's not reducing diffs, it's still changing the behavior.

Andarist:
It changes one diff and adds it to the accepted changes. I could port this **faithfully** (to print `(anonymous)` instead of `(Missing)`) but then that isn't exactly consistent with some of the other presented cases (IIRC). So I chose to just unify this here while addressing this baseline diff

jakebailey (approved):
On second thought, I feel like this is a reasonable improvement, however it would probably make sense to stick this in Strada too, given the improvement should apply there?

Andarist:
Sure, I can port this to Strada

Andarist:
I created a PR porting this to Strada: https://github.com/microsoft/TypeScript/pull/61943 so I have reverted new tests etc from here given they will come to Corsa automatically once you bump the submodule after that port lands

jakebailey:
I'd leave the tests here; it's fine for tests to be "duplicated" between the two temporarily. The names will not conflict.

PR #671, opened 2025-03-19
Fixed: [bug/lsp]panic: vfs: path  is not absolute #670

rezwanahmedsami:
Added a validation step to filter out empty strings when processing file paths. For example, in the `getFileNamesFromConfigSpecs` function, validated like this:

   ```go
   for _, fileName := range validatedFilesSpec {
       if fileName == "" {
           continue // Skip empty strings
       }
       file := tspath.GetNormalizedAbsolutePath(fileName, basePath)
       literalFileMap.Set(keyMapper(fileName), file)
   }
   ```

PR #738, opened 2025-04-01
Fixed: [bug/lsp]panic: vfs: path is not absolute #671

gabrielluizsf:
@jakebailey 
@gabritto
@andrewbranch

[Old PR](https://github.com/microsoft/typescript-go/pull/671)

### DIFF
--- a/internal/tsoptions/tsconfigparsing.go
+++ b/internal/tsoptions/tsconfigparsing.go
@@ -1423,7 +1423,6 @@ func removeWildcardFilesWithLowerPriorityExtension(file string, wildcardFiles co
 // options is the Compiler options.
 // host is the host used to resolve files and directories.
 // extraFileExtensions optionally file extra file extension information from host
-
 func getFileNamesFromConfigSpecs(
 	configFileSpecs configFileSpecs,
 	basePath string, // considering this is the current directory
--- a/internal/tsoptions/tsconfigparsing.go
+++ b/internal/tsoptions/tsconfigparsing.go
@@ -2,6 +2,8 @@ package tsoptions
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"slices"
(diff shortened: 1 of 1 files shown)


i9si (approved):

internal/tsoptions/tsconfigparsing.go
```diff
@@ -1432,32 +1434,41 @@ func getFileNamesFromConfigSpecs(
 	extraFileExtensions []fileExtensionInfo,
 ) []string {
 	extraFileExtensions = []fileExtensionInfo{}
+
+	if basePath == "" {
+		basePath, _ = os.Getwd()
```
jakebailey: This is definitely incorrect, sorry. This package should have no knowledge about the OS package, nor should any package use the filepath package.

You should be rewriting this using the provided FS and the tspath package.

gabrielluizsf: ok

jakebailey:
This really needs some sort of test. Nobody's said how this problem surfaces.

internal/tsoptions/tsconfigparsing.go
waelantar: @gabrielluizsf  this may help
```
if basePath == "" {
    basePath = host.GetCurrentDirectory() // Use FS abstraction
} else {
    basePath = tspath.NormalizePath(
        tspath.CombinePaths(host.GetCurrentDirectory(), basePath)
    )
}

basePath = tspath.NormalizePath(basePath)
keyMapper := func(value string) string { 
    return tspath.GetCanonicalFileName(value, host.UseCaseSensitiveFileNames()) 
}

for _, fileName := range validatedFilesSpec {
    file := tspath.GetNormalizedAbsolutePath(fileName, basePath)
    
    // Validate using tspath's internal absolute check
    if !tspath.IsAbsolute(file) {
        host.ReportError(fmt.Errorf("non-absolute path %q detected", file))
        continue
    }
    
    literalFileMap.Set(keyMapper(fileName), file)
}

// File collection without OS path manipulation
files := make([]string, 0)
for file := range literalFileMap.Values() {
    files = append(files, file) // Keep all entries
}
for file := range wildcardFileMap.Values() {
    files = append(files, file)
}
for file := range wildCardJsonFileMap.Values() {
    files = append(files, file)
}

return files
```

gabrielluizsf: > if basePath == "" {
>     basePath = host.GetCurrentDirectory() // Use FS abstraction
> } else {
>     basePath = tspath.NormalizePath(
>         tspath.CombinePaths(host.GetCurrentDirectory(), basePath)
>     )
> }

GetCurrentDirectory does not exist, I will try to understand how this project works to see if I can implement it

jakebailey: It does exist on the host.

jakebailey:
Most of the changes in this PR are unrelated to the actual fix, which I believe is just `basePath` being modified. Can you please undo everything that is unrelated?

I will still say like last time that I don't think this is the right fix. The original `getFileNamesFromConfigSpecs` code does not do this; it only normalizes the path. If there's a bad path being passed into this function, the problem must be somewhere else.

gabrielluizsf:
> Most of the changes in this PR are unrelated to the actual fix, which I believe is just `basePath` being modified. Can you please undo everything that is unrelated?
> 
> I will still say like last time that I don't think this is the right fix. The original `getFileNamesFromConfigSpecs` code does not do this; it only normalizes the path. If there's a bad path being passed into this function, the problem must be somewhere else.



> Most of the changes in this PR are unrelated to the actual fix, which I believe is just `basePath` being modified. Can you please undo everything that is unrelated?
> 
> I will still say like last time that I don't think this is the right fix. The original `getFileNamesFromConfigSpecs` code does not do this; it only normalizes the path. If there's a bad path being passed into this function, the problem must be somewhere else.

Thank you for your feedback. I understand your concerns regarding the scope of the changes. I'll revert the unrelated modifications to keep the PR focused on the actual fix.

Regarding the approach, I see your point about getFileNamesFromConfigSpecs only normalizing the path. I'll take a closer look to identify where the incorrect path is being introduced and determine if the issue originates elsewhere. Let me know if you have any additional insights.

PR #740, opened 2025-04-01
wire up @ts-nocheck/ts-check pragmas

tmm1:
follow up to #719

cc @jakebailey


internal/parser/parser.go
```diff
@@ -6495,12 +6495,25 @@ func extractPragmas(commentRange ast.CommentRange, text string) []ast.Pragma {
 		}
 		return []ast.Pragma{pragma}
 	}
-	return nil
 
 	// const singleLine = range.kind === SyntaxKind.SingleLineCommentTrivia && singleLinePragmaRegEx.exec(text);
```
jakebailey: Seems like the comments here can be deleted now that they are implemented?

jakebailey:
@ahejlsberg I think you're working on pragma parsing, not sure what you're modifying at this point, though. Maybe it'd be fine to take this and you just delete the ported code in favor of your regex-free version?

internal/parser/parser.go
```diff
@@ -6556,6 +6571,18 @@ func processPragmasIntoFields(context *ast.SourceFile /* !!! reportDiagnostic fu
 			} else {
 				// reportDiagnostic(argMap.Pos, argMap.End-argMap.Pos, "Invalid reference directive syntax")
 			}
+		case "ts-nocheck":
+			fallthrough
+		case "ts-check":
```
```suggestion
		case "ts-check", "ts-nocheck":
```

internal/parser/parser.go
```diff
@@ -6495,12 +6495,25 @@ func extractPragmas(commentRange ast.CommentRange, text string) []ast.Pragma {
...
+		return []ast.Pragma{
+			{
+				Name:      matches[1],
+				Args:      make(map[string]ast.PragmaArgument),
+				ArgsRange: commentRange,
+			},
+		}
+	}
 
 	// if (range.kind === SyntaxKind.MultiLineCommentTrivia) {
```
jakebailey: This comment too.

jakebailey:
This seems correct to me, just, bad timing with Anders porting more pragma code for JSX.

tmm1:
If this is in-progress I'm happy to wait. Support for `@ts-nocheck` is the only thing preventing me from using `tsgo` atm.

jakebailey:
> If this is in-progress I'm happy to wait. Support for `@ts-nocheck` is the only thing preventing me from using `tsgo` atm.

Er, well, we don't have JSDoc support, so JS support is going to be wonky at best.

tmm1:
Even though tests are passing locally, when I try using the built tsgo on my project.. its still reporting errors for files with `// @ts-nocheck` at the top. Not immediately clear to me why.

tmm1:
Alright I have this working to my satisfaction. If there are other PRs I need to resolve merge conflicts against, lmk.

internal/parser/parser.go
jakebailey: Or, is this not done? Maybe it's only ported for one pragma?

internal/compiler/program.go
```diff
@@ -321,6 +321,9 @@ func (p *Program) getSemanticDiagnosticsForFile(sourceFile *ast.SourceFile) []*a
 	}
 
 	diags := slices.Clip(sourceFile.BindDiagnostics())
+	if sourceFile.CheckJsDirective != nil && !sourceFile.CheckJsDirective.Enabled {
+		return diags
+	}
```
jakebailey: Is this the right place for this? I thought the checker itself checked this option?

internal/parser/parser.go
tmm1: Correct, did not implement multi line

internal/compiler/program.go
tmm1: That's what I thought but it wasn't working for some cases without this.

tmm1: Some of the test failures are due to additional comment parsing at the end of this function (i.e for `@ts-expect-error`), which happens outside the checkers.

jakebailey: The thing I'm considering is just that we should be looking to see where `SourceFile.checkJsDirective` was used in the old codebase, then ensure we mirror that here.

tmm1: That is what I did for the rest of the PR. But I still needed this for tests to pass. Not sure why.

tmm1: Okay I dug in some more here and found the equivalent typescript-side check:

https://github.com/microsoft/TypeScript/blob/0693cc72e6ddb6c7468a14e05888efa2f43ac7b0/src/compiler/program.ts#L3058-L3066

```ts
    function getBindAndCheckDiagnosticsForFileNoCache(
    ): readonly Diagnostic[] {
            if (skipTypeChecking(sourceFile, options, program)) {
                return emptyArray;
            }
```

In the typescript impl, `skipTypeChecking` is reused between program and checker. But on the golang side, `skipTypeChecking` only exists in the `internal/checker` package. I can try moving it into `internal/core`.. might make more sense as `(*core.CompilerOptions).SkipTypeChecking(sourceFile)` ?

jakebailey: I think I just fixed this in #763. Anders also just redid the pragma parsing for JSX reasons, so I think your PR will need an update but end up simpler as a result.

tmm1: Oh, fantastic. I merged main so this PR should be ready.

jakebailey (approved):
Looks great, thank you!

PR #742, opened 2025-04-03
Update lsp generation for more accurate documentation

dyl10s:
I was poking around the lsp server and found that you guys generate a good bit of the types using the meta model. I noticed that when I re-ran the generation, I got different results than the currently committed file, and some of the documentation had weird indenting/spacing. 

My efforts here were to normalize it and make it map more closely to the `metaModel.json`. I then ran the generation script so that it updated the `lsp_generated.go`


dyl10s:

jakebailey:
> I noticed that when I re-ran the generation, I got different results than the currently committed file

This shouldn't be possible; the script is deterministic. Did you run the scripts in the dir to grab the pinned copy? Are you missing a format step maybe? Note how CI fails.

I'm not sure I like the code added for this, honestly. It's a lot of code to duplicate for all of the places we write things.

dyl10s:
> This shouldn't be possible; the script is deterministic. Did you run the scripts in the dir to grab the pinned copy? Are you missing a format step maybe? Note how CI fails.

I did run the `fetchModel` script in the directory to grab the pinned copy using node. 

The primary fix for this PR was to resolve the issue where the since documentation would be formatted correctly. I ran the initial script with no changes and got a problem where additional comment lines were added in places they shouldn't be (see image). This is why there was a slightly larger refactor.
image

Let me know if there is a specific way I should be running the `fetchModel` or `generate` script. I could be missing something here.

internal/lsp/lsproto/_generate/generate.mjs
```diff
@@ -60,19 +60,34 @@ function writeLine(s) {
 
 /**
  * @param {string | undefined} doc
+ * @param {string | undefined} proposed,
+ * @param {string | undefined} since,
+ * @param {string | undefined} sinceTags,
+ * @param {string | undefined} deprecated
  */
-function writeDocumentation(doc) {
+function writeDocumentation(doc, proposed, since, sinceTags, deprecated) {
```
jakebailey: This should really just accept a single parameter of a shape that includes all of these props, looking at each of the calls.

dyl10s: I swapped this to an inline object type. I looked in `metaModelSchema.mts` to see if there was a base type that each of these objects extended but didn't find one. Let me know if you have a better idea here.

jakebailey:
Merging #789 and rerunning CI here shows that the formatting is in fact being changed in an unexpected way.

PR #743, opened 2025-04-03
🛠️  Refactors binder declareSymbolEx to remove sonar cognitive complexity warning

Imran109:
The function, `declareSymbolEx` had cognitive complexity: 100

Only broken down this function into small chunks for better readability. Used extractions and guard clauses. Did not make any changes to the actual implementation's logic. I've even kept all the existing comments in place and made sure all test cases are passing and has 0 complexities now.

This is my first contribution here, so let me know if I am missing anything with respect to how PRs are done for this repo

### DIFF
--- a/internal/binder/binder.go
+++ b/internal/binder/binder.go
@@ -147,156 +147,267 @@ func (b *Binder) declareSymbol(symbolTable ast.SymbolTable, parent *ast.Symbol, 
 	return b.declareSymbolEx(symbolTable, parent, node, includes, excludes, false /*isReplaceableByMethod*/, false /*isComputedName*/)
 }
 
-func (b *Binder) declareSymbolEx(symbolTable ast.SymbolTable, parent *ast.Symbol, node *ast.Node, includes ast.SymbolFlags, excludes ast.SymbolFlags, isReplaceableByMethod bool, isComputedName bool) *ast.Symbol {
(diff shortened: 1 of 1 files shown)


internal/binder/binder.go
```diff
@@ -147,156 +147,267 @@ func (b *Binder) declareSymbol(symbolTable ast.SymbolTable, parent *ast.Symbol,
 	return b.declareSymbolEx(symbolTable, parent, node, includes, excludes, false /*isReplaceableByMethod*/, false /*isComputedName*/)
 }
 
-func (b *Binder) declareSymbolEx(symbolTable ast.SymbolTable, parent *ast.Symbol, node *ast.Node, includes ast.SymbolFlags, excludes ast.SymbolFlags, isReplaceableByMethod bool, isComputedName bool) *ast.Symbol {
+func (b *Binder) declareSymbolEx(symbolTable ast.SymbolTable, parent *ast.Symbol, node *ast.Node,
```
Imran109, line 150: Only broken down this function into small chunks for better readability. Used extractions and guard clauses. Did not make any changes to the actual implementation. I've even kept all the existing comments in place.

Imran109:
> @microsoft-github-policy-service agree

jakebailey:
Thanks for the contribution; this codebase is a port of http://github.com/microsoft/TypeScript and so intentionally shares the same code structure. I don't think now is the right time to make structural changes like this as it only make it more challenging to double check what we're doing and find differences.

Maybe at a later date, we could do something like this, but not right now.

PR #746, opened 2025-04-03
Use env vars for herebyfile options in CI

jakebailey:
This is a lot easier to manage.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -21,24 +21,48 @@ const isCI = !!process.env.CI;
 const $pipe = _$({ verbose: "short" });
 const $ = _$({ verbose: "short", stdio: "inherit" });
 
+/**
+ * @param {string} name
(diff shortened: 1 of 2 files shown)


andrewbranch (approved):

Herebyfile.mjs
```diff
@@ -21,24 +21,48 @@ const isCI = !!process.env.CI;
...
+        concurrentTestPrograms: { type: "boolean", default: parseEnvBoolean("CONCURRENT_TEST_PROGRAMS") },
+        coverage: { type: "boolean", default: parseEnvBoolean("COVERAGE") },
     },
     strict: false,
     allowPositionals: true,
     allowNegative: true,
-    noembed: false,
-    debug: false,
-    concurrentTestPrograms: false,
-    coverage: false,
```
jakebailey, line 41: This was wrong but the types don't indicate so, very cool

PR #748, opened 2025-04-03
Add basic CommonJS support

sandersn:
This PR adds basic CommonJS imports and exports. It's aiming to cover 80% of usage--a subset of TS semantics in most cases.

CommonJS imports are implemented basically the same way as in Strada. The binder and checker treat a variable declaration that matches a syntactic pattern as if it were the same as `import x = require('x')` in Typescript.
- `const x = require('x')` is supported.
- `const { a, b, c } = require('x')` is also supported (this is different than TS, but works the same as in Strada).
- `const a = require('x').a` is **no longer** supported.

CommonJS export assignments are implemented as a synthetic variant of Typescript export assignments. When the parser sees `module.exports = ...`, it emits a synthetic JS `export = ...`.
- `module.exports = x` is supported, with the same semantics as `export = x` in TS. (This means that literal types don't widen because `export=` isn't a mutable location.)

CommonJS exports are implemented as a separate synthetic declaration. I don't create synthetic variable declarations because `export var a = 1` both exports `a` and declares a local `a`. `module.exports.a = 1` only does the former. More on that in Open Questions.
- `module.exports.a = 1` exports `a`, and allows references to `module.exports.a` in that file.

Other changes:
- I deleted some code that is no longer needed because of (1) simpler `const x = require` support (2) synthetic export support. But I haven't completely cleaned house, especially in the binder.
- I added JS support to the incomplete module resolution in fileloader/resolveImportsAndModuleAugmentations, which makes test results a lot better, even though it's still not how the finished port will work.
- I moved the code to `reparser.go`, and will move the JSDoc tag code there later.

## Open questions
- Should we support local declaration of `module.exports.x=`? It is possible to hack the checker and name resolution to make it possible to refer to `module.exports.x` inside the CommonJS file, but it won't be pretty. I'm leaning toward not supporting this, at least until I can run top400 JS on Corsa after this PR. Based on the top100 JS corpus I looked at earlier, it's rare to reference an exported object within the exporting file.

- A secondary question is: should `module`, `exports` and `require` error without a declaration, suggesting that you install `@types/node`? Strada never does this for JS, but in Corsa I think it makes sense to ask people with `checkJs` on to install `@types/node` if they're using CommonJS. This does introduce lots of errors into the baselines, but remember that plain JS files report only syntax errors.

- How much of the ES/CJS interop failures in modulePreserve2 and 4, and importNonExportedMember12 are due to mistakes/omissions in this PR and how much are due to unfinished module exports code? I need somebody who understands ES/CJS interop well to look at these baselines.

- Naming: The synthetic node for `module.exports=` is `KindJSExportAssignment`, based on `export=`'s `KindExportAssignment`. For `module.exports.x=`, I chose `KindCommonJSExport`, but maybe it should be `KindCommonJSExportAssignment` or `KindCommonJSExportVariableDeclaration`. Opinions?

- Should baselines skip all reparsed nodes? Just non-duplicate reparsed nodes? Skipping all of them makes the Strada baseline diffs smaller but omits, especially for `@typedef`, which was not baselined before.

## Future work
- Move other reparsing code into reparser.go
- `module.exports['illegal name']=` -- requires generating a temp name to print it in a .d.ts file.
- The combination
```js
/** @typedef {number} ImplicitlyExported */
module.exports = { a, b, c }
```
Requires exporting `ImplicitlyExported` with the rest of the exports `a,b,c`. The most straightforward way is to synthesise a temp name for `{a,b,c}` and a namespace with the same name that merges with it. The synthetic namespace contains the implicitly exported type.


internal/printer/printer.go
```diff
@@ -3689,6 +3725,24 @@ func (p *Printer) emitExportAssignment(node *ast.ExportAssignment) {
...
 }
 
+// export declare var <name> = <initialiser>;
+func (p *Printer) emitCommonJSExport(node *ast.CommonJSExport) {
+	state := p.enterNode(node.AsNode())
+	p.emitTokenWithComment(ast.KindExportKeyword, node.Pos(), WriteKindKeyword, node.AsNode())
+	p.writeSpace()
+	p.writeKeyword("var")
+	p.writeSpace()
+	if node.Name().Kind == ast.KindStringLiteral {
```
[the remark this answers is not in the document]
sandersn: @rbuckton I'm not sure if I need a printer for commonjs exports, or any other reparsed node. If so, I'll open a bug to remind myself to emit a temp rename, or some other correct way to emit exports with arbitrary names.

internal/ast/utilities.go
```diff
@@ -1392,22 +1393,6 @@ func IsFunctionPropertyAssignment(node *Node) bool {
 	return false
 }
```
sandersn, line 1416: we have a private and public duplicate of this function so I deleted the private one

internal/ast/utilities.go
```diff
@@ -1473,10 +1458,6 @@ func IsEffectiveExternalModule(node *SourceFile, compilerOptions *core.CompilerO
 	return IsExternalModule(node) || (isCommonJSContainingModuleKind(compilerOptions.GetEmitModuleKind()) && node.CommonJSModuleIndicator != nil)
 }
```
sandersn, line 1481: this is unused

internal/ast/utilities.go
```diff
@@ -1646,20 +1627,7 @@ func IsEnumConst(node *Node) bool {
...
 
 func ExportAssignmentIsAlias(node *Node) bool {
```
sandersn, line 1651: I believe this was for syntactic `module.exports=` support and will never be used.

internal/binder/binder.go
```diff
@@ -378,12 +380,16 @@ func GetSymbolNameForPrivateIdentifier(containingClassSymbol *ast.Symbol, descri
 }
 
 func (b *Binder) declareModuleMember(node *ast.Node, symbolFlags ast.SymbolFlags, symbolExcludes ast.SymbolFlags) *ast.Symbol {
+	container := b.container
```
sandersn, line 380: commonjs exports are file exports, even if they're nested inside something else (though Corsa, not Strada, gives a diagnostic too)

internal/binder/binder.go
```diff
@@ -400,21 +406,21 @@ func (b *Binder) declareModuleMember(node *ast.Node, symbolFlags ast.SymbolFlags
 	//       during global merging in the checker. Why? The only case when ambient module is permitted inside another module is module augmentation
 	//       and this case is specially handled. Module augmentations should only be merged with original module definition
 	//       and should never be merged directly with other augmentation, and the latter case would be possible if automatic merge is allowed.
-	if !ast.IsAmbientModule(node) && (hasExportModifier || b.container.Flags&ast.NodeFlagsExportContext != 0) {
-		if !ast.IsLocalsContainer(b.container) || (ast.HasSyntacticModifier(node, ast.ModifierFlagsDefault) && b.getDeclarationName(node) == ast.InternalSymbolNameMissing) {
-			return b.declareSymbol(ast.GetExports(b.container.Symbol()), b.container.Symbol(), node, symbolFlags, symbolExcludes)
+	if !ast.IsAmbientModule(node) && (hasExportModifier || container.Flags&ast.NodeFlagsExportContext != 0) {
```
sandersn, line 406: commonjs exports need to bind `module.exports.x` locally not just `x`, so I skipped it for now. See Open Questions

internal/binder/binder.go
```diff
@@ -922,6 +936,19 @@ func (b *Binder) bindFunctionExpression(node *ast.Node) {
 	b.bindAnonymousDeclaration(node, ast.SymbolFlagsFunction, bindingName)
 }
```
sandersn, line 935: this is a port, with some advanced module code not done

internal/binder/nameresolver.go
```diff
@@ -12,6 +12,7 @@ type NameResolver struct {
 	Error                            func(location *ast.Node, message *diagnostics.Message, args ...any) *ast.Diagnostic
 	Globals                          ast.SymbolTable
 	ArgumentsSymbol                  *ast.Symbol
+	RequireSymbol                    *ast.Symbol
```
sandersn, line 15: this allows error-free use of `require` even without `@types/node` installed. This is here as part of the port of `const x = require` code from Strada. I'm torn about including it; `module.exports=` *does* error if you don't have node types installed, and I think that's better. Thoughts? I'll make a note to remove it for now.

internal/checker/checker.go
```diff
@@ -13547,15 +13553,6 @@ func (c *Checker) resolveSymbolEx(symbol *ast.Symbol, dontResolveAlias bool) *as
 
 func (c *Checker) getTargetOfImportEqualsDeclaration(node *ast.Node, dontResolveAlias bool) *ast.Symbol {
 	// Node is ImportEqualsDeclaration | VariableDeclaration
```
sandersn: `const a = require('x').a` is no longer supported

internal/checker/checker.go
```diff
@@ -16980,16 +17023,13 @@ func (c *Checker) getTypeOfAlias(symbol *ast.Symbol) *Type {
 		}
 		targetSymbol := c.resolveAlias(symbol)
 		exportSymbol := c.getTargetOfAliasDeclaration(c.getDeclarationOfAliasSymbol(symbol), true /*dontRecursivelyResolve*/)
-		declaredType := c.getExportAssignmentType(exportSymbol)
 		// It only makes sense to get the type of a value symbol. If the result of resolving
 		// the alias is not a value, then it has no type. To get the type associated with a
 		// type symbol, call getDeclaredTypeOfSymbol.
 		// This check is important because without it, a call to getTypeOfSymbol could end
 		// up recursively calling getTypeOfAlias, causing a stack overflow.
 		if links.resolvedType == nil {
```
sandersn, line 17031: this was only needed to support old syntactic `module.exports=`

jakebailey:
> `const a = require('x').a` is no longer supported.

This surprises me; how are we choosing to drop this?

> A secondary question is: should module, exports and require error without a declaration, suggesting that you install `@types/node`?

I would not do this. It is not uncommon to depend on these working in CJS code without the node types.

> How much of the ES/CJS interop failures in modulePreserve2 and 4, and importNonExportedMember12 are due to mistakes/omissions in this PR and how much are due to unfinished module exports code? I need somebody who understands ES/CJS interop well to look at these baselines.

I haven't looked either, but we're missing most of the code that makes `externalModuleIndicator` work, so a lot of this doesn't behave as expected.

> Should baselines skip all reparsed nodes? Just non-duplicate reparsed nodes? Skipping all of them makes the Strada baseline diffs smaller

I would definitely be eliminating diffs.

internal/parser/reparser.go
```diff
@@ -0,0 +1,88 @@
...
+
+const (
+	jsDeclarationKindNone jsDeclarationKind = iota
+	/// module.exports = expr
+	jsDeclarationKindModuleExports
+	/// exports.name = expr
+	/// module.exports.name = expr
+	jsDeclarationKindExportsProperty
+	/// className.prototype.name = expr
+	jsDeclarationKindPrototypeProperty
```
sandersn, line 18: I included the last 3 entries for future use, although I'm not 100% sure that they will be needed, or stay private to this module.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsTypeDefNoUnusedLocalMarked.errors.txt.diff
```diff
@@ -0,0 +1,26 @@
...
++    
++    export = Foo;
++==== something.js (1 errors) ====
++    /** @typedef {typeof import("./file")} Foo */
++    
++    /** @typedef {(foo: Foo) => string} FooFun */
++    
++    module.exports = /** @type {FooFun} */(void 0);
++    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
++!!! error TS2309: An export assignment cannot be used in a module with other exported elements.
```
jakebailey, line 26: This seems like a bug?

testdata/baselines/reference/submoduleAccepted/compiler/commonJsExportTypeDeclarationError.errors.txt.diff
```diff
@@ -21,21 +18,17 @@
      }
      
 -==== ./types1.ts (2 errors) ====
-+==== types1.ts (2 errors) ====
++==== types1.ts (1 errors) ====
      import test from "./test";
-+                     ~~~~~~~~
-+!!! error TS2306: File 'test.js' is not a module.
      export type test
 -
```
jakebailey, line 21: Do we have a parser bug?

testdata/baselines/reference/submoduleAccepted/compiler/localRequireFunction.errors.txt.diff
```diff
@@ -0,0 +1,20 @@
...
+-<no content>
+@@= skipped --1, +1 lines =@@
++app.js(1,10): error TS2441: Duplicate identifier 'require'. Compiler reserves name 'require' in top level scope of a module.
++app.js(5,20): error TS2307: Cannot find module 'fs' or its corresponding type declarations.
++
++
++==== app.js (2 errors) ====
++    function require(a) {
++             ~~~~~~~
++!!! error TS2441: Duplicate identifier 'require'. Compiler reserves name 'require' in top level scope of a module.
```
jakebailey, line 13: Hm, why is this happening? Maybe a module mode related problem?

testdata/baselines/reference/submoduleAccepted/compiler/moduleResolutionWithExtensions_notSupported2.types.diff
```diff
@@ -0,0 +1,8 @@
+--- old.moduleResolutionWithExtensions_notSupported2.types
++++ new.moduleResolutionWithExtensions_notSupported2.types
+@@= skipped -3, +3 lines =@@
+ import jsx from "./jsx";
+ >jsx : any
+ 
++=== /jsx.jsx ===
++
```
jakebailey, line 8: Odd baseline, maybe a bug in the printer?

sandersn (commented):
Some commentary from the last pass I made over the diff.

sandersn:
Some replies: 

> > const a = require('x').a is no longer supported.

> This surprises me; how are we choosing to drop this?

The only examples I found in top200 JS were webpack test cases.

> > A secondary question is: should module, exports and require error without a declaration, suggesting that you install @types/node?

> I would not do this. It is not uncommon to depend on these working in CJS code without the node types.

Even if you have `checkJs: true`? But you're right, it is pretty annoying just in the baseline diffs.

> > Should baselines skip all reparsed nodes? Just non-duplicate reparsed nodes? Skipping all of them makes the Strada baseline diffs smaller

> I would definitely be eliminating diffs.

I moved this up to the type tag PR and will consider re-adding non-duplicate reparsed nodes after Corsa is done and we don't care about baseline diffs anymore.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsTypeDefNoUnusedLocalMarked.errors.txt.diff
sandersn: typedefs are implicitly exported, and need to **not** be in the presence of `module.exports=`. I'm going to fix this separately because it's going to be tricky.

testdata/baselines/reference/submoduleAccepted/compiler/localRequireFunction.errors.txt.diff
sandersn: I'm not sure why the dupe is happening, but the checker is supposed to avoid resolving `require` to the global one if there is a local one available, so I need to stop that from happening.

In other words, the `const fs = require('fs')` below is **not** supposed to be an import, just a call to the local function.

andrewbranch:
Haven’t looked at the code yet, but from the description, I will note that

```js
exports.a = "a";
```

is more common than

```js
module.exports.a = "a";
```

in my experience. (Our own CommonJS emit, for example, does this.) (The way to reason about this, if it helps: the module system will use the value of `module.exports` as the source of truth, but both `module` and `exports` are free variables in the CJS script scope, with `module.exports` initialized to `exports` and `exports` initialized to an empty object. So by default, when you mutate the properties of `exports`, you mutate the properties of `module.exports`. But if you assign `exports = ...`, that rebinds the free variable and `module.exports` is unaffected. Similarly, if you reassign to `module.exports = {}`, subsequent mutations of `exports.a` will be ineffective.)

> Should we support local declaration of module.exports.x=? It is possible to hack the checker and name resolution to make it possible to refer to module.exports.x inside the CommonJS file, but it won't be pretty. I'm leaning toward not supporting this, at least until I can run top400 JS on Corsa after this PR. Based on the top100 JS corpus I looked at earlier, it's rare to reference an exported object within the exporting file.

When searching for this, make sure you're looking for the `exports.x` form too. I think this is fairly common.

> A secondary question is: should module, exports and require error without a declaration, suggesting that you install @types/node? Strada never does this for JS, but in Corsa I think it makes sense to ask people with checkJs on to install @types/node if they're using CommonJS. This does introduce lots of errors into the baselines, but remember that plain JS files report only syntax errors.

I would not do this—it seems like we’re making CommonJS much less powerful, so I’m not sure it makes sense to make it more nitpicky at the same time.

I think the broad questions about what can be dropped depends on the high level goal/philosophy here. If I were writing new CommonJS code (??), I would be happy with the level of support you have here. But existing code does all kinds of weird stuff, and I don’t think you’re going to find much of it in the `topNNN` suite. If we had a `bottom10%`, though... Are those users actively maintaining their code and upgrading TS versions? I doubt it.

sandersn:
> I will note that `exports.a = "a";` is more common than `module.exports.a = "a";`in my experience.

Thanks, my intuitions about that are out-of-date.

> I think the broad questions about what can be dropped depends on the high level goal/philosophy here. 

The customers I'm trying to support are: people writing TS-in-JSDoc, people writing miscellaneous unchecked JS, and people browsing emitted JS in node_modules. The only code base in topNNN that I've seen with commonjs is webpack, and it uses only a single form (`module.exports=` + `@typedef`). CommonJS is basically for understanding JS in node_modules.

andrewbranch:
> people browsing emitted JS in node_modules

This contingency will likely be harmed by the removal of any require/exports-related analysis, unfortunately.

testdata/baselines/reference/submoduleAccepted/compiler/modulePreserve4.errors.txt.diff
```diff
@@ -120,67 +106,64 @@
...
-+!!! error TS2305: Module '"/a"' has no exported member 'y'.
++!!! error TS1286: ESM syntax is not allowed in a CommonJS module when 'verbatimModuleSyntax' is enabled.
      import a1 = require("./a"); // Error in JS
 -    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 -!!! error TS8002: 'import ... =' can only be used in TypeScript files.
      const a2 = require("./a"); // { x: 0 }
-+               ~~~~~~~
-+!!! error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
++          ~~
++!!! error TS1286: ESM syntax is not allowed in a CommonJS module when 'verbatimModuleSyntax' is enabled.
```
andrewbranch: This is likely a mistake in this PR—this error is flagging `require` as ESM syntax, which is incorrect.

sandersn:
The latest commit resolves `module.exports` and `exports` more or less the same way that Strada did -- by treating them as the source file's symbol.

I still need to look through the baselines for other bugs and limitations, and fix up the errors that you all have already pointed out.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsTypeDefNoUnusedLocalMarked.errors.txt.diff
sandersn: Thinking about this more, I'll probably need to bring back `delayedBindJSDocTypedef` from Strada. Unfortunate, but it will still be simpler than Strada's, I promise.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsTypeDefNoUnusedLocalMarked.types.diff
```diff
@@ -10,9 +10,9 @@
 ->exports : (foo: Foo) => string
 ->(void 0) : FooFun
 +>module.exports = /** @type {FooFun} */(void 0) : (foo: typeof Foo) => string
-+>module.exports : any
-+>module : any
-+>exports : any
++>module.exports : (foo: typeof Foo) => string
```
sandersn, line 13: `typeof Foo` is correct -- the typeof-insertion insertion shenanigans in Strada were doing the wrong thing.

testdata/baselines/reference/submoduleAccepted/compiler/commonJsExportTypeDeclarationError.errors.txt.diff
sandersn: I suspect it's a difference in error reporting ("Type expected" and "'=' expected" appear to be on the same location), but I'll confirm.

testdata/baselines/reference/submoduleAccepted/compiler/commonJsExportTypeDeclarationError.errors.txt.diff
```diff
@@ -45,9 +35,5 @@
 +!!! error TS1110: Type expected.
 +==== types3.ts (1 errors) ====
```
sandersn, line 36: weird difference in spacing between Strada and Corsa baselines.

testdata/baselines/reference/submoduleAccepted/compiler/commonJsExportTypeDeclarationError.errors.txt.diff
sandersn: Yep. The code is working as expected in Corsa but not in Strada. The two messages are both on the EOF token in Corsa. I'm not sure why they evade position-based deduping in Strada.

testdata/baselines/reference/submoduleAccepted/compiler/commonjsAccessExports.types.diff
```diff
@@ -1,37 +1,16 @@
...
  {
      // 'exports' does not provide a contextual type to a function-class
      exports.Cls = function() {
 ->exports.Cls = function() {        this.x = 0;    } : typeof Cls
 ->exports.Cls : typeof Cls
-->exports : typeof import("/a")
-->Cls : typeof Cls
-->function() {        this.x = 0;    } : typeof Cls
 +>exports.Cls = function() {        this.x = 0;    } : () => void
 +>exports.Cls : any
```
sandersn, line 10: I'm not sure why `exports.Cls : any` but it's not a problem with `exports` resolution, because `exports.x` works above. I suspect it's something wrong with partially implemented construction function typing.

jakebailey:
Is there anything left in this PR? It's hard to view and the thread is long so I'm not sure where we are at; I tried to skim the code until the tab crashed, and I'm not sure there's anything that jumps out at me, especially if this is a net improvement over the current state that we can update later.

sandersn:
@jakebailey You're probably right, I've just been fixing bugs that would be easier to review individually.

Still,
1. I want to look over the baselines one more time.
2. I have many many unused baselines I need to remove.
3. I'll ping you directly when I'm done with that.

jakebailey (approved):
LGTM just to get this going.

PR #749, opened 2025-04-04
LSP Notification Message Type

dyl10s:
- Added lsp notification message type, as the todo comment stated the difference is if an ID is provided or not. 
- A few of the messages that were being handled were notifications, so they are all now parsed/typed correctly.


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -43,56 +43,88 @@ func (id *ID) UnmarshalJSON(data []byte) error {
...
+type RequestOrNotificationMessage struct {
+	NotificationMessage *NotificationMessage
+	RequestMessage      *RequestMessage
+}
+
+func (r *RequestOrNotificationMessage) UnmarshalJSON(data []byte) error {
 	var raw struct {
 		JSONRPC JSONRPCVersion  `json:"jsonrpc"`
-		ID      *ID             `json:"id"`
+		ID      *ID             `json:"id,omitempty"`
```
jakebailey: Generally speaking, we should be using `omitzero` (new in Go 1.24) rather than `omitempty`.

jakebailey: There are other of these, FWIW, I just commented on the one.

dyl10s: I went ahead and normalized this file to always use omitzero!

jakebailey:
I am currently redoing this code to deal with the fact that we need to handle requests, notifications, and responses going both directions over the pipe. For this, I don't actually think we should split these types apart, but rather we'll have one thing that's just "Message" that represents whatever can be sent over the wire. With that in mind, I think I'm going to close this; sorry I was slow to review this but I don't think it will line up with what we're going to need right away.

PR #750, opened 2025-04-04
Add a new cached VFS wrapper

jakebailey:
`cachedvfs.From` wraps another FS implementation with caches that can be cleared.

We should be able to use this within a Program (where we assume the FS is unchanging).

To make testing easier, I've finally introduced `moq` into the repo to produce test mocks for interfaces. Most of the added code is just that.


internal/vfs/cachedvfs/cachedvfs.go
```diff
@@ -0,0 +1,82 @@
...
+func (vfs *FS) WalkDir(root string, walkFn vfs.WalkDirFunc) error {
+	return vfs.fs.WalkDir(root, walkFn)
+}
+
+func (vfs *FS) WriteFile(path string, data string, writeByteOrderMark bool) error {
+	// TODO: should this call ClearCache?
+	return vfs.fs.WriteFile(path, data, writeByteOrderMark)
+}
+
+func cached[Arg any, Ret any](cache *sync.Map, key Arg, fn func(Arg) Ret) Ret {
```
jakebailey: I thought this would get inlined, but it unfortunately isn't, which means we pay the closure cost on callers. I'll inline this manually, I guess.

andrewbranch (approved):

PR #751, opened 2025-04-04
Fixed Uri length

armanio123:
`documentUriToFileName` was returning an incorrect when the Uri contained a drive letter. ie:: 

`file:///C:/Folder1/file.ts` --> `C:/lder1/file.ts`


jakebailey:
We should really add some tests for these converters 😄

Mustafiz04 (approved):
Agree with @jakebailey to have a test for this.

andrewbranch (approved):

armanio123:
@jakebailey Agree, I'll see to adding some tests.

PR #758, opened 2025-04-07
Copy full /.lib dir into testfs when needed

jakebailey:
Add all of `/.lib` into the test FS when needed; the previous PR did this for what was mentioned in `@libFiles` but that option seemingly is meant to control if the files are included in the Program by default, and `/.lib` is just always mounted in every test anyway.

This fixes a whole lot of diffs.


testdata/baselines/reference/submodule/compiler/callsOnComplexSignatures.symbols.diff
```diff
@@ -57,7 +57,7 @@
  >url : Symbol(url, Decl(callsOnComplexSignatures.tsx, 75, 20))
  >React : Symbol(React, Decl(callsOnComplexSignatures.tsx, 1, 6))
 ->ReactNode : Symbol(React.ReactNode, Decl(react16.d.ts, 216, 49))
-+>ReactNode : Symbol(ReactNode)
++>ReactNode : Symbol(ReactNode, Decl(react16.d.ts, 216, 49))
```
jakebailey, line 60: Note the improvement here; this now just differs due to symbol naming.

(And of course, the loads of deleted errors.)

ahejlsberg (approved):

PR #761, opened 2025-04-07
s/Js/JS/g

sandersn:
I chose the spellings

- `jsdoc` not `jsDoc`
- `JS` not `Js`
- `CommonJS` not `CommonJs`

I am OK with Js instead of JS, but I like JS better and I am the one typing it all the time right now.

- I didn't change compiler options since those are external facing
- I didn't change the tspath Extension enum because there are lot of related Ts, Dts, Dcts, ... entries.
- I didn't change JSX, but only because I'm typing JS all the time right now, not JSX.
- I changed harnessutils because, in the triple `{ Js, Dts, Maps }`, only one of those is plural, which was enough additional justification for a change.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -9467,35 +9467,35 @@ func (node *JSDocTypedefTag) Clone(f *NodeFactory) *Node {
 type JSDocTypeLiteral struct {
 	TypeNodeBase
 	DeclarationBase
-	JsDocPropertyTags []*Node
+	JSDocPropertyTags []*Node
 	IsArrayType       bool
 }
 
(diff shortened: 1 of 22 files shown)


jakebailey:
I definitely like the more Go-ish form (JS), but I do wonder if we should actually change the style or how this would affect the public API we're constructing (which would be better roughly matching the old API).

andrewbranch (approved):

sandersn:
I'd rather have a public/internal distinction and nice internals. The API isn't going to change very much once it's done and the names are mostly already in place from the old API.

sandersn:
It's worth pointing out that this is easy to change back and forth if we change our minds later.

PR #762, opened 2025-04-07
Port JSX type checking

ahejlsberg:

PR #763, opened 2025-04-07
Return empty diags for files with skipped checking

jakebailey:
This mirrors `getBindAndCheckDiagnosticsForFileNoCache`, which looks like:

```ts
    function getBindAndCheckDiagnosticsForFileNoCache(
        sourceFile: SourceFile,
        cancellationToken: CancellationToken | undefined,
        nodesToCheck?: Node[],
    ): readonly Diagnostic[] {
        return runWithCancellationToken(() => {
            if (skipTypeChecking(sourceFile, options, program)) {
                return emptyArray;
            }

            const typeChecker = getTypeChecker();
```

Notably I am _not_ doing this in `getBindDiagnosticsForFile` since that isn't actually called by anyone except `tsgo` for timing.

This fixes the baseline checks that were commented out in #762.


internal/compiler/program.go
```diff
@@ -311,10 +311,19 @@ func (p *Program) getSyntacticDiagnosticsForFile(sourceFile *ast.SourceFile) []*
 }
 
 func (p *Program) getBindDiagnosticsForFile(sourceFile *ast.SourceFile) []*ast.Diagnostic {
+	// TODO: restore this; tsgo's main depends on this function binding all files for timing.
```
ahejlsberg, line 314: When and why would we restore this? Seems odd for `skipLibChecks` to affect binding.

ahejlsberg (approved):

internal/compiler/program.go
jakebailey: We have one function that gets both bind and check diagnostics (`getBindAndCheckDiagnosticsForFileNoCache`), and it's silenced by this same function in Strada. Why? I have no idea, but that's how it was. It's just that they are split apart in Corsa for timing reasons at the moment.

PR #766, opened 2025-04-08
Fix pattern ambient modules

### LINKED ISSUE #697: [bug] `tsgo` failing to resolve custom declaration files
The company I work for has a TS project with over 90k+ files. I recently ran `tsgo` against this project and was very impressed with how quick it was. However, it failed to resolve **our custom declaration files** resulting in a huge number of type errors like this:

```
Cannot find module './some_file.css' or its corresponding type declarations
...
Found 30381 errors in 16139 files
```

For example, we use CSS modules and have an associated `css_modules.d.ts` file of form:

```typescript
declare type HtmlEncodedCssList = ...;

declare module '*.css' {
  ...
}
```

When I run `tsgo -listFilesOnly` this declarations file does show up:

```
./typescript-go/built/local/tsgo -p <omitted>/web/tsconfig.json -listFilesOnly | grep css_modules
<omitted>/declarations/css_modules.d.ts
```

jakebailey:
Fixes #697

A lot of this function hasn't been implemented but this seems to do the trick for pattern-based ambient modules.


andrewbranch (approved):

jakebailey:
Confirmed with people externally that this fixes their issues, yay.

PR #772, opened 2025-04-08
Move LSP↔️LS converters to ls package

andrewbranch:


internal/core/core.go
```diff
@@ -524,3 +524,7 @@ func levenshteinWithMax(s1 []rune, s2 []rune, maxValue float64) float64 {
 func Identity[T any](t T) T {
 	return t
 }
+
+func PtrTo[T any](v T) *T {
```
jakebailey: At some level I wish we would just copy and paste this into packages we use it, since `core.PtrTo` is a long word for a simple thing, and it means we keep importing core everywhere which makes modifying it for unrelated reasons cause rebuilds of everything, but, I guess it's fine

internal/lsp/lspconv/converters.go
```diff
@@ -13,39 +13,47 @@ import (
...
-	projectService   *project.Service
-	positionEncoding lsproto.PositionEncodingKind
+type ScriptInfo interface {
+	Text() string
+	LineMap() *LineMap
 }
 
-func (c *converters) toLspRange(fileName string, textRange core.TextRange) (lsproto.Range, error) {
-	scriptInfo := c.projectService.GetScriptInfo(fileName)
+type ConvertersHost interface {
```
jakebailey: This could just be called `Host` or something (package is already called `lspconv`), or we could just provide the func only to `Converters`.

jakebailey (approved):

internal/core/core.go
andrewbranch: I mean, everything uses `core` for many other things (why is `TextPos` a named type??) so this doesn’t make material progress toward unwinding the dependency, but sure

jakebailey: No, it's alright, just leave it; the PR seems okay as-is.

andrewbranch:
I forgot the converters reference types defined in `ls`, so I put them in `ls` instead for now. But maybe we should delete the `ls` types and avoid those conversions altogether.

gabritto (approved):

jakebailey (approved):

PR #773, opened 2025-04-08
Add support for SourceMap emit and baselines

rbuckton:
This wires up source map emit to the emitter and enables source map baselines. After some fixes, the few source map diffs seem to be mostly due to missing transforms.

This also makes a minor change to the `ast.SourceFile` type to make `Text()` a method rather than a field. This was partially done to support `sourcemap.Source`, but also to make it easier to access the source file text while debugging as the Go debugger gets confused by `Text` being a field on `SourceFile` but a method on `Node`, which is observed when adding `file.Text` as a Watch expression or when referencing it in the Debug console.


internal/sourcemap/util.go
```diff
@@ -0,0 +1,26 @@
+package sourcemap
+
+import (
+	"regexp"
+	"strings"
+	"unicode"
+)
+
+var (
+	sourceMapCommentRegExp       = regexp.MustCompile(`^\/\/[@#] sourceMappingURL=(.+)\r?\n?$`)
```
jakebailey: Eventually we'll want to eliminate these since Go's regex perf is pretty bad.

internal/stringutil/util.go
```diff
@@ -2,6 +2,7 @@
 package stringutil
 
 import (
+	"net/url"
```
jakebailey, line 5: I am suspicious of using this for anything for JS stuff; can we stick a `!!!` onto the uses of this so we can say "come back and see if we should change this"? By using it we're somewhat at the whim of what Go does here, which could be good but could be bad.

testdata/baselines/reference/submodule/compiler/commonSourceDirectory.js.map.diff
```diff
@@ -0,0 +1,8 @@
+--- old.commonSourceDirectory.js.map
```
jakebailey, line 1: I'm sort of wondering how useful this is when the maps are single lines so will always exist. Maybe we should do something special to format these at some point? (Might be annoying though).

internal/stringutil/util.go
rbuckton: This should be equivalent to `encodeURIComponent`. I went through and compared the encoding rules Go uses to the ones JS uses, and these two mechanisms are equivalent. However, Go has no equivalent to `encodeURI`, which is why I implemented it myself.

testdata/baselines/reference/submodule/compiler/commonSourceDirectory.js.map.diff
rbuckton: Something other than the `.sourcemap.txt` files? I would venture to guess that 99% of the time the diffs we see for `.js.map` are going to be in the `"mappings"` field, which is still going to be on a single line and will be just as inscrutable even in a formatted JSON file.

internal/compiler/emitter.go
```diff
@@ -258,7 +399,12 @@ func sourceFileMayBeEmitted(sourceFile *ast.SourceFile, host EmitHost, forceDtsE
 		return false
 	}
 
-	// !!! Source file from node_modules are not emitted
+	// !!! Source file from node_modules are not emitted. In Strada, this depends on module resolution and uses
+	// `sourceFilesFoundSearchingNodeModules` in `createProgram`. For now, we will just check for `/node_modules/` in
+	// the file name.
+	if strings.Contains(sourceFile.FileName(), "/node_modules/") {
```
jakebailey, line 405: I would suggest using `.Pah()` here but I assume this is just temporary?

rbuckton: Yes, this is temporary.

jakebailey (approved):

PR #777, opened 2025-04-08
Add typesafe sync.Map wrapper

jakebailey:
I think there's a proposal to add this exact type to the stdlib eventually, but I'm finding it increasingly annoying to deal with `any` when making new `sync.Map`s.

### DIFF
--- a/internal/collections/syncmap.go
+++ b/internal/collections/syncmap.go
@@ -1,0 +1,59 @@
+package collections
+
+import "sync"
+
+type SyncMap[K comparable, V any] struct {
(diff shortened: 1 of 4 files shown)


internal/collections/syncmap.go
```diff
@@ -0,0 +1,59 @@
+package collections
```
jakebailey, line 1: This is here in `core` so that leaf packages like `cachedvfs` don't need to import `core`.

andrewbranch (approved):

PR #779, opened 2025-04-09
Reorganize fileLoader, make it lock-free via SyncMaps

jakebailey:
I'm currently modifying this to work on JSX stuff; sending this refactor in early.

I'm going to add more stuff to the file loader for JSX, so converting its return into a struct will make a future PR simpler.

Additionally, there's no real need to use full locks here; sync maps are sufficient for all uses, so we can do everything in the loader lock-free (more or less).

### DIFF
--- a/internal/compiler/fileloader.go
+++ b/internal/compiler/fileloader.go
@@ -16,24 +16,22 @@ import (
 )
 
 type fileLoader struct {
-	host            CompilerHost
-	programOptions  ProgramOptions
-	compilerOptions *core.CompilerOptions
+	host                CompilerHost
+	programOptions      ProgramOptions
+	compilerOptions     *core.CompilerOptions
+	resolver            *module.Resolver
+	defaultLibraryPath  string
+	comparePathsOptions tspath.ComparePathsOptions
+	wg                  core.WorkGroup
+	supportedExtensions []string
 
-	resolver        *module.Resolver
-	resolvedModules collections.SyncMap[tspath.Path, module.ModeAwareCache[*module.ResolvedModule]]
+	rootTasks []*parseTask
 
-	sourceFileMetaDatasMutex sync.RWMutex
-	sourceFileMetaDatas      map[tspath.Path]*ast.SourceFileMetaData
+	resolvedModules     collections.SyncMap[tspath.Path, module.ModeAwareCache[*module.ResolvedModule]]
+	sourceFileMetaDatas collections.SyncMap[tspath.Path, *ast.SourceFileMetaData]
 
-	mu                      sync.Mutex
-	wg                      core.WorkGroup
-	tasksByFileName         map[string]*parseTask
-	currentNodeModulesDepth int
-	defaultLibraryPath      string
-	comparePathsOptions     tspath.ComparePathsOptions
-	rootTasks               []*parseTask
-	supportedExtensions     []string
+	tasksByFileNameMu sync.Mutex
+	tasksByFileName   map[string]*parseTask
 }
 
 type processedFiles struct {
@@ -88,7 +86,7 @@ func processAllProgramFiles(
 	return processedFiles{
 		files:               append(libFiles, files...),
 		resolvedModules:     loader.resolvedModules.ToMap(),
-		sourceFileMetaDatas: loader.sourceFileMetaDatas,
+		sourceFileMetaDatas: loader.sourceFileMetaDatas.ToMap(),
 	}
 }
 
@@ -121,8 +119,8 @@ func (p *fileLoader) addAutomaticTypeDirectiveTasks() {
 
 func (p *fileLoader) startTasks(tasks []*parseTask) {
 	if len(tasks) > 0 {
-		p.mu.Lock()
-		defer p.mu.Unlock()
+		p.tasksByFileNameMu.Lock()
+		defer p.tasksByFileNameMu.Unlock()
 		for i, task := range tasks {
 			// dedup tasks to ensure correct file order, regardless of which task would be started first
 			if existingTask, ok := p.tasksByFileName[task.normalizedFilePath]; ok {
@@ -235,9 +233,7 @@ func (t *parseTask) start(loader *fileLoader) {
 }
 
 func (p *fileLoader) loadSourceFileMetaData(path tspath.Path) {
-	p.sourceFileMetaDatasMutex.RLock()
-	_, ok := p.sourceFileMetaDatas[path]
-	p.sourceFileMetaDatasMutex.RUnlock()
+	_, ok := p.sourceFileMetaDatas.Load(path)
 	if ok {
 		return
 	}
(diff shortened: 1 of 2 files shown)


jakebailey:
Bug in the lint rule, nice.

EDIT: I can't fix it without https://github.com/golang/go/issues/73278 or a bunch of finagling to fix up the CFG, so I'll just rename the var to make the lint rule happy.

internal/compiler/fileloader.go
```diff
@@ -277,12 +273,7 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile)
 
 		resolutionsInFile := make(module.ModeAwareCache[*module.ResolvedModule], len(resolutions))
 
-		p.resolvedModulesMutex.Lock()
-		defer p.resolvedModulesMutex.Unlock()
-		if p.resolvedModules == nil {
-			p.resolvedModules = make(map[tspath.Path]module.ModeAwareCache[*module.ResolvedModule])
-		}
-		p.resolvedModules[file.Path()] = resolutionsInFile
+		p.resolvedModules.Store(file.Path(), resolutionsInFile)
```
andrewbranch: Isn't this the worst use case for a `sync.Map`? `p.resolvedModules` is write-only until file loading is complete, and every write will occur under a different key. Isn't this just offloading our mutex locking onto the one in `sync.Map`, with the added overhead of checking every time whether the key exists in the readonly map?

jakebailey: It shouldn't be; `sync.Map` is a sharded map to allow for concurrent use with fewer locks. The docs say:

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.

So I expect we are in category 2.

andrewbranch: Have you measured the effect?

andrewbranch: My understanding is that writing to a non-existent key always locks, and reading the source code it sure looks that way

jakebailey: I'll do a proper benchmark, but I had hyperfine'd it before/after and saw a small improvement.

jakebailey: As far as I can tell it's all noise; making a program load benchmark for `src/compiler`, I see:

```
goos: linux
goarch: amd64
Program-20   29.97m ± 3%   29.16m ± 2%  -2.72% (p=0.019 n=10)

Program-20   39.86Mi ± 0%   39.90Mi ± 0%  +0.09% (p=0.002 n=10)

           │  allocs/op  │  allocs/op   vs base               │
Program-20   128.2k ± 0%   129.2k ± 0%  +0.84% (p=0.000 n=10)
```

But rerunning I get:

```
goos: linux
goarch: amd64
           │   old.txt   │            new.txt            │
           │   sec/op    │   sec/op     vs base          │
Program-20   29.97m ± 3%   29.39m ± 1%  ~ (p=0.063 n=10)

Program-20   39.86Mi ± 0%   39.89Mi ± 0%  +0.08% (p=0.004 n=10)

           │  allocs/op  │  allocs/op   vs base               │
Program-20   128.2k ± 0%   129.2k ± 0%  +0.82% (p=0.000 n=10)
```

And a previous run (whose data I lost in my terminal scrollback) showed like 3% worse one run.

So, I don't know if it's actually better or worse.

jakebailey: Hm, this is way too short to be real, I may have done a bad benchmark.

jakebailey: Okay, this time actually testing correctly:

```
goos: linux
goarch: amd64
        │   old.txt   │            new.txt            │
        │   sec/op    │   sec/op     vs base          │
Program   245.1m ± 2%   242.0m ± 1%  ~ (p=0.143 n=10)

Program   132.4Mi ± 0%   132.5Mi ± 0%  +0.04% (p=0.000 n=10)

        │  allocs/op  │  allocs/op   vs base               │
Program   236.5k ± 0%   238.2k ± 0%  +0.71% (p=0.000 n=10)
```

jakebailey: With more benchmarks running `go test -run=- -bench=. -benchmem -count=10 -cpu=1 ./internal/compiler` (`-cpu=1` tells go to not parallelize benchmarks itself, since this benchmark is already internally concurrent).

```
goos: linux
goarch: amd64
NewProgram/BasicFileOrdering     28.00m ± 3%   27.59m ± 2%       ~ (p=0.315 n=10)
NewProgram/FileOrderingImports   27.76m ± 2%   28.08m ± 1%  +1.13% (p=0.035 n=10)
NewProgram/FileOrderingCycles    27.81m ± 1%   28.07m ± 2%       ~ (p=0.052 n=10)
NewProgram/compiler              245.7m ± 7%   246.7m ± 6%       ~ (p=0.971 n=10)
geomean                          48.01m        48.13m       +0.25%

NewProgram/BasicFileOrdering     12.46Mi ± 0%   12.48Mi ± 0%  +0.13% (p=0.000 n=10)
NewProgram/FileOrderingImports   12.49Mi ± 0%   12.50Mi ± 0%  +0.13% (p=0.000 n=10)
NewProgram/FileOrderingCycles    12.49Mi ± 0%   12.51Mi ± 0%  +0.13% (p=0.000 n=10)
NewProgram/compiler              132.5Mi ± 0%   132.5Mi ± 0%  +0.04% (p=0.000 n=10)
geomean                          22.52Mi        22.55Mi       +0.11%

                               │  allocs/op  │  allocs/op   vs base               │
NewProgram/BasicFileOrdering     32.71k ± 0%   33.17k ± 0%  +1.40% (p=0.000 n=10)
NewProgram/FileOrderingImports   33.16k ± 0%   33.64k ± 0%  +1.44% (p=0.000 n=10)
NewProgram/FileOrderingCycles    33.24k ± 0%   33.72k ± 0%  +1.44% (p=0.000 n=10)
NewProgram/compiler              236.5k ± 0%   238.2k ± 0%  +0.71% (p=0.000 n=10)
geomean                          54.04k        54.71k       +1.25%
```

The timing is noisy each time. The memory is not, however.

andrewbranch: Is this main vs. the branch? I’m not sure about the metadata `sync.Map`, I just think this one on its own isn’t helpful. I think a sharded map would be better, but the locking is so brief that it might not matter.

jakebailey: Yes, this is main versus the branch.

I guess if this seems not so helpful I can just split this out into the main change I wanted; having the file loader return a single struct so I didn't need to keep adding stuff.

But, I'm not sure that it's actually any worse besides some confusing memory allocation increases; I find this form a lot clearer than using maps and mutexes where we're _definitely_ locking the entire thing.

jakebailey:
Realized that none of this even needs to be locked; we can just put the info onto the tasks, which we collect later anyway.

andrewbranch (approved):

PR #780, opened 2025-04-09
Handle import helpers and JSX factory imports

### LINKED ISSUE #765: [bug] JSX support in `preserve` mode w/ custom `jsxImportSource` causing `TS2875`
I have a tsconfig.json approximately like:

```json
"compilerOptions": {
        "module": "es2022",
        "jsx": "preserve",
        "jsxImportSource": "external/solid",
        "baseUrl": ".",
        "paths": {
                "external/*": ["./external/*"]
        },
        "target": "es2022",
```

```
error TS2875: This JSX tag requires the module path 'external/solid/jsx-runtime' to exist, but none could be found. Make sure you have types for the appropriate package installed.
```

even though this file exists:

```
❯ cat external/solid/jsx-runtime.d.ts
export * from "./solid/jsx";

❯ ls -l external/solid/solid/jsx.d.ts
-rw-r--r--  1 tmm1  staff  73590 Apr  4 17:34 external/solid/solid/jsx.d.ts
```

Is this due to `compilerOptions.paths`?

### LINKED ISSUE #767: [bug] failing to resolve `react/jsx-runtime` when type checking TSX files
Apologies if this is considered a duplicate of https://github.com/microsoft/typescript-go/issues/765 -- but I believe it is slightly different due to a lack of `compilerOptions.jsx` support in typescript-go (vs `compilerOptions.jsxImportSource`)...

The company I work for has a TS project with over 90k+ files. I recently ran `tsgo` against this project and was very impressed with how quick it was. However, it reporting over 38k errors when type checking our TSX files with form:

```
This JSX tag requires the module path 'react/jsx-runtime' to exist, but none could be found. Make sure you have types for the appropriate package installed.
...
Found 41250 errors in 35304 files.
```

I believe this could be due to `tsgo` currently not supporting our usage of the `compilerOptions.jsx` tsconfig property so wanted to create an issue for tracking support.

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
  }
}
```

jakebailey:
Fixes #765
Fixes #767
Includes #779 until that's merged.

This implements import helper and JSX factory import node synthesis and resolution. Since the AST node is ummutable, we cannot modify `SourceFile.Imports`. This info is now stored on the `Program` rather than at special fixed instances in `imports`.

This should allow us to reuse ASTs more often, and means we stop having to work around these synthezied imports in other parts of the code (LS features, mainly).

This PR isn't perfect; it turns out we're missing implementations of `moduleDetection` and `externalModuleIndicator`, so we're not correctly determining that JSX files are modules because the `jsx` setting plays a part in determining if a file is a module (yay).

Fixing that will have to be a follow-up PR but that's going to be a lot more invasive.


testdata/baselines/reference/submodule/compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=auto).errors.txt.diff
```diff
@@ -0,0 +1,44 @@
+--- old.commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=auto).errors.txt
```
jakebailey, line 1: This is broken because we do not correctly set the module indicators.

testdata/baselines/reference/submodule/compiler/tslibMissingHelper.symbols.diff
```diff
@@ -1,10 +1,9 @@
...
- >Promise : Symbol(Promise, Decl(lib.es5.d.ts, --, --), Decl(lib.es2015.iterable.d.ts, --, --), Decl(lib.es2015.promise.d.ts, --, --), Decl(lib.es2015.symbol.wellknown.d.ts, --, --))
+@@= skipped -17, +17 lines =@@
  
--=== /node_modules/tslib/tslib.d.ts ===
--export const notAHelper: any;
+ === /node_modules/tslib/tslib.d.ts ===
+ export const notAHelper: any;
 ->notAHelper : Symbol(notAHelper, Decl(tslib.d.ts, --, --))
--
++>notAHelper : Symbol(notAHelper, Decl(tslib.d.ts, 0, 12))
```
jakebailey, line 8: Funny, Strada's regex here is bad, while in Corsa we "fixed" this.

andrewbranch (approved):

jakebailey:
Future TODO: stop using synthetic nodes for this, because we only use them to then plug into module resolution anyway. There's no real reason to use a Node here other than "the code used to work this way and it's not very trivial to change", but changing it will be faster and make the loader lock-free again.

PR #783, opened 2025-04-09
cmd/tsgo: add -targetFile to dev cli for easier testing of EmitOptions.TargetSourceFile

tmm1:


jakebailey:
Why is this needed? Our emit suite is extensive, and `outDir` works to look at emit as well.

tmm1:
This lets me check/emit single files instead of the whole project. The alternative is pass a custom configuration with [modified files/imports](https://github.com/gustavopch/tsc-files/blob/3065ba8266c280980344a3f9cbfe7159c4a963ae/cli.js#L55-L58)

cc https://github.com/microsoft/TypeScript/issues/41865 https://github.com/microsoft/TypeScript/issues/27379

Basically just a hack, to let me recompile files which I know have been modified (using external file watcher). Stop gap until real incremental mode.

The `TargetSourceFile` does seem to work as expected, so maybe I can look at wiring it up to optimize `tsc -w` mode.

tmm1:
whole project (6k files):

```
Parse time:       0.707s
Bind time:        0.047s
Check time:       3.468s
Emit time:        2.193s
Total time:       6.415s
```

single file only:

```
Parse time:       0.666s
Bind time:        0.000s
Check time:       0.069s
Emit time:        0.012s
Total time:       0.747s
```

It appears the whole project is still parsed.

tmm1:
My goal is to try to integrate tsgo into the vscode build system.

My initial idea was to shell out to tsgo in a debounced fashion, passing in lists of modified files to re-generate. These can be sent to a tmp outDir, and then slurped back up off the filesystem when the command completes.

But maybe there's a better solution I'm missing- I looked briefly into the `LanguageClient` lsp approach, but it wasn't clear to me if the required functionality was present there yet.

---

You may be familiar, but the way it currently works is by instantiating typescript lsp, uploading file contents, and then asking for diagnostics and emitted outputs.

https://github.com/microsoft/vscode/blob/459f974eb62bcecf3ffd36e35a6df3588c5d3614/build/lib/tsb/builder.ts#L49
```ts
export function createTypeScriptBuilder(config: IConfiguration, projectFile: string, cmd: ts.ParsedCommandLine): ITypeScriptBuilder {
	const host = new LanguageServiceHost(cmd, projectFile, _log);
	const service = ts.createLanguageService(host, ts.createDocumentRegistry());
```

https://github.com/microsoft/vscode/blob/459f974eb62bcecf3ffd36e35a6df3588c5d3614/build/lib/tsb/builder.ts#L70
```ts
			host.addScriptSnapshot(file.path, new VinylScriptSnapshot(file));
```

https://github.com/microsoft/vscode/blob/459f974eb62bcecf3ffd36e35a6df3588c5d3614/build/lib/tsb/builder.ts#L132
```ts
					const output = service.getEmitOutput(fileName);
```

---

jakebailey:
> Basically just a hack, to let me recompile files which I know have been modified (using external file watcher). Stop gap until real incremental mode.
> 
> The `TargetSourceFile` does seem to work as expected, so maybe I can look at wiring it up to optimize `tsc -w` mode.

> It appears the whole project is still parsed.

Basically none of the infrastructure is there to make a "full" incremental watch mode work. The current watch mode always does a full program load, and module resolution is not yet cached.

I think it's a bit early to actually put something like this PR into the command, and we're also likely to transition the main entrypoint into the actual tsc CLI instead in the near future, once stats and other functionality is implemented.

tmm1:
> I think it's a bit early to actually put something like this PR into the command

Sounds good, that's what I figured. I can keep this on a branch for my own experiments.

PR #784, opened 2025-04-10
Add -sourceMap option to the dev cli

tmm1:
cc @rbuckton @jakebailey https://github.com/microsoft/typescript-go/pull/773


jakebailey:
Does `tsgo tsc --sourceMap` not work?

tmm1:
Yea that works. I was using `-outDir` which is only in the dev cli, but it sounds like I should add that to tsc instead.

EDIT: n/m I missed it in the help output.

PR #786, opened 2025-04-10
More type checking fixes

ahejlsberg:


internal/checker/checker.go
```diff
@@ -15198,16 +15198,26 @@ func (c *Checker) getTypeOfVariableOrParameterOrPropertyWorker(symbol *ast.Symbo
 	if symbol == c.requireSymbol {
 		return c.anyType
 	}
-	// !!! Handle SymbolFlagsModuleExports
+	if symbol.Flags&ast.SymbolFlagsModuleExports != 0 && symbol.ValueDeclaration != nil {
```
ahejlsberg, line 15201: We're not currently creating any `SymbolFlagsModuleExports` symbols, but presumably we will as we implement JS/JSDoc checking. Otherwise, we can delete this if statement.

internal/checker/checker.go
```diff
@@ -17993,8 +18003,9 @@ func (c *Checker) addInheritedMembers(symbols ast.SymbolTable, baseSymbols []*as
 func (c *Checker) resolveDeclaredMembers(t *Type) *InterfaceType {
 	d := t.AsInterfaceType()
 	if !d.declaredMembersResolved {
+		members := c.getMembersOfSymbol(t.symbol)
 		d.declaredMembersResolved = true
-		d.declaredMembers = c.getMembersOfSymbol(t.symbol)
+		d.declaredMembers = members
```
ahejlsberg, line 18008: This change enables recursion through `getMembersOfSymbol` which we rely on to first resolve early-bound members so they can be referenced in resolution of late-bound members.

testdata/baselines/reference/submodule/compiler/genericIndexedAccessVarianceComparisonResultCorrect.types
```diff
@@ -46,15 +46,15 @@ type PickA_extends_PickB = Pick<A, 'x'> extends Pick<B, 'x'> ? true : false;
 >false : false
 
 type TA_extends_TB = T<A> extends T<B> ? true : false;                          // should be true
->TA_extends_TB : true
+>TA_extends_TB : false
```
jakebailey, line 49: "should be true" but the correct baselines are `false`, fun 😄

jakebailey (approved):

gabrielluizsf (approved):

testdata/baselines/reference/submodule/compiler/genericIndexedAccessVarianceComparisonResultCorrect.types
ahejlsberg: Indeed. It's basically a test to show that variance computation is sometimes too conservative.

PR #788, opened 2025-04-10
Make SourceFileAffectingCompilerOptions hold fully computed values, use in binder

jakebailey:
This PR makes `SourceFileAffectingCompilerOptions` consist of only _fully computed_ values (in prep for some other changes).

- We can now prove that the binder (and in the future, parser) only touch things that they need.
- The values can be computed once and reused.
- `SourceFileAffectingCompilerOptions` is smaller, allowing much more AST reuse.

The AST reuse results in the tests running 20-25% faster, and using 1GB less peak memory.

psrecord before:

old

psrecord after:

new

Before, it's maybe 8GB, before maybe 7GB.

### DIFF
--- a/internal/binder/binder.go
+++ b/internal/binder/binder.go
@@ -40,7 +40,7 @@ const (
 
 type Binder struct {
 	file                    *ast.SourceFile
-	options                 *core.CompilerOptions
+	options                 *core.SourceFileAffectingCompilerOptions
 	languageVersion         core.ScriptTarget
 	bindFunc                func(*ast.Node) bool
 	unreachableFlow         *ast.FlowNode
@@ -86,7 +86,7 @@ type ActiveLabel struct {
 func (label *ActiveLabel) BreakTarget() *ast.FlowNode    { return label.breakTarget }
 func (label *ActiveLabel) ContinueTarget() *ast.FlowNode { return label.continueTarget }
 
-func BindSourceFile(file *ast.SourceFile, options *core.CompilerOptions) {
+func BindSourceFile(file *ast.SourceFile, options *core.SourceFileAffectingCompilerOptions) {
 	// This is constructed this way to make the compiler "out-line" the function,
 	// avoiding most work in the common case where the file has already been bound.
 	if !file.IsBound() {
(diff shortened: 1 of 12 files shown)


internal/core/compileroptions.go
```diff
@@ -269,34 +269,25 @@ func (options *CompilerOptions) HasJsonModuleEmitEnabled() bool {
...
-		Jsx:                  options.Jsx,
-		JsxImportSource:      options.JsxImportSource,
-		ImportHelpers:        options.ImportHelpers,
-		AlwaysStrict:         options.AlwaysStrict,
-		ModuleDetection:      options.ModuleDetection,
-		AllowUnreachableCode: options.AllowUnreachableCode,
-		AllowUnusedLabels:    options.AllowUnusedLabels,
-		PreserveConstEnums:   options.PreserveConstEnums,
-		IsolatedModules:      options.IsolatedModules,
+func (options *CompilerOptions) SourceFileAffecting() *SourceFileAffectingCompilerOptions {
```
jakebailey, line 283: This is not cached anywhere since we'd have to put it on CompilerOptions, which means `sync.Once`, which means no more `DeepEqual`.

jakebailey, line 283: I'm not sure if this should continue to be a value type or not, since we pass it around and so on. But it's not that large, actually.

jakebailey: I tested and it keeping it a value doesn't actually hurt or help perf, so I could do that.

andrewbranch (approved):

internal/binder/binder.go
```diff
@@ -1331,7 +1331,7 @@ func (b *Binder) checkStrictModeWithStatement(node *ast.Node) {
 
 func (b *Binder) checkStrictModeLabeledStatement(node *ast.Node) {
 	// Grammar checking for labeledStatement
-	if b.inStrictMode && b.options.Target >= core.ScriptTargetES2015 {
+	if b.inStrictMode && b.options.EmitScriptTarget >= core.ScriptTargetES2015 {
```
jakebailey, line 1334: This is a porting bug I'm fixing, FWIW.

PR #790, opened 2025-04-11
Use main CI workflow for merge queue

jakebailey:
This is a little tricky but I think works.

Basically, just skip all of the macOS jobs (runners are in short supply) and Windows jobs (slow). Running the tests generally speaking is slow enough that we can do basically everything else.


andrewbranch (approved):

jakebailey:
I hit the merge button to watch the run in action; race mode in particular takes longer than I'd like, so I'm going to disable that check.

https://github.com/microsoft/typescript-go/actions/runs/14451205530/job/40524114981

PR #791, opened 2025-04-11
`createNodeBuilder` and associated utility port

weswigham:
TODO, annotated with current thoughts on each:
* [ ] Mapped type node creation (Generates tons of custom scopes and symbols, so useless without other in-progress work anyway. Also kinda just lost it among the symbol functions 'til I went back and surveyed what was still missing. Might hop back to this and get it in so at least all the type structures are done.) 
* [ ] JSDoc (mostly in ID related content, but a little bit in js declaration and ts @overload content)
* [ ] JS Declarations more generally (`symbolTableToDeclarationStatements` and friends)
* [ ] Synthetic comments (scattered blocks of functionality throughout the node builder to attach comments to the output - mostly for quickfixes)
* [ ] String quote style preservation (a recent TS feature is preserving single vs double quotes through the emit pipeline, but some API changes precluded initial support)
* [ ] Isolated declarations (`expressionToTypeNode.ts`, plus some wrappers)
  * [ ] Contained node reuse logic also used by non-ID, so also needed for good declaration map support now that sourcemaps are in (but maybe we unwind the extraction to a separate syntactic resolver object to also fix all the bugs the extraction introduced?)
* [ ] Symbol name construction (`symbolToNode` and other flavors - lots of very similar but subtly different functions with little corner cases for services support)
  * [ ] Module specifier generation (`moduleSpecifiers.ts`) <-- current work item, looking at generating this more automatically with our tooling since it's fairly standalone and doesn't interact with many refactored APIs
  * [ ] Entity name accessibility checking and chain generation (`isSymbolAccessible`, `getAccessibleSymbolChain` and the like - these are unimplemented stubs right now, for want of the above, and are a bit unwieldy - also likely candidates for a perf rewrite at some point in the future, as the historically slow part of declaration emit)
 * [ ] Any unported factory/checker/utils that those depend on implicitly that I haven't run into yet (insert mining meme here)
 * [ ] Updates from expandable hover after that merges to strada (notably, lots of scattered approximatelength additions that may not show up as any test failures if not ported)
 * [ ] Make `singlelinetextwriter` usage threadsafe to support concurrent checking (likely one writer per checker instead of a global)
 * [ ] Factor all the node building logic into a subfolder so the growing set of files/loc is less painful on the compile times of checker (probably do this very last)
 * [ ] Allow threading through `EmitContext` to individual factory calls (via symbol tracker?) so the context used by the declaration transform can be correctly used within the node builder, while it still falls back to the checker's internal emit context for diagnostic production or quickfixes and the like that don't need their own emit context
 * [ ] Memoize printer setup per-checker, rather than making a new printer on each builder request to stay threadsafe (maybe the checker should have its' own emit context wrapper to track emit/print related state like this?)
   * [ ] Investigate if `OmitTrailingSemicolon` printer option is redundant with the trailing semicolon eliding wrapper-writer we use in strada - only one of these two may need to be ported
 * [ ] Reintroduce some way to smuggle signature type arguments out of the node builder for quickinfo (previously we patched a `.typeArguments` onto every signature kind)
 * [ ] Actual `declarations.ts` transform once those are all (or mostly) in
   * [ ] TS declarations
   * [ ] JS declarations
   * [ ] Isolated declarations 

I've updated my branch to replace the current `typeToString` implementation with the ported node-builder backed implementation, which is the top-down goal of this PR, ultimately. CI on it should not pass until the node builder is mostly (if not totally) completely ported, due to interdependencies between many of its' subsystems. I'll try to keep this PR in a state where it at least compiles without error, which is usually in between big chunks of functionality, but the tests are going to keep throwing `unimplemented` panics until all the missing functions are either fully ported or at least sufficiently stubbed (which doesn't feel worth it, really).

Structurally, `nodebuilder.go` is the inner contents of `createNodeBuilder`'s closure environment, lifted into members of a struct. All `context` parameters have been removed and replaced with lookups of `context` on the struct itself, since we're OO now, but that's about the only refactor. `nodebuilderapi.go` is the wrapper returned by `createNodeBuilder` which maps all the internal closed over methods to the internal node builder API shape (which is recorded as the `NodeBuilderInterface`... interface)- basically it's the logic that handles setting up `context` objects for each request. Some of that might get renamed to reduce confusion eventually, but the structure seems sound. `nodebuilderscopes.go` may or may not go away - `NodeBuilder.enterNewScope` was pretty big but isolated (used by mapped type and signature construction), so felt like it could stand alone, and it has some utilities only it uses. `symbolaccessibility.go` is for the checker's symbol accessibility functionality - these are also mostly self-contained, though do depend on one-another and some common utilities (though I only have stubs here right now - my previous attempts to optimize them as I ported them have broken them, so we're just gonna port them straight as we can for now).

This is already a bit of a beast to review, size-wise, and I'd say there's still a fair bit left for a full port - but if we add some extra unit tests, some subsystems, like the specifier generation and maybe accessibility, *can* reasonably stand alone as changes. Those things just aren't currently unit tested outside of their integrations into the builder in strada, though, so those tests'd all be additional greenfield work.

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -1435,10 +1435,10 @@ func IsDynamicName(name *Node) bool {
 }
 
 func IsEntityNameExpression(node *Node) bool {
-	return node.Kind == KindIdentifier || isPropertyAccessEntityNameExpression(node)
+	return node.Kind == KindIdentifier || IsPropertyAccessEntityNameExpression(node)
 }
 
-func isPropertyAccessEntityNameExpression(node *Node) bool {
+func IsPropertyAccessEntityNameExpression(node *Node) bool {
 	if node.Kind == KindPropertyAccessExpression {
 		expr := node.AsPropertyAccessExpression()
 		return expr.Name().Kind == KindIdentifier && IsEntityNameExpression(expr.Expression)
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -821,6 +821,7 @@ type Checker struct {
 	markNodeAssignments                        func(*ast.Node) bool
 	emitResolver                               *emitResolver
 	emitResolverOnce                           sync.Once
+	nodeBuilder                                NodeBuilderInterface
 	_jsxNamespace                              string
 	_jsxFactoryEntity                          *ast.Node
 }
@@ -1027,6 +1028,8 @@ func NewChecker(program Program) *Checker {
 	c.getGlobalClassAccessorDecoratorTargetType = c.getGlobalTypeResolver("ClassAccessorDecoratorTarget", 2 /*arity*/, true /*reportErrors*/)
 	c.getGlobalClassAccessorDecoratorResultType = c.getGlobalTypeResolver("ClassAccessorDecoratorResult", 2 /*arity*/, true /*reportErrors*/)
 	c.getGlobalClassFieldDecoratorContextType = c.getGlobalTypeResolver("ClassFieldDecoratorContext", 2 /*arity*/, true /*reportErrors*/)
+	diagnosticConstructionContext := printer.NewEmitContext()
+	c.nodeBuilder = NewNodeBuilderAPI(c, diagnosticConstructionContext)
 	c.initializeClosures()
 	c.initializeIterationResolvers()
 	c.initializeChecker()
@@ -17923,7 +17926,7 @@ func (c *Checker) areAllOuterTypeParametersApplied(t *Type) bool {
 }
 
 func (c *Checker) reportCircularBaseType(node *ast.Node, t *Type) {
-	c.error(node, diagnostics.Type_0_recursively_references_itself_as_a_base_type, c.typeToStringEx(t, nil, TypeFormatFlagsWriteArrayAsGenericType))
+	c.error(node, diagnostics.Type_0_recursively_references_itself_as_a_base_type, c.typeToStringEx(t, nil, TypeFormatFlagsWriteArrayAsGenericType, nil))
 }
 
 // A valid base type is `any`, an object type or intersection of object types.
@@ -20196,11 +20199,11 @@ func (c *Checker) elaborateNeverIntersection(chain *ast.Diagnostic, node *ast.No
 	if t.flags&TypeFlagsIntersection != 0 && t.objectFlags&ObjectFlagsIsNeverIntersection != 0 {
 		neverProp := core.Find(c.getPropertiesOfUnionOrIntersectionType(t), c.isDiscriminantWithNeverType)
 		if neverProp != nil {
-			return NewDiagnosticChainForNode(chain, node, diagnostics.The_intersection_0_was_reduced_to_never_because_property_1_has_conflicting_types_in_some_constituents, c.typeToStringEx(t, nil, TypeFormatFlagsNoTypeReduction), c.symbolToString(neverProp))
+			return NewDiagnosticChainForNode(chain, node, diagnostics.The_intersection_0_was_reduced_to_never_because_property_1_has_conflicting_types_in_some_constituents, c.typeToStringEx(t, nil, TypeFormatFlagsNoTypeReduction, nil), c.symbolToString(neverProp))
 		}
 		privateProp := core.Find(c.getPropertiesOfUnionOrIntersectionType(t), isConflictingPrivateProperty)
 		if privateProp != nil {
-			return NewDiagnosticChainForNode(chain, node, diagnostics.The_intersection_0_was_reduced_to_never_because_property_1_exists_in_multiple_constituents_and_is_private_in_some, c.typeToStringEx(t, nil, TypeFormatFlagsNoTypeReduction), c.symbolToString(privateProp))
+			return NewDiagnosticChainForNode(chain, node, diagnostics.The_intersection_0_was_reduced_to_never_because_property_1_exists_in_multiple_constituents_and_is_private_in_some, c.typeToStringEx(t, nil, TypeFormatFlagsNoTypeReduction, nil), c.symbolToString(privateProp))
 		}
 	}
 	return chain
--- a/internal/checker/nodebuilder.go
+++ b/internal/checker/nodebuilder.go
@@ -1,0 +1,1927 @@
+package checker
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/jsnum"
+	"github.com/microsoft/typescript-go/internal/printer"
+	"github.com/microsoft/typescript-go/internal/scanner"
+)
+
+// NOTE: If modifying this enum, must modify `TypeFormatFlags` too!
+// dprint-ignore
+type NodeBuilderFlags int32
+
+const (
+	NodeBuilderFlagsNone NodeBuilderFlags = 0
+	// Options
+	NodeBuilderFlagsNoTruncation                        NodeBuilderFlags = 1 << 0
+	NodeBuilderFlagsWriteArrayAsGenericType             NodeBuilderFlags = 1 << 1
+	NodeBuilderFlagsGenerateNamesForShadowedTypeParams  NodeBuilderFlags = 1 << 2
+	NodeBuilderFlagsUseStructuralFallback               NodeBuilderFlags = 1 << 3
+	NodeBuilderFlagsForbidIndexedAccessSymbolReferences NodeBuilderFlags = 1 << 4
+	NodeBuilderFlagsWriteTypeArgumentsOfSignature       NodeBuilderFlags = 1 << 5
+	NodeBuilderFlagsUseFullyQualifiedType               NodeBuilderFlags = 1 << 6
+	NodeBuilderFlagsUseOnlyExternalAliasing             NodeBuilderFlags = 1 << 7
+	NodeBuilderFlagsSuppressAnyReturnType               NodeBuilderFlags = 1 << 8
+	NodeBuilderFlagsWriteTypeParametersInQualifiedName  NodeBuilderFlags = 1 << 9
+	NodeBuilderFlagsMultilineObjectLiterals             NodeBuilderFlags = 1 << 10
+	NodeBuilderFlagsWriteClassExpressionAsTypeLiteral   NodeBuilderFlags = 1 << 11
+	NodeBuilderFlagsUseTypeOfFunction                   NodeBuilderFlags = 1 << 12
+	NodeBuilderFlagsOmitParameterModifiers              NodeBuilderFlags = 1 << 13
+	NodeBuilderFlagsUseAliasDefinedOutsideCurrentScope  NodeBuilderFlags = 1 << 14
+	NodeBuilderFlagsUseSingleQuotesForStringLiteralType NodeBuilderFlags = 1 << 28
+	NodeBuilderFlagsNoTypeReduction                     NodeBuilderFlags = 1 << 29
+	NodeBuilderFlagsOmitThisParameter                   NodeBuilderFlags = 1 << 25
+	// Error handling
+	NodeBuilderFlagsAllowThisInObjectLiteral              NodeBuilderFlags = 1 << 15
+	NodeBuilderFlagsAllowQualifiedNameInPlaceOfIdentifier NodeBuilderFlags = 1 << 16
+	NodeBuilderFlagsAllowAnonymousIdentifier              NodeBuilderFlags = 1 << 17
+	NodeBuilderFlagsAllowEmptyUnionOrIntersection         NodeBuilderFlags = 1 << 18
+	NodeBuilderFlagsAllowEmptyTuple                       NodeBuilderFlags = 1 << 19
+	NodeBuilderFlagsAllowUniqueESSymbolType               NodeBuilderFlags = 1 << 20
+	NodeBuilderFlagsAllowEmptyIndexInfoType               NodeBuilderFlags = 1 << 21
+	// Errors (cont.)
+	NodeBuilderFlagsAllowNodeModulesRelativePaths NodeBuilderFlags = 1 << 26
+	NodeBuilderFlagsIgnoreErrors                  NodeBuilderFlags = NodeBuilderFlagsAllowThisInObjectLiteral | NodeBuilderFlagsAllowQualifiedNameInPlaceOfIdentifier | NodeBuilderFlagsAllowAnonymousIdentifier | NodeBuilderFlagsAllowEmptyUnionOrIntersection | NodeBuilderFlagsAllowEmptyTuple | NodeBuilderFlagsAllowEmptyIndexInfoType | NodeBuilderFlagsAllowNodeModulesRelativePaths
+	// State
+	NodeBuilderFlagsInObjectTypeLiteral NodeBuilderFlags = 1 << 22
+	NodeBuilderFlagsInTypeAlias         NodeBuilderFlags = 1 << 23
+	NodeBuilderFlagsInInitialEntityName NodeBuilderFlags = 1 << 24
+)
+
+/** @internal */
+// dprint-ignore
+
+type InternalNodeBuilderFlags int32
+
+const (
+	InternalNodeBuilderFlagsNone                    InternalNodeBuilderFlags = 0
+	InternalNodeBuilderFlagsWriteComputedProps      InternalNodeBuilderFlags = 1 << 0
+	InternalNodeBuilderFlagsNoSyntacticPrinter      InternalNodeBuilderFlags = 1 << 1
+	InternalNodeBuilderFlagsDoNotIncludeSymbolChain InternalNodeBuilderFlags = 1 << 2
+	InternalNodeBuilderFlagsAllowUnresolvedNames    InternalNodeBuilderFlags = 1 << 3
+)
+
+type CompositeSymbolIdentity struct {
+	isConstructorNode bool
+	symbolId          ast.SymbolId
+	nodeId            ast.NodeId
+}
+
+type TrackedSymbolArgs struct {
+	symbol               *ast.Symbol
+	enclosingDeclaration *ast.Node
+	meaning              ast.SymbolFlags
+}
+
+type SerializedTypeEntry struct {
+	node           *ast.Node
+	truncating     bool
+	addedLength    int
+	trackedSymbols []*TrackedSymbolArgs
+}
+
+type CompositeTypeCacheIdentity struct {
+	typeId        TypeId
+	flags         NodeBuilderFlags
+	internalFlags InternalNodeBuilderFlags
+}
+
+type NodeBuilderLinks struct {
+	serializedTypes                  map[CompositeTypeCacheIdentity]*SerializedTypeEntry // Collection of types serialized at this location
+	fakeScopeForSignatureDeclaration *string                                             // If present, this is a fake scope injected into an enclosing declaration chain.
+}
+
+type NodeBuilderContext struct {
+	tracker                         SymbolTracker
+	approximateLength               int
+	encounteredError                bool
+	truncating                      bool
+	reportedDiagnostic              bool
+	flags                           NodeBuilderFlags
+	internalFlags                   InternalNodeBuilderFlags
+	depth                           int
+	enclosingDeclaration            *ast.Node
+	enclosingFile                   *ast.SourceFile
+	inferTypeParameters             []*Type
+	visitedTypes                    map[TypeId]bool
+	symbolDepth                     map[CompositeSymbolIdentity]int
+	trackedSymbols                  []*TrackedSymbolArgs
+	mapper                          *TypeMapper
+	reverseMappedStack              []*ast.Symbol
+	enclosingSymbolTypes            map[ast.SymbolId]*Type
+	suppressReportInferenceFallback bool
+
+	// per signature scope state
+	mustCreateTypeParameterSymbolList     bool
+	mustCreateTypeParametersNamesLookups  bool
+	typeParameterNames                    any
+	typeParameterNamesByText              any
+	typeParameterNamesByTextNextNameCount any
+	typeParameterSymbolList               any
+}
+
+type NodeBuilder struct {
+	// host members
+	f  *ast.NodeFactory
+	ch *Checker
+	e  *printer.EmitContext
+
+	// cache
+	links core.LinkStore[*ast.Node, NodeBuilderLinks]
+
+	// closures
+	typeToTypeNodeClosure               func(t *Type) *ast.TypeNode
+	typeReferenceToTypeNodeClosure      func(t *Type) *ast.TypeNode
+	conditionalTypeToTypeNodeClosure    func(t *Type) *ast.TypeNode
+	createTypeNodeFromObjectTypeClosure func(t *Type) *ast.TypeNode
+	isStringNamedClosure                func(d *ast.Declaration) bool
+	isSingleQuotedStringNamedClosure    func(d *ast.Declaration) bool
+
+	// state
+	ctx *NodeBuilderContext
+}
+
+const defaultMaximumTruncationLength = 160
+const noTruncationMaximumTruncationLength = 1_000_000
+
+// Node builder utility functions
+
+// You probably don't mean to use this - use `NewNodeBuilderAPI` instead
+func NewNodeBuilder(ch *Checker, e *printer.EmitContext) NodeBuilder {
+	result := NodeBuilder{f: e.Factory, ch: ch, e: e, typeToTypeNodeClosure: nil, typeReferenceToTypeNodeClosure: nil, conditionalTypeToTypeNodeClosure: nil, ctx: nil}
+	result.initializeClosures()
+	return result
+}
+
+func (b *NodeBuilder) initializeClosures() {
+	b.typeToTypeNodeClosure = b.typeToTypeNode
+	b.typeReferenceToTypeNodeClosure = b.typeReferenceToTypeNode
+	b.conditionalTypeToTypeNodeClosure = b.conditionalTypeToTypeNode
+	b.createTypeNodeFromObjectTypeClosure = b.createTypeNodeFromObjectType
+	b.isStringNamedClosure = b.isStringNamed
+	b.isSingleQuotedStringNamedClosure = b.isSingleQuotedStringNamed
+}
+
+func (b *NodeBuilder) saveRestoreFlags() func() {
+	flags := b.ctx.flags
+	internalFlags := b.ctx.internalFlags
+	depth := b.ctx.depth
+
+	return func() {
+		b.ctx.flags = flags
+		b.ctx.internalFlags = internalFlags
+		b.ctx.depth = depth
+	}
+}
+
+func (b *NodeBuilder) checkTruncationLength() bool {
+	if b.ctx.truncating {
+		return b.ctx.truncating
+	}
+	b.ctx.truncating = b.ctx.approximateLength > (core.IfElse((b.ctx.flags&NodeBuilderFlagsNoTruncation != 0), noTruncationMaximumTruncationLength, defaultMaximumTruncationLength))
+	return b.ctx.truncating
+}
+
+func (b *NodeBuilder) appendReferenceToType(root *ast.TypeNode, ref *ast.TypeNode) *ast.TypeNode {
+	if ast.IsImportTypeNode(root) {
+		// first shift type arguments
+
+		// !!! In the old emitter, an Identifier could have type arguments for use with quickinfo:
+		// typeArguments := root.TypeArguments
+		// qualifier := root.AsImportTypeNode().Qualifier
+		// if qualifier != nil {
+		// 	if ast.IsIdentifier(qualifier) {
+		// 		if typeArguments != getIdentifierTypeArguments(qualifier) {
+		// 			qualifier = setIdentifierTypeArguments(b.f.CloneNode(qualifier), typeArguments)
+		// 		}
+		// 	} else {
+		// 		if typeArguments != getIdentifierTypeArguments(qualifier.Right) {
+		// 			qualifier = b.f.UpdateQualifiedName(qualifier, qualifier.Left, setIdentifierTypeArguments(b.f.cloneNode(qualifier.Right), typeArguments))
+		// 		}
+		// 	}
+		// }
+		// !!! Without the above, nested type args are silently elided
+		imprt := root.AsImportTypeNode()
+		// then move qualifiers
+		ids := getAccessStack(ref)
+		var qualifier *ast.Node
+		for _, id := range ids {
+			if qualifier != nil {
+				qualifier = b.f.NewQualifiedName(qualifier, id)
+			} else {
+				qualifier = id
+			}
+		}
+		return b.f.UpdateImportTypeNode(imprt, imprt.IsTypeOf, imprt.Argument, imprt.Attributes, qualifier, ref.AsTypeReferenceNode().TypeArguments)
+	} else {
+		// first shift type arguments
+		// !!! In the old emitter, an Identifier could have type arguments for use with quickinfo:
+		// typeArguments := root.TypeArguments
+		// typeName := root.AsTypeReferenceNode().TypeName
+		// if ast.IsIdentifier(typeName) {
+		// 	if typeArguments != getIdentifierTypeArguments(typeName) {
+		// 		typeName = setIdentifierTypeArguments(b.f.cloneNode(typeName), typeArguments)
+		// 	}
+		// } else {
+		// 	if typeArguments != getIdentifierTypeArguments(typeName.Right) {
+		// 		typeName = b.f.UpdateQualifiedName(typeName, typeName.Left, setIdentifierTypeArguments(b.f.cloneNode(typeName.Right), typeArguments))
+		// 	}
+		// }
+		// !!! Without the above, nested type args are silently elided
+		// then move qualifiers
+		ids := getAccessStack(ref)
+		var typeName *ast.Node = root.AsTypeReferenceNode().TypeName
+		for _, id := range ids {
+			typeName = b.f.NewQualifiedName(typeName, id)
+		}
+		return b.f.UpdateTypeReferenceNode(root.AsTypeReferenceNode(), typeName, ref.AsTypeReferenceNode().TypeArguments)
+	}
+}
+
+func getAccessStack(ref *ast.Node) []*ast.Node {
+	var state *ast.Node = ref.AsTypeReferenceNode().TypeName
+	ids := []*ast.Node{}
+	for !ast.IsIdentifier(state) {
+		entity := state.AsQualifiedName()
+		ids = append([]*ast.Node{entity.Right}, ids...)
+		state = entity.Left
+	}
+	ids = append([]*ast.Node{state}, ids...)
+	return ids
+}
+
+func isClassInstanceSide(c *Checker, t *Type) bool {
+	return t.symbol != nil && t.symbol.Flags&ast.SymbolFlagsClass != 0 && (t == c.getDeclaredTypeOfClassOrInterface(t.symbol) || (t.flags&TypeFlagsObject != 0 && t.objectFlags&ObjectFlagsIsClassInstanceClone != 0))
+}
+
+func (b *NodeBuilder) createElidedInformationPlaceholder() *ast.TypeNode {
+	b.ctx.approximateLength += 3
+	if b.ctx.flags&NodeBuilderFlagsNoTruncation == 0 {
+		return b.f.NewTypeReferenceNode(b.f.NewIdentifier("..."), nil /*typeArguments*/)
+	}
+	// addSyntheticLeadingComment(b.f.NewKeywordTypeNode(ast.KindAnyKeyword), ast.KindMultiLineCommentTrivia, "elided") // !!!
+	return b.f.NewKeywordTypeNode(ast.KindAnyKeyword)
+}
+
+func (b *NodeBuilder) mapToTypeNodes(list []*Type) *ast.NodeList {
+	if len(list) == 0 {
+		return nil
+	}
+	contents := core.Map(list, b.typeToTypeNodeClosure)
+	return b.f.NewNodeList(contents)
+}
+
+func (b *NodeBuilder) setCommentRange(node *ast.Node, range_ *ast.Node) {
+	if range_ != nil && b.ctx.enclosingFile != nil && b.ctx.enclosingFile == ast.GetSourceFileOfNode(range_) {
+		// Copy comments to node for declaration emit
+		b.e.AssignCommentRange(node, range_)
+	}
+}
+
+func (b *NodeBuilder) tryReuseExistingTypeNodeHelper(existing *ast.TypeNode) *ast.TypeNode {
+	return nil // !!!
+}
+
+func containsNonMissingUndefinedType(c *Checker, t *Type) bool {
+	var candidate *Type
+	if t.flags&TypeFlagsUnion != 0 {
+		candidate = t.AsUnionType().types[0]
+	} else {
+		candidate = t
+	}
+	return candidate.flags&TypeFlagsUndefined != 0 && candidate != c.missingType
+}
+
+func (b *NodeBuilder) tryReuseExistingTypeNode(typeNode *ast.TypeNode, t *Type, host *ast.Node, addUndefined bool) *ast.TypeNode {
+	originalType := t
+	if addUndefined {
+		t = b.ch.getOptionalType(t, !ast.IsParameter(host))
+	}
+	clone := b.tryReuseExistingNonParameterTypeNode(typeNode, t, host, nil)
+	if clone != nil {
+		// explicitly add `| undefined` if it's missing from the input type nodes and the type contains `undefined` (and not the missing type)
+		if addUndefined && containsNonMissingUndefinedType(b.ch, t) && !someType(b.getTypeFromTypeNode(typeNode, false), func(t *Type) bool {
+			return t.flags&TypeFlagsUndefined != 0
+		}) {
+			return b.f.NewUnionTypeNode(b.f.NewNodeList([]*ast.TypeNode{clone, b.f.NewKeywordTypeNode(ast.KindUndefinedKeyword)}))
+		}
+		return clone
+	}
+	if addUndefined && originalType != t {
+		cloneMissingUndefined := b.tryReuseExistingNonParameterTypeNode(typeNode, originalType, host, nil)
+		if cloneMissingUndefined != nil {
+			return b.f.NewUnionTypeNode(b.f.NewNodeList([]*ast.TypeNode{cloneMissingUndefined, b.f.NewKeywordTypeNode(ast.KindUndefinedKeyword)}))
+		}
+	}
+	return nil
+}
+
+func (b *NodeBuilder) typeNodeIsEquivalentToType(annotatedDeclaration *ast.Node, t *Type, typeFromTypeNode *Type) bool {
+	if typeFromTypeNode == t {
+		return true
+	}
+	if annotatedDeclaration == nil {
+		return false
+	}
+	// !!!
+	// used to be hasEffectiveQuestionToken for JSDoc
+	if isOptionalDeclaration(annotatedDeclaration) {
+		return b.ch.getTypeWithFacts(t, TypeFactsNEUndefined) == typeFromTypeNode
+	}
+	return false
+}
+
+func (b *NodeBuilder) existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount(existing *ast.TypeNode, t *Type) bool {
+	// In JS, you can say something like `Foo` and get a `Foo<any>` implicitly - we don't want to preserve that original `Foo` in these cases, though.
+	if t.objectFlags&ObjectFlagsReference == 0 {
+		return true
+	}
+	if !ast.IsTypeReferenceNode(existing) {
+		return true
+	}
+	// `type` is a reference type, and `existing` is a type reference node, but we still need to make sure they refer to the _same_ target type
+	// before we go comparing their type argument counts.
+	b.ch.getTypeFromTypeReference(existing)
+	// call to ensure symbol is resolved
+	links := b.ch.symbolNodeLinks.TryGet(existing)
+	if links == nil {
+		return true
+	}
+	symbol := links.resolvedSymbol
+	if symbol == nil {
+		return true
+	}
+	existingTarget := b.ch.getDeclaredTypeOfSymbol(symbol)
+	if existingTarget == nil || existingTarget != t.AsTypeReference().target {
+		return true
+	}
+	return len(existing.TypeArguments()) >= b.ch.getMinTypeArgumentCount(t.AsTypeReference().target.AsInterfaceType().TypeParameters())
+}
+
+func (b *NodeBuilder) tryReuseExistingNonParameterTypeNode(existing *ast.TypeNode, t *Type, host *ast.Node, annotationType *Type) *ast.TypeNode {
+	if host == nil {
+		host = b.ctx.enclosingDeclaration
+	}
+	if annotationType == nil {
+		annotationType = b.getTypeFromTypeNode(existing, true)
+	}
+	if annotationType != nil && b.typeNodeIsEquivalentToType(host, t, annotationType) && b.existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount(existing, t) {
+		result := b.tryReuseExistingTypeNodeHelper(existing)
+		if result != nil {
+			return result
+		}
+	}
+	return nil
+}
+
+func (b *NodeBuilder) getResolvedTypeWithoutAbstractConstructSignatures(t *StructuredType) *Type {
+	if len(t.ConstructSignatures()) == 0 {
+		return t.AsType()
+	}
+	if t.objectTypeWithoutAbstractConstructSignatures != nil {
+		return t.objectTypeWithoutAbstractConstructSignatures
+	}
+	constructSignatures := core.Filter(t.ConstructSignatures(), func(signature *Signature) bool {
+		return signature.flags&SignatureFlagsAbstract == 0
+	})
+	if len(constructSignatures) == len(t.ConstructSignatures()) {
+		t.objectTypeWithoutAbstractConstructSignatures = t.AsType()
+		return t.AsType()
+	}
+	typeCopy := b.ch.newAnonymousType(t.symbol, t.members, t.CallSignatures(), core.IfElse(len(constructSignatures) > 0, constructSignatures, []*Signature{}), t.indexInfos)
+	t.objectTypeWithoutAbstractConstructSignatures = typeCopy
+	typeCopy.AsStructuredType().objectTypeWithoutAbstractConstructSignatures = typeCopy
+	return typeCopy
+}
+
+func (b *NodeBuilder) symbolToNode(symbol *ast.Symbol, meaning ast.SymbolFlags) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToName(symbol *ast.Symbol, meaning ast.SymbolFlags, false bool) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToEntityNameNode(symbol *ast.Symbol) *ast.EntityName {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToTypeNode(symbol *ast.Symbol, mask ast.SymbolFlags, typeArguments *ast.NodeList) *ast.TypeNode {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToExpression(symbol *ast.Symbol, mask ast.SymbolFlags) *ast.Expression {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) typeParameterToDeclarationWithConstraint(typeParameter *Type, constraintNode *ast.TypeNode) *ast.TypeParameterDeclarationNode {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) typeParameterToName(typeParameter *Type) *ast.Identifier {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) createMappedTypeNodeFromType(type_ *Type) *ast.TypeNode {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) typePredicateToTypePredicateNode(predicate *TypePredicate) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) typeParameterToDeclaration(parameter *Type) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToTypeParameterDeclarations(symbol *ast.Symbol) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolToParameterDeclaration(symbol *ast.Symbol, preserveModifierFlags bool) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) symbolTableToDeclarationStatements(symbolTable *ast.SymbolTable) []*ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) serializeTypeForExpression(expr *ast.Node) *ast.Node {
+	panic("unimplemented") // !!!
+}
+
+func (b *NodeBuilder) serializeInferredReturnTypeForSignature(signature *Signature, returnType *Type) *ast.Node {
+	oldSuppressReportInferenceFallback := b.ctx.suppressReportInferenceFallback
+	b.ctx.suppressReportInferenceFallback = true
+	typePredicate := b.ch.getTypePredicateOfSignature(signature)
+	var returnTypeNode *ast.Node
+	if typePredicate != nil {
+		var predicate *TypePredicate
+		if b.ctx.mapper != nil {
+			predicate = b.ch.instantiateTypePredicate(typePredicate, b.ctx.mapper)
+		} else {
+			predicate = typePredicate
+		}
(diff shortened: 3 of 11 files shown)


internal/printer/singlelinestringwriter.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/stringutil"
 )
 
+// TODO: Definitely not threadsafe - make one per checker instead of one global one (threadlocal storage would be neat)
```
jakebailey: Yeah, a `sync.Pool` of these would be best.

jakebailey (commented):
I'll be interested to see if we can move any of this out of the checker package; it's getting pretty big in there.

(I'm not sure how much you want eyes on the contents of the PR yet but I'm happy to look if you do.)

weswigham:
I have https://github.com/weswigham/typescript-go/pull/1 separately open with the baseline changes, because if I merge those into this PR, this change goes from "practically unreviewable" to "actually impossible to navigate in the github web UI", so it's probably best if reviews of this are done before any baseline updates get merged in.

weswigham:
> I'm not sure how much you want eyes on the contents of the PR yet but I'm happy to look if you do

At this point, eyes are welcome, since we could probably merge this any time we wanted given the completeness bar for corsa right now - full stack declaration emit is functional and non-panicing on our test suite, despite some missing features (js support, for example). I've already worked through some of the more egregious bugs and differences from the initial version, too. If you wanted to look at this + baselines, there's https://github.com/weswigham/typescript-go/pull/1, but I can't recommend opening that in the github web UI.

internal/checker/symbolaccessibility.go
```diff
@@ -0,0 +1,762 @@
...
+		if symbolFromSymbolTable.Name == ctx.symbol.Name && symbolFromSymbolTable.ExportSymbol != nil {
+			if ch.isAccessible(ctx, ch.getMergedSymbol(symbolFromSymbolTable.ExportSymbol) /*resolvedAliasSymbol*/, nil, ignoreQualification) {
+				candidateChains = append(candidateChains, []*ast.Symbol{ctx.symbol})
+			}
+		}
+	}
+
+	if len(candidateChains) > 0 {
+		// pick first, shortest
+		slices.SortStableFunc(candidateChains, ch.compareSymbolChains)
```
weswigham, line 508: I feel obligated to point out that this logic - exhaustively traversing symbol tables looking for a way to name a symbol - is the most expensive part of declaration emit, and here in the `go` port, it is *always* worst case performance if anything is exported through a renaming alias and thus dodges the direct lookup fast path - we can't early bail on finding a matching symbol because of the need to sort the results for stability.

So this logic is *incredibly* slow. You *will* notice how much slower tests are with this PR (though to some degree that was guaranteed!). This stuff has been a candidate for a rewrite for approximately forever - it dates all the way back to the old string-based declaration emitter logic, and has just grown in weird preference-encoding and edge-case-handling ever since. Making a more efficient equivalent that behaves the same way is almost impossible (I've tried), but the randomly iterated and-then-sorted symbol tables are already potentially going to break strict equivalence with old output (since the symbols we select will first be sorted by declaration location and alphabetically according to the new symbol sort algorithm, and not by creation time). Given that, this whole suite of functions (and wrappers in the node builder) is ripe for a rewrite at some point, while we're here breaking things.

weswigham:
The baselines at https://github.com/weswigham/typescript-go/pull/1 have something like 2026 deleted `.types.diff` and `.js.diff` files, so there's certainly some amount of progress here~

internal/ast/utilities.go
```diff
@@ -2801,3 +2789,332 @@ func ForEachChildAndJSDoc(node *Node, sourceFile *SourceFile, v Visitor) bool {
 	}
 	return node.ForEachChild(v)
 }
+
+func GetExternalModuleImportEqualsDeclarationExpression(node *Node) *Node {
+	// Debug.assert(isExternalModuleImportEqualsDeclaration(node))
```
gabritto, line 2953: Any reason why this isn't an actual assertion/panic?

internal/ast/utilities.go
```diff
@@ -2801,3 +2789,332 @@ func ForEachChildAndJSDoc(node *Node, sourceFile *SourceFile, v Visitor) bool {
...
+		result = append(result, createModifier(KindInKeyword))
+	}
+	if flags&ModifierFlagsOut != 0 {
+		result = append(result, createModifier(KindOutKeyword))
+	}
+	return result
+}
+
+func GetThisParameter(signature *Node) *Node {
+	// callback tags do not currently support this parameters
```
gabritto, line 3008: Is this comment still relevant in light of the new approach to JSDoc?

internal/checker/checker.go
```diff
@@ -5192,9 +5204,7 @@ func (c *Checker) checkExportSpecifier(node *ast.Node) {
 	hasModuleSpecifier := node.Parent.Parent.AsExportDeclaration().ModuleSpecifier != nil
 	c.checkModuleExportName(node.AsExportSpecifier().PropertyName, hasModuleSpecifier)
 	c.checkModuleExportName(node.Name(), true /*allowStringLiteral*/)
-	if c.compilerOptions.GetEmitDeclarations() {
```
gabritto: Is the emit resolver responsible for doing this now?

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+	declarationFileLinks    core.LinkStore[*ast.Node, DeclarationFileLinks]
+}
+
+func (r *emitResolver) IsOptionalParameter(node *ast.Node) bool {
+	return r.isOptionalParameter(node)
+}
+
+func (r *emitResolver) IsLateBound(node *ast.Node) bool {
+	// TODO: Require an emitContext to construct an EmitResolver, remove all emitContext arguments
+	// node = r.emitContext.ParseNode(node)
```
gabritto, line 46: Why do we need to do this?

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+
+	r.checker.computeEnumMemberValues(node.Parent)
+	if !r.checker.enumMemberLinks.Has(node) {
+		return evaluator.NewResult(nil, false, false, false)
+	}
+	return r.checker.enumMemberLinks.Get(node).value
+}
+
+func (r *emitResolver) IsDeclarationVisible(node *ast.Node) bool {
+	// Only lock on external API func to prevent deadlocks
```
gabritto, line 78: How is this part of a deadlock?

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+		panic("Node cannot possibly require adding undefined")
+	}
+}
+
+func (r *emitResolver) requiresAddingImplicitUndefined(parameter *ast.Node, enclosingDeclaration *ast.Node) bool {
+	return (r.isRequiredInitializedParameter(parameter, enclosingDeclaration) || r.isOptionalUninitializedParameterProperty(parameter)) && !r.declaredParameterTypeContainsUndefined(parameter)
+}
+
+func (r *emitResolver) declaredParameterTypeContainsUndefined(parameter *ast.Node) bool {
+	// typeNode := getNonlocalEffectiveTypeAnnotationNode(parameter); // !!! JSDoc Support
```
gabritto, line 501: I thought we wouldn't need this anymore given the new approach to JSDoc. I thought a JSDoc annotation would become a synthetic type node on the parameter, but I might be wrong.

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+		return false
+	}
+	if ast.HasSyntacticModifier(parameter, ast.ModifierFlagsParameterPropertyModifier) {
+		return enclosingDeclaration != nil && ast.IsFunctionLikeDeclaration(enclosingDeclaration)
+	}
+	return true
+}
+
+func (r *emitResolver) isOptionalParameter(node *ast.Node) bool {
+	// !!! TODO: JSDoc support
```
gabritto, line 531: Same as above, I don't think we'll need this.

internal/checker/emitresolver.go
```diff
@@ -225,3 +805,150 @@ func (r *emitResolver) GetReferencedValueDeclarations(node *ast.IdentifierNode)
 
 	return r.getReferenceResolver().GetReferencedValueDeclarations(node)
 }
+
+// TODO: the emit resolver being respoinsible for some amount of node construction is a very leaky abstraction,
```
gabritto:
```suggestion
// TODO: the emit resolver being responsible for some amount of node construction is a very leaky abstraction,
```

internal/checker/emitresolver.go
```diff
@@ -225,3 +805,150 @@ func (r *emitResolver) GetReferencedValueDeclarations(node *ast.IdentifierNode)
...
+	requestNodeBuilder := NewNodeBuilderAPI(r.checker, emitContext) // TODO: cache per-context
+	// // Get type of the symbol if this is the valid symbol otherwise get type at location
+	symbol := r.checker.getSymbolOfDeclaration(declaration)
+	return requestNodeBuilder.SerializeTypeForDeclaration(declaration, symbol, enclosingDeclaration, flags|nodebuilder.FlagsMultilineObjectLiterals, internalFlags, tracker)
+}
+
+func (r *emitResolver) CreateLiteralConstValue(emitContext *printer.EmitContext, node *ast.Node, tracker nodebuilder.SymbolTracker) *ast.Node {
+	node = emitContext.ParseNode(node)
+	r.checkerMu.Lock()
+	type_ := r.checker.getTypeOfSymbol(r.checker.getSymbolOfDeclaration(node))
```
gabritto: I think we've been using `t` instead of `type` since we can't use `type`. It'd be nice to be consistent, but it's up to you.

internal/checker/emitresolver.go
```diff
@@ -225,3 +805,150 @@ func (r *emitResolver) GetReferencedValueDeclarations(node *ast.IdentifierNode)
...
+			)
+		}
+		return emitContext.Factory.NewNumericLiteral(value.Base10Value)
+	case bool:
+		if literalValue.(bool) {
+			return emitContext.Factory.NewKeywordExpression(ast.KindTrueKeyword)
+		}
+		return emitContext.Factory.NewKeywordExpression(ast.KindFalseKeyword)
+	}
+	return nil /// !!! TODO: panic? impossible?
```
gabritto: I think we should panic here if this should be impossible. It's better than returning `nil` and crashing somewhere else with a vague `nil` dereference code.

internal/checker/checker.go
```diff
@@ -3057,12 +3069,12 @@ func (c *Checker) checkTemplateLiteralType(node *ast.Node) {
 func (c *Checker) checkImportType(node *ast.Node) {
 	c.checkSourceElement(node.AsImportTypeNode().Argument)
 	if attributes := node.AsImportTypeNode().Attributes; attributes != nil {
-		c.getResolutionModeOverride(attributes.AsImportAttributes(), true /*reportErrors*/)
+		c.GetResolutionModeOverride(attributes.AsImportAttributes(), true /*reportErrors*/)
```
gabritto: So, it seems we now want to wrap checker methods in a public method in `exports.go` instead of renaming existing ones to make them public... That's what I did in #858 (see https://github.com/microsoft/typescript-go/pull/858#pullrequestreview-2840864050).

internal/checker/nodebuilderapi.go
```diff
@@ -0,0 +1,199 @@
...
+	b.impl.initializeClosures() // recapture ctx
+	b.ctxStack = append(b.ctxStack, b.impl.ctx)
+}
+
+func (b *NodeBuilderAPI) popContext() {
+	b.impl.ctx = nil
+	if len(b.ctxStack) > 1 {
+		b.impl.ctx = b.ctxStack[len(b.ctxStack)-1]
+	}
+	b.ctxStack = b.ctxStack[0 : len(b.ctxStack)-1]
```
gabritto:
```suggestion
	b.ctxStack = b.ctxStack[: len(b.ctxStack)-1]
```

internal/checker/nodebuilderapi.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (b *NodeBuilderAPI) exitContextCheck() {
+	if b.impl.ctx.truncating && b.impl.ctx.flags&nodebuilder.FlagsNoTruncation != 0 {
+		b.impl.ctx.tracker.ReportTruncationError()
+	}
+}
+
+// IndexInfoToIndexSignatureDeclaration implements NodeBuilderInterface.
+func (b *NodeBuilderAPI) IndexInfoToIndexSignatureDeclaration(info *IndexInfo, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node {
+	b.enterContext(enclosingDeclaration, flags, internalFlags, tracker)
```
gabritto: This is like `withContext` in Strada, right? Are we now doing it this way to avoid allocating lambdas?

internal/checker/nodebuilderscopes.go
```diff
@@ -0,0 +1,245 @@
...
+		context.hasCreatedTypeParametersNamesLookups = oldMustCreateTypeParametersNamesLookups
+	}
+}
+
+type localsRecord struct {
+	name      string
+	oldSymbol *ast.Symbol
+}
+
+func (b *NodeBuilder) enterNewScope(declaration *ast.Node, expandedParams *[]*ast.Symbol, typeParameters *[]*Type, originalParameters *[]*ast.Symbol, mapper *TypeMapper) func() {
```
gabritto: I couldn't find a place where we're mutating `typeParameters` or `originalParameters`, so I think those can be regular slice types.

internal/checker/nodebuilderscopes.go
```diff
@@ -0,0 +1,245 @@
...
+	//
+	// If the parameters or return type include "typeof globalThis.paramName", using the wrong scope will lead
+	// us to believe that we can emit "typeof paramName" instead, even though that would refer to the parameter,
+	// not the global. Make sure we are in the right scope by changing the enclosingDeclaration to the function.
+	//
+	// We can't use the declaration directly; it may be in another file and so we may lose access to symbols
+	// accessible to the current enclosing declaration, or gain access to symbols not accessible to the current
+	// enclosing declaration. To keep this chain accurate, insert a fake scope into the chain which makes the
+	// function's parameters visible.
+	var cleanupParams *func()
```
gabritto: I think this can just be `func()`, AFAIK function types are references in Go.

internal/modulespecifiers/specifiers.go
```diff
@@ -0,0 +1,1313 @@
...
+
+			if indexOfStar != -1 {
+				prefix := pattern[0:indexOfStar]
+				suffix := pattern[indexOfStar+1:]
+				for _, c := range candidates {
+					value := c.value
+					if len(value) >= len(prefix)+len(suffix) &&
+						hasPrefix(value, prefix, caseSensitive) && // TODO: possible strada bug: these are not case-switched in strada
+						hasSuffix(value, suffix, caseSensitive) &&
+						validateEncoding(c, relativeToBaseUrl, compilerOptions, host) {
```
andrewbranch:
```suggestion
						validateEnding(c, relativeToBaseUrl, compilerOptions, host) {
```

internal/modulespecifiers/specifiers.go
```diff
@@ -0,0 +1,1313 @@
...
+						matchedStar := value[len(prefix) : len(value)-len(suffix)]
+						if !tspath.PathIsRelative(matchedStar) {
+							return replaceFirstStar(key, matchedStar)
+						}
+					}
+				}
+			} else if core.Some(candidates, func(c specPair) bool { return c.ending != ModuleSpecifierEndingMinimal && pattern == c.value }) ||
+				core.Some(candidates, func(c specPair) bool {
+					return c.ending == ModuleSpecifierEndingMinimal && pattern == c.value && validateEncoding(c, relativeToBaseUrl, compilerOptions, host)
```
andrewbranch:
```suggestion
					return c.ending == ModuleSpecifierEndingMinimal && pattern == c.value && validateEnding(c, relativeToBaseUrl, compilerOptions, host)
```

internal/modulespecifiers/specifiers.go
```diff
@@ -0,0 +1,1313 @@
...
+				}) {
+				return key
+			}
+		}
+	}
+	return ""
+}
+
+func validateEncoding(c specPair, relativeToBaseUrl string, compilerOptions *core.CompilerOptions, host ModuleSpecifierGenerationHost) bool {
```
andrewbranch:
```suggestion
func validateEnding(c specPair, relativeToBaseUrl string, compilerOptions *core.CompilerOptions, host ModuleSpecifierGenerationHost) bool {
```

internal/checker/printer.go
```diff
@@ -6,642 +6,394 @@ import (
...
+		result = append(result, c.nullType)
+	}
+	if flags&TypeFlagsUndefined != 0 {
+		result = append(result, c.undefinedType)
+	}
+	return result
 }
 
-func (p *Printer) printSourceFileWithTypes(sourceFile *ast.SourceFile) {
+func (c *Checker) SourceFileWithTypes(sourceFile *ast.SourceFile) string {
```
gabritto: should that be `PrintSourceFileWithTypes`?

internal/checker/types.go
```diff
@@ -236,6 +267,18 @@ type MarkedAssignmentSymbolLinks struct {
...
 
+type accessibleChainCacheKey struct {
+	useOnlyExternalAliasing bool
+	location                *ast.Node
+	meaning                 ast.SymbolFlags
+}
+
+type ContainingSymbolLinks struct {
+	extendedContainersByFile map[ast.NodeId][]*ast.Symbol // Symbols of nodes which which logically contain this one, cached by file the request is made within
+	extendedContainers       *[]*ast.Symbol               // Containers (other than the parent) which this symbol is aliased in
```
gabritto, line 278: Like elsewhere, I don't think this needs to be a `*[]*ast.Symbol`, but I may be wrong.

internal/checker/nodebuilder.go
```diff
@@ -0,0 +1,2988 @@
...
+
+const (
+	defaultMaximumTruncationLength      = 160
+	noTruncationMaximumTruncationLength = 1_000_000
+)
+
+// Node builder utility functions
+
+// You probably don't mean to use this - use `NewNodeBuilderAPI` instead
+func NewNodeBuilder(ch *Checker, e *printer.EmitContext) NodeBuilder {
```
andrewbranch: If this (and the `NodeBuilder` struct altogether) are not to be used directly, does it need to be exported? If this were `nodeBuilder`, then `NodeBuilderInterface` could become `NodeBuilder`. All the small structs above in this file also seem not to escape the `checker` package either.

internal/compiler/emitter.go
```diff
@@ -378,3 +462,22 @@ func getSourceFilesToEmit(host printer.EmitHost, targetSourceFile *ast.SourceFil
 		return sourceFileMayBeEmitted(sourceFile, host, forceDtsEmit)
 	})
 }
+
+func isJsonSourceFile(file *ast.SourceFile) bool {
```
gabritto: We have that in `ast/utilities.go`. 🙂

internal/checker/nodebuilderapi.go
```diff
@@ -0,0 +1,199 @@
...
+	SymbolToTypeParameterDeclarations(symbol *ast.Symbol, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) []*ast.Node
+	SymbolToParameterDeclaration(symbol *ast.Symbol, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node
+	TypeParameterToDeclaration(parameter *Type, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node
+	SymbolTableToDeclarationStatements(symbolTable *ast.SymbolTable, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) []*ast.Node
+	SymbolToNode(symbol *ast.Symbol, meaning ast.SymbolFlags, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node
+}
+
```
andrewbranch:
```suggestion

var _ NodeBuilderInterface = (*NodeBuilderAPI)(nil)

```

internal/core/compileroptions.go
```diff
@@ -269,6 +271,41 @@ func (options *CompilerOptions) HasJsonModuleEmitEnabled() bool {
...
+func moduleResolutionSupportsPackageJsonExportsAndImports(moduleResolution ModuleResolutionKind) bool {
+	return moduleResolution >= ModuleResolutionKindNode16 && moduleResolution <= ModuleResolutionKindNodeNext || moduleResolution == ModuleResolutionKindBundler
+}
+
+func (options *CompilerOptions) GetResolvePackageJsonImports() bool {
+	moduleResolution := options.GetModuleResolutionKind()
+	if !moduleResolutionSupportsPackageJsonExportsAndImports(moduleResolution) {
+		return false
+	}
+	if options.ResolvePackageJsonExports != TSUnknown {
```
gabritto: Shouldn't this be `options.ResolvePackageJsonImports`? Or is this the Strada bug alluded to below? Definitely doesn't look right though.

internal/declarations/util.go
```diff
@@ -0,0 +1,243 @@
...
+		ast.KindVariableDeclaration,
+		ast.KindExportAssignment,
+		ast.KindJSExportAssignment,
+		ast.KindPropertyAssignment,
+		ast.KindShorthandPropertyAssignment,
+		ast.KindJSDocParameterTag,
+		ast.KindJSDocPropertyTag:
+		return true
+	default:
+		// assertType<never>(node); // !!!
```
gabritto: Couldn't those be panics?

internal/declarations/util.go
```diff
@@ -0,0 +1,243 @@
...
+func getFirstConstructorWithBody(node *ast.Node) *ast.Node {
+	for _, member := range node.Members() {
+		if ast.IsConstructorDeclaration(member) && ast.NodeIsPresent(member.Body()) {
+			return member
+		}
+	}
+	return nil
+}
+
+func getEffectiveBaseTypeNode(node *ast.Node) *ast.Node {
```
gabritto: Same as elsewhere, I don't think we'll need this with the new jsdoc approach.

internal/declarations/util.go
```diff
@@ -0,0 +1,243 @@
...
+	//     // Prefer an @augments tag because it may have type parameters.
+	//     const tag = getJSDocAugmentsTag(node);
+	//     if (tag) {
+	//         return tag.class;
+	//     }
+	// }
+	return baseType
+}
+
+func getClassExtendsHeritageElement(node *ast.Node) *ast.Node {
```
gabritto: I recently added this function to `ast/utilities.go` in one of my completions PR.

gabritto (commented):
75% done looking through the files. I just had a few questions and some comments about Go stuff.

internal/checker/nodebuilder.go
```diff
@@ -0,0 +1,2988 @@
...
+	if index > 0 {
+		return b.f.NewQualifiedName(
+			b.createEntityNameFromSymbolChain(chain, index-1),
+			identifier,
+		)
+	}
+	return identifier
+}
+
+// TODO: Audit usages of symbolToEntityNameNode - they should probably all be symbolToName
```
gabritto: +1 for this, I remember being confused about this when working on expandable hover.

internal/checker/nodebuilder.go
```diff
@@ -0,0 +1,2988 @@
...
+}
+
+/**
+ * Strip off existed surrounding single quotes, double quotes, or backticks from a given string
+ *
+ * @return non-quoted string
+ *
+ * @internal
+ */
+func stripQuotes(name string) string {
```
gabritto: I think this is the same as `core.StripQuotes`.

internal/core/compileroptions.go
andrewbranch: Confirmed this is a Strada bug. Luckily I think nobody ever changes imports/exports support independently. I’ll go ahead and open a Strada PR to fix.

Note also that you don’t need to check `moduleResolution` in Corsa, because the only values that don’t support imports/exports are gone.

internal/modulespecifiers/util.go
```diff
@@ -0,0 +1,325 @@
...
+		return false
+	}
+	return range_.Test(&typeScriptVersion)
+}
+
+func replaceFirstStar(s string, replacement string) string {
+	return strings.Replace(s, "*", replacement, 1)
+}
+
+func hasPrefix(s string, prefix string, caseSensitive bool) bool {
```
gabritto: If there's any chance we'll use this elsewhere, I'd move it to `stringutil`.

internal/checker/symbolaccessibility.go
```diff
@@ -0,0 +1,762 @@
...
+ */
+func (ch *Checker) getAccessibleSymbolChainFromSymbolTable(ctx accessibleSymbolChainContext, t ast.SymbolTable, ignoreQualification bool, isLocalNameLookup bool) []*ast.Symbol {
+	symId := ast.GetSymbolId(ctx.symbol)
+	visitedSymbolTables, ok := ctx.visitedSymbolTablesMap[symId]
+	if !ok {
+		visitedSymbolTables = make(map[uintptr]struct{})
+		ctx.visitedSymbolTablesMap[symId] = visitedSymbolTables
+	}
+
+	id := reflect.ValueOf(t).Pointer() // TODO: Is this seriously the only way to check reference equality of maps?
```
jakebailey: This is definitely unsafe. The symbol tables could get GC'd because this is hiding the pointer behind a `uintptr`. Do we really need to be checking reference equality?

The only way that this could be okay is if we used `.UnsafePointer()` (the recommended function) instead of `uintptr` to ensure we actually have a reference. But, this is all quite spooky.

internal/checker/checker.go
```diff
@@ -821,22 +827,25 @@ type Checker struct {
...
 	emitResolverOnce                           sync.Once
+	diagnosticConstructionContext              *printer.EmitContext
+	nodeBuilder                                NodeBuilderInterface
 	_jsxNamespace                              string
 	_jsxFactoryEntity                          *ast.Node
 	skipDirectInferenceNodes                   core.Set[*ast.Node]
 }
 
-func NewChecker(program Program) *Checker {
+func NewChecker(program Program, host Host) *Checker {
```
andrewbranch, line 837: I don’t love this—the only implementation of `Host` is the private `compiler.emitHost`, which means you can’t easily make a fully functional checker from outside the `compiler` package, which will cause problems in my async LSP branch (the Project system’s checker pool will break with no clear resolution), even though `compiler.emitHost` is just trivially a wrapper around `Program`.

internal/checker/emitresolver.go
andrewbranch: `isDeclarationVisible` is recursive through `determineIfDeclarationIsVisible`, so you can’t lock at the beginning of that one, or else you’d try to acquire a lock you’re already holding on the recursive calls.

gabritto (commented):
Done now

internal/checker/nodebuilder.go
weswigham: My goal was to move the implementation into the `nodebuilder` package by abstracting the `Checker` APIs it uses, but turns out the builder uses over 120(!) checker APIs - which isn't itself an issue beyond the effort involved, but they also require moving `Type`/`Signature`/`IndexInfo` (and all the derived subtypes of `Type`) up the stack, too (and, in turn, making their fields public or into methods), which I know I'll get... _opinions_ on, so I'm saving that for a followup. I can lowercase it for now... but it doesn't matter either way, since it's in the `internal` package. Honestly, I don't know why we wouldn't default to package exported in this environment.

internal/checker/nodebuilderapi.go
weswigham: There's some usages that effectively accomplish this already, but I can readd this if you'd like.

internal/ast/utilities.go
weswigham: Do we have a canonical replacement for calls to `Debug.assert` or other debug methods (perhaps involving conditional compilation)? I've been putting `panic`s in for `Debug.fail`s, but assertions/typechecks I haven't seen a canonical replacement for, so I've been leaving them around as comments to note places where we used to check more invariants.

internal/ast/utilities.go
weswigham: I dunno! Maybe, maybe not? Some things are still done the old way. I implemented as much of the JS declaration emit as it seemed like we supported, but ultimately it's gonna need some touch-ups when we actually enable testing it.

internal/checker/checker.go
weswigham: Yep - there's an emit resolver function the declaration emitter calls when it visits a file. Much neater, doesn't require knowledge during emit that something needs to conditionally be done. It's also *never* done during check anymore, and is cached inside the emit resolver instead - it's just always done as a short walk during declaration emit.

internal/checker/emitresolver.go
weswigham: Emit resolver APIs have, historically, been hardened to incorrect inputs, with the expectation that the nodes they're fed by API consumers might not quite follow all the conventions the ones we pass do. It's a cleanup we should do when we make a public API surface out of this.

internal/checker/emitresolver.go
weswigham: Precisely.  So we only lock on the public entrypoints, so they only lock on initial entry.

internal/ast/utilities.go
jakebailey: We don't yet, we had found them to be slow so were deferring that.

internal/checker/emitresolver.go
weswigham: It should - I left these around as a reminder of what to doublecheck when we unflag and start testing js declarations.

internal/checker/nodebuilderapi.go
weswigham: Yep, this wrapper serves to adapt the internal APIs shapes to the exposed API shapes and construct and manage node builder contexts. Exactly what `withContext` did before.

internal/checker/types.go
weswigham: In this case, we need to distinguish between not-yet-set and set-but-empty, so we don't recalculate that there's no extended containers of a symbol repeatedly.

internal/checker/symbolaccessibility.go
weswigham: We use the identity of the symbol tables to check which ones we've already traversed, and by rights they should never mutate _or_ be GC'd under us here, since we should have calculated and cached the tables by this point (which should only lose their last referrer when the checker itself is collected). Can't use symbols or symbol ids for identity, since they contain multiple tables (and, surprisingly, the same table can end up in multiple places). And exhaustive map comparison is obviously a nonstarter, as these are often huge. I can swap the pointer kind in use, if that'll help, though.

internal/checker/checker.go
weswigham: I *could* flatten all the methods needed to satisfy all the various host interfaces onto `Program`, but then that headache just gets coupled to implementing a `Program`, no? This at least gives you the option to provide them separately, as we already do. Basically all the `Host` functionality requested by the checker has to do with file metadata that should be sharable cross-program...

internal/modulespecifiers/util.go
weswigham: `strings.HasPrefix` is canonical in the general case - you only need this if you need to dynamically swap case handling, and in our old codebase, we only ever did that here... (though not doing it elsewhere is probably inaccurate). I can move it, regardless.

internal/ast/utilities.go
gabritto: So we're not having the equivalent of `Debug.assert`s in Corsa, or are we still doing it in the future when we figure out a faster way?

jakebailey: I think we will in the future, either as a build tag guarded thing, or making it faster generally speaking

internal/checker/nodebuilder.go
andrewbranch: Yeah, I think the reasons to keep stuff unexported are just for us, not for future importers. It keeps completions in other packages cleaner and faster, and my understanding is it’s good for whole project builds (you can edit signatures of unexported stuff without invalidating the build of downstream packages). But mainly having to name interfaces `FooInterface` makes me sad 😄

internal/checker/checker.go
andrewbranch: I don’t think there’s actually an expectation that a Checker will be backed by anything other than a `compiler.Program` struct; I think the interface is just there to untangle the circular dependency. (The only implementation of `TypeCheckerHost` in Strada is the `Program` that gets returned from `createProgram`.) I think minimal contracts are good, but I don’t see a need to pick them apart if they only ever get used in combination by one monolith. At the very least, for practical immediate purposes, there needs to be a convenient way to create a checker with just a `compiler.Program` from outside the `compiler` or `checker` packages.

internal/checker/printer.go
weswigham: Neither! It's simply gone now~

jakebailey:
I don't want to spam the PR with this comment, but I think we're trying to put any exported function shims in `exports.go` rather than rename functions to be exported, just to make it super clear where "public" things are sitting.

internal/checker/symboltracker.go
```diff
@@ -0,0 +1,140 @@
...
+		inner = tracker.GetInnerSymbolTracker()
+		if inner == nil {
+			inner = tracker
+		}
+	}
+
+	return &SymbolTrackerImpl{context, inner, false, tchost}
+}
+
+func (this *SymbolTrackerImpl) GetModuleSpecifierGenerationHost() modulespecifiers.ModuleSpecifierGenerationHost {
```
andrewbranch, line 28: `this` 🤨

internal/checker/symboltracker.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/modulespecifiers"
+	"github.com/microsoft/typescript-go/internal/nodebuilder"
+)
+
+type SymbolTrackerImpl struct {
+	context            *NodeBuilderContext
+	inner              nodebuilder.SymbolTracker
```
andrewbranch, line 11: The number of things that are types that are `Impl` or `Interface` and members that are `impl` or `inner` makes me feel like something is in the wrong place and/or we’re trying to be too clever with polymorphism. How much of this is temporary until more things are moved out of checker.ts?

internal/checker/symboltracker.go
weswigham: Hey, it is analogous and it's not a keyword, so why not keep it similar where we actually *did* use objects/classes in Strada?

internal/checker/symboltracker.go
weswigham: The `SymbolTracker` interface defined outside `checker` *is* required - otherwise the printer and checker end up circularly dependent on one another, so don't work. I did just push a removal of the `NodeBuilderInterface` since, for now at least, it's not strictly required.

jakebailey:
`hugeDeclarationOutputGetsTruncatedWithError` appears to time out on my machine; maybe we need to skip its output? Or, there's a perf bug?

weswigham:
Definite perf issue - it completes locally for me in *checks* around 50 seconds.

weswigham:
I think the perf issue is in the test suite - pprof shows almost all 50 seconds spent in `myers.Diff`...

jakebailey:
Hm, unfortunate; the test suite doesn't have a way to skip a diff yet, but could a la `fixupOld` plumbed downward.

weswigham:
Can we just specify a higher timeout for an individual test in the interim? For now, the diff _is_ technically still useful - just expensive to calculate for a large output such as that test.

jakebailey:
No, unfortunately the test timeouts are global as golang/go#48157 isn't yet implemented.

internal/bundled/generate.go
```diff
@@ -133,7 +133,7 @@ func readLibs() []lib {
 		paths map[string]string
 	}
 
-	libsFile := filepath.Join(libInputDir, "libs.json")
+	libsFile := tspath.NormalizeSlashes(filepath.Join(libInputDir, "libs.json"))
```
jakebailey, line 136: Ah, good catch.

internal/checker/nodebuilder.go
```diff
@@ -0,0 +1,180 @@
...
+
+// EmitContext implements NodeBuilderInterface.
+func (b *NodeBuilder) EmitContext() *printer.EmitContext {
+	return b.impl.e
+}
+
+func (b *NodeBuilder) enterContext(enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) {
+	b.impl.ctx = &NodeBuilderContext{
+		tracker:                  tracker,
+		approximateLength:        0,
```
jakebailey: FWIW you can freely leave zero/false/nil as-is and not specify them.

internal/checker/nodebuilderscopes.go
```diff
@@ -0,0 +1,245 @@
...
+					if originalParameters != nil && originalParam != param {
+						// Can't reference parameters that come from an expansion
+						add(param.Name, b.ch.unknownSymbol)
+						// Can't reference the original expanded parameter either
+						if originalParam != nil {
+							add(originalParam.Name, b.ch.unknownSymbol)
+						}
+					} else if !core.Some(param.Declarations, func(d *ast.Node) bool {
+						var bindElement *(func(e *ast.BindingElement))
+						var bindPattern *(func(e *ast.BindingPattern))
```
jakebailey: Functions are already nilable; I think you could safely omit the `*`.

internal/checker/symboltracker.go
jakebailey: I guess this is the _only_ class in the old codebase, but I still don't like it 😄

internal/checker/types.go
```diff
@@ -75,6 +75,37 @@ const (
 	TypeFormatFlagsInTypeAlias         TypeFormatFlags = 1 << 23 // Writing type in type alias declaration
 )
 
+const TypeFormatFlagsNodeBuilderFlagsMask = TypeFormatFlagsNoTruncation | TypeFormatFlagsWriteArrayAsGenericType | TypeFormatFlagsGenerateNamesForShadowedTypeParams | TypeFormatFlagsUseStructuralFallback | TypeFormatFlagsWriteTypeArgumentsOfSignature |
+	TypeFormatFlagsUseFullyQualifiedType | TypeFormatFlagsSuppressAnyReturnType | TypeFormatFlagsMultilineObjectLiterals | TypeFormatFlagsWriteClassExpressionAsTypeLiteral |
+	TypeFormatFlagsUseTypeOfFunction | TypeFormatFlagsOmitParameterModifiers | TypeFormatFlagsUseAliasDefinedOutsideCurrentScope | TypeFormatFlagsAllowUniqueESSymbolType | TypeFormatFlagsInTypeAlias |
+	TypeFormatFlagsUseSingleQuotesForStringLiteralType | TypeFormatFlagsNoTypeReduction | TypeFormatFlagsOmitThisParameter
+
+// dprint-ignore
```
jakebailey: Shouldn't need this anymore, given dprint isn't in control of this.

internal/compiler/emitter.go
```diff
@@ -298,7 +372,17 @@ type outputPaths struct {
...
 }
 
-func getOutputPathsFor(sourceFile *ast.SourceFile, host printer.EmitHost, forceDtsEmit bool) *outputPaths {
+// DeclarationFilePath implements declarations.OutputPaths.
+func (o *outputPaths) DeclarationFilePath() string {
+	return o.declarationFilePath
+}
+
+// JsFilePath implements declarations.OutputPaths.
+func (o *outputPaths) JsFilePath() string {
```
jakebailey, line 381: FWIW it's more canonical in Go to write `JSFilePath`; I think we've mostly done that elsewhere?

jakebailey: But, we also write `Json` in places still so, shrug

internal/compiler/program.go
```diff
@@ -70,6 +71,105 @@ type Program struct {
 	unsupportedExtensions []string
 }
 
+// FileExists implements checker.Program.
+func (p *Program) FileExists(path string) bool {
```
jakebailey, line 76: I don't think we should be making Program implement some of this stuff, right? I had kept this all on FS intentionally, to avoid every object turning into some sort of `Host` thing like the old codebase.

internal/core/compileroptions.go
jakebailey: Do we fix this here now? Probably worth doing?

internal/core/core.go
```diff
@@ -557,3 +557,22 @@ func StripQuotes(name string) string {
 	}
 	return name
 }
+
+func GetPathsBasePath(options *CompilerOptions, currentDirectory string) string {
```
jakebailey: This could just be a method on `CompilerOptions`, I think.

internal/declarations/transform.go
jakebailey, line 1: Should this package be placed under the transforms package?

internal/ast/utilities.go
```diff
@@ -903,6 +904,11 @@ func SetParentInChildren(node *Node) {
 	fn(node)
 }
 
+// This should never be called outside the parser
+func SetImportsOfSourceFile(node *SourceFile, imports []*LiteralLikeNode) {
```
jakebailey, line 908: Should this just be a method? Or just provide imports via `NewSourceFile` like we do the rest? (Probably can't because it's a second pass ☹️)

internal/checker/checker.go
```diff
@@ -519,6 +520,7 @@ type WideningContext struct {
 }
 
 type Program interface {
+	Host
```
jakebailey, line 523: I don't really like us going back to the old days of "everything's a host you have to forward though"; is there a way to avoid this?

internal/transformers/commonjsmodule.go
```diff
@@ -38,7 +38,7 @@ func NewCommonJSModuleTransformer(emitContext *printer.EmitContext, compilerOpti
 	tx.assignmentPatternVisitor = emitContext.NewNodeVisitor(tx.visitAssignmentPattern)
 	tx.languageVersion = compilerOptions.GetEmitScriptTarget()
 	tx.moduleKind = compilerOptions.GetEmitModuleKind()
-	return tx.newTransformer(tx.visit, emitContext)
+	return tx.NewTransformer(tx.visit, emitContext)
```
jakebailey, line 41: Is this just because the declaration emitter is not in this same package?

internal/transformers/importelision_test.go
```diff
@@ -22,6 +23,42 @@ type fakeProgram struct {
 	getResolvedModule           func(currentSourceFile *ast.SourceFile, moduleReference string) *ast.SourceFile
 }
 
+func (p *fakeProgram) FileExists(path string) bool {
```
jakebailey, line 26: Yeah, having to have all of these is sort of a shame but maybe there's no way around it ☹️

jakebailey (commented):
I did a pass of some of the files; have not ventured into the larger files quite yet.

internal/compiler/program.go
weswigham: Well, @andrewbranch said they'd rather see host functionality merged into `Program` so it can provide all the capabilities needed without needing to go out of your way to construct another host interface-providing object... Rather than writing passthru functions, you can have methods on `Program` (as required on the `Host`) that expose the program's `FS` directly (basically an `AsFS()` method like we do for `Node` and `NodeFactory`)), but that just grows the accidental dependency list in this case, when we only need a single shard of the underlying functionality, no? (Certainly it makes writing the test fake impl harder)

The layering is wishy-washy right now, but once `-b` work is underway, we'll have secondary `Program` implementations (multiple) which'll actually be impacted by these decisions.

jakebailey: Well, at least it's all a lot less obscure than the old code 😄

internal/declarations/transform.go
weswigham: It can be, if you'd rather see it there than at the top level. All that really changes would be the import path an internal user would use to pull it out, afaik.

internal/ast/utilities.go
weswigham: Yeah, can't be in the constructor because of the late-calculation thing, and I'm not sure something like this, which aught not be used without very good reason, should be on the method completion list. I'd rather not put mutating methods on AST nodes.

internal/checker/checker.go
weswigham: ...? No? The `program` is, by request, _the only argument to making a checker_, ergo it must provide all external, configurable/queryable state functionality the checker needs. All the metadata/state must flow through it, since it's the only point of contact between the checker and anything else. None of the methods added by this host are onerous for a normal `program` to implement, either (hence why the emit host, which wraps a program, can also easily provide them) - they're either methods on its own internal FS/host, or metadata it's already calculating during setup (like package json cache info). Like, you can expose the underlying things the `program` owns to implement *some* of the additional methods directly, like the `FS` or `Resolver`, but all that does is push complexity farther down and make implementing the interface for test/secondary use harder by accidentally exposing more scope...

jakebailey: Well, this goes back to my question about "is a Program a host, or does it _have_ a host"? And I prefer the latter but I can get why the former is more convenient

internal/transformers/commonjsmodule.go
weswigham: Yeah, but defining transforms outside the core `transform` package could also be done in some quickfixes and would be done for some lint rule (fixes) - these methods on the base transform should be exported, imo, to support defining transforms where's they're needed, rather than only in one place.

internal/transformers/importelision_test.go
weswigham: Yup. I certainly don't wanna leak all the stuff on `program`'s internal bits into the `checker` scope - minimally it'd make writing the interfaces incredibly painful.

internal/transformers/commonjsmodule.go
jakebailey: Yeah, I think declaration emit being "different" compared to the other transforms seems fine. It really does do something different than downleveling transforms. (I kinda want the other transforms in a different package, too.)

internal/checker/checker.go
weswigham: Historically, it *is* a host (with an internal host of it's own for fs access) that we wrap/window/subset/augment as needed. In strada it's the primary implementation of `TypeCheckerHost` - we've just changed the term in use here - rather than naming the interface `checker.TypeCheckerHost` or `checker.Host`, we named it `checker.Program`. Just because, I guess.

jakebailey: It used to be that they were in the same package (compiler), so it was concrete before and was just left as an interface when the refactor happened.

internal/core/core.go
andrewbranch: It's already a private function in `module`, could you remove it from there if you make it a CompilerOptions method?

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+				}
+			}
+		}
+	}
+	return false
+}
+
+func (r *emitResolver) RequiresAddingImplicitUndefined(declaration *ast.Node, symbol *ast.Symbol, enclosingDeclaration *ast.Node) bool {
+	// node = r.emitContext.ParseNode(node)
+	if !ast.IsParseTreeNode(declaration) {
```
jakebailey, line 476: All of these `IsParseTreenode` calls appear to be dead (at least, uncovered); is this just since we don't have anyone calling us with weird nodes?

internal/checker/emitresolver.go
```diff
@@ -1,21 +1,586 @@
...
+	switch declaration.Kind {
+	case ast.KindPropertyDeclaration, ast.KindPropertySignature, ast.KindJSDocPropertyTag:
+		r.checkerMu.Lock()
+		defer r.checkerMu.Unlock()
+		if symbol == nil {
+			symbol = r.checker.getSymbolOfDeclaration(declaration)
+		}
+		t := r.checker.getTypeOfSymbol(symbol)
+		r.checker.mappedSymbolLinks.Has(symbol)
+		return !!((symbol.Flags&ast.SymbolFlagsProperty != 0) && (symbol.Flags&ast.SymbolFlagsOptional != 0) && isOptionalDeclaration(declaration) && r.checker.ReverseMappedSymbolLinks.Has(symbol) && r.checker.ReverseMappedSymbolLinks.Get(symbol).mappedType != nil && containsNonMissingUndefinedType(r.checker, t))
```
jakebailey, line 488: This code here seems to be uncovered, which is interesting.

internal/checker/nodebuilderimpl.go
```diff
@@ -0,0 +1,2972 @@
...
+	links       core.LinkStore[*ast.Node, NodeBuilderLinks]
+	symbolLinks core.LinkStore[*ast.Symbol, NodeBuilderSymbolLinks]
+
+	// closures
+	typeToTypeNodeClosure               func(t *Type) *ast.TypeNode
+	typeReferenceToTypeNodeClosure      func(t *Type) *ast.TypeNode
+	conditionalTypeToTypeNodeClosure    func(t *Type) *ast.TypeNode
+	createTypeNodeFromObjectTypeClosure func(t *Type) *ast.TypeNode
+	isStringNamedClosure                func(d *ast.Declaration) bool
+	isSingleQuotedStringNamedClosure    func(d *ast.Declaration) bool
```
jakebailey: I would suggest inverting this and give these the "nice names", since they are what should be called by everyone.

internal/checker/nodebuilderimpl.go
```diff
@@ -0,0 +1,2972 @@
...
+	}
+	// TODO: in strada, this only used `isIdentifierStart` on the first character, while this checks the whole string for validity
+	// - possible strada bug?
+	if strings.HasPrefix(name, "#") {
+		return len(name) > 1 && scanner.IsIdentifierText(name[1:], languageVersion, core.LanguageVariantStandard)
+	}
+	return scanner.IsIdentifierText(name, languageVersion, core.LanguageVariantStandard)
+}
+
+func unquoteString(str string) string {
```
jakebailey: This would be good to shift into `stringutil` or something; I suspect we have it (or will need it for something else in the future).

internal/checker/nodebuilderimpl.go
```diff
@@ -0,0 +1,2972 @@
...
+				name := associatedNames[i]
+				flags := restType.Target().AsTupleType().elementInfos[i].flags
+				var checkFlags ast.CheckFlags
+				switch {
+				case flags&ElementFlagsVariable != 0:
+					checkFlags = ast.CheckFlagsRestParameter
+				case flags&ElementFlagsOptional != 0:
+					checkFlags = ast.CheckFlagsOptionalParameter
+				default:
+					checkFlags = 0
```
jakebailey: FWIW you could just leave this out, given zero initialization.

internal/checker/types.go
jakebailey: Could you just use `nil` vs not `nil`?

internal/transformers/declarations/diagnostics.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+	case ast.KindFunctionDeclaration:
+		return selectDiagnosticBasedOnModuleName(
+			symbolAccessibilityResult,
+			diagnostics.Return_type_of_exported_function_has_or_is_using_name_0_from_external_module_1_but_cannot_be_named,
+			diagnostics.Return_type_of_exported_function_has_or_is_using_name_0_from_private_module_1,
+			diagnostics.Return_type_of_exported_function_has_or_is_using_private_name_0,
+		)
+	default:
+		panic(fmt.Sprintf("This is unknown kind for signature: %d", node.Kind))
```
jakebailey:
```suggestion
		panic("This is unknown kind for signature: " + node.Kind.String())
```

jakebailey: (and similarly below)

jakebailey (approved):
Definitely a "so much code I can't possibly review every line" situation, but I skimmed things and pointed out the couple of tiny non-blocking nits I noticed.

If things don't crash, and baselines seem like improvement, LGTM. Has to happen at some point, so!

internal/checker/emitresolver.go
weswigham: Correct - the only internal use of the emit resolver is the declaration emitter, which already takes care to pass only parsed nodes and not transformed ones. But all the `EmitResolver` APIs have historically been hardened vs non-parse input to support their use in custom transforms.

internal/checker/emitresolver.go
weswigham: AFAIK, the primary caller of this function is ID error generation, which isn't in yet. The current calls are only to handle some edge cases in declaration emit, the tested cases for which probably don't overlap with this specific condition. I know this line in strada was added to fix an ID bug, at least.

internal/checker/types.go
weswigham: I could, but if general guidance is to consider nil and empty slices the same, isn't this more explicit?

jakebailey: Usually it's not so good but sometimes it's not that bad... But it's fine to leave this

jakebailey:
Seems like there's still some raciness and non-determinism; the smoke tests have a race failure and the baselines appear to have some ordering differences when using concurrent program with info from multiple checkers.

jakebailey:
```
Write at 0x00c0154f5e90 by goroutine 747:
  runtime.mapassign_fast64ptr()
      /opt/hostedtoolcache/go/1.24.3/x64/src/internal/runtime/maps/runtime_fast64_swiss.go:367 +0x0
  github.com/microsoft/typescript-go/internal/core.(*LinkStore[go.shape.*github.com/microsoft/typescript-go/internal/ast.Node,go.shape.struct { github.com/microsoft/typescript-go/internal/checker.isVisible github.com/microsoft/typescript-go/internal/core.Tristate }]).Get()
      /home/runner/work/typescript-go/typescript-go/internal/core/linkstore.go:19 +0x15b
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).isDeclarationVisible()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:91 +0x5e
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).IsDeclarationVisible()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:79 +0x9b
  github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformImportDeclaration.func1()
      /home/runner/work/typescript-go/typescript-go/internal/transformers/declarations/transform.go:1672 +0x4d
  github.com/microsoft/typescript-go/internal/core.Filter[go.shape.*uint8]()
      /home/runner/work/typescript-go/typescript-go/internal/core/core.go:20 +0x9c
```

```
Previous read at 0x00c0154f5e90 by goroutine 739:
  runtime.mapaccess1_fast64()
      /opt/hostedtoolcache/go/1.24.3/x64/src/internal/runtime/maps/runtime_fast64_swiss.go:17 +0x0
  github.com/microsoft/typescript-go/internal/core.(*LinkStore[go.shape.*github.com/microsoft/typescript-go/internal/ast.Node,go.shape.struct { github.com/microsoft/typescript-go/internal/checker.isVisible github.com/microsoft/typescript-go/internal/core.Tristate }]).Get()
      /home/runner/work/typescript-go/typescript-go/internal/core/linkstore.go:11 +0x7b
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).isDeclarationVisible()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:91 +0x5e
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).hasVisibleDeclarations()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:350 +0x187
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).isEntityNameVisible()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:315 +0x424
  github.com/microsoft/typescript-go/internal/checker.(*emitResolver).IsEntityNameVisible()
      /home/runner/work/typescript-go/typescript-go/internal/checker/emitresolver.go:278 +0x84
  github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).checkEntityNameVisibility()
      /home/runner/work/typescript-go/typescript-go/internal/transformers/declarations/transform.go:1039 +0x7a
```

(for example)

weswigham:
> Seems like there's still some raciness and non-determinism; the smoke tests have a race failure and the baselines appear to have some ordering differences when using concurrent program with info from multiple checkers.

Probably missing some locks on some emit resolver APIs, I'm guessing. I fixed all the run-to-run output stability issues locally, but hadn't actually run the race tests. Looks like `IsEntityNameVisible` is missing a lock, from that output.

jakebailey:
Interestingly, this is not a problem during compiler test runs it seems, only in the CLI, which we should definitely have more tests for than just the one-off smoke tests.

weswigham:
I think I've quashed all the `race` issues (and gotten race mode testing running locally), but the last fix for the last one might not fix the underlying raciness, just hide it in our tests by fixing the logic bug I saw while looking over the relevant code closely - there was a missing `!` in computed name elision in the declaration emitter, so weather we elided or kept late bound names was partially inverted (which, worryingly, affects surprisingly few tests). That error, in turn, removed a location where we marked late-painted statements (since we were eliding the name that definitely marks them) and caused the late painted statement visibility to only depend on weather a prior checker API call has witnessed the link between the type node and the statement and painted the visibility of the target. On reflection, I... think it's fine, with the fixed logic? So long as visibility is painted at-least-once, it's fine - so with the fixed logic, it's no longer maybe-visibility-painted-zero-times, as the declaration emitter itself is ensuring at least one check.

weswigham:
And CI finally green 👍

gabritto (approved):

hamidrezahanafi:
hey @weswigham I filed an issue [here](https://github.com/microsoft/typescript-go/issues/1012) and it seems this PR is culprit (local TSGO build with commit before this PR works fine), let me know if I can help in anyway

PR #792, opened 2025-04-13
LSP: Jump to the function name

shinichy:
Improved the "Go to Definition" command to jump to the function name instead of the beginning of the `function` keyword.

### DIFF
--- a/internal/ls/definition.go
+++ b/internal/ls/definition.go
@@ -3,6 +3,7 @@ package ls
 import (
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/astnav"
+	"github.com/microsoft/typescript-go/internal/checker"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/scanner"
 )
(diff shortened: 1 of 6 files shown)


internal/ls/definition.go
```diff
@@ -15,25 +16,70 @@ func (l *LanguageService) ProvideDefinitions(fileName string, position int) []Lo
 	}
 
 	checker := program.GetTypeChecker()
+	calledDeclaration := tryGetSignatureDeclaration(checker, node)
```
shinichy, line 24: I tried to port the logic from `src/services/goToDefinition.ts` in the TypeScript repository.

Mustafiz04 (approved):

sandersn:
@iisaduan is working on find-all-refs, so she may have already looked at this code.

Should some tests change? I'm not sure if fourslash tests are available yet.

iisaduan:
It's pretty separate from the find all refs work I've been doing, so there's no duplication. But the PR probably should update tests, so we have some sort of reference for what has been ported already/what we expect to work

iisaduan:
I could probably test this with my find all refs testing, but it's not quite ready yet

shinichy:
It's good to know that there's no duplication, as that's something I was concerned about since your team seems to be actively working on the `ls` package. I'll also try to add a test.

shinichy:
I added a basic test. Please let me know if I should cover more test cases, as there are currently no tests for `definition.go`.

jakebailey:
This has unfortunately gotten out of date as main has changed. Any chance you could update this so we can review it?

shinichy:
Yes, let me merge main and resolve the conflicts.

shinichy:
@jakebailey I've merged main and resolved the conflicts.

jakebailey (approved):

PR #797, opened 2025-04-14
Replace LSP generator with new rewritten version

jakebailey:
This fixes a bunch of issues:

- Structs are now consistently referenced as pointers.
  - Exceptions: `Position`, `Location`, `Range`, etc, are value types.
- Nullability is now correctly modeled where needed.
- Comments are now closer to the metaModel

Future changes I've reverted out for a smaller diff:

- `omitzero` instead of `omitempty`.
- Union unmarshallers use blocks to scope values


internal/lsp/lsproto/_generate/generate.mjs
```diff
@@ -26,524 +26,805 @@ if (!fs.existsSync(metaModelPath)) {
...
+        // Use Nullable[T] instead of pointer for null union with one other type
+        return {
+            name: `Nullable[${resolvedType.name}]`,
+            needsPointer: false,
+        };
+    }
 
-/** @type {Map<string, UnionMember[]>} */
-const unionTypes = new Map();
+    const types = orType.items.filter((_, index) => index !== nullIndex);
```
gabritto: Maybe I'm missing something, but do we need to filter this? We're already dealing with the case where `nullIndex !== -1` above.

jakebailey: Ah, you're totally right

gabritto (approved):
Looks good to me, except for the tiny comment I left.

PR #804, opened 2025-04-17
Add `findPrecedingToken`

gabritto:
The new code ports the original `findPrecedingToken` as faithfully as possible.

From the baselines, there are two differences:
1. The range of some returned tokens is different in Corsa.
2. In Strada, when the position is inside a JSDoc comment that has no tags, `findPrecedingToken` returns `undefined`. In Corsa, right now, we return the JSDoc. I am not sure if this difference matters or how to properly deal with that, so I thought I should leave this as a future work item.


andrewbranch (approved):
Baseline diff looks good—I think those are Strada bugs (that probably never matter in practice).

navya9singh (approved):

PR #805, opened 2025-04-17
Add ability to build extension VSIX and swap for built-in extension

andrewbranch:
This PR sets up a `hereby` command to build and persistently install the tsgo VS Code extension. It does nothing if the default typescript-language-features extension is active. https://github.com/microsoft/vscode/pull/246858 allows for disabling that one via configuration/command, allowing the tsgo LSP to take over. Also adds a warning-colored status bar item to make it obvious when you’re running tsgo instead of stable TypeScript in the editor.

https://github.com/user-attachments/assets/ea9481c9-8c04-462d-966c-c3435170bd6d


jakebailey:
Ah, seems like you have the npm bug that Kat fixed; probably want to restore your lockfile to main, delete all node_modules, and then install again.

Herebyfile.mjs
```diff
@@ -348,6 +350,17 @@ export const testAll = task({
     },
 });
 
+export const installExtension = task({
+    name: "install-extension",
```
jakebailey, line 354: Theoretically long-term we're going to have to do something special here, since the VSIX is technically platform dependent.

andrewbranch: Huh? This is for local development/dogfooding.

jakebailey: Right, this is more rumination about "how we're going to build stuff" at some point.

package-lock.json
jakebailey, line 1: What in the heck is vsce doing with 5000 lines of deps added???

jakebailey: wowwww https://npmgraph.js.org/?q=%40vscode%2Fvsce

RyanCavanaugh: J***** ******d 🥲

gabritto (approved):

PR #806, opened 2025-04-17
Watch files over LSP

andrewbranch:
This adds LSP file watching requests for wildcard directories, failed/affecting lookup locations, and tsconfig files.


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -54,11 +65,20 @@ func (id *ID) MustInt() int32 {
...
 	JSONRPC JSONRPCVersion `json:"jsonrpc"`
-	ID      *ID            `json:"id"`
+	ID      *ID            `json:"id,omitempty"`
 	Method  Method         `json:"method"`
 	Params  any            `json:"params"`
 }
 
+func NewRequestMessage(method Method, id *ID, params any) *RequestMessage {
+	return &RequestMessage{
+		JSONRPC: JSONRPCVersion{},
```
jakebailey:
```suggestion
```

No need to specify this since it's zero'd.

internal/lsp/server.go
```diff
@@ -62,36 +67,100 @@ type Server struct {
 	initializeParams *lsproto.InitializeParams
 	positionEncoding lsproto.PositionEncodingKind
 
+	watcheEnabled  bool
+	watcherID      int
+	watchers       map[project.WatcherHandle]struct{}
```
jakebailey: Is this just a `core.Set[project.WatcherHandle]`?

internal/lsp/server.go
```diff
@@ -262,9 +363,14 @@ func (s *Server) handleInitialize(req *lsproto.RequestMessage) error {
 }
 
 func (s *Server) handleInitialized(req *lsproto.RequestMessage) error {
+	if s.initializeParams.Capabilities.Workspace.DidChangeWatchedFiles != nil && *s.initializeParams.Capabilities.Workspace.DidChangeWatchedFiles.DynamicRegistration {
+		s.watcheEnabled = true
```
jakebailey: Typo in `watcheEnabled`

internal/lsp/server.go
andrewbranch: Oh yeah, I forgot we had something other than OrderedSet

jakebailey (approved):

internal/tsoptions/parsedcommandline.go
```diff
@@ -45,3 +63,32 @@ func (p *ParsedCommandLine) GetConfigFileParsingDiagnostics() []*ast.Diagnostic
...
+		return path == tspath.ToPath(f, comparePathsOptions.CurrentDirectory, comparePathsOptions.UseCaseSensitiveFileNames)
+	}) {
+		return true
+	}
+
+	if p.ConfigFile == nil {
+		return false
+	}
+
+	if slices.ContainsFunc(p.ConfigFile.configFileSpecs.validatedFilesSpec, func(f string) bool {
```
sheetalkamat: You also want to check if extension is valid to avoid matching js files when they may not be included https://github.com/microsoft/TypeScript/blob/main/src/server/editorServices.ts#L4494

internal/project/project.go
```diff
@@ -205,6 +235,95 @@ func (p *Project) LanguageService() *ls.LanguageService {
...
+		for dir, recursive := range wildcardDirectories {
+			result = append(result, fmt.Sprintf("%s/%s", dir, core.IfElse(recursive, recursiveFileGlobPattern, fileGlobPattern)))
+		}
+		return result
+	}
+	return nil
+}
+
+func (p *Project) getModuleResolutionWatchGlobs() (failedLookups map[tspath.Path]string, affectingLocaions map[tspath.Path]string) {
+	failedLookups = make(map[tspath.Path]string)
```
sheetalkamat, line 257: This may be ok for initial implementation but iterating over so many resolutions and so many failed lookups has been perf bottleneck. It was also memory issue for bigger code bases to store all the failed lookups as is and not watch the directories instead

internal/project/project.go
```diff
@@ -205,6 +235,95 @@ func (p *Project) LanguageService() *ls.LanguageService {
...
+	} else if updated {
+		p.log("Failed lookup watches updated:\n" + formatFileList(p.failedLookupsWatch.globs, "\t", hr))
+	}
+
+	if updated, err := p.affectingLocationsWatch.update(affectingLocationGlobs); err != nil {
+		p.log(fmt.Sprintf("Failed to update affecting location watch: %v", err))
+	} else if updated {
+		p.log("Affecting location watches updated:\n" + formatFileList(p.affectingLocationsWatch.globs, "\t", hr))
+	}
+}
```
sheetalkamat, line 306: Need missing file watches .. On that note currently we add root files that are "present" but we need to be adding all root files and let program handle it so that it can issue "missing file" diagnostic and we can add watch as well

internal/project/project.go
```diff
@@ -205,6 +235,95 @@ func (p *Project) LanguageService() *ls.LanguageService {
...
+// This could be a case of one of the following:
+//   - A file is being created that will be added to the project.
+//   - An affecting location was changed.
+//   - A file is being created that matches a watch glob, but is not actually
+//     part of the project, e.g., a .js file in a project without --allowJs.
+func (p *Project) onWatchEventForNilScriptInfo(fileName string) {
+	path := p.toPath(fileName)
+	if p.kind == KindConfigured {
+		if p.rootFileNames.Has(path) || p.parsedCommandLine.MatchesFileName(fileName, p.comparePathsOptions) {
+			p.pendingConfigReload = true
```
sheetalkamat: For now this is ok but it is unnecessary to reparse the "config file" and calculate everything if config file has not changed. Just getting new list of root file names is pattern matching and is usually faster.

sheetalkamat (changes requested):

internal/project/watch.go
```diff
@@ -0,0 +1,61 @@
...
+func (w *watchedFiles[T]) update(newData T) (updated bool, err error) {
+	newGlobs := w.getGlobs(newData)
+	w.data = newData
+	if slices.Equal(w.globs, newGlobs) {
+		return false, nil
+	}
+
+	w.globs = newGlobs
+	if w.watcherID != "" {
+		if err = w.client.UnwatchFiles(w.watcherID); err != nil {
```
sheetalkamat, line 41: Fyi. Removing watches and then adding again for unchanged file list in past has been pretty costly from perf perspective.

internal/tsoptions/parsedcommandline.go
andrewbranch: I didn’t port this line because I couldn’t understand the intent. It makes `isMatchedByConfig` return `false` if the file name is supported. So only _unsupported file names_ can go on to be checked against the include and exclude specs, which doesn’t make any sense to me.

internal/project/project.go
andrewbranch: Yeah, I want to do the simplest thing to start with and then optimize based on measurements

internal/project/watch.go
andrewbranch: This follows a `slices.Equal` check that bails out if the list is unchanged.

internal/tsoptions/parsedcommandline.go
andrewbranch: Updated with tests adapted from matchFiles.ts

internal/testutil/projecttestutil/projecttestutil.go
```diff
@@ -12,12 +12,15 @@ import (
 	"github.com/microsoft/typescript-go/internal/vfs/vfstest"
 )
 
+//go:generate go tool github.com/matryer/moq -stub -fmt goimports -pkg projecttestutil -out clientmock.go ../../project Client
```
jakebailey:
```suggestion
//go:generate go tool github.com/matryer/moq -stub -fmt goimports -pkg projecttestutil -out clientmock_generated.go ../../project Client
```

internal/tsoptions/parsedcommandline.go
```diff
@@ -2,20 +2,52 @@ package tsoptions
...
+	wildcardDirectoriesOnce sync.Once
+	wildcardDirectories     map[string]bool
+	extraFileExtensions     []fileExtensionInfo
+}
+
+// WildcardDirectories returns the cached wildcard directories, initializing them if needed
+func (p *ParsedCommandLine) WildcardDirectories() map[string]bool {
+	if p.wildcardDirectories != nil {
+		return p.wildcardDirectories
+	}
```
jakebailey, line 32: You could actually eliminate this branch entirely, as the Do call is already a branch that checks that this has been set.

andrewbranch: Not exactly—that’s how I had it originally, but `ReloadFileNamesOfParsedCommandLine` clones `wildcardDirectories`, but doesn’t copy the `sync.Once`. So the new ParsedCommandLine has pre-seeded `wildcardDirectories` but its `sync.Once` never ran.

jakebailey: Oh, I see, funky

andrewbranch: Open to other suggestions. Is there a safe way to copy the Once state?

jakebailey: There isn't, no; if it can be accessed atomically then it can't be traditionally copied by value.

sheetalkamat (approved):

PR #808, opened 2025-04-18
Remove possible allocation of `NodeList` by eliminating `IfElse`.

DanielRosenwasser:
Just a drive-by change. I'm not sure if the compiler is "smart enough" to get rid of the allocation, but it felt better to be more explicit anyway.


jakebailey (approved):
The compiler is smart enough, but it would be an illegal transform because it would change the runtime behavior

PR #812, opened 2025-04-21
More LSP gen cleanup, some changes for bidirectional messages

jakebailey:
A bunch of refactors and cleanups from the code gen as I work to get async bidirectional LSP communication working. Most of this is cleaning up cruft leftover from me testing out agent mode, but also fixes some bugs surrounding `any` decoding, method generation / duplication, which were preexisting.


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -73,27 +72,9 @@ func (r *RequestMessage) UnmarshalJSON(data []byte) error {
 
 	r.ID = raw.ID
 	r.Method = raw.Method
-	if r.Method == MethodShutdown || r.Method == MethodExit {
-		// These methods have no params.
-		return nil
-	}
-
-	if strings.HasPrefix(string(r.Method), "@ts/") {
```
gabritto, line 81: For my own understanding: were we planning on having `@ts/...`/custom requests and changed our minds, or something else?

gabritto (approved):

internal/lsp/lsproto/jsonrpc.go
jakebailey: This is leftover from the API testing; I talked to Andrew about it last week and he said it was fine to remove.

If we are going to tack on our own requests, we'll want to then start patching the meta model to have those custom messages be generated too.

PR #819, opened 2025-04-23
Add `esnext` transform

rbuckton:
This adds the ESNext transform, which handles the downlevel transformation of `using` and `await using` declarations.

This PR is dependent on #815, so most feedback related to ast/ast.go, printer/emitcontext.go, and printer/factory.go should be addressed there.


internal/transformers/namedevaluation.go
```diff
@@ -0,0 +1,556 @@
...
+ */
+func isClassNamedEvaluationHelperBlock(emitContext *printer.EmitContext, node *ast.Node) bool {
+	if !ast.IsClassStaticBlockDeclaration(node) || len(node.AsClassStaticBlockDeclaration().Body.AsBlock().Statements.Nodes) != 1 {
+		return false
+	}
+
+	statement := node.AsClassStaticBlockDeclaration().Body.AsBlock().Statements.Nodes[0]
+	if ast.IsExpressionStatement(statement) {
+		expression := statement.AsExpressionStatement().Expression
+		if emitContext.IsCallToHelper(expression, "___setFunctionName") {
```
jakebailey: Three underscores here; is this a typo?

jakebailey: Huh, the original code has this too. Is this due to the old compiler's different symbol name escaping mechanism? The func generally seems like it should be just `__setFunctionName`, so...

rbuckton: Ah yes, this was due to the underscore-escaping in Strada.

jakebailey (approved):

PR #820, opened 2025-04-23
Fix `findRightmostValidToken`

gabritto:
Turns out, for `findPrecedingToken`/`findRightmostValidToken`, we have to look at every token that is not part of the AST and that occurs between "empty" AST nodes.

A simplified illustration of the problem present in the test I added:

```ts
variable.|

function foo() { ... }
```

With position at `|`, we should return the dot token, because it is the rightmost valid (non-empty) token that precedes the position.
But to find it, we have to look at tokens that occur between the identifier node `variable`, and the empty identifier node that the parser adds after the dot to construct a valid property access expression.

Before this PR, we were only looking at tokens after the last visited node, i.e. after the empty identifier node.
Now, we look at every token that occurs in between visited nodes, after the rightmost non-empty valid node:

| Rightmost valid node | ... tokens ... | Invalid node 1 | ... more tokens ... | Invalid node n | ... more tokens ...

What this means for the example is that the rightmost non-empty valid node is identifier `variable`, and to the right of that we have the empty identifier node, and we look in between those two nodes for tokens:

| Identifier 'variable' | ... dot token ... | Identifier '' | ...

### DIFF
--- a/internal/astnav/tokens.go
+++ b/internal/astnav/tokens.go
@@ -510,9 +510,9 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
 			}
 			scanner := scanner.GetScannerForSourceFile(sourceFile, startPos)
 			var tokens []*ast.Node
-			for i := range rightmostVisitedNodes {
+			for _, visitedNode := range rightmostVisitedNodes {
 				// Trailing tokens that occur before this node.
-				for startPos < min(rightmostVisitedNodes[i].Pos(), position) {
+				for startPos < min(visitedNode.Pos(), position) {
 					tokenStart := scanner.TokenStart()
 					if tokenStart >= position {
 						break
@@ -524,7 +524,7 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
 					tokens = append(tokens, sourceFile.GetOrCreateToken(token, tokenFullStart, tokenEnd, n))
 					scanner.Scan()
 				}
-				startPos = rightmostVisitedNodes[i].End()
+				startPos = visitedNode.End()
 				scanner.ResetPos(startPos)
 			}
 			// Trailing tokens after last visited node.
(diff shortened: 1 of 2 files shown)


internal/astnav/tokens_test.go
```diff
@@ -83,7 +83,7 @@ func baselineTokens(t *testing.T, testName string, includeEOF bool, getTSTokens
 			fileText, err := os.ReadFile(fileName)
 			assert.NilError(t, err)
 
-			positions := make([]int, len(fileText))
+			positions := make([]int, len(fileText)+core.IfElse(includeEOF, 1, 0))
```
gabritto, line 86: Drive-by fix, I forgot to do this in my previous PR.

jakebailey:
Is this difference just a problem with the ported code, or was this some bug in the old code too? I assume the former?

gabritto:
> Is this difference just a problem with the ported code, or was this some bug in the old code too? I assume the former?

Yep, a problem with the ported code. The old code doesn't have this problem because it uses `node.getChildren()`, which automatically includes tokens that are not part of the AST, so it can just iterate through children that way.

The added complexity of the new implementation is because it has to simulate the old `findPrecedingToken` while also doing the extra work of considering tokens that are not part of the AST, since we are avoiding a port of `node.getChildren()` for now, because it was inefficient.

internal/astnav/tokens.go
```diff
@@ -485,19 +497,37 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
...
-			if rightmostVisitedNode != nil {
-				startPos = rightmostVisitedNode.End()
+			if rightmostValidNode != nil {
+				startPos = rightmostValidNode.End()
 			} else {
 				startPos = n.Pos()
 			}
 			scanner := scanner.GetScannerForSourceFile(sourceFile, startPos)
 			var tokens []*ast.Node
+			for i := range rightmostVisitedNodes {
```
jakebailey: Seems like this loop could also include the variable instead of accessing `rightmostVisitedNodes[i]` below, right?

jakebailey (approved):

PR #822, opened 2025-04-24
Multiple fixes to checker

ahejlsberg:
In this PR:

* Multiple small fixes motivated by tests.
* We now merge related information for otherwise identical diagnostics (the old compiler simply discards information here).
* Processing of `@ts-ignore` and `@ts-expect-error` now closely matches old compiler.
* Elaboration logic for JSX children ported (JSX port is now 100% complete).


internal/compiler/program.go
```diff
@@ -394,9 +394,42 @@ func isCommentOrBlankLine(text string, pos int) bool {
...
+	j := 0
+	for i < len(diagnostics) {
+		d := diagnostics[i]
+		n := 1
+		for i+n < len(diagnostics) && ast.EqualDiagnosticsNoRelatedInfo(d, diagnostics[i+n]) {
+			n++
+		}
+		if n > 1 {
+			var relatedInfos []*ast.Diagnostic
+			for k := range n {
```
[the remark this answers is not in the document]
ahejlsberg: Hmm, Copilot needs to catch up on Go's latest features.

internal/ast/diagnostic.go
```diff
@@ -60,6 +60,18 @@ func (d *Diagnostic) AddRelatedInfo(relatedInformation *Diagnostic) *Diagnostic
 	return d
 }
 
+func (d *Diagnostic) Clone() *Diagnostic {
+	return &Diagnostic{
```
jakebailey: Seems like this could just be:

```go
cpy := *d
return &cpy
```

Since no fields have been omitted.

internal/diagnostics/diagnostics.go
```diff
@@ -54,3 +54,15 @@ func (m *Message) Format(args ...any) string {
 	}
 	return text
 }
+
+func FormatMessage(m *Message, args ...any) *Message {
+	return &Message{
```
jakebailey: Probably could be a `m.Clone()` followed by a write to `text`, right?

jakebailey (approved):
LGTM other than the diagnostic cloning suggestions.

internal/ast/diagnostic.go
ahejlsberg: Agreed.

internal/diagnostics/diagnostics.go
ahejlsberg: I don't think we have a `Clone` method on `diagnostics.Message`. But it could just be an assignment copy followed by a write to the `text` property

jakebailey: Ah, right, I was mistaking this for the type you just added `Clone` to above.

ryoeuyo (approved):

PR #824, opened 2025-04-25
post CommonJS PR cleanup

sandersn:
1. Move tag reparsing from jsdoc.go to reparser.go
2. Improve a couple of names in the binder and reparser.
3. Restore mistakenly deleted contextual type for export=.
4. Remove TODO in nameresolver.

I tried addressing the TODO and it makes the result more correct but less useful: when a globally declared `module` or `exports` overrides the fallback we provide, and `exports: { [s: string]: any }` (which I believe is the node definition), `exports.x: any` when referred to inside the exporting file. Pro: goto-def works on `exports`. Con: `exports.x` loses type information it could have had.

I considered moving `require` up above the global resolution, but it's not useful to have a good type for `require` and I'd like to keep goto-def going to types/node if present.


internal/checker/checker.go
```diff
@@ -27158,6 +27158,8 @@ func (c *Checker) getContextualType(node *ast.Node, contextFlags ContextFlags) *
 		return c.getContextualType(parent, contextFlags)
 	case ast.KindSatisfiesExpression:
 		return c.getTypeFromTypeNode(parent.AsSatisfiesExpression().Type)
+	case ast.KindExportAssignment:
+		return c.tryGetTypeFromTypeNode(parent)
```
jakebailey, line 27162: No test change?

jakebailey (approved):

internal/checker/checker.go
sandersn: Nope!

sandersn: My theory, which I haven't confirmed: in Strada, this code was added by analogy with

```ts
/** @type {X} */
exports.x = (x,y,z) => { }
```

(which is currently broken in Corsa, it's on my list to fix)

But it does nothing because there's no type annotation on `export=`.

PR #826, opened 2025-04-25
Fix typedef binding with CJS `exports=`

sandersn:
With this PR, top-level typedefs in file with commonjs exports (`module.exports=x` or `exports=x`) are no longer added as exports of the file. They're added as exports of whatever is `export=`. I'm not sure that I declared the local and export='d symbols in the right way, so I'd appreciate expert opinions there.

The big change compared to Strada is that **only** top-level typedefs are exported. But it never made sense for these two scoped type aliases to be exported, let alone *both*:

```js
function one() {
  /** @typedef {string} T */
  /** @type {T} */
  var s = 's'
}
function two() {
  /** @typedef {number} T */
  /** @type {T} */
  var n = 1
}
/* @type {T} */
var error = "I AM ERROR"
```

It's especially weird that a scoped alias would be inaccessible at the top-level, but then accessible in a different file.


internal/binder/binder.go
```diff
@@ -686,8 +710,14 @@ func (b *Binder) bind(node *ast.Node) bool {
 		b.bindBlockScopedDeclaration(node, ast.SymbolFlagsInterface, ast.SymbolFlagsInterfaceExcludes)
 	case ast.KindCallExpression:
 		b.bindCallExpression(node)
-	case ast.KindTypeAliasDeclaration, ast.KindJSTypeAliasDeclaration:
+	case ast.KindTypeAliasDeclaration:
 		b.bindBlockScopedDeclaration(node, ast.SymbolFlagsTypeAlias, ast.SymbolFlagsTypeAliasExcludes)
+	case ast.KindJSTypeAliasDeclaration:
+		if b.file.AsNode() == b.container {
```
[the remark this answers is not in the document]
sandersn: So weird, this is *exactly* the function you hallucinated when I first tried to write this code! (I still don't think it's a good idea.)

internal/binder/binder.go
```diff
@@ -124,9 +125,32 @@ func bindSourceFile(file *ast.SourceFile, options *core.SourceFileAffectingCompi
...
+	if b.delayedTypeAliases == nil {
+		return
+	}
+	if b.file.Symbol != nil {
+		if exportEq := b.file.Symbol.Exports[ast.InternalSymbolNameExportEquals]; exportEq != nil && b.file.CommonJSModuleIndicator != nil {
+			for _, node := range b.delayedTypeAliases {
+				b.declareSymbol(ast.GetSymbolTable(&exportEq.Exports), exportEq /*parent*/, node, ast.SymbolFlagsTypeAlias, ast.SymbolFlagsTypeAliasExcludes)
+				b.declareSymbol(ast.GetLocals(b.file.AsNode()), b.file.Symbol, node, ast.SymbolFlagsTypeAlias, ast.SymbolFlagsTypeAliasExcludes)
+			}
+			return
```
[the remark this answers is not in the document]
sandersn: Pretty sure that `putBinder` does this with `*b = Binder{bindFunc: b.bindFunc}` but paging @jakebailey to confirm.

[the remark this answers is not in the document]
jakebailey: Yes, it's zero'd out. We should go and remove any explicit clearing we have leftover on the parser, scanner, binder, etc, since they are now pooled and zero'd.

jakebailey (approved):

internal/binder/binder.go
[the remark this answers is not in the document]
sandersn: Good, because I wrote and then deleted quite a bit of clearing code for this PR before I convinced myself it wasn't necessary. So copilot isn't completely incompetent here.

internal/binder/binder.go
```diff
@@ -124,9 +125,32 @@ func bindSourceFile(file *ast.SourceFile, options *core.SourceFileAffectingCompi
...
 
+// top-level typedef binding is delayed because it changes based on whether `module.exports = x` is bound
+func (b *Binder) delayedBindJSDocTypedefTag() {
+	if b.delayedTypeAliases == nil {
+		return
+	}
+	if b.file.Symbol != nil {
+		if exportEq := b.file.Symbol.Exports[ast.InternalSymbolNameExportEquals]; exportEq != nil && b.file.CommonJSModuleIndicator != nil {
+			for _, node := range b.delayedTypeAliases {
+				b.declareSymbol(ast.GetSymbolTable(&exportEq.Exports), exportEq /*parent*/, node, ast.SymbolFlagsTypeAlias, ast.SymbolFlagsTypeAliasExcludes)
```
weswigham: This works for one level of indirection to a local namespace-y thing, but what about an aliased namespacey thing? Eg
```js
const mod = require("./mod.js")
/**
* @typedef {string} T
*/
module.exports = mod
```
or, abusing a ts construct to do it locally,
```ts
namespace ns {
  export namespace inner {}
}
import mod = ns.inner
/**
* @typedef {string} T
*/
module.exports = mod
```
`mod`'s symbol will be an alias, and this will just patch the typedefs onto the alias symbol, and not the alias symbol's ultimate target, which, dollars to donuts, means it's going to effectively go missing without extra lookup logic in the checker, similar to what we used to have. 🥹

sandersn: Even simpler, this also reproduces the problem.

```js
function f() { }
/** @typedef {string} T */
module.exports = f
```

internal/checker/checker.go
```diff
@@ -14545,9 +14545,24 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 }
 
 func (c *Checker) resolveQualifiedName(name *ast.Node, left *ast.Node, right *ast.Node, meaning ast.SymbolFlags, ignoreErrors bool, dontResolveAlias bool, location *ast.Node) *ast.Symbol {
-	namespace := c.resolveEntityName(left, ast.SymbolFlagsNamespace, ignoreErrors, false /*dontResolveAlias*/, location)
+	namespace := c.resolveEntityName(left, ast.SymbolFlagsNamespace, true /*ignoreErrors*/, false /*dontResolveAlias*/, location)
 	if namespace == nil || ast.NodeIsMissing(right) {
-		return nil
+		var immediate *ast.Symbol
+		alias := c.resolveEntityName(left, ast.SymbolFlagsAlias, true /*ignoreErrors*/, true /*dontResolveAlias*/, location)
```
weswigham: Hm. So if the resolution of the LHS of a qualified name fails, we try to look it up with only one level of aliasing? This seems... weird? While this might work in some scenarios, if `getExportsOfSymbol` doesn't return the merged in aliases, declaration emit is going to struggle to figure out how to name these typedefs, as traversing the exports is how symbols are named. Basically, we'd like to interpret

```
/**
* @typedef {number} Foo
*/
module.exports = name
```
as
```ts
export * from name // even though this isn't a thing in ESM (yet)
export type Foo = number
```

I think if we want to handle this in a non-jank way, we need to bind a `module.exports = name` in a file with typedefs not as an alias but instead as a (new?) namespacey symbol kind that has logic in `getExportsOfSymbol` to merge together the typedef set and the resolved, late-bound alias's set. (Once again, a callback to the `.cjsSymbolMerged` logic we used to have)

I'm also honestly a lil unsure what a syntactically driven declaration emit for this is.

weswigham: Maybe we just add support for parsing/checking `export * from name` in declaration files and actually do that transform on reparse? In theory it'll be added to esm alongside the module declarations proposal, probably.

sandersn: I tried this in 1cffb6e, but it results in just as many hacks, in the same places, as the previous commit. That's because `export *` isn't callable, so I need the same kind of stop-in-the-middle hack to return the symbol of a function initialiser, plus an additional hack to look for and add call signatures when resolving members of the `export *`.

So I reverted it. Can you come up with a way to improve the current hack? I think one-level resolution is fine even if not; commonjs usage is ossified and I think the support can be accordingly restrictive.

weswigham: > So I reverted it. Can you come up with a way to improve the current hack? 

Honestly? https://github.com/microsoft/typescript-go/commit/1cffb6ec0ec1329f6eac0012b65dc7f1b7b48b56 but go a little bit farther - with slightly more accompanying checker complexity - rather than tossing it into the `export*` symbol table member and fishing it out, make a new special symbol table member for cjs export assignments that does the whole "copy the whole target and add the other exports on, too" rigamarole on export table lookup, complete with custom conflict resolution logic. Unfortunately, it definitely doesn't behave like any other AST construct we have at present because it's an `export=` and a `export*` symbol all rolled into one (or perhaps a mutable `export=`).

I think a lot of why the strada implementation feels jank isn't because of the weird expression binding, but rather is because of our refusal to handle it like it's its own thing and instead patching its' behaviors on top of the existing ones. I'd rather handle it explicitly as its' own thing through the whole process and lower my worries that we're either shoehorning it into something it's not quite equivalent to and breaking a bunch of legacy code (which is all this is here for! "new jsdoc code" is almost an oxymoron) or patching exceptions on in an ad-hoc way like we did in strada.

weswigham: Or the other option is to loosen how we handle `export=` internally to behave like the cjs `module.exports=` in strada - then it could be equivalent to that, I guess. There's honestly very little reason to disallow
```ts
export = class Foo {}
export const a = 1
```
in cjs-targeting typescript other than historic strictness and taste (since it transpiles to the same declarations we're checking on the input side here).

jakebailey:
Old PR, but is this still needed?

sandersn:
Yes, it's still a problem when a typedef is in a file with `module.exports=`. I should look at this again, because the declaration emit and reparser are finished now.

sandersn:
What does a failure in the generate shtage mean? There's now an unexpected diff, but I don't know what causes diffs to appear.

jakebailey:
```
 internal/fourslash/_scripts/failingTests.txt                     | 1 -
 internal/fourslash/tests/gen/javascriptModulesTypeImport_test.go | 2 +-
```

means you fixed one failing test, and should run `npm run updatefailing` and commit that result

internal/checker/checker.go
```diff
@@ -15247,10 +15254,10 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 		symbol = c.getMergedSymbol(c.resolveName(resolveLocation, name.AsIdentifier().Text, meaning, message, true /*isUse*/, false /*excludeGlobals*/))
 	case ast.KindQualifiedName:
 		qualified := name.AsQualifiedName()
-		symbol = c.resolveQualifiedName(name, qualified.Left, qualified.Right, meaning, ignoreErrors, dontResolveAlias, location)
+		symbol = c.resolveQualifiedName(name, qualified.Left, qualified.Right, meaning, ignoreErrors, location)
```
sandersn, line 15360: this is a driveby fix--dontResolveAlias wasn't used even before this PR

internal/checker/checker.go
```diff
@@ -15624,12 +15640,26 @@ func (c *Checker) getExportsOfModuleWorker(moduleSymbol *ast.Symbol) (exports as
...
 	moduleSymbol = c.resolveExternalModuleSymbol(moduleSymbol, false /*dontResolveAlias*/)
 	exports = visit(moduleSymbol, nil, false)
 	if exports == nil {
 		exports = make(ast.SymbolTable)
 	}
+	// A CommonJS module defined by an 'export=' might also export JSTypeAliasDeclarations with
+	if originalModule != nil && len(originalModule.Exports) > 1 {
+		for _, symbol := range originalModule.Exports {
+			if symbol.Flags&ast.SymbolFlagsType != 0 && symbol.Name != ast.InternalSymbolNameExportEquals {
+				exports[symbol.Name] = symbol
```
ahejlsberg, line 15753: Since we're not merging symbols here, shouldn't we only promote the `JSTypeAliasDeclaration` symbols when they have unique names?

ahejlsberg: Here, it would overwrite an existing symbol with that name.

sandersn: I think this is only possible in Typescript, because nothing in JS creates namespaces. You could get a conflict with `/** @typedef {1} x */ module.exports.x = 1` but this only works with `module.exports=`.

However, we may allow this in TS for 7.1 or later, so it's a good idea to add a check.

internal/checker/checker.go
jakebailey: In Strada I see `dontResolveAlias` being given both true and false all over the place. Are you sure we should remove it?

ahejlsberg (approved):

internal/checker/checker.go
sandersn: Yes. As best I can tell, it's a translation artifact--resolveQualifiedName is a function extracted from resolveEntityName -- it reads way better in Go and probably would in TS too -- but its code never actually refers to dontResolveAlias, even in Strada. It's only checked once in resolveEntityName, actually, right at the end, when deciding whether to resolveAlias on the symbol we just resolved.

PR #831, opened 2025-04-28
Reparse `@import` as synthetic type-only import declaration

sandersn:
It is slightly stricter than Strada's treatment because Corsa now treats the synthetic node *exactly* as a type-only declaration.


testdata/baselines/reference/submoduleAccepted/conformance/importTag16.types.diff
```diff
@@ -0,0 +1,12 @@
+--- old.importTag16.types
++++ new.importTag16.types
+@@= skipped -12, +12 lines =@@
+  * @param {I} b
+  */
+ export function foo(a, b) {}
+->foo : (a: Foo, b: I) => void
+->a : Foo
++>foo : (a: default, b: I) => void
```
sandersn, line 9: I don't know how typescript behaves, but I've changed so little code that I guess that `a: default` is closer to TS than `a: Foo` was.

testdata/baselines/reference/submoduleAccepted/conformance/importTag17.errors.txt.diff
```diff
@@ -3,17 +3,21 @@
...
-@@= skipped -25, +25 lines =@@
+@@= skipped -21, +21 lines =@@
      
-     /**
-      * @returns { Import }
+ ==== /a.js (2 errors) ====
+     /** @import { Import } from 'foo' with { 'resolution-mode': 'import' } */
 +                  ~~~~~~
-+!!! error TS2304: Cannot find name 'Import'.
++!!! error TS2305: Module '"/node_modules/@types/foo/index"' has no exported member 'Import'.
```
sandersn, line 16: this looks wrong.

sandersn: But it is broken with `import type { Import } from 'foo' with { 'resolution-mode': 'import' }` as well, so I think it's a missing feature of module resolution, not of this PR.

sandersn:
@a-tarasyuk you wrote the original implementation of `@import`, so you might be interested to see how it looks in Corsa.

internal/ast/ast.go
```diff
@@ -4036,13 +4034,21 @@ type ImportDeclaration struct {
...
+
+func (f *NodeFactory) NewImportDeclaration(modifiers *ModifierList, importClause *ImportClauseNode, moduleSpecifier *Expression, attributes *ImportAttributesNode) *Node {
+	return f.newImportOrJSImportDeclaration(KindImportDeclaration, modifiers, importClause, moduleSpecifier, attributes)
+}
+
+func (f *NodeFactory) NewJSImportDeclaration(modifiers *ModifierList, importClause *ImportClauseNode, moduleSpecifier *Expression, attributes *ImportAttributesNode) *Node {
+	return f.newImportOrJSImportDeclaration(KindJSImportDeclaration, modifiers, importClause, moduleSpecifier, attributes)
 }
 
 func (f *NodeFactory) UpdateImportDeclaration(node *ImportDeclaration, modifiers *ModifierList, importClause *ImportClauseNode, moduleSpecifier *Expression, attributes *ImportAttributesNode) *Node {
```
rbuckton, line 4064: `VisitEachChild` is going to have issues if you don't differentiate by `Kind` in `UpdateImportDeclaration`, since it would turn a `JSImportDeclaration` into a normal `ImportDeclaration`.

internal/printer/printer.go
```diff
@@ -3953,7 +3953,7 @@ func (p *Printer) emitStatement(node *ast.Statement) {
 		p.emitNamespaceExportDeclaration(node.AsNamespaceExportDeclaration())
 	case ast.KindImportEqualsDeclaration:
 		p.emitImportEqualsDeclaration(node.AsImportEqualsDeclaration())
-	case ast.KindImportDeclaration:
```
rbuckton: Printing `KindJSImportDeclaration` should either panic (if you expect it to always be transformed), or have its own distinct emit.

internal/transformers/commonjsmodule.go
```diff
@@ -63,6 +63,8 @@ func (tx *CommonJSModuleTransformer) visitTopLevel(node *ast.Node) *ast.Node {
 	switch node.Kind {
 	case ast.KindImportDeclaration:
 		node = tx.visitTopLevelImportDeclaration(node.AsImportDeclaration())
+	case ast.KindJSImportDeclaration:
```
rbuckton: IMO, these should be handled by something like `typeeraser.go` or `importellision.go` so that other transformers shouldn't have to deal with special cases like `JSImportDeclaration`. Handling them in the module transforms means they have to exist as part of all other transforms first, and its better if we limit syntax in those transforms to ones expected by that edition of JS.

internal/transformers/esmodule.go
```diff
@@ -39,6 +39,8 @@ func (tx *ESModuleTransformer) visit(node *ast.Node) *ast.Node {
 		node = tx.visitSourceFile(node.AsSourceFile())
 	case ast.KindImportDeclaration:
 		node = tx.visitImportDeclaration(node.AsImportDeclaration())
+	case ast.KindJSImportDeclaration:
+		node = nil
```
rbuckton: This should be handled in an earlier transform.

internal/transformers/esnext.go
```diff
@@ -368,6 +368,7 @@ func (tx *ESNextTransformer) transformUsingDeclarations(statementsIn []*ast.Stat
 
 		switch node.Kind {
 		case ast.KindImportDeclaration,
+			ast.KindJSImportDeclaration,
```
rbuckton: This should be handled in an earlier transform.

internal/transformers/externalmoduleinfo.go
```diff
@@ -47,11 +47,12 @@ func (c *externalModuleInfoCollector) collect() *externalModuleInfo {
 	hasImportDefault := false
 	for _, node := range c.sourceFile.Statements.Nodes {
 		switch node.Kind {
-		case ast.KindImportDeclaration:
+		case ast.KindImportDeclaration, ast.KindJSImportDeclaration:
```
rbuckton: Is this necessary? Aren't these imports type-only and just elided?

internal/transformers/importelision.go
```diff
@@ -28,7 +28,7 @@ func (tx *ImportElisionTransformer) visit(node *ast.Node) *ast.Node {
 			return nil
 		}
 		return tx.visitor.VisitEachChild(node)
-	case ast.KindImportDeclaration:
+	case ast.KindImportDeclaration, ast.KindJSImportDeclaration:
```
rbuckton: This should be handled by typeeraser.go and should never reach this point

internal/transformers/importelision.go
```diff
@@ -166,7 +166,7 @@ func (tx *ImportElisionTransformer) isElisionBlocked(node *ast.Node /*ImportDecl
 	}
 
 	switch node.Kind {
-	case ast.KindImportDeclaration:
+	case ast.KindImportDeclaration, ast.KindJSImportDeclaration:
```
rbuckton: This shouldn't be necessary as it should be handled by typeeraser.go

internal/transformers/typeeraser.go
```diff
@@ -255,7 +255,7 @@ func (tx *TypeEraserTransformer) visit(node *ast.Node) *ast.Node {
 		}
 		return tx.visitor.VisitEachChild(node)
 
-	case ast.KindImportDeclaration:
+	case ast.KindImportDeclaration, ast.KindJSImportDeclaration:
```
rbuckton: This should have its own branch and should just be elided.

rbuckton (changes requested):

testdata/baselines/reference/submoduleAccepted/conformance/importTag16.types.diff
andrewbranch: Isn’t this saying that TS has `a: Foo` where tsgo has `a: default`? We definitely shouldn’t be calling that type `default`.

sandersn: Oh, yeah, I misread this as diffing from the previous version of Strada, but the fact that it's a new diff file means that the type output was identical before -- even though the error output was different.

sandersn: This is a limitation of our type printer. Here's the equivalent TS output, which is wrong in the same way:

```
=== b.ts ===
import type Foo, { type I } from "./a"
>Foo : default
>I : I

/**
 */
export function foo(a: Foo, b: I) {}
>foo : (a: default, b: I) => void
>a : default
>b : I
```

internal/transformers/externalmoduleinfo.go
sandersn: Yes, they are type-only. In general KindJS* should only affect .d.ts emit since they're generated from real statements in .js files.

gabrielluizsf (approved):

internal/ast/ast.go
```diff
@@ -4036,18 +4024,26 @@ type ImportDeclaration struct {
 	Attributes      *ImportAttributesNode // ImportAttributesNode. Optional
 }
 
-func (f *NodeFactory) NewImportDeclaration(modifiers *ModifierList, importClause *ImportClauseNode, moduleSpecifier *Expression, attributes *ImportAttributesNode) *Node {
+func (f *NodeFactory) newImportOrJSImportDeclaration(kind Kind, modifiers *ModifierList, importClause *ImportClauseNode, moduleSpecifier *Expression, attributes *ImportAttributesNode) *Node {
```
rbuckton, line 4047: minor nit: I think `newTypeAliasOrJSTypeAliasDeclaration` reads better than `newTypeOrJSTypeAliasDeclaration`.

rbuckton (approved):

internal/ast/kind.go
```diff
@@ -381,6 +381,7 @@ const (
 	KindJSTypeAliasDeclaration
 	KindJSExportAssignment
 	KindCommonJSExport
+	KindJSImportDeclaration
```
andrewbranch, line 384: Since we don’t yet have codegen in place, can you update `_packages/ast/src/syntaxKind.ts` and `_packages/ast/src/syntaxKind.enum.ts`?

andrewbranch (approved):
Looks good after the _packages/ast update

internal/ast/ast.go
sandersn: Thanks, that does read much better.

PR #811, opened 2025-04-21
Basic completion structure + dot completions

gabritto:
This PR introduces the basic completion workflow (`getCompletionData`, `completionInfoFromData`, etc), and dot completions. Global completions remain unimplemented, i.e. we're not collecting symbols for them.

I haven't tested the implementation yet beyond a very basic test, and some helpers are unimplemented, but I'd like to get the code in because there's helper code that may be shared by other services that are also being ported now.

PR #833, opened 2025-04-28
Port more of member completions + tests

gabritto:
Follow-up of #811.
In this PR:
- More member completions unit tests.
- Ported code from the vscode TS extension that computed replacement range (i.e. LSP text edit) and filter text.
- Implemented `findChildOfKind`, which was needed for the tuple completion case.

The code in `ls/testsetup.go` is temporary while we don't have real fourslash test infra, and is used for parsing markers in the unit tests. I got it from @navya9singh's signature help branch. I also tested the scenarios in the unit tests in vscode.


internal/ls/testsetup.go
```diff
@@ -0,0 +1,179 @@
+package ls
```
andrewbranch: Can this go in `ls_test`?

internal/ls/testsetup.go
```diff
@@ -0,0 +1,179 @@
...
+	sourceColumn   int
+}
+
+type TestFileInfo struct { // for FourSlashFile
+	Filename string
+	// The contents of the file (with markers, etc stripped out)
+	Content string
+}
+
+func parseFileContent(content string, filename string, markerMap *map[string]Marker, markers *[]Marker) TestFileInfo {
```
andrewbranch: Pointers to maps and slices look fishy

andrewbranch (approved):
Looks good except for `testsetup`, which I guess we should address now if this is going to be ready to merge before @navya9singh’s branch.

internal/ls/testsetup.go
gabritto: Moved it to its own package in `testutil`. I'd have to rename the file to something like `ls_test.go` to have it be part of `ls_test`.

internal/ls/testsetup.go
gabritto: Fixed the pointer to map, but the pointer to slice is needed because we append to the slice.

andrewbranch (approved):

Knightsaket007:
😁

PR #835, opened 2025-04-29
Add support for extendedDiagnostics

maschwenk:


cmd/tsgo/main.go
```diff
@@ -295,7 +298,44 @@ func runMain() int {
 	}
 	stats.add("Total time", totalTime)
 
-	stats.print()
+	if opts.devel.extendedDiagnostics != "" {
```
jakebailey: This isn't really the same thing as the original tsc flag; I would not personally want to add this flag, at least not under this name.

Why not just parse out the current output, which is what needs to be done with old tsc as well?

cmd/tsgo/diagnostics.go
```diff
@@ -0,0 +1,107 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+// max returns the larger of x or y
+func max(x, y int) int {
```
jakebailey: This is already built into the language.

cmd/tsgo/main.go
maschwenk: > Why not just parse out the current output, which is what needs to be done with old tsc as well?

For exactly that reason, the code we wrote on our side to parse it feels pretty goofy and brittle. Happy to think of other names! The context for me is that we send this information to Datadog, so it's pretty nice to not have to maintain that code that parses out of stdout because we have to scan through potentially thousands of lines of errors to find what were looking for. When we were working on reducing our tsc memory usage, we were even keeping track of these metrics for builds with failures.

cmd/tsgo/diagnostics.go
maschwenk: Sorry, this is is admittedly AI slop

cmd/tsgo/main.go
jakebailey: Needing to send this information somewhere is making me more nervous than before. If we add `devel` flags here, they're temporary until we can discuss and agree on a real flag that will also be present in tsc. tsgo itself is going to go away and be replaced tsc (with extra flags), so I would be wary about adding a flag that we'll just remove later...

PR #838, opened 2025-04-30
add a couple of missing options to be parsed

srabraham:
As things are, tsgo basically ignores these fields in a tsconfig, so rewriting imports doesn't work.

Could there be a "default: panic" or similar in this switch? Or some test? That'd presumably help make this kind of bug less easy to come up again.

**EDIT:** This PR is mutually incompatible with my other one, #839. If #839 makes sense, I suggest you merge that one and close #838.


jakebailey:
I think I'd prefer #839, or at least some other test for this.

PR #839, opened 2025-04-30
add missing compiler options, add a test to help prevent this in future

### LINKED ISSUE #842: Some CompilerOptions don't work due to deficient `switch-case` in parsinghelpers.go
I noticed that RewriteRelativeImportExtensions does not work when I use `tsgo` locally, even though all the code seemed to be in place in this repo. The problem is in a big switch-case in parsinghelpers.go, in which that field (and others) aren't getting converted. I put up a couple of mutually exclusive options as PRs to address this: #839 or #838.

https://github.com/microsoft/typescript-go/blob/62c88967c1a5402dbdc17c3d9ebd69fc376c8c24/internal/tsoptions/parsinghelpers.go#L143-L151

srabraham:
This is a more comprehensive fix than #838. Future work would involve doing the same sort of tests for non-Tristate types.

This isn't exactly a pretty test, but if we take for a given that the big switch-case needs to exist as is, then you're better off having a nasty test like this than not having it.

Fixes #842


srabraham:

internal/tsoptions/parsinghelpers.go
```diff
@@ -185,6 +187,12 @@ func ParseCompilerOptions(key string, value any, allOptions *core.CompilerOption
 		allOptions.DeclarationMap = parseTristate(value)
 	case "declaration":
 		allOptions.Declaration = parseTristate(value)
+	case "downlevelIteration":
+		allOptions.DownlevelIteration = parseTristate(value)
+	case "emitDeclarationOnly":
+		allOptions.EmitDeclarationOnly = parseTristate(value)
+	case "eSModuleInterop":
```
jakebailey:
```suggestion
	case "esModuleInterop":
```

jakebailey: This means the test will fail, since these are not just "first letter lowercase".

What I would do in your test instead is use reflection to extract out the `json` struct field tag.

internal/tsoptions/parsinghelpers_test.go
```diff
@@ -0,0 +1,36 @@
+package tsoptions
+
+import (
+	"github.com/microsoft/typescript-go/internal/core"
+	"reflect"
+	"testing"
+	"unicode"
+)
+
+func TestParseCompilerOptions_noMissingTristates(t *testing.T) {
```
jakebailey:
```suggestion
func TestParseCompilerOptionNoMissingTristates(t *testing.T) {
```

We don't typically name tests like this.

jakebailey:
For the test, we may want to instead just loop over `OptionsDeclarations` instead.

internal/tsoptions/parsinghelpers.go
srabraham: oh, /facepalm...I should've noticed those keys come from the json tags. Done

internal/tsoptions/parsinghelpers_test.go
srabraham: Fixed

srabraham:
> For the test, we may want to instead just loop over `OptionsDeclarations` instead.

I spent some time in the code trying to figure out what you mean by this, but I couldn't get there

jakebailey:
> > For the test, we may want to instead just loop over `OptionsDeclarations` instead.
> 
> I spent some time in the code trying to figure out what you mean by this, but I couldn't get there

That's fair. I can send a follow-up PR to do this, building on your test.

srabraham:
> > > For the test, we may want to instead just loop over `OptionsDeclarations` instead.
> > 
> > 
> > I spent some time in the code trying to figure out what you mean by this, but I couldn't get there
> 
> That's fair. I can send a follow-up PR to do this, building on your test.

Thanks! Yeah, this was just meant to be a starting point for you all. Presumably you'd want tests for types other than tristates too. Go ahead and merge this if it makes sense.

I'm eager to switch over to tsgo in my own work. Thanks for your efforts!

EDIT: fixing tests...

srabraham:
I think the tests/formatting should work now

EDIT: looking again...

srabraham:
Thanks for your patience. I'm still cloning the submodules locally, which are apparently required in order to run the particular failing test in this PR. Sigh...

jakebailey:
FYI, no need to force push; we squash on merge so your history can be as bad as you want it. The force pushing really only makes it hard to see what's changing when reviewing.

srabraham:
> FYI, no need to force push; we squash on merge so your history can be as bad as you want it. The force pushing really only makes it hard to see what's changing when reviewing.

Ah, true. Yeah, bad habit from a past job

jakebailey:
Once you have cloned the submodule, you should be able to do `hereby test && hereby baseline-accept` and commit the changes.

srabraham:
> Once you have cloned the submodule, you should be able to do `hereby test && hereby baseline-accept` and commit the changes.

done, but wow, that just made so much diff I assumed it was wrong

srabraham:
Wunderbar! Please go ahead and merge at your leisure (looks like I can't do so myself). Thank you again

PR #844, opened 2025-05-02
Typings Intaller initial work

sheetalkamat:
This ports the typescript installer code from Strada

Instead of different process we use go routine to queue discovery of the packages install and then another go routine to do actual install. 

Here are things that are still not ported and are todo for later
- Js file names to search typings for - need to exclude source files from external module search - currently this data is not threaded so left it for later
- custom npm path, safelist path etc that earlier we use to take - review and add the options that we really need
- Strada use to use events to notify client about typings installation. we need to use notification but will do that later
- TypingsInstaller currently uses "latest" as version to install typings for, to change this to correct version when we have typings for those published

While working on this, found out current bug in strada where typings already installed are not cached correctly if lockfile version is newer. This i havent fixed but will do that separately later. This means even if  say `@types/chai` is already in our cache location installed, we will reinstall it. (Because `package-lock.json` structure change)


sheetalkamat:
looking into test failures

cmd/tsgo/lsp.go
```diff
@@ -45,10 +48,78 @@ func runLSP(args []string) int {
...
 	})
 
 	if err := s.Run(); err != nil {
 		return 1
 	}
 	return 0
 }
+
+func getGlobalTypingsCacheLocation() string {
+	switch runtime.GOOS {
```
jakebailey, line 61: Glad to see this code living out here; best to keep the platform specific code out of the real code.

internal/compiler/program.go
```diff
@@ -137,6 +140,8 @@ func NewProgram(options ProgramOptions) *Program {
 	}
 
 	p.resolver = module.NewResolver(p.host, p.compilerOptions)
+	p.resolver.TypingsLocation = p.programOptions.TypingsLocation
+	p.resolver.ProjectName = p.programOptions.ProjectName
```
jakebailey: Should these just be parameters/options on `NewResolver` rather than patched on?

internal/core/nodemodules.go
```diff
@@ -0,0 +1,94 @@
...
+			nodeCoreModules[unprefixed] = true
+			nodeCoreModules["node:"+unprefixed] = true
+		}
+		maps.Copy(nodeCoreModules, ExclusivelyPrefixedNodeCoreModules)
+	})
+}
+
+func NonRelativeModuleNameForTypingCache(moduleName string) string {
+	ensureNodeCoreModules()
+	if nodeCoreModules[moduleName] {
```
jakebailey:
```suggestion
var nodeCoreModules = sync.OnceFunc(func() map[string]bool {
	nodeCoreModules := make(map[string]bool, len(UnprefixedNodeCoreModules)*2 + len(ExclusivelyPrefixedNodeCoreModules))
	for unprefixed := range UnprefixedNodeCoreModules {
		nodeCoreModules[unprefixed] = true
		nodeCoreModules["node:"+unprefixed] = true
	}
	maps.Copy(nodeCoreModules, ExclusivelyPrefixedNodeCoreModules)
	return nodeCoreModules
})

func NonRelativeModuleNameForTypingCache(moduleName string) string {
	if nodeCoreModules()[moduleName] {
```

This is a simpler way to do this same thing.

internal/project/discovertypings.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	if hasJsxFile {
+		log("TI:: Inferred 'react' typings due to presence of '.jsx' extension")
+		addInferredTyping(inferredTypings, "react")
+	}
+}
+
+func getManifestNamesFromDependencies(manifestNames []string, dependencies map[string]string) []string {
+	if dependencies != nil {
+		for dependency := range dependencies {
```
jakebailey: There's no need to check for nil when reading from maps (including iterating); they are treated as empty.

internal/project/discovertypings.go
```diff
@@ -0,0 +1,342 @@
...
+	if ok {
+		var manifest packagejson.DependencyFields
+		filesToWatch = append(filesToWatch, manifestPath)
+		// var manifest map[string]any
+		err := json.Unmarshal([]byte(manifestContents), &manifest)
+		if err == nil {
+			manifestTypingNames = getManifestNamesFromDependencies(manifestTypingNames, manifest.Dependencies.Value)
+			manifestTypingNames = getManifestNamesFromDependencies(manifestTypingNames, manifest.DevDependencies.Value)
+			manifestTypingNames = getManifestNamesFromDependencies(manifestTypingNames, manifest.OptionalDependencies.Value)
+			manifestTypingNames = getManifestNamesFromDependencies(manifestTypingNames, manifest.PeerDependencies.Value)
```
jakebailey: FWIW I think this could be written like:

```go
manifestTypingNames = slices.AppendSeq(manifestTypingNames, maps.Keys(manifest.Dependencies.Value))
```

And avoid needing to handwrite the loops.

internal/project/discovertypings.go
```diff
@@ -0,0 +1,342 @@
...
+ * @internal
+ */
+func removeMinAndVersionNumbers(fileName string) string {
+	// We used to use the regex /[.-]((min)|(\d+(\.\d+)*))$/ and would just .replace it twice.
+	// Unfortunately, that regex has O(n^2) performance because v8 doesn't match from the end of the string.
+	// Instead, we now essentially scan the filename (backwards) ourselves.
+
+	end := len(fileName)
+	for pos := end - 1; pos > 0; pos-- {
+		ch := rune(fileName[pos])
```
jakebailey: I am suspicious of this code's unicode handling; indexing is by byte so it's possible to chop up a longer sequence accidentally.

If this doesn't have to be fast, it may be simpler to just convert the string to `[]rune` and back? Or maybe, the bad case can't happen here anyway.

internal/project/validatepackagename.go
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Validates package name using rules defined at https://docs.npmjs.com/files/package.json
+ *
+ * @internal
+ */
+func ValidatePackageName(packageName string) (result NameValidationResult, name string, isScopeName bool) {
+	return validatePackageNameWorker(packageName /*supportScopedPackage*/, true)
+}
+
+var packageNameRegexp = regexp.MustCompile("^@([^/]+)/([^/]+)$")
```
jakebailey: If possible we should avoid using the `regexp` package, which is unfortunately pretty slow.

This seems like it could just be some `strings` package `Cut` calls given this is a full string match?

internal/project/validatepackagename.go
```diff
@@ -0,0 +1,97 @@
...
+
+func validatePackageNameWorker(packageName string, supportScopedPackage bool) (result NameValidationResult, name string, isScopeName bool) {
+	packageNameLen := len(packageName)
+	if packageNameLen == 0 {
+		return EmptyName, "", false
+	}
+	if packageNameLen > maxPackageNameLength {
+		return NameTooLong, "", false
+	}
+	firstChar := rune(packageName[0])
```
jakebailey: I'm suspicious of this too, but maybe we'll never have a non-ascii package name?

internal/vfs/utilities.go
```diff
@@ -1,4 +1,4 @@
-package tsoptions
+package vfs
```
jakebailey, line 1: Should these just be in `tspath`, given they're path operations? `vfs` is more of the actual FS abstraction, not utilities.

internal/project/project.go
```diff
@@ -636,8 +930,52 @@ func (p *Project) clearSourceMapperCache() {
...
+	//             for (const f of configFile.extendedSourceFiles) {
+	//                 result.push(asNormalizedPath(f));
+	//             }
+	//         }
+	//     }
+	// }
+	return result
+}
+
+func (p *Project) print(writeFileNames bool, writeFileExplanation bool, writeFileVersionAndText bool, inputBuilder ...strings.Builder) string {
```
jakebailey: `string.Builder` should never be passed by value, so this would have to be `...*string.Builder`, or not passed at all.

Why does this need to take in more builders anyway? Is this just a hack to pass an optional parameter?

internal/project/ti.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+	return result
+}
+
+func (ti *TypingsInstaller) init(p *Project) {
+	ti.initializedMu.Lock()
+	if ti.initialized {
+		ti.initializedMu.Unlock()
+		return
+	}
```
jakebailey: This would is more canonically handled as a `sync.Once`, like:

```go
type TypingsInstaller struct {
	initOnce sync.Once
}

func (ti *TypingsInstaller) init(p *Project) {
	init.initOnce.Do(func() {
		// ...
	})
}
```

internal/project/ti.go
```diff
@@ -0,0 +1,651 @@
...
+				}
+				p.Log("TI:: Adding entry into typings cache: " + packageName + " => " + typingFile)
+				version := npmLockValue.Version
+				if version == "" {
+					continue
+				}
+
+				newTyping := &CachedTyping{
+					TypingsLocation: typingFile,
+					Version:         semver.MustParse(version),
```
jakebailey, line 548: Is this version user-controlled? IIRC the package.json can contain basically anything, even a bad version, since that's not actually used by package managers directly. It might be that this needs to not be a `MustParse` to avoid a panic.

internal/project/project.go
```diff
@@ -477,6 +545,225 @@ func (p *Project) updateProgram() bool {
...
+		}
+	}
+	if p.unresolvedImportsPerFile == nil {
+		p.unresolvedImportsPerFile = make(map[*ast.SourceFile][]string, len(p.program.GetSourceFiles()))
+	}
+	p.unresolvedImportsPerFile[file] = unresolvedImports
+	return hasChanges
+}
+
+func (p *Project) UpdateTypingFiles(typingsInfo *TypingsInfo, typingFiles []string) {
```
jakebailey, line 681: I suspect that this function is not concurrent safe when the LSP is actually running; I think that we need to be creating a channel and sending this stuff as an event to the `dispatchLoop`, allowing that operation to be serial. `dispatchLoop` would then select on that new channel and we'd be concurrency safe.

internal/project/discovertypings.go
sheetalkamat: yeah this was first `go` change so little rusty..

internal/project/discovertypings.go
sheetalkamat: I was trying to port this https://github.com/microsoft/TypeScript/blob/main/src/compiler/core.ts#L2276 .. 

This is called on js files in program to find typings for those files eg from `jquery-min.2.3.4.js` to `jquery`

jakebailey: Ah; what you would probably want to do then is to not loop over the len, but rather use `utf8.DecodeRune` and increment the size it returns. We do have some examples in the scanner and others of this.

Then again, this is a reverse iteration, so would need to be `DecodeLastRune`? Or, just make it slightly slower and do `[]rune(fileName)` and iterate those...

jakebailey: But, maybe none of the characters being searched for are actually possible inside of a longer unicode character such that this code is fine.

internal/project/validatepackagename.go
sheetalkamat: this could be from unresolved import, types array in config, include in config or type name derived from js file name

internal/vfs/utilities.go
sheetalkamat: This contains ReadDirectory etc implementation so makes sense to keep in vfs as it works with FS

jakebailey: I suppose so, but most of it seems to be globbing related?

The vfs package today only has declarations, no behavior, so if anything it feels like it should live somewhere else still, but...

internal/project/project.go
sheetalkamat: This is used in printProjects in service. I removed the group logic from logger to make it thread safe with logging (found during --race test). Changed this to always take `*strings.builder`

internal/project/ti.go
sheetalkamat: I just added this for `const result = Debug.checkDefined(tryParseComponents(major), "Invalid version");` porting this portion of the code.

sheetalkamat: This also comes from our `@types` package.json soi should be ok right ?

jakebailey: Sure, yeah, that's fine. Just verifying, since panicking in Go is more like "something extremely bad has happened and I would like to crash", versus exceptions in JS which are actually reasonable to handle.

But, if it was a debug assert before, then we don't handle that either really.

internal/project/project.go
sheetalkamat: Do we? Wont just aquiring lock on  `p.mu` be enough ?
Also this is the only place that sets this and and is read when updating program so i thought marking dirty lock should be ok and hence had it without p.mu lock

jakebailey: Hm, maybe; @andrewbranch would probably be able to speak better to it. I guess maybe what I suggested isn't so good when all of the functionality is within the project package?

sheetalkamat: Yeah and its all i believe is thread safe with "dirty" mutex that project maintains. i think its just like any watch event - so i just need to make sure its marked with lock aquired i believe but i will chat with @andrewbranch to make sure we are on same page

internal/project/discovertypings.go
```diff
@@ -287,34 +288,33 @@ func removeMinAndVersionNumbers(fileName string) string {
 	// We used to use the regex /[.-]((min)|(\d+(\.\d+)*))$/ and would just .replace it twice.
 	// Unfortunately, that regex has O(n^2) performance because v8 doesn't match from the end of the string.
 	// Instead, we now essentially scan the filename (backwards) ourselves.
-
 	end := len(fileName)
-	for pos := end - 1; pos > 0; pos-- {
-		ch := rune(fileName[pos])
+	for pos := end; pos > 0; {
+		ch, size := utf8.DecodeLastRuneInString(fileName[0:pos])
```
jakebailey: No need for `0`; you can omit it and write `:pos`, as the zero is implicit.

internal/project/project.go
sheetalkamat: I think this is right approach as this is similar to adding /removing root files.

jakebailey:
Works on my machine 😄

PR #847, opened 2025-05-06
[LSP] fix crash on inferred project JS files

andrewbranch:


internal/project/service.go
```diff
@@ -619,7 +619,9 @@ func (s *Service) getDefaultProjectForScript(scriptInfo *ScriptInfo) *Project {
 
 func (s *Service) createInferredProject(currentDirectory string, projectRootPath tspath.Path) *Project {
 	// !!!
-	compilerOptions := core.CompilerOptions{}
+	compilerOptions := core.CompilerOptions{
+		AllowJs: core.TSTrue,
```
sheetalkamat, line 623: We normally set this to true when there are js root files. i got this working in my branch as its kind of what i do for type Acq defaults too.

andrewbranch: It appears to be unconditional https://github.com/microsoft/TypeScript/blob/d88d3a46810bfae0b072beb023a2b09b8026b82c/src/server/project.ts#L2422

sheetalkamat: You are right. thats the maxNodeModule depth part thats conditional

sheetalkamat (approved):

PR #851, opened 2025-05-07
Add support for resolution mode

EricCornelson:


andrewbranch:
Looking at the error diffs that got modified but not deleted, a few big categories stand out:

- Missing program-level diagnostics (`verifyCompilerOptions` not ported)
- Module specifiers in errors mismatch in formatting, which will be fixed by #791
- Some tests accidentally include trailing commas in package.json, which we successfully parse in Strada by using our own JSON parser and evaluator, but fail to parse in Corsa. npm itself crashes on invalid JSON in package.json files and doesn’t allow it to be pushed to the registry, so I decided to use Go’s built-in JSON parsing for better performance.

testdata/baselines/reference/submoduleAccepted/conformance/importTag15(module=esnext).errors.txt.diff
```diff
@@ -0,0 +1,22 @@
...
+     export interface I { }
+     
+-==== 1.js (2 errors) ====
++==== 1.js (3 errors) ====
+     /** @import { I } from './0' with { type: "json" } */
+                                  ~~~~~~~~~~~~~~~~~~~~~
+ !!! error TS2857: Import attributes cannot be used with type-only imports or exports.
+     /** @import * as foo from './0' with { type: "json" } */
++                              ~~~~~
++!!! error TS2307: Cannot find module './0' or its corresponding type declarations.
```
jakebailey: This is odd; wonder why this appears.

andrewbranch: There are a number of bad effects from #791 and #970 that I’m working through.

PR #853, opened 2025-05-08
Improve the missing CompilerOptions check

srabraham:
This mucks with the real code a bit, but it works better and is simpler.

@jakebailey FYI, and I swear I'm done bugging you 😝 

See https://github.com/microsoft/typescript-go/pull/839 https://github.com/microsoft/typescript-go/issues/842

### DIFF
--- a/internal/tsoptions/parsinghelpers_test.go
+++ b/internal/tsoptions/parsinghelpers_test.go
@@ -1,7 +1,6 @@
 package tsoptions
 
 import (
-	"fmt"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/tsoptions/parsinghelpers_test.go
```diff
@@ -1,41 +1,44 @@
...
-		if isTristate {
-			// Set the field on a CompilerOptions to something other than the
-			// default (i.e. not TSUnknown), then check whether
-			// ParseCompilerOptions does actually update the value for that key.
-			testValue := core.TSTrue
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					if strings.Contains(fmt.Sprint(r), "interface conversion") {
+						// this is a success case, amazingly
```
jakebailey: I can't say I like this testing method. I would much rather go and do the work to make these functions generated at this point.

srabraham: Ya, go for it! It just wasn't sitting well with me that I'd only done Tristates. Do note though that there's another field not being converted currently, as this PR shows.

srabraham: And the panicking thing is kind of awful...that could be averted with more reflection, since one could conceivably create a zero value of the required field type, rather than this generic "something" empty interface{}

jakebailey: Creating a zero value seems like a better idea; that might be fine to accept, though no compiler tests were affected by this PR.

srabraham: done...that was easier than I thought. I mean search this PR for "moduleDetectionKind", because I had to add that to the switch statement to make this test pass. But you're right, I guess nothing important uses that field downstream of parsinghelpers

gabrielluizsf (approved):

jakebailey (approved):

jakebailey:
@gabrielluizsf FYI your reviews don't actually have any effect since you don't have merge perms on this repo; if anything they just spam us with notifications we have to ignore. I'd prefer if you didn't just approve every PR on this repo.

gabrielluizsf:
> @gabrielluizsf FYI your reviews don't actually have any effect since you don't have merge perms on this repo; if anything they just spam us with notifications we have to ignore. I'd prefer if you didn't just approve every PR on this repo.


I apologize for any inconvenience caused. My intention was simply to contribute constructively, but I understand that it may have generated unnecessary notifications. I appreciate the feedback and will be more mindful going forward.

PR #856, opened 2025-05-09
Support cancellation in checker

ahejlsberg:
With this PR the `CheckSourceFile` and `GetDiagnostics` methods on `Checker` and `GetSyntacticDiagnostics`, `GetBindDiagnostics`, and `GetSemanticDiagnostics` on `Program` support cancellation through a `context.Context` which is passed as the first parameter to those methods.

When an operation is cancelled, any subsequent call to `CheckSourceFile`, `GetDiagnostics`, `GetDiagnosticsWithoutCheck`, or `GetGlobalDiagnostics` on that checker will panic regardless of which context the checker is invoked on. (Following cancellation, a checker contains only a partial list of diagnostics, and without more extensive work to track which source elements were checked and which weren't, that checker can no longer deliver a consistent set of diagnostics.)

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -1,6 +1,7 @@
 package compiler
 
 import (
+	"context"
 	"fmt"
 	"slices"
 	"sync"
@@ -206,13 +207,13 @@ func (p *Program) BindSourceFiles() {
 	wg.RunAndWait()
 }
 
-func (p *Program) CheckSourceFiles() {
+func (p *Program) CheckSourceFiles(ctx context.Context) {
 	p.createCheckers()
 	wg := core.NewWorkGroup(p.programOptions.SingleThreaded)
 	for index, checker := range p.checkers {
 		wg.Queue(func() {
 			for i := index; i < len(p.files); i += len(p.checkers) {
-				checker.CheckSourceFile(p.files[i])
+				checker.CheckSourceFile(ctx, p.files[i])
 			}
 		})
 	}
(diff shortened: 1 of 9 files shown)


internal/compiler/program.go
```diff
@@ -323,26 +324,27 @@ func (p *Program) getBindDiagnosticsForFile(sourceFile *ast.SourceFile) []*ast.D
...
 	// These are deduplicated later.
 	for _, checker := range p.checkers {
 		if sourceFile == nil || checker == fileChecker {
-			diags = append(diags, checker.GetDiagnostics(sourceFile)...)
+			diags = append(diags, checker.GetDiagnostics(ctx, sourceFile)...)
 		} else {
 			diags = append(diags, checker.GetDiagnosticsWithoutCheck(sourceFile)...)
 		}
 	}
+	if ctx != nil && ctx.Err() != nil {
```
jakebailey: No need to check `nil`; besides the struct field in the checker we are allowing to be nil, all other uses (as parameters) should always be non-nil.

internal/ls/diagnostics.go
```diff
@@ -1,14 +1,15 @@
...
 
 	"github.com/microsoft/typescript-go/internal/ast"
 )
 
 func (l *LanguageService) GetDocumentDiagnostics(fileName string) []*ast.Diagnostic {
 	program, file := l.getProgramAndFile(fileName)
-	syntaxDiagnostics := program.GetSyntacticDiagnostics(file)
-	semanticDiagnostics := program.GetSemanticDiagnostics(file)
+	syntaxDiagnostics := program.GetSyntacticDiagnostics(context.Background(), file)
+	semanticDiagnostics := program.GetSemanticDiagnostics(context.Background(), file)
```
jakebailey, line 13:
```suggestion
	syntaxDiagnostics := program.GetSyntacticDiagnostics(context.TODO(), file)
	semanticDiagnostics := program.GetSemanticDiagnostics(context.TODO(), file)
```

I think this is one place we're going to want to plumb it.

internal/compiler/program.go
ahejlsberg: Right, but there are other `Checker` methods that don't take a context so there's no assurance that a context has been provided. We could default to `context.Background()` in initialization (and restore `context.Background()` in `CheckSourceFile`) but I sort of prefer the way it is done now since it makes it clear that the checker was invoked through a non-cancellable method.

internal/ls/diagnostics.go
ahejlsberg: Indeed.

internal/compiler/program.go
ahejlsberg: An alternative here would be to associate a context with a `Checker` upon construction and then keep that association for the lifetime of the checker. But that seems less idiomatic.

jakebailey: We definitely do not want to associate a checker with a context like that, no, for sure.

What I mean above is that `getSemanticDiagnosticsForFile`'s context is a parameter, and no caller should ever be passing in a nil context. The only nil context is the one internal to the checker (in another package).

ahejlsberg: Ah, sorry, I missed that this wasn't a `c.ctx` check. Will fix.

ahejlsberg:
I should mention that I verified cancellation works by compiling the VS Code project with a context that times out after two seconds. It consistently bailed out within a few milliseconds.

internal/ls/diagnostics.go
ahejlsberg: I'm going to leave that to a separate PR though.

jakebailey: To be clear my suggestion wasn't to plumb it in this PR, but to just leave these as TODO calls to make it clear that we're intending on changing it later.

jakebailey: I'll approve it in any case, this is just a nit and I'm sure Andrew will figure this out when he plugs it in

jakebailey (approved):

andrewbranch (approved):
We're going to have a lot of merge conflicts, but I can't think of a better way 🙃

PR #858, opened 2025-05-12
Port global completions

gabritto:
This PR ports global completions, a.k.a. completions you get from symbols collected by `tryGetGlobalSymbols`.


andrewbranch:
Are there any notable deviations from Strada that should get special attention in a code review?

gabritto:
> Are there any notable deviations from Strada that should get special attention in a code review?

Good question. I don't think there are any for this PR, at least not that I can think of. I'll try to keep that in mind and call it out on future PRs.

internal/ls/utilities.go
```diff
@@ -425,3 +530,200 @@ func isArgumentOfElementAccessExpression(node *ast.Node) bool {
...
+			return isCompletedNode(n.AsIndexSignatureDeclaration().Type, sourceFile)
+		}
+		return hasChildOfKind(n, ast.KindCloseBracketToken, sourceFile)
+
+	case ast.KindCaseClause,
+		ast.KindDefaultClause:
+		// there is no such thing as terminator token for CaseClause/DefaultClause so for simplicity always consider them non-completed
+		return false
+
+	case ast.KindForStatement,
```
RyanCavanaugh, line 630:
```suggestion
    // Add a corresponding case to ast.go if modifying this
	case ast.KindForStatement,
```

gabritto: ??

internal/checker/flow.go
```diff
@@ -333,7 +333,7 @@ func (c *Checker) narrowTypeByAssertion(f *FlowState, t *Type, expr *ast.Node) *
 			return c.narrowTypeByAssertion(f, c.narrowTypeByAssertion(f, t, node.AsBinaryExpression().Left), node.AsBinaryExpression().Right)
 		}
 		if node.AsBinaryExpression().OperatorToken.Kind == ast.KindBarBarToken {
-			return c.getUnionType([]*Type{c.narrowTypeByAssertion(f, t, node.AsBinaryExpression().Left), c.narrowTypeByAssertion(f, t, node.AsBinaryExpression().Right)})
+			return c.GetUnionType([]*Type{c.narrowTypeByAssertion(f, t, node.AsBinaryExpression().Left), c.narrowTypeByAssertion(f, t, node.AsBinaryExpression().Right)})
```
jakebailey: Nit, but maybe we should just declare a new one and keep the existing calls internal to reduce the diff? Maybe it's pointless.

jakebailey: Though it is a lot of lines in this PR...

internal/checker/services.go
```diff
@@ -18,7 +18,7 @@ func (c *Checker) getSymbolsInScope(location *ast.Node, meaning ast.SymbolFlags)
 		return nil
 	}
 
-	symbols := createSymbolTable(nil)
+	symbols := make(ast.SymbolTable)
```
jakebailey, line 21: Is this intentional? I assume so to get non-nil?

internal/ls/utilities.go
jakebailey: Yeah I'm not sure what this comment is

internal/scanner/scanner.go
```diff
@@ -2020,6 +2021,7 @@ func init() {
 	for text, key := range textToToken {
 		tokenToText[key] = text
 	}
+	maps.Copy(textToToken, textToKeyword)
```
jakebailey: I'd consider putting this up by `textToToken`; packages can have more than one init function so you can just write it there. Or honestly, in these cases if I had written them originally I would have used an IIFE.

jakebailey (commented):
I think that something must be missing from the response; I'm doing global completions and VS Code is saying that there aren't any matches:

image

internal/checker/services.go
gabritto: Yes, the later code may crash with a nil symbol table.

gabritto:
> I think that something must be missing from the response; I'm doing global completions and VS Code is saying that there aren't any matches:
> 
> image

Yeah, for some reason the insert/replace positions are messed up I think, I don't know yet where the positions go wrong. I'll debug and update with a fix tomorrow.

Update: fixed by #866.

andrewbranch (approved):

jakebailey (changes requested):
Before we merge this, I believe that @ahejlsberg wanted to see about not renaming the methods but instead declaring exported wrappers in a new file in the checker package. Would probably be good to do that now instead of churning back and forth.

jakebailey (approved):

jakebailey (approved):

PR #852, opened 2025-05-08
Port type acquisition parsing from `tsconfig.json`

iisaduan:
This uses the beginning of the type acquisition definitions from #844

PR #859, opened 2025-05-12
add unknown option error for TypeAcquisition

iisaduan:
This PR adds some of the error handling that was missed in #852 and wasn't ported over in the initial tsconfig parsing port


jakebailey (approved):

internal/tsoptions/tsconfigparsing.go
```diff
@@ -520,7 +528,7 @@ func convertOptionsFromJson[O optionParser](optionsNameMap map[string]*CommandLi
 		opt, ok := optionsNameMap[key]
 		if !ok {
 			// !!! TODO?: support suggestion
-			errors = append(errors, ast.NewCompilerDiagnostic(diagnostics.Unknown_compiler_option_0, key))
+			errors = append(errors, createUnknownOptionError(key, result.UnknownOptionDiagnostic(), "", nil, nil, nil))
```
sheetalkamat, line 567: isnt this just `result.UnknownOptionDiagnostic()` after adding that method to `optionParser`

sheetalkamat (changes requested):

internal/tsoptions/tsconfigparsing.go
iisaduan: `result.UnknownOptionDiagnostic()` only returns the diagnostic message. We need the call to create the full error (and format if needed)

internal/tsoptions/errors.go
```diff
@@ -67,3 +84,16 @@ func createDiagnosticForNodeInSourceFileOrCompilerDiagnostic(sourceFile *ast.Sou
 	}
 	return ast.NewCompilerDiagnostic(message, args...)
 }
+
+func extraKeyDiagnostics(s string) *diagnostics.Message {
```
sheetalkamat, line 88: sorry i meant here shouldnt this be from parser.unknownDiagnostics ? i dont understand why we have two ways of doing same thing

iisaduan: They are implemented in two different ways because of how it was implemented in strada, and then the two different code paths that use the messages.
When we are parsing from a `map` to a `struct`, we only have an `optionParser`, we don't know what type of `optionParser` it is, so we can't use `extraKeyDiagnostics` directly. 
When we are parsing the tsconfig, we don't use `optionParser`, so we are not able call `optionsParser.UnknownOptionDiagnostic()`.

I could refactor each of the `optionsParser.UnknownOptionDiagnostic()` to be like 
```
func (o *typeAcquisitionParser) UnknownOptionDiagnostic() *diagnostics.Message {
	return extraKeyDiagnostics("typeAcquisition")
}
```
but I'm not sure if it's worth the indirection

iisaduan: I thought about it some more, and ended up doing that refactor

PR #861, opened 2025-05-12
Signature help implementation

navya9singh:
This pr covers mostly everything that is needed for signature help except for signature help in .js files. Some tests that I haven't yet ported from Strada are tests for:
1. Tagged templates
2. Verify no signature help
3. Filtered triggers
4. Tests with JSDoc comments and other comments. It needs some changes in the test setup


_submodules/TypeScript
jakebailey, line 1: Something went wrong in merging main somewhere; I'd merge main and then do `git checkout main -- _submodules/TypeScript` then `git submodule update` to fix this.

internal/checker/utilities.go
```diff
@@ -2073,3 +2084,211 @@ func (c *Checker) checkNotCanceled() {
 		panic("Checker was previously cancelled")
 	}
 }
+
```
navya9singh: These are some functions I added temporarily to make Signature help work. They will be removed once the node builder pr is merged. Also, these do not give the best result for generics, and returns `object` instead. This is why some of the tests have a comments `//returns object` because this part returns `object`.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1120 @@
...
+
+func createTypeHelpItems(symbol *ast.Symbol, argumentInfo *argumentListInfo, sourceFile *ast.SourceFile, clientOptions *lsproto.SignatureHelpClientCapabilities, c *checker.Checker) *lsproto.SignatureHelp {
+	typeParameters := c.GetLocalTypeParametersOfClassOrInterfaceOrTypeAlias(symbol)
+	if typeParameters == nil {
+		return nil
+	}
+	item := getTypeHelpItem(symbol, typeParameters, getEnclosingDeclarationFromInvocation(argumentInfo.invocation), sourceFile, c)
+
+	// Converting signatureHelpParameter to *lsproto.ParameterInformation
+	var parameters []*lsproto.ParameterInformation
```
navya9singh: Some of these functions are converting from local types to lsproto type because they needed some extra parameters like `isRest`, `isOptional` or `isVariadic` to be able to correctly calculate the parameter count and the parameter index that would be highlighted.

internal/ls/utilities.go
```diff
@@ -425,3 +511,116 @@ func isArgumentOfElementAccessExpression(node *ast.Node) bool {
 func isTagName(node *ast.Node) bool {
 	return node.Parent != nil && ast.IsJSDocTag(node.Parent) && node.Parent.TagName() == node
 }
+
+func IsInString(sourceFile *ast.SourceFile, position int, previousToken *ast.Node) bool {
+	if previousToken != nil && ast.IsStringTextContainingNode(previousToken) {
+		start := previousToken.Pos()
```
gabritto:
```suggestion
		start := astnav.GetStartOfNode(previousToken, sourceFile, false /*includeJSDoc*/)
```

internal/ls/utilities.go
```diff
@@ -425,3 +511,116 @@ func isArgumentOfElementAccessExpression(node *ast.Node) bool {
...
+		//   1. entirely within the token text.
+		//   2. at the end position of an unterminated token.
+		//   3. at the end of a regular expression (due to trailing flags like '/foo/g').
+		if start < position && position < end {
+			return true
+		}
+
+		if position == end {
+			return ast.IsUnterminatedLiteral(previousToken)
+			//return !!(previousToken as LiteralExpression).isUnterminated;
```
gabritto:
```suggestion
```

internal/astnav/tokens.go
```diff
@@ -602,3 +602,13 @@ func getNodeVisitor(
 		},
 	})
 }
+
+// !!!
+func HasQuestionToken(node *ast.Node) bool {
```
jakebailey: Should this go into the ast package?

internal/checker/checker.go
```diff
@@ -17392,7 +17392,7 @@ func (c *Checker) getTypeOfPropertyOfType(t *Type, name string) *Type {
 	return nil
 }
 
-func (c *Checker) getSignaturesOfType(t *Type, kind SignatureKind) []*Signature {
+func (c *Checker) GetSignaturesOfType(t *Type, kind SignatureKind) []*Signature {
```
jakebailey: Don't rename this method; just declare `GetSignaturesOfType` in `exports.go` and then forward to this unexported method.

I think the same should go for all of the other unexported functions; we're trying to consolidate these into a clearer API and also avoid huge diffs due to renames (which this PR does at the moment).

internal/ast/utilities.go
```diff
@@ -2964,3 +2964,13 @@ func GetPropertyNameForPropertyNameNode(name *Node) string {
 	}
 	panic("Unhandled case in getPropertyNameForPropertyNameNode")
 }
+
+// !!!
+func HasQuestionToken(node *Node) bool {
+	switch node.Kind {
+	case KindParameter, KindMethodDeclaration, KindMethodSignature, KindShorthandPropertyAssignment,
+		KindPropertyAssignment, KindPropertyDeclaration, KindPropertySignature:
+		return node.AsParameterDeclaration().QuestionToken != nil
```
gabritto, line 3395: That doesn't look right: if the node kind is, say, method declaration, `node.AsParameterDeclaration()` will fail. So I'd suggest either adding a separate cast for each node kind here, or maybe creating a `QuestionToken()` method in `ast.go`, in the style of the other getters we have there already.

internal/checker/checker.go
```diff
@@ -29422,7 +29429,7 @@ func (c *Checker) GetSymbolAtLocation(node *ast.Node) *ast.Symbol {
 // be used only by the language service and external tools. The semantics of the function are deliberately "fuzzy"
 // and aim to just return *some* symbol for the node. To obtain the symbol associated with a node for type checking
 // purposes, use appropriate function for the context, e.g. `getResolvedSymbol` for an expression identifier,
-// `getSymbolOfDeclaration` for a declaration, etc.
```
gabritto: These are probably leftover from the old rename.

internal/checker/exports.go
```diff
@@ -52,3 +52,59 @@ func GetDeclarationModifierFlagsFromSymbol(s *ast.Symbol) ast.ModifierFlags {
...
+
+func IsTupleType(t *Type) bool {
+	return isTupleType(t)
+}
+
+func (c *Checker) GetReturnTypeOfSignature(sig *Signature) *Type {
+	return c.getReturnTypeOfSignature(sig)
+}
+
+func IsCallOrNewExpression(node *ast.Node) bool {
```
gabritto: I would just move those node predicates to `ast/utilities.go`, like this one, `GetInvokedExpression`, `IsJsxOpeningLikeElement`, etc.

internal/checker/inference.go
```diff
@@ -1235,7 +1235,7 @@ func (c *Checker) inferFromIntraExpressionSites(n *InferenceContext) {
 		if ast.IsMethodDeclaration(site.node) {
 			contextualType = c.getContextualTypeForObjectLiteralMethod(site.node, ContextFlagsNoConstraints)
 		} else {
-			contextualType = c.getContextualType(site.node, ContextFlagsNoConstraints)
```
gabritto: Also looks like leftover from a rename?

internal/checker/checker.go
```diff
@@ -25962,16 +25962,23 @@ func (c *Checker) markPropertyAsReferenced(prop *ast.Symbol, nodeForCheckWriteOn
 	c.symbolReferenceLinks.Get(target).referenceKinds |= ast.SymbolFlagsAll
 }
 
-func (c *Checker) GetExpandedParameters(signature *Signature /* !!! skipUnionExpanding */) []*ast.Symbol {
+func (c *Checker) getExpandedParameters(signature *Signature, skipUnionExpanding *bool) [][]*ast.Symbol {
```
gabritto: `skipUnionExpanding` should just be `bool`, since we can't avoid passing arguments in Go like we can in TS with optional parameters.

internal/checker/printer.go
```diff
@@ -585,7 +585,8 @@ func (p *Printer) printSignature(sig *Signature, returnSeparator string) {
 		p.printType(p.c.getTypeOfSymbol(sig.thisParameter))
 		tail = true
 	}
-	expandedParameters := p.c.GetExpandedParameters(sig)
+	skipUnionExpanding := true
+	expandedParameters := p.c.getExpandedParameters(sig, &skipUnionExpanding)[0]
```
gabritto: Once the parameter is a regular `bool`, I think this can be inlined.

internal/checker/services.go
```diff
@@ -481,3 +482,50 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
 func (c *Checker) GetContextualTypeForJsxAttribute(attribute *ast.JsxAttributeLike) *Type {
 	return c.getContextualTypeForJsxAttribute(attribute, ContextFlagsNone)
 }
+
+func (c *Checker) runWithoutResolvedSignatureCaching(node *ast.Node, fn func() *Signature) *Signature {
```
gabritto: I think I added this function already (the version you shared with me earlier, but made generic) in one of the completion PRs.

internal/checker/checker.go
DanielRosenwasser: I would avoid making this a `*bool`. It implies one of two things to  me:

1. You intend to mutate the state after the fact (which is why it's a pointer)
2. You need `nil` to be treated distinctly from `false`.

The only reason the old codebase had to account for `undefined` was so it could support an optional parameter which Go doesn't have.

internal/checker/services.go
```diff
@@ -481,3 +482,50 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
...
+	ancestorNode := ast.FindAncestor(node, func(n *ast.Node) bool {
+		return ast.IsCallLikeOrFunctionLikeExpression(n)
+	})
```
DanielRosenwasser:
```suggestion
	ancestorNode := ast.FindAncestor(node, ast.IsCallLikeOrFunctionLikeExpression)
```

internal/checker/services.go
```diff
@@ -481,3 +482,50 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
...
+	if ancestorNode != nil {
+		cachedResolvedSignatures := make(map[*SignatureLinks]*Signature)
```
DanielRosenwasser: I might just do two slices here instead of a map, but this might be cleaner.

Either way, I would just add a comment like

```suggestion
		// We're going to temporarily reset resolved signatures for all surrounding calls.
		cachedResolvedSignatures := make(map[*SignatureLinks]*Signature)
```

internal/checker/services.go
```diff
@@ -481,3 +482,50 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
...
+			if ast.IsFunctionExpressionOrArrowFunction(ancestorNode) {
+				symbolLinks := c.valueSymbolLinks.Get(c.getSymbolOfDeclaration(ancestorNode))
+				resolvedType := symbolLinks.resolvedType
+				cachedTypes[symbolLinks] = resolvedType
+				symbolLinks.resolvedType = nil
+			}
+			ancestorNode = ast.FindAncestor(ancestorNode.Parent, ast.IsCallLikeOrFunctionLikeExpression)
+		}
+		result := fn()
+		for signatureLinks, resolvedSignature := range cachedResolvedSignatures {
```
DanielRosenwasser:
```suggestion

		result := fn()

		// Once our work is over, we'll restore any signatures we previously resolved.
		for signatureLinks, resolvedSignature := range cachedResolvedSignatures {
```

internal/ls/utilities.go
```diff
@@ -245,13 +245,99 @@ type PossibleTypeArgumentInfo struct {
...
 func getPossibleTypeArgumentsInfo(tokenIn *ast.Node, sourceFile *ast.SourceFile) *PossibleTypeArgumentInfo {
-	return nil
-}
+	// This is a rare case, but one that saves on a _lot_ of work if true - if the source file has _no_ `<` character,
+	// then there obviously can't be any type arguments - no expensive brace-matching backwards scanning required
+	// searchPosition := len(sourceFile.Text())
+	// if tokenIn != nil {
+	// 	searchPosition = scanner.GetTokenPosOfNode(tokenIn, sourceFile, false /*includeJSDoc*/)
+	// }
+	if strings.LastIndex(sourceFile.Text(), "<") == -1 {
```
DanielRosenwasser:
```suggestion
	if strings.LastIndexByte(sourceFile.Text(), '<') == -1 {
```

internal/ls/utilities.go
```diff
@@ -245,13 +245,99 @@ type PossibleTypeArgumentInfo struct {
...
+	}
 
-// !!! signature help
-func getPossibleGenericSignatures(called *ast.Expression, typeArgumentCount int, checker *checker.Checker) []*checker.Signature {
+	token := tokenIn
+	// This function determines if the node could be type argument position
+	// Since during editing, when type argument list is not complete,
+	// the tree could be of any shape depending on the tokens parsed before current node,
+	// scanning of the previous identifier followed by "<" before current node would give us better result
+	// Note that we also balance out the already provided type arguments, arrays, object literals while doing so
```
DanielRosenwasser:
```suggestion
	// This function determines if the node could be a type argument position
	// When editing, it is common to have an incomplete type argument list (e.g. missing ">"),
	// so the tree can have any shape depending on the tokens before the current node.
	// Instead, scanning for an identifier followed by a "<" before current node
	// will typically give us better results than inspecting the tree.
	// Note that we also balance out the already provided type arguments, arrays, object literals while doing so.
```

internal/ls/utilities.go
```diff
@@ -245,13 +245,99 @@ type PossibleTypeArgumentInfo struct {
...
+		case ast.KindLessThanToken:
+			// Found the beginning of the generic argument expression
+			token = astnav.FindPrecedingToken(sourceFile, token.Pos())
+			if token != nil && token.Kind == ast.KindQuestionDotToken {
+				token = astnav.FindPrecedingToken(sourceFile, token.Pos())
+			}
+			if token == nil || !ast.IsIdentifier(token) {
+				return nil
+			}
+			if remainingLessThanTokens <= 0 {
```
DanielRosenwasser: No need to check less-than, right?

internal/ls/utilities.go
```diff
@@ -245,13 +245,99 @@ type PossibleTypeArgumentInfo struct {
 	nTypeArguments int
 }
 
-// !!! signature help
+// Get info for an expression like `f <` that may be the start of type arguments.
 func getPossibleTypeArgumentsInfo(tokenIn *ast.Node, sourceFile *ast.SourceFile) *PossibleTypeArgumentInfo {
```
DanielRosenwasser, line 267: Minor preference for this to just return an `bool, PossibleTypeArgumentInfo` - it doesn't need to allocate. Not sure what others think.

internal/checker/types.go
```diff
@@ -1094,6 +1109,38 @@ type Signature struct {
...
+
+func (s *Signature) Parameters() []*ast.Symbol {
+	if len(s.parameters) == 0 {
+		return nil
+	}
+	return s.parameters
+}
+
+func (s *Signature) Target() *Signature {
+	if s.target == nil {
```
gabritto: This and the method below could just be `return s.target`.

internal/ls/utilities.go
```diff
@@ -724,3 +810,114 @@ func getAllSuperTypeNodes(node *ast.Node) []*ast.TypeNode {
...
+	// we can only use the textual result directly if we didn't have to count any close tokens within the range
+	if strings.LastIndex(sourceFile.Text(), closeTokenText) < bestGuessIndex {
+		nodeAtGuess := astnav.FindPrecedingToken(sourceFile, bestGuessIndex+1)
+		if nodeAtGuess != nil && nodeAtGuess.Kind == matchingTokenKind {
+			return nodeAtGuess
+		}
+	}
+	tokenKind := token.Kind
+	remainingMatchingTokens := 0
+	for true {
```
DanielRosenwasser: I think the test of the codebases uses 

```suggestion
	for {
```

DanielRosenwasser (changes requested):
Mostly stylistic feedback. I haven't fully reviewed but I haven't seen anything glaringly different from the original so far!

internal/checker/types.go
```diff
@@ -1094,6 +1109,38 @@ type Signature struct {
 	composite                *CompositeSignature
 }
 
+func (s *Signature) TypeParameters() []*Type {
+	if len(s.typeParameters) == 0 {
```
gabritto: I think here and below, this can also just be `return s.typeParameters`. `len(nil)` just returns 0, so typically there's no need to distinguish between a `nil` and an empty slice.
Then the code in signature help can just do `if len(signature.TypeParameters() != 0 ...`.

internal/checker/utilities.go
```diff
@@ -1130,6 +1131,15 @@ func isThisTypeParameter(t *Type) bool {
 	return t.flags&TypeFlagsTypeParameter != 0 && t.AsTypeParameter().isThisType
 }
 
+func isCallLikeExpression(node *ast.Node) bool {
```
gabritto: There is already a similar function in `ast/utilities.go`.

internal/ls/utilities.go
```diff
@@ -245,13 +245,99 @@ type PossibleTypeArgumentInfo struct {
...
 }
 
-// !!! signature help
+// Get info for an expression like `f <` that may be the start of type arguments.
 func getPossibleTypeArgumentsInfo(tokenIn *ast.Node, sourceFile *ast.SourceFile) *PossibleTypeArgumentInfo {
-	return nil
-}
+	// This is a rare case, but one that saves on a _lot_ of work if true - if the source file has _no_ `<` character,
+	// then there obviously can't be any type arguments - no expensive brace-matching backwards scanning required
+	// searchPosition := len(sourceFile.Text())
```
gabritto: Is this missing a `// !!!`, or are we getting rid of this code?

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1122 @@
...
+
+	// Decide whether to show signature help
+	startingToken := astnav.FindPrecedingToken(sourceFile, position)
+	if startingToken == nil {
+		// We are at the beginning of the file
+		return nil
+	}
+
+	// Only need to be careful if the user typed a character and signature help wasn't showing.
+	onlyUseSyntacticOwners := context.TriggerKind == 2
```
gabritto:
```suggestion
	onlyUseSyntacticOwners := context.TriggerKind == lsproto.SignatureHelpTriggerKindTriggerCharacter
```

internal/ls/utilities.go
gabritto: I think completions uses the detailed return info, so we'll probably need to return a struct here.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1122 @@
...
+		})
+	}
+	return result
+}
+
+func itemInfoForParameters(candidateSignature *checker.Signature, c *checker.Checker, enclosingDeclaratipn *ast.Node, sourceFile *ast.SourceFile) []*signatureHelpItemInfo {
+	printer := printer.NewPrinter(printer.PrinterOptions{NewLine: core.NewLineKindLF}, printer.PrintHandlers{}, nil)
+
+	var getTypeParameters []signatureHelpParameter = []signatureHelpParameter{}
+	if candidateSignature.TypeParameters() != nil && len(candidateSignature.TypeParameters()) != 0 {
```
gabritto: This here is what I mentioned elsewhere that could become just `len(candidateSignature.TypeParameters()) != 0`.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1122 @@
...
+	if !checker.IsCallOrNewExpression(node) {
+		return false
+	}
+	invocationChildren := getTokensFromNode(node, sourceFile)
+	switch startingToken.Kind {
+	case ast.KindOpenParenToken:
+		return containsNode(invocationChildren, startingToken)
+	case ast.KindCommaToken:
+		return containsNode(invocationChildren, startingToken)
+		// const containingList = findContainingList(startingToken);
```
gabritto, line 572: Should this have a `// !!!`?

gabritto (commented):
Some minor comments, but in general I think this is looking good.

internal/ls/utilities.go
navya9singh: Yes, getting rid of it.

internal/ls/utilities.go
navya9singh: Yeah, it can instead be changed to check if `remainingLessThanTokens == 0`  (in Strada it had `(!remainingLessThanTokens)`) because this part is trying to check if we have found the start of the type argument.

internal/ast/utilities.go
navya9singh: Right, I wasn't quite sure how to add this getter, so I added a `// !!!`. Most of these kinds, after a case, do not have a token property. They do have a `PostfixToken ` which has a token node. Is that what should be checked here if it's a questionToken?

internal/checker/checker.go
navya9singh: You're right. I was testing this out by passing in `nil` at some point, but just a `bool` works now.

internal/ast/utilities.go
gabritto: Yes, I think checking `PostfixToken` should be the new way to do this.

internal/ast/utilities.go
```diff
@@ -2965,12 +2965,22 @@ func GetPropertyNameForPropertyNameNode(name *Node) string {
...
 
-// !!!
 func HasQuestionToken(node *Node) bool {
 	switch node.Kind {
-	case KindParameter, KindMethodDeclaration, KindMethodSignature, KindShorthandPropertyAssignment,
-		KindPropertyAssignment, KindPropertyDeclaration, KindPropertySignature:
+	case KindParameter:
 		return node.AsParameterDeclaration().QuestionToken != nil
+	case KindMethodDeclaration:
+		return IsQuestionToken(node.AsMethodDeclaration().PostfixToken)
```
gabritto, line 3397: I think `IsQuestionToken` needs to be modified to not crash when given a `nil` node.

internal/checker/checker.go
```diff
@@ -25962,16 +25962,23 @@ func (c *Checker) markPropertyAsReferenced(prop *ast.Symbol, nodeForCheckWriteOn
...
 		restIndex := len(signature.parameters) - 1
 		restSymbol := signature.parameters[restIndex]
 		restType := c.getTypeOfSymbol(restSymbol)
 		if isTupleType(restType) {
-			return c.expandSignatureParametersWithTupleMembers(signature, restType.AsTypeReference(), restIndex, restSymbol)
+			return [][]*ast.Symbol{c.expandSignatureParametersWithTupleMembers(signature, restType.AsTypeReference(), restIndex, restSymbol)}
+		} else if !skipUnionExpanding && restType.flags&TypeFlagsUnion != 0 && core.Every(restType.AsUnionType().Types(), isTupleType) {
+			result := [][]*ast.Symbol{}
+			for _, t := range restType.AsUnionType().Types() {
+				result = append(result, c.expandSignatureParametersWithTupleMembers(signature, t.AsTypeReference(), restIndex, restSymbol))
```
andrewbranch: You can allocate this array with the right size:

```suggestion
			types := restType.AsUnionType().Types()
			result := make([][]*ast.Symbol, len(types))
			for i, t := range types {
				result[i] = c.expandSignatureParametersWithTupleMembers(signature, t.AsTypeReference(), restIndex, restSymbol)
```

internal/checker/printer.go
```diff
@@ -585,7 +585,7 @@ func (p *Printer) printSignature(sig *Signature, returnSeparator string) {
 		p.printType(p.c.getTypeOfSymbol(sig.thisParameter))
 		tail = true
 	}
-	expandedParameters := p.c.GetExpandedParameters(sig)
+	expandedParameters := p.c.getExpandedParameters(sig, true)[0]
```
andrewbranch:
```suggestion
	expandedParameters := p.c.getExpandedParameters(sig, true /*skipUnionExpanding*/)[0]
```

internal/checker/services.go
```diff
@@ -481,3 +486,13 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
 func (c *Checker) GetContextualTypeForJsxAttribute(attribute *ast.JsxAttributeLike) *Type {
 	return c.getContextualTypeForJsxAttribute(attribute, ContextFlagsNone)
 }
+
+func (c *Checker) getResolvedSignatureWorker(node *ast.Node, candidatesOutArray *[]*Signature, checkMode CheckMode, argumentCount int) *Signature {
+	parsedNode := printer.NewEmitContext().ParseNode(node)
+	c.apparentArgumentCount = &argumentCount
+	var res *Signature = nil
```
andrewbranch:
```suggestion
	var res *Signature
```

internal/checker/services.go
```diff
@@ -481,3 +486,13 @@ func (c *Checker) GetJsxIntrinsicTagNamesAt(location *ast.Node) []*ast.Symbol {
 func (c *Checker) GetContextualTypeForJsxAttribute(attribute *ast.JsxAttributeLike) *Type {
 	return c.getContextualTypeForJsxAttribute(attribute, ContextFlagsNone)
 }
+
+func (c *Checker) getResolvedSignatureWorker(node *ast.Node, candidatesOutArray *[]*Signature, checkMode CheckMode, argumentCount int) *Signature {
+	parsedNode := printer.NewEmitContext().ParseNode(node)
+	c.apparentArgumentCount = &argumentCount
```
andrewbranch, line 523: You forgot to unset this at the end of the function.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+// a set of parameters.
+type signatureInformation struct {
+	// The Label of this signature. Will be shown in
+	// the UI.
+	Label string
+	// The human-readable doc-comment of this signature. Will be shown
+	// in the UI but can be omitted.
+	Documentation *string
+	// The Parameters of this signature.
+	Parameters *[]signatureHelpParameter
```
andrewbranch: Don't use a pointer to a slice if you can avoid it. You range over this without checking for nil, so it seems like you can avoid it.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+	// Needed only here, not in lsp
+	IsVariadic bool
+}
+
+type signatureHelpItemInfo struct {
+	isVariadic   bool
+	parameters   *[]signatureHelpParameter
```
andrewbranch: Pointer to slice

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+
+func getExpressionFromInvocation(argumentInfo *argumentListInfo) *ast.Node {
+	if argumentInfo.invocation.callInvocation != nil {
+		return ast.GetInvokedExpression(argumentInfo.invocation.callInvocation.node)
+	}
+	return argumentInfo.invocation.typeArgsInvocation.called.AsNode()
+}
+
+type candidateInfo struct {
+	candidates        *[]*checker.Signature
```
andrewbranch: Pointer to slice

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+
+	return signatureInformation{
+		Label:         strings.Join(displayParts, ""),
+		Documentation: nil,
+		Parameters:    &parameters,
+		IsVariadic:    false,
+	}
+}
+
+func createSignatureHelpItems(candidates *[]*checker.Signature, resolvedSignature *checker.Signature, argumentInfo *argumentListInfo, sourceFile *ast.SourceFile, c *checker.Checker, useFullPrefix bool, clientOptions *lsproto.SignatureHelpClientCapabilities) *lsproto.SignatureHelp {
```
andrewbranch: Pointer to slice

internal/ls/signaturehelp.go
andrewbranch: This slice can be allocated with the correct length/capacity

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+	var parameters []*lsproto.ParameterInformation
+	for _, param := range *item.Parameters {
+		parameters = append(parameters, param.parameterInfo)
+	}
+	signatureInformation := []*lsproto.SignatureInformation{}
+	signatureInformation = append(signatureInformation, &lsproto.SignatureInformation{
+		Label:         item.Label,
+		Documentation: nil,
+		Parameters:    &parameters,
+	})
```
andrewbranch: This can be a single initialization

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+	var symbolString strings.Builder
+	symbolString.WriteString(c.SymbolToString(symbol))
+
+	var parameters []signatureHelpParameter = []signatureHelpParameter{}
+	for _, typeParam := range typeParameter {
+		parameters = append(parameters, createSignatureHelpParameterForTypeParameter(typeParam, sourceFile, enclosingDeclaration, c, printer))
+	}
+
+	// Creating display label
+	typeParameterDisplayParts := []string{}
```
andrewbranch: I would use a `strings.Builder` here

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+		}
+	}
+
+	callTargetDisplayParts := []string{}
+	if callTargetSymbol != nil {
+		callTargetDisplayParts = append(callTargetDisplayParts, c.SymbolToString(callTargetSymbol))
+	}
+	var items [][]signatureInformation
+	for _, candidateSignature := range *candidates {
+		items = append(items, getSignatureHelpItem(candidateSignature, argumentInfo.isTypeParameterList, callTargetDisplayParts, enclosingDeclaration, sourceFile, c))
```
andrewbranch: This is the only call to `getSignatureHelpItem`, and `callTargetDisplayParts` is always length 0 or 1. It could just be a string. We know we're not going to have display parts anymore, so it doesn't seem like there's a need to keep the slices around. A lot of things can probably just become strings.

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+		if callTargetSymbol == nil && useFullPrefix && resolvedSignature.Declaration() != nil {
+			callTargetSymbol = resolvedSignature.Declaration().Symbol()
+		}
+	}
+
+	callTargetDisplayParts := []string{}
+	if callTargetSymbol != nil {
+		callTargetDisplayParts = append(callTargetDisplayParts, c.SymbolToString(callTargetSymbol))
+	}
+	var items [][]signatureInformation
```
andrewbranch: This can be allocated with the correct length/capacity

internal/ls/signaturehelp.go
```diff
@@ -0,0 +1,1123 @@
...
+					if j.IsVariadic || len(*j.Parameters) >= argumentInfo.argumentCount {
+						selectedItemIndex = itemSeen + count
+						break
+					}
+					count++
+				}
+			}
+		}
+		itemSeen = itemSeen + len(item)
+	}
```
andrewbranch: This looks terrifying but I see it's ported exactly 😅

andrewbranch (changes requested):
This is looking pretty good overall. Beyond the comments I left, I suggest you look at every declaration of a slice and ask these questions:

- is this a `displayParts` slice that should just be a string or a `strings.Builder`?
- if not, can I allocate it with the correct capacity up front?

internal/ls/signaturehelp.go
navya9singh: I have removed the pointer here but still had to create a new `candidatesOutArray := &[]*Signature{}` in `getResolvedSignatureWorker` as `checker.getResolvedSignature()` needs a pointer. I tried removing a pointer from `getResolvedSignature` as well but there are other functions that need it and are not yet being used fully in Corsa but are used at different positions in Strada, so it is hard to say if the pointer will or will not be needed in the future.

internal/checker/services.go
```diff
@@ -299,10 +300,17 @@ func runWithInferenceBlockedFromSourceNode[T any](c *Checker, node *ast.Node, fn
 	return result
 }
 
+func GetResolvedSignatureForSignatureHelp(node *ast.Node, argumentCount int, c *Checker) (*Signature, []*Signature) {
+	var candidatesOutArray []*Signature
+	return runWithoutResolvedSignatureCaching(c, node, func() *Signature {
+		var res *Signature
+		res, candidatesOutArray = c.getResolvedSignatureWorker(node, CheckModeIsForSignatureHelp, argumentCount)
+		return res
+	}), candidatesOutArray
```
jakebailey: I think this technically works due to the evaluation order of multi-returns, but it kinda hurts my head. I'd personally rather this callback return a struct with the two results (which are then unpacked or returned as-is), or this call be split out into its own line before the return.

jakebailey: e.g.

```go
func GetResolvedSignatureForSignatureHelp(node *ast.Node, argumentCount int, c *Checker) (*Signature, []*Signature) {
	type result struct {
		signature *signature
		candidates []*signature
	}
	result := runWithoutResolvedSignatureCaching(c, node, func() result {
		signature, candidates := c.getResolvedSignatureWorker(node, CheckModeIsForSignatureHelp, argumentCount)
		return result{signature, candidates}
	})
	return result.signature, result.candidates
}
```

PR #863, opened 2025-05-13
JSX completions

gabritto:
Fill in the JSX-related places in completions.
Follow-up of #858.

### DIFF
--- a/internal/ls/completions.go
+++ b/internal/ls/completions.go
@@ -3875,7 +3875,8 @@ func (l *LanguageService) getJsxClosingTagCompletion(
 	location *ast.Node,
 	file *ast.SourceFile,
 	position int,
-	clientOptions *lsproto.CompletionClientCapabilities) *lsproto.CompletionList {
+	clientOptions *lsproto.CompletionClientCapabilities,
+) *lsproto.CompletionList {
 	// We wanna walk up the tree till we find a JSX closing element.
 	jsxClosingElement := ast.FindAncestorOrQuit(location, func(node *ast.Node) ast.FindAncestorResult {
 		switch node.Kind {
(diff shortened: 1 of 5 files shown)


internal/ls/completions.go
```diff
@@ -1907,128 +1920,29 @@ func (l *LanguageService) createCompletionItem(
 		// Otherwise use the completion list default.
 	}
 
-	// Text edit
```
gabritto, line 1910: This all got moved to `createLSPCompletionItem` so it can be reused in other places that also created completion entries and now need to create an LSP completion item.

internal/ls/completions.go
```diff
@@ -1907,128 +1920,29 @@ func (l *LanguageService) createCompletionItem(
...
-		SortText:         ptrTo(string(sortText)),
-		FilterText:       strPtrTo(filterText),
-		InsertText:       strPtrTo(insertText),
-		InsertTextFormat: insertTextFormat,
-		TextEdit:         textEdit,
-		CommitCharacters: commitCharacters,
-		Data:             nil, // !!! auto-imports
-	}
+	return l.createLSPCompletionItem(
+		name,
```
jakebailey, line 1927: Man that's a lot of parameters 😅

internal/ls/completions.go
```diff
@@ -1622,7 +1635,7 @@ func (l *LanguageService) getCompletionEntriesFromSymbols(
 	for name := range maps.Keys(uniques) {
 		uniqueSet.Add(name)
 	}
-	return uniqueSet, sortedEntries
+	return *uniqueSet, sortedEntries
```
jakebailey, line 1638: Is this just so that the uniqueSet can be optional in some way? Could we just use nil/empty?

internal/ls/completions.go
gabritto: Yeah. The alternative I guess would be to create a struct? We usually haven't done it for completions in Strada, but idk.

internal/ls/completions.go
gabritto: What do you mean? In general I'm avoiding using `*core.Set` because it crashes if you call `set.Has()`. The dereference is here just because `core.NewSetWithSizeHint` returns a pointer.

jakebailey: I see, yeah. I guess we just have a mix of places that need to read or write this so what pointer-ness to use isn't super clear.

internal/ls/completions.go
jakebailey: I don't mind right now, so long as it's unlikely they get swapped around wrong. I guess it's probably not too much less safe than Strada

jakebailey (approved):

PR #864, opened 2025-05-13
Align Quick Info display more closely with old language service

ahejlsberg:
This PR more closely aligns Quick Info display (a.k.a. "hover") with the old language service.


internal/checker/printer.go
```diff
@@ -85,6 +85,99 @@ func (c *Checker) SourceFileWithTypes(sourceFile *ast.SourceFile) string {
...
+	}
+	if symbol == nil || symbol == c.unknownSymbol {
+		return ""
+	}
+	flags := symbol.Flags
+	if flags&ast.SymbolFlagsType != 0 && !ast.IsInExpressionContext(node) {
+		// If the symbol has a type meaning and we're not in an expression context, remove any value meanings
+		flags &^= ast.SymbolFlagsValue
+	}
+	p := c.newPrinter(TypeFormatFlagsNone)
```
jakebailey, line 97: Presumably this is just temporary given I'm sure we'll want to stick this in the LS proper at some point?

internal/checker/printer.go
```diff
@@ -79,6 +79,99 @@ func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags
 	return p.string()
 }
 
+func (c *Checker) GetQuickInfoAtLocation(node *ast.Node) string {
+	symbol := c.GetSymbolAtLocation(node)
+	if symbol != nil && symbol.Flags&ast.SymbolFlagsAlias != 0 {
+		symbol = c.resolveAlias(symbol)
```
jakebailey, line 87: Maybe this is just extra, but I think that if you remembered if this path were hit, you could get `(alias)` too, like tsserver:

image

Versus the current PR's:

image

jakebailey (approved):

internal/checker/printer.go
ahejlsberg: Yup, will do.

internal/checker/printer.go
ahejlsberg: Yes, pretty much everything in `printer.go` is temporary.

jakebailey (approved):
Nice!

PR #865, opened 2025-05-14
Properly visit the `Expression` node of a `TypeParameterDeclaration`

ahejlsberg:
The missing `visit(v, node.Expression)` caused a crash in the language service for a type parameter with an incomplete `extends` clause.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -2063,8 +2063,8 @@ type TypeParameterDeclaration struct {
 	typeSyntaxBase
 	name        *IdentifierNode // IdentifierNode
 	Constraint  *TypeNode       // TypeNode. Optional
+	Expression  *Expression     // Expression. Optional, for error recovery purposes
 	DefaultType *TypeNode       // TypeNode. Optional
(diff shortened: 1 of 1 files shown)


internal/ast/ast.go
```diff
@@ -2084,7 +2084,7 @@ func (f *NodeFactory) UpdateTypeParameterDeclaration(node *TypeParameterDeclarat
 }
 
 func (node *TypeParameterDeclaration) ForEachChild(v Visitor) bool {
-	return visitModifiers(v, node.modifiers) || visit(v, node.name) || visit(v, node.Constraint) || visit(v, node.DefaultType)
+	return visitModifiers(v, node.modifiers) || visit(v, node.name) || visit(v, node.Constraint) || visit(v, node.DefaultType) || visit(v, node.Expression)
```
jakebailey: I think this actually needs to come before `DefaultType`, given where the parser parses it.

ahejlsberg: Got it. Now fixed.

PR #866, opened 2025-05-14
Fix LSP line/pos at start of a line

jakebailey:
Before:

image

After:

image

We really need tests for this conversion but this should be sufficient to fix this for demo purposes.

Probably fixes the global completions PR?


gabritto (approved):

jakebailey:
Yep, works:

image

PR #869, opened 2025-05-16
Concurrent LSP with cancelation

andrewbranch:
This PR asyncifies the LSP serve and enables parallel execution of requests.

At the top level, one goroutine reads and parses messages from STDIN and puts them onto a queue to be handled. Another pulls from that queue and dispatches them to the project service, and puts the response objects onto another queue. A third goroutine pulls the response messages and writes them to STDOUT.

The dispatcher categorizes requests into ones that modify state and ones that only query the existing state. The former are handled synchronously within the dispatcher goroutine, while the latter are spun off into their own goroutines so they can execute in parallel.

Executing multiple requests in parallel agains the same program introduces two challenges:

1. Type checkers are not concurrency safe, so each request needs to use a different one. Further, a request must always get the _same_ checker if it asks the Program for one at two different times. For this, I’ve introduced a `CheckerPool` interface whose methods for getting a checker take a `context.Context` and return an available checker, along with a function to return it to the pool when done. The implementation in `compiler` used for `tsc` is just the logic that was already built into Program, while the implementation in `project` creates checkers dynamically and tracks which ones are in use by which requests.
2. The text content of files the request handler inspects must be consistent over the duration of the request. Previously, line/character positions were converted to and from absolute positions in the server layer by looking at the text storage layer owned by the project system, which is mutable over time. With this PR, I’ve made all the LanguageService methods accept and return `lsproto` types (which are exclusively line/character) so all the conversions have to happen in that layer. Then, I removed LanguageService’s ability to access any mutable state (basically). When it asks for the line map of a file, its host attempts to pull the cached one from the text storage layer only if the versions match. Otherwise, the text storage has been updated by another request, and the line map for the old version is computed on demand from the SourceFile text, which is immutable.

### DIFF
--- a/internal/compiler/checkerpool.go
+++ b/internal/compiler/checkerpool.go
@@ -1,0 +1,90 @@
+package compiler
+
+import (
+	"context"
+	"iter"
+	"slices"
+	"sync"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/checker"
+	"github.com/microsoft/typescript-go/internal/core"
+)
+
+type CheckerPool interface {
+	GetChecker(ctx context.Context) (*checker.Checker, func())
+	GetCheckerForFile(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func())
+	GetAllCheckers(ctx context.Context) ([]*checker.Checker, func())
+	Files(checker *checker.Checker) iter.Seq[*ast.SourceFile]
+}
+
+type checkerPool struct {
+	checkerCount int
+	program      *Program
+
+	createCheckersOnce sync.Once
+	checkers           []*checker.Checker
+	fileAssociations   map[*ast.SourceFile]*checker.Checker
+}
+
+var _ CheckerPool = (*checkerPool)(nil)
+
+func newCheckerPool(checkerCount int, program *Program) *checkerPool {
+	pool := &checkerPool{
+		program:      program,
+		checkerCount: checkerCount,
+		checkers:     make([]*checker.Checker, checkerCount),
+	}
+
+	return pool
+}
+
+func (p *checkerPool) GetCheckerForFile(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func()) {
+	p.createCheckers()
+	checker := p.fileAssociations[file]
+	return checker, noop
+}
+
+func (p *checkerPool) GetChecker(ctx context.Context) (*checker.Checker, func()) {
+	p.createCheckers()
+	checker := p.checkers[0]
+	return checker, noop
+}
+
+func (p *checkerPool) createCheckers() {
+	p.createCheckersOnce.Do(func() {
+		wg := core.NewWorkGroup(p.program.singleThreaded())
+		for i := range p.checkerCount {
+			wg.Queue(func() {
+				p.checkers[i] = checker.NewChecker(p.program)
+			})
+		}
+
+		wg.RunAndWait()
+
+		p.fileAssociations = make(map[*ast.SourceFile]*checker.Checker, len(p.program.files))
+		for i, file := range p.program.files {
+			p.fileAssociations[file] = p.checkers[i%p.checkerCount]
+		}
+	})
+}
+
+func (p *checkerPool) GetAllCheckers(ctx context.Context) ([]*checker.Checker, func()) {
+	p.createCheckers()
+	return p.checkers, noop
+}
+
+func (p *checkerPool) Files(checker *checker.Checker) iter.Seq[*ast.SourceFile] {
+	checkerIndex := slices.Index(p.checkers, checker)
+	return func(yield func(*ast.SourceFile) bool) {
+		for i, file := range p.program.files {
+			if i%p.checkerCount == checkerIndex {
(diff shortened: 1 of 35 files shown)


internal/compiler/emitHost.go
```diff
@@ -32,7 +34,8 @@ func (host *emitHost) WriteFile(fileName string, text string, writeByteOrderMark
 }
 
 func (host *emitHost) GetEmitResolver(file *ast.SourceFile, skipDiagnostics bool) printer.EmitResolver {
-	checker := host.program.GetTypeCheckerForFile(file)
+	checker, done := host.program.GetTypeCheckerForFile(context.Background(), file)
```
jakebailey: Do we need a TODO here to indicate that this is subtly wrong?

internal/ls/utilities.go
```diff
@@ -268,10 +268,7 @@ func (l *LanguageService) createLspRangeFromNode(node *ast.Node, file *ast.Sourc
 }
 
 func (l *LanguageService) createLspRangeFromBounds(start, end int, file *ast.SourceFile) *lsproto.Range {
-	lspRange, err := l.converters.ToLSPRange(file.FileName(), core.NewTextRange(start, end))
-	if err != nil {
-		panic(err)
-	}
+	lspRange := l.converters.ToLSPRange(file, core.NewTextRange(start, end))
 	return &lspRange
```
jakebailey, line 272: Not related to anything you're doing here but I think it's weird that we have this pointer to range when most everyone actually just dereferences it immediately. (I think it's used in one place where it's optional but it seems to me like the pointer-ness should be moved down there.)

internal/lsp/server.go
```diff
@@ -151,105 +172,199 @@ func (s *Server) UnwatchFiles(handle project.WatcherHandle) error {
...
+	g.Go(func() error { return s.readLoop(ctx) })
+	return g.Wait()
+}
+
+func (s *Server) readLoop(ctx context.Context) error {
 	for {
-		req, err := s.read()
+		msg, err := s.read()
 		if err != nil {
+			if err == io.EOF {
```
jakebailey: I wonder when this regressed; errorlint used to complain about it, then it was fixed, and I guess now it's broken again?

internal/ls/utilities.go
andrewbranch: I'll leave it alone in this PR. cc @gabritto if she wants to consider changing it.

internal/lsp/server.go
andrewbranch: Wait, the linter _is_ complaining about this; I thought you told me to change it to be this way?

jakebailey: I did, but I guess that linter was not fully fixed to stop complaining about this ☹️

I guess feel free to `errors.Is` it and we can come back to this later.

jakebailey (approved):
LGTM with a `go mod tidy`.

gabritto (approved):

jakebailey:
I approved this but I think you said that there's some bug with duplicate program creation you want to address (so I'll happy look again after too)

ahejlsberg (approved):

internal/compiler/checkerpool.go
```diff
@@ -0,0 +1,90 @@
...
+		wg := core.NewWorkGroup(p.program.singleThreaded())
+		for i := range p.checkerCount {
+			wg.Queue(func() {
+				p.checkers[i] = checker.NewChecker(p.program)
+			})
+		}
+
+		wg.RunAndWait()
+
+		p.fileAssociations = make(map[*ast.SourceFile]*checker.Checker, len(p.program.files))
```
sheetalkamat, line 65: this is good for now but in theory the questions are asked only on open files so assign only for those ?

sheetalkamat: I meant in editor scenario not cli

internal/project/service.go
```diff
@@ -177,13 +179,30 @@ func (s *Service) OpenFile(fileName string, fileContent string, scriptKind core.
...
-	if info == nil {
-		panic("scriptInfo not found")
+	scriptInfo := s.GetScriptInfoByPath(path)
+	if scriptInfo == nil {
+		return fmt.Errorf("file %s not found", fileName)
+	}
+
+	textChanges := make([]ls.TextChange, len(changes))
+	for i, change := range changes {
+		if partialChange := change.TextDocumentContentChangePartial; partialChange != nil {
```
sheetalkamat, line 192: There is delayUpdateProjectGraph in this file- in strada we use to schedule the update (unless new change came in - then we would cancel and restart the timer) - with this change does this mean - send the refresh diagnostics notification - instead of updating as we want to update on next request instead ?

internal/compiler/checkerpool.go
andrewbranch: This implementation is only used in the CLI. In the editor this map is created with no capacity hint.

jakebailey (approved):

internal/project/service.go
andrewbranch: We don't send the diagnostic refresh notification from here. The client will ask for new diagnostics after an update, so I don't think there's any need for us to schedule anything.

PR #873, opened 2025-05-16
Add missing bounds check in `isReachableFlowNodeWorker`

### LINKED ISSUE #871: Crash on assertion predicate with optional argument
```ts
export function assertWeird(value?: string): asserts value {
}

assertWeird();
assertWeird("hello");
```

Note that order of the calls is important to reproduce the issue.

ahejlsberg:
Fixes #871.

### DIFF
--- a/internal/checker/flow.go
+++ b/internal/checker/flow.go
@@ -2484,13 +2484,12 @@ func (c *Checker) isReachableFlowNodeWorker(f *FlowState, flow *ast.FlowNode, no
 		case flags&(ast.FlowFlagsAssignment|ast.FlowFlagsCondition|ast.FlowFlagsArrayMutation) != 0:
 			flow = flow.Antecedent
 		case flags&ast.FlowFlagsCall != 0:
-			signature := c.getEffectsSignature(flow.Node)
-			if signature != nil {
-				predicate := c.getTypePredicateOfSignature(signature)
-				if predicate != nil && predicate.kind == TypePredicateKindAssertsIdentifier && predicate.t == nil {
-					predicateArgument := flow.Node.Arguments()[predicate.parameterIndex]
-					if predicateArgument != nil && c.isFalseExpression(predicateArgument) {
-						return false
+			if signature := c.getEffectsSignature(flow.Node); signature != nil {
+				if predicate := c.getTypePredicateOfSignature(signature); predicate != nil && predicate.kind == TypePredicateKindAssertsIdentifier && predicate.t == nil {
+					if arguments := flow.Node.Arguments(); int(predicate.parameterIndex) < len(arguments) {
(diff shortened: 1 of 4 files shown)


internal/checker/flow.go
```diff
@@ -2484,13 +2484,12 @@ func (c *Checker) isReachableFlowNodeWorker(f *FlowState, flow *ast.FlowNode, no
...
-			if signature != nil {
-				predicate := c.getTypePredicateOfSignature(signature)
-				if predicate != nil && predicate.kind == TypePredicateKindAssertsIdentifier && predicate.t == nil {
-					predicateArgument := flow.Node.Arguments()[predicate.parameterIndex]
-					if predicateArgument != nil && c.isFalseExpression(predicateArgument) {
-						return false
+			if signature := c.getEffectsSignature(flow.Node); signature != nil {
+				if predicate := c.getTypePredicateOfSignature(signature); predicate != nil && predicate.kind == TypePredicateKindAssertsIdentifier && predicate.t == nil {
+					if arguments := flow.Node.Arguments(); int(predicate.parameterIndex) < len(arguments) {
+						if predicateArgument := arguments[predicate.parameterIndex]; predicateArgument != nil && c.isFalseExpression(predicateArgument) {
```
DanielRosenwasser: I don't think you need `predicateArgument != nil` once you have the bounds check.

```suggestion
						if predicateArgument := arguments[predicate.parameterIndex]; c.isFalseExpression(predicateArgument) {
```

testdata/tests/cases/compiler/assertionWithNoArgument.ts
```diff
@@ -0,0 +1,8 @@
+// @strict: true
+// @noemit: true
+
+export function assertWeird(value?: string): asserts value {
+}
+
+assertWeird();
+assertWeird("hello");
```
DanielRosenwasser, line 8: Might be worth having another test that swaps the order?

DanielRosenwasser:
Any idea why two calls are necessary to trigger this crash?

internal/checker/flow.go
ahejlsberg: Agreed.

ahejlsberg:
> Any idea why two calls are necessary to trigger this crash?

You just need _some_ other statement after the argument-less call to trigger the situation. It's because in the call to `isReachableFlowNode` from `checkSourceElementWorker`, we're only going to examine the flow node graph for nodes that _precede_ the current node.

testdata/tests/cases/compiler/assertionWithNoArgument.ts
ahejlsberg: Nah, isn't really relevant. See my comment below.

jakebailey (approved):

PR #877, opened 2025-05-17
fix: avoid treating await in jsdoc as top-level await

### LINKED ISSUE #874: JSDoc parsing breaks when declaring a property named `await`
Found checking webpack. Create a JS file and use `checkJs`.

```js
/**
 * @typedef {object} Settings
 * @property {boolean} await
 */

const x = 42;
```

a-tarasyuk:
Fixes #874

---

`parseJSDocIdentifierName` uses `newIdentifier`, which may flag `await` identifier as a top-level declaration. @sandersn should this be suppressed when parsing JSDoc?

https://github.com/microsoft/typescript-go/blob/cab7a0bf793f3ba0e83619798f2eb191673fe4c7/internal/parser/jsdoc.go#L1253

https://github.com/microsoft/typescript-go/blob/cab7a0bf793f3ba0e83619798f2eb191673fe4c7/internal/parser/parser.go#L2860-L2867

### DIFF
--- a/internal/parser/jsdoc.go
+++ b/internal/parser/jsdoc.go
@@ -129,6 +129,7 @@ func (p *Parser) parseJSDocComment(parent *ast.Node, start int, end int, fullSta
 	saveScannerState := p.scanner.Mark()
 	saveDiagnosticsLength := len(p.diagnostics)
 	saveHasParseError := p.hasParseError
+	saveHasAwaitIdentifier := p.statementHasAwaitIdentifier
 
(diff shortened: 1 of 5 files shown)


sandersn (approved):

sandersn:
The trailing `p.statementHasAwaitIdentifier = saveHasAwaitIdentifier` should be the right code to suppress it; it should discard the true value before parseToplevelStatement examines it and adds the statement to the possibleAwaitSpans.

PR #879, opened 2025-05-18
Introduce fast path for program updates in language service

ahejlsberg:
This PR introduces a fast path for program updates in the language service. When only a single file has been modified and none of the changes impact the program's structure, we create a new program by just cloning the existing program and replacing the single file. This dramatically improves responsiveness in the typical path of typing in a file.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -2,6 +2,7 @@ package compiler
 
 import (
 	"context"
+	"maps"
 	"slices"
 	"sync"
 
@@ -78,11 +79,7 @@ func NewProgram(options ProgramOptions) *Program {
 	if p.compilerOptions == nil {
 		p.compilerOptions = &core.CompilerOptions{}
 	}
-	if p.programOptions.CreateCheckerPool != nil {
-		p.checkerPool = p.programOptions.CreateCheckerPool(p)
-	} else {
-		p.checkerPool = newCheckerPool(core.IfElse(p.singleThreaded(), 1, 4), p)
-	}
+	p.initCheckerPool()
 
 	// p.maxNodeModuleJsDepth = p.options.MaxNodeModuleJsDepth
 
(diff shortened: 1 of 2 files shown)


internal/project/project.go
```diff
@@ -372,14 +373,23 @@ func (p *Project) getScriptKind(fileName string) core.ScriptKind {
 }
 
 func (p *Project) markFileAsDirty(path tspath.Path) {
-	p.markAsDirty()
+	p.dirtyStateMu.Lock()
+	defer p.dirtyStateMu.Unlock()
+	if !p.dirty {
+		p.dirty = true
+		p.dirtyFilePath = path
```
jakebailey, line 380: I think this may need to be a `Set` or something; file watching can cause this to be overwritten with subsequent updates, or even open files on branch switches or similar.

ahejlsberg: Right, but the minute *anything* else changes, we clear out `p.dirtyFilePath` and `updateGraph` will then completely re-create the program.

internal/project/project.go
```diff
@@ -432,6 +442,7 @@ func (p *Project) updateGraph() bool {
 	p.hasAddedOrRemovedSymlinks = false
 	p.updateProgram()
 	p.dirty = false
+	p.dirtyFilePath = ""
```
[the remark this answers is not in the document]
jakebailey: why have these comments gotten so bad ☹️

internal/project/project.go
jakebailey: Ah, I see what you mean in `updateProgram`; you're right that it's safe. I guess we don't need to be super optimal (though I think it might just be a couple lines to keep track of multiple dirty files).

internal/compiler/program.go
```diff
@@ -174,6 +171,75 @@ func NewProgram(options ProgramOptions) *Program {
...
+		file1.IsDeclarationFile == file2.IsDeclarationFile &&
+		file1.HasNoDefaultLib == file2.HasNoDefaultLib &&
+		file1.UsesUriStyleNodeCoreModules == file2.UsesUriStyleNodeCoreModules &&
+		slices.EqualFunc(file1.Imports, file2.Imports, compareImports) &&
+		slices.EqualFunc(file1.ModuleAugmentations, file2.ModuleAugmentations, compareModuleAugmentations) &&
+		slices.Equal(file1.AmbientModuleNames, file2.AmbientModuleNames) &&
+		slices.EqualFunc(file1.ReferencedFiles, file2.ReferencedFiles, compareFileReferences) &&
+		slices.EqualFunc(file1.TypeReferenceDirectives, file2.TypeReferenceDirectives, compareFileReferences) &&
+		slices.EqualFunc(file1.LibReferenceDirectives, file2.LibReferenceDirectives, compareFileReferences) &&
+		file1.CheckJsDirective == file2.CheckJsDirective
```
jakebailey: These are pointers; probably needs to be a more complicated function which checks pointer equality, nilness, and the underlying values?

ahejlsberg: It sure does.

internal/project/project.go
```diff
@@ -454,22 +465,26 @@ func (p *Project) updateGraph() bool {
...
-		CreateCheckerPool: func(program *compiler.Program) compiler.CheckerPool {
-			p.checkerPool = newCheckerPool(4, program, p.log)
-			return p.checkerPool
-		},
-	})
-
+	if p.program == nil || p.dirtyFilePath == "" {
+		rootFileNames := p.GetRootFileNames()
+		compilerOptions := p.compilerOptions
+		p.program = compiler.NewProgram(compiler.ProgramOptions{
```
andrewbranch, line 475: If we somehow propagate out whether the old program was cloned or not, we can use that to skip reevaluating the file watchers at the end of `updateGraph` too.

internal/project/project.go
sheetalkamat: Is this temporary? because we have this https://github.com/microsoft/TypeScript/blob/main/src/compiler/program.ts#L2458 in strada thats much more resilient to multiple file changes (can happen as part of open and close when navigating source code)

andrewbranch: But as a trade-off, typing performance in Strada is extremely poor compared to what it could be. We definitely need new optimizations for the 99 percentile case of typing in a single file. Program construction shouldn’t have to loop over every source file to determine what’s changed when the project infrastructure already _knows_ what changed.

internal/project/project.go
ahejlsberg: Now in my latest commit.

andrewbranch (approved):

ahejlsberg:
Just as an FYI, with this PR, diagnostics requests in `checker.ts` (a 50K line file in the old TS code base) take about 60ms following a change in the file. In that time we do full parsing, binding, and checking of the file. With that level of performance it is clear that we no longer need the incremental parse logic we had in the old compiler.

internal/project/project.go
```diff
@@ -372,14 +373,23 @@ func (p *Project) getScriptKind(fileName string) core.ScriptKind {
...
+		p.dirtyFilePath = ""
+	}
 }
 
 func (p *Project) markAsDirty() {
 	p.dirtyStateMu.Lock()
 	defer p.dirtyStateMu.Unlock()
 	if !p.dirty {
 		p.dirty = true
+		p.dirtyFilePath = ""
```
gabritto, line 392: I think we need to reset `p.dirtyFilePath` unconditionally here. Suppose we call `p.markFileAsDirty()` and it sets `p.dirtyFilePath` to some path, and then right after we call `p.markAsDirty()`: we won't reset `p.dirtyFilePath` because `p.dirty` is already true.

ahejlsberg: Good catch!

PR #881, opened 2025-05-20
feat: display goos & goarch

sxzz:
Since the build of TypeScript Go depends on the operating system and CPU architecture, it also outputs `GOOS` and `GOARCH` accordingly.

```
❯ tsc -v
Version 7.0.0-dev darwin/arm64

❯ tsc -h
tsc: The TypeScript Compiler - Version 7.0.0-dev - darwin/arm64
...
```

### DIFF
--- a/internal/execute/outputs.go
+++ b/internal/execute/outputs.go
@@ -95,7 +95,7 @@ func reportStatistics(sys System, program *compiler.Program, result compileAndEm
 }
 
 func printVersion(sys System) {
-	fmt.Fprint(sys.Writer(), diagnostics.Version_0.Format(core.Version)+sys.NewLine())
+	fmt.Fprint(sys.Writer(), diagnostics.Version_0.Format(core.Version)+" "+runtime.GOOS+"/"+runtime.GOARCH+sys.NewLine())
 	sys.EndWrite()
 }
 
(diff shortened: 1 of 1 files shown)


jakebailey:
I don't think this is useful; a user probably already knows what kind of system they're running on. That and, this text cannot change as it's a "load-bearing" string that people parse.

LukeAbby:
> That and, this text cannot change as it's a "load-bearing" string that people parse.

Horrifying but I guess I shouldn't be surprised.

sxzz:
Some operating systems, including macOS and Windows, support running x86/x64 programs on Arm hardware through compatibility layers, but performance is significantly reduced. Providing this information helps users identify and choose the binary that best suits their system.

sxzz:
Considering compatibility with 5.x, do you think this minor change is acceptable?
  
As for parsing, the `tsc -v` command still separates the version with a space, ensuring that commands like `tsc -v | awk '{print $2}'` (or similar approaches) will continue to work without any changes.

jakebailey:
I just don't think it's required. I don't know of any other application which does this, outside `go` itself, which is a compiler that produces native code. The GOOS/GOARCH that people see will just match their version of Node, because it'll be installed from npm.

sxzz:
This information is already included in all CLI programs that use CAC. See https://www.npmjs.com/browse/depended/cac

```
❯ pnpx vite -v
vite/6.3.5 darwin-arm64 node-v22.16.0
❯ pnpx vitest -v
vitest/3.1.4 darwin-arm64 node-v22.16.0
❯ pnpx tsup -v
tsup/8.5.0 darwin-arm64 node-v22.16.0
❯ pnpx @unocss/cli -v
unocss/66.1.2 darwin-arm64 node-v22.16.0
```

Please feel free to close the PR if you don't find it helpful.

RyanCavanaugh:
Let's revisit this when the need arises. For now, we'd like to keep the TS6/TS7 delta as small as possible.

PR #882, opened 2025-05-20
findAllReferences inital port

iisaduan:
This port focuses on the logic needed to resolving symbols within a single file. 

Todos after this PR:
- Search for references through module resolving
- Search for references through imported and exported symbol resolving
  - Some imported and global symbols return references, but it is hard to test with the current `lstestutil` capabilities. 
- Jsdoc references (not sure how much of the currently un-ported parts will be needed, since @sandersn has been working to simplify jsdoc checking)
- Special cases for object literals
- Special cases for classes and keywords
- String references


internal/ast/utilities.go
```diff
@@ -2456,6 +2552,104 @@ func GetDeclarationContainer(node *Node) *Node {
...
+	AssignmentDeclarationKindObjectDefinePropertyValue
+	// Object.defineProperty(exports || module.exports, 'name', ...);
+	AssignmentDeclarationKindObjectDefinePropertyExports
+	// Object.defineProperty(Foo.prototype, 'name', ...);
+	AssignmentDeclarationKindObjectDefinePrototypeProperty
+)
+
+// / Given a BinaryExpression, returns SpecialPropertyAssignmentKind for the various kinds of property
+// / assignments we treat as special in the binder
+func GetAssignmentDeclarationKind(expr *Expression /*BinaryExpression | CallExpression*/) AssignmentDeclarationKind {
```
gabritto: I think we didn't port this to Corsa on purpose, because in Strada this existed because of all the JS weirdness, and the hope is we won't need this in Corsa, or it will be simplified. @sandersn can you confirm?

internal/checker/utilities.go
```diff
@@ -1007,6 +1007,23 @@ func getPropertyNameFromType(t *Type) string {
 	panic("Unhandled case in getPropertyNameFromType")
 }
 
+func GetEffectiveImplementsTypeNodes(node *ast.Node) []*ast.TypeNode {
```
gabritto: I think we won't need to port this given the new JSDoc approach is to add synthetic nodes precisely to avoid this kind of function. Same for `GetEffectiveBaseTypeNode` below. You should probably use the existing `GetImplementsTypeNodes`.

internal/core/core.go
```diff
@@ -327,6 +338,12 @@ func Coalesce[T *U, U any](a T, b T) T {
 	}
 }
 
+// Returns the first element that is not `nil`; CoalesceList(a, b, c) is roughly analogous to `a ?? b ?? c` in JS, except that it
+// non-shortcutting, so it is advised to only use a constant or precomputed value for non-first values in the list
+func CoalesceList[T *U, U any](a ...T) T {
```
gabritto, line 343: Given how easy it is to misuse `core.IfElse` already for the same reason, I wonder if it's a good idea to have this.

internal/ls/findallreferencesexport_test.go
```diff
@@ -0,0 +1,31 @@
+package ls
+
+import (
+	"github.com/microsoft/typescript-go/internal/astnav"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+)
+
+func (l *LanguageService) GetExpectedReferenceFromMarker(fileName string, pos int) *lsproto.Location {
```
gabritto, line 9: This is unusual but I guess it's ok if it's only temporary until we have fourslash.

internal/ls/utilities.go
```diff
@@ -11,13 +13,56 @@ import (
...
+	}
+	// If the symbol is declared as part of a declaration like `{ type: "a" } | { type: "b" }`, use the property on the union type to get more references.
+	return core.FirstNonNil(originalSymbol.Declarations, func(decl *ast.Node) *ast.Symbol {
+		if decl.Parent == nil {
+			// Ignore UMD module and global merge
+			if originalSymbol.Flags&ast.SymbolFlagsTransient != 0 {
+				return nil
+			}
+			// Assertions for GH#21814. We should be handling SourceFile symbols in `getReferencedSymbolsForModule` instead of getting here.
+			panic(`Unexpected symbol at ${Debug.formatast.Kind(node.Kind)}: ${Debug.formatSymbol(symbol)}`)
```
gabritto:
```suggestion
			panic(fmt.Sprintf("Unexpected symbol at %s: %s", node.Kind.String(), symbol.Name))
```

internal/ls/utilities.go
```diff
@@ -41,6 +86,58 @@ func tryGetImportFromModuleSpecifier(node *ast.StringLiteralLike) *ast.Node {
 	return nil
 }
 
+func isModuleSpecifierLike(node *ast.Node) bool {
+	// return node.Kind == ast.KindStringLiteral || node.Kind == ast.KindNoSubstitutionTemplateLiteral ||
```
gabritto: If this was just a refactoring replacing the explicit return with `ast.IsStringLiteralLike`, I'd just get rid of the comment.

internal/checker/checker.go
```diff
@@ -13588,6 +13588,36 @@ func (c *Checker) getExportSymbolOfValueSymbolIfExported(symbol *ast.Symbol) *as
 	return c.getMergedSymbol(symbol)
 }
 
+func (c *Checker) GetExportSpecifierLocalTargetSymbol(node *ast.Node) *ast.Symbol {
```
gabritto: This should probably be in `checker/services.go`.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1490 @@
...
+
+	// !!! check fall throughs
+	switch searchSpaceNode.Kind {
+	case ast.KindMethodDeclaration, ast.KindMethodSignature:
+		if ast.IsObjectLiteralMethod(searchSpaceNode) {
+			staticFlag &= searchSpaceNode.ModifierFlags()
+			searchSpaceNode = searchSpaceNode.Parent // re-assign to be the owning object literals
+			break
+		}
+		// falls through
```
gabritto: Is the comment here because you're not sure if the original code should be falling through?

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1490 @@
...
+// 		singleReferences []*ast.Node // ientifier | stringliteral
+// 		indirectUsers    []*ast.SourceFile
+// 	}
+// )
+
+type seenTracker[T comparable] struct {
+	m core.Set[T]
+}
+
+func (t *seenTracker[T]) mark(key T) bool {
```
gabritto: I think you could just get rid of the new struct and use `core.Set`, and add this as a set method. I recently added method `AddIfAbsent` to replace Strada's `addToSeen`.

internal/ls/findallreferencesexport_test.go
iisaduan: Yes, the testing exports is intended to be temporary, at least until we get fourslash more fleshed out

internal/core/core.go
iisaduan: I found this useful because expanding chains of || was getting really messy, and there were quite a few `baseSymbol|rootSymbol|originalSymbol|symbol` like statements in find all references, where the list is all precomputed variables. I could change it into just a `findAllReferences` helper function?

internal/ast/utilities.go
iisaduan: @sandersn I did notice there was a `getAssignmentDeclarationKind` included in #860, but wasn't sure what I should use in its place since it didn't include call expressions.

internal/checker/services.go
```diff
@@ -399,6 +400,26 @@ func (c *Checker) GetExportSymbolOfSymbol(symbol *ast.Symbol) *ast.Symbol {
...
+	classDeclaration := parameter.Parent.Parent
+
+	parameterSymbol := c.getSymbol(constructorDeclaration.Locals(), parameterName, ast.SymbolFlagsValue)
+	propertySymbol := c.getSymbol(c.getMembersOfSymbol(classDeclaration.Symbol()), parameterName, ast.SymbolFlagsValue)
+
+	if parameterSymbol != nil && propertySymbol != nil {
+		return parameterSymbol, propertySymbol
+	}
+
+	panic("There should exist two symbols, one as property declaration and one as parameter declaration")
```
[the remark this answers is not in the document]
iisaduan: this comment is ported directly from strada

internal/ast/utilities.go
sandersn: unfortunately, we still need GetAssignmentDeclarationKind =(

However, it is going to be much simpler, so you should use the one that's in main.

internal/ls/findallreferences.go
iisaduan: yup, was a reminder, I've now checked 👍

internal/ls/findallreferences.go
```diff
@@ -1040,9 +1020,9 @@ type refState struct {
 	result        []*SymbolAndEntries
 
 	inheritsFromCache            map[inheritKey]bool
-	seenContainingTypeReferences seenTracker[*ast.Node] // node seen tracker
-	// seenReExportRHS              seenTracker[*ast.Node] // node seen tracker
-	// importTracker                ImportTracker
+	seenContainingTypeReferences *core.Set[*ast.Node] // node seen tracker
```
jakebailey, line 1023: You can just make these `core.Set`, they'll auto initialize (and then you can drop it from the struct literal)

internal/ast/utilities.go
```diff
@@ -1029,6 +1029,26 @@ func IsStatic(node *Node) bool {
 	return IsClassElement(node) && HasStaticModifier(node) || IsClassStaticBlockDeclaration(node)
 }
 
+func CanHaveSymbol(node *Node) bool {
+	switch node.Kind {
```
sandersn, line 1040: All the JSDoc tags should be dropped from here, replaced with JSExportAssignment JSTypeAliasDeclaration, CommonJSExport (I think).
Also I'm pretty sure Identifier is JS-only and unsupported in Corsa, so can be dropped as well.

internal/ast/utilities.go
```diff
@@ -1290,6 +1314,37 @@ func IsThisParameter(node *Node) bool {
 	return IsParameter(node) && node.Name() != nil && IsThisIdentifier(node.Name())
 }
```
sandersn, line 1322: a lot of this syntax isn't supported, or supported yet, in Corsa. Merge fom main and use GetAssignmentDeclarationKind as-is.

internal/ast/utilities.go
```diff
@@ -1417,6 +1479,19 @@ const (
 	JSDeclarationKindThisProperty
 	/// F.name = expr, F[name] = expr
 	JSDeclarationKindProperty
+
+	// PropertyAccessKinds
+	// F.prototype = { ... }
+	JSDeclarationKindPrototype
```
sandersn: JSDeclarationKindPrototype isn't supported in Corsa, and Object.defineProperty is at the very bottom of my TODO list. Same advice as above.

internal/ast/utilities.go
```diff
@@ -1766,13 +1906,13 @@ func IsExpressionNode(node *Node) bool {
 		for node.Parent.Kind == KindQualifiedName {
 			node = node.Parent
 		}
-		return IsTypeQueryNode(node.Parent) || isJSDocLinkLike(node.Parent) || isJSXTagName(node)
+		return IsTypeQueryNode(node.Parent) || IsJSDocLinkLike(node.Parent) || isJSXTagName(node)
```
sandersn, line 1854: I have to wonder if I came up with the name `IsJSDocLinkLike` about the time I was reading Fox in Socks' "Luke Luck likes lakes, Luke's duck licks lakes" twice a day.

internal/ast/utilities.go
```diff
@@ -2456,6 +2596,10 @@ func GetDeclarationContainer(node *Node) *Node {
 	}).Parent
 }
 
+func IsPrototypeAccess(node *Node) bool {
```
sandersn: same advice as above

internal/core/set.go
```diff
@@ -48,6 +48,17 @@ func (s *Set[T]) AddIfAbsent(key T) bool {
 	return true
 }
 
+// Returns true if the key is arleady in the set. Adds the key and returns false otherwise.
```
sandersn:
```suggestion
// Returns true if the key is already in the set. Adds the key and returns false otherwise.
```

internal/checker/checker.go
```diff
@@ -13673,7 +13673,7 @@ func (c *Checker) getSymbolOfPartOfRightHandSideOfImportEquals(entityName *ast.N
 	//     import a = |b|; // Namespace
 	//     import a = |b.c|; // Value, type, namespace
 	//     import a = |b.c|.d; // Namespace
-	if entityName.Kind == ast.KindIdentifier && isRightSideOfQualifiedNameOrPropertyAccess(entityName) {
+	if entityName.Kind == ast.KindIdentifier && IsRightSideOfQualifiedNameOrPropertyAccess(entityName) {
```
sandersn: I *love* the way changing accessibility of a thing requires changing all references to it everywhere.

internal/ls/utilities.go
```diff
@@ -11,6 +13,24 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
+// p.Compare(other) == cmp.Compare(p, other)
```
sandersn: I don't understand these comments.

internal/ls/utilities.go
```diff
@@ -41,6 +61,56 @@ func tryGetImportFromModuleSpecifier(node *ast.StringLiteralLike) *ast.Node {
 	return nil
 }
 
+func isModuleSpecifierLike(node *ast.Node) bool {
+	if !ast.IsStringLiteralLike(node) {
+		return false
+	}
+
+	if ast.IsRequireCall(node.Parent /*requireStringLiteralLikeArgument*/) || ast.IsImportCall(node.Parent) {
```
sandersn: Are all parent pointers set at this point? IsVariableDeclarationInitializedToRequire is a more complete check but you need to be able to pass node.Parent.Parent to it. (see checker's GetSymbolAtLocation for a similar usage.)
```suggestion
	if ast.IsVariableDeclarationInitializedToRequire(node.Parent.Parent) || ast.IsImportCall(node.Parent) {
```

internal/ls/utilities.go
```diff
@@ -41,6 +61,56 @@ func tryGetImportFromModuleSpecifier(node *ast.StringLiteralLike) *ast.Node {
...
+		return node.Parent.AsCallExpression().Arguments.Nodes[0] == node
+	}
+
+	return node.Parent.Kind == ast.KindExternalModuleReference ||
+		node.Parent.Kind == ast.KindImportDeclaration ||
+		node.Parent.Kind == ast.KindJSDocImportTag
```
sandersn:
```suggestion
		node.Parent.Kind == ast.KindJSImportDeclaration
```

internal/ls/utilities.go
```diff
@@ -391,6 +547,36 @@ func nodeIsASICandidate(node *ast.Node, file *ast.SourceFile) bool {
 	return startLine != endLine
 }
 
+func getNextJSDocCommentLocation(node *ast.Node) *ast.Node {
```
sandersn: I don't think this should be needed, but if it is, it'll likely need to be completely different.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+		}
+
+		if container == nil || (container.Kind == ast.KindSourceFile && !ast.IsExternalOrCommonJSModule(container.AsSourceFile())) {
+			// This is a global variable and not an external module, any declaration defined
+			// within this scope is visible outside the file
+			return nil
+		}
+
+		scope = container
+		if scope.Kind == ast.KindFunctionExpression {
```
sandersn: Unless JSDoc comments are being used as a proxy for something else, this is not needed. Function expressions will have synthetic types in situ instead of needing to look outward to JSDoc.

internal/ls/utilities.go
sandersn: No, it looks unneeded.

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		return getContainingNodeIfInHeritageClause(node.Parent)
+	}
+	if node.Kind == ast.KindExpressionWithTypeArguments && (ast.IsClassLike(node.Parent.Parent) || node.Parent.Parent.Kind == ast.KindInterfaceDeclaration) {
+		return node.Parent.Parent
+	}
+	return nil
+}
+
+func getContainerNode(node *ast.Node) *ast.Node {
+	// if (isJSDocTypeAlias(node)) {
```
sandersn: JSTypeAliasDeclaration doesn't behave this way anymore, so this code is not needed.

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+	// /**/<var|let| [|n:ame|] ...
+	if node.Kind == ast.KindVarKeyword || node.Kind == ast.KindConstKeyword || node.Kind == ast.KindLetKeyword &&
+		ast.IsVariableDeclarationList(parent) && len(parent.AsVariableDeclarationList().Declarations.Nodes) == 1 {
+		if decl := parent.AsVariableDeclarationList().Declarations.Nodes[0].AsVariableDeclaration(); ast.IsIdentifier(decl.Name()) {
+			return decl.Name()
+		}
+	}
+
+	if node.Kind == ast.KindTypeKeyword {
+		// import /**/type [|name|] from ...;
```
sandersn, line 1162: what is this function called on? could `node` ever be synthetic? asking because I see KindExportAssignment below and I don't know if KindJSExportAssignment is relevant or not.

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		return node.Name()
+	}
+	if forRename {
+		return nil
+	}
+	switch node.Kind {
+	case ast.KindClassDeclaration, ast.KindFunctionDeclaration:
+		// for class and function declarations, use the `default` modifier
+		// when the declaration is unnamed.
+		if node.Modifiers() != nil {
```
sandersn, line 1357: This is currently only called one place, inside a check that the original node is a modifier, and the node in this function is its parent. So this check isn't strictly needed. (But is probably a port, and probably a good idea in case this function is used elsewhere.)

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+func getAdjustedLocationForDeclaration(node *ast.Node, forRename bool, sourceFile *ast.SourceFile) *ast.Node {
+	if node.Name() != nil {
+		return node.Name()
+	}
+	if forRename {
+		return nil
+	}
+	switch node.Kind {
+	case ast.KindClassDeclaration, ast.KindFunctionDeclaration:
+		// for class and function declarations, use the `default` modifier
```
sandersn, line 1355: TIL `export default class { }` is a ClassDeclaration not a ClassExpression

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+	} else if ast.IsEntityName(node) && ast.FindAncestor(node, func(*ast.Node) bool {
+		return node.Kind == ast.KindJSDocNameReference || ast.IsJSDocLinkLike(node) || node.Kind == ast.KindJSDocMemberName
+	}) != nil {
+		return ast.SemanticMeaningAll
+	} else if isTypeReference(node) {
+		return ast.SemanticMeaningType
+	} else if isNamespaceReference(node) {
+		return ast.SemanticMeaningNamespace
+	} else if parent.Kind == ast.KindTypeParameter {
+		// todo jsdoc not implemented
```
sandersn: not needed

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+	if !forRename {
+		// /**/export * from "[|module|]";
+		// export /**/type * from "[|module|]";
+		return node.ModuleSpecifier
+	}
+	return nil
+}
+
+func getMeaningFromLocation(node *ast.Node) ast.SemanticMeaning {
+	node = getAdjustedLocation(node, false /*forRename*/, nil)
```
sandersn, line 1442: I don't know if this needs to handle JS-reparsed nodes, but if so, it should handle JSExportAssignment here and 

	KindJSTypeAliasDeclaration (type)
	KindCommonJSExport (value)
	KindJSImportDeclaration (all)

in getMeaningFromDeclaration

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		return ast.SemanticMeaningType | ast.SemanticMeaningValue
+	} else {
+		return ast.SemanticMeaningValue
+	}
+}
+
+func getMeaningFromDeclaration(node *ast.Node) ast.SemanticMeaning {
+	switch node.Kind {
+	case ast.KindVariableDeclaration:
+		// if ast.IsInJSFile(node) { // !!! && ast.GetJSDocEnumTag(node) {
```
sandersn: not needed

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		// This might be T["name"], which is actually referencing a property and not a type. So allow both meanings.
+		return ast.SemanticMeaningType | ast.SemanticMeaningValue
+	} else {
+		return ast.SemanticMeaningValue
+	}
+}
+
+func getMeaningFromDeclaration(node *ast.Node) ast.SemanticMeaning {
+	switch node.Kind {
+	case ast.KindVariableDeclaration:
```
sandersn:
```suggestion
	case ast.KindVariableDeclaration, ast.KindCommonJSExport:
```

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		// if ast.IsInJSFile(node) { // !!! && ast.GetJSDocEnumTag(node) {
+		// 	return ast.SemanticMeaningAll
+		// }
+		return ast.SemanticMeaningValue
+
+	case ast.KindParameter, ast.KindBindingElement, ast.KindPropertyDeclaration, ast.KindPropertySignature, ast.KindPropertyAssignment, ast.KindShorthandPropertyAssignment, ast.KindMethodDeclaration, ast.KindMethodSignature, ast.KindConstructor, ast.KindGetAccessor, ast.KindSetAccessor, ast.KindFunctionDeclaration, ast.KindFunctionExpression, ast.KindArrowFunction, ast.KindCatchClause, ast.KindJsxAttribute:
+		return ast.SemanticMeaningValue
+
+	case ast.KindTypeParameter, ast.KindInterfaceDeclaration, ast.KindTypeAliasDeclaration, ast.KindTypeLiteral:
```
sandersn:
```suggestion
	case ast.KindTypeParameter, ast.KindInterfaceDeclaration, ast.KindTypeAliasDeclaration, ast.KindJSTypeAliasDeclaration, ast.KindTypeLiteral:
```

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+		return ast.SemanticMeaningType
+
+	case ast.KindJSDocTypedefTag:
```
sandersn: this whole case is replaced with the JSTypeAliasDeclaration case above, and fortunately anonymous typedefs are not supported in Corsa

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+	case ast.KindModuleDeclaration:
+		if ast.IsAmbientModule(node) {
+			return ast.SemanticMeaningNamespace | ast.SemanticMeaningValue
+		} else if ast.GetModuleInstanceState(node) == ast.ModuleInstanceStateInstantiated {
+			return ast.SemanticMeaningNamespace | ast.SemanticMeaningValue
+		} else {
+			return ast.SemanticMeaningNamespace
+		}
+
+	case ast.KindEnumDeclaration, ast.KindNamedImports, ast.KindImportSpecifier, ast.KindImportEqualsDeclaration, ast.KindImportDeclaration, ast.KindExportAssignment, ast.KindExportDeclaration:
```
sandersn:
```suggestion
	case ast.KindEnumDeclaration, ast.KindNamedImports, ast.KindImportSpecifier, ast.KindImportEqualsDeclaration, ast.KindImportDeclaration, ast.KindJSImportDeclaratoin, ast.KindExportAssignment, ast.KindExportDeclaration:
```

internal/ls/utilities.go
```diff
@@ -711,6 +1006,512 @@ func nodeEndsWith(n *ast.Node, expectedLastToken ast.Kind, sourceFile *ast.Sourc
...
+				m |= declarationMeaning
+			}
+		}
+		return m
+	}
+	meaning = iteration(meaning)
+
+	for meaning != lastIterationMeaning {
+		// The result is order-sensitive, for instance if initialMeaning == Namespace, and declarations = [class, instantiated module]
+		// we need to consider both as they initialMeaning intersects with the module in the namespace space, and the module
```
sandersn:
```suggestion
		// we need to consider both as the initialMeaning intersects with the module in the namespace space, and the module
```

internal/ls/utilities.go
```diff
@@ -724,3 +1525,106 @@ func getAllSuperTypeNodes(node *ast.Node) []*ast.TypeNode {
...
+			return core.IfElse(t.Symbol() != nil && t.Symbol().Flags&(ast.SymbolFlagsClass|ast.SymbolFlagsInterface) != 0, t.Symbol(), nil)
+		},
+	)
+	if len(res) == 0 {
+		return nil
+	}
+	return res
+}
+
+/**
```
sandersn: I think the Go format for doc comments is just `//`. Not certain though.

internal/ls/utilities.go
```diff
@@ -724,3 +1525,106 @@ func getAllSuperTypeNodes(node *ast.Node) []*ast.TypeNode {
...
+* Find symbol of the given property-name and add the symbol to the given result array
+* @param symbol a symbol to start searching for the given propertyName
+* @param propertyName a name of property to search for
+* @param result an array of symbol of found property symbols
+* @param previousIterationSymbolsCache a cache of symbol from previous iterations of calling this function to prevent infinite revisiting of the same symbol.
+*                                The value of previousIterationSymbol is undefined when the function is first called.
```
sandersn: these don't appear to match the current parameters

internal/ls/utilities.go
```diff
@@ -724,3 +1525,106 @@ func getAllSuperTypeNodes(node *ast.Node) []*ast.TypeNode {
...
+		return nil
+	}
+
+	lhsType := ch.GetTypeAtLocation(propertyAccessExpression.Expression())
+	if lhsType == nil {
+		return nil
+	}
+
+	res := core.MapNonNil(
+		core.IfElse(lhsType.Flags() != 0, lhsType.Types(), core.IfElse(lhsType.Symbol() == symbol.Parent, nil, []*checker.Type{lhsType})),
```
sandersn: this feels like it does too much work for a pair of IfElses, though it would be much longer to write out as if/elses.

internal/ls/utilities.go
```diff
@@ -724,3 +1525,106 @@ func getAllSuperTypeNodes(node *ast.Node) []*ast.TypeNode {
...
+
+func getPropertySymbolOfObjectBindingPatternWithoutPropertyName(symbol *ast.Symbol, checker *checker.Checker) *ast.Symbol {
+	bindingElement := ast.GetDeclarationOfKind(symbol, ast.KindBindingElement)
+	if bindingElement != nil && isObjectBindingElementWithoutPropertyName(bindingElement) {
+		return getPropertySymbolFromBindingElement(checker, bindingElement)
+	}
+	return nil
+}
+
+func getSourceOfDefaultedAssignment(node *ast.Node) *ast.Node {
```
sandersn: not needed because this is called from unneeded code.
also, defaulted assignments (`window.x = window.x || "thing"`) are not supported in Corsa

sandersn (changes requested):
I haven't read findallreferences.go yet but here are my comments for the rest. I focussed on changes needed for JS. I'll follow up with comments for the big file tomorrow.

internal/checker/checker.go
gabritto: the gift that keeps on giving 😅

jakebailey: For this package, just declare the exported one in `exports.go`.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+	if node.Parent == nil {
+		return nil
+	}
+
+	if !ast.IsDeclaration(node.Parent) && node.Parent.Kind != ast.KindExportAssignment {
+		// Special property assignment in javascript
+		if ast.IsInJSFile(node) {
+			binaryExpression := core.IfElse(node.Parent.Kind == ast.KindBinaryExpression,
+				node.Parent,
+				core.IfElse(ast.IsAccessExpression(node.Parent) && node.Parent.Parent.Kind == ast.KindBinaryExpression && node.Parent.Parent.AsBinaryExpression().Left == node.Parent,
```
sandersn, line 156: this branch *might* not be needed, but I can't tell. better to wait until we have tests to find out.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+func getContextNodeForNodeEntry(node *ast.Node) *ast.Node {
+	if ast.IsDeclaration(node) {
+		return getContextNode(node)
+	}
+
+	if node.Parent == nil {
+		return nil
+	}
+
+	if !ast.IsDeclaration(node.Parent) && node.Parent.Kind != ast.KindExportAssignment {
```
sandersn: I'm not 100% sure this is needed, but:
```suggestion
	if !ast.IsDeclaration(node.Parent) && node.Parent.Kind != ast.KindExportAssignment && node.Parent.Kind != ast.KindJSExportAssignment {
```

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+		propertyName := ast.FindAncestor(node, ast.IsComputedPropertyName)
+		if propertyName != nil {
+			return getContextNode(propertyName.Parent)
+		}
+		return nil
+	}
+
+	if node.Parent.Name() == node || // node is name of declaration, use parent
+		node.Parent.Kind == ast.KindConstructor ||
+		node.Parent.Kind == ast.KindExportAssignment ||
```
sandersn, line 194: JSExportAssignment here too

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+
+	if len(symbol.Declarations) == 0 {
+		return nil
+	}
+
+	declarations := symbol.Declarations
+	// If this is private property or method, the scope is the containing class
+	if symbol.Flags&(ast.SymbolFlagsProperty|ast.SymbolFlagsMethod) != 0 {
+		privateDeclaration := core.Find(declarations, func(d *ast.Node) bool {
+			return checker.HasModifier(d, ast.ModifierFlagsPrivate) || ast.IsPrivateIdentifierClassElementDeclaration(d)
```
sandersn, line 345: aside--if this port didn't skip any Strada jsdoc checks for `@private`, then Corsa's going to be more a lot more accurate for JS, since my next PR is to convert `@private` into ModifierFlagsPrivate. (Nothing for you to do here, just an observation.)

internal/ls/findallreferences.go
sandersn: Later: pretty sure this isn't needed.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+		if (searchSpaceNode.Kind == ast.KindMethodDeclaration || searchSpaceNode.Kind == ast.KindMethodSignature) && ast.IsObjectLiteralMethod(searchSpaceNode) {
+			staticFlag &= searchSpaceNode.ModifierFlags()
+			searchSpaceNode = searchSpaceNode.Parent // re-assign to be the owning object literals
+			break
+		}
+		staticFlag &= searchSpaceNode.ModifierFlags()
+		searchSpaceNode = searchSpaceNode.Parent // re-assign to be the owning class
+		break
+	case ast.KindSourceFile:
+		if ast.IsExternalModule(searchSpaceNode.AsSourceFile()) || isParameterName(thisOrSuperKeyword) {
```
sandersn, line 690: I don't see how `GetThisContainer` could return a SourceFile for a parameter name but also `this` never fails to surprise me.

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+	}
+
+	if parent.Kind == ast.KindExportSpecifier {
+		// !!! not implemented
+		// debug.Assert(referenceLocation.Kind == ast.KindIdentifier || referenceLocation.Kind == ast.KindStringLiteral)
+		// state.getReferencesAtExportSpecifier(referenceLocation /* Identifier | StringLiteral*/, referenceSymbol, parent.AsExportSpecifier(), search, addReferencesHere, false /*alwaysGetReferences*/)
+		return
+	}
+
+	// !!! not implemented
```
sandersn: not needed

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+		// state.addConstructorReferences(referenceLocation, sourceFile, search)
+	case "class":
+		// !!! not implemented
+		// state.addClassStaticThisReferences(referenceLocation, search)
+	}
+
+	// Use the parent symbol if the location is commonjs require syntax on javascript files only.
+	if ast.IsInJSFile(referenceLocation) && referenceLocation.Parent.Kind == ast.KindBindingElement &&
+		ast.IsVariableDeclarationInitializedToRequire(referenceLocation.Parent.Parent.Parent) {
+		// !!! when findAllReferences has been fully implemented, check if the behavior is the same since isVariableDeclarationInitializedToBareOrAccessedRequire has been removed
```
sandersn: only the tests that use a property access on a require should fail, eg `const x = require('mod').x`

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+	)
+	return result
+}
+
+func (state *refState) getRelatedSymbol(search *refSearch, referenceSymbol *ast.Symbol, referenceLocation *ast.Node) (*ast.Symbol, entryKind) {
+	return state.forEachRelatedSymbol(
+		referenceSymbol,
+		referenceLocation,
+		false, /*isForRenamePopulateSearchSymbolSet*/
+		state.options.use != referenceUseRename || !!state.options.providePrefixAndSuffixTextForRename, /*onlyIncludeBindingElementAtReferenceLocation*/
```
sandersn:
```suggestion
		state.options.use != referenceUseRename || state.options.providePrefixAndSuffixTextForRename, /*onlyIncludeBindingElementAtReferenceLocation*/
```

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1493 @@
...
+		// !!! not implemented
+
+		// When renaming at an export specifier, rename the export and not the thing being exported.
+		// state.getReferencesAtExportSpecifier(exportSpecifier.Name(), symbol, exportSpecifier.AsExportSpecifier(), state.createSearch(node, originalSymbol, comingFromUnknown /*comingFrom*/, "", nil), true /*addReferencesHere*/, true /*alwaysGetReferences*/)
+	} else if node != nil && node.Kind == ast.KindDefaultKeyword && symbol.Name == ast.InternalSymbolNameDefault && symbol.Parent != nil {
+		state.addReference(node, symbol, entryKindNone)
+		// !!! not implemented
+		// state.searchForImportsOfExport(node, symbol, &ExportInfo{exportingModuleSymbol: symbol.Parent, exportKind: ExportKindDefault})
+	} else {
+		search := state.createSearch(node, symbol, comingFromUnknown /*comingFrom*/, "", state.populateSearchSymbolSet(symbol, node, options.use == referenceUseRename, !!options.providePrefixAndSuffixTextForRename, !!options.implementations))
```
sandersn:
```suggestion
		search := state.createSearch(node, symbol, comingFromUnknown /*comingFrom*/, "", state.populateSearchSymbolSet(symbol, node, options.use == referenceUseRename, options.providePrefixAndSuffixTextForRename, options.implementations))
```

sandersn (changes requested):
Minor comments on findallreferences.go, nothing substantive.

internal/ls/utilities.go
iisaduan: huh. It's directly copied over from strada, and it appears to be equally wrong there lol

internal/ls/utilities.go
iisaduan: does it also need `ast.KindJsExportAssignment`?

internal/ls/utilities.go
iisaduan: I'm not totally sure, as `getAdjustedLocation` is called at the entry point into a bunch of `findAllReference` uses, each of which could use it differently. If it helps, in Strada, the function doesn't special case `isInJsFile`/`isJsDoc`

iisaduan: maybe the best thing to do is leave a `// todo, check this out later once jsdoc find all ref tests are ported

internal/ls/utilities.go
iisaduan: oops, chanaged the function signature and did not change the comments to match

internal/ast/utilities.go
iisaduan: Should I also just replace `GetAssignmentDeclarationPropertyAccessKind` with `GetAssignmentDeclarationKind`?

internal/ls/findallreferences.go
sandersn: oh, uh: `const f = x=>x` at the top of a file? Arrows don't bind `this`

internal/ast/utilities.go
sandersn: Yes

internal/ls/utilities.go
sandersn: Yes, I missed that.

internal/ls/utilities.go
iisaduan: I'm going to leave the same comment here as left with `getAdjustedLocation`, because I'm not sure about jsdoc

internal/checker/services.go
```diff
@@ -396,6 +396,56 @@ func (c *Checker) GetExportSymbolOfSymbol(symbol *ast.Symbol) *ast.Symbol {
...
+	// node should be ExportSpecifier | Identifier
+	switch node.Kind {
+	case ast.KindExportSpecifier:
+		if node.Parent.Parent.AsExportDeclaration().ModuleSpecifier != nil {
+			return c.getExternalModuleMember(node.Parent.Parent, node, false /*dontResolveAlias*/)
+		}
+		name := node.AsExportSpecifier().PropertyName
+		if name == nil {
+			name = node.Name()
+		}
```
andrewbranch: This is super common, may be worth adding methods on ImportSpecifier and ExportSpecifier

internal/compiler/program.go
```diff
@@ -875,6 +887,96 @@ func (p *Program) GetSourceFiles() []*ast.SourceFile {
...
+		sourceFile.FileName(),
+		p.getModeForTypeReferenceDirectiveInFile(typeRef, sourceFile),
+		nil,
+	)
+}
+
+func (p *Program) getModeForTypeReferenceDirectiveInFile(ref *ast.FileReference, sourceFile *ast.SourceFile) core.ResolutionMode {
+	return ref.ResolutionMode
+	// TODO
+	// !!! || p.getDefaultResolutionModeForFile(sourceFile)
```
andrewbranch: These functions are now in `main`

internal/compiler/program.go
```diff
@@ -875,6 +887,96 @@ func (p *Program) GetSourceFiles() []*ast.SourceFile {
...
+}
+
+/** This should have similar behavior to 'processSourceFile' without diagnostics or mutation. */
+// func (p *Program) GetSourceFileFromReference(referencingFile *ast.SourceFile, ref *ast.FileReference) *ast.SourceFile {
+// 	return p.getSourceFileFromReferenceWorker(tspath.ResolveTripleslashReference(ref.FileName, referencingFile.FileName()), p.GetSourceFile, nil /*fail*/)
+// }
+// !!! todo: check against new `getSourceFileFromReference`
+func (p *Program) getSourceFileFromReferenceWorker(
```
andrewbranch: Yeah, Wesley ported this :(

internal/core/set.go
```diff
@@ -48,6 +48,17 @@ func (s *Set[T]) AddIfAbsent(key T) bool {
 	return true
 }
 
+// Returns true if the key is already in the set. Adds the key and returns false otherwise.
+//
+//	"marks" keys for in "seen" sets in services
+func (s *Set[T]) HasAndAdd(key T) bool {
```
andrewbranch: This is just `!s.AddIfAbsent(key)` 😅

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1486 @@
...
+const (
+	entryKindNone                       entryKind = 0
+	entryKindRange                      entryKind = 1
+	entryKindNode                       entryKind = 2
+	entryKindStringLiteral              entryKind = 3
+	entryKindSearchedLocalFoundProperty entryKind = 4
+	entryKindSearchedPropertyFoundLocal entryKind = 5
+)
+
+type Entry struct {
```
andrewbranch: Does this and its constructor need to be exported?

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1486 @@
...
+			node.Parent.Kind == ast.KindBreakStatement ||
+			node.Parent.Kind == ast.KindContinueStatement {
+			return node.Parent
+		} else if node.Parent.Kind == ast.KindStringLiteral || node.Parent.Kind == ast.KindNoSubstitutionTemplateLiteral {
+			if validImport := tryGetImportFromModuleSpecifier(node); validImport != nil {
+				declOrStatement := ast.FindAncestor(validImport, func(*ast.Node) bool {
+					return ast.IsDeclaration(node) || ast.IsStatement(node) || ast.IsJSDocTag(node)
+				})
+				return core.IfElse(ast.IsDeclaration(declOrStatement),
+					getContextNode(declOrStatement),
```
andrewbranch: Make this a real `if` statement; `getContextNode` does some amount of work

internal/ls/findallreferences.go
```diff
@@ -0,0 +1,1486 @@
...
+	referenceUseReferences referenceUse = 2
+	referenceUseRename     referenceUse = 3
+)
+
+type refOptions struct {
+	findInStrings                       bool
+	findInComments                      bool
+	use                                 referenceUse // other, references, rename
+	implementations                     bool
+	providePrefixAndSuffixTextForRename bool
```
andrewbranch: Can we take this opportunity to rename one of the worst and most confusingly named user preferences we have? This is going to be implemented differently too (rename just returns a generic workspace edit, no explicit prefix and suffix text properties), becoming even more of a misnomer. This is what it maps to:

image

and it defaults to `true` from the client side.

iisaduan: I'm always down for making user preference names less confusing :D

internal/ls/findallreferences.go
iisaduan: renamed to `referenceEntry`

andrewbranch (approved):

internal/checker/utilities.go
```diff
@@ -1202,7 +1202,7 @@ func isInAmbientOrTypeNode(node *ast.Node) bool {
 	}) != nil
 }
 
-func getAncestor(node *ast.Node, kind ast.Kind) *ast.Node {
+func GetAncestor(node *ast.Node, kind ast.Kind) *ast.Node {
```
jakebailey: Should this just be in `ast`? Same for `HasModifier`, etc?

internal/lsp/server.go
```diff
@@ -583,6 +588,18 @@ func (s *Server) handleDefinition(ctx context.Context, req *lsproto.RequestMessa
...
 }
 
+func (s *Server) handleReferences(ctx context.Context, req *lsproto.RequestMessage) error {
+	// findAllReferences
+	params := req.Params.(*lsproto.ReferenceParams)
+	project := s.projectService.EnsureDefaultProjectForURI(params.TextDocument.Uri)
+	languageService, done := project.GetLanguageServiceForRequest(ctx)
+	defer done()
+
+	locations := languageService.ProvideReferences(params)
```
jakebailey, line 606: May want to do the panic recovery here too (same as completions)

jakebailey:
Seems like this fails to compile, probably due to more code getting added in other PRs.

PR #876, opened 2025-05-17
Create preview builds for publishing to npm and VS Code marketplace

jakebailey:
This PR adds a bunch of infrastructure to produce _preview_ npm packages and VS Code extensions.

At a high level, users will be able to `npm install @typescript/native-preview`, then do `npx tsgo` to try out the CLI. The CLI is not called `tsc` to prevent conflicts at this stage.

There will be an extension "TypeScript (Native Preview)" published to the VS Code marketplace to be able to try out the language server implementation.

Eventually, these will all be marked as deprecated and users will be able to use the real deal (whenever we're "done" and move the codebase over).

---

This all requires building the package for a slew of platforms, binary signing, and so on. It's a bit challenging to put together and test out, but I think it should work at this point. I am explicitly _not_ building for every platform Go supports (what esbuild does), but rather starting out building for all platforms VS Code supports. That should be sufficient for now.

TODO:

- [x] Extension
  - [x] The extension needs a README. This will be visible on the marketplace.
  - [x] The extension description needs a look.
  - [x] Need to double check the content root set for the extension, also for the readme.
  - [x] The extension still needs updates to rebrand it to "TypeScript (Native Preview)". I've only changed the name so far.
- [x] Packages
  - [x] The package READMEs deserve a look. They're mostly generated.
  - [x] Export an importable interface that gives the path to the executable, for `tsdk` loading reasons.
- [x] The main repo README has been modified to reflect the new stuff; also needs to be double checked.
- [x] Need to double check the certs (ARM is weird).
- [x] Finish the release pipeline (located elsewhere).

PR #884, opened 2025-05-20
Prepare VS Code extension for marketplace preview

andrewbranch:
This builds on top of #876 to update strings in the extension, allow resolution of the exe from the @typescript/native-preview package, and make minor UX improvements.

### DIFF
--- a/_extension/src/extension.ts
+++ b/_extension/src/extension.ts
@@ -11,15 +11,28 @@ export async function activate(context: vscode.ExtensionContext) {
     const client = new Client(output, traceOutput);
     registerCommands(context, client, output, traceOutput);
 
-    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(event => {
+    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(async event => {
         if (event.affectsConfiguration("typescript.experimental.useTsgo")) {
-            vscode.commands.executeCommand("workbench.action.restartExtensionHost");
+            const selected = await vscode.window.showInformationMessage("TypeScript Native Preview setting has changed. Restart extensions to apply changes.", "Restart Extensions");
+            if (selected) {
+                vscode.commands.executeCommand("workbench.action.restartExtensionHost");
+            }
         }
     }));
 
-    if (context.extensionMode !== vscode.ExtensionMode.Development && !vscode.workspace.getConfiguration("typescript").get<boolean>("experimental.useTsgo")) {
-        output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
-        return;
+    const useTsgo = vscode.workspace.getConfiguration("typescript").get<boolean>("experimental.useTsgo");
+    if (!useTsgo) {
+        if (context.extensionMode === vscode.ExtensionMode.Development) {
+            if (useTsgo === false) {
+                vscode.window.showWarningMessage(
+                    'TypeScript Native Preview is running in development mode. Ignoring "typescript.experimental.useTsgo": false.',
+                );
+            }
+        }
+        else {
+            output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
+            return;
+        }
     }
 
     await client.initialize(context);
--- a/_extension/src/extension.ts
+++ b/_extension/src/extension.ts
@@ -1,242 +1,32 @@
-import * as path from "path";
 import * as vscode from "vscode";
 
-import {
-    LanguageClient,
-    LanguageClientOptions,
-    NotebookDocumentFilter,
-    ServerOptions,
-    TextDocumentFilter,
-    TransportKind,
-} from "vscode-languageclient/node";
-
-let client: LanguageClient;
-let statusBarItem: vscode.StatusBarItem;
-
-const BUILTIN_TS_EXTENSION_ID = "vscode.typescript-language-features";
+import { Client } from "./client";
+import { registerCommands } from "./commands";
+import { setupStatusBar } from "./statusBar";
+import { builtinTSExtensionId } from "./util";
+import { setupVersionStatusItem } from "./versionStatusItem";
 
 export async function activate(context: vscode.ExtensionContext) {
-    const tsExtension = vscode.extensions.getExtension(BUILTIN_TS_EXTENSION_ID);
-    if (tsExtension?.isActive && !vscode.workspace.getConfiguration("typescript").get<boolean>("experimental.useTsgo")) {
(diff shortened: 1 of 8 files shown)


_extension/src/util.ts
```diff
@@ -0,0 +1,91 @@
...
+    const exeName = `tsgo${process.platform === "win32" ? ".exe" : ""}`;
+
+    let exe = config.get<string>("tsdk");
+    if (exe) {
+        if (exe.endsWith("/@typescript/native-preview")) {
+            try {
+                const packagePath = workspaceResolve(exe);
+                const packageJsonPath = vscode.Uri.joinPath(packagePath, "package.json");
+                const packageJson = JSON.parse(await vscode.workspace.fs.readFile(packageJsonPath).then(buffer => buffer.toString()));
+                const getExePath = (await import(vscode.Uri.joinPath(packagePath, "lib", "getExePath.js").toString())).default;
```
jakebailey, line 47: Does this require modifying the package to be importable? IIRC this isn't exported on the package. Does this only work because the emitted code is CJS?

We should probably just formally export it (and maybe drop the default export?).

_extension/src/util.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const builtinTSExtensionId = "vscode.typescript-language-features";
+
+export interface ExeInfo {
+    path: string;
+    version: string;
+}
+
+export function getBuiltinExePath(context: vscode.ExtensionContext): string {
+    return context.asAbsolutePath(path.join("./lib", `tsgo${process.platform === "win32" ? ".exe" : ""}`));
```
jakebailey, line 22: I'd like to be able to move this to `node_modules` so we can just use resolve, but vsce is making it very hard to do that due to https://github.com/microsoft/vscode-vsce/issues/970, so I think we'll have to keep this for now. Thankfully it's in the VSIX so is changeable, unlike the other resolution algorithm.

_extension/src/util.ts
andrewbranch: No, this works because it’s an absolute path, so the package.json is irrelevant.

jakebailey: Huh, I didn't realize that absolute resolution bypassed export rules.

_extension/src/extension.ts
```diff
@@ -1,242 +1,48 @@
...
-                            // But, we're only called on selectors passed above, so just ignore this for now.
-                            throw new Error("Not implemented");
-                        }
-
-                        return true;
-                    }
+    const output = vscode.window.createOutputChannel("typescript-native-preview", "log");
+    const traceOutput = vscode.window.createOutputChannel("typescript-native-preview (LSP)");
+    const client = new Client(output, traceOutput);
+    registerCommands(context, client, output, traceOutput);
```
jakebailey, line 12: Can we defer output window creation until Client actually starts the language server?

For the "show output" command, I think we could just make the command ask the Client for its output window (or undefined).

Probably not required for this PR, though, just noticing it now.

_extension/src/extension.ts
```diff
@@ -1,242 +1,48 @@
...
-    ], {
-        placeHolder: "TypeScript Go Options",
-    });
-
-    if (selected) {
-        if (selected.label.includes("Restart Server")) {
-            await vscode.commands.executeCommand("typescript-go.restart");
+    const useTsgo = vscode.workspace.getConfiguration("typescript").get<boolean>("experimental.useTsgo");
+    if (!useTsgo) {
+        if (context.extensionMode === vscode.ExtensionMode.Development) {
```
jakebailey, line 29: Is this extension mode per-extension? I haven't seen this option before, so, curious.

_extension/src/extension.ts
```diff
@@ -1,242 +1,48 @@
...
+            if (useTsgo === false) {
+                vscode.window.showInformationMessage(
+                    'TypeScript Native Preview is running in development mode. Ignoring "typescript.experimental.useTsgo": false.',
+                );
+            }
         }
-        else if (selected.label.includes("Show TS Server Log")) {
-            await vscode.commands.executeCommand("typescript-go.output.focus");
+        else {
+            output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
```
jakebailey, line 37: Ah, I guess this logging means we couldn't conditionally enable the output, yeah.

_extension/src/extension.ts
andrewbranch: Maybe we can just defer the trace output; I have a message in the main output window saying that the extension is disabled whenever `typescript.experimental.useTsgo` isn’t set.

_extension/src/extension.ts
andrewbranch: Yep. `Development` means it’s running from `--extension-development-path`

_extension/src/util.ts
andrewbranch: Also relative. Only bare specifiers do the complicated thing. The idea being that ESM resolution should generally be predictable and cheap, but package specifier resolution is what it is.

_extension/src/extension.ts
jakebailey: I don't really mind; this seems fine for now.

jakebailey (approved):
LGTM; will defer to @DanielRosenwasser about the `package.json` naming 
and stuff.

Herebyfile.mjs
```diff
@@ -1208,6 +1208,6 @@ export const installExtension = task({
         console.log(pc.yellowBright("\nExtension installed. ") + "To enable this extension, set:\n");
         console.log(pc.whiteBright(`    "typescript.experimental.useTsgo": true\n`));
```
DanielRosenwasser, line 1209: Is this going to be the final setting name?

jakebailey: It's hardcoded in https://github.com/microsoft/vscode/blob/35d9d31a8c96ba3e545a1db67a91dc9876551cc9/extensions/typescript-language-features/src/extension.ts#L37 so we can't really change it immediately

DanielRosenwasser: @mjbvz can we at least change how this looks on the UI so `Tsgo` is not awkwardly Pascal-cased?

image

mjbvz: Mentioned here that extensions may not be able to change it: https://github.com/microsoft/vscode/issues/73374

@rzhao271  Can let the TS team know if there's a way around the auto casing in the settings UI

rzhao271: I see the Tsgo setting in the [package.json](https://github.com/microsoft/vscode/blob/35d9d31a8c96ba3e545a1db67a91dc9876551cc9/extensions/typescript-language-features/package.json#L206) file.

There's no workaround yet. What is the expected display name, "Use TS Go"?

andrewbranch: Worth noting users don’t need to see this setting if they use this
image

PR #885, opened 2025-05-20
Add {}/class-expression expandos

sandersn:
This PR adds support for expando properties on `{}` and class expressions in Javascript. It also adds support for expandos on expandos in Javascript:

```js
var o = {}
o.C = class { // expando on o
  constructor() {
    this.o = {}
    this.o.p = 12 // expando on this.o
  }
}
o.C.x = 13 // expando on o.C
```

It also works with element accesses as well as property accesses.

### Implementation notes
- The PR, unfortunately, brings back some of the complexity that the binder uses to detect these different patterns. But it's still nothing like the checks in Strada.
- I discovered that Strada's `thisParentContainer` tracking was not complete, so it relies on getThisContainer, which isn't safe to use in the binder because not all parent pointers are set. I added a specific ContainerFlags for `this`, and the resulting code should be simpler and safer.
- Checking of `{}` expandos is different because the exports are on the object literal, not the declaring variable. The binding is now consistent between `{}, class, function` expressions, and the checker code to support the old way would be at least as bad.
- References to other properties in the same expando cause a circularity in widening. I put a hack in there to stop it but I decided I didnt like it and left it out for now.
- While adding the code to contextually type expando properties when there's a type annotation for the container, I made the code more like the Strada original.

The binder diff is a lot easier to read with whitespace ignored since I pulled out some code into a separate function.


internal/checker/checker.go
```diff
@@ -16926,6 +16931,11 @@ func (c *Checker) getWidenedTypeWithContext(t *Type, context *WideningContext) *
 		case t.flags&(TypeFlagsAny|TypeFlagsNullable) != 0:
 			result = c.anyType
 		case isObjectLiteralType(t):
+			// expando object literal types do not widen to prevent circular checking (but should, ideally)
```
sandersn: this is a TODO that I'll get back to later

internal/binder/binder.go
```diff
@@ -1209,22 +1240,35 @@ func (b *Binder) bindTypeParameter(node *ast.Node) {
...
+	}
+	return nil
+}
+
 func (b *Binder) lookupName(name string, container *ast.Node) *ast.Symbol {
-	localsContainer := container.LocalsContainerData()
-	if localsContainer != nil {
-		local := localsContainer.Locals[name]
-		if local != nil {
+	if localsContainer := container.LocalsContainerData(); localsContainer != nil {
```
sandersn, line 1269: Just a style improvement--at least I hope it's an improvement.

internal/binder/binder.go
```diff
@@ -1010,74 +1011,104 @@ func addLateBoundAssignmentDeclarationToSymbol(node *ast.Node, symbol *ast.Symbo
...
+	if symbol == nil || symbol.ValueDeclaration == nil {
+		return nil
+	}
+	declaration := symbol.ValueDeclaration
+	// For an assignment 'fn.xxx = ...', where 'fn' is a previously declared function or a previously
+	// declared const variable initialized with a function expression or arrow function, we add expando
+	// property declarations to the function's symbol.
+	// This also applies to class expressions and empty object literals.
+	switch {
+	case ast.IsFunctionDeclaration(declaration) || ast.IsInJSFile(declaration) && ast.IsClassDeclaration(declaration):
```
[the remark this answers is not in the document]
sandersn: No, this function only takes symbols from top-level statements. The BinaryExpression case follows top-level BinaryExpressions to their initialisers, which could be class expressions.

internal/binder/binder.go
```diff
@@ -1010,74 +1011,104 @@ func addLateBoundAssignmentDeclarationToSymbol(node *ast.Node, symbol *ast.Symbo
...
+		}
+	case ast.IsBinaryExpression(declaration) && ast.IsInJSFile(declaration):
+		initializer := declaration.AsBinaryExpression().Right
+		if isExpandoInitializer(initializer) {
+			return initializer.Symbol()
 		}
 	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
sandersn: The functions this function calls are as easy to read as the english comment would be, and won't fall out of date. A comment is not helpful here.

internal/checker/checker.go
sandersn: I decided the hack was not good enough so I left in the circular type checking errors for now.

testdata/baselines/reference/submodule/conformance/jsDeclarationsClassStatic2.symbols.diff
```diff
@@ -27,7 +27,7 @@
 
  Bar.foo = "foo";
 ->Bar.foo : Symbol(Foo.foo, Decl(Foo.js, 3, 32))
-+>Bar.foo : Symbol(foo, Decl(Foo.js, 0, 12))
++>Bar.foo : Symbol(foo, Decl(Foo.js, 3, 32))
```
jakebailey, line 30: I do wonder how much work it'd be to fix the symbol baseline namings to make these diffs go away.

sandersn: I thought that it would be improved with Wesley's declaration emit PR, but upon thinking about it, that might only help type display and not symbol display. I'll merge from main tomorrow and see if there's a difference.

sandersn: Nope. No change. I'll investigate this as part of burning down JS test diffs.

internal/binder/binder.go
```diff
@@ -49,7 +50,7 @@ type Binder struct {
 
 	parent                 *ast.Node
 	container              *ast.Node
-	thisParentContainer    *ast.Node
+	thisContainer          *ast.Node
```
jakebailey, line 53: Hopefully this doesn't come back to bite us in porting code. (Pun intended)

testdata/baselines/reference/submodule/compiler/jsEnumCrossFileExport.symbols.diff
```diff
@@ -47,7 +55,9 @@
 ->Other.Cls : Symbol(Other.Cls, Decl(index.js, 0, 15))
 ->Other : Symbol(Other, Decl(index.js, 0, 3), Decl(index.js, 0, 15))
 ->Cls : Symbol(Other.Cls, Decl(index.js, 0, 15))
++>Other.Cls : Symbol(Cls, Decl(index.js, 0, 15))
```
jakebailey, line 58: I feel like so many of the leftover diffs are just this symbol naming problem

testdata/baselines/reference/submoduleAccepted/compiler/classFieldSuperAccessibleJs1.errors.txt.diff
```diff
@@ -5,27 +5,18 @@
...
+-      static blah1 = 123;
+-    }
+-    C.blah2 = 456;
+-    
+-    class D extends C {
+-      static {
+-        console.log(super.blah1);
+-        console.log(super.blah2);
+-                          ~~~~~
 -!!! error TS2565: Property 'blah2' is used before being assigned.
```
jakebailey, line 18: This is an interesting one; is the reparser creating nodes in a fashion that means that we no longer detect use-before-def? E.g. if I declare a prop on a class, it's technically "declared" and wouldn't error.

jakebailey (approved):

internal/binder/binder.go
sandersn: it's only used in expando binding, and that's been entirely rewritten, so (1) it's guaranteed to have all-new bugs anyway (2) I have the context loaded in my head, so it'll be easier to fix.

testdata/baselines/reference/submodule/compiler/jsEnumCrossFileExport.symbols.diff
sandersn: yepppp. I need to look at it to see if it's jsdoc's fault or our symbol display.

testdata/baselines/reference/submoduleAccepted/compiler/classFieldSuperAccessibleJs1.errors.txt.diff
sandersn: I think there was special-case code for class expandos before that has changed just enough to not trigger. I'll check on it during my burndown of diffs.

PR #886, opened 2025-05-20
Unconditionally set dirtyFilePath in markAsDirty

andrewbranch:
Fixes a bug caught by @gabritto in https://github.com/microsoft/typescript-go/pull/879/files/a30a84c490a5d49555a8f46b1516836b75c560e8#r2098483287.

I had to make the other changes in order to make the test fail; there was an unnecessary double update on config file changes that was clearing out the dirty state on the second update.

### DIFF
--- a/internal/project/service.go
+++ b/internal/project/service.go
@@ -305,11 +305,11 @@ func (s *Service) onConfigFileChanged(project *Project, changeKind lsproto.FileC
 		project.deferredClose = true
 	}
 
-	s.delayUpdateProjectGraph(project)
 	if !project.deferredClose {
 		project.pendingReload = PendingReloadFull
 		project.markAsDirty()
 	}
+	project.updateIfDirty()
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


gabritto (approved):

internal/project/service.go
```diff
@@ -305,11 +305,11 @@ func (s *Service) onConfigFileChanged(project *Project, changeKind lsproto.FileC
 		project.deferredClose = true
 	}
 
-	s.delayUpdateProjectGraph(project)
 	if !project.deferredClose {
 		project.pendingReload = PendingReloadFull
 		project.markAsDirty()
 	}
+	project.updateIfDirty()
```
sheetalkamat, line 312: Seems to me this shouldnt be called here and should either send notification to client that things have changed and/or update on next request ?

internal/project/service.go
```diff
@@ -402,25 +402,13 @@ func (s *Service) OnDiscoveredSymlink(info *ScriptInfo) {
...
-}
-
-func (s *Service) delayUpdateProjectGraph(project *Project) {
-	if project.deferredClose {
-		return
-	}
-	project.markAsDirty()
-	if project.kind == KindAutoImportProvider || project.kind == KindAuxiliary {
-		return
+		project.updateGraph()
```
sheetalkamat, line 410: same?

internal/project/service.go
andrewbranch: I think we’ll need this because we’ll have to do push diagnostics for config file errors, which aren’t hooked up yet, but I think we could do that for now. I think I’d like to do that in a separate PR though, as this one is just trying to prove that there was a bug and fix it. We’re trying to get things stable for an npm release.

jakebailey (approved):

PR #887, opened 2025-05-20
Copy default compiler options for inferred projects from VS Code client

andrewbranch:
Taken from a TSServer config message, replacing `allowSyntheticDefaultImports` with `esModuleInterop`

### DIFF
--- a/internal/project/service.go
+++ b/internal/project/service.go
@@ -716,9 +716,19 @@ func (s *Service) getDefaultProjectForScript(scriptInfo *ScriptInfo) *Project {
 }
 
 func (s *Service) createInferredProject(currentDirectory string, projectRootPath tspath.Path) *Project {
-	// !!!
 	compilerOptions := core.CompilerOptions{
-		AllowJs: core.TSTrue,
+		AllowJs:                    core.TSTrue,
+		Module:                     core.ModuleKindESNext,
+		ModuleResolution:           core.ModuleResolutionKindBundler,
+		Target:                     core.ScriptTargetES2022,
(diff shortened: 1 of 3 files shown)


internal/project/service.go
```diff
@@ -716,9 +716,19 @@ func (s *Service) getDefaultProjectForScript(scriptInfo *ScriptInfo) *Project {
 }
 
 func (s *Service) createInferredProject(currentDirectory string, projectRootPath tspath.Path) *Project {
-	// !!!
 	compilerOptions := core.CompilerOptions{
```
sheetalkamat, line 719: is it possible for the extension to kind of use "SetCompilerOptionsForProjectRoot" - think tsserver uses that and there was some need of this in past though i am not aware of what as i wasnt around the time this was designed.

jakebailey: Personally I think we should try and avoid doing that for as long as possible; we're owning the LS now, so I think it makes sense for us to try and set this to what we think is a good idea. I think we'll need an extension point for sure, but I don't think it's needed until maybe VS or some other editor.

jakebailey:
Hm, tests do not like this 😄

internal/project/service.go
andrewbranch: We will eventually need the ability for the client to set certain options (e.g. `strictNullChecks` is exposed as a VS Code option), but we should also encode good defaults ourself. Empty options are really bad, and I’m guessing there are existing third party TS Server clients that don’t use `setCompilerOptionsForInferredProjects` and have really bad default behavior.

PR #893, opened 2025-05-21
Gabritto/stringcompletion

gabritto:


internal/ls/string_completions.go
```diff
@@ -0,0 +1,548 @@
...
+	completion stringLiteralCompletions,
+	contextToken *ast.StringLiteralLike,
+	file *ast.SourceFile,
+	position int,
+	program *compiler.Program,
+	options *core.CompilerOptions,
+	preferences *UserPreferences,
+	clientOptions *lsproto.CompletionClientCapabilities,
+) *lsproto.CompletionList {
+	if completion == nil {
```
gabritto, line 90: This check is insufficient because `completion` might still be an interface-wrapped `nil`. I'll refactor to avoid this problem.

jakebailey (approved):
There's a lot of uncovered code (based on the codecov extension), are you just waiting for fourslash to have more tests?

I skimmed it and it seems fine but, checking if there are quick things we could throw in to get more coverage.

gabritto:
> There's a lot of uncovered code (based on the codecov extension), are you just waiting for fourslash to have more tests?
> 
> I skimmed it and it seems fine but, checking if there are quick things we could throw in to get more coverage.

Yes, basically. It's a bit time consuming to figure out what tests to port (e.g. what tests cover what, what tests don't depend on yet unported code, etc), write them as unit tests, etc, and the idea is we'll have fourslash tests sometime soon, so I didn't want to duplicate that work. But I could be convinced to go and add a few more tests. Do you think it's worth it?

jakebailey:
I think if it's easy, it might be; I just notice stuff like:

image

Where it seems like one or two tests might add a good bit of coverage.

jakebailey (approved):

PR #897, opened 2025-05-21
[LSP] Fix data races on dirty state and script info version

andrewbranch:
Changes `project` to operate with a single mutex, and avoids reading `scriptInfo.Version()` multiple times during parsing.

### DIFF
--- a/internal/project/documentregistry.go
+++ b/internal/project/documentregistry.go
@@ -89,12 +89,14 @@ func (r *DocumentRegistry) getDocumentWorker(
 	key registryKey,
 ) *ast.SourceFile {
 	scriptTarget := core.IfElse(scriptInfo.scriptKind == core.ScriptKindJSON, core.ScriptTargetJSON, compilerOptions.GetEmitScriptTarget())
+	scriptInfoVersion := scriptInfo.Version()
+	scriptInfoText := scriptInfo.Text()
 	if entry, ok := r.documents.Load(key); ok {
 		// We have an entry for this file. However, it may be for a different version of
 		// the script snapshot. If so, update it appropriately.
-		if entry.sourceFile.Version != scriptInfo.Version() {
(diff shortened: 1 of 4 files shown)


internal/project/project.go
```diff
@@ -385,51 +382,37 @@ func (p *Project) markFileAsDirty(path tspath.Path) {
...
-// updateIfDirty returns true if the project was updated.
-func (p *Project) updateIfDirty() bool {
-	// !!! p.invalidateResolutionsOfFailedLookupLocations()
-	return p.dirty && p.updateGraph()
-}
-
-func (p *Project) onFileAddedOrRemoved(isSymlink bool) {
-	p.dirtyStateMu.Lock()
-	defer p.dirtyStateMu.Unlock()
-	p.hasAddedOrRemovedFiles = true
```
andrewbranch, line 406: These weren’t yet used for anything but logging, and they complicate the locking because they are sometimes, but not always, called during program construction when the mutex is already locked.

internal/project/documentregistry.go
```diff
@@ -89,12 +89,14 @@ func (r *DocumentRegistry) getDocumentWorker(
 	key registryKey,
 ) *ast.SourceFile {
 	scriptTarget := core.IfElse(scriptInfo.scriptKind == core.ScriptKindJSON, core.ScriptTargetJSON, compilerOptions.GetEmitScriptTarget())
+	scriptInfoVersion := scriptInfo.Version()
+	scriptInfoText := scriptInfo.Text()
```
andrewbranch, line 93: Technically this should be one atomic access

jakebailey: Do you want to address this now or later?

jakebailey (approved):
I couldn't reproduce the issue before, but I tested and it still works for me, so 😄

ahejlsberg (approved):
I can confirm that this fixes the issue on Windows. 🎉

internal/project/documentregistry.go
andrewbranch: I just tried protecting all the ScriptInfo fields with a mutex and accessing correlated fields within the same lock, but that had a ~20% penalty on project initialization

PR #900, opened 2025-05-21
Add overload tag

sandersn:
`@overload` emits a synthetic bodyless function declaration, which then behaves just like a real overload.

To make errors behave nicely, I had to omit some position checks; those errors can't apply to `@overload` tags anyway. I also removed one `@overload`-specific error and went back to the standard one.

I ran into the same problem with `@template` (type parameters) as I saw with `@typedef`. The fix requires setting Host to a parent synthetic node on nested synthetic nodes, because name resolution always prefers Host for those nodes. I created the signature first, with Parameters and Type nil, so that I could use `makeNewType` to set the Host in the usual way when creating parameters and return type. Then I set the Parameters and Type afterward.

I applied that same fix to `@typedef` but it still didn't instantiate the type parameters correctly, so I'm going to fix it in a separate PR.

There is a good deal of churn because I also removed JSDocPropertyTag as a separate type -- it's now a JSDocParameterTag with a different kind, just like JSExportAssignment and other reparsed nodes.


testdata/baselines/reference/submoduleAccepted/conformance/jsdocParseStarEquals.errors.txt.diff
```diff
@@ -3,19 +3,20 @@
 @@= skipped -0, +-1 lines =@@
 -<no content>
 @@= skipped --1, +1 lines =@@
++a.js(1,5): error TS1047: A rest parameter cannot be optional.
```
weswigham, line 5: Is this an intentional change in behavior? Previously a postfix `=` meant optional-if-able-but-`| undefined`-otherwise.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileMethodOverloads3.errors.txt.diff
```diff
@@ -1,35 +1,26 @@
...
+      * @overload
+         ~~~~~~~~
 -!!! error TS7012: This overload implicitly returns the type 'any' because it lacks a return type annotation.
--     * @param {number} x
--     */
--    
--    /**
--     * @overload
--        ~~~~~~~~
++!!! error TS7010: 'id', which lacks return-type annotation, implicitly has an 'any' return type.
```
weswigham: This is just the wrong error message, no? JSDoc overloads are supposed to use the version without a declaration name because they're nameless in the source, right? Even if the synthetic one we make now copies the host function's name. As-is, this error is kinda ambiguous, since nowhere does `id` appear in the overload descriptor.

internal/ast/ast.go
```diff
@@ -9113,93 +9105,46 @@ func (node *JSDocTemplateTag) Clone(f NodeFactoryCoercible) *Node {
...
 	name           *EntityName
 	IsBracketed    bool
 	TypeExpression *TypeNode
 	IsNameFirst    bool
 }
 
-func (f *NodeFactory) NewJSDocPropertyTag(tagName *IdentifierNode, name *EntityName, isBracketed bool, typeExpression *TypeNode, isNameFirst bool, comment *NodeList) *Node {
-	data := &JSDocPropertyTag{}
+func (f *NodeFactory) newJSDocParameterOrPropertyTag(kind Kind, tagName *IdentifierNode, name *EntityName, isBracketed bool, typeExpression *TypeNode, isNameFirst bool, comment *NodeList) *Node {
+	data := &JSDocParameterTag{}
```
weswigham: I feel like we should probably rename the base struct type here to `JSDocParameterOrPropertyTag` and alias `type JSDocParameterTag = JSDocParameterOrPropertyTag` - likewise, the `As` cast function should change to refer to the `Or` type, but it should be aliased to the subtypes for per-use clarity. As-is, something like
```
switch (node.Kind) {
  case ast.KindJSDocParamterTag, ast.KindJSDocPropertyTag:
    node.AsJSDocParameterTag() /// ???
    
}
```
just looks super sus, even though it (currently) works because of the underlying shapes - the aliased types and methods should help with that. Should also help if we ever get around to paying off the massive debt all this type erasure is incurring and start statically checking unions again...

sandersn: sigh...you're probably right. I hate that `@param` and `@property` are the same thing except that they appear in mutually exclusive contexts. Feels bad to make them a single kind like true synonyms of jsdoc, `@augments` and `@extends` for example.

internal/parser/reparser.go
```diff
@@ -95,10 +88,48 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
...
+						parameter.Loc = jsparam.Loc
+						parameter.Flags = p.contextFlags | ast.NodeFlagsReparsed
+						parameters = append(parameters, parameter)
+					}
+
+					if jsSignature.Type != nil {
+						signature.FunctionLikeData().Type = p.makeNewType(jsSignature.Type.AsJSDocReturnTag().TypeExpression, signature)
+					}
+					signature.FunctionLikeData().Parameters = p.newNodeList(jsSignature.Parameters.Loc, parameters)
+					signature.Loc = tag.AsJSDocOverloadTag().TagName.Loc
```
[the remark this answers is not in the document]
sandersn: This is how Strada reports errors. The entire tag span is too large, especially when printed on the command line.

internal/binder/binder.go
```diff
@@ -2874,9 +2874,15 @@ func GetErrorRangeForNode(sourceFile *ast.SourceFile, node *ast.Node) core.TextR
 		}
 		return scanner.GetRangeOfTokenAtPosition(sourceFile, pos)
 	// This list is a work in progress. Add missing node kinds to improve their error spans
+	case ast.KindFunctionDeclaration, ast.KindMethodDeclaration:
```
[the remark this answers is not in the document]
sandersn: no, this is a bad idea

testdata/baselines/reference/submoduleAccepted/conformance/jsdocParseStarEquals.errors.txt.diff
sandersn: Yes, this is an intentional simplification. `=` always means both (1) addition of syntactic `?` and (2) `getOptionalType` in `getTypeFromTypeNode`. I don't plan to port the code that drops (1) where it's illegal.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileMethodOverloads3.errors.txt.diff
sandersn: I disagree and collapsed the two intentionally, but if you feel strongly I can change it back.

The reason I like having the name in the error message is that the function name appears a few lines after the tag in the source text. When I read the diff, I feel like the Strada error message withholds useful information on a technicality.

weswigham: As-is, to me, the error feels like it's talking about the host declaration rather than the underlined one, and the span's in the wrong place.

weswigham: If you wanted to compromise, the message `This overload of 'id' lacks a return-type annotation and implicitly has an 'any' return type.` would be more apt, I think.

sandersn: It's easier and more consistent with Strada to use the old message. The messages are still generated from the Typescript repo last I checked, so I'd have toc hange it there first.

testdata/baselines/reference/submoduleAccepted/conformance/templateInsideCallback.errors.txt.diff
```diff
@@ -64,12 +66,14 @@
...
-@@= skipped -9, +7 lines =@@
+     
      /**
       * @overload
-      * @template T
--        ~~~~~~~~
+-     * @template T
+         ~~~~~~~~
 -!!! error TS8039: A JSDoc '@template' tag may not follow a '@typedef', '@callback', or '@overload' tag
++!!! error TS7012: This overload implicitly returns the type 'any' because it lacks a return type annotation.
```
sandersn, line 75: Previously the parse error from nested type literal parsing blocked this error, but I haven't ported that yet.

sandersn:
https://github.com/microsoft/typescript-go/issues/1042

apendua:
@sandersn It's really great to see how `@overload` is taking shape in Corsa. I've just done a little bit of search in the issues history and I appreciate how much work the team has to put into this single feature alone. When I first added it a couple years ago, I never imagined that it would generate so much maintenance burden. Thank you!

Please let me know if there's anything I can do to support this and possibly other efforts.

PR #927, opened 2025-05-23
Add the 5 modifier jsdoc tags

sandersn:
`@public @private @protected @readonly @override`

Three notes:
- I did not add support for BinaryExpressions yet because they don't have a Modifiers list. I think the right thing is to add it (along with a Type node), but I want to bring it up at a design meeting first.
- I dropped the error messages with tag-specific wording for `@override`. This seems unnecessary to me -- people who use JS as TS are aware that's what they're doing.
- I added a setModifiers method on nodeData and exposed it via a public AsMutable method on Node. I don't know if that's exactly the right way
  to do it.


internal/parser/reparser.go
```diff
@@ -105,39 +105,43 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
 			}
 			switch tag.Kind {
 			case ast.KindJSDocTypeTag:
-				if parent.Kind == ast.KindVariableStatement && parent.AsVariableStatement().DeclarationList != nil {
-					for _, declaration := range parent.AsVariableStatement().DeclarationList.AsVariableDeclarationList().Declarations.Nodes {
-						if declaration.AsVariableDeclaration().Type == nil {
-							declaration.AsVariableDeclaration().Type = p.makeNewType(tag.AsJSDocTypeTag().TypeExpression, declaration)
-							break
+				switch parent.Kind {
```
sandersn: this is a drive-by reformat. As usual, best reviewed ignoring whitespace.

internal/parser/reparser.go
```diff
@@ -178,6 +182,39 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
 						fun.FunctionLikeData().Type = p.makeNewType(tag.AsJSDocReturnTag().TypeExpression, fun)
 					}
 				}
+			case ast.KindJSDocReadonlyTag, ast.KindJSDocPrivateTag, ast.KindJSDocPublicTag, ast.KindJSDocProtectedTag, ast.KindJSDocOverrideTag:
+				switch parent.Kind {
+				case ast.KindPropertyDeclaration, ast.KindMethodDeclaration, ast.KindGetAccessor, ast.KindSetAccessor:
+					// !!! BinaryExpression (this.p assignments)
```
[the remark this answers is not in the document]
sandersn: I can't implement handling until I decide how to store modifiers for `this.p=`. Waiting on a decision at a design meeting.

testdata/baselines/reference/submodule/conformance/jsdocReadonly.types
```diff
@@ -51,8 +51,8 @@ var l = new LOL()
 
 l.y = 12
 >l.y = 12 : 12
->l.y : number
+>l.y : any
 >l : LOL
->y : number
+>y : any
 >12 : 12
```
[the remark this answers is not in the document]
sandersn: No, it's because `y` is readonly now, so assignments to it are not allowed.

internal/parser/reparser.go
```diff
@@ -178,6 +182,39 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
...
+					var loc core.TextRange
+					if parent.Modifiers() == nil {
+						nodes = p.nodeSlicePool.NewSlice(1)
+						nodes[0] = modifier
+						loc = tag.Loc
+					} else {
+						nodes = append(parent.Modifiers().Nodes, modifier)
+						loc = parent.Modifiers().Loc
+					}
+					parent.AsMutable().SetModifiers(p.newModifierList(loc, nodes))
```
jakebailey: Does this ever happen on _any_ parent kind, or is there a specific subset of nodes that do this? I'm not entirely certain how I feel about the `AsMutable` pattern, so curious if it's just shorthand for a couple of types.

sandersn: It is the 5 node kinds you see here (including BinaryExpression), unless I missed 1 or 2. Would you prefer a specific SetModifiers on those types?

jakebailey: In a way I think it'd be fine to just have a top level ast package function with a switch case in it; just wasn't sure if it was that simple or if the method dispatch here was simplifying things.

sandersn: It is the 5 you see here: class members and expando class members. Based on experience, I predict were going to need this to generalise someday, but definitely not today.

jakebailey (approved):

PR #940, opened 2025-05-26
Non-circular type resolution for expando assignment declarations

### LINKED ISSUE #921: `tsgo` typechecking error in tsx code that passes when using `tsc`
I'm trying out tsgo on my large work codebase and found it errors on a pattern we use very often that does pass `tsc`:

```ts
import type { ComponentType, JSX } from "react"

export type Page<P = NonNullable<unknown>> = ComponentType<P> & {
    getLayout?: (component: JSX.Element) => JSX.Element
}

export const FooPage: Page = () => {
    return (
        <div>
            <p>Foo</p>
        </div>
    )
}

// `tsgo` errors here: "'getLayout' implicitly has type 'any' because it does not have a type annotation
//                                  and is referenced directly or indirectly in its own initializer."
FooPage.getLayout = () => {
    return <></>
}
```

(_And on importing vite's `defineConfig`, but I see someone else beat me to that in #904._)

I was able to reproduce my error in a much more barebones project: https://github.com/justinsmid/tsgo-test
I'm sure the `tsconfig.json` could be simplified, and I'm not sure if it's related to any setting in there, but I wanted to keep the config as close to that of my work repo as I could.

Running `pnpm typecheck` (which uses `tsc`) passes, but `pnpm typecheck-go` does not:

```
> tsgo --noEmit

src/components/Foo.tsx:15:1 - error TS7022: 'getLayout' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

15 FooPage.getLayout = () => {
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~
16     return <></>
   ~~~~~~~~~~~~~~~~
17 }
   ~
vite.config.ts:1:10 - error TS2305: Module '"[REDACTED]/tsgo-test/node_modules/.pnpm/vite@6.3.5/node_modules/vite/index"' has no exported member 'defineConfig'.

1 import { defineConfig } from 'vite';
           ~~~~~~~~~~~~

Found 2 errors in 2 files.

Errors  Files
     1  src/components/Foo.tsx:15
     1  vite.config.ts:1

 ELIFECYCLE  Command failed with exit code 2.
```

ahejlsberg:
Fixes #921.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -27686,7 +27686,10 @@ func (c *Checker) getContextualTypeForBinaryOperand(node *ast.Node, contextFlags
 	case ast.KindEqualsToken, ast.KindAmpersandAmpersandEqualsToken, ast.KindBarBarEqualsToken, ast.KindQuestionQuestionEqualsToken:
 		// In an assignment expression, the right operand is contextually typed by the type of the left operand
 		// unless it's an assignment declaration.
-		if node == binary.Right && !c.isReferenceToModuleExports(binary.Left) && (binary.Symbol == nil || c.canGetContextualTypeForAssignmentDeclaration(node.Parent)) {
+		if node == binary.Right && !c.isReferenceToModuleExports(binary.Left) {
+			if binary.Symbol != nil {
+				return c.getContextualTypeForAssignmentDeclaration(node.Parent)
+			}
 			return c.getTypeOfExpression(binary.Left)
 		}
 	case ast.KindBarBarToken, ast.KindQuestionQuestionToken:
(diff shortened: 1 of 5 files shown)


ahejlsberg:
@jakebailey I didn't add a regression test because it seems we have no way to reference React from tests in `testdata/tests` (we currently have no `.tsx` files in there). In the old tests we have a `lib/react.d.ts`, do we need something similar?

jakebailey:
That's all set up and should "just work" even in the non-submodules test with just a reference directive:

```
/// <reference path="/.lib/react16.d.ts" />
```

The code we have for it isn't specific to where the test lives.

(#730, #758)

ahejlsberg:
@jakebailey Hmm, something is not happy in the test runner when I add that directive.

jakebailey:
Ohh, the test runner doesn't skip the test when the submodule isn't cloned; should be easy to fix.

jakebailey:
Applying this patch should do it:

```diff
diff --git a/internal/testutil/harnessutil/harnessutil.go b/internal/testutil/harnessutil/harnessutil.go
index 3f4474789..9c02222fb 100644
--- a/internal/testutil/harnessutil/harnessutil.go
+++ b/internal/testutil/harnessutil/harnessutil.go
@@ -146,6 +146,10 @@ func CompileFilesEx(
 		}
 	}
 
+	if includeLibDir {
+		repo.SkipIfNoTypeScriptSubmodule(t)
+	}
+
 	// !!!
 	// ts.assign(options, ts.convertToOptionsWithAbsolutePaths(options, path => ts.getNormalizedAbsolutePath(path, currentDirectory)));
 	if compilerOptions.OutDir != "" {
```

jakebailey (approved):

PR #947, opened 2025-05-27
refactor: use slices.Contains and slices.ContainsFunc to simplify code

tongjicoder:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/internal/checker/mapper.go
+++ b/internal/checker/mapper.go
@@ -1,6 +1,10 @@
 package checker
 
-import "github.com/microsoft/typescript-go/internal/core"
+import (
+	"slices"
+
+	"github.com/microsoft/typescript-go/internal/core"
+)
 
 // TypeMapperKind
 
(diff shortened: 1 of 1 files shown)


tongjicoder:

internal/ast/ast.go
```diff
@@ -22,12 +23,7 @@ func visit(v Visitor, node *Node) bool {
 }
 
 func visitNodes(v Visitor, nodes []*Node) bool {
-	for _, node := range nodes {
-		if v(node) {
-			return true
-		}
-	}
-	return false
+	return slices.ContainsFunc(nodes, v)
```
jakebailey: This is not really an improvement; while this function has the same form as ContainsFunc, it's actually doing something different than "finding" an element in a slice, so I would not change it. The same can be said for many of these refactors.

jakebailey (changes requested):

internal/ast/ast.go
tongjicoder: Thank you for your review and suggestions. I have reverted the changes related to slices.ContainsFunc.

jakebailey:
I'm going to close this; I think we're going to apply these using the new `modernize` tool by the Go team in bulk.

I'll also note that you sent this PR from your `main` branch; that is likely to cause problems if you want to send code to this repo, so I would suggest resetting your fork and sending future code via a different branch.

PR #953, opened 2025-05-27
Port and use `GetTypeOfSymbolAtLocation` in quick info

### LINKED ISSUE #912: Bug Report: Type narrowing not working as expected
**Expected Behavior:**

When using type guards like `Array.isArray` or truthiness checks in an `if` block, the type of the variable within that block should be narrowed down to the more specific type.

For `Array.isArray(foo)`:
```typescript
const foo: string | string[] = 'foo'; // or ['foo']
if (Array.isArray(foo)) {
    // foo should be inferred as string[]
    foo; 
}
```

For truthiness check on `bar`:
```typescript
const bar: string | undefined = 'bar'; // or undefined
if (bar) {
    // bar should be inferred as string
    bar;
}
```

**Actual Behavior:**

The type of the variable inside the `if` block is not being narrowed correctly.

For `Array.isArray(foo)`:
```typescript
const foo: string | string[] = 'foo'; // or string[]
if (Array.isArray(foo)) {
    // foo is inferred as string | string[]
    foo; 
}
```
The type of `foo` remains `string | string[]` inside the `if` block, even after the `Array.isArray(foo)` check.

For truthiness check on `bar`:
```typescript
const bar: string | undefined = 'bar'; // or undefined
if (bar) {
    // bar is inferred as string | undefined
    bar.charAt(0); // Error: Object is possibly 'undefined'.
}
```
The type of `bar` remains `string | undefined` inside the `if` block, even after the truthiness check. This can lead to unnecessary errors or require explicit type assertions.

Image

Image

**Additional Context:**

@typescript/native-preview: "7.0.0-dev.20250522.2"
TypeScript (Native Preview): 0.20250522.2

ahejlsberg:
Fixes #912.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -15170,8 +15170,39 @@ func (c *Checker) getWriteTypeOfSymbol(symbol *ast.Symbol) *Type {
 }
 
 func (c *Checker) GetTypeOfSymbolAtLocation(symbol *ast.Symbol, location *ast.Node) *Type {
-	// !!!
-	return c.getTypeOfSymbol(symbol)
+	symbol = c.getExportSymbolOfValueSymbolIfExported(symbol)
+	// If we have an identifier or a property access at the given location, if the location is
+	// an dotted name expression, and if the location is not an assignment target, obtain the type
+	// of the expression (which will reflect control flow analysis). If the expression indeed
+	// resolved to the given symbol, return the narrowed type.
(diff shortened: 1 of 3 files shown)


jakebailey (approved):

jakebailey:
Hm, seems like this must be crashing in the API tests?

jakebailey (approved):

PR #955, opened 2025-05-27
Expose bind diags in LSP, show dead/deprecated code

### LINKED ISSUE #941: tsgo does not detect unreachable code
#### Reproduction Steps

1. Follow the setup instructions from this blog post to enable `tsgo` in VSCode(if you haven't already):  
   https://devblogs.microsoft.com/typescript/announcing-typescript-native-previews/

2. Create a `.ts` file and add the following code:

   ```ts
   function doSomethingAwesome() {
     throw new Error("we messed up");
     return [];
   }

#### Expected Behavior
The return statement should be reported as unreachable when it follows a throw

#### Actual Behavior
No unreachable code warning is shown.


#### Environment 

- **@typescript/native-preview version:** `7.0.0-dev.20250525.1`
- **Node.js version:** `v20.18.2`
- **OS:** Ubuntu 24.04 (if applicable)

jakebailey:
Fixes #941

image

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -348,6 +348,10 @@ func (p *Program) GetSemanticDiagnostics(ctx context.Context, sourceFile *ast.So
 	return p.getDiagnosticsHelper(ctx, sourceFile, true /*ensureBound*/, true /*ensureChecked*/, p.getSemanticDiagnosticsForFile)
 }
 
+func (p *Program) GetSuggestionDiagnostics(ctx context.Context, sourceFile *ast.SourceFile) []*ast.Diagnostic {
+	return p.getDiagnosticsHelper(ctx, sourceFile, true /*ensureBound*/, true /*ensureChecked*/, p.getSuggestionDiagnosticsForFile)
+}
+
 func (p *Program) GetGlobalDiagnostics(ctx context.Context) []*ast.Diagnostic {
 	var globalDiagnostics []*ast.Diagnostic
 	checkers, done := p.checkerPool.GetAllCheckers(ctx)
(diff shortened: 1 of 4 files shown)


gabritto (approved):

internal/ls/diagnostics.go
```diff
@@ -13,17 +13,24 @@ func (l *LanguageService) GetDocumentDiagnostics(ctx context.Context, documentUR
...
-		lspDiagnostics = make([]*lsproto.Diagnostic, len(syntaxDiagnostics))
-		for i, diag := range syntaxDiagnostics {
-			lspDiagnostics[i] = toLSPDiagnostic(diag, l.converters)
+		lspDiagnostics = make([]*lsproto.Diagnostic, 0, len(syntaxDiagnostics))
+		for _, diag := range syntaxDiagnostics {
+			lspDiagnostics = append(lspDiagnostics, toLSPDiagnostic(diag, l.converters))
 		}
 	} else {
+		// !!! suggestion diagnostics?
+		// !!! In Strada, we consolidate bind/check errors into semantic diagnostics rather than
```
gabritto: Wait, we already consolidate them in `program.GetSemanticDiagnostics()`. Why are we not using it here? Is it because of the multiple checkers thing?

jakebailey: The code below asks for a checker and asks it for diagnostics directly, so nothing here is asking for them anymore.

gabritto: Yes, but why can't it all be replaced by `program.GetSemanticDiagnostics(ctx, file)`?

jakebailey: I had thought becuase `getSemanticDiagnosticsForFile` loops through all the checkers, and we can't do that in the LSP. But I see now that it might actually work fine as it will return just the context's checker. Let me give it a try.

jakebailey: Seems to be fine, yeah. Thanks.

Unfortunately, this PR doesn't seem to be complete and doesn't work when `allowUnreachableCode` is unset (no darkened text), so will have to figure that out.

jakebailey: Oh, because that's also a suggestion diagnostic. Yay.

jakebailey: Okay, that wasn't that hard.

internal/compiler/program.go
```diff
@@ -452,6 +456,39 @@ func (p *Program) getSemanticDiagnosticsForFile(ctx context.Context, sourceFile
 	return filtered
 }
 
+func (p *Program) getSuggestionDiagnosticsForFile(ctx context.Context, sourceFile *ast.SourceFile) []*ast.Diagnostic {
```
jakebailey, line 459: This all is pretty duplicative but I'm not about to attempt to refactor it into some generic version, honestly.

PR #958, opened 2025-05-27
Build VSIX as non-prerelease

jakebailey:
I use typically insiders (which opts people in to prereleases by default), so I never noticed that attempting to install our extension pops up an error in regular VS Code:

image

https://github.com/microsoft/vscode/issues/251730

Given we slap the word "preview" all over this extension, I think it would be better to not mark it as a prerelease, ensuring this error does not appear for the bulk of users.


jakebailey:
Though, maybe this is just a VS Code stable bug.

andrewbranch (approved):

jakebailey:
Undrafting, but we can wait a little to see if https://github.com/microsoft/vscode/issues/251730 is likely to be fixed instead.

Anoesj:
@jakebailey As the issue is bypassed by clicking the "Install Pre-Release version" button in VS Code stable, maybe add that in the [readme](https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.native-preview) of the extension for now. Took me some time to stumble upon this issue :smile:

DanielRosenwasser (approved):

PR #962, opened 2025-05-28
Simplify and fix logic for assignment expression contextual types

### LINKED ISSUE #943: Error with deriving implicit type from initializer
In the following example, I get the error "implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer" with the native preview, whereas in regular typescript this type checks:

```ts
class ReactiveMap<K, V> {
  private readonly _map;

  constructor(entries?: readonly (readonly [K, V])[] | null) {
    this._map = new Map(entries);
  }

  clear(): void {
    this._map.clear();
  }
}
```

Image

I also tried with a local build from `fix-921` branch for (#921) since the error looks the same, but it still produces the error also with that branch build.

ahejlsberg:
Fixes #943.


ahejlsberg:
We already have a failing test case for this issue, so no need to add a regression test.

jakebailey (approved):

PR #970, opened 2025-05-28
Ensure nodes have real parents, fix parent race

andrewbranch:
This is an alternative to #966 that doesn’t do any cloning. Any nodes that are shared between the real tree and reparsed nodes have parent pointers into the real tree.


testdata/baselines/reference/submodule/conformance/jsDeclarationsFunctionClassesCjsExportAssignment.types
```diff
@@ -30,22 +30,22 @@ module.exports = Timer;
  * @param {HookHandler} handle
  */
 function Hook(handle) {
->Hook : (handle: HookHandler) => void
->handle : HookHandler
+>Hook : (handle: (arg: any) => void) => void
```
sandersn, line 33: the main different effect of this alternate-parent technique is that symbols no longer print as aliases. I'll need to figure out why when I start burning down the jsdoc diffs.

testdata/baselines/reference/submodule/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt
```diff
@@ -0,0 +1,39 @@
+index.js(2,28): error TS2304: Cannot find name 'B'.
+index.js(2,42): error TS2304: Cannot find name 'A'.
+index.js(2,48): error TS2304: Cannot find name 'B'.
+index.js(2,67): error TS2304: Cannot find name 'B'.
+
+
+==== index.js (4 errors) ====
+    /**
+     * @typedef {{ [K in keyof B]: { fn: (a: A, b: B) => void; thing: B[K]; } }} Funcs
+                               ~
```
sandersn, line 10: my best guess is that the something set the type expression parents to the synthetic type alias, inconsistent with other type expressions, and the binder change now makes it consistent with the others. I have a branch that sets `Host` on typedef children which should fix this.

Edit:actually it may be that all  type aliases were working by  mistake?

sandersn: actually it may be that *all*  typedefs were setting their type's parent to the synthetic node. Not in the reparser, though, so through the binder's existing code.

jakebailey:
It's probably a temporary thing, though we'll see how much this affects the node builder and declaration emit PR.

andrewbranch:
waiting for #791 to merge first, probably will have to update baselines

andrewbranch:
There seems to be a new logical conflict with #791 around declaration emit for JSDoc `@import` tags.

PR #971, opened 2025-05-29
Watch with way fewer globs in LSP

### LINKED ISSUE #956: Running the lsp causes my editor to rapidly consume all available memory
Hi! 
Thanks for working on the typescript go project. I'm having an issue with the LSP. When i start it together with Neovim it causes Neovim to freeze and rapidly consume all memory available and crash. 

I have tried installing via npm and building tsgo from source, both causes the same behaviour.

Before the preview was released I was able to build tsgo from source and run tsgo with neovim without problems so I'm guessing it's a recent change.

Below is a screenshot of neovim running with the tsgo lsp. The top process is the nvim process which spawned tsgo.
Image

I'm starting the lsp with this command:
```sh
tsgo --lsp --stdio
```

Any help to resolve this would be greatly appreciated! If you need any more information from me please reach out!

jakebailey:
Fixes #956

Before, opening the VS Code source would net a super large number of file watchers: 

image
image

With this PR, this is much reduced:

image

And, the root file globs have been fixed (though it's possible VS Code was normalizing them).

This is not particularly fast; for VS Code it takes some 60ms to process all of the failed lookup locations, but it doesn't change often.

### DIFF
--- a/internal/project/watch.go
+++ b/internal/project/watch.go
@@ -2,7 +2,9 @@ package project
 
 import (
 	"context"
+	"maps"
 	"slices"
+	"strings"
 
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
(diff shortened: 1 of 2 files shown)


internal/project/watch.go
```diff
@@ -60,3 +65,267 @@ func (w *watchedFiles[T]) update(ctx context.Context, newData T) (updated bool,
...
+
+type directoryOfFailedLookupWatch struct {
+	dir            string
+	dirPath        tspath.Path
+	nonRecursive   bool
+	packageDir     *string
+	packageDirPath *tspath.Path
+}
+
+func getDirectoryToWatchFailedLookupLocation(
```
jakebailey, line 121: This is a pure port, basically line for line. I made no attempts to make it go any faster.

ctrlplusb:
thank you for this. my machine was absolutely crunching.

internal/project/watch.go
```diff
@@ -60,3 +65,267 @@ func (w *watchedFiles[T]) update(ctx context.Context, newData T) (updated bool,
...
+			if g.nonRecursive {
+				globs = append(globs, g.dir+"/"+fileGlobPattern)
+			} else {
+				globs = append(globs, g.dir+"/"+recursiveFileGlobPattern)
+			}
+		}
+		slices.Sort(globs)
+
+		took := time.Since(start)
+		host.Log(fmt.Sprintf("createGlobMapper took %s to create %d globs for %d failed lookups", took, len(globs), len(data)))
```
DanielRosenwasser:
```suggestion
		timeTaken := time.Since(start)
		host.Log(fmt.Sprintf("createGlobMapper took %s to create %d globs for %d failed lookups", timeTaken, len(globs), len(data)))
```

internal/project/watch.go
```diff
@@ -60,3 +65,267 @@ func (w *watchedFiles[T]) update(ctx context.Context, newData T) (updated bool,
...
+				rootPath,
+				rootPathComponents,
+				isRootWatchable,
+				rootDir,
+				true,
+			)
+			if w == nil {
+				continue
+			}
+			globSet.Add(glob{dir: w.dir, nonRecursive: w.nonRecursive})
```
DanielRosenwasser: If you've seen a non-recursive watcher followed by a recursive watcher, would it be better to only rely on the recursive watcher glob?

Maybe instead of a `Set[glob]` you use something like a `map[string]bool` to track directories, but map to whether a recursive glob has ever been encountered.

internal/project/watch.go
```diff
@@ -60,3 +64,263 @@ func (w *watchedFiles[T]) update(ctx context.Context, newData T) (updated bool,
...
+func getDirectoryToWatchFailedLookupLocation(
+	failedLookupLocation string,
+	failedLookupLocationPath tspath.Path,
+	rootDir string,
+	rootPath tspath.Path,
+	rootPathComponents []string,
+	isRootWatchable bool,
+	currentDirectory string,
+	preferNonRecursiveWatch bool,
+) *directoryOfFailedLookupWatch {
```
sheetalkamat, line 130: Strada has baselines on what to watch : https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/canWatch/getDirectoryToWatchFailedLookupLocationIndirDos.baseline.md like these. do we want these atleast till we have this logic and dont make this any faster.

internal/project/watch.go
```diff
@@ -60,3 +64,263 @@ func (w *watchedFiles[T]) update(ctx context.Context, newData T) (updated bool,
...
+				true,
+			)
+			if w == nil {
+				continue
+			}
+			globSet[w.dir] = globSet[w.dir] || !w.nonRecursive
+		}
+
+		globs := make([]string, 0, len(globSet))
+		for dir, recursive := range globSet {
```
sheetalkamat, line 98: In strada there is code to watch "packageDir" if it exits and dir otherwise so we can handle the symlinks .. a note here to handle package symlinks later would be good

internal/project/watch.go
jakebailey: We can do that, though I'm wary of continuing to have these sorts of super large baselines.

internal/project/watch.go
jakebailey: Ah, I should have added that, yeah. Will follow up with that.

jakebailey: I was dogfooding `tsgo` so couldn't find all refs on it; is it just `packageDir ?? dir`?

sheetalkamat: its packageDir if packageDir exists and is symlink otherwise dir.   Its in `createDirectoryWatcherForPackageDir` in strada. Its little complicated with having to manage state but since you dont need to, it should make it easier.

PR #975, opened 2025-05-29
Update “Declaration emit” in README progress table

controversial:
Follow up to update the README now that #791 is merged!

I tried to write notes based on the TODOs in the PR but please correct me if I’m misrepresenting the status of this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ This is still a work in progress and is not yet at full feature parity with Type
 | Type checking | done | Same errors, locations, and messages as TS5.8. Types printback in errors may display differently (in progress) |
 | JavaScript-specific inference and JS Doc | not ready | - |
 | JSX | done | - |
-| Declaration emit | in progress | Most common features are in-place, but some edge cases and feature flags are still unhandled |
+| Declaration emit | in progress | Most common features are in place, but some edge cases and feature flags are still unhandled |
 | Emit (JS output) | in progress | `target: esnext` well-supported, other targets may have gaps |
 | Watch mode | prototype | Watches files and rebuilds, but no incremental rechecking |
 | Build mode / project references | not ready | - |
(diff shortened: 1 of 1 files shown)


controversial:
Actually, it seems like declaration support for JS files / JSDoc is working; a fresh build of tsgo transforms this `.js` file
```js
/**
 * @param {number} a
 * @param {number} b
 */
export function func1(a, b) {
    return a + b;
}
export function func2() {
    const a = 1;
    return a;
}
```

into this `.d.ts` file:
```ts
/**
 * @param {number} a
 * @param {number} b
 */
export declare function func1(a: number, b: number): number;
export declare function func2(): unknown;
```

so maybe I’m misinterpreting 
> - [ ]  JS declaration emit support (will need to be substantively rewritten given different upfront parsing and checking of JSDoc structures - likely for the better)

jakebailey:
The new codebase uses an all-new method to handle JSDoc where we actually rewrite the AST during parse to inject type nodes where TS files would have them, so some JS stuff is likely to "just work" without any other work. But, not everything works with that method.

jakebailey:
It's not incorrect; it's definitely not thoroughly tested. We're not even claiming JS/JSDoc is done, so it would be weird to say that its declaration emit is already done.

controversial:
got it, awesome!

I did notice that tsgo emits `export declare function` statements for that .js file, where tsc emitted plain `export function` statements (but both use `export declare` in declarations for typescript source files)

controversial:
> We're not even claiming JS/JSDoc is done, so it would be weird to say that its declaration emit is already done

I removed the part of the note that makes claims about JS/JSDoc for now, since it seems untrue that it *doesn’t* work.

> it's definitely not thoroughly tested

Understand if you want to hold off on making claims about declaration emit in the status table until there’s been more thorough testing! Excited about the progress here.

README.md
```diff
@@ -84,7 +84,7 @@ This is still a work in progress and is not yet at full feature parity with Type
 | Type checking | done | Same errors, locations, and messages as TS5.8. Types printback in errors may display differently (in progress) |
 | JavaScript-specific inference and JS Doc | not ready | - |
 | JSX | done | - |
-| Declaration emit | not ready | Coming soon |
+| Declaration emit | done | Some types may be formatted slightly differently |
```
weswigham:
```suggestion
| Declaration emit | in progress | Most common features are in-place, but some edge cases and feature flags are still unhandled |
```

weswigham (approved):

PR #979, opened 2025-05-29
Implement externalModuleIndicator

jakebailey:
The infamous `externalModuleIndicator` field requires access to info like the module setting of the closest package.json in order to work in some situations. This is normally "fine"; just make a sidecar that the Program holds onto. But unfortunately, `externalModuleIndicator` is needed to know whether or not a given file needs to be reparsed to deal with top-level await, so needs to be there at parse time. This means that it's technically an _input_ to parse, not something on the side.

To deal with this, I've eliminated the `SourceFileMetaData` sidecar, moving it back onto `SourceFile` and populating it during parse. This means less AST reuse, however, the info here is effectively entirely derived from the FS (the path and what's on disk), so it's unlikely files that _only_ differed by this metadata were actually sharing anything.

Fixing this eliminates a bunch of random diffs.


internal/parser/parser.go
```diff
@@ -355,6 +358,83 @@ func (p *Parser) finishSourceFile(result *ast.SourceFile, isDeclarationFile bool
...
+	if file.IsDeclarationFile {
+		return nil
+	}
+	return walkTreeForJSXTags(file.AsNode())
+}
+
+// This is a somewhat unavoidable full tree walk to locate a JSX tag - `import.meta` requires the same,
+// but we avoid that walk (or parts of it) if at all possible using the `PossiblyContainsImportMeta` node flag.
+// Unfortunately, there's no `NodeFlag` space to do the same for JSX.
+func walkTreeForJSXTags(node *ast.Node) *ast.Node {
```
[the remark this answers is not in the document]
jakebailey: Pal, `SubtreeContains` _is_ the cache

andrewbranch:
I need to look in more detail later, but my immediate thought is that the cache key could have the pre-computed value of “force this file to be a module” instead of all the compiler options that _might_ contribute to that value. Right now, differing `moduleResolution` or `jsx` settings will prevent source file reuse, but that needn’t be true if both projects have `--moduleDetection force`.

jakebailey:
That was my thought too, but the trouble is that the indicator is an AST node, because we need to be able to point diagnostics at it in some situations. But, for these new cases they're basically all just what used to be "true", so potentially that would work as a plain boolean passed in.

I'm just not sure that we'll actually end up in the situation where that reuse would have happened, though.

andrewbranch:
I think it’s pretty common for monorepos to have projects where these settings differ, but the actual parse doesn’t. It seems silly to not be able to share common declaration files between those projects, when the declaration files themselves are never affected by `moduleDetection` in the first place.

jakebailey:
This PR is pretty broken now; I'm going to redo it, not eliminate the metadata sidecar and instead still pass stuff into the parser as an input, which should produce the smallest diff.

We discussed a new way to do reparsing, however that's a much more difficult undertaking, not just because it means redoing things, but also because reparsing happening in the parser during the initial parse means that the Parser struct is still there and populated with stuff like the identifier list and so on (maybe other important state), so it needs a little bit of thinking.

internal/project/project.go
```diff
@@ -1068,7 +1070,7 @@ func (p *Project) Close() {
 
 	if p.program != nil {
 		for _, sourceFile := range p.program.GetSourceFiles() {
-			p.host.DocumentRegistry().ReleaseDocument(sourceFile, p.program.Options())
+			p.host.DocumentRegistry().ReleaseDocument(sourceFile, p.program.Options(), p.program.GetSourceFileMetaData(sourceFile.Path()))
```
jakebailey, line 1073: I really don't like how much jumping through hoops I have to do for this.

Honestly, it seems to me like the the entire set of parameters to `ParseSourceFile` _are_ the key, and that we could store a copy on the source files, rather than constantly asking the program for them.

jakebailey: Of course, that would not work out for that differing parse plan.

andrewbranch: It's correct that the key and the parse params are the same and storing a copy/pointer on the file makes sense. It's just that we may want to try to trim down the key/params later, with `SourceFileMetadata` in particular containing some stuff we might want to pull off. But I think it's fine to make a type that is the cache key and store it on the file.

jakebailey: Should I make that change now? Would theoretically simplify things, though at that point I would find it strange that we'd _also_ have the sidecar and plumb that access through everywhere.

andrewbranch: No, I might see if I can follow up with a way to just pass in the module status, so let’s do that first

jakebailey: Yeah, the change to make the above work is extensive, I spent a while trying to make it work and it exposed a bunch of other weirdness. I'll send a different PR for that.

PR #981, opened 2025-05-29
Fix up hover code, mostly temporary

jakebailey:
The writing and clearing and passing down of things doesn't seem to be working; for now just flip this all to use the returned string until we can do something faster. (I'm not doing something faster because some of the code does some truncation that is not easily doable via streaming.)

Also the unknown symbol flip.

### DIFF
--- a/internal/checker/printer.go
+++ b/internal/checker/printer.go
@@ -171,6 +171,10 @@ func toNodeBuilderFlags(flags TypeFormatFlags) nodebuilder.Flags {
 	return nodebuilder.Flags(flags & TypeFormatFlagsNodeBuilderFlagsMask)
 }
 
+func (c *Checker) TypeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags) string {
+	return c.typeToStringEx(t, enclosingDeclaration, flags)
+}
+
 func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags) string {
 	writer := printer.NewTextWriter("")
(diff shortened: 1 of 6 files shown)


jakebailey:
Last commit optional, that's mainly there per request (presumably to make the markdown easier)

internal/checker/printer.go
```diff
@@ -343,20 +315,20 @@ func (c *Checker) valueToString(value any) string {
...
-func (c *Checker) WriteSymbol(symbol *ast.Symbol, enclosingDeclaration *ast.Node, meaning ast.SymbolFlags, flags SymbolFormatFlags, writer printer.EmitTextWriter) string {
-	return c.symbolToStringEx(symbol, enclosingDeclaration, meaning, flags, writer)
+func (c *Checker) WriteSymbol(symbol *ast.Symbol, enclosingDeclaration *ast.Node, meaning ast.SymbolFlags, flags SymbolFormatFlags) string {
+	return c.symbolToStringEx(symbol, enclosingDeclaration, meaning, flags)
 }
 
-func (c *Checker) WriteType(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags, writer printer.EmitTextWriter) string {
-	return c.typeToStringEx(t, enclosingDeclaration, flags, writer)
+func (c *Checker) WriteType(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags) string {
+	return c.typeToStringEx(t, enclosingDeclaration, flags)
```
ahejlsberg: Just get rid of these `WriteXXX` functions now? They're just needless wrappers.

ahejlsberg:
Otherwise, looks good to me.

internal/checker/printer.go
jakebailey: I can but then everything they call has to be exported. I guess I can make that change.

PR #991, opened 2025-05-30
noslash

gabritto:
This PR adds the basics of fourslash tests to Corsa.

- There's now a `fourslash` package with the basic setup for running fourslash tests.
   + The new fourslash setup creates a server and talks to it, so this would be closer to "fourslash server" tests in Strada. I think we won't need the non-server tests.
- There's a script for converting between Strada fourslash in `internal/fourslash/_scripts/convertFourslash.mts`.
  + You can do `npm run convertfourslash` to run the script.
  + The script optionally takes in a file containing a list of Strada tests to translate. This is useful when working on the script to make it attempt to translate only tests that are related to e.g. completions.
  + The script outputs a list of files it failed to parse, and it logs the reason things failed to stderr.
  + Generated tests that currently fail are skipped via `t.Skip()`. The current list of tests expected to fail is in `internal/fourslash/_scripts/failedTests.txt`.
  
Didn't yet include all the generated tests in the PR because it'd break GitHub.
Included example of a test the script generates: https://github.com/microsoft/typescript-go/pull/991/files#diff-9c51d0662c595fdc5d7eb5e3521e692c744a7711906f1bcceee767b32cfb9252R11

Planning to do in future PRs:
- CI action to verify that the checked-in generated fourslash tests match what the script generates.
- Better panic handling for the server, and better error message for fourslash tests when server panics.
For reference, here's what a panic in the server currently looks like:
image

- Improve script to parse more completions tests.
- Completion baselines.

@iisaduan is working on handling ranges.


internal/project/project.go
```diff
@@ -192,6 +213,11 @@ func (p *Project) GetCompilerOptions() *core.CompilerOptions {
 
 // GetSourceFile implements compiler.CompilerHost.
 func (p *Project) GetSourceFile(fileName string, path tspath.Path, languageVersion core.ScriptTarget) *ast.SourceFile {
+	if p.getCachedSourceFile != nil {
```
jakebailey: This is just a draft, but I would think we'd want to actually do this down where we do the parse call, since this will prevent us from testing the doc registry logic.

gabritto: What doc registry logic?

jakebailey: The code just below my comment that this is bypassing.

internal/ls/completions_test.go
```diff
@@ -1387,12 +1387,12 @@ export function isAnyDirectorySeparator(charCode: number): boolean {
 									InsertReplaceEdit: &lsproto.InsertReplaceEdit{
 										NewText: "CharacterCodes",
 										Insert: lsproto.Range{
-											Start: lsproto.Position{Line: 27, Character: 0},
-											End:   lsproto.Position{Line: 27, Character: 1},
+											Start: lsproto.Position{Line: 33, Character: 0},
```
gabritto, line 1390: Result of adding marker character validation (the file now correctly includes a block comment that was parsed as a marker before).

internal/ls/completions.go
```diff
@@ -31,12 +31,16 @@ func (l *LanguageService) ProvideCompletion(
 	preferences *UserPreferences,
 ) (*lsproto.CompletionList, error) {
 	program, file := l.getProgramAndFile(documentURI)
+	var triggerCharacter *string
```
gabritto, line 35: Refactored this file to get rid of nil reference crashes when accessing client capabilities and trigger character.

internal/ls/utilities.go
```diff
@@ -987,7 +986,7 @@ func findPrecedingMatchingToken(token *ast.Node, matchingTokenKind ast.Kind, sou
 	tokenKind := token.Kind
 	remainingMatchingTokens := 0
 	for {
-		preceding := astnav.FindPrecedingToken(sourceFile, tokenFullStart)
+		preceding := astnav.FindPrecedingToken(sourceFile, token.Pos())
```
gabritto, line 1585: Bug fix for an infinite loop bug caught by one of the generated tests.

internal/lsp/server.go
```diff
@@ -639,3 +712,18 @@ func ptrIsTrue(v *bool) bool {
 	}
 	return *v
 }
+
```
gabritto, line 845: Same as in `completions.go`: we were accessing several properties of the initialization params without first checking for `nil`, now we are not.

internal/lsp/server.go
```diff
@@ -194,7 +256,7 @@ func (s *Server) Run() error {
 	g, ctx := errgroup.WithContext(ctx)
 	g.Go(func() error { return s.dispatchLoop(ctx) })
 	g.Go(func() error { return s.writeLoop(ctx) })
-	g.Go(func() error { return s.readLoop(ctx) })
+	go func() { _ = s.readLoop(ctx) }() // !!! do something with the error here?
```
gabritto: We can't wait for the read loop to return here because the other two loops may have crashed but the read loop could be blocked on a `read()` call, such that the server never closes, and then fourslash/the client also remains blocked.

internal/testrunner/test_case_parser.go
```diff
@@ -42,19 +41,88 @@ var optionRegex = regexp.MustCompile(`(?m)^\/{2}\s*@(\w+)\s*:\s*([^\r\n]*)`)
...
+		tsConfigFileUnitData: tsConfigFileUnitData,
+		symlinks:             symlinks,
+	}
+}
+
+// !!! move this to a different package
+// Given a test file containing // @FileName and // @symlink directives,
+// return an array of named units of code to be added to an existing compiler instance,
+// along with a map of symlinks and the current directory.
+func ParseTestFilesAndSymlinks[T any](
```
gabritto, line 108: I extracted this function from the old `makeUnitsFromTest` so that I can reuse it for fourslash test parsing.

_scripts/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "private": true,
+    "name": "scripts",
+    "version": "1.0.0",
+    "description": "Test conversion script",
+    "scripts": {
+        "build": "npx tsc -p .",
+        "runjs": "node dist/convertFourslash.js",
+        "run": "npx tsc -p . && node dist/convertFourslash.js"
```
jakebailey: I think I'd prefer if this code were moved into a subdir within the fourslash package, similarly to the LSP generation script, rather than introducing another top-level directory. You can also just use node's type stripping (or mjs), and avoid package.json or any deps entirely. See also the API tests.

cmd/tsgo/lsp.go
```diff
@@ -42,8 +42,8 @@ func runLSP(args []string) int {
 	typingsLocation := getGlobalTypingsCacheLocation()
 
 	s := lsp.NewServer(&lsp.ServerOptions{
-		In:                 os.Stdin,
-		Out:                os.Stdout,
+		In:                 lsp.ToLSPReader(os.Stdin),
+		Out:                lsp.ToLSPWriter(os.Stdout),
```
jakebailey: In context, it seems like these names are duplicative (LSP twice).

internal/fourslash/test_parser.go
```diff
@@ -0,0 +1,258 @@
...
+
+import (
+	"strings"
+	"testing"
+	"unicode/utf8"
+
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/ls"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	runner "github.com/microsoft/typescript-go/internal/testrunner"
```
jakebailey: We should rename the `testrunner` package internally, it shouldn't be called `runner`. I think I must have missed it when I moved it up a dir.

internal/project/project.go
gabritto: Working on a fix.

internal/fourslash/test_parser.go
```diff
@@ -0,0 +1,258 @@
...
+const (
+	stateNone parserState = iota
+	stateInSlashStarMarker
+	stateInObjectMarker
+)
+
+func parseFileContent(filename string, content string, fileOptions map[string]string) *testFileWithMarkers {
+	filename = tspath.GetNormalizedAbsolutePath(filename, "/")
+	// Any slash-star comment with a character not in this string is not a marker.
+	validMarkerChars := core.NewSetFromItems(
```
jakebailey: This seems like it could just be some calls to `stringutil`?

internal/fourslash/tests/basicInterfaceMembers_test.go
```diff
@@ -0,0 +1,54 @@
...
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `export {};
+interface Point {
+	x: number;
+	y: number;
+}
+declare const p: Point;
+p./*a*/`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
```
jakebailey: Rather than returning a `done` func to defer, you can just use `t.Cleanup()` from inside `NewFourslash`, which will tell the testing package to run it after the test completes.

internal/lsp/server.go
```diff
@@ -63,9 +68,60 @@ type pendingClientRequest struct {
 	cancel context.CancelFunc
 }
 
-type Server struct {
+type LSPReader interface {
+	Read() (*lsproto.Message, error)
+}
+
+type LSPWriter interface {
```
jakebailey: I'd just call these `Reader` and `Writer`.

internal/lsp/server.go
```diff
@@ -91,6 +147,12 @@ type Server struct {
 	watchers       core.Set[project.WatcherHandle]
 	logger         *project.Logger
 	projectService *project.Service
+
+	// enables tests to share a cache of parsed source files
+	getCachedSourceFile func(string, tspath.Path, core.ScriptTarget) *ast.SourceFile
+
+	// !!! temporary; remove when we have `handleDidChangeConfiguration`/implicit project config support
+	compilerOptionsForInferredProjects *core.CompilerOptions
```
jakebailey, line 153: It's likely that we'll do this differently, probably via configuration watches and gets.

internal/lsp/server.go
```diff
@@ -91,6 +147,12 @@ type Server struct {
 	watchers       core.Set[project.WatcherHandle]
 	logger         *project.Logger
 	projectService *project.Service
+
+	// enables tests to share a cache of parsed source files
+	getCachedSourceFile func(string, tspath.Path, core.ScriptTarget) *ast.SourceFile
```
jakebailey: I'm wary of this but I'm not sure what to do otherwise. The trouble is that this would bypass all of the logic the server/service/project that do their own caching, but perhaps we just need to test that without fourslash.

internal/lsp/server.go
jakebailey: This is probably something we should just go take into main, honestly. (Same for many of the fixes you've found in this PR.)

internal/fourslash/tests/basicInterfaceMembers_test.go
```diff
@@ -0,0 +1,54 @@
...
+	f.VerifyCompletions(t, "a", &fourslash.VerifyCompletionsExpectedList{
+		IsIncomplete: false,
+		ItemDefaults: &lsproto.CompletionItemDefaults{
+			CommitCharacters: &fourslash.DefaultCommitCharacters,
+		},
+		Items: &fourslash.VerifyCompletionsExpectedItems{
+			Exact: []fourslash.ExpectedCompletionItem{
+				&lsproto.CompletionItem{
+					Label:      "x",
+					Kind:       fourslash.PtrTo(lsproto.CompletionItemKindField),
```
jakebailey: Rather than exporting `ptrTo`, we should just declare it in this package locally; otherwise we're just writing this long thing over and over again.

(At least, until [golang/go#45624](https://redirect.github.com/golang/go/issues/45624))

internal/lsp/server.go
```diff
@@ -288,6 +345,14 @@ func (s *Server) dispatchLoop(ctx context.Context) error {
 			}
 
 			handle := func() {
+				defer func() {
```
jakebailey, line 355: Probably it's worth doing this on main; we really only want to kill the LS when a panic happens during the handling of a critical message, otherwise it should just be returning an error or something.

It's also possible to store off onto the LS the error and then in the test harness check after each use that no new errors appeared.

internal/project/service.go
```diff
@@ -787,3 +795,13 @@ func (s *Service) printProjects() {
 func (s *Service) logf(format string, args ...any) {
 	s.Log(fmt.Sprintf(format, args...))
 }
+
+// !!! per root compiler options
+func (s *Service) SetCompilerOptionsForInferredProjects(compilerOptions *core.CompilerOptions) {
+	s.compilerOptionsForInferredProjects = compilerOptions
```
jakebailey, line 985: This is probably going to require some form of synchronization when we do start calling it.

Just to expand further on what I think we should do as an alternative:

The LSP allows servers to call back into the client and ask for configuration (e.g., VS Code settings), and also ask the client to send a notification when that configuration has changed. This means that if we want to allow the user to override the default configuration, we'd call back into the client for the info and register a "watcher" on it. Then, any configuration changes would come in as plain messages and be correctly synchronized by the handler loop.

This is not something I would expect in this PR whatsoever, obviously.

internal/fourslash/fourslash.go
```diff
@@ -0,0 +1,478 @@
...
+	testData *TestData
+
+	currentCaretPosition lsproto.Position
+	currentFilename      string
+	lastKnownMarkerName  string
+	activeFilename       string
+}
+
+type lspReader struct {
+	c <-chan *lsproto.Message
```
jakebailey, line 41: I feel suspicious of the concurrency in this, solely because it will become normal for the server to make calls back into the client (which can then make calls back into the server, ad infinitum).

Channels are the right thing of course, maybe this is just something to consider when writing the harness itself, which sends off a request and then is presumably sitting there waiting for the reply to the request.

gabritto: I think this would be part of what I was thinking when I wrote this to do:
https://github.com/microsoft/typescript-go/pull/991/files#diff-635a5113ad9af7a728a8dc541629f50378e06c4afd9e8247533d2452981a6896R227

internal/lsp/server.go
```diff
@@ -288,6 +345,14 @@ func (s *Server) dispatchLoop(ctx context.Context) error {
...
+					if r := recover(); r != nil {
+						stack := debug.Stack()
+						s.Log("panic obtaining completions:", r, string(stack))
```
weswigham: This applies to more than just completions here, right? Should it log a bit more about the request that triggered a crash?

jakebailey: Yeah, I think it's time for a generic solution to this.

internal/project/project.go
gabritto: Hopefully fixed now; it was also entirely not working before because I stopped halfway through and forgot about it.

internal/lsp/server.go
jakebailey: I pulled this out into https://github.com/microsoft/typescript-go/pull/1074

internal/lsp/server.go
```diff
@@ -204,40 +264,45 @@ func (s *Server) Run() error {
...
 func (s *Server) readLoop(ctx context.Context) error {
 	for {
-		msg, err := s.read()
-		if err != nil {
-			if errors.Is(err, lsproto.ErrInvalidRequest) {
-				s.sendError(nil, err)
-				continue
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
jakebailey: For this, you don't need a select; just do `if err := ctx.Err(); err != nil { return err }`.

internal/fourslash/test_parser.go
gabritto: What calls? I know `unicode` has `IsDigit` and `IsLetter`, but not e.g. `IsASCIILetter`. I'm not sure this distinction matters though.

jakebailey: I was just thinking `stringutil.IsASCIILetter(c) || stringutil.IsDigit(c) || c == '$' || c == '_'` or something

gabritto:
Race mode test failure should be fixed by #1075.

package.json
```diff
@@ -27,15 +27,16 @@
...
-        "node": "node --no-warnings --conditions @typescript/source"
+        "node": "node --no-warnings --conditions @typescript/source",
+        "convertfourslash": "node --experimental-strip-types --no-warnings internal/fourslash/_scripts/convertFourslash.mts"
     },
     "workspaces": [
         "./_extension",
         "./_packages/*"
     ],
-    "packageManager": "npm@8.19.4",
+    "packageManager": "npm@10.8.2",
```
jakebailey: Can you leave the npm versions? I'd like to update it in a separate PR, since this will require regenerating the lockfile.

.gitignore
```diff
@@ -196,3 +196,5 @@ custom-gcl.hash
 !testdata/submoduleAccepted.txt
 
 !NOTICE.txt
+
+!internal/fourslash/_scripts/failingTests.txt
```
jakebailey: Nit: missing the newline at the end of this file (you may not have that option configured in vscode)

internal/fourslash/fourslash.go
```diff
@@ -0,0 +1,507 @@
...
+		DefaultLibraryPath: bundled.LibPath(),
+
+		ParsedFileCache: &parsedFileCache{},
+	})
+
+	go func() {
+		defer func() {
+			outputWriter.Close()
+		}()
+		_ = server.Run()
```
jakebailey: Does this error need to be checked with a `t.Error`? (Doesn't need to be t.Fatal since that wouldn't work in the wrong goroutine, but it would at least error out if the exit was bad)

internal/fourslash/fourslash.go
```diff
@@ -0,0 +1,507 @@
...
+		t.Fatalf("Expected completion item to be a string or *lsproto.CompletionItem, got %T", item)
+		return ""
+	}
+}
+
+func assertDeepEqual(t *testing.T, actual any, expected any, prefix string) {
+	t.Helper()
+
+	diff := cmp.Diff(actual, expected)
+	if diff != "" {
```
jakebailey, line 508: Is this equivalent to just doing `assert.DeepEqual`? Or is that just bad looking?

gabritto: It should be, I just needed a way to print a custom message before printing the actual diff, which `assert.DeepEqual` doesn't allow.

jakebailey (approved):

jakebailey (approved):

PR #995, opened 2025-05-30
Use packageDir in watch globbing

jakebailey:
Followup to https://github.com/microsoft/typescript-go/pull/971#discussion_r2114352757, but this makes creating globs a bit slower due to the FS accesses to see if things are symlinked.


sheetalkamat:
Should wait on this till we do watching in separate go routine #1217

jakebailey:
Yeah, I don't intend to do this now, but I was just updating the branch before it got so out of date my editor crashed trying to merge `main` 😄

PR #996, opened 2025-05-30
Cache FS in LS

jakebailey:
I think this is right; within a Project, cache the FS accesses and clear them out on graph update for safety. Then, things get faster in load time in module resolution and so on.

Not a mega fan of the whole host wrapping but I don't see a better way.


internal/project/project.go
```diff
@@ -141,28 +142,46 @@ func NewInferredProject(compilerOptions *core.CompilerOptions, currentDirectory
...
+	if project.host.IsWatchEnabled() && client != nil {
+		globMapper := createGlobMapper(project.host)
 		project.failedLookupsWatch = newWatchedFiles(client, lsproto.WatchKindCreate, globMapper)
 		project.affectingLocationsWatch = newWatchedFiles(client, lsproto.WatchKindChange|lsproto.WatchKindCreate|lsproto.WatchKindDelete, globMapper)
 	}
 	project.markAsDirty()
 	return project
 }
 
+type projectHostWithCachedFS struct {
```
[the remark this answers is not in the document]
andrewbranch: You're telling me `projectHostWithCachedFS` wraps a project host with a cached FS?

sheetalkamat (approved):

sheetalkamat:
We did this in Strada too (though we did it before creating program in service) this is known to help perf.

jakebailey:
Unfortunately this method doesn't really work, the project test system replaces the FS out from under the cache, but the cache FS grabs the FS on creation.

I've been able to change the tests to not do that, but I'm not clearing the cache in places that seemingly we need it. So this will need more work.

jakebailey:
Oh oops, sorry, this was supposed to be a draft when I sent it 😄

jakebailey:
CI fails because the new ATA code runs before updateGraph and therefore reads from the cached FS before it's cleared.

internal/project/project.go
```diff
@@ -462,6 +486,7 @@ func (p *Project) updateGraph() bool {
 
 	start := time.Now()
 	p.Log("Starting updateGraph: Project: " + p.name)
+	p.host.fs.ClearCache()
```
sheetalkamat: I think we should keep the cache only for updateGraph duration as otherwise we dont know if its correct or not.

sheetalkamat: you probably saw this with typings installer tests but in practice this could be watch events as well

jakebailey: I'm not sure where you're telling me to put the call; I don't know where this would go otherwise.

Are you saying to clear the cache at the end of updateGraph?

I can do it that way, but I don't think that would affect ATA, since that code can happen any time concurrently (part of the reason why I am not giving it a cached FS, and plan to decouple it from the Project in a followup to fix the races)

sheetalkamat: i meant to keep the cache only for duration of update graph " so uncached before and after that call. (eg of how we did this in strada - this ensures the correctness if for some reason we need to check with FS on file existence etc  https://github.com/microsoft/TypeScript/blob/main/src/services/services.ts#L1782 where cache lives only for program creation and then goes away

jakebailey: So, defer this call so it runs when updateGraph returns?

sheetalkamat: start caching at start of updateGraph and end cacheing when returning from it

jakebailey: Updated, PTAL

internal/project/project.go
```diff
@@ -460,6 +484,8 @@ func (p *Project) updateGraph() bool {
 		return false
 	}
 
+	defer p.host.fs.ClearCache()
```
sheetalkamat: But this still keeps cache outside of upate graph right. You probably want cache to be present only during updateGraph and use uncached version outside of it, that is it should not cache not consult cache when outside the updateGraph call

jakebailey: I see, you want me to pass a custom host into the Program?

jakebailey: The trouble is that the entire project is "a host", so it's pretty tough to do that, unless I make the cached FS able to be disabled entirely on some sort of toggle, which seems odd...

(This is why I hate every object turning into a host, but here we are...)

jakebailey: Maybe what I should do is create the cache in Program, which is how it's done for the other code (`NewCachedFSCompilerHost`).

sheetalkamat: i think if you create cache in program you wont be able to use it for your globs etc for watching.. you would want cache host to be "enabled" only for duration of the update graph. we did this eg in tsbuild in strada where we would cache only  when build was in progress

jakebailey: I'll give it a shot.

jakebailey: Well, I did it, but it's not pretty 😄

jakebailey:
Another ATA test race 😦

internal/project/project.go
sheetalkamat: I think you want to set  and unset cacheHost inself ? so that its not checked all over the place?

also i think ata needs to used uncached version all the time? since it could be doing things in background while project is updating

jakebailey: I really do not want to swap the hosts out from under anything. That was done once before in testing and it was a pain because many things call `FS()`, etc, storing off the value, and so are not updated, which then has to be chased down. Ensuring everything is always static is very simplifying.

I'm less concerned about the extra check just because it's only a single atomic access most of the time.

> also i think ata needs to used uncached version all the time? since it could be doing things in background while project is updating

I thought about this, but if both are happening concurrently, there's no harm in both using the same cached info, since updateGraph will clear it on exit anyhow.

gabritto (approved):

PR #999, opened 2025-05-30
Program to take ParsedCommandLine as config instead of separate options, root names, diagnostics etc

sheetalkamat:


internal/project/project.go
```diff
@@ -512,16 +514,37 @@ func (p *Project) updateProgram() bool {
 	}
 	var oldProgramReused bool
 	if p.program == nil || p.dirtyFilePath == "" {
-		rootFileNames := p.GetRootFileNames()
-		compilerOptions := p.compilerOptions
+		if p.programConfig == nil {
+			if p.parsedCommandLine != nil {
```
jakebailey: Trying to understand this bit of code; is it just that the parsed config we get from the config file needs to be revised to have only the files we want in the root? Why can't we just pass it straight in? Presumably we've already loaded the other files anyway, they're just not being queried?

sheetalkamat: this is to copy typings installer typings files that are set as root

jakebailey (approved):

PR #1001, opened 2025-05-30
Various node builder related performance optimizations

jakebailey:
- Eliminate the closures; this can be done through the same tricks used in the parser to avoid binding. This reduces some 10%+ allocations.
- Presize `nonTypeOnlyNames`.
- Eliminate the closures from the emit context.
- Reuse emit contexts in type baselines.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -14904,7 +14904,7 @@ type ExportCollisionTable = map[string]*ExportCollision
 
 func (c *Checker) getExportsOfModuleWorker(moduleSymbol *ast.Symbol) (exports ast.SymbolTable, typeOnlyExportStarMap map[string]*ast.Node) {
 	var visitedSymbols []*ast.Symbol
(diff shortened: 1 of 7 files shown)


weswigham (approved):
Is this now our preferred method closure pattern?

jakebailey:
Kinda; in this case I'm changing it because the rebinding was very expensive (the most allocation done by the new code).

For ephemeral stuff like the checker, just using fields is more ergonomic and not much slower. For these kinds of pooled / reused / reset resources, managing the fields is kind of annoying, so passing in a method directly `(*Parser).parseSomething` to another method on the same type for a simple callback is not too inergonomic.

The `core` helpers I like the least and arguably could just be removed from this PR as I don't think they actually helped perf in my testing.

jakebailey:
I checked, and the captures I "avoided" with the `core` package changes don't matter. The functions are so small that I believe the compiler is figuring it out (either by inlining, but most likely by realizing that the closure is not leaked to the heap).

image

PR #1002, opened 2025-05-30
Hardcode contents of typesMap.json

sheetalkamat:


internal/project/typesmap.go
```diff
@@ -0,0 +1,505 @@
+package project
+
+type SafeListEntry struct {
+	Match   string
+	Exclude [][]any
+	Types   []string
+}
```
jakebailey: FWIW this all could be unexported, given it's only used within this package. (e.g, `safeListEntry`).

jakebailey (approved):

PR #1003, opened 2025-05-30
Ensure we don't copy OrderedMaps

### LINKED ISSUE #978: lsp hangs when a javascript and typescript file have the same name
repro: https://github.com/joshcartme/tsgo-duplicate-file-bug

If you have a js and ts file with the same name, `rootNames` in the computed config ends up with a `""` entry. This seems to hang the server, it never starts responding.  For example in the repro posted above I have `lib.ts` and `lib.js`, `lib.js` appears to get resolved as `""` in `rootNames`. 

This may be the same as https://github.com/microsoft/typescript-go/issues/670

Output from the VSCode extension:
```
Resolved to /home/repro/.vscode-server/extensions/typescriptteam.native-preview-0.20250529.1-linux-x64/lib/tsgo
Starting language server...
Info 0    [14:36:15.239] currentDirectory:: /home/repro/development/tsgo-duplicate-file-bug useCaseSensitiveFileNames:: true
Info 1    [14:36:15.240] libs Location:: /home/repro/.vscode-server/extensions/typescriptteam.native-preview-0.20250529.1-linux-x64/lib
Info 2    [14:36:15.244] getConfigFileNameForFile:: File: /home/repro/development/tsgo-duplicate-file-bug/index.ts ProjectRootPath: :: Result: /home/repro/development/tsgo-duplicate-file-bug/tsconfig.json
Info 3    [14:36:15.244] Creating KindConfiguredProject: /home/repro/development/tsgo-duplicate-file-bug/tsconfig.json, currentDirectory: /home/repro/development/tsgo-duplicate-file-bug
Info 4    [14:36:15.244] Config: /home/repro/development/tsgo-duplicate-file-bug/tsconfig.json : {
      "options": {
        "allowJs": true,
        "configFilePath": "/home/repro/development/tsgo-duplicate-file-bug/tsconfig.json"
      },
      "projectReferences": [],
      "rootNames": [
        "/home/repro/development/tsgo-duplicate-file-bug/index.ts",
        "",
        "/home/repro/development/tsgo-duplicate-file-bug/lib.ts"
      ]
    }
```

It's very easy to end up with a js and ts file with the same name if you don't have an `outDir` in your `tsconfig`

jakebailey:
Fixes #978

#978 broke because `removeWildcardFilesWithLowerPriorityExtension` passed the map by value and then attempted to delete from it. That will cause the internal `map` to be modified, but the `key` slice to _not_ be modified in the caller. The fix is to pass by pointer.

To prevent this bug from happening in the future, use the `noCopy` pattern to make `go vet` check it. This reveals another place we weren't passing the map by pointer. That one's not a bug, but it could have been.

I don't have a real test for the problem as I think it's self explanatory and `go vet` will catch this, but I could figure out how to test it if required.

### DIFF
--- a/internal/collections/ordered_map.go
+++ b/internal/collections/ordered_map.go
@@ -18,9 +18,21 @@ import (
 
 // OrderedMap is an insertion ordered map.
 type OrderedMap[K comparable, V any] struct {
+	_    noCopy
 	keys []K
 	mp   map[K]V
 }
+
+// noCopy may be embedded into structs which must not be copied
+// after the first use.
+//
+// See https://golang.org/issues/8005#issuecomment-190753527
+// for details.
+type noCopy struct{}
+
+// Lock is a no-op used by -copylocks checker from `go vet`.
+func (*noCopy) Lock()   {}
+func (*noCopy) Unlock() {}
 
 // NewOrderedMapWithSizeHint creates a new OrderedMap with a hint for the number of elements it will contain.
 func NewOrderedMapWithSizeHint[K comparable, V any](hint int) *OrderedMap[K, V] {
@@ -191,7 +203,7 @@ func (m *OrderedMap[K, V]) clone() OrderedMap[K, V] {
 	}
 }
 
-func (m OrderedMap[K, V]) MarshalJSON() ([]byte, error) {
+func (m *OrderedMap[K, V]) MarshalJSON() ([]byte, error) {
 	if len(m.mp) == 0 {
 		return []byte("{}"), nil
 	}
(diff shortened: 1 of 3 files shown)


sheetalkamat (approved):

internal/collections/ordered_map.go
```diff
@@ -18,10 +18,22 @@ import (
 
 // OrderedMap is an insertion ordered map.
 type OrderedMap[K comparable, V any] struct {
+	_    noCopy
 	keys []K
 	mp   map[K]V
 }
 
+// noCopy may be embedded into structs which must not be copied
```
sheetalkamat, line 26: May be we should do this for orderedSet and syncmap as well ?

Issue #980, opened 2025-05-29
globbing in include isn't working properly under tsgo


joshcartme:
repro: https://github.com/joshcartme/tsgo-glob-bug
using: `7.0.0-dev.20250529.1`

In the above repo I've got a nested `tsconfig.json` that includes some files by having `"../../ambient-types/**/*"` in it's `include`.  This works fine under `tsc` but does not work with `tsgo`.  It does work with `tsgo` if the include is changed to `"../../ambient-types/*"` or `"../../ambient-types"`

tsc:
```
$ yarn run tsc --project lib/core/tsconfig.json
yarn run v1.22.22
$ /home/repro/development/tsgo-glob-bug/node_modules/.bin/tsc --project lib/core/tsconfig.json
Done in 0.59s.
```

tsgo:
```
$ yarn run tsgo --project lib/core/tsconfig.json
yarn run v1.22.22
$ /home/repro/development/tsgo-glob-bug/node_modules/.bin/tsgo --project lib/core/tsconfig.json
lib/core/index.ts:1:14 - error TS2304: Cannot find name '$FixMe'.

1 const fixMe: $FixMe = '';
               ~~~~~~

Found 1 error in lib/core/index.ts:1
```

Using the root `tsconfig.json` which has `ambient-types/**/*` in it's include works find with `tsc` and `tsgo` so it seems to be something about the combination of a relative url that has a glob pattern in it.

joshcartme:
I think I've tracked this down, it is because an include like `../../ambient-types/**/*` is turned into an absolute path, let's say `/home/dev/project/ambient-types/**/*` and passed here https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L241-L252
That strips the last chunk of the path, so returns `/home/dev/project/ambient-types/**`.  Then `patterns` in https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L425 ends up with one of it's base paths as  `/home/dev/project/ambient-types/**` and that is not a valid path to visit in https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L370-L382 so `systemEntries.Files` and `systemEntries.Directories` are empty and any files represented by that include are not found.

[...]

PR #1004, opened 2025-05-30
Strip glob patterns from include basePath

joshcartme:
quick explanation: 
Absolute paths like "/some/path/glob/**" were getting through when the correct path would be "/some/path/glob"

long: 
An include like `../../ambient-types/**/*` is turned into an absolute path, let's say `/home/dev/project/ambient-types/**/*` and passed here https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L241-L252
That strips the last chunk of the path, so returns `/home/dev/project/ambient-types/**`.  Then `patterns` in https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L425 ends up with one of it's base paths as  `/home/dev/project/ambient-types/**` and that is not a valid path to visit in https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L370-L382 so `systemEntries.Files` and `systemEntries.Directories` are empty and any files represented by that include are not found.

Thing do work with implicit globs, so if that include was `../../ambient-types` the `basePath` ends up as `/home/dev/project/ambient-types` and the files and/or directories are found.  

This is attempting to address #980

### DIFF
--- a/internal/vfs/utilities.go
+++ b/internal/vfs/utilities.go
@@ -248,7 +248,7 @@ func getIncludeBasePath(absolute string) string {
 			return tspath.RemoveTrailingDirectorySeparator(tspath.GetDirectoryPath(absolute))
 		}
 	}
-	return absolute[:strings.LastIndex(absolute, string(tspath.DirectorySeparator))]
+	return absolute[:strings.LastIndex(absolute[:wildcardOffset], string(tspath.DirectorySeparator))]
 }
 
 // getBasePaths computes the unique non-wildcard base paths amongst the provided include patterns.
--- a/internal/vfs/utilities.go
+++ b/internal/vfs/utilities.go
@@ -248,7 +248,7 @@ func getIncludeBasePath(absolute string) string {
 			return tspath.RemoveTrailingDirectorySeparator(tspath.GetDirectoryPath(absolute))
 		}
 	}
-	return absolute[:strings.LastIndex(absolute[:wildcardOffset], string(tspath.DirectorySeparator))]
+	return absolute[:max(strings.LastIndex(absolute[:wildcardOffset], string(tspath.DirectorySeparator)), 0)]
 }
 
 // getBasePaths computes the unique non-wildcard base paths amongst the provided include patterns.


internal/vfs/utilities.go
```diff
@@ -248,7 +248,7 @@ func getIncludeBasePath(absolute string) string {
 			return tspath.RemoveTrailingDirectorySeparator(tspath.GetDirectoryPath(absolute))
 		}
 	}
-	return absolute[:strings.LastIndex(absolute, string(tspath.DirectorySeparator))]
+	return absolute[:strings.LastIndex(absolute[:wildcardOffset], string(tspath.DirectorySeparator))]
```
jakebailey: The original code was:

```ts
function getIncludeBasePath(absolute: string): string {
    const wildcardOffset = indexOfAnyCharCode(absolute, wildcardCharCodes);
    if (wildcardOffset < 0) {
        // No "*" or "?" in the path
        return !hasExtension(absolute)
            ? absolute
            : removeTrailingDirectorySeparator(getDirectoryPath(absolute));
    }
    return absolute.substring(0, absolute.lastIndexOf(directorySeparator, wildcardOffset));
}
```

So, this looks correct to me; we need the last index starting at `wildcardOffset`. This slice cuts off the end, so is still a valid index for `absolute` itself.

That being said, you can't negative index a slice in Go, so the thing I'd wonder here is if the `LastIndex` is guaranteed to always succeed.

joshcartme: Hmm yeah, using the debugger I just forced `absolute` to be a string with a `*` but no `/` and it does indeed panic.  But I don't think that could happen in any real scenario because `absolute` is from 
https://github.com/microsoft/typescript-go/blob/e8d352c89b733779275a364d6e5de7df65ac9a42/internal/vfs/utilities.go#L266-L270
and as a real absolute path (as far as I can tell), I think it _must_ start with `/`.

I'm very new to go, so I could be totally wrong and will happily close this or add more safety as you think is necessary.

jakebailey: Yeah, it's possible that it's possible. The safe thing would be to write `min(strings.LastIndex(absolute[:wildcardOffset], string(tspath.DirectorySeparator)), 0)` but, I don't know that it matters.

Sort of surprised no tests change, but I guess we don't have a test for this.

joshcartme: Done, I used `max` rather than `min` as `min` would still give us a negative slice, or always give us `0`

jakebailey: Oh, yes, oops. 😄

jakebailey (approved):

PR #1007, opened 2025-05-31
Add cleanup code for projects and script infos

sheetalkamat:
Changes include:
- Release unused scriptinfos during updateGraph
- Release unused configured projects and script infos after open - this could be a go routine instead that does work after certain time and checking if no new projects are created or something - but for now following same thing as strada - after opening the file - which allows close -open which is a pattern in the 
- Made inferred projects as map for projectRoot to make it easier and use sync maps for configured and inferred projects to ensure concurrency safety 
- Added mutex for containing projects of scriptInfo
- Configured project creation doesnt need rootFiles to have script info (Inferred does because it could be dynamic script info - but may be it doesnt given we dont have wierd project root and current directories to handle - TODO for later investigation)

Verified that memory usage goes down after opening 2-3 projects and then closing all of them and opening new one or file from one of them.  (Currently memory usage is printed on open file in log)


internal/project/scriptinfo.go
```diff
@@ -194,3 +228,11 @@ func (s *ScriptInfo) delayReloadNonMixedContentFile() {
 	s.pendingReloadFromDisk = true
 	s.markContainingProjectsAsDirty()
 }
+
+func (s *ScriptInfo) containedByDeferredClosedProject() bool {
+	s.containingProjectsMu.RLock()
+	defer s.containingProjectsMu.RUnlock()
+	return slices.IndexFunc(s.containingProjects, func(project *Project) bool {
+		return project.deferredClose
+	}) != -1
```
jakebailey: I'd just use `slices.ContainsFunc` here.

internal/project/service.go
```diff
@@ -748,42 +893,52 @@ func (s *Service) createInferredProject(currentDirectory string, projectRootPath
...
 	s.Log(fmt.Sprintf(format, args...))
 }
+
+func (s *Service) printMemoryUsage() {
+	runtime.GC() // Force garbage collection to get accurate memory stats
+	var memStats runtime.MemStats
+	runtime.ReadMemStats(&memStats)
+
+	s.logf("Alloc: %v KB", memStats.Alloc/1024)
+	s.logf("TotalAlloc: %v KB", memStats.TotalAlloc/1024)
```
jakebailey: I would avoid printing this one, as it's the cumulative allocs which is just going to increase and is probably meaningless.

internal/project/service.go
```diff
@@ -748,42 +893,52 @@ func (s *Service) createInferredProject(currentDirectory string, projectRootPath
...
 func (s *Service) logf(format string, args ...any) {
 	s.Log(fmt.Sprintf(format, args...))
 }
+
+func (s *Service) printMemoryUsage() {
+	runtime.GC() // Force garbage collection to get accurate memory stats
+	var memStats runtime.MemStats
+	runtime.ReadMemStats(&memStats)
+
+	s.logf("Alloc: %v KB", memStats.Alloc/1024)
```
jakebailey: Do we want to print these all as one message like we do the "Open files" above?

With the separate calls, we print out many numbered lines:

image

internal/project/service.go
```diff
@@ -613,70 +646,182 @@ func (s *Service) assignProjectToOpenedScriptInfo(info *ScriptInfo) assignProjec
...
+	for _, project := range toRemoveProjects {
+		s.removeProject(project)
+	}
 }
 
-func (s *Service) getOrCreateUnrootedInferredProject() *Project {
-	if s.unrootedInferredProject == nil {
-		s.unrootedInferredProject = s.createInferredProject(s.host.GetCurrentDirectory(), "")
+func (s *Service) removeProject(project *Project) {
+	s.Log("`remove Project:: " + project.name)
```
jakebailey:
```suggestion
	s.Log("remove Project:: " + project.name)
```

internal/project/project.go
```diff
@@ -496,6 +512,7 @@ func (p *Project) updateGraph() bool {
 			for _, oldSourceFile := range oldProgram.GetSourceFiles() {
 				if p.program.GetSourceFileByPath(oldSourceFile.Path()) == nil {
 					p.host.DocumentRegistry().ReleaseDocument(oldSourceFile, oldProgram.Options())
+					p.detachScriptInfoIfNotInferredRoot(oldSourceFile.Path())
```
andrewbranch, line 518: How do you get a scenario where a file disappeared from the program after an update but it’s an inferred project root?

sheetalkamat: Inferred project roots are detached when file is closed => makes it remove as root of inferred project root and detach the script info.
After that when updateGraph happens - program will not have file and old program will. But since scriptInfo is already detached, we should not detach it again.

andrewbranch: In that case, when would this line _not_ result in a no-op for an inferred project?

sheetalkamat: if the file is not the root of inferred project = file through import of the root or some other file

andrewbranch:
How is it that the project service’s maps of projects can be mutated concurrently? I wonder if instead of using sync maps, we should be coordinating project synchronization to happen synchronously, like we do with document-editing requests.

jakebailey:
Yeah, I was a little confused by the introduction of syncmaps there. I mean it's great to do more stuff concurrently, but I wasn't sure if we needed to do that...

sheetalkamat:
I had that before but project creation of one isn’t dependent on another project .. so sync maps suffice in my opinion

sheetalkamat:
Are you saying project create , updates , deletes should’ve on same thread ? That seems unnecessary

andrewbranch:
Presumably a project update would not change the map. I think we’d want different projects to be able to handle updates concurrently. But create/delete seems reasonable to synchronize and hold other requests back for. To me, the introduction of sync maps is not itself a problem, but I _definitely_ want to understand why and how it’s needed, both to evaluate whether it’s worth the complexity and to understand what else needs to be guarded when making future changes.

jakebailey (approved):
Tested it out locally, and things seem to work as expected; I can see it opening new projects, then they disappear after I close files and start typing.

It'd be cool in the future to not gate things to updateGraph and so on, instead have some sort of delay for unused Projects/ScriptInfos to clear them out asynchonrously (since such a thing is now possible).

sheetalkamat:
Yeah that’s a todo for later as I mentioned in the description

andrewbranch (approved):

PR #1013, opened 2025-06-02
Type nodes for JSDoc

sandersn:
In Javascript, you can give a type and modifiers (for classes) to expandos, exports and object literal properties:

```js
/** @type {number} */
module.exports = 12
const o = {
  /** @type {number} */
  p: 12
}
/** @type {number} */
f.p = 12
class C {
  constructor() {
    /** 
     * @private
     * @type {number}
     */
    this.p = 12
  }
}
```

Alternate PR name: Types where they shouldn't be.

image

This PR adds Type nodes to nodes that can only get types from jsdoc. Then it checks those types, replacing the previous hack of reparsing a type assertion. In order to make this work on commonjs exports it starts calling `withJSDoc` on synthetic commonjs nodes. It *also* (sorry), refactors reparser.go into separate functions, and stops cloning type nodes since we're moving to a no-clone reparsing. I'll remove the other 3 clones later, as part of the PR that corrects synthetic-parent walks for `@template`.

This is the first infrastructure/fix PR of the ones I'll spend most of June on. I hope they'll get more coherent but this one sprawls a bit.

#### Nodes with new Type nodes
- PropertyAssignment
- ShorthandPropertyAssignment
- \[JS\]ExportAssignment
- CommonJSExport
- BinaryExpression, which also adds `Modifiers` to support `private` on `this.p` assignments.
These additions are going to cost memory, which we agreed on at the design meeting.

Returns do not add type annotations because I think this is a low-value feature, and TS type checking has no equivalent whatsoever.
#### Notes
To fix contextual typing on commonjs exports, which of course walks up the parent chain, I should also have to introduce a Host property on BinaryExpression, because CommonJSExport is reparsed from BinaryExpression. This was a bridge too far for me, so I cheated: both CommonJSExport and BinaryExpression have the same type annotation, so I let the `@type` double-attach and have special-case code only in `getContextualTypeForBinaryOperand` to recognise the BinaryExpression syntax for `module.exports`.

ShorthandPropertyAssignment type checking was inconsistent in the two places it existed. When I refactored it to a single function, I left the (unannotated) behaviour the same, but it means that the `inDestructuringPattern` argument seems backwards. We should revisit this because I think the consistent behaviour may be right.

I started removing Clones in reparsing, starting with type nodes. This required me to add a synthetic-parent lookup in 2 more places (the existing one is in nameresolver.go). I expect there are a few more, so I created a utility. In order to nudge us toward making `Parent` a method, I gave it the name `getEffectiveTypeParent`. But I'm not sure how hard/how important a `Parent` method is.

The reparser refactor obscures a few changes (sorry!), but not many. The cases in `reparseHosted` for PropertyAssignment, etc just change from `makeNewTypeAssertion` to `makeNewType`.


testdata/baselines/reference/submoduleAccepted/compiler/argumentsReferenceInConstructor3_Js.types.diff
```diff
@@ -30,13 +29,12 @@
  		this.bar = super.arguments.foo;
 ->this.bar = super.arguments.foo : error
 ->this.bar : any
-+>this.bar = super.arguments.foo : object
++>this.bar = super.arguments.foo : any
```
sandersn, line 32: `any` is better than `object` because it shows that the sythetic cast is gone. There's an error, so `error` would be even better though.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsdocTypeTagOnExportAssignment1.errors.txt.diff
```diff
@@ -1,25 +1,11 @@
 --- old.checkJsdocTypeTagOnExportAssignment1.errors.txt
 +++ new.checkJsdocTypeTagOnExportAssignment1.errors.txt
 @@= skipped -0, +0 lines =@@
--a.js(8,18): error TS2353: Object literal may only specify known properties, and 'c' does not exist in type 'Foo'.
--
--
+ a.js(8,18): error TS2353: Object literal may only specify known properties, and 'c' does not exist in type 'Foo'.
+
+
 -==== checkJsdocTypeTagOnExportAssignment1.js (0 errors) ====
```
sandersn: all that's left is a diff in the test framework I think

testdata/baselines/reference/submoduleAccepted/compiler/checkJsdocTypeTagOnExportAssignment3.types.diff
```diff
@@ -1,17 +1,12 @@
...
  === b.js ===
  import a from "./a";
 ->a : import("a").Foo
-+>a : import("./a").Foo
++>a : { c: number; }
 
  a;
 ->a : import("a").Foo
-+>a : import("./a").Foo
++>a : { c: number; }
```
sandersn, line 12: same type printing diff as for all reparsed imports. Needs later investigation.

testdata/baselines/reference/submoduleAccepted/compiler/jsdocTypeNongenericInstantiationAttempt.errors.txt.diff
```diff
@@ -4,120 +4,39 @@
  index.js(2,19): error TS2315: Type 'Boolean' is not generic.
 -index2.js(2,19): error TS2315: Type 'Void' is not generic.
 -index3.js(2,19): error TS2315: Type 'Undefined' is not generic.
-+index.js(2,27): error TS2304: Cannot find name 'T'.
 +index2.js(2,19): error TS2304: Cannot find name 'Void'.
```
sandersn, line 7: this name resolution improvement is a result of not cloning the  type anymore

testdata/baselines/reference/submoduleAccepted/conformance/exportNestedNamespaces2.errors.txt.diff
```diff
@@ -42,4 +44,15 @@
 +!!! error TS2304: Cannot find name 'exports'.
          return v
      }
+     
+-==== use.js (0 errors) ====
++==== use.js (2 errors) ====
+     import * as debug from './mod'
+     
+     debug.formatters.j
++                     ~
```
sandersn, line 53: this is an improvement in imports, it's just that multi-file expandos are not supported in Corsa. I'll send a later PR adding to the list of accepted diffs.

testdata/baselines/reference/submoduleAccepted/conformance/instantiateTemplateTagTypeParameterOnVariableStatement.errors.txt.diff
```diff
@@ -0,0 +1,29 @@
...
++     */
++    const seq = a => b => b;
++    
++    const text1 = "hello";
++    const text2 = "world";
++    
++    /** @type {string} */
++    var text3 = seq(text1)(text2);
++        ~~~~~
++!!! error TS2322: Type 'T' is not assignable to type 'string'.
```
sandersn, line 24: `@template` still needs a fix to type instantation -- I believe it's inspecting parents at some point and failing.

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsFunctionsCjs.errors.txt.diff
```diff
@@ -2,18 +2,25 @@
...
 +index.js(45,23): error TS2580: Cannot find name 'module'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
 +
 +
-+==== index.js (2 errors) ====
++==== index.js (5 errors) ====
 +    module.exports.a = function a() {}
 +    
 +    module.exports.b = function b() {}
 +    module.exports.b.cat = "cat";
++                     ~~~
```
sandersn, line 17: same typing improvement, but nested expandos aren't supported yet

testdata/baselines/reference/submoduleAccepted/conformance/typeTagModuleExports.types.diff
```diff
@@ -1,11 +1,13 @@
 --- old.typeTagModuleExports.types
 +++ new.typeTagModuleExports.types
-@@= skipped -4, +4 lines =@@
+@@= skipped -2, +2 lines =@@
+ === bug27327.js ===
+ /** @type {string} */
  module.exports = 0;
- >module.exports = 0 : string
+->module.exports = 0 : string
++>module.exports = 0 : 0
```
sandersn, line 8: I think this is an improvement, in which the binary expression source of the export ignores the annotation, but usages of it have the annotated type.

internal/ast/ast.go
```diff
@@ -366,7 +366,7 @@ func (n *Node) Expression() *Node {
 	case KindJsxSpreadAttribute:
 		return n.AsJsxSpreadAttribute().Expression
 	}
-	panic("Unhandled case in Node.Expression")
+	panic("Unhandled case in Node.Expression" + n.Kind.String())
```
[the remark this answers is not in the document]
sandersn: Yeah, good idea.

testdata/baselines/reference/submoduleAccepted/compiler/checkJsdocTypeTagOnExportAssignment1.errors.txt.diff
jakebailey: https://github.com/microsoft/typescript-go/pull/1014 will fix this.

internal/binder/binder.go
```diff
@@ -1649,7 +1649,7 @@ func (b *Binder) bindChildren(node *ast.Node) {
 		b.inAssignmentPattern = saveInAssignmentPattern
 		b.bindEachChild(node)
 	case ast.KindJSExportAssignment, ast.KindCommonJSExport:
-		return // Reparsed nodes do not double-bind children, which are not reparsed
+		// Reparsed nodes do not double-bind children, which are not reparsed
```
sandersn, line 1656: I think this is an improvement--now setJSDocParents runs and inAssignmentPattern correctly follows dynamic scope

internal/checker/checker.go
```diff
@@ -15352,11 +15374,15 @@ func (c *Checker) getTypeOfVariableOrParameterOrPropertyWorker(symbol *ast.Symbo
 	case ast.KindPropertyAssignment:
 		result = c.checkPropertyAssignment(declaration, CheckModeNormal)
 	case ast.KindShorthandPropertyAssignment:
-		result = c.checkExpressionForMutableLocation(declaration.Name(), CheckModeNormal)
+		result = c.checkShorthandPropertyAssignment(declaration, true /*inDestructuringPattern*/, CheckModeNormal)
```
sandersn: this isn't necessarily in a destructuring pattern, but true=never use the object assigmebt initialiser. Sometimes this shows up outside a destructuring though! I've passed true here to keep the behaviour the same but we should check this out in more detail.

internal/parser/jsdoc.go
```diff
@@ -60,7 +60,9 @@ func (p *Parser) withJSDoc(node *ast.Node, hasJSDoc bool) {
 			p.reparseTags(node, jsdoc)
 		}
 		p.jsdocCache[node] = jsdoc
+		return jsdoc
 	}
```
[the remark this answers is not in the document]
sandersn: No. jsdocCache is write-only, write-once here by the nature of the parser. The current bode's jsdoc will never be cached.

internal/parser/reparser.go
```diff
@@ -268,13 +292,10 @@ func (p *Parser) makeNewType(typeExpression *ast.TypeNode, host *ast.Node) *ast.
 	}
 	if typeExpression.AsJSDocTypeExpression().Host == nil {
 		typeExpression.AsJSDocTypeExpression().Host = host
-	} else {
+	} else if host.Kind != ast.KindJSExportAssignment && host.Kind != ast.KindCommonJSExport {
 		panic("JSDoc type expression already has a host: " + typeExpression.AsJSDocTypeExpression().Host.Kind.String())
 	}
-	t := typeExpression.Type().Clone(&p.factory)
+	t := typeExpression.Type()
```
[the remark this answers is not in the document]
sandersn: That's what the panic is for! The JSExportAssignment/CommonJSExport exemption avoids a Host pointer on BinaryExpression, at the cost of special handling later.

testdata/baselines/reference/submodule/conformance/typeFromContextualThisType.symbols.diff
```diff
@@ -6,7 +6,7 @@
  >this.b : Symbol(b, Decl(bug25926.js, 0, 23))
 ->this : Symbol(__type, Decl(bug25926.js, 0, 11))
 ->b : Symbol(b, Decl(bug25926.js, 2, 9))
-+>this : Symbol(o1, Decl(bug25926.js, 0, 11))
++>this : Symbol(�type, Decl(bug25926.js, 0, 11))
```
jakebailey: This is getting some sort of goofy symbol name, it seems.

jakebailey:
> In Javascript, you can give a type and modifiers (for classes) to expandos, exports and object literal properties:

And functions 😭

sandersn:
I know, I know. But the function annotation code is *so* complicated. It was a bug farm for years in Strada.

testdata/baselines/reference/submodule/conformance/typeFromContextualThisType.symbols.diff
sandersn: this is the usual string for the intenral name of types. Does it not normally print?
Edit: It prints in Strada, whereit's `__type` not `th'type`.

sandersn:
Agreed. I think theres a change needed in outer type parameter code in the checker.

PR #1031, opened 2025-06-04
Fix more parent races

andrewbranch:
#970 was incomplete because I misread `SetParentsInChildren` as not setting the parent if the node already has one.

### DIFF
--- a/internal/ast/utilities.go
+++ b/internal/ast/utilities.go
@@ -885,7 +885,8 @@ func newParentInChildrenSetter() func(node *Node) bool {
 	}
 
 	state.visit = func(node *Node) bool {
-		if state.parent != nil {
+		if state.parent != nil && node.Parent != state.parent {
+			// Avoid data races on no-ops
 			node.Parent = state.parent
 		}
 		saveParent := state.parent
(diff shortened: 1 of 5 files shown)


internal/ast/utilities.go
```diff
@@ -885,7 +885,8 @@ func newParentInChildrenSetter() func(node *Node) bool {
 	}
 
 	state.visit = func(node *Node) bool {
-		if state.parent != nil {
+		if state.parent != nil && node.Parent != state.parent {
```
jakebailey, line 888: Does this actually avoid a race? A write can happen somewhere other than this, and then it's a read-write race.

andrewbranch: The only other concurrent write is what I protected in #970

andrewbranch: And to be clear, it’s always been a read-write race. https://github.com/microsoft/typescript-go/actions/runs/15429131343/job/43423235122

jakebailey: Right, but then this part of the code doesn't need to change, since if there were a race, it'd be a race either way, right? We're just incurring an extra read before the write

jakebailey: Ah, I understand. If someone is reading the parents assuming they're set up, and then this code comes along, it would today unconditionally write even though it even though it's already done. Gotcha.

jakebailey:
Tests fail, fun

andrewbranch:
The baseline change is actually moving in the right direction—before, module resolution failed completely. Now it succeeds, which reveals another problem that I assume is unrelated to this change. In JSDoc, you’re allowed to refer to values as types (we insert an implicit `typeof`). It looks like that’s either not working here or not yet been set up at all (cc @sandersn).

jakebailey (approved):

PR #1048, opened 2025-06-04
Fix [checker]: Added missing callback fields to createNameResolverForSuggestion

DrHazemAli:
Greeting,
I hope you’re all doing great! 
I was digging through the code today and stumbled upon something worth sharing. the `createNameResolverForSuggestion()` function was missing a couple of callbacks that its sibling, `createNameResolver()`, already has. In short, that means when TypeScript tried to offer those handy "Did you mean…?" suggestions, it wasn’t running the same checks it does for regular symbol resolution. As a result, users could end up seeing suggestions that would actually break their code if they picked them.

## Provided Fix
Added the missing callbacks:
- `OnPropertyWithInvalidInitializer`
- `OnSuccessfullyResolvedSymbol`

Now both functions are consistent and suggestions get properly validated.

##  Why it matters
Before: Suggestions might recommend symbols that would actually cause errors
After: Suggestions are validated the same way as normal code, giving users better recommendations

## What changed

```diff
 func (c *Checker) createNameResolverForSuggestion() *binder.NameResolver {
     return &binder.NameResolver{
         // ... existing fields ...
+        OnPropertyWithInvalidInitializer: c.checkAndReportErrorForInvalidInitializer,
+        OnFailedToResolveSymbol:          c.onFailedToResolveSymbol,
+        OnSuccessfullyResolvedSymbol:     c.onSuccessfullyResolvedSymbol,
     }
 }
```

Simple fix for better TypeScript 😅


DrHazemAli:
@jakebailey can you look into this?
Thanks in advance 🙏

jakebailey:
No tests change, so I'm not exactly sure how to evaluate this.

DrHazemAli:
Hey @jakebailey, thanks for looking into this

This change simply adds two callbacks `OnPropertyWithInvalidInitializer` and `OnSuccessfullyResolvedSymbol` so that suggestion-only name resolution runs through the `same validation checks` as normal symbol resolution. 

Since it doesn’t alter existing behavior or introduce any new public APIs, the current tests already cover it (they’d break if suggestion validation ever regressed). That’s why you don’t see any test files here. 😅

If you’d rather have an explicit test to prove everything is wired up correctly, I can add a minimal unit test that ensures `createNameResolverForSuggestion()` fires those callbacks when it sees invalid symbols. 

Please let me know if that sounds helpful for you.
Cheers 🙏

jakebailey:
The original TS code this was ported from said:

```ts
    var resolveName = createNameResolver({
        compilerOptions,
        requireSymbol,
        argumentsSymbol,
        globals,
        getSymbolOfDeclaration,
        error,
        getRequiresScopeChangeCache,
        setRequiresScopeChangeCache,
        lookup: getSymbol,
        onPropertyWithInvalidInitializer: checkAndReportErrorForInvalidInitializer,
        onFailedToResolveSymbol,
        onSuccessfullyResolvedSymbol,
    });
```

```ts
    var resolveNameForSymbolSuggestion = createNameResolver({
        compilerOptions,
        requireSymbol,
        argumentsSymbol,
        globals,
        getSymbolOfDeclaration,
        error,
        getRequiresScopeChangeCache,
        setRequiresScopeChangeCache,
        lookup: getSuggestionForSymbolNameLookup,
    });
```

Note that the the three things you've added are explicitly _not_ on the resolver for suggestions.

So, I do not think this PR is correct.

(IIRC this resolver needs to _not_ have side effects like noting failures, otherwise the editor will cause random other breakages.)

DrHazemAli:
Thank you for your valuable time reviewing this PR and your clarification.
You can close it, or let me know if you’d like me to draft it. 🙏

jakebailey:
I'm going to close this, sorry. I don't think it's correct.

PR #1049, opened 2025-06-04
Remove unnecessary compilerOptions and use directly from config

sheetalkamat:
Refactor after #999

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -299,12 +299,12 @@ func (p *Program) GetConfigFileParsingDiagnostics() []*ast.Diagnostic {
 }
 
 func (p *Program) singleThreaded() bool {
-	return p.programOptions.SingleThreaded.DefaultIfUnknown(p.programOptions.Config.CompilerOptions().SingleThreaded).IsTrue()
+	return p.programOptions.SingleThreaded.DefaultIfUnknown(p.Options().SingleThreaded).IsTrue()
 }
 
 func (p *Program) getSourceAffectingCompilerOptions() *core.SourceFileAffectingCompilerOptions {
 	p.sourceAffectingCompilerOptionsOnce.Do(func() {
-		p.sourceAffectingCompilerOptions = p.programOptions.Config.CompilerOptions().SourceFileAffecting()
+		p.sourceAffectingCompilerOptions = p.Options().SourceFileAffecting()
 	})
 	return p.sourceAffectingCompilerOptions
 }
@@ -427,7 +427,7 @@ func (p *Program) getBindDiagnosticsForFile(ctx context.Context, sourceFile *ast
 }
 
 func (p *Program) getSemanticDiagnosticsForFile(ctx context.Context, sourceFile *ast.SourceFile) []*ast.Diagnostic {
-	compilerOptions := p.programOptions.Config.CompilerOptions()
+	compilerOptions := p.Options()
 	if checker.SkipTypeChecking(sourceFile, compilerOptions) {
 		return nil
 	}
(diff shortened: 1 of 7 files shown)


internal/compiler/program.go
```diff
@@ -169,8 +168,8 @@ func (p *Program) GetSourceFileFromReference(origin *ast.SourceFile, ref *ast.Fi
 func NewProgram(options ProgramOptions) *Program {
 	p := &Program{}
 	p.programOptions = options
-	p.compilerOptions = options.Config.CompilerOptions()
-	if p.compilerOptions == nil {
+	compilerOptions := p.programOptions.Config.CompilerOptions()
```
jakebailey: Is there a reason to move the object off of Program? It sems like we'd access it all the time.

But, maybe this call is static enough to be inlined everywhere anyway.

internal/compiler/program.go
```diff
@@ -294,19 +292,19 @@ func equalCheckJSDirectives(d1 *ast.CheckJsDirective, d2 *ast.CheckJsDirective)
...
 
 func (p *Program) singleThreaded() bool {
-	return p.programOptions.SingleThreaded.DefaultIfUnknown(p.compilerOptions.SingleThreaded).IsTrue()
+	return p.programOptions.SingleThreaded.DefaultIfUnknown(p.programOptions.Config.CompilerOptions().SingleThreaded).IsTrue()
 }
 
 func (p *Program) getSourceAffectingCompilerOptions() *core.SourceFileAffectingCompilerOptions {
 	p.sourceAffectingCompilerOptionsOnce.Do(func() {
-		p.sourceAffectingCompilerOptions = p.compilerOptions.SourceFileAffecting()
+		p.sourceAffectingCompilerOptions = p.programOptions.Config.CompilerOptions().SourceFileAffecting()
```
jakebailey: For all of these I would probably consider just writing `p.Options()` rather than inlining it everywhere.

internal/compiler/program.go
sheetalkamat: just wanted to avoid having info in two places and consolidate this with the usage of config- and per offline discussion seemed like compilerOptions on program and in  programoptions wasnt intentional.

jakebailey: Ah, so what I was trying to say in the last thread was that it's unintentional that we're storing `ProgramOptions` itself instead of just copying stuff directly into the `Program`, not that we should be using `ProgramOptions` _more_. Sorry if it sounded like the opposite.

sheetalkamat: ok i will do other way around then .. i would still like to keep things as "Config" instead of copying compilerOptions off of it

sheetalkamat: Btw storing program options as is nicer than having to copy all the data

jakebailey: What's the rationale for keeping it there?

sheetalkamat: Also having programOptions is needed for `UpdateProgram` so i think we shouldnt get rid of it. we should instead remove the duplicate stuff on the program instead

jakebailey: Sure, that's fine; we can keep it that way.

sheetalkamat: > What's the rationale for keeping it there?

https://github.com/microsoft/typescript-go/blob/main/internal/compiler/program.go#L229  this is nice and easier to use than having to copy

internal/compiler/program.go
```diff
@@ -670,23 +669,23 @@ func (p *Program) GetSourceFileMetaData(path tspath.Path) *ast.SourceFileMetaDat
...
 
 func (p *Program) GetModeForUsageLocation(sourceFile *ast.SourceFile, location *ast.Node) core.ResolutionMode {
-	return getModeForUsageLocation(sourceFile, p.sourceFileMetaDatas[sourceFile.Path()], location, p.compilerOptions)
+	return getModeForUsageLocation(sourceFile, p.sourceFileMetaDatas[sourceFile.Path()], location, p.programOptions.Config.CompilerOptions())
 }
 
 func (p *Program) GetDefaultResolutionModeForFile(sourceFile modulespecifiers.SourceFileForSpecifierGeneration) core.ResolutionMode {
-	return getDefaultResolutionModeForFile(sourceFile, p.sourceFileMetaDatas[sourceFile.Path()], p.compilerOptions)
+	return getDefaultResolutionModeForFile(sourceFile, p.sourceFileMetaDatas[sourceFile.Path()], p.programOptions.Config.CompilerOptions())
 }
```
jakebailey: Seems like a couple of these are still missed.

jakebailey (approved):

PR #1051, opened 2025-06-04
More module specifier generation

andrewbranch:
Follow-up to #851 and #791 eliminating more diffs


testdata/baselines/reference/submodule/compiler/declarationEmitDoesNotUseReexportedNamespaceAsLocal.types.diff
```diff
@@ -15,7 +15,7 @@
 
  export * as Q from "./sub";
 ->Q : typeof import("sub")
-+>Q : typeof import("./sub.js")
++>Q : typeof import("./sub")
```
andrewbranch, line 18: I think this is something that @weswigham mentioned, that type baselines in Strada did not use real module specifier generation, but they (intentionally?) do in Corsa. That significantly limits the number of diffs that will get completely deleted. IMO, it would be nice to align them and see a pure diff deletion, whether by changing Corsa or changing Strada now that we have a target `tsgo-port` branch.

internal/checker/checker.go
```diff
@@ -14366,7 +14365,7 @@ func (c *Checker) resolveExternalModule(location *ast.Node, moduleReference stri
 
 	var sourceFile *ast.SourceFile
 	resolvedModule := c.program.GetResolvedModule(importingSourceFile, moduleReference, mode)
-	if resolvedModule != nil && resolvedModule.IsResolved() {
+	if resolvedModule.IsResolved() {
```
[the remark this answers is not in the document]
andrewbranch: No it may not

internal/transformers/impliedmodule.go
```diff
@@ -42,20 +42,15 @@ func (tx *ImpliedModuleTransformer) visitSourceFile(node *ast.SourceFile) *ast.N
...
 		}
 		transformer = tx.cjsTransformer
 	}
 
 	return transformer.TransformSourceFile(node).AsNode()
 }
-
-func (tx *ImpliedModuleTransformer) getEmitModuleFormatOfFile(node *ast.SourceFile) core.ModuleKind {
-	// !!! host.getEmitModuleFormatOfFile?
-	return ast.GetEmitModuleFormatOfFileWorker(node, tx.compilerOptions, tx.sourceFileMetaDataProvider.GetSourceFileMetaData(node.Path()))
```
jakebailey, line 60: This (and similar dupes) really bugged me when I was trying to do the module indicator 👍

jakebailey (approved):

testdata/baselines/reference/submodule/compiler/declarationEmitDoesNotUseReexportedNamespaceAsLocal.types.diff
jakebailey: Do you know where that might be? Happy to fix that and push it up.

internal/modulespecifiers/specifiers.go
```diff
@@ -288,29 +289,39 @@ func computeModuleSpecifiers(
...
-	//         },
-	//     ));
-	// if (existingSpecifier) {
-	//     return { kind: undefined, moduleSpecifiers: [existingSpecifier], computedWithoutCache: true };
-	// }
+	var existingSpecifier string
+	for _, modulePath := range modulePaths {
+		targetPath := tspath.ToPath(modulePath.FileName, host.GetCurrentDirectory(), info.UseCaseSensitiveFileNames)
+		var existingImport *ast.StringLiteralLike
+		for _, importSpecifier := range importingSourceFile.Imports() {
```
andrewbranch, line 296: In Strada, this code uses program file include reasons instead of `file.imports`, which I found very puzzling. I changed it in Strada and the only tests that fail were ones that use `--outFile`. (The existing imports it finds are synthesized ambient module imports inside ambient module declarations in the generated .d.ts bundle. Since those are attached to module declaration blocks instead of a source file, `importingSourceFile.Imports()` doesn’t find them. This code is supposed to be an optimization, so missing that edge case seems fine.)

testdata/baselines/reference/submodule/compiler/declarationEmitDoesNotUseReexportedNamespaceAsLocal.types.diff
weswigham: The `typeToTypeNode` call in `typeWriter.ts` provides no symbol tracker, and thus no module resolution host. That's fine, but in such a case, the `createBasicNodeBuilderModuleSpecifierResolutionHost` call inside `checker` that makes a fallback only partially wraps the `host` `program`, and only when `InternalNodeBuilderFlags.DoNotIncludeSymbolChain` is passed (which, it is not, in this case). Make that fallback unconditional, and make the wrapper more complete, and you should match the new Corsa behavior.

internal/modulespecifiers/specifiers.go
weswigham: I dunno. I'd have guessed something like "so it works for files which weren't assembled by the default program, since `.imports` is set by the program and not the parser", but in such a scenario, I kinda doubt the host would still provide file include reasons. I guess it *could* happen...?

PR #903, opened 2025-05-21
Block completions in invalid positions

gabritto:
Ports the code that blocks completions in positions such as inside strings when the possible strings aren't constrained by the context.
Completions not yet blocked inside comments because `isInComment` is not ported yet (it lives in the formatter 😬)

PR #1052, opened 2025-06-04
Port formatter, enable in LSP

weswigham:
This isn't extensively tested yet, as the fourslash tests aughta do that once they're in, but I've got a test/bench running the formatter over `checker.ts` just to check that it doesn't crash, at least.


internal/core/text.go
```diff
@@ -54,3 +54,13 @@ func (t TextRange) WithPos(pos int) TextRange {
 func (t TextRange) WithEnd(end int) TextRange {
 	return TextRange{pos: t.pos, end: TextPos(end)}
 }
+
+func (t TextRange) ContainedBy(t2 TextRange) bool {
+	return t2.pos <= t.pos && t2.end >= t.end
+}
+
+func (t TextRange) Overlaps(t2 TextRange) bool {
```
jakebailey, line 62: I think some of these may have just been added in #903, maybe?

internal/core/textchange.go
```diff
@@ -0,0 +1,10 @@
+package core
+
+type TextChange struct {
+	TextRange
+	NewText string
+}
+
+func (t TextChange) ApplyTo(text string) string {
+	return text[:t.Pos()] + t.NewText + text[t.End():]
```
jakebailey, line 9: Yeah, in JS this is fast because ropes, but in Go this is super super slow. I think we're going to have to batch these. I haven't read onward from here but hopefully the formatter doesn't heavily depend on being able to repeatedly apply edits...

internal/format/api.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	// get start position for the previous line
+	startPos := int(scanner.GetLineStarts(sourceFile)[line-1])
+	// After the enter key, the cursor is now at a new line. The new line may or may not contain non-whitespace characters.
+	// If the new line has only whitespaces, we won't want to format this line, because that would remove the indentation as
+	// trailing whitespaces. So the end of the formatting span should be the later one between:
+	//  1. the end of the previous line
+	//  2. the last non-whitespace character in the current line
+	endOfFormatSpan := scanner.GetEndLinePosition(sourceFile, line)
+	for endOfFormatSpan > startPos {
```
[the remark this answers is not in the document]
weswigham: The thumbs-down button doesn't work (that can't be good), so I'll just say: `scanner.GetEndLinePosition` is guaranteed to return a position within the file's text, and `startPos` is guaranteed to be zero or positive, so the bounds are already checked.

internal/core/textchange.go
weswigham: Only testing does. Nothing we actually do applies edits - we leave that to the editor. (Also notable - I just moved this; it's already in the code)

internal/format/api.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+type formatContextKey int
+
+const (
+	formatOptionsKey formatContextKey = iota
+	formatNewlineKey
+)
+
+func NewContext(ctx context.Context, options *FormatCodeSettings, newLine string) context.Context {
```
jakebailey: Is using context here because we're going to cancel something? Or just out of convenience?

(canonically these funcs are called something like `WithFormattingOptions` plus typesafe methods to extract the value).

internal/format/api.go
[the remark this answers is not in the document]
jakebailey: you have to refresh the page to click the button half the time ☹️

internal/core/text.go
weswigham: I only see `ContainsExclusive` - which isn't either of these.

internal/format/rule.go
```diff
@@ -0,0 +1,109 @@
+package format
+
+import "github.com/microsoft/typescript-go/internal/ast"
+
+type Rule struct {
```
jakebailey: Very optional but I suspect a lot of these types don't have to be exported, given the rules are sort of internal details

internal/lsp/server.go
```diff
@@ -611,6 +627,87 @@ func (s *Server) handleCompletion(ctx context.Context, req *lsproto.RequestMessa
...
 }
 
+func (s *Server) handleDocumentFormat(ctx context.Context, req *lsproto.RequestMessage) error {
+	params := req.Params.(*lsproto.DocumentFormattingParams)
+	project := s.projectService.EnsureDefaultProjectForURI(params.TextDocument.Uri)
+	languageService, done := project.GetLanguageServiceForRequest(ctx)
+	defer done()
+	// !!! remove this after formatting is fully ported/tested
+	defer func() {
+		if r := recover(); r != nil {
```
jakebailey, line 637: Not in the slightest needed to be done in this PR, but I've been wanting to apply this to all "non critical" requests

internal/format/indent.go
```diff
@@ -0,0 +1,554 @@
...
+		}
+		return getList((node).ParameterList(), r, node, sourceFile)
+	case ast.KindGetAccessor:
+		return getList((node).ParameterList(), r, node, sourceFile)
+	case ast.KindClassDeclaration,
+		ast.KindClassExpression,
+		ast.KindInterfaceDeclaration,
+		ast.KindTypeAliasDeclaration,
+		ast.KindJSDocTemplateTag:
+		return getList((node).TypeParameterList(), r, node, sourceFile)
```
jakebailey: Seems like there are some extra parens leftover in some of these (maybe code gen artifacts?)

internal/format/rulecontext.go
```diff
@@ -0,0 +1,644 @@
...
+func isOptionEnabledOrUndefined(optionName optionSelector) ContextPredicate {
+	return func(context *formattingContext) bool {
+		if context.Options == nil {
+			return true
+		}
+		return optionName(context.Options).IsTrueOrUnknown()
+	}
+}
+
+func isForContext(context *formattingContext) bool {
```
jakebailey, line 153: Nit but I usually avoid the name "context" given its conflict with the package name, but it's probably fine.

internal/format/rulecontext.go
```diff
@@ -0,0 +1,644 @@
...
+	return false
+}
+
+func isFunctionDeclContext(context *formattingContext) bool {
+	switch context.contextNode.Kind {
+	case ast.KindFunctionDeclaration,
+		ast.KindMethodDeclaration,
+		ast.KindMethodSignature:
+		// case ast.KindMemberFunctionDeclaration:
+		fallthrough
```
jakebailey, line 295: Is this function intentionally like this? Everything falls through.

internal/format/rules.go
```diff
@@ -0,0 +1,446 @@
...
+		rule("SpaceBeforeOpenBraceInFunction", functionOpenBraceLeftTokenRange, ast.KindOpenBraceToken, []ContextPredicate{isOptionDisabledOrUndefinedOrTokensOnSameLine(placeOpenBraceOnNewLineForFunctionsOption), isFunctionDeclContext, isBeforeBlockContext, isNotFormatOnEnter, isSameLineTokenOrBeforeBlockContext}, RuleActionInsertSpace, RuleFlagsCanDeleteNewLines),
+		rule("SpaceBeforeOpenBraceInTypeScriptDeclWithBlock", typeScriptOpenBraceLeftTokenRange, ast.KindOpenBraceToken, []ContextPredicate{isOptionDisabledOrUndefinedOrTokensOnSameLine(placeOpenBraceOnNewLineForFunctionsOption), isTypeScriptDeclWithBlockContext, isNotFormatOnEnter, isSameLineTokenOrBeforeBlockContext}, RuleActionInsertSpace, RuleFlagsCanDeleteNewLines),
+
+		rule("NoSpaceBeforeComma", anyToken, ast.KindCommaToken, []ContextPredicate{isNonJsxSameLineTokenContext}, RuleActionDeleteSpace),
+
+		// No space before and after indexer `x[]ast.Kind{}`
+		rule("NoSpaceBeforeOpenBracket", anyTokenExcept(ast.KindAsyncKeyword, ast.KindCaseKeyword), ast.KindOpenBracketToken, []ContextPredicate{isNonJsxSameLineTokenContext}, RuleActionDeleteSpace),
```
jakebailey: I'm not sure whether to be happy or sad to see that this pattern still works even when ported 😄

internal/format/api.go
weswigham: We *should* be checking the cancellation token in the formatter. It doesn't yet, though (and doesn't in strada). My thinking was that the formatting options need to be passed around by multiple APIs (anything using the formatter aughta set/provide them, such as quickfixes), so it's easier to keep them in the context than flow them everywhere (in strada, they're managed by the project service).

internal/format/indent.go
weswigham: Honestly surprised the formatter doesn't remove these.

internal/format/rulecontext.go
weswigham: I left it structurally the same as in Strada - where some blocks have comments and explicit `fallthrough`s, rather than just one big block. Don't *really* know why, other than to logically group the `kind`s.

internal/format/rules.go
weswigham: The only thing that needed any modification was the option-dependent context checks, which all had to be swapped from dynamic property lookup to hardcoded selectors.

jakebailey:
One thing that I think we'll have to document is that formatting won't work by default unless users set something like:

```json
"editor.defaultFormatter": "TypeScriptTeam.native-preview"
```

Formatting is a request that can only be handled by one provider at a time, so it has to know which one to choose, and by default that is surely the built-in TS extension, which will fail.

jakebailey:
Though, maybe we can abuse:

```jsonc
        "configurationDefaults": {
            "[typescript]": {
                "editor.defaultFormatter": "TypeScriptTeam.native-preview"
            },
            "[typescriptreact]": {
                "editor.defaultFormatter": "TypeScriptTeam.native-preview"
            },
            "[javascript]": {
                "editor.defaultFormatter": "TypeScriptTeam.native-preview"
            },
            "[javascriptreact]": {
                "editor.defaultFormatter": "TypeScriptTeam.native-preview"
            }
        },
```

Though ideally this would only be the case if the LSP were running. Not sure if these can be conditioned like other package.json level things.

weswigham:
> One thing that I think we'll have to document is that formatting won't work by default unless users set something like:

Do they? With the native preview enabled, before I added the capability, I got a "no formatting provider" error, and then after I added the capability to the server, it "just worked". Dunno if that's an artifact of the extension dev mode, though.

jakebailey:
Ah, you know, maybe having the tsgo option enabled means the builtin TS extension never registers the format provider, and therefore the format request comes to us naturally. That seems fine!

weswigham:
@jakebailey With the bulk edit application in place (so `go` isn't making hundreds of thousands of multi-megabyte strings), https://github.com/microsoft/typescript-go/pull/1052/commits/03015d602c3061a5fccd573a3a66f89fc3f9a152 removes another source of memory pressure - that change should be good even outside the formatter, too, no?

Even after https://github.com/microsoft/typescript-go/pull/1052/commits/24bbe7d54036ca4403c1e7677ca4893c0880c5b1 and  https://github.com/microsoft/typescript-go/pull/1052/commits/2d519b3e60852fd6e6a45d35b711f0f6b32fd6ab, formatting `checker.ts` still takes around 2x longer than just pretty printing it via the printer (not that the printer is a complete replacement at present, since it discards a bunch of text and lacks configuration), but I expect that gap can be narrowed further - in theory the formatter really aughta be pretty close, since it's not building any strings itself (just a list of edits), and is a pretty similar AST visit. Unless I'm reading it wrong, though, the majority of the time is now spent in scanner util functions like utf8 rune decoding, counting, and position calculation (things the pretty printer generally doesn't need to bother with, since it's not rescanning and checking adjacent tokens for whitespace rules).

jakebailey:
I assume the 2x is not "2x slower than Strada", but rather 2x slower than Corsa emit?

I haven't profiled it myself yet, but I think that as long as it works and doesn't crash, it's fine to leave like this for future PRs. This all gets to run asynchronously too.

jakebailey:
IIRC formatting checker.ts in Strada took way more than a couple seconds, so 😄

weswigham:
> but rather 2x slower than Corsa emit?

Yeah.

> IIRC formatting checker.ts in Strada took way more than a couple seconds, so 😄

Well, now it's 0.21 sec to format `checker.ts` (when it's already available parsed) on my machine (vs 0.1s to emit). It was only in the "seconds" order of magnitude with unoptimized edit application included (which there is now a helper to avoid, so).

PR #1053, opened 2025-06-04
Port JSX fragment checking changes

Andarist:
ports https://github.com/microsoft/TypeScript/pull/59933
ports https://github.com/microsoft/TypeScript/pull/60122
ports https://github.com/microsoft/TypeScript/pull/60615
ports https://github.com/microsoft/TypeScript/pull/60687


internal/checker/checker.go
```diff
@@ -26921,10 +26926,15 @@ func (c *Checker) markJsxAliasReferenced(node *ast.Node /*JsxOpeningLikeElement
...
 	}
-	// allow null as jsxFragmentFactory
+	shouldFactoryRefErr := c.compilerOptions.Jsx != core.JsxEmitPreserve && c.compilerOptions.Jsx != core.JsxEmitReactNative
+	// #38720/60122, allow null as jsxFragmentFactory
 	var jsxFactorySym *ast.Symbol
 	if !(ast.IsJsxOpeningFragment(node) && jsxFactoryNamespace == "null") {
-		jsxFactorySym = c.resolveName(jsxFactoryLocation, jsxFactoryNamespace, ast.SymbolFlagsValue, jsxFactoryRefErr, true /*isUse*/, false /*excludeGlobals*/)
+		flags := ast.SymbolFlagsValue
+		if !shouldFactoryRefErr {
+			flags &= ^ast.SymbolFlagsEnum
```
jakebailey: `&= ^...` should be `&^=` in Go. (same in other new places)

Andarist: fixed

Andarist: actually, this isn't true about the compound assignment like this - after applying the proposed change format job failed

jakebailey: It was just mad because you didn't write `x &^= y`, but instead wrote `x&^=y`. Just needed a `dprint fmt`.

Andarist: I ran the format task. I can also see other ` &= ^` instances in the codebase 🤔

jakebailey: They're equivalent, it's just that the Go operator is nicer to read as "AND NOT equal".

PR #1059, opened 2025-06-05
Port "Error on too many parameters for iterator method"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/60321


testdata/baselines/reference/submodule/compiler/iteratorExtraParameters.errors.txt.diff
Andarist, line 1: this is still different for 2 reasons:
- diagnostics/error output containers don't support `skipLogging` (yet?)
- there is no logic to ignore cached values for `IterationTypes{}` (aka `noIterationTypes`) and thus repeated errors are not reported. I'll fix this in a followup PR

internal/checker/checker.go
```diff
@@ -6047,10 +6052,17 @@ func (c *Checker) getIterationTypesOfIterableSlow(t *Type, r *IterationTypesReso
 		if IsTypeAny(methodType) {
 			return IterationTypes{c.anyType, c.anyType, c.anyType}
 		}
-		if signatures := c.getSignaturesOfType(methodType, SignatureKindCall); len(signatures) != 0 {
-			iteratorType := c.getIntersectionType(core.Map(signatures, c.getReturnTypeOfSignature))
+		allSignatures := c.getSignaturesOfType(methodType, SignatureKindCall)
+		validSignatures := core.Filter(allSignatures, func(sig *Signature) bool {
+			return c.getMinArgumentCount(sig) == 0
+		})
+		if len(validSignatures) != 0 {
```
jakebailey: It feels suspicious that this condition is flipped from the original code...

Andarist: It's just because the actual branches here are flipped, this allowed me to nearly fall through to the `IterationTypes{}` case in the case when valid signatures are missing

testdata/baselines/reference/submodule/compiler/iteratorExtraParameters.errors.txt.diff
jakebailey: For the first thing, I'm pretty sure this is a place where we're differing because the code wasn't ported but redone from scratch. So I don't think that's a simple fix.

jakebailey: I think the PR needs a rebase after #1061, yes?

Andarist: done

jakebailey (approved):
Seems correct modulo the larger errors.

PR #1061, opened 2025-06-05
Report iteration errors on all error nodes

Andarist:


internal/checker/checker.go
```diff
@@ -5882,18 +5881,26 @@ func (c *Checker) getIterationTypesOfIterable(t *Type, use IterationUse, errorNo
 	if IsTypeAny(t) {
 		return IterationTypes{c.anyType, c.anyType, c.anyType}
 	}
-	key := IterationTypesKey{typeId: t.id, use: use&IterationUseCacheFlags | core.IfElse(errorNode != nil, IterationUseReportError, 0)}
+	key := IterationTypesKey{typeId: t.id, use: use&IterationUseCacheFlags}
+	// If we are reporting errors and encounter a cached `noIterationTypes`, we should ignore the cached value and continue as if nothing was cached.
+	// In addition, we should not cache any new results for this call.
+	noCache := false
```
Andarist: alternatively, I could use `nodeId` as part of the `IterationTypesKey` to get the same results

internal/checker/checker.go
```diff
@@ -5882,18 +5881,26 @@ func (c *Checker) getIterationTypesOfIterable(t *Type, use IterationUse, errorNo
 	if IsTypeAny(t) {
 		return IterationTypes{c.anyType, c.anyType, c.anyType}
 	}
-	key := IterationTypesKey{typeId: t.id, use: use&IterationUseCacheFlags | core.IfElse(errorNode != nil, IterationUseReportError, 0)}
```
Andarist: was this even concurrent-checking safe? wouldn't we, potentially, get a different set of errors with multiple error nodes given each concurrent checker could report the error for a different firt encountered `errorNode`?

jakebailey: So, yes, I agree, though I believe it's handled because we spawn a specific number of checkers and then round robin them, and then they all walk their files in order, so it's always consistent.

But, I think we should probably change it, since this affects the editor and would change if we allow custom checker counts.

internal/checker/checker.go
jakebailey: I have a feeling that we might want to do that, just because we can hit this path through regular checker methods externally and then repeatedly tack more diagnostics onto the list.

Andarist: done

jakebailey: Actually, won't this be like one entry for every single loop, yield, etc, that we check?

jakebailey: Both options seem bad, kinda...

jakebailey: Hm, I guess the way you had it originally was more similar to the original code, right?

Andarist: yes, it was more similar - and it only was re-retrieving the iteration types when the cached value had no types. So it seems it should be faster than including the nodeId in the cache key

jakebailey: Okay, then I guess go back, sorry 😄

Andarist: I pushed out a revert going back to the `noCache` variant

PR #1064, opened 2025-06-05
Issue implicit any in JS files on widened inferred types from initializers in binding patterns

Andarist:
ports https://github.com/microsoft/TypeScript/pull/60083


jakebailey:
Can you remerge main here?

Andarist:
done

testdata/baselines/reference/submoduleAccepted/conformance/destructuringParameterDeclaration9(strict=true).errors.txt.diff
```diff
@@ -1,53 +1,39 @@
...
++==== index.js (4 errors) ====
+     /**
+      * @param {Object} [config]
+      * @param {Partial<Record<'json' | 'jsonc' | 'json5', string[]>>} [config.additionalFiles]
+@@= skipped -8, +11 lines =@@
+     export function prepareConfig({
+         additionalFiles: {
+             json = []
++            ~~~~
++!!! error TS7031: Binding element 'json' implicitly has an 'any[]' type.
```
jakebailey: I also tried porting this and hit these same issues. I'm pretty sure these are exactly the things that we are aiming to prevent, but now I think the fix doesn't quite work anymore?

Andarist: I'll take a look at this tomorrow~ then

Andarist: This PR aims to issue implicit any error - and it does that... to some extent 😉 It is overly eager though. From what I can see, this one is still not handled by the reparser - and that's why we are seeing errors here. The compiler thinks it doesn't have a type annotation

sandersn: I think @a-tarasyuk had a recent PR that improves handling of destructured parameters. Do things improve if you merge from main?

testdata/baselines/reference/submodule/conformance/typeFromJSInitializer.errors.txt
```diff
@@ -37,6 +36,8 @@ a.js(41,12): error TS2345: Argument of type '"ok"' is not assignable to paramete
     
     // should get any on parameter initialisers
```
sandersn, line 37: I don't understand what is going on here--the differing errors appear to match the comments better than before, so maybe the old baselines are wrong?

sandersn: oh, this is a .errors.txt, not a .errors.txt.diff

testdata/baselines/reference/submoduleAccepted/conformance/destructuringParameterDeclaration9(strict=true).errors.txt.diff
Andarist: yes! `destructuringParameterDeclaration9(strict=true).errors.txt.diff` is completely gone now

internal/checker/checker.go
```diff
@@ -17212,7 +17227,7 @@ func (c *Checker) getTypeFromBindingElement(element *ast.Node, includePatternInT
 		if ast.IsBindingPattern(element.Name()) {
 			contextualType = c.getTypeFromBindingPattern(element.Name(), true /*includePatternInType*/, false /*reportErrors*/)
 		}
-		return c.addOptionality(c.widenTypeInferredFromInitializer(element, c.checkDeclarationInitializer(element, CheckModeNormal, contextualType)))
```
sandersn, line 17215: this new function is called from a couple of other places in Strada. Why not in Corsa too?

sandersn (dismissed):
diffs now look good but the new function is called two other places in Strada, so maybe there are more diffs that could be eliminated?

testdata/baselines/reference/submoduleAccepted/conformance/destructuringParameterDeclaration9(strict=true).types.diff
```diff
@@ -9,24 +9,7 @@
 
      additionalFiles: {
  >additionalFiles : any
-@@= skipped -26, +26 lines =@@
- >additionalFiles : any
-
-         json = []
-->json : any[]
```
jakebailey, line 16: Wonder why this is still wrong...

internal/checker/checker.go
Andarist: I only see 3 `getWidenedLiteralTypeForInitializer` hits in both codebases 🤔 And they seem to be in the same places. Same for `widenTypeInferredFromInitializer` (4 hits)

testdata/baselines/reference/submoduleAccepted/conformance/destructuringParameterDeclaration9(strict=true).types.diff
Andarist: this has been **removed** from the diff - so this discrepancy is actually being fixed by this PR. Maybe you have meant to comment on some other line?

jakebailey: I'm dumb, yes

internal/checker/checker.go
jakebailey: How is that possible? This PR is introducing getWidenedLiteralTypeForInitializer and only adds one call

Andarist: it splits the old `widenTypeInferredFromInitializer` into `widenTypeInferredFromInitializer` and `getWidenedLiteralTypeForInitializer`. You have the second call to `getWidenedLiteralTypeForInitializer` at the top of the new version of `widenTypeInferredFromInitializer`

jakebailey: Right, but the old code had other calls to the get variant

jakebailey: No, I'm wrong, it seems fine

PR #1075, opened 2025-06-05
Move version from source file to document registry

gabritto:
The version really only concerns the project and document registry, so I moved it there to avoid a data race in #991 where we cache `SourceFile`s.

### DIFF
--- a/internal/project/documentregistry.go
+++ b/internal/project/documentregistry.go
@@ -27,6 +27,7 @@ func newRegistryKey(options *core.CompilerOptions, path tspath.Path, scriptKind 
 
 type registryEntry struct {
 	sourceFile *ast.SourceFile
+	version    int
 	refCount   int
 	mu         sync.Mutex
 }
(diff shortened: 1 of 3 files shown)


internal/project/project.go
```diff
@@ -46,7 +46,7 @@ type snapshot struct {
 func (s *snapshot) GetLineMap(fileName string) *ls.LineMap {
 	file := s.program.GetSourceFile(fileName)
 	scriptInfo := s.project.host.GetScriptInfoByPath(file.Path())
-	if file.Version == scriptInfo.Version() {
+	if s.project.getFileVersion(file, s.program.Options()) == scriptInfo.Version() {
```
jakebailey, line 51: Doesn't the left side of this just return the script info too, so this will always be true?

gabritto: It returns the version in the document registry's entry for that file. Do you mean that somehow that version is gonna end up always being the script info version? If so, do you have a scenario in mind? I can't really see it.

jakebailey: I think I'm just wrong, the doc registry maps paths/options to source files, not scriptinfos.

sheetalkamat (approved):

internal/project/documentregistry.go
```diff
@@ -118,6 +119,14 @@ func (r *DocumentRegistry) getDocumentWorker(
 	}
 }
 
+func (r *DocumentRegistry) getFileVersion(file *ast.SourceFile, options *core.CompilerOptions) int {
+	key := newRegistryKey(options, file.Path(), file.ScriptKind)
+	if entry, ok := r.documents.Load(key); ok {
+		return entry.version
```
sheetalkamat, line 125: This is not enough i think. The source file may have been updated for the compilerOptions by other program and hence the source file may be newer = script info version. You actually want to check if `file == entry.sourceFile` as well

sheetalkamat (dismissed):

internal/project/documentregistry.go
gabritto: Good catch, thanks!

andrewbranch (approved):

PR #1078, opened 2025-06-06
Project references in a program

sheetalkamat:


internal/ast/utilities.go
```diff
@@ -3580,3 +3580,23 @@ func IsTrivia(token Kind) bool {
 func HasDecorators(node *Node) bool {
 	return HasSyntacticModifier(node, ModifierFlagsDecorator)
 }
+
+type HasFileNameImpl struct {
```
jakebailey: This doesn't need to be exported, given its only ever used via its values and through an interface.

internal/checker/utilities.go
```diff
@@ -1471,10 +1471,15 @@ func forEachYieldExpression(body *ast.Node, visitor func(expr *ast.Node)) {
 	traverse(body)
 }
 
-func SkipTypeChecking(sourceFile *ast.SourceFile, options *core.CompilerOptions) bool {
+type SkipTypeCheckingHost interface {
+	IsSourceFromProjectReference(path tspath.Path) bool
+}
+
+func SkipTypeChecking(sourceFile *ast.SourceFile, options *core.CompilerOptions, host SkipTypeCheckingHost) bool {
```
jakebailey: Seems like `host` could safely be `checker.Program`? Is this ever used in another way?

internal/compiler/fileloader.go
```diff
@@ -2,45 +2,53 @@ package compiler
...
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/module"
+	"github.com/microsoft/typescript-go/internal/modulespecifiers"
 	"github.com/microsoft/typescript-go/internal/tsoptions"
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
 type fileLoader struct {
-	opts                ProgramOptions
+	opts                *ProgramOptions
```
jakebailey: Does this need to be a pointer? Within reason, memcopies like this are not that expensive.

internal/compiler/fileloader.go
```diff
@@ -59,24 +66,39 @@ func processAllProgramFiles(
...
 
 	loader.addRootTasks(rootFiles, false)
 	loader.addRootTasks(libs, true)
 	loader.addAutomaticTypeDirectiveTasks()
 
-	loader.startTasks(loader.rootTasks)
-
-	loader.wg.RunAndWait()
+	loader.parseTasks.runAndWait(&loader, loader.rootTasks)
+	loader.projectReferenceFileMapper.loader = nil
```
jakebailey, line 93: Why nil this out? For GC reasons?

internal/compiler/host.go
```diff
@@ -74,9 +91,17 @@ func (h *compilerHost) Trace(msg string) {
 }
 
 func (h *compilerHost) GetSourceFile(fileName string, path tspath.Path, languageVersion core.ScriptTarget) *ast.SourceFile {
-	text, _ := h.FS().ReadFile(fileName)
+	text, ok := h.FS().ReadFile(fileName)
```
jakebailey: Ah, you found the same thing I found 😅 https://github.com/microsoft/typescript-go/pull/1148

internal/compiler/program.go
```diff
@@ -262,7 +254,8 @@ func (p *Program) initCheckerPool() {
 }
 
 func canReplaceFileInProgram(file1 *ast.SourceFile, file2 *ast.SourceFile) bool {
-	return file1.FileName() == file2.FileName() &&
+	return file2 != nil &&
```
jakebailey: I had this nil check outside this func in #1148, not sure which is better.

internal/core/commonsourcedirectory.go
```diff
@@ -0,0 +1,86 @@
+package core
```
jakebailey: In a way I wonder if this would be better in the `outputpaths` package.

internal/core/projectreference.go
```diff
@@ -1,7 +1,20 @@
...
 	Path         string
 	OriginalPath string
 	Circular     bool
 }
+
+func ResolveProjectReferencePath(ref *ProjectReference) string {
+	return resolveConfiFileNameOfProjectReference(ref.Path)
+}
+
+func resolveConfiFileNameOfProjectReference(path string) string {
```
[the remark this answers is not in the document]
jakebailey: Yeah, typo.

internal/execute/tsc.go
```diff
@@ -115,8 +116,8 @@ func executeCommandLineWorker(sys System, cb cbType, commandLine *tsoptions.Pars
 
 	if configFileName != "" {
 		configStart := sys.Now()
-		extendedConfigCache := map[tspath.Path]*tsoptions.ExtendedConfigCacheEntry{}
-		configParseResult, errors := tsoptions.GetParsedCommandLineOfConfigFile(configFileName, compilerOptionsFromCommandLine, sys, extendedConfigCache)
+		extendedConfigCache := collections.SyncMap[tspath.Path, *tsoptions.ExtendedConfigCacheEntry]{}
```
jakebailey, line 119: Do we actually do this concurrently enough to require a SyncMap?

internal/module/resolver.go
```diff
@@ -129,6 +135,10 @@ func NewResolver(
 	}
 }
 
+func (r *Resolver) GetHost() ResolutionHost {
```
jakebailey: This seems like an internal detail, do we need the resolver to provide this to someone?

internal/modulespecifiers/symlinkcache.go
```diff
@@ -0,0 +1,53 @@
+package modulespecifiers
+
+import (
+	"github.com/microsoft/typescript-go/internal/collections"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type SymlinkedDirectory struct {
```
jakebailey: I feel moderately suspicious of needing this again; why is the FS caching not enough?

internal/core/commonsourcedirectory.go
jakebailey: Oh, this code does seem duplicated with code added to `outputpaths`?

internal/compiler/fileloader.go
sheetalkamat: And to ensure that loader isnt used after program creation is complete

internal/core/commonsourcedirectory.go
sheetalkamat: yeah sorry about that. i forgot to delete when i moved this code around

internal/execute/tsc.go
sheetalkamat: we will with project references - many projects could be using same extends cache and parsing at the same time.

internal/compiler/fileloader.go
jakebailey: I'd be good to leave a comment about it, just to make it clear that it's for assurances later.

internal/modulespecifiers/symlinkcache.go
andrewbranch: We sometimes take a file path and ask for all the symlinks that resolve to it.

sheetalkamat: For now moved this to compiler - we can move it specifiers when we need it 
Called it knownLinks to not confuse with cache per jake's suggestion

internal/compiler/fileloadertask.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+		}
+	}
+}
+
+func (w *fileLoaderWorker[K]) collect(loader *fileLoader, tasks []K, iterate func(K, []tspath.Path)) []tspath.Path {
+	return w.collectWorker(loader, tasks, iterate, collections.Set[K]{})
+}
+
+func (w *fileLoaderWorker[K]) collectWorker(loader *fileLoader, tasks []K, iterate func(K, []tspath.Path), seen collections.Set[K]) []tspath.Path {
```
jakebailey, line 48: I am suspicious of `collections.Set[K]` just because it's an auto-initializing struct containing a map, which means passing it by value will cause separate call chains to not actually share that underlying map. Perhaps a pointer is warranted here?

jakebailey: I guess this probably works out because the first thing it does is `add` the task, so any recursive calls will have preinitialized the map.

sheetalkamat: Yeah. which is why i didnt make it address but i can if thats more clear

jakebailey: I guess it's fine for now. I'll probably `noCopy` these in the future to make the bug impossible, but there's technically no bug here.

jakebailey (approved):

PR #1082, opened 2025-06-06
Port "Fix flow node improper reuse"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/60662


ahejlsberg:
We're not doing incremental parsing in Corsa. When a source file is modified, we always completely rebuild the AST. So, I don't think there's a scenario where the code in this PR is needed.

Andarist:
That's cool 👍 I was wondering if the AST reuse is still done here

jakebailey:
Marked this as not needed on the PR board.

PR #1085, opened 2025-06-06
Improve type discrimination algorithm

### LINKED ISSUE #1020: Discrepancies in behavior of spread of union type
```ts
type State =
  | { loaded: true; type: "loader"; value: string }
  | { loaded: false; type: "loader" }
  | { type: "other" };

declare class Store {
  state: State;
  setState(state: State): void;
}

function unload(store: Store): void {
  store.setState({ ...store.state, loaded: false });
}
```

This passes in TS, in Go we get:
```ts
src/union.ts:12:36 - error TS2353: Object literal may only specify known properties, and 'loaded' does not exist in type '{ type: "other"; }'.

12   store.setState({ ...store.state, loaded: false });
                                      ~~~~~~
```

(It's a bit debatable who's philosophically more correct -- it depends on which objects are allowed to have unknown properties.)

ahejlsberg:
Implements what I suggest [here](https://github.com/microsoft/typescript-go/issues/1020#issuecomment-2949341867).

We might consider back-porting this to the old code base.

Fixes #1020.

### DIFF
--- a/testdata/baselines/reference/compiler/missingDiscriminants.errors.txt
+++ b/testdata/baselines/reference/compiler/missingDiscriminants.errors.txt
@@ -1,0 +1,28 @@
+missingDiscriminants.ts(17,23): error TS2353: Object literal may only specify known properties, and 'subkind' does not exist in type '{ kind: "b"; }'.
+missingDiscriminants.ts(18,34): error TS2353: Object literal may only specify known properties, and 'subkind' does not exist in type '{ kind: "b"; }'.
+
+
+==== missingDiscriminants.ts (2 errors) ====
+    // https://github.com/microsoft/typescript-go/issues/1020
+    
+    type Thing =
+      | { str: "a", num: 0 }
+      | { str: "b" }
+      | { num: 1 }
+    
+    const thing1: Thing = { str: "a", num: 0 }
+    const thing2: Thing = { str: "b", num: 1 } // Shouldn't be error
+    const thing3: Thing = { num: 1, str: "b" } // Shouldn't be error
+    
+    type Item =
+      | { kind: "a", subkind: 0, value: string }
+      | { kind: "a", subkind: 1, value: number }
+      | { kind: "b" }
+    
+    const item1: Item = { subkind: 1, kind: "b" } // Error, type "b" not assignable to type "a"
+                          ~~~~~~~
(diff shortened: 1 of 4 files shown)


testdata/baselines/reference/submodule/compiler/discriminateWithMissingProperty.errors.txt
```diff
@@ -1,22 +0,0 @@
...
-        data: number,
-    } | {
-        mode: "alphabetic",
-        data: string,
-    } | {
-        data: string | Uint8Array;
-    }
-    
-    declare function foo(arg: Arg): void;
-    foo({ mode: "numeric", data: new Uint8Array([30]) }); // Should error
```
jakebailey, line 18: Removing this error is sort of interesting; I suppose it's technically correct to pass this extra prop in, but I'm surprised it doesn't hit EPC?

ahejlsberg: For context, here's the test that no longer errors:
```ts
type Arg =
    | { mode: "numeric", data: number }
    | { mode: "alphabetic", data: string }
    | { data: string | Uint8Array }

declare function foo(arg: Arg): void;

foo({ mode: "numeric", data: new Uint8Array([30]) }); // Now no error
```

Previously, discrimination would remove all but the first variant from the target type and then report an error because the `data` types are mismatched. Now, discrimination only removes the `"alphabetic"` variant (which is correct). Since the source type is assignable to the last variant, there's no type mismatch error. But there is also no excess property error because one of the variants does have a `mode` property.

Note that the `data` property _isn't_ considered a discriminant property because it contains no unit types. Had it contained a unit type, we would have further discriminated the target down to just the last variant and then reported an excess property error on `mode`. There have been suggestions to relax the unit type restriction, but that's beyond this PR.

jakebailey:
Ported it over in https://github.com/microsoft/TypeScript/pull/61828 and it fails the self test. I wonder why this doesn't fail in CI here.

jakebailey:
Ah, because the failure isn't in src/compiler, which is what we test here given we don't have build mode.

jakebailey (approved):

Eigilak:
The man, the myth, the legend 🫡

PR #1089, opened 2025-06-06
Add copilot-setup-steps

andrewbranch:


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,27 @@
...
+    permissions:
+      # If you want to clone the repository as part of your setup steps, for example to install dependencies, you'll need the `contents: read` permission. If you don't clone the repository in your setup steps, Copilot will do this for you automatically after the steps complete.
+      contents: read
+
+    # You can define any steps you want, and they will run before the agent starts.
+    # If you do not check out your code, Copilot will do this for you.
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: true
```
jakebailey, line 21:
```suggestion
          submodules: true
          filter: blob:none
          fetch-depth: 0
```

I was going to suggest this but https://github.com/actions/checkout/blob/09d2acae674a48949e3602304ab46fd20ae0c42f/src/git-command-manager.ts#L57 implies that filters are not passed down, only the fetch depth :(

andrewbranch: Since I've been giving it web access to PRs/patches, I don't think it uses the submodule at all, so we could probably remove this. But the shallow clone is pretty fast.

PR #1097, opened 2025-06-07
Add handling for 2-3 invalid overload candidates

### LINKED ISSUE #1088: Inconsistent behavior between `tsc` and `tsgo` for `@ts-expect-error` directive placement
There appears to be an inconsistency in how `tsc` and `tsgo` handle the placement of `@ts-expect-error` directives when dealing with `TS2769` (No overload matches this call) errors.

``` ts
import { reduce } from 'lodash';

export function foo(state: any) {
	// tsc is happy with @ts-expect-error here
	// @ts-expect-error FIXME (TS2769)  No overload matches this call.; Remove this comment to see full error message
	return reduce(
		{ foo: 'bar' },
		(fuzz: number) => {
			return fuzz;
		},
		// tsgo is happy with @ts-expect-error here
		{}
	);
}

```

Repo to see the error with: https://github.com/cartond/tsgo-ts-expect-err-repro

muglug:
Fixes #1088 by filling in missing logic for handling 2-3 overload errors.


muglug:
Actually it doesn't fix it, yet. Boo.

jakebailey:
I'm pretty sure that this difference is intentional, specifically the "not showing a billion messages anymore" in favor of "just report the last one". Though, obviously this does appear to improve the diffs.

jakebailey:
I don't quite see how this is related to #1088?

PR #1098, opened 2025-06-07
Add handling for 2-3 invalid overload candidates

### LINKED ISSUE #1088: Inconsistent behavior between `tsc` and `tsgo` for `@ts-expect-error` directive placement

muglug:
Fixes #1088 by adding back the 2-3 override arg code.

The overload descriptions in errors _seem_ wonky still (they're adding "new (..." in a bunch of places but I don't think that's touched in the specific code here

### DIFF
--- a/internal/project/configfileregistry.go
+++ b/internal/project/configfileregistry.go
@@ -3,6 +3,7 @@ package project
 import (
 	"context"
 	"fmt"
+	"maps"
 	"slices"
 	"sync"
 
@@ -14,7 +15,7 @@ import (
 )
 
 type ConfigFileEntry struct {
-	mu             sync.Mutex
+	mu             sync.RWMutex
 	commandLine    *tsoptions.ParsedCommandLine
 	projects       collections.Set[*Project]
 	infos          collections.Set[*ScriptInfo]
@@ -114,6 +115,8 @@ func (c *ConfigFileRegistry) acquireConfig(fileName string, path tspath.Path, pr
 func (c *ConfigFileRegistry) getConfig(path tspath.Path) *tsoptions.ParsedCommandLine {
 	entry, ok := c.ConfigFiles.Load(path)
 	if ok {
+		entry.mu.RLock()
+		defer entry.mu.RUnlock()
 		return entry.commandLine
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


jakebailey:
I said this in https://github.com/microsoft/typescript-go/pull/1097#issuecomment-2952945701 but I'm not 100% certain that this particular thing was a bug. Note how much more text the old compiler had to complain about roughly the same thing.

jakebailey:
Though moving the error lines is not super great. Not sure why this change is needed exactly, given I would have expected the overload error to be placed in the same place but...

muglug:
Moving the lines is the problematic thing for us because it invalidates our existing `@ts-expect-error` comments, which allows us to run both Go & TypeScript checkers on the codebase simultaneously until it's safe to switch over.

I think that specific linked issue could be fixed by applying [these lines](https://github.com/microsoft/typescript-go/pull/1098/files#diff-8e542d9de5c6b06221168642e2d9dd2df1f2d42d9c79ab60dc28fa6eac85fe39R9390-R9406) in the case of 2 or 3 overload errors, with a comment explaining that it's done to be backwards-compatible with existing error placement.

Making this change needn't prohibit you simplifying this behaviour and/or changing error-reported lines in TypeScript 7.1+, but matching the existing typechecker's error output reduces the conformances diffs to make meaningful differences stand out more while the port is getting its finishing touches.

jakebailey:
I do agree that the errors moving is a problem. If this is intentional (I believe so, based on past discussions), perhaps what we should do instead is port this into 5.9.

To be clear, the positioning of errors is not really a guaranteed thing. In your case, I would really suggest not using `ts-ignore` or `ts-expect-error`, instead using casts (or `as Foo satisfies Foo`) to get the overload you need.

jakebailey:
Just FYI, I don't think we're going to take this PR as-is; the changes here were intentional and argugably should be backported to Strada.

That being said, I think @ahejlsberg was going to look at https://github.com/microsoft/typescript-go/issues/1088 to see whether or not the error positioning itself is unintentional for this issue.

PR #1099, opened 2025-06-08
Type width in error messages

bjatkin:
Fixes https://github.com/microsoft/TypeScript/issues/45896

Currently the complier will print "... is not assignable..." errors on a single line
```
prettyFormatLongTypes.ts: - error TS2882: Type '{ b: { c: { e: { f: string; }; }; }; }' is not assignable to type '{ b: { c: { e: { f: number; }; }; }; }'
 The types of 'b.c.e.f' are incompatible between these types.
    Type 'string' is not assignable to type 'number'.
```

This PR puts types onto their own lines when they are >30 characters in length
```
prettyFormatLongTypes.ts: - error TS2882: Type:
  { b: { c: { e: { f: string; }; }; }; }

is not assignable to type:
  { b: { c: { e: { f: number; }; }; }; }

  The types of 'b.c.e.f' are incompatible between these types.
    Type 'string' is not assignable to type 'number'.
```

This PR is the typescrpt-go version of this PR https://github.com/microsoft/TypeScript/pull/61829 and needs to be merged after that PR has been merged to main so the internal submodule can be updated.

### DIFF
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -4705,7 +4705,14 @@ func (r *Relater) reportRelationError(message *diagnostics.Message, source *Type
 					return
 				}
 			}
-			message = diagnostics.Type_0_is_not_assignable_to_type_1
+
+			includesWideTypes := len(sourceType) > 30 && len(targetType) > 30 
+			isPretty := r.c.compilerOptions.Pretty == core.TSTrue
+			if isPretty && includesWideTypes {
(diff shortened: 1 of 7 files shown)


jakebailey:
I'm going to mark this as a draft. I don't think we've accepted this feature, and the direction of new features like this needs to be from the TypeScript repo into this one.

jakebailey:
Thanks, but I'm going to close this; we're not accepting this kind of change at this point, but we can discuss it on the main repo after 7.0.

PR #1107, opened 2025-06-09
Add extra test entrypoint for formatter for profiling

weswigham:
And throw in a bugfix and a few memory use improvements for the formatter. Combined, these lower the total count of allocations from around 2 million to 750k per `checker.ts` format. Byte-wise, it goes from around 113MB allocated to 83MB (36% savings). These, in turn, help performance on cache-limited systems substantially.

### DIFF
--- a/internal/execute/fmt_test.go
+++ b/internal/execute/fmt_test.go
@@ -1,0 +1,95 @@
+package execute
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/microsoft/typescript-go/internal/bundled"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/repo"
+	"github.com/microsoft/typescript-go/internal/tspath"
+	"github.com/microsoft/typescript-go/internal/vfs"
+	"github.com/microsoft/typescript-go/internal/vfs/osvfs"
+)
+
+func BenchmarkFormat(b *testing.B) {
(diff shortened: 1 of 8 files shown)


internal/execute/fmt_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/microsoft/typescript-go/internal/bundled"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/repo"
+	"github.com/microsoft/typescript-go/internal/tspath"
+	"github.com/microsoft/typescript-go/internal/vfs"
+	"github.com/microsoft/typescript-go/internal/vfs/osvfs"
+)
+
+func BenchmarkFormat(b *testing.B) {
```
jakebailey: I realized after the fact that you had already added formatting benchmarks to the `format` package. Could these just go there? Or, is this just a dupe?

jakebailey (commented):
The actual perf changes seem good, though I think the PR title should be updated to talk just about the perf given the code for the formatting test entrypoint is commented out. (Do we need that code? Can we just keep it locally or just use regular benchmarks?)

weswigham:
> Do we need that code?

I'd like to make it a public entrypoint to the compiler at some point in the future (alongside other LS features), but for now it's useful to keep around as a quick edit for testing and making changes to the formatter without waiting for a full vscode extension reload.

jakebailey:
Ah, I broke this via #979

jakebailey (approved):

PR #1148, opened 2025-06-10
Do not create SourceFiles for non-existent files

jakebailey:
Noticed this while working on another PR. We were treating all not-found files as empty.

~Perhaps this is just papering over some places we don't precheck that a file exists?~ No, Strada had `getSourceFile(...): SourceFile | undefined`.


sheetalkamat (approved):
I had done this change in #1078  where some of my changes werent working properly because of this

internal/compiler/program.go
```diff
@@ -229,7 +229,7 @@ func NewProgram(opts ProgramOptions) *Program {
 func (p *Program) UpdateProgram(changedFilePath tspath.Path) (*Program, bool) {
 	oldFile := p.filesByPath[changedFilePath]
 	newFile := p.Host().GetSourceFile(oldFile.FileName(), changedFilePath, oldFile.LanguageVersion)
-	if !canReplaceFileInProgram(oldFile, newFile) {
+	if newFile == nil || !canReplaceFileInProgram(oldFile, newFile) {
```
sheetalkamat: You can move this in canReplaceFileInProgram instead

jakebailey: I guess, I just thought it was clearer to have it here where we are getting a source file.

sheetalkamat (approved):

jakebailey:
CI race failure is #1151.

PR #1156, opened 2025-06-10
Cache CompilerOptions.SourceFileAffecting

jakebailey:
Enabled by #1106, we can now cache the SourceFileAffecting function's results on the CompilerOptions struct itself.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -304,7 +304,7 @@ func (p *Program) BindSourceFiles() {
 	for _, file := range p.files {
 		if !file.IsBound() {
 			wg.Queue(func() {
-				binder.BindSourceFile(file, p.Options().SourceFileAffecting())
+				binder.BindSourceFile(file, p.projectReferenceFileMapper.getCompilerOptionsForFile(file).SourceFileAffecting())
 			})
 		}
 	}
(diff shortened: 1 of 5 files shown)


internal/compiler/program.go
```diff
@@ -302,19 +299,12 @@ func (p *Program) singleThreaded() bool {
...
-	return p.sourceAffectingCompilerOptions
-}
-
 func (p *Program) BindSourceFiles() {
 	wg := core.NewWorkGroup(p.singleThreaded())
 	for _, file := range p.files {
 		if !file.IsBound() {
 			wg.Queue(func() {
-				binder.BindSourceFile(file, p.getSourceAffectingCompilerOptions())
+				binder.BindSourceFile(file, p.Options().SourceFileAffecting())
```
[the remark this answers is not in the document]
jakebailey: No, a `sync.Once.Do` only takes a single atomic read to verify that it's already been run.

sheetalkamat: I think this needs to be optionsForFile ?

internal/compiler/program.go
```diff
@@ -605,7 +595,7 @@ func compactAndMergeRelatedInfos(diagnostics []*ast.Diagnostic) []*ast.Diagnosti
 func (p *Program) getDiagnosticsHelper(ctx context.Context, sourceFile *ast.SourceFile, ensureBound bool, ensureChecked bool, getDiagnostics func(context.Context, *ast.SourceFile) []*ast.Diagnostic) []*ast.Diagnostic {
 	if sourceFile != nil {
 		if ensureBound {
-			binder.BindSourceFile(sourceFile, p.getSourceAffectingCompilerOptions())
+			binder.BindSourceFile(sourceFile, p.Options().SourceFileAffecting())
```
sheetalkamat: Same options for file

internal/compiler/program.go
jakebailey: I don't think so, then we'd have different options within the same `Program`. All of the other uses of `getCompilerOptionsForFile` appear to be immediately passed to other helpers to deal with module stuff?

jakebailey: Notably, #1158 moves this info onto the SourceFile, so these funcs should actually _not_ be consulting the Program at all.

jakebailey: You know, I think you're correct, yes. It just needs to match whatever was passed to `GetSourceFile`. #1158 makes this happen automatically, but the PR that introduced `getCompilerOptionsForFile` just missed the binder stuff.

jakebailey: There are some other places which seemingly don't have this call, and should probably be changed too.

PR #1157, opened 2025-06-10
Add this/satisfies tag

sandersn:
These are the last hosted tags!

I'm not sure about the name `makeNewCast`; copilot originally suggested two separate functions, which felt a little repetitive to me.


internal/parser/reparser.go
```diff
@@ -265,6 +270,32 @@ func (p *Parser) reparseHosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Node)
 				}
 			}
 		}
+	case ast.KindJSDocThisTag:
```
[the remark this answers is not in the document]
sandersn: There is an error for this case. I think it is fine to have weird semantics when we report an error.

testdata/baselines/reference/submodule/conformance/checkJsdocSatisfiesTag14.symbols.diff
```diff
@@ -0,0 +1,7 @@
+--- old.checkJsdocSatisfiesTag14.symbols
++++ new.checkJsdocSatisfiesTag14.symbols
+@@= skipped -14, +14 lines =@@
+
+ const t2 = /** @satisfies T1 */ ({ a: 1 });
+ >t2 : Symbol(t2, Decl(a.js, 9, 5))
+->a : Symbol(a, Decl(a.js, 9, 34))
```
jakebailey: Are we forgetting to walk these nodes?

sandersn: We intentionally skip synthetic nodes in the baselines since most synth nodes would cause dupes. Id like to undo this but after we have stopped relying on diffs.

sandersn: I could special case assertions and satisfies I suppose.

jakebailey: Right, but this means we entirely skip full trees, since these new cast nodes have real nodes underneath them. I think all of the existing cases are not like that, no?

jakebailey: (This is the first example, but there are worse diff additions below in the PR)

sandersn: There already is one for assertions so I will add one for satisfies as well.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag2.types.diff
```diff
@@ -7,8 +7,9 @@
 ->p : { isEven: (n: number) => boolean; isOdd: (n: number) => boolean; }
 ->({    isEven: n => n % 2 === 0,    isOdd: n => n % 2 === 1}) : { isEven: (n: number) => boolean; isOdd: (n: number) => boolean; }
 ->{    isEven: n => n % 2 === 0,    isOdd: n => n % 2 === 1} : { isEven: (n: number) => boolean; isOdd: (n: number) => boolean; }
-+>p : { isEven: (n: any) => boolean; isOdd: (n: any) => boolean; }
-+>({    isEven: n => n % 2 === 0,    isOdd: n => n % 2 === 1}) : { isEven: (n: any) => boolean; isOdd: (n: any) => boolean; }
++>p : any
```
sandersn, line 10: this is now `any` because Corsa no longer supports index signature syntax like `Object.<string, T>`

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag3.types.diff
```diff
@@ -5,32 +5,16 @@
  let obj = /** @satisfies {{ g(s: string): void } & Record<string, unknown>} */ ({
  >obj : { f(s: string): void; } & Record<string, unknown>
 ->({    f(s) { }, // "incorrect" implicit any on 's'    g(s) { }}) : { f(s: string): void; } & Record<string, unknown>
-->{    f(s) { }, // "incorrect" implicit any on 's'    g(s) { }} : { f(s: any): void; g(s: string): void; }
-+>({    f(s) { }, // "incorrect" implicit any on 's'    g(s) { }}) : { f(s: string): void; g(s: any): void; }
-+>{    f(s) { }, // "incorrect" implicit any on 's'    g(s) { }} : { f(s: string): void; g(s: any): void; }
++>({    f(s) { }, // "incorrect" implicit any on 's'    g(s) { }}) : { f(s: any): void; g(s: string): void; }
```
sandersn, line 8: I'm not sure what is going on with Strada's result here, because Corsa's is consistent with the errors for this file.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag8.types.diff
```diff
@@ -7,8 +7,9 @@
 ->x : { m: true; s: string; }
 ->({    m: true,    s: "false"}) : { m: true; s: string; }
 ->{    m: true,    s: "false"} : { m: true; s: string; }
-+>x : { m: boolean; s: string; }
-+>({    m: true,    s: "false"}) : { m: boolean; s: string; }
++>x : any
```
sandersn, line 10: same; more unsupported syntax

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag3.types.diff
Andarist: If I'm reading this correctly... it seems the difference is in "type at location". The parenthesized expression in Strada gets the "outer" type (the annoted type coming from `@type`) but in Corsa, according to this baseline diff, it gets the "inner" type (the RHS expression's type computed based on the contextual type introduced by `@satisfies`).

If that's correct then I'd say Corsa's result here makes more sense

sandersn: That explanation makes sense to me.

Andarist: @sandersn do u agree Corsa's result are better and what you intend to stick with? I could look into "porting" this to Strada so we could get the same behavior in both and get rid of the related baseline diffs at some point

jakebailey: We technically have a porting branch that a change like that could go on, though if this affects only a few baselines, it may not be worth it. Happy to test it out, though. Or, someone else can just by modifying their local submodule to update the tests.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag9.types.diff
```diff
@@ -0,0 +1,10 @@
+--- old.checkJsdocSatisfiesTag9.types
++++ new.checkJsdocSatisfiesTag9.types
+@@= skipped -12, +12 lines =@@
+ >Palette : { white: { r: number; g: number; b: number; }; black: { r: number; g: number; d: number; }; blue: { r: number; g: number; b: number; }; }
+ >({    white: { r: 255, g: 255, b: 255 },    black: { r: 0, g: 0, d: 0 }, // <- oops! 'd' in place of 'b'    blue: { r: 0, g: 0, b: 255 },}) : { white: { r: number; g: number; b: number; }; black: { r: number; g: number; d: number; }; blue: { r: number; g: number; b: number; }; }
+ >{    white: { r: 255, g: 255, b: 255 },    black: { r: 0, g: 0, d: 0 }, // <- oops! 'd' in place of 'b'    blue: { r: 0, g: 0, b: 255 },} : { white: { r: number; g: number; b: number; }; black: { r: number; g: number; d: number; }; blue: { r: number; g: number; b: number; }; }
++>{    white: { r: 255, g: 255, b: 255 },    black: { r: 0, g: 0, d: 0 }, // <- oops! 'd' in place of 'b'    blue: { r: 0, g: 0, b: 255 },} : { white: { r: number; g: number; b: number; }; black: { r: number; g: number; d: number; }; blue: { r: number; g: number; b: number; }; }
```
jakebailey: Some of these baselines only differ because they have an extra duplicate line. Is this just something that could be skipped in the type baseliner?

sandersn: probably. right now it intentionally does not skip as/satisfies, but I can see if it's possible to skip the node itself but not its children

sandersn: ok, done. the code is a little goofy but the baseline diffs are all gone.

testdata/baselines/reference/submodule/conformance/inferThis.symbols.diff
```diff
@@ -9,7 +9,7 @@
 
          return this;
 ->this : Symbol(this)
-+>this : Symbol(C, Decl(a.js, 0, 0))
++>this : Symbol((Missing), Decl(a.js, 3, 8))
```
jakebailey, line 12: I am curious what the deal is with `(Missing)` here, hopefully not a problem.

weswigham: iirc, it means the `Name` node of the symbol is missing/undefined, (and this isn't a class (it's the class static side), so it isn't `(anonymous class)`). Probably a bug somewhere, since `this` is jsdoc assigned to type `T`, so it really aughta be `T`, if not the generic unresolved `this` symbol. Perhaps lookup of the names of jsdoc template type params is broken?

internal/parser/reparser.go
```diff
@@ -266,6 +271,31 @@ func (p *Parser) reparseHosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Node)
...
+		if fun, ok := getFunctionLikeHost(parent); ok {
```
weswigham, line 275: What's this supposed to do for function-like-expressions that don't have a `this`, like arrow functions?

testdata/baselines/reference/submodule/conformance/inferThis.symbols.diff
jakebailey: https://github.com/microsoft/typescript-go/pull/731 potentially

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag3.types.diff
sandersn: For things that are better, or JS features that are cut, there is a file with accepted diffs. I haven't added anything to it yet, though I plan to soon.

internal/parser/reparser.go
sandersn: Copilot already noticed this and [I replied that there's already an error logged so it's OK with me to have wonky semantics in a file with errors](https://github.com/microsoft/typescript-go/pull/1157#discussion_r2138823013)

testdata/baselines/reference/submodule/conformance/inferThis.symbols.diff
sandersn: yes,    lookup of template types isbroken. I'm working on a fix right now as part of adding `@callback` and refactoring `@typedef` support.

PR #1158, opened 2025-06-10
Simplify parser options passing, eliminate ParseJSONText

jakebailey:
This PR adds `ast.SourceFileParseOptions`, a value type which is the input to file parsing. The parser now takes it instead of many parameters that have to be plumbed around, and then stores a copy on the SourceFile for later use in map keys and similar. This greatly simplifies `GetSourceFile` implementors.

Unfortunately, this PR is broken for JSON imports in specific resolution modes, and I haven't been able to spot the bug.


internal/compiler/host.go
```diff
@@ -90,15 +89,12 @@ func (h *compilerHost) Trace(msg string) {
...
+	text, ok := h.FS().ReadFile(opts.FileName)
 	if !ok {
 		return nil
 	}
-	if tspath.FileExtensionIs(fileName, tspath.ExtensionJson) {
-		return parser.ParseJSONText(fileName, path, text)
-	}
-	return parser.ParseSourceFile(fileName, path, text, options, metadata, scanner.JSDocParsingModeParseForTypeErrors)
+	return parser.ParseSourceFile(opts, text, core.GetScriptKindFromFileName(opts.FileName))
 }
```
jakebailey, line 98: Note that the `ScriptKind` and the text are _not_ parameters to `GetSourceFile` (not options). Implementations of `GetSourceFile` are in charge of fetching the file contents, and then they also override `ScriptKind` depending. Here, the core compiler only cares about the path to things, but in the LS the `ScriptKind` is determined by the language server (since the editor can say "actually, this is JS").

internal/parser/parser.go
```diff
@@ -172,39 +168,40 @@ func ParseJSONText(fileName string, path tspath.Path, sourceText string) *ast.So
...
 	p.nextToken()
 	entityName := p.parseEntityName(true, nil)
 	return core.IfElse(p.token == ast.KindEndOfFile && len(p.diagnostics) == 0, entityName, nil)
 }
 
-func (p *Parser) initializeState(fileName string, path tspath.Path, sourceText string, options *core.SourceFileAffectingCompilerOptions, metadata *ast.SourceFileMetaData, scriptKind core.ScriptKind, jsdocParsingMode scanner.JSDocParsingMode) {
+func (p *Parser) initializeState(opts ast.SourceFileParseOptions, sourceText string, scriptKind core.ScriptKind) {
+	if scriptKind == core.ScriptKindUnknown {
+		panic("ScriptKind must be specified when parsing source files.")
+	}
```
jakebailey, line 195: Allowing the compiler to determine the ScriptKind itself is a problem because it means we save onto the SourceFiles the "wrong" value, which means that using it for keys in DocumentRegistry and so on fails later because those caches are unaware of what the parser used to do to it.

internal/modulespecifiers/types.go
```diff
@@ -9,14 +9,6 @@ import (
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
-type SourceFileForSpecifierGeneration interface {
```
jakebailey: These interfaces (this and `HasFileName`) appear to do nothing. I could theoretically move the metadata calls onto these and keep it working, but I don't know if it's actually worth it right now.

internal/ast/ast.go
```diff
@@ -9978,17 +9995,13 @@ type SourceFile struct {
 	compositeNodeBase
 
 	// Fields set by NewSourceFile
-
-	text       string
-	fileName   string
```
jakebailey, line 9990: It may be the case that this is annoying for debugging. I could duplicate it back onto the file.

internal/ast/ast.go
```diff
@@ -9965,6 +9965,23 @@ type CommentDirective struct {
...
+type JSDocParsingMode int
+
+const (
+	JSDocParsingModeParseAll JSDocParsingMode = iota
+	JSDocParsingModeParseNone
+	JSDocParsingModeParseForTypeErrors
+	JSDocParsingModeParseForTypeInfo
+)
+
+type SourceFileParseOptions struct {
```
jakebailey: Potentially this should be called `SourceFileInfo` or something.

internal/ast/externalmoduleindicator.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type ExternalModuleIndicatorOptions struct {
+	jsx   bool
+	force bool
+}
```
jakebailey: Shifting this code around now means that the entire cache key for the external module indicator is just these two booleans! So much more reuse.

internal/ast/parseoptions.go
```diff
@@ -0,0 +1,162 @@
...
+	Path                           tspath.Path
+	CompilerOptions                core.SourceFileAffectingCompilerOptions
+	ExternalModuleIndicatorOptions ExternalModuleIndicatorOptions
+	JSDocParsingMode               JSDocParsingMode
+}
+
+func GetSourceFileAffectingCompilerOptions(fileName string, options *core.CompilerOptions) core.SourceFileAffectingCompilerOptions {
+	// Declaration files are not parsed/bound differently depending on compiler options.
+	if tspath.IsDeclarationFileName(fileName) {
+		return core.SourceFileAffectingCompilerOptions{}
```
jakebailey, line 28: It turns out that it's entirely safe to not use any options (source affecting, external module indicator) at all for declaration files, so we can now share these ASTs everywhere.

andrewbranch (approved):

PR #1159, opened 2025-06-11
Port JSX Emit

weswigham:
I'm still going through the baseline diffs and trying to minimize the bad ones, but there are two notable things I've seen already:
* `printer.EFStartOnNextLine` is not yet implemented, so all tag emit occurs on one line (this is also used lightly by the cjs module transform)


jakebailey:
> * `SourceFile.ExternalModuleIndicator` isn't set on files containing JSX tags, so if this is the only indicator a file is a module, implicit jsx runtime imports do not emit correctly in that file in jsx emit modes which use them (I think this is being worked on already?)

This should now be there in main as of earlier today, so that should just work if you merge.

weswigham:
> This should now be there in main as of earlier today, so that should just work if you merge.

Yes, but on investigation, a weeee bit off. One of those `true` early returns shoulda been a `false` - as-is, finding a subtree with no JSX stops it from looking for JSX in any other nodes, which, in turn, prevents finding a JSX tag pretty reliably. I've pushed the fix and updated baselines here.

jakebailey:
Great, thanks! 😄

testdata/baselines/reference/submodule/compiler/callsOnComplexSignatures.js.diff
```diff
@@ -38,21 +38,19 @@
      // Pair of non-like intrinsics
      function render(url) {
 -        var Tag = url ? 'a' : 'button';
```
jakebailey, line 40: I'm working on dropping ES5 out of the submodule, so theoretically that'd eliminate a load of diffs here too.

testdata/baselines/reference/submodule/compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=auto).js.diff
```diff
@@ -1,24 +1,16 @@
...
 +    render() {
-+        return <div>
-+            
-+            {null /* preserved */}
-+            
-+            
-+            
-+            
-+        </div>;
++        return jsx_runtime_1.jsx("div", { children: null /* preserved */ });
```
jakebailey, line 13: One difference I'm seeing is that the `(0, ...)` is missing. I'm not 100% sure who's in charge of adding that.

testdata/baselines/reference/submodule/compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=legacy).js.diff
```diff
@@ -8,14 +8,7 @@
...
-+            {null /* preserved */}
-+            
-+            
-+            
-+            
-+        </div>;
++        return _jsx("div", { children: null /* preserved */ });
      }
 -    Component.prototype.render = function () {
 -        return (0, _a.jsx)("div", { children: null /* preserved */ });
```
jakebailey, line 14: Seemingly missing whatever `_a` was?

internal/transformers/jsx.go
```diff
@@ -0,0 +1,1130 @@
...
+
+func (tx *JSXTransformer) visitJsxExpression(expression *ast.JsxExpression) *ast.Node {
+	e := tx.visitor.Visit(expression.Expression)
+	if expression.DotDotDotToken != nil {
+		return tx.factory.NewSpreadElement(e)
+	}
+	return e
+}
+
+var htmlEntityMatcher = regexp2.MustCompile(`&((#((\d+)|x([\da-fA-F]+)))|(\w+));`, regexp2.ECMAScript)
```
jakebailey, line 836: I sure hope this isn't in a hot path 😄

internal/transformers/jsx.go
```diff
@@ -0,0 +1,1130 @@
...
+ */
+func decodeEntities(text string) string {
+	res, err := htmlEntityMatcher.ReplaceFunc(text, htmlEntityReplacer, -1, -1)
+	if err != nil {
+		panic(err.Error())
+	}
+	return res
+}
+
+var entities = map[string]rune{
```
jakebailey, line 876: I would recommend turning this into a switch, since map lookups for stuff like this are not perfect.

jakebailey (commented):
Seems like the diffs fall into a couple of buckets.

(Most are ES5, or AMD/System/etc, which I am working to remove from our submodule)

testdata/baselines/reference/submodule/compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=auto).js.diff
weswigham: No idea tbh - not the JSX transform, tho, that's for sure. I think it's a downleveling one, though.

testdata/baselines/reference/submodule/compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=legacy).js.diff
weswigham: This is the emit when the file's `module` transform isn't the same as in Strada, as the substitution from `_jsx` to `mod.jsx` is done by the cjs transform. It implies something is still off with the module detection logic or substitution compared to strada - in this case, when `moduleDetection: legacy` is set. In this case, though, I don't think it's an important diff - the behavior of jsx auto-imports in script files (which this is) is explicitly not actually stable.

internal/transformers/jsx.go
weswigham: ehhh, emit for `JsxText` nodes? Not that hot.

internal/transformers/jsx.go
weswigham: Map lookups are hash-based but switches do binary searches, no? Unless switches also compile to a hash/perfect hash by the compiler when possible? What's "not perfect" mean here?

weswigham:
> Seems like the diffs fall into a couple of buckets.

Yeah, the vast majority remaining are `const` instead of `var` (or less so, other downleveling) and the multiline emit.

internal/transformers/jsx.go
jakebailey: They should be perfect hashes as of a few releases. It's fine to keep this for now, though.

tmm1:
Happy to see someone working on the emit side! 🙏🏽

PR #1176, opened 2025-06-12
Add configurable concurrency option

jakebailey:
This adds an experimental `--concurrency` flag. Example usage:

- `--concurrency <default|auto|on|yes>` - default behavior, uses 4 checkers, plus all cores for parse/bind/emit. Theoretically we could change this to some magic algorithm.
- `--concurrency <single|off|no>`, `--singleThreaded` - one checker, single threaded parse/bind/emit.
- `--concurrency max` - Uses `GOMAXPROCS` checkers, plus all cores for parse/bind/emit.
- `--concurrency half` - Uses `GOMAXPROCS/2` checkers, plus all cores for parse/bind/emit. (For me, this option is best for checking VS Code.)
- `--concurrency 10` - Uses 10 checkers, plus all cores for parse/bind/emit. Can be any number all the way up to the number of files.
- For testing, `--concurrency checker-per-file` creates a checker for every file; max concurrency, for testing race conditions.

Maybe this is overkill, I can drop some of this.


DanielRosenwasser:
Instead of `concurrency`, I would recommend `checkConcurrency`.

jakebailey:
Except that this also controls the concurrency of the parse/bind/emit stages too... I wasn't happy with any name I came up with.

jakebailey:
#1985

PR #1180, opened 2025-06-13
Convert read file content directly to strings

jakebailey:
I don't like unsafe, but in this case I believe it to be reasonable.

When reading from a filesystem, the returned bytes must necessarily never be touched by the underlying filesystem ever again. Otherwise, the caller would experience data races or be unable to modify the data. This means that in effect, the FS is guaranteed to no longer "own" that data, and the caller can do whatever it wants with it.

Our FS abstraction works with strings, but the underlying FSs all provide `[]byte`. Given the above, we can skip the typical `string` to `[]byte` copy and "safely" use `unsafe.String` to directly convert it.

This nets about a 50% improvement in time and memory allocation on non-trivial files. It's not much, but for larger files like `checker.ts`, it means going from a 1 ms read to a 0.5 ms read.

I'm not totally sure if we should do this (other stuff is more expensive), but this is an improvement, so...

```
goos: linux
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/vfs
                             │    sec/op    │   sec/op     vs base                │
ReadFile/MapFS_small-20         296.9n ± 0%   277.5n ± 1%   -6.52% (p=0.000 n=10)
ReadFile/OS_small-20            4.657µ ± 0%   4.723µ ± 1%   +1.41% (p=0.000 n=10)
ReadFile/MapFS_checker.ts-20   1067.8µ ± 6%   498.6µ ± 3%  -53.31% (p=0.000 n=10)
ReadFile/OS_checker.ts-20       833.4µ ± 7%   315.8µ ± 2%  -62.11% (p=0.000 n=10)
geomean                         33.30µ        21.31µ       -36.00%
```
```
                             │   old.txt    │               new.txt                │
                             │     B/op     │     B/op      vs base                │
ReadFile/MapFS_small-20          152.0 ± 0%     136.0 ± 0%  -10.53% (p=0.000 n=10)
ReadFile/OS_small-20            1016.0 ± 0%    1000.0 ± 0%   -1.57% (p=0.000 n=10)
ReadFile/MapFS_checker.ts-20   5.906Mi ± 0%   2.953Mi ± 0%  -50.00% (p=0.000 n=10)
ReadFile/OS_checker.ts-20      5.907Mi ± 0%   2.954Mi ± 0%  -50.00% (p=0.000 n=10)
geomean                        48.18Ki        33.00Ki       -31.50%
```
```
                             │  old.txt   │              new.txt               │
                             │ allocs/op  │ allocs/op   vs base                │
ReadFile/MapFS_small-20        5.000 ± 0%   4.000 ± 0%  -20.00% (p=0.000 n=10)
ReadFile/OS_small-20           11.00 ± 0%   10.00 ± 0%   -9.09% (p=0.000 n=10)
ReadFile/MapFS_checker.ts-20   5.000 ± 0%   4.000 ± 0%  -20.00% (p=0.000 n=10)
ReadFile/OS_checker.ts-20      11.00 ± 0%   10.00 ± 0%   -9.09% (p=0.000 n=10)
geomean                        7.416        6.325       -14.72%
```

### DIFF
--- a/internal/repo/paths.go
+++ b/internal/repo/paths.go
@@ -46,6 +46,10 @@ var typeScriptSubmoduleExists = sync.OnceValue(func() bool {
 	return true
 })
 
+func TypeScriptSubmoduleExists() bool {
+	return typeScriptSubmoduleExists()
+}
+
 type skippable interface {
 	Helper()
 	Skipf(format string, args ...any)
--- a/internal/vfs/internal/internal.go
+++ b/internal/vfs/internal/internal.go
@@ -148,7 +148,7 @@ func (vfs *Common) ReadFile(path string) (contents string, ok bool) {
 	}
 
 	// An invariant of any underlying filesystem is that the bytes returned
-	// are immtuable, otherwise anyone using the filesystem would end up
+	// are immutable, otherwise anyone using the filesystem would end up
 	// with data races.
 	//
 	// This means that we can safely convert the bytes to a string directly,
--- a/internal/vfs/internal/internal.go
+++ b/internal/vfs/internal/internal.go
@@ -1,11 +1,12 @@
 package internal
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
 	"io/fs"
+	"strings"
 	"unicode/utf16"
+	"unsafe"
 
 	"github.com/microsoft/typescript-go/internal/tspath"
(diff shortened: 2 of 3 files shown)


jakebailey:
Oh excellent, it seems like the `execute` package is violating this invariant. Actually, more. Blah.

PR #1184, opened 2025-06-13
Multiple improvements to Quick Info

ahejlsberg:
Improvements include:

* Show dotted names for members of classes and interfaces.
* Show the resolved overload in calls to overloaded functions and methods.
* Include type arguments in quick info for calls to generic functions and methods.

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -1404,7 +1404,11 @@ func (p *Printer) emitTypeParameter(node *ast.TypeParameterDeclaration) {
 }
 
 func (p *Printer) emitTypeParameterNode(node *ast.TypeParameterDeclarationNode) {
-	p.emitTypeParameter(node.AsTypeParameter())
+	if ast.IsTypeParameterDeclaration(node) {
+		p.emitTypeParameter(node.AsTypeParameter())
+	} else {
+		p.emitTypeArgument(node)
+	}
 }
 
 func (p *Printer) emitParameterName(node *ast.BindingName) {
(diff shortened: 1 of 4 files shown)


internal/printer/printer.go
```diff
@@ -1404,7 +1404,11 @@ func (p *Printer) emitTypeParameter(node *ast.TypeParameterDeclaration) {
 }
 
 func (p *Printer) emitTypeParameterNode(node *ast.TypeParameterDeclarationNode) {
-	p.emitTypeParameter(node.AsTypeParameter())
+	if ast.IsTypeParameterDeclaration(node) {
+		p.emitTypeParameter(node.AsTypeParameter())
+	} else {
+		p.emitTypeArgument(node)
+	}
```
jakebailey, line 1413: This felt spooky until I read the original code:

```
function emitTypeParameters(parentNode: SignatureDeclaration | InterfaceDeclaration | TypeAliasDeclaration | ClassDeclaration | ClassExpression, typeParameters: NodeArray<TypeParameterDeclaration> | undefined) {
    if (isFunctionLike(parentNode) && parentNode.typeArguments) { // Quick info uses type arguments in place of type parameters on instantiated signatures
        return emitTypeArguments(parentNode, parentNode.typeArguments);
    }
    emitList(parentNode, typeParameters, ListFormat.TypeParameters | (isArrowFunction(parentNode) ? ListFormat.AllowTrailingComma : ListFormat.None));
}
```

Can we leave a comment here saying why it's okay that this function is attempting to handle the wrong kind of thing?

ahejlsberg: Yeah, it's a bit ugly, but way less ugly than what we used to do. The main issue is that we're wanting to display a declaration, but with type arguments instead of type parameters--and type arguments only occur in calls. Allowing type arguments instead of type parameters in node building and emit seems like a reasonable way to accomplish this.

ahejlsberg: Anyway, I'll add a comment.

jakebailey (approved):

PR #1187, opened 2025-06-13
Add callback tag, fix jsdoc type parameter resolution

sandersn:
This PR adds support for the `@callback` tag. More importantly, it fixes type parameter resolution and number of other bugs in `@typedef` and `@template`. It does some of this by simplifying the rules for how `@template` attaches its type parameter to its host.

Specifically, all `@template` tags would previously apply to *all* `@overloads`, `@typedef`, function declarations, etc, that they shared a jsdoc with. They didn't even have to come before their usage:
```ts
/**
 * @template T
 * @typedef {T} Alias
 * @template U
 * @overload
 * @param {[T,U]} tu
 * @overload
 * @param {number} n
 * @return {string}
 * @param {any} tun
 * @return {string}
 */
function f(tun) {}

// is equivalent to TS
type Alias<T,U> = T
function f<T,U>(tu: [T,U])
function f<T,U>(n: number): string
function f<T,U>(tun: any): string {}
```
Not only is this nonsensical, it's really hard to implement, and I don't think people write their types in one single jsdoc like this in the first place. Corsa, as implemented in this PR, requires `@template` to precede its usage and for it to apply in only one place:
```ts
type Alias<T> = T
function f<U>(tu: [T,U])
function f(n: number): string
function f(tun: any): string
```
This will break any intentional sharing, but those will have to be rewritten. Sharing reparsed nodes to more than one other place in the tree would require `Host` to be a slice, plus some way to choose the correct host when resolving names.

Speaking of `Host`, `JSDocTemplateTag` now has a host, and the grossly-named `GetEffectiveTypeParent` now has two cases. It's also used consistently in the 5 places it makes sense. I do think we're nearing the total number of places it'll be needed, but there are 512 uses of `Node.Parent` in the checker alone, so I could be wrong.

Despite the improvements in typing, there is little code churn outside reparser.go. Inside, I refactored callback/typedef/overload quite a bit to share code, so there are almost 200 lines changed. Best reviewed ignoring whitespace, as always.


testdata/baselines/reference/submoduleAccepted/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt.diff
```diff
@@ -27,6 +28,8 @@
 +     * @template A
 +     * @template {Record<string, unknown>} B
 +     * @param {Funcs<A, B>} fns
++               ~~~~~~~~~~~
```
sandersn, line 31: template tags have to come before their usage now. If this restriction is lifted, then this test passes. But I really don't want to because it makes so little sense and complicates the code for choosing between possible hosts so much.

testdata/baselines/reference/submoduleAccepted/compiler/expandoFunctionContextualTypesJs.types.diff
```diff
@@ -6,27 +6,22 @@
  const MyComponent = () => /* @type {any} */(null);
 ->MyComponent : StatelessComponent<MyComponentProps>
 ->() => /* @type {any} */(null) : { (): any; defaultProps: Partial<MyComponentProps>; }
-+>MyComponent : { (): any; defaultProps?: P; }
-+>() => /* @type {any} */(null) : { (): any; defaultProps: { color: string; }; }
++>MyComponent : { (): any; defaultProps?: Partial<{ color: "blue" | "red"; }>; }
```
sandersn, line 9: type printing still de-aliases typedef when it shouldn't. It's on my list of things to investigate.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileFunctionOverloads2.types.diff
```diff
@@ -1,6 +1,19 @@
 --- old.jsFileFunctionOverloads2.types
 +++ new.jsFileFunctionOverloads2.types
-@@= skipped -83, +83 lines =@@
+@@= skipped -54, +54 lines =@@
+  * @returns {unknown[]}
+  */
+ function flatMap(array, iterable = identity) {
+->flatMap : { <T, U>(array: T[], iterable: (x: T) => U[]): U[]; <T, U>(array: T[][]): T[]; }
++>flatMap : { <T, U>(array: T[], iterable: (x: T) => U[]): U[]; (array: T[][]): T[]; }
```
sandersn, line 8: no type parameter sharing means that the second overload would need to add a `@template T` in order to be correct.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileFunctionOverloads2.types.diff
```diff
@@ -1,6 +1,19 @@
...
+ >array : unknown[]
+ >iterable : (x: unknown) => unknown
+->identity : <T_1>(x: T_1) => T_1
```
sandersn, line 11: pretty sure T_1 was because the implementation signature had its own T in scope, but that's no longer shared.

testdata/baselines/reference/submoduleAccepted/compiler/unmetTypeConstraintInJSDocImportCall.errors.txt.diff
```diff
@@ -5,26 +5,20 @@
...
+-    export default {};
+-    
 -==== file2.js (1 errors) ====
-+==== file2.js (0 errors) ====
-     /**
-      * @template T
-      * @typedef {import('./file1').Foo<T>} Bar
+-    /**
+-     * @template T
+-     * @typedef {import('./file1').Foo<T>} Bar
```
sandersn, line 18: still not checking tags on EOF

testdata/baselines/reference/submoduleAccepted/conformance/asyncFunctionDeclaration16_es5.errors.txt.diff
```diff
@@ -31,34 +30,29 @@
...
+     }
      
      /** @type {T1} */
-                ~~
+-               ~~
 -!!! error TS1065: The return type of an async function or method must be the global Promise<T> type.
 -!!! related TS1055 /a.js:4:14: Type 'string' is not a valid async function return type in ES5 because it does not refer to a Promise-compatible constructor value.
-+!!! error TS2304: Cannot find name 'T1'.
      const f2 = async str => {
++          ~~
```
sandersn, line 40: This error moved because Strada's *complex and very hard to support* `@type`-on-`function` feature is gone in Corsa so the error comes from contextual typing now.

testdata/baselines/reference/submoduleAccepted/conformance/asyncFunctionDeclaration16_es5.errors.txt.diff
```diff
@@ -31,34 +30,29 @@
...
-+!!! error TS2304: Cannot find name 'T2'.
-     const f4 = async str => {
-         return str;
+@@= skipped -28, +28 lines =@@
      }
      
      /** @type {T3} */
-                ~~
+-               ~~
 -!!! error TS1065: The return type of an async function or method must be the global Promise<T> type.
```
sandersn, line 51: not sure why this is missing--maybe because a contextually-typed Promise function is assignable to Thenable?

testdata/baselines/reference/submoduleAccepted/conformance/callbackTag2.errors.txt.diff
```diff
@@ -5,39 +5,26 @@
...
+@@= skipped -19, +20 lines =@@
      class SharedClass {
          constructor() {
              /** @type {SharedId<S>} */
 -                                ~
 -!!! error TS2304: Cannot find name 'S'.
              this.id;
 +                 ~~
-+!!! error TS2339: Property 'id' does not exist on type 'SharedClass<S>'.
++!!! error TS2339: Property 'id' does not exist on type 'SharedClass'.
```
sandersn, line 24: no more template tag sharing (but no constructor functions yet either)

testdata/baselines/reference/submoduleAccepted/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt.diff
jakebailey, line 31: This seems like it's regressing? And the types too?

testdata/baselines/reference/submoduleAccepted/compiler/jsFileMethodOverloads2.types.diff
```diff
@@ -17,7 +17,7 @@
     */
    transform(fn) {
 ->transform : { <U>(fn: (y: T) => U): U; <U>(): T; }
-+>transform : { <U>(fn: (y: T) => U): U; <U_1>(): T; }
++>transform : { <U>(fn: (y: T) => U): U; (): T; }
```
jakebailey, line 20: This is spooky; now there's a func without type params.

testdata/baselines/reference/submoduleAccepted/conformance/callbackTagVariadicType.errors.txt.diff
```diff
@@ -13,8 +13,8 @@
 +     */
 +    
 +    /** @type {Foo} */
-+               ~~~
-+!!! error TS2304: Cannot find name 'Foo'.
 +    export const x = () => 1
 +    var res = x('a', 'b')
++                     ~~~
```
sandersn, line 18: `...T` no longer makes an argument with its type into a rest type

testdata/baselines/reference/submoduleAccepted/compiler/unmetTypeConstraintInJSDocImportCall.errors.txt.diff
```diff
@@ -5,26 +5,20 @@
...
 -                                       ~
 -!!! error TS2344: Type 'T' does not satisfy the constraint 'string'.
```
jakebailey, line 20: We must still not be checking something inside these typedef nodes.

testdata/baselines/reference/submoduleAccepted/conformance/jsdocTemplateConstructorFunction.errors.txt.diff
```diff
@@ -5,25 +5,31 @@
...
+-     * @param {T} v
+-     * @param {Id<T>} id
+-     */
+-    Zet.prototype.add = function(v, id) {
+-        this.u = v || this.t
+-        return id(this.u)
+-    }
+-    var z = new Zet(1)
+-    z.t = 2
+-    z.u = false
```
sandersn, line 31: still no constructor functions

testdata/baselines/reference/submoduleAccepted/conformance/jsdocTemplateTag6.errors.txt.diff
```diff
@@ -0,0 +1,116 @@
...
++a.js(48,14): error TS1277: 'const' modifier can only appear on a type parameter of a function, method or class
++a.js(59,14): error TS1277: 'const' modifier can only appear on a type parameter of a function, method or class
++a.js(68,18): error TS1277: 'const' modifier can only appear on a type parameter of a function, method or class
++
++
++==== a.js (7 errors) ====
++    /**
++     * @template const T
++                 ~~~~~
++!!! error TS1277: 'const' modifier can only appear on a type parameter of a function, method or class
```
sandersn, line 18: oh nooo, I completely forgot about this feature. Not sure what is broken here exactly.

testdata/baselines/reference/submoduleAccepted/conformance/templateInsideCallback.types.diff
```diff
@@ -1,24 +1,15 @@
...
-->identity : any
-+>identity : Call<T>
- >x => x : <T>(x: any) => any
- >x : any
- >x : any
-@@= skipped -37, +37 lines =@@
- >flatMap : { <U>(): any; (): any; }
+ function flatMap(array, iterable = identity) {
+->flatMap : { <U>(): any; (): any; }
++>flatMap : { (): any; (): any; }
```
sandersn, line 8: no more sharing (unintended in this case, so it's an improvement)

testdata/baselines/reference/submoduleAccepted/conformance/typeTagNoErasure.types.diff
```diff
@@ -5,21 +5,19 @@
...
++>test : <T1 extends number>(data: T1) => T1
+ >dibbity => dibbity : <T1 extends number>(dibbity: T1) => T1
  >dibbity : T1
  >dibbity : T1
 
  test(1) // ok, T=1
  >test(1) : 1
 ->test : Test<number>
-+>test : <T1 extends T>(data: T1) => T1
++>test : <T1 extends number>(data: T1) => T1
```
sandersn, line 16: type printing still following aliases incorrectly

testdata/baselines/reference/submoduleAccepted/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt.diff
sandersn: intentional break, see my comment

testdata/baselines/reference/submoduleAccepted/compiler/jsFileMethodOverloads2.types.diff
sandersn: intentional change that's an improvement in this case (T comes from the class not the function). see my comment.

testdata/baselines/reference/submoduleAccepted/compiler/unmetTypeConstraintInJSDocImportCall.errors.txt.diff
sandersn: yep, still missing reparsing on EOF

testdata/baselines/reference/submoduleAccepted/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt.diff
jakebailey: It's better than nothing to have it work, but I do wonder how breaky this would be.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileFunctionOverloads2.types.diff
jakebailey: This seems like it's going to end up being even more breaky...

testdata/baselines/reference/submoduleAccepted/compiler/jsFileMethodOverloads2.types.diff
jakebailey: I guess it's that sharing thing again. I feel like it can't be good to break it...

testdata/baselines/reference/submoduleAccepted/conformance/asyncFunctionDeclaration16_es5.errors.txt.diff
jakebailey: > _complex and very hard to support_

this will not stop me from asking for the feature back 🫠

jakebailey (approved):

testdata/baselines/reference/submoduleAccepted/compiler/contravariantOnlyInferenceFromAnnotatedFunctionJs.errors.txt.diff
sandersn: I will try to support this specific case (only one possible host in the comment, template after typedef) if somebody complains about it. But I really don't think anybody expects this to work.

testdata/baselines/reference/submoduleAccepted/compiler/jsFileFunctionOverloads2.types.diff
sandersn: `@overload` is usually written with separated jsdoc comments to avoid the ambiguity that arises when looking for the end of a JSDocSignature, so most code will already be written in a way that requires the non-shared template tags (template tags are not shared between jsdoc comments)

PR #1189, opened 2025-06-14
Implement maxNodeModuleJsDepth, noResolve

### LINKED ISSUE #931: Some dependency JS files appear to be erroneously typechecked
When I built my project in TS-Go I ended up getting thousands of errors. While I'm still working through some of them to make more reports, the ones that struck me as most obviously problematic is thousands of errors in `.js` files in dependencies.

I decided to minimize my repo's code into something easier to investigate.
```
// index.ts
import "vitest/config";
import "simple-peer";

// package.json
{
  "dependencies": {
    "@types/simple-peer": "~9.11.1",
    "@typescript/native-preview": "^7.0.0-dev.20250523.1",
    "typescript": "^5.8.3",
    "vite": "^6.3.5",
    "vitest": "^3.0.5"
  }
}

// tsconfig.json
{
  "$schema": "https://json.schemastore.org/tsconfig.json",
  "compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "noEmit": true,
    "noUncheckedSideEffectImports": true, // Optional
    "allowJs": true,
    "checkJs": true
  }
}
```

Here's that same code as a zip: [dependency-js-errors.zip](https://github.com/user-attachments/files/20421076/dependency-js-errors.zip).

When `npx tsc` is run there are 0 errors, when `npx tsgo` is run this is the summary of the errors:
```
Found 321 errors in 4 files.

Errors  Files
   305  node_modules/chai/chai.js:13
     8  node_modules/vitest/dist/chunks/reporters.d.C-cu31ET.d.ts:5
     2  node_modules/vitest/dist/chunks/vite.d.iXCEVtFP.d.ts:1
     6  node_modules/vitest/dist/config.d.ts:1
```

My original repo experienced many more errors due to stricter tsconfig rules but I suspect that fixing these errors will fix the JS errors in my repo.

I stopped my reduction scripts once they had reduced down to minimal code in my own repo but if you'd my to help reducing the dependencies as well I'd be happy to help.

jakebailey:
Fixes #931

This was tricky to deal with in a concurrent loader.

The gist of it is that we keep track of the current "JS file depth" as we recurse. If a subtask about to be enqueued is a JS file from `node_modules`, we skip it if the current depth of JS files is too high. If the depth is okay, we'll load it, then recurse. If we ever revisit the same file again, we check to see if the new depth is lower than the last time the file is checked, and if it is we'll recheck its subtasks at that new depth (potentially loading more files).

While here, I've simplified a few things; namely I remembered the trick to have a self-referencing generic method and implemented `noResolve`.

Sadly, there's now a mutex per file, but I can't think of a cleaner method. The only place where this would matter is when the same file is referenced multiple times, in which case we need to recheck it serially anyhow.

We don't seem to have any tests for `maxNodeModuleJsDepth`; I'll think about how to make some. I'm also not setting up `maxNodeModuleJsDepth` for the the editor. I'm not sure where that goes.


testdata/baselines/reference/submoduleAccepted/compiler/jsEnumTagOnObjectFrozen.errors.txt.diff
```diff
@@ -0,0 +1,54 @@
+--- old.jsEnumTagOnObjectFrozen.errors.txt
++++ new.jsEnumTagOnObjectFrozen.errors.txt
+@@= skipped -0, +0 lines =@@
+-<no content>
++index.js(10,12): error TS2749: 'Thing' refers to a value, but is being used as a type here. Did you mean 'typeof Thing'?
++index.js(17,16): error TS2749: 'Thing' refers to a value, but is being used as a type here. Did you mean 'typeof Thing'?
```
jakebailey, line 6: Expected, since `@enum` doesn't do anything anymore and so no type is declared.

testdata/baselines/reference/submoduleAccepted/conformance/checkExportsObjectAssignProperty.errors.txt.diff
```diff
@@ -0,0 +1,147 @@
...
+-validator.ts(21,1): error TS2322: Type 'number' is not assignable to type 'string'.
+-validator.ts(37,4): error TS2540: Cannot assign to 'readonlyProp' because it is a read-only property.
+-validator.ts(38,4): error TS2540: Cannot assign to 'readonlyAccessor' because it is a read-only property.
+-validator.ts(39,1): error TS2322: Type 'number' is not assignable to type 'string'.
+-validator.ts(40,1): error TS2322: Type 'string' is not assignable to type 'number'.
+-validator.ts(41,1): error TS2322: Type 'number' is not assignable to type 'string'.
+-
+-
+-==== validator.ts (10 errors) ====
++index.js(4,11): error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
```
jakebailey: `mod1` looks like:

```
Object.defineProperty(exports, "thing", { value: 42, writable: true });
Object.defineProperty(exports, "readonlyProp", { value: "Smith", writable: false });
Object.defineProperty(exports, "rwAccessors", { get() { return 98122 }, set(_) { /*ignore*/ } });
Object.defineProperty(exports, "readonlyAccessor", { get() { return 21.75 } });
Object.defineProperty(exports, "setonlyAccessor", {
    /** @param {string} str */
    set(str) {
        this.rwAccessors = Number(str) 
    }
});
```

This is not being treated as a module for some reason? I don't think I did anything to affect this, probably another bug.

DanielRosenwasser:
> I'm also not setting up `maxNodeModuleJsDepth` for the the editor. I'm not sure where that goes.

That was previously set for `InferredProject`s that have at least one JavaScript file. I'm not sure if it is now.

jakebailey:
Theoretically, the tsconfig parser will set this default for `jsconfig` files, but that path does not appear to be used for the server, which has its own inferred options. In the new code, `createInferredProject` is the thing to modify, but I'm not sure how to get "there's a JS file" at that point; it's too early?

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
 	}
 
+	importsStart := len(moduleNames)
+
+	moduleNames = append(moduleNames, file.Imports()...)
+	for _, imp := range file.ModuleAugmentations {
+		if imp.Kind == ast.KindStringLiteral {
+			moduleNames = append(moduleNames, imp)
+		}
+		// Do nothing if it's an Identifier; we don't need to do module resolution for `declare global`.
```
DanielRosenwasser, line 372: I really dislike how global augmentations are module augmentations with an identifier.

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
-			// TODO(ercornel): !!!: check if from node modules
+			if !resolvedModule.IsResolved() {
+				continue
+			}
 
-			mode := getModeForUsageLocation(file.FileName(), meta, resolution.node, optionsForFile)
-			resolutionsInFile[module.ModeAwareCacheKey{Name: resolution.node.Text(), Mode: mode}] = resolution.resolvedModule
+			resolvedFileName := resolvedModule.ResolvedFileName
+			isFromNodeModulesSearch := resolvedModule.IsExternalLibraryImport
+			// If this is js file source of project reference, dont treat it as js file but as d.ts
```
DanielRosenwasser: I know this is from an existing comment, but can we fix it up so it's clearer what the intent is?

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
+			isJsFile := !tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsWithJsonFlat) && p.projectReferenceFileMapper.getRedirectForResolution(ast.NewHasFileName(resolvedFileName, p.toPath(resolvedFileName))) == nil
+			isJsFileFromNodeModules := isFromNodeModulesSearch && isJsFile && (resolvedFileName == "" || strings.Contains(resolvedFileName, "/node_modules/"))
```
DanielRosenwasser: When is `resolvedModuleName` the empty string?

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
-			if optionsForFile.GetResolveJsonModule() {
-				hasAllowedExtension = tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsWithJsonFlat)
-			} else if optionsForFile.AllowJs.IsTrue() {
-				hasAllowedExtension = tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedJSExtensionsFlat) || tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsFlat)
-			} else {
-				hasAllowedExtension = tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsFlat)
-			}
-			shouldAddFile := resolution.resolvedModule.IsResolved() && hasAllowedExtension
-			// TODO(ercornel): !!!: other checks on whether or not to add the file
+			shouldAddFile := moduleName != "" &&
```
DanielRosenwasser, line 407: When is `moduleName` supposed to be `""` other than when the user wrote that?

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
 	return toParse, resolutionsInFile, importHelpersImportSpecifier, jsxRuntimeImportSpecifier_
 }
 
-func (p *fileLoader) resolveModuleNames(entries []*ast.Node, file *ast.SourceFile, meta *ast.SourceFileMetaData, redirect *tsoptions.ParsedCommandLine) []*resolution {
-	if len(entries) == 0 {
-		return nil
+// Returns a DiagnosticMessage if we won't include a resolved module due to its extension.
+// The DiagnosticMessage's parameters are the imported module name, and the filename it resolved to.
+// This returns a diagnostic even if the module will be an untyped module.
+func getResolutionDiagnostic(options *core.CompilerOptions, resolvedModule *module.ResolvedModule, file *ast.SourceFile) *diagnostics.Message {
```
DanielRosenwasser, line 429: Is this only written this way because future code will eventually need the diagnostic?

internal/compiler/fileloader.go
jakebailey: Hey I just work here

internal/compiler/fileloader.go
DanielRosenwasser: Also- do we resolve modules based on module augmentations? That's how I'm reading this code.

internal/compiler/fileloadertask.go
```diff
@@ -1,43 +1,75 @@
...
+		w.wg.Queue(func() {
+			loadedTask.mu.Lock()
+			defer loadedTask.mu.Unlock()
+
+			if !task.isLoaded() {
+				task.load(loader)
+			}
+
+			if currentDepth < loadedTask.lowestDepth {
+				// If we're seeing this task at a lower depth than before,
```
DanielRosenwasser, line 66: How does this happen?

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
-				hasAllowedExtension = tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedJSExtensionsFlat) || tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsFlat)
-			} else {
-				hasAllowedExtension = tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsFlat)
-			}
-			shouldAddFile := resolution.resolvedModule.IsResolved() && hasAllowedExtension
-			// TODO(ercornel): !!!: other checks on whether or not to add the file
+			shouldAddFile := moduleName != "" &&
+				getResolutionDiagnostic(optionsForFile, resolvedModule, file) == nil &&
+				!optionsForFile.NoResolve.IsTrue() &&
+				!(isJsFile && !optionsForFile.GetAllowJS()) &&
```
DanielRosenwasser, line 410:
```suggestion
				(!isJsFile || optionsForFile.GetAllowJS()) &&
```

internal/compiler/fileloader.go
```diff
@@ -352,68 +355,118 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
...
+			if !resolvedModule.IsResolved() {
+				continue
+			}
 
-			mode := getModeForUsageLocation(file.FileName(), meta, resolution.node, optionsForFile)
-			resolutionsInFile[module.ModeAwareCacheKey{Name: resolution.node.Text(), Mode: mode}] = resolution.resolvedModule
+			resolvedFileName := resolvedModule.ResolvedFileName
+			isFromNodeModulesSearch := resolvedModule.IsExternalLibraryImport
+			// If this is js file source of project reference, dont treat it as js file but as d.ts
+			isJsFile := !tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedTSExtensionsWithJsonFlat) && p.projectReferenceFileMapper.getRedirectForResolution(ast.NewHasFileName(resolvedFileName, p.toPath(resolvedFileName))) == nil
```
DanielRosenwasser, line 396: Why isn't this just the following? Just a faithful translation?


```suggestion
			isJsFile := tspath.FileExtensionIsOneOf(resolvedFileName, tspath.SupportedJSExtensionsFlat) && p.projectReferenceFileMapper.getRedirectForResolution(ast.NewHasFileName(resolvedFileName, p.toPath(resolvedFileName))) == nil
```

internal/compiler/fileloader.go
jakebailey: This is a direct port, but I could change it.

internal/compiler/fileloadertask.go
jakebailey: Imagine:

```
A -> B -> C -> D -> E
     \---------^
```

If we walk A-B-C-D first and D exceeds the max depth (so is skipped, we'll eventually walk through B instead and have to redo D, and and then E, and so on.

internal/compiler/fileloader.go
jakebailey: Yes, all of this is a faithful port.

internal/compiler/fileloader.go
jakebailey: Yes, this code is supposed to be used in the checker, but for some reason is currently not.

internal/compiler/fileloader.go
jakebailey: Module name shouldn't be the empty string unless the user wrote it, yes.

We do resolve modules based on augmentation, it's just that augmenting a module and then failing to resolve that module is legal, as the augmentation would have no effect.

internal/compiler/fileloader.go
jakebailey: Probably never, and the original code just had `string | undefined` for a failed resolution. I just directly ported it, but could change it.

testdata/baselines/reference/submoduleAccepted/conformance/checkExportsObjectAssignProperty.errors.txt.diff
```diff
@@ -0,0 +1,147 @@
...
+             this.rwAccessors = Number(str) 
+         }
+     });
+     
+-==== mod2.js (0 errors) ====
++==== mod2.js (5 errors) ====
+     Object.defineProperty(module.exports, "thing", { value: "yes", writable: true });
++                          ~~~~~~
++!!! error TS2580: Cannot find name 'module'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
+     Object.defineProperty(module.exports, "readonlyProp", { value: "Smith", writable: false });
```
jakebailey, line 114: Seemingly this is broken because the new code does not yet handle `defineProperty` based exports. But, it's not clear to me why this succeeded in the past.

internal/compiler/fileloadertask.go
```diff
@@ -1,43 +1,75 @@
...
+				// reprocess its subtasks to ensure they are loaded.
+				loadedTask.lowestDepth = currentDepth
+				subTasks := task.getSubTasks()
+				w.start(loader, subTasks, currentDepth)
```
andrewbranch, line 70: The subtasks may or may not have been elided when they were seen at a lower depth—is there a mechanism here to skip them if they’ve already been run?

internal/compiler/parsetask.go
```diff
@@ -92,11 +96,30 @@ func (t *parseTask) redirect(loader *fileLoader, fileName string) {
 	t.subTasks = []*parseTask{{normalizedFilePath: tspath.NormalizePath(fileName), isLib: t.isLib}}
```
sheetalkamat, line 97: This needs to copy all the fields - increaseDepth, elideOnDepth etc

jakebailey: Copying it would mean redirects would double count depths, so this omission is intentional.

internal/compiler/fileloadertask.go
jakebailey: They'll be deduped by the `start` function; if they're at a lower depth the code in this block will skip walking any further, and the `loaded` check will have also skipped loading the source file and other info again.

Issue #1038, opened 2025-06-04
Hashbang comments are stripped from JS output


AlCalzone:
This is a followup to #541. Most cases were fixed when the issue was solved, except for hashbangs at the start of the file. Those are still omitted from the JS output, even when `removeComments` is `false`.

For example:
TS:
```ts
#!/usr/bin/env node

// ...code
```

JS:
```js
// ...code
```

PR #1193, opened 2025-06-15
Emit shebangs when printing

AlCalzone:
Fixes: https://github.com/microsoft/typescript-go/issues/1038

Ideally I'd like to add a test for this, but I have no clue how this works in this repo.


jakebailey:
> Ideally I'd like to add a test for this, but I have no clue how this works in this repo.

Most likely we already have a test, such that running `hereby test` and then `hereby baseline-accept` will update the baselines.

Otherwise, there are printer tests in the printer package you could add to.

AlCalzone:
Done. I wasn't sure how to interpret the diff, but after accepting the output files look correct to me.

jakebailey (approved):

PR #1195, opened 2025-06-16
Fix code that should have used IsRequireCall

jakebailey:
`IsVariableDeclarationInitializedToRequire` is not an equivalent replacement for `IsRequireCall`. They have differing behavior.

Noticed some of this in looking at new errors exposed by #1189.

Best to filter this PR to just diffs.


testdata/baselines/reference/submodule/conformance/jsDeclarationsTypeReferences2.symbols.diff
```diff
@@ -1,17 +1,14 @@
...
-@@= skipped -3, +3 lines =@@
- const{ a, m } = require("./something").o;
- >a : Symbol(a, Decl(index.js, 0, 6))
+@@= skipped -5, +5 lines =@@
  >m : Symbol(m, Decl(index.js, 0, 9))
-->require("./something").o : Symbol(o, Decl(something.ts, 0, 12))
-->require : Symbol(require)
+ >require("./something").o : Symbol(o, Decl(something.ts, 0, 12))
+ >require : Symbol(require)
 ->"./something" : Symbol("something", Decl(something.ts, 0, 0))
```
jakebailey, line 7: Still missing _something_ here, as this is mad about the actual string not having resolved to the other module (which would break something like go to def).

testdata/baselines/reference/submoduleAccepted/conformance/emitModuleCommonJS(module=commonjs).errors.txt.diff
```diff
@@ -16,10 +15,4 @@
 +!!! error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
```
jakebailey: This error is still here because `IsRequireCall` is missing the `requireStringLiteralLikeArgument` arg, which would normally let callers decide if the argument is even checked. Will try and update the PR to do this.

testdata/baselines/reference/submoduleAccepted/compiler/modulePreserve4.errors.txt.diff
```diff
@@ -12,29 +12,18 @@
...
 -/main3.cjs(2,1): error TS8002: 'import ... =' can only be used in TypeScript files.
--/main3.cjs(5,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--/main3.cjs(8,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--/main3.cjs(10,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--/main3.cjs(12,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--/main3.cjs(14,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--/main3.cjs(17,8): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
--
--
++/main3.cjs(1,13): error TS1293: ESM syntax is not allowed in a CommonJS module when 'module' is set to 'preserve'.
```
jakebailey, line 18: This is probably a piece of the checker that hasn't been updated to ignore reparsed exports.

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypeReferences2.symbols.diff
weswigham: It's probably still missing because `getSymbolAtLocation` doesn't support js declaration lookups yet (since it's bound entirely differently), or something similar.

weswigham (approved):

Issue #778, opened 2025-04-08
[bug] watch mode triggers continuously


tmm1:
```
❯ time tsgo tsc -noEmit -w -p src
build triggered due to config change
build starting at 2025-04-08 16:46:43.150158 -0700 PDT m=+0.937383709
build finished in 5.714491375s
build triggered due to config change
build starting at 2025-04-08 16:46:50.457319 -0700 PDT m=+8.244531876
build finished in 5.544717875s
build triggered due to config change
build starting at 2025-04-08 16:46:57.955792 -0700 PDT m=+15.742991834
build finished in 5.286759958s
build triggered due to config change
```

```diff
diff --git i/internal/execute/watcher.go w/internal/execute/watcher.go
index af20829e6..087daf754 100644
--- i/internal/execute/watcher.go
+++ w/internal/execute/watcher.go
@@ -1,6 +1,7 @@
 package execute
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/microsoft/typescript-go/internal/compiler"
@@ -50,6 +51,7 @@ func (w *watcher) hasErrorsInTsConfig() bool {
 			return true
 		}
 		if w.options.CompilerOptions() != configParseResult.CompilerOptions() {
+			fmt.Fprint(w.sys.Writer(), "build triggered due to config change", w.sys.NewLine())
 			w.configModified = true
 		}
 		w.options = configParseResult
@@ -73,12 +75,14 @@ func (w *watcher) hasBeenModified(program *compiler.Program) bool {
 		currState[fileName] = s.ModTime()
 		if !filesModified {
 			if currState[fileName] != w.prevModified[fileName] {
+				fmt.Fprint(w.sys.Writer(), "build triggered from ", fileName, ": ", w.prevModified[fileName], " -> ", currState[fileName], w.sys.NewLine())
 				filesModified = true
 			}
-			delete(w.prevModified, fileName)
 		}
+		delete(w.prevModified, fileName)
 	}
 	if len(w.prevModified) > 0 {
+		fmt.Fprint(w.sys.Writer(), "build triggered due to ", len(w.prevModified), " files deleted", w.sys.NewLine())
 		filesModified = true
 	}
 	w.prevModified = currState

```

cc @iisaduan

tmm1:
Probably due to pointer compare here?

https://github.com/microsoft/typescript-go/blob/74c01ca0f30c0117e927760871ffc0fef23cff5f/internal/execute/watcher.go#L52

[...]

PR #1197, opened 2025-06-16
Only compare source-file affecting changes in watcher configuration

ephemer:
Similar to the issue reported in https://github.com/microsoft/typescript-go/issues/778, #1156 appears to have introduced an issue such that `--watch` mode continuously rebuilds a project due to configuration changes, even if no config changes have occurred.

This PR aims to address that issue by only comparing the configuration subsets that affect source files, while ignoring differences in other parts. It appears from #1156 (although I did not fully understand the scope of those changes) that only the `SourceFileAffecting` changes are passed to the watcher command in the first place, so this change "should" be correct.

I don't claim to have deep knowledge of the full implications of the change otherwise however.

### DIFF
--- a/internal/execute/watcher.go
+++ b/internal/execute/watcher.go
@@ -54,7 +54,7 @@ func (w *watcher) hasErrorsInTsConfig() bool {
 			}
 			return true
 		}
-		if !reflect.DeepEqual(w.options.CompilerOptions(), configParseResult.CompilerOptions()) {
+		if !reflect.DeepEqual(w.options.CompilerOptions().SourceFileAffecting(), configParseResult.CompilerOptions().SourceFileAffecting()) {
 			// fmt.Fprint(w.sys.Writer(), "build triggered due to config change", w.sys.NewLine())
 			w.configModified = true
 		}


ephemer:
Tagging @jakebailey because of his involvement in both #778 and #1156

Also tagging @sheetalkamat because of her involvement in #1078, which appears to be tangentially related to this issue: the difference in configuration appears to come from the cache not being included in one of the configuration objects being compared, while it is included in another.

internal/execute/watcher.go
```diff
@@ -54,7 +54,7 @@ func (w *watcher) hasErrorsInTsConfig() bool {
 			}
 			return true
 		}
-		if !reflect.DeepEqual(w.options.CompilerOptions(), configParseResult.CompilerOptions()) {
```
jakebailey, line 57: This is definitely not correct, sorry. We need to be comparing all of the options, as using only this subset means we are missing many more options that matter for loading things or any checking option.

I entirely forgot that this code was here. I'll come up with a better fix.

ephemer: Thanks for looking into this @jakebailey!

internal/execute/watcher.go
```diff
@@ -54,7 +54,7 @@ func (w *watcher) hasErrorsInTsConfig() bool {
...
+		if !reflect.DeepEqual(w.options.CompilerOptions().SourceFileAffecting(), configParseResult.CompilerOptions().SourceFileAffecting()) {
```
jakebailey:
```suggestion
		// CompilerOptions contain fields which should not be compared; clone to get a copy without those set.
		if !reflect.DeepEqual(w.options.CompilerOptions().Clone(), configParseResult.CompilerOptions().Clone()) {
```

Actually, this should be good enough for a hotfix.

jakebailey: (and then retitle the pr)

ephemer: From what I could see, the issue was that one of the two `CompilerOptions` objects being compared had `<nil>` for one of its last fields (which I think is the config cache?), whereas the other one has the cache defined.

I don't see how `.Clone()` could fix this, unless `Clone` doesn't actually do what it says (and for example Clones everything except for the Cache)?

ephemer: In #778 and its subsequent fix, the issue appeared to be to use DeepEqual to make the comparison between these objects. In some sense, that appears to be the opposite fix to using `Clone`?

internal/execute/watcher.go
```diff
@@ -54,7 +54,7 @@ func (w *watcher) hasErrorsInTsConfig() bool {
 			}
 			return true
 		}
-		if !reflect.DeepEqual(w.options.CompilerOptions(), configParseResult.CompilerOptions()) {
+		if !reflect.DeepEqual(w.options.CompilerOptions().Clone(), configParseResult.CompilerOptions().Clone()) {
```
jakebailey, line 58:
```suggestion
		// CompilerOptions contains fields which should not be compared; clone to get a copy without those set.
		if !reflect.DeepEqual(w.options.CompilerOptions().Clone(), configParseResult.CompilerOptions().Clone()) {
```

Oops, meant for this to be the suggestion too.

DanielRosenwasser (approved):

internal/execute/watcher.go
ephemer: I tested your suggestion in practice and it works, but it doesn't make a lot of sense to me "why" from just looking at the code.

I updated the PR title accordingly

internal/execute/watcher.go
ephemer: Added it. That helps but it's a bit spooky to me that `.Clone()` does not produce an actual clone

ephemer:
Hi @DanielRosenwasser thanks for your review. I just updated the PR with an additional comment. Could you approve again?

internal/execute/watcher.go
jakebailey: It does produce an actual clone, it's just not a deep copy.

ephemer: I wonder if `DeepEqual` is the correct check if we're only interested in a shallow comparison? Doesn't bother me either way too much, I'm just happy there's a fix going in :)

jakebailey: It has to be `DeepEqual` so it descends into slices, maps, pointers. But for the purpose of this hack, we don't need to clone deeply, only ignore CompilerOptions' caches.

ephemer: Alright. Maybe the comment should reflect that more directly, but I wouldn't want to stop CI for it

Issue #1190, opened 2025-06-15
Erroneously recommending `import "./index.d.ts"` be turned into `import type "./index.d.ts"`


LukeAbby:
With `verbatimModuleSyntax` and `noUncheckedSideEffectImports` both set to true when I write `import "./index.d.ts"` I get an error of this form:
```ts
main.ts:1:8 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './types.js' instead?

1 import "./types.d.ts";
         ~~~~~~~~~~~~~~
```

Whereas in the stable version of TypeScript I get no such errors. `import type "./types.d.ts"` is simply a syntax error in both TypeScript-Go and stable TypeScript (as it should be asfik).

jakebailey:
Yeah, I think we're actually accidentally always checking these imports.

PR #1198, opened 2025-06-16
Assume clause-less imports to be unemittable

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/1190

### DIFF
--- a/testdata/tests/cases/conformance/moduleResolution/allowImportingTypesDtsExtension2.ts
+++ b/testdata/tests/cases/conformance/moduleResolution/allowImportingTypesDtsExtension2.ts
@@ -1,0 +1,13 @@
+// @strict: true
+// @verbatimModuleSyntax: true,false
+// @noUncheckedSideEffectImports: true,false
+// @noEmit: true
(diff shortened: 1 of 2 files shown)


testdata/tests/cases/conformance/moduleResolution/allowImportingTypesDtsExtension2.ts
Andarist, line 1: For some reason, this doesn't create new baselines. cc @jakebailey

jakebailey: I'd assume not, you disabled emit and types?

Andarist: Ah, this makes sense - dummy me. I'm not used to `noTypesAndSymbols` - but the other test like this used this so I just copied it over. This leaves us with only errors and there are no errors here with the fix so there are no baselines either.

All good 👍

jakebailey (approved):

PR #1205, opened 2025-06-16
Remove ScriptTarget from source affecting options

jakebailey:
This warrants some discussion, but after #1203, the language version is actually no longer needed at all for parsing, and can be dropped from binding if we assume no ES5 support. This means we could share ASTs no matter which language version they're used with, which is pretty neat.

The only gotcha is that the scanner is still missing the regex parse error code, and the language version is needed there to give correct errors.

Honestly, I think we should do this PR to get even more AST reuse, and then go put the regex checking code elsewhere, like in the checker itself or something.


internal/scanner/scanner.go
```diff
@@ -218,7 +218,6 @@ type ScannerState struct {
 type Scanner struct {
 	text string
 	// TODO: remove this? unused except for (unimplemented) regex checking
```
```suggestion
```

internal/scanner/scanner.go
```diff
@@ -231,7 +229,7 @@ func defaultScanner() Scanner {
 	// Using a function rather than a global is intentional; this function is
 	// inlined as pure code (zeroing + moves), whereas a global requires write
 	// barriers since the memory is mutable.
-	return Scanner{languageVersion: core.ScriptTargetLatest, skipTrivia: true}
+	return Scanner{skipTrivia: true}
```
jakebailey, line 223: If I make skipTrivia a method, then flip the default, then we can also reset scanners with zeroing, which would be amazing.

jakebailey:
The verdict was "do this, and keep removing options" plus "reintroduce regexp checking as grammar checks".

PR #1211, opened 2025-06-17
Unskip JS emit tests

jakebailey:
After #1189, we no longer fall over in JS code. Unskip all of these tests so we can start working on them and catch any crashes.


testdata/baselines/reference/conformance/deepElementAccessExpressionInJS.js
```diff
@@ -0,0 +1,24 @@
...
+//// [elementAccessExpressionInJS.js]
+if (module[calculatePropertyName(1)]) {
+}
+function calculatePropertyName(index) {
+    // this would be some webpack index in real life
+    return `property${index}`;
+}
+
+
+//// [elementAccessExpressionInJS.js]
```
jakebailey, line 18: Something is wrong here; these are not declaration files, these are JS files.

weswigham: For sure, double js emit and no declaration file seems pretty wrong.

jakebailey:
Yeah, I want to look at this a little harder before, because it's failing race checks due to emit ordering problems, which is I think must be fixed before we are able to merge it.

testdata/baselines/reference/conformance/deepElementAccessExpressionInJS.js
jakebailey: Oh, this is just the input JS + the output JS. This test does not have declaration emit enabled at all.

jakebailey:
Okay, I triaged the two failures (and cleaned up skipped tests a little bit). Theoretically this should pass CI.

PR #1213, opened 2025-06-17
Set maxNodeModuleJsDepth in JS-containing projects

### LINKED ISSUE #1209: Set maxNodeModuleJsDepth in editor
After #1189, `maxNodeModuleJsDepth` now actually works. But, we're not setting it to a higher number in the editor. We need to do so on inferred projects, presumably only those containing JS files?

jakebailey:
Could use a test, but this looks correct to me in theory.

Fixes #1209


internal/project/project.go
```diff
@@ -561,6 +562,12 @@ func (p *Project) updateProgram() bool {
 			} else {
 				rootFileNames := p.GetRootFileNames()
 				compilerOptions := p.compilerOptions
+
+				if compilerOptions.MaxNodeModuleJsDepth == nil && p.rootJSFileCount > 0 {
+					compilerOptions = compilerOptions.Clone()
```
sheetalkamat, line 577: There is method GetCompilerOptions - do you need to modify that instead and scan through other usage of p.compilerOptions ?

jakebailey: That method is actually unused except for the API package, so I opted to not modify that. I could theoretically do that, though, yes.

internal/project/project.go
```diff
@@ -943,12 +949,32 @@ func (p *Project) setRootFiles(rootFileNames []string) {
...
-				p.rootFileNames.Delete(root)
+				p.deleteRootFileName(root)
 			}
 		}
 	}
 }
 
+func (p *Project) setRootFileName(path tspath.Path, fileName string) {
+	p.rootFileNames.Set(path, fileName)
+	if p.kind == KindInferred {
```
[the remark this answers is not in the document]
jakebailey: this is definitely true, I'll have to fix it.

internal/project/project.go
jakebailey: Tried, but the concurrency story on this is pretty weird. I'd rather just leave this here for now and fix it if we need this option set for something else eventually. It'll probably get easier with the snapshot refactor.

internal/project/project.go
```diff
@@ -929,7 +935,7 @@ func (p *Project) setRootFiles(rootFileNames []string) {
 		// !!! updateNonInferredProjectFiles uses a fileExists check, which I guess
 		// could be needed if a watcher fails?
 		newRootScriptInfos[path] = struct{}{}
-		p.rootFileNames.Set(path, file)
+		p.setRootFileName(path, file)
```
sheetalkamat: this is used only by configured project instead of changing this may be rename the methods so its clear which project kind they are operating on

sheetalkamat (changes requested):

internal/project/project.go
```diff
@@ -964,6 +970,29 @@ func (p *Project) setRootFiles(rootFileNames []string) {
 	}
 }
 
+func (p *Project) setRootFileNameOfInferred(path tspath.Path, fileName string) {
+	has := p.rootFileNames.Has(path)
+	p.rootFileNames.Set(path, fileName)
+	if p.kind == KindInferred {
```
sheetalkamat: no need to check this ?

jakebailey: I feel like I should at least assert it for these functions? There's no guards elsewhere.

sheetalkamat (approved):

PR #1214, opened 2025-06-17
Fix nested dist when referenced package has a `.ts` file next to a `.js` file

maschwenk:
Reproduction case:

- https://github.com/dimitropoulos/tsgo-repros/pull/2

I prompted Cursor with:

```
I'm having an issue where if we have this structure

package A
   depends on package B and imports from package B
package B
   has a dist folder with both an dist/index.js and dist/index.ts

when I build package A, the dist folder of A becomes nested. the Go implementation behaves this way but the javascript implementation doesn't. Can you figure this out?
```

and after a few turns got this, so still working out if this is valid or not but it does fix my issue.

Simpler fix for this is just not having `.ts` files in `dist`, which is generally a good thing in general, but I was just curious why this was behaving differently in strada/corsa.


jakebailey:
I'm not sure what's going on here; this code is very much unlike the original code. I think the fix is going to look very different to this.

maschwenk:
Per Discord, we've determined that the actual root cause of the issue here is the differences in implementation of 

`_submodules/TypeScript/src/compiler/utilities.ts`'s `sourceFileMayBeEmitted` versus

`internal/compiler/emitter.go`'s `sourceFileMayBeEmitted`

I believe the actual case that should be returning false here is one of these:

```ts
    // Source file from node_modules are not emitted
    if (host.isSourceFileFromExternalLibrary(sourceFile)) return false;
    // forcing dts emit => file needs to be emitted
    if (forceDtsEmit) return true;
    // Check other conditions for file emit
    // Source files from referenced projects are not emitted
    if (host.isSourceOfProjectReferenceRedirect(sourceFile.fileName)) return false;
```

and I'm fairly sure it's `isSourceFileFromExternalLibrary` which is still not implemented in Go-land

jakebailey:
All of the code needed to implement `isSourceFileFromExternalLibrary` is actually there since #1189, actually. We just aren't collecting that list of files.

PR #1217, opened 2025-06-17
Watch type refs and instead of watching directly use go routine

sheetalkamat:


jakebailey (commented):
The file loader code seems correct to me, though I am unsure about the concurrency stuff in the Project given our ongoing logical races.

I assume no tests change with the loader change since the main effect here is to actually keep track of these files for watching?

jakebailey:
Is this still needed after #1505?

PR #1218, opened 2025-06-17
Refactor some of the code with lsp and api layer to use DocumentStore directly now that it exists

sheetalkamat:


internal/project/util_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/project"
+	"github.com/microsoft/typescript-go/internal/tspath"
+	"gotest.tools/v3/assert"
+)
+
+func configFileExists(t *testing.T, service *project.Service, path tspath.Path, exists bool) {
+	_, loaded := service.ConfigFileRegistry().ConfigFiles.Load(path)
```
jakebailey, line 13: I recommend calling `t.Helper()` in this func, to move the assert out to the caller.

PR #1223, opened 2025-06-18
Don't change `global` keyword to `namespace` in `transformModuleDeclaration`

Andarist:


jakebailey:
I think this is probably fine, but does require a merge from main and baseline update.

PR #1227, opened 2025-06-18
Add CI job for checking generated fourslash tests

gabritto:
Have CI check that the fourslash tests generated by the conversion script match the ones checked into the repo.
This should be temporary while we're still working on the conversion script.


.github/workflows/ci.yml
```diff
@@ -355,11 +355,61 @@ jobs:
           name: fix_baselines.patch
           path: fix_baselines.patch
 
+  checkfourslash:
```
jakebailey: Does this need to be new? Can you just stick this into the `generate` job?

gabritto: I moved it. Just didn't occur to me that we already had that job.

jakebailey (approved):

PR #1231, opened 2025-06-18
Order control flow loop and branch antecedents the same as Strada

### LINKED ISSUE #987: New implicit any error from self referencing in initializer
I'm not sure if this is a bug in the current or in the new compiler, but there is a scenario where some amount of logic involved determining the type from an initializer is showing an error in the new compiler where it doesn't in the current compiler:

```ts
class A {
  next: A | null = null;

  constructor(readonly children: (A | null)[]) {}
}

function getNodes(): A[] {
  const out: A[] = [];

  let current: A | null = new A([]);

  while (current !== null) {
    let firstChild = null;

    // If the following if block is commented out, then current TS also shows an error
    if (out.length) {
      current = current.next;
      continue;
    }

    for (let i = 0; i < current.children.length; i++) {
      const child = current.children[i];

      if (child) {
        if (!firstChild) {
          firstChild = child;
          firstChild.next = current.next;
        }

        child.next = current.next;
      }
    }

    current = firstChild || current.next;
  }

  return out;
}
```

(example is nonsensical, just tried to reduce it to a minimal repro from the real code)

> src/indirect-init.ts:21:13 - error TS7022: 'child' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.

When I comment out the if condition it gives this error in both old and new compiler.

[playground](https://www.typescriptlang.org/play/?#code/MYGwhgzhAECC0G8BQ1oDsCmAPALgLjmgB90BXEEaAXjIoG4kVpgB7NCHAJ1OBxc4AUnDGAAmbEAE9mACwCWIUcLQEB8EmnIgAlAG0AutsQBfJKaQAzUml5y20AOYYcAORaiMEAdoKwDiJlZ2HGgWUnw4fxoDBiYQZ2ZSTmUI9VpKGkwAdzgBA21Y1Cz5eOgBYCSU6ABCKkytI2RUVHiQizlODgBhEtFqdMLmgHoh6ABJC2gcGQxoCxYKFiy5NAdoOUmAIxAWYABrdZhWAFtjjDQcDD6wnAAaKZm0ROTzkIAVAGVoMBAIFmgIDIljAwE8MMl+ExUBsyjcAHTxVbTRpQ5oVF4XfrolJwzC4BjNNFsHArUgYAnNcyE+acMqtdb9AAMdAZAB5njjgL1lAjzg5piy5ABqIUowmoIIcWQKPo0bGvOFcmXKXRyfSDQkw8q9MXi6GTATVdqdHA9GW6vWoY3dXpY3oUy1zDo2mW47AhOWVBV4nAO8VUy1KxRu3BYr0XEO+1GoUyEgMS8Mep0ms2KYgkeURn0UgPCHBJJ43BimIA)

### LINKED ISSUE #1229: typescript-go errors with "too large for control flow analysis" when tsc does not
Image

I should note this example is the minimal size at which `tsgo` starts complaining when `tsc` does not. I think the upper size that tsc would accept is much higher than this

Repro: https://github.com/issacgerges/tsgo-too-large-control-flow

ahejlsberg:
Fixes #987.
Fixes #1229.

### DIFF
--- a/internal/binder/binder.go
+++ b/internal/binder/binder.go
@@ -550,21 +550,24 @@ func setFlowNodeReferenced(flow *ast.FlowNode) {
 	}
 }
 
-func hasAntecedent(list *ast.FlowList, antecedent *ast.FlowNode) bool {
-	for list != nil {
-		if list.Flow == antecedent {
-			return true
-		}
-		list = list.Next
-	}
-	return false
-}
-
 func (b *Binder) addAntecedent(label *ast.FlowLabel, antecedent *ast.FlowNode) {
-	if antecedent.Flags&ast.FlowFlagsUnreachable == 0 && !hasAntecedent(label.Antecedents, antecedent) {
-		label.Antecedents = b.newFlowList(antecedent, label.Antecedents)
-		setFlowNodeReferenced(antecedent)
+	if antecedent.Flags&ast.FlowFlagsUnreachable != 0 {
+		return
 	}
+	// If antecedent isn't already on the Antecedents list, add it to the end of the list
+	var last *ast.FlowList
+	for list := label.Antecedents; list != nil; list = list.Next {
+		if list.Flow == antecedent {
+			return
+		}
+		last = list
+	}
+	if last == nil {
+		label.Antecedents = b.newFlowList(antecedent, nil)
+	} else {
+		last.Next = b.newFlowList(antecedent, nil)
+	}
+	setFlowNodeReferenced(antecedent)
 }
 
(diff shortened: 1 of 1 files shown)


ahejlsberg:
I don't think we need additional tests since all this does is to eliminate a difference with Strada.

jakebailey (approved):
Both of the linked issues have minimal test cases, but we probably wouldn't have even experienced this if the code were ported this way originally, so, yeah... 😄

PR #1070, opened 2025-06-05
Port `node18` and `nodenext` changes

andrewbranch:
Ports
- https://github.com/microsoft/TypeScript/pull/60722
- https://github.com/microsoft/TypeScript/pull/60761

PR #1234, opened 2025-06-18
Port IsSourceFileFromExternalLibrary, fix node18 emit

### LINKED ISSUE #1215: When referenced external dependency contains `.ts` alongside `.js` of the same name, builds are nested
Repro case here:

- https://github.com/dimitropoulos/tsgo-repros/pull/2

Where a fairly vanilla monorepo setup created a nested `dist`. Solve-able by just NOT having `.ts` files in `dist` (and this is good practice) but it is a weird regression case.

A (very rough) attempt at a fix is here:

- https://github.com/microsoft/typescript-go/pull/1214

maschwenk:
Fixes #1215


internal/project/project.go
```diff
@@ -970,6 +970,7 @@ func (p *Project) GetFileNames(excludeFilesFromExternalLibraries bool, excludeCo
 	result := []string{}
 	sourceFiles := p.program.GetSourceFiles()
 	for _, sourceFile := range sourceFiles {
+		// !! This is probably ready to be implemented now?
```
maschwenk, line 986: Because we have `IsSourceFileFromExternalLibrary`

internal/compiler/program.go
```diff
@@ -820,6 +831,16 @@ func (p *Program) getModeForTypeReferenceDirectiveInFile(ref *ast.FileReference,
 	return p.GetDefaultResolutionModeForFile(sourceFile)
 }
 
+// IsSourceFileFromExternalLibrary returns true if the source file is from an external library.
+// This mirrors the TypeScript implementation which tracks files found while searching node_modules.
```
jakebailey: I assume some LLM added the comments; I would just remove all of these as they just describe exactly what the code says.

internal/compiler/fileloader.go
```diff
@@ -412,9 +428,10 @@ func (p *fileLoader) resolveImportsAndModuleAugmentations(file *ast.SourceFile,
 
 			if shouldAddFile {
 				toParse = append(toParse, resolvedRef{
-					fileName:      resolvedFileName,
-					increaseDepth: resolvedModule.IsExternalLibraryImport,
-					elideOnDepth:  isJsFileFromNodeModules,
+					fileName:              resolvedFileName,
+					increaseDepth:         resolvedModule.IsExternalLibraryImport,
+					elideOnDepth:          isJsFileFromNodeModules,
+					isFromExternalLibrary: isFromNodeModulesSearch,
```
jakebailey: Not just `resolvedModule.IsExternalLibraryImport`?

internal/compiler/fileloader.go
```diff
@@ -301,7 +313,10 @@ func (p *fileLoader) resolveTripleslashPathReference(moduleName string, containi
 		referencedFileName = tspath.CombinePaths(basePath, moduleName)
 	}
 	return resolvedRef{
-		fileName: tspath.NormalizePath(referencedFileName),
+		fileName:              tspath.NormalizePath(referencedFileName),
+		increaseDepth:         false,
+		elideOnDepth:          false,
+		isFromExternalLibrary: false, // Triple-slash references are always local files
```
jakebailey: These are all defaulted to `false` if not present.

internal/compiler/program.go
```diff
@@ -820,6 +831,16 @@ func (p *Program) getModeForTypeReferenceDirectiveInFile(ref *ast.FileReference,
...
+func (p *Program) IsSourceFileFromExternalLibrary(file *ast.SourceFile) bool {
+	if file == nil {
+		return false
+	}
+	isExternal, exists := p.sourceFilesFoundSearchingNodeModules.Load(file.Path())
+	return exists && isExternal
```
jakebailey: I don't think this should have `Load`. It's just a `map`.

internal/compiler/fileloader.go
```diff
@@ -50,6 +50,9 @@ type processedFiles struct {
 	importHelpersImportSpecifiers map[tspath.Path]*ast.Node
 	// List of present unsupported extensions
 	unsupportedExtensions []string
+	// Track which source files were found while searching node_modules
+	// Similar to TypeScript's sourceFilesFoundSearchingNodeModules map
+	sourceFilesFoundSearchingNodeModules map[tspath.Path]bool
```
jakebailey: Use `collections.Set[tspath.Path]` here.

internal/compiler/program.go
```diff
@@ -203,6 +205,14 @@ func NewProgram(opts ProgramOptions) *Program {
 
 	p.processedFiles = processAllProgramFiles(p.opts, libs, p.singleThreaded())
 
+	// Populate sourceFilesFoundSearchingNodeModules from the processed files
+	// This mirrors the TypeScript implementation's tracking behavior
+	if p.processedFiles.sourceFilesFoundSearchingNodeModules != nil {
+		for path, isExternal := range p.processedFiles.sourceFilesFoundSearchingNodeModules {
+			p.sourceFilesFoundSearchingNodeModules.Store(path, isExternal)
+		}
+	}
```
jakebailey: This should not be required, the info should already be on processedFiles and therefore be accessible.

testdata/baselines/reference/submodule/conformance/typingsLookup4.js.diff
```diff
@@ -1,6 +1,19 @@
...
++"use strict";
++Object.defineProperty(exports, "__esModule", { value: true });
++exports.l = void 0;
++exports.l = 2;
++//// [index.js]
++"use strict";
++Object.defineProperty(exports, "__esModule", { value: true });
++exports.m = void 0;
++exports.m = 3;
  //// [a.js]
```
jakebailey: These files appearing indicates that there's still some sort of bug here.

jakebailey:
I still have more stuff to push; there are other weird things here that need to be fixed before this is ready.

internal/compiler/parsetask.go
```diff
@@ -98,14 +102,21 @@ func (t *parseTask) redirect(loader *fileLoader, fileName string) {
...
 
 func (t *parseTask) addSubTask(ref resolvedRef, isLib bool) {
 	normalizedFilePath := tspath.NormalizePath(ref.fileName)
-	subTask := &parseTask{normalizedFilePath: normalizedFilePath, isLib: isLib, increaseDepth: ref.increaseDepth, elideOnDepth: ref.elideOnDepth}
+	subTask := &parseTask{
+		normalizedFilePath:    normalizedFilePath,
+		isLib:                 isLib,
+		increaseDepth:         ref.increaseDepth,
+		elideOnDepth:          ref.elideOnDepth,
+		isFromExternalLibrary: ref.isFromExternalLibrary || t.isFromExternalLibrary, // Propagate external library status
```
sheetalkamat: Not enough and needs to be handled differently to make sure https://github.com/microsoft/TypeScript/blob/main/src/compiler/program.ts#L3620 part is working as well

jakebailey: The redoing is what the `start` block does with the depth, but we do need to reexamine things, yes.

sheetalkamat: But here we are propagating is from external import .. we would need to change those other tasks too ?

jakebailey: Yes, that's the super annoying thing I'm trying to fix now 😭

jakebailey:
Okay, well that was annoying

testdata/baselines/reference/submodule/conformance/nodeModulesExportsSourceTs(module=node18).js.diff
```diff
@@ -1,29 +1,37 @@
...
-     "exports": "./index.ts"
+@@= skipped -27, +27 lines =@@
  }
 
--//// [other.js]
+ //// [other.js]
 -export const x = null;
--//// [index.js]
--// esm format file
++"use strict";
```
jakebailey: This is broken because node18 is not implemented.

jakebailey: Hm, but #1070 should have done it. Need to recheck.

jakebailey: Ah, one liner.

weswigham (approved):

PR #1237, opened 2025-06-19
Handle `@captureSuggestions` in the harness

Andarist:
I was surprised only a single baseline got updated but it makes sense - this option was introduced somewhat recently, here: https://github.com/microsoft/TypeScript/pull/59519


jakebailey (approved):
Thanks, been meaning to do this!

PR #1238, opened 2025-06-19
Add support for workspace

sheetalkamat:


internal/lsp/server.go
```diff
@@ -591,12 +602,20 @@ func (s *Server) handleInitialized(ctx context.Context, req *lsproto.RequestMess
 		s.projectService.SetCompilerOptionsForInferredProjects(s.compilerOptionsForInferredProjects)
 	}
 
+	if s.initializeParams.RootUri.Value != "" {
+		s.projectService.Workspace.SetRoot(ls.DocumentURIToFileName(s.initializeParams.RootUri.Value))
+	}
+	if s.initializeParams.WorkspaceFolders != nil {
+		for _, folder := range s.initializeParams.WorkspaceFolders.Value {
+			s.projectService.Workspace.AddFolder(ls.DocumentURIToFileName(lsproto.DocumentUri(folder.Uri)))
```
jakebailey, line 610: Do you know why we need to handle this kind of thing? Typically this stuff is only ever used if we plan to explicitly handle multi-root workspaces, but I don't think we have a reason to do that, since we're great at just accepting whatever gets opened anyway?

sheetalkamat: this was used for creating better inferred projects  - thats when projectRootPath was introduced
It is also used to not lookup tsconfig beyond the project root locations. I added this because with multi project thing the search locations increase and wanted to make sure i am not going beyond the point it needs to.

PR #1241, opened 2025-06-19
fix(1196): extend parameter matching with index fallback when name matching fails

### LINKED ISSUE #1196: Missing optional JSDoc param parsing
Hello, 

we are in the process in porting our jsx codebase to ts, we are getting there, but there are still some JS[X] files that need to be rewritten

In the mean time, we are relying on jsdoc to [very] pseudo type the remaining files, 

with something like this :
```
/**
 * @param {Object} props
 * @param {string|null} [props.hint]
 * @param {ReactNode} props.value
 * @param {number} [props.iconSize]
 * @param {string} [props.className]
 * @param {boolean} [props.preventDefault]
 * @param {boolean} [props.external]
 * @param {string} [props.schemaBuiltinKey]
 *
 */
const AttributeStringHint = ({ hint, value, iconSize = 14, className, preventDefault, external, schemaBuiltinKey }) => {
```
note the [ ] , that mark the prop as optional here in the react component

then if i can from a tsx file like this 
```	<AttributeStringHint key={index} hint={attr.hint} value={value} />```
i got with the current preview

```
Attribute.tsx:160:8 - error TS2739: Type '{ key: number; hint: SchemaAttributeHint | null | undefined; value: string; }' is missing the following properties from type '{ className: any; external: any; hint: any; iconSize?: number | undefined; preventDefault: any; schemaBuiltinKey: any; value: any; }': className, external, preventDefault, schemaBuiltinKey

160       <AttributeStringHint key={index} hint={attr.hint} value={value} />
```

in current ts, everything is ok of course

### LINKED ISSUE #1225: Destructured JSDoc Arguments No Longer Collectively Typeable Using Arbitrary Name
On Strada this snippet works fine:
```js
/**
 * @param {{ foo: number; bar: string }} args
 */
function foo({ foo, bar }) {}
```

Resulting in this `.d.ts` file:
```ts
/**
 * @param {{ foo: number; bar: string }} args
 */
declare function foo({ foo, bar }: {
    foo: number;
    bar: string;
}): void;
```

However in Corsa it appears that this has no effect and `foo` and `bar` are typed as `any`.

CC @sandersn. Was this something intentionally left out of the JSDoc port or just something not yet implemented?

a-tarasyuk:
Fixes #1196
Fixes #1225

--- 

ℹ️ This patch extends parameter matching by adding a [fallback to index-based resolution](https://github.com/microsoft/TypeScript/blob/479285d0ac293c35a926508d17f0bb5eca7e0303/src/compiler/utilitiesPublic.ts#L1041-L1047) when JSDoc parameter tag names do not match function parameter names, and also improves handling of nested array literals.

---

:white_check_mark: Verification test - [jsdocParamTagTypeLiteral](https://github.com/microsoft/TypeScript/blob/479285d0ac293c35a926508d17f0bb5eca7e0303/tests/cases/conformance/jsdoc/jsdocParamTagTypeLiteral.ts
)


testdata/baselines/reference/submoduleAccepted/conformance/paramTagNestedWithoutTopLevelObject.types.diff
```diff
@@ -0,0 +1,17 @@
...
+ function g(xyz) {
+->g : (xyz: any) => any
+->xyz : any
++>g : (xyz: number) => any
++>xyz : number
+
+     return xyz.p;
+ >xyz.p : any
+->xyz : any
++>xyz : number
```
jakebailey: This seems suspect?

a-tarasyuk: This should be treated as an error [paramTagNestedWithoutTopLevelObject](https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/paramTagNestedWithoutTopLevelObject.errors.txt)

However, it seems that `checkUnmatchedJSDocParameters` hasn’t been migrated yet.

testdata/baselines/reference/submoduleAccepted/compiler/jsdocParamTagInvalid.types.diff
```diff
@@ -0,0 +1,10 @@
+--- old.jsdocParamTagInvalid.types
++++ new.jsdocParamTagInvalid.types
+@@= skipped -2, +2 lines =@@
+ === /a.js ===
+ /** @param {string} colour */
+ function f(color) {}
+->f : (color: any) => void
+->color : any
++>f : (color: string) => void
++>color : string
```
jakebailey: Perhaps this is like the other comment, but why is this ending up succeeding? I'm confused how fixing a missing error (in another PR) would cause this to flip back to `any`.

testdata/baselines/reference/submoduleAccepted/conformance/paramTagNestedWithoutTopLevelObject.types.diff
a-tarasyuk: I didn't mean it eliminates type changes, I only meant to point out that certain cases involving qualified names are not yet covered.

jakebailey: Yeah. I mean, this all seems like an improvement in any case. But, will leave to @sandersn to double check.

jakebailey:
This PR needs a merge from main after other PRs were merged.

testdata/baselines/reference/submoduleAccepted/conformance/jsdocParamTagTypeLiteral.types.diff
```diff
@@ -5,177 +5,113 @@
...
-->opts5[0].what.bad[0] : { idea: string; oh: boolean; }
-->opts5[0].what.bad : { idea: string; oh: boolean; }[]
+ >opts5[0].what.bad[0].idea : string
+ >opts5[0].what.bad[0] : { idea: string; oh: boolean; }
+ >opts5[0].what.bad : { idea: string; oh: boolean; }[]
 ->opts5[0].what : { a: string; bad: {    idea: string;    oh: boolean;}; }
 ->opts5[0] : { help: string; what: {    a: string;    bad: {        idea: string;        oh: boolean;    };}; unnest: number; }
 ->opts5 : { help: string; what: {    a: string;    bad: {        idea: string;        oh: boolean;    };}; unnest: number; }[]
-->0 : 0
++>opts5[0].what : { a: string; bad: { idea: string; oh: boolean; }[]; }
```
sandersn, line 91: `bad` being an array looks correct; I just can't figure out how Strada thought it *wasn't*

sandersn:
Since there's a merge needed, could you also delete the unused `host` parameter from `reparseJSDocTypeLiteral`?

jakebailey (approved):

PR #1242, opened 2025-06-19
Port implicit `any` type arguments in JavaScript

### LINKED ISSUE #1090: Using a class based react component imported from a `jsx` files produce errors
repro: https://github.com/joshcartme/tsgo-jsx-prop-bug

Under tsgo, using a class based react component imported from a `jsx` files produce errors complaining that the props are not assignable to the component's type.

In the output below `MyComponent` is a class based component imported from `./component.jsx`

tsgo:

```
$ yarn run tsgo

index.tsx:6:25 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type '{ text: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<MyComponent> & Readonly<{}>'.
      Property 'text' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<MyComponent> & Readonly<{}>'.

6     return <MyComponent text="hello" />
                          ~~~~

  node_modules/@types/react/index.d.ts:1021:9 - The last overload is declared here.
    1021         constructor(props: P, context: any);
                 ~~~~~~~~~~~

Found 1 error in index.tsx:6

```

tsc:

```
$ yarn run tsc
Done in 1.93s.
```

DanielRosenwasser:
This restores the behavior of filling in missing type arguments (and default type arguments set to `{}` and `unknown`) as `any` in JavaScript files.

It should fix the behavior in #1090, but and unbreaks JS code I've tested elsewhere where things like `new Set` result in a `Set<unknown>` instead of a `Set<any>`.

Some of this is already partially controlled by flags like `noImplicitAny` - however, I do wonder if we should consider adding another `--strict` flag for this behavior in JavaScript files. My intuition is that JS devs using checking and `--strict` *do* want the same behavior as TypeScript, and would opt in for tighter checking. The use of JS is circumstantial around their build needs/preferences, not a signal of whether they want looser checking.

Fixes #1090


DanielRosenwasser:
Minimal repro for the crash is

```js
// @filename: foo.js
export class A<T> {}

export class B extends A {
    constructor() {
        super();
    }
}
```

I haven't fixed it in the latest commits, but I was in the neighborhood and found a few places I had missed (or might as well have cleared out some diffs)

jakebailey (approved):
Based on the diffs, this sure seems correct to me. I'm not sure if this stuff was intentionally removed or not, though, but I feel like this should be here. Maybe @sanders has an opinion, however.

(Also, baselines need a reup.)

testdata/baselines/reference/submodule/conformance/jsDeclarationsMissingGenerics.js.diff
```diff
@@ -10,4 +10,27 @@
...
 -declare function y(x: Promise<any>): void;
-+declare function y(x: Promise): void;
\ No newline at end of file
++declare function y(x: Promise): void;
++
++
++//// [DtsFileErrors]
++
++
++out/file.d.ts(4,23): error TS2314: Generic type 'Array<T>' requires 1 type argument(s).
```
jakebailey, line 19: Hm, this is exciting; wonder how easy it'd be to fix this?

DanielRosenwasser: @weswigham can correct me, but I think this needs `symbolTableToDeclarationStatements` and others before we expect JS `.d.ts` to *really* work - see `existingTypeNodeIsNotReferenceOrIsReferenceWithCompatibleTypeArgumentCount`, `canReuseTypeNode` etc.

weswigham: Actually, not so - since we changed how JS code is parsed, we may _never_ implement the `symbolTableToDeclarationStatements` stub and might just remove it (which is maybe a problem for expandable hover, since that also leverages it). Since JS is syntactically transformed into something the checker likes now, similarly, the declaration emitter aughta just have those special syntaxes handled directly in declaration emit rather than always falling back to a semantic printback - which basically just boils down to handling the new reparsed node structures and kinds in the transform itself. I added a draft of whatever js support was in while I was working on the initial declaration emitter, but it really does just need someone to go in and add the entirely new emit logic.

weswigham: In this case, the declaration emitter is picking up that we now just shove the type nodes directly on the `.Type` of the parameter and mostly just doing the right thing, but the declaration emitter is missing the logic to ensure minimum type argument counts are met that the node builder used to handle implicitly. That's a slightly semantic question ("how many type arguments should this reference have, at a minimum?"), so it'll probably need to get answered through a new `emitResolver` hook.

jakebailey (approved):

PR #1245, opened 2025-06-20
Finish implementing sourceFileMayBeEmitted, JSON emit

### LINKED ISSUE #660: JSON file includes do not change lowest common ancestor for outDir
Example project to reproduce the issue: [repro.zip](https://github.com/user-attachments/files/19317861/repro.zip)

Result when running `tsgo` here:
```
build\
  index.js
```

Expected result / what `tsc` outputs:
```
build\
  static\
    data.json
  src\
    index.js
```

jakebailey:
Fixes #660

I'm not happy with the emit change, but I don't know what the right fix is. Strada did parenthesizing completely differently.


internal/printer/printer.go
```diff
@@ -3175,7 +3175,10 @@ func (p *Printer) emitEmptyStatement(node *ast.EmptyStatement, isEmbeddedStateme
 func (p *Printer) emitExpressionStatement(node *ast.ExpressionStatement) {
 	state := p.enterNode(node.AsNode())
 
-	if isImmediatelyInvokedFunctionExpressionOrArrowFunction(node.Expression) {
+	if p.currentSourceFile != nil && p.currentSourceFile.ScriptKind == core.ScriptKindJSON {
```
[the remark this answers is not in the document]
jakebailey: Ah yes, it's that simple

internal/compiler/emitter.go
```diff
@@ -323,8 +340,19 @@ func sourceFileMayBeEmitted(sourceFile *ast.SourceFile, host printer.EmitHost, f
 		return true
 	}
 
-	// !!! Should JSON input files be emitted
-	return false
```
sheetalkamat, line 399:
```
// Json file is not emitted if outDir is not specified
    if (!options.outDir) return false;
```

this one is missing ?

sheetalkamat (changes requested):

internal/compiler/emitter.go
jakebailey: Good catch, I missed that one. Probably misread it as `outFile`

PR #1248, opened 2025-06-20
Port more resolveExternalModule, fix copy paste error in GetResolutionDiagnostic

jakebailey:
This fixes this error from the VS Code repo:

```
node_modules/@vscode/vscode-languagedetection/dist/lib/index.d.ts:1:8 - error TS2307: Cannot find module '@tensorflow/tfjs-backend-cpu' or its corresponding type declarations.

1 import '@tensorflow/tfjs-backend-cpu';
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 1 error in node_modules/@vscode/vscode-languagedetection/dist/lib/index.d.ts:1
```

Among other problems.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -14500,7 +14500,6 @@ func (c *Checker) resolveExternalModule(location *ast.Node, moduleReference stri
 			c.error(errorNode, resolutionDiagnostic, moduleReference, resolvedModule.ResolvedFileName)
 		}
 
-		// !!!
 		if errorNode != nil {
 			if resolvedModule.ResolvedUsingTsExtension && tspath.IsDeclarationFileName(moduleReference) {
 				if ast.FindAncestor(location, ast.IsEmittableImport) != nil {


internal/checker/checker.go
```diff
@@ -14482,13 +14482,24 @@ func (c *Checker) resolveExternalModule(location *ast.Node, moduleReference stri
...
 		sourceFile = c.program.GetSourceFileForResolvedModule(resolvedModule.ResolvedFileName)
 	}
 
 	if sourceFile != nil {
+		// If there's a resolutionDiagnostic we need to report it even if a sourceFile is found.
+		if resolutionDiagnostic != nil {
+			c.error(errorNode, resolutionDiagnostic, moduleReference, resolvedModule.ResolvedFileName)
+		}
+
 		// !!!
```
DanielRosenwasser, line 14503: I think you filled in the `!!!`

internal/checker/checker.go
```diff
@@ -14663,29 +14673,36 @@ func (c *Checker) resolveExternalModule(location *ast.Node, moduleReference stri
...
-			c.diagnostics.Add(NewDiagnosticChainForNode(errorInfo, errorNode, moduleNotFoundError, moduleReference))
-		} else {
-			c.error(errorNode, moduleNotFoundError, moduleReference)
 		}
 	}
 
 	return nil
 }
 
+func resolutionExtensionIsTSOrJson(ext string) bool {
```
DanielRosenwasser, line 14702: idk if we're consistent buy maybe just here

```suggestion
func resolutionExtensionIsTsOrJson(ext string) bool {
```

internal/checker/checker.go
jakebailey: Yeah, I think this function is actually now complete?

internal/checker/checker.go
```diff
@@ -14663,29 +14673,36 @@ func (c *Checker) resolveExternalModule(location *ast.Node, moduleReference stri
...
-			c.error(errorNode, diagnostics.Cannot_find_module_0_Consider_using_resolveJsonModule_to_import_module_with_json_extension, moduleReference)
-		} else if mode == core.ResolutionModeESM && resolutionIsNode16OrNext && isExtensionlessRelativePathImport {
-			absoluteRef := tspath.GetNormalizedAbsolutePath(moduleReference, tspath.GetDirectoryPath(importingSourceFile.FileName()))
-			if suggestedExt := c.getSuggestedImportExtension(absoluteRef); suggestedExt != "" {
-				c.error(errorNode, diagnostics.Relative_import_paths_need_explicit_file_extensions_in_ECMAScript_imports_when_moduleResolution_is_node16_or_nodenext_Did_you_mean_0, moduleReference+suggestedExt)
+		if resolutionDiagnostic != nil {
+			c.error(errorNode, resolutionDiagnostic, moduleReference, resolvedModule.ResolvedFileName)
+		} else {
+			isExtensionlessRelativePathImport := tspath.PathIsRelative(moduleReference) && !tspath.HasExtension(moduleReference)
+			resolutionIsNode16OrNext := c.moduleResolutionKind == core.ModuleResolutionKindNode16 || c.moduleResolutionKind == core.ModuleResolutionKindNodeNext
```
DanielRosenwasser, line 14680: This is in the current snapshot, but does this not kick in for `node20` in Strada?

sheetalkamat (approved):

DanielRosenwasser (approved):
I'll admit I don't know how/why this fixes the error in VS Code (maybe you can comment on why), but the implementation looks like it matches the existing Strada behavior.

jakebailey:
> I'll admit I don't know how/why this fixes the error in VS Code (maybe you can comment on why), but the implementation looks like it matches the existing Strada behavior.

The VS Code problem is specifically that a side effect import was being resolved to a JS file (lacking type definitions), but we weren't handling that case.

internal/checker/checker.go
jakebailey: It was a pure copy paste from the original code; we're not consistent at all about anything ☹️

DanielRosenwasser: At least in that we're consistent!

internal/checker/checker.go
jakebailey: See the description on https://github.com/microsoft/TypeScript/pull/61805; it would if Corsa had node20.

PR #1256, opened 2025-06-21
feat: handle unmatched jsdoc parameters

a-tarasyuk:


internal/checker/checker.go
```diff
@@ -30482,6 +30483,36 @@ func (c *Checker) getRegularTypeOfExpression(expr *ast.Node) *Type {
...
+				return visit(node.Expression())
+			case ast.KindPropertyAssignment:
+				return visit(node.AsPropertyAssignment().Initializer)
+			}
+			if nodeStartsNewLexicalEnvironment(node) || ast.IsPartOfTypeNode(node) {
+				return false
+			}
+			return node.ForEachChild(visit)
+		}
+		links.containsArgumentsReference = core.IfElse(visit(node.Body()), core.TSTrue, core.TSFalse)
```
jakebailey: You can just use `boolToTristate` here.

internal/checker/checker.go
```diff
@@ -30482,6 +30483,36 @@ func (c *Checker) getRegularTypeOfExpression(expr *ast.Node) *Type {
 	return c.getRegularTypeOfLiteralType(c.getTypeOfExpression(expr))
 }
 
+func (c *Checker) containsArgumentsReference(node *ast.Node) bool {
+	links := c.nodeLinks.Get(node)
```
jakebailey: Is there a more specific `links` that would make more sense here? I don't think we need to store this bit on every node's links, just functions, right?

a-tarasyuk: @jakebailey, thanks for the feedback. could the restriction apply only to declarations with a body, as that’s the starting point for traversal?

internal/checker/checker.go
a-tarasyuk: I've updated to use `boolToTristate`, however, `IfElse` looks ok to me :)

internal/checker/checker.go
jakebailey: It doesn't matter whether or not this is conditional because the field is still declared on the links type so will be paid for all node links. I was mainly wondering if this should go on `c.signatureLinks`.

a-tarasyuk: @jakebailey Yes, that makes sense. It would be cheaper. I initially followed the way TS handles it with `nodeLinks`

internal/checker/types.go
```diff
@@ -362,6 +362,7 @@ type NodeLinks struct {
 	flags                                NodeCheckFlags // Set of flags specific to Node
 	declarationRequiresScopeChange       core.Tristate  // Set by `useOuterVariableScopeInParameter` in checker when downlevel emit would change the name resolution scope inside of a parameter.
 	hasReportedStatementInAmbientContext bool           // Cache boolean if we report statements in ambient context
+	containsArgumentsReference           core.Tristate  // Whether a function-like declaration contains an 'arguments' reference
```
jakebailey: This can now be removed.

internal/checker/types.go
```diff
@@ -401,9 +402,10 @@ type SourceFileLinks struct {
 // Signature specific links
 
 type SignatureLinks struct {
-	resolvedSignature  *Signature // Cached signature of signature node or call expression
-	effectsSignature   *Signature // Signature with possible control flow effects
-	decoratorSignature *Signature // Signature for decorator as if invoked by the runtime
+	resolvedSignature          *Signature    // Cached signature of signature node or call expression
+	effectsSignature           *Signature    // Signature with possible control flow effects
+	decoratorSignature         *Signature    // Signature for decorator as if invoked by the runtime
+	containsArgumentsReference core.Tristate // Whether the signature contains an 'arguments' reference
```
jakebailey: You know, honestly, this might be fine as a map on the checker, given it only stores true or false...

jakebailey (approved):

internal/checker/jsdoc.go
```diff
@@ -0,0 +1,86 @@
+package checker
```
sandersn, line 1: 👍🏼 on creating a new file

internal/checker/jsdoc.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if ast.IsBindingPattern(name) {
+			excludedParameters.Add(i)
+		}
+	}
+	if c.containsArgumentsReference(node) {
+		lastJSDocParamIndex := len(jsdocParameters) - 1
+		lastJSDocParam := jsdocParameters[lastJSDocParamIndex].AsJSDocParameterOrPropertyTag()
+		if isJs && lastJSDocParam != nil && ast.IsIdentifier(lastJSDocParam.Name()) && lastJSDocParam.TypeExpression != nil &&
+			lastJSDocParam.TypeExpression.Type() != nil && !parameters.Has(lastJSDocParam.Name().Text()) && !excludedParameters.Has(lastJSDocParamIndex) {
```
sandersn: I don't see the check for whether the last param's type is an array type

internal/checker/jsdoc.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+	}
+}
+
+func getAllJSDocTags(node *ast.Node) []*ast.Node {
+	if node == nil {
+		return nil
+	}
+	tags := []*ast.Node{}
+	for _, jsdoc := range node.JSDoc(nil) {
```
sandersn: only the last jsdoc should provide tags

testdata/baselines/reference/submoduleAccepted/conformance/paramTagOnFunctionUsingArguments.errors.txt.diff
```diff
@@ -1,23 +1,28 @@
 --- old.paramTagOnFunctionUsingArguments.errors.txt
 +++ new.paramTagOnFunctionUsingArguments.errors.txt
 @@= skipped -0, +0 lines =@@
--a.js(2,20): error TS8029: JSDoc '@param' tag has name 'first', but there is no parameter with that name. It would match 'arguments' if it had an array type.
+ a.js(2,20): error TS8029: JSDoc '@param' tag has name 'first', but there is no parameter with that name. It would match 'arguments' if it had an array type.
 -a.js(19,9): error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.
++a.js(13,23): error TS8029: JSDoc '@param' tag has name 'strings', but there is no parameter with that name. It would match 'arguments' if it had an array type.
```
sandersn: this error should go away after the exemption for array type is added.

sandersn (changes requested):

internal/checker/jsdoc.go
a-tarasyuk: @sandersn Oops, missed that. Thanks for pointing that out. I've added support for array types.

internal/checker/jsdoc.go
a-tarasyuk: @sandersn I've changed to use tags from last jsdoc

jakebailey (approved):

sandersn (approved):

jakebailey:
Seems like this may have a bug; I see this in `checker.ts`:

image

But, `ignoreQualification` is there and spelled correctly.

jakebailey:
Oh, no, I'm wrong, the doc comment is backwards! That's hilarious.

PR #1257, opened 2025-06-21
Bring back `SourceFile.EndOfFileToken`

Andarist:


internal/astnav/tokens_test.go
```diff
@@ -20,7 +20,6 @@ import (
 )
 
 var testFiles = []string{
-	// !!! EOFToken JSDoc parsing is missing
 	// filepath.Join(repo.TestDataPath, "fixtures/astnav/eofJSDoc.ts"),
```
Andarist: even though I brought back `EndOfFileToken`, I'm not sure what to do here - this commented out file doesnt exist and I can't find its equivalent in Strada either

internal/parser/reparser.go
```diff
@@ -44,6 +44,10 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
 			continue
 		}
 		for _, tag := range tags.Nodes {
+			if parent.Kind == ast.KindEndOfFile {
+				p.reparseUnhosted(tag, parent, j)
+				continue
+			}
```
Andarist: this seems to make sense to me that htis should be treated as unhosted JSDoc. But despite this... issues like this did not go away (yet): https://github.com/microsoft/typescript-go/pull/1187#discussion_r2146155726

I need to investigate further, cc @sandersn

testdata/baselines/reference/astnav/GetTouchingPropertyName.mapCode.ts.baseline.txt
Andarist, line 1: tbh, I don't know how to assess if this is a correct change or not

internal/parser/reparser.go
Andarist: Ah, the reparsed nodes are simply not flushed in this case. I'm working on a fix for this

internal/parser/parser.go
```diff
@@ -315,12 +316,20 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
...
 	eof := p.parseTokenNode()
+	p.withJSDoc(eof, endHasJSDoc)
 	if eof.Kind != ast.KindEndOfFile {
 		panic("Expected end of file token from scanner.")
 	}
-	node := p.factory.NewSourceFile(p.opts, p.sourceText, statements)
+	if len(p.reparseList) > 0 {
+		nodes = append(nodes, p.reparseList...)
+		p.reparseList = nil
+		end = p.nodePos()
```
Andarist, line 347: I'm not entirely sure if this dance with `end` is fully correct here. It seems to me that `end` for the statement list should get shifted if some reparsed nodes were pushed to it - otherwise some weird data mismatches would occur and the code would have to, likely, handle them in a special way. cc @sandersn

_packages/api/src/node.ts
```diff
@@ -20,6 +20,7 @@ declare module "@typescript/ast" {
 const popcount8 = [0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4, 1, 2, 2, 3, 2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5, 1, 2, 2, 3, 2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 1, 2, 2, 3, 2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7, 1, 2, 2, 3, 2, 3, 3, 4, 2, 3, 3, 4, 3, 4, 4, 5, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7, 2, 3, 3, 4, 3, 4, 4, 5, 3, 4, 4, 5, 4, 5, 5, 6, 3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7, 3, 4, 4, 5, 4, 5, 5, 6, 4, 5, 5, 6, 5, 6, 6, 7, 4, 5, 5, 6, 5, 6, 6, 7, 5, 6, 6, 7, 6, 7, 7, 8];
 
 const childProperties: Readonly<Partial<Record<SyntaxKind, readonly string[]>>> = {
+    [SyntaxKind.SourceFile]: ["statements", "endOfFileToken"],
```
Andarist, line 23: @andrewbranch could you advise what further changes should be made to the api (probably in the encoder) to support this?

at the moment, it fails to recognize that a source file could have children, see https://github.com/microsoft/typescript-go/actions/runs/15805213937/job/44549073881

internal/astnav/tokens.go
```diff
@@ -49,6 +49,9 @@ func getTokenAtPosition(
 	left := 0
 
 	testNode := func(node *ast.Node) int {
+		if node.Kind == ast.KindEndOfFile {
+			return 0
```
sandersn: how does this translate from the code in services/utilities.ts? I don't see a specific EOF case in that code so I assume it's simplified from it in some way.

internal/astnav/tokens_test.go
sandersn: my best guess is that this is copilot-generated code, or some other variant of a todo. I think it's fine to delete this line too since all the jsdoc tests are proof that it works. (and mapCode's baselines should change too, actually)

_packages/api/src/node.ts
andrewbranch: Can you try making `get childMask()` return `-1` instead of `0` when the data type isn’t the child mask?

internal/parser/parser.go
sandersn: At first I thought you were wrong but after I tried replacing the mutation with a a panic I see you are right. I think it's the best option but it still might cause some weirdness when getTokenAtPosition should return an EOF. However, the reparsed nodes are marked as such and should be skipped by that code.

internal/parser/parser.go
```diff
@@ -315,12 +316,20 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 		p.contextFlags |= ast.NodeFlagsAmbient
 	}
 	pos := p.nodePos()
-	statements := p.parseListIndex(PCSourceElements, (*Parser).parseToplevelStatement)
+	nodes := p.parseListIndex(PCSourceElements, (*Parser).parseToplevelStatement)
```
sandersn: please keep this `statements`, since a list of statements are still statements

internal/parser/reparser.go
```diff
@@ -44,6 +44,10 @@ func (p *Parser) reparseTags(parent *ast.Node, jsDoc []*ast.Node) {
 			continue
 		}
 		for _, tag := range tags.Nodes {
+			if parent.Kind == ast.KindEndOfFile {
```
sandersn: no need for this special-case code, since there are no cases in reparseHosted for EndOfFile

internal/astnav/tokens_test.go
jakebailey: I think this is just leftover from someone's testing, likely it could just be removed.

internal/parser/reparser.go
```diff
@@ -102,6 +106,9 @@ func (p *Parser) reparseUnhosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Nod
 	case ast.KindJSDocImportTag:
 		importTag := tag.AsJSDocImportTag()
 		importClause := importTag.ImportClause
+		if importClause == nil {
```
sandersn, line 113: I think ModuleSpecifier is actually the critical property to require non-nil, although an import tag that imports no identifiers is pointless, so keep checking importClause for nil too.

internal/parser/parser.go
jakebailey: `end` should be a real location, so I would very much assume that we don't need to do anything here. The end of file token should always come after the statement list, and IIRC any reparsed nodes would have to as well, and so nothing special needs to happen?

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefault.js.diff
```diff
@@ -58,15 +74,87 @@
...
++    import Fab from "./index3";
++    declare class Bar extends Fab {
++        x: Bar;
++    }
++    export default Bar;
++    
++==== out/index5.d.ts (6 errors) ====
++    declare const _default: number;
++    export default _default;
++    export type default = string | number;
```
jakebailey, line 130: Hm, this looks like a syntax error. Curious whether or not this is a quick fix. What should have happened here is that the type was renamed to `_default` too. But, that wouldn't be in the parser, but the declaration transform.

internal/parser/parser.go
Andarist: Yeah, all of that makes sense and matches my expectations. I believe one of the previous iterations was not correct because it computed `end` before the reparsed nodes on the eof were handled. That's why I have changed the `parseListIndex`'s return type - so I could get control over the assigned `end` position.

I've just done this because it intuitively made sense but I have not fully verified why the previous version of this code I had have not worked. And because of that, I just want the reviewers to be extra cautious here - even though I think this makes sense :)

_packages/api/src/node.ts
Andarist: 🪄 

Thank you! This worked

testdata/baselines/reference/submoduleAccepted/conformance/importTag14.errors.txt.diff
```diff
@@ -1,18 +1,20 @@
...
--                                    ~~~~
--!!! error TS1464: Type import attributes should have exactly one key - 'resolution-mode' - with value 'import' or 'require'.
--                                    ~~~~
--!!! error TS2823: Import attributes are only supported when the '--module' option is set to 'esnext', 'node18', 'nodenext', or 'preserve'.
+
+
+ ==== /foo.js (3 errors) ====
+     /** @import * as f from "./foo" with */
++                            ~~~~~~~
++!!! error TS2306: File '/foo.js' is not a module.
```
jakebailey, line 13: Weird, but we also have other follow-on errors anyway.

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefault.js.diff
Andarist: I'll look into it. Btw. this doesn't correctly generate a unique identifier in Strada (:

```js
// @allowJs: true
// @checkJs: true
// @target: es5
// @outDir: ./out
// @declaration: true

export const _default = class {}
export default 12;
/**
 * @typedef {string | number} default
 */
```

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsFunctionPrototypeStatic.errors.txt.diff
```diff
@@ -0,0 +1,23 @@
+--- old.jsDeclarationsFunctionPrototypeStatic.errors.txt
++++ new.jsDeclarationsFunctionPrototypeStatic.errors.txt
+@@= skipped -0, +0 lines =@@
+-<no content>
++source.js(1,1): error TS2309: An export assignment cannot be used in a module with other exported elements.
```
jakebailey, line 5: This seems preexisting; the reparsed code triggers some other errors about addtional exports.

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefault.js.diff
jakebailey: [Playground Link](https://www.typescriptlang.org/play/?filetype=js#code/KYDwDg9gTgLgBAE2AMwIYFcA28CMAmAbgCgB6AKjKLjLgAEYBPMYJZOAbwGcYoBLAOwDmcAD5x+6ALYAjYFAC+iFBmxUyJIkA) is what I was testing; putting your code into that also seems to work?

internal/parser/parser.go
jakebailey: I actually more confused as to why we need to append `reparseList` at all. I thought that was an internal detail of the reparser?

jakebailey: Hm, I guess that's where it sticks its statements, so.

jakebailey: In other words, I think what you've written here is actually all correct.

Or at least, at worst maybe this position should just be the old end before the EOF.

sandersn (commented):
Still reviewing baselines, but here are some out of order comments.

testdata/baselines/reference/astnav/GetTouchingPropertyName.mapCode.ts.baseline.txt
```diff
@@ -1,3 +1,39 @@
+〚Positions: [729, 730]〛
+【TS: SourceFile [0, 9772)】
+《Go: EndOfFileToken [9770, 9772)》
```
sandersn: @andrewbranch I don't know how to read these baselines, but this could be the difference in GetTokenAtPosition that I was worried about.

sandersn: Later: no, there's no jsdoc at the end; this is just a difference in GetTokenAtPosition or something similar between Strada and Corsa. Based on my reading of the diff--still not an expert--it looks like it is incorrect and Strada is correct.

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefault.js.diff
Andarist: In my sample, there is an explicit `_default` exported and the merged default export is assigned a `_default` name too. Both of those should refer to different identifiers/exports

testdata/baselines/reference/submoduleAccepted/compiler/uniqueSymbolJs.errors.txt.diff
```diff
@@ -3,17 +3,19 @@
...
+     /** @typedef {{ [foo]: boolean }} A */
+     /** @typedef {{ [key: foo] boolean }} B */
+                      ~~~
 -!!! error TS1337: An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead.
 -                               ~~~~~~~
 -!!! error TS1005: ';' expected.
--    
-+<no content>
\ No newline at end of file
++!!! error TS1268: An index signature parameter type must be 'string', 'number', 'symbol', or a template literal type.
```
sandersn, line 18: this error seems more accurate than the Strada one, although it's missing the suggestion to use a mapped type.

testdata/baselines/reference/submoduleAccepted/conformance/callbackTag2.types.diff
```diff
@@ -17,13 +17,7 @@
...
-->noctis : { heroes: any; } & { fantasy: any; }
-+>(barts, tidus, noctis) => "cecil" : (barts: any, tidus: any, noctis: any) => string
-+>barts : any
-+>tidus : any
-+>noctis : any
- >"cecil" : "cecil"
-
- /**
\ No newline at end of file
++>(barts, tidus, noctis) => "cecil" : (barts: { fantasy: any; }, tidus: { heroes: any; }, noctis: { heroes: any; } & { fantasy: any; }) => "cecil"
```
sandersn, line 20: this diff is because the type annotation is now a contextual type, not a type annotation that attaches to the arrow

testdata/baselines/reference/submoduleAccepted/conformance/importTag13.errors.txt.diff
```diff
@@ -2,16 +2,15 @@
...
--    
--==== /types.ts (0 errors) ====
--    export interface Foo {
--        a: number;
--    }
--    
-+<no content>
\ No newline at end of file
++                  ~~~~~~~~~~~~~~~~~~
++!!! error TS1141: String literal expected.
```
sandersn, line 13: I don't know why this parse error is different, but it's not related to this PR

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsFunctionPrototypeStatic.errors.txt.diff
sandersn: yep. I need to get back to the PR that fixes these now that declaration emit is done, and I'm done adding (most) JS features.

sandersn (dismissed):
Requesting changes based on my previous comments, since this batch is just guesses for why diffs have happened.

internal/astnav/tokens.go
Andarist: This is a good question. It's an indirect port of the EOF case checked by `nodeContainsPosition`. The related code in Corsa is rewritten sligthly so it's not easy to see this. There is a chance this second `nodeContainsPosition` call might have to be ported too... but, so far, I have not found a need for it (at least not in the case of EOF):
```ts
            // first element whose start position is before the input and whose end position is after or equal to the input
            if (nodeContainsPosition(children[middle], start, end)) {
                if (children[middle - 1]) {
                    // we want the _first_ element that contains the position, so left-recur if the prior node also contains the position
                    if (nodeContainsPosition(children[middle - 1])) {
                        return Comparison.GreaterThan;
                    }
                }
                return Comparison.EqualTo;
            }
```

internal/parser/reparser.go
Andarist: I'm not sure if I understand correctly. Have you meant I should adjust it to this?
```go
if importClause == nil || importTag.ModuleSpecifier == nil  {
  break
}
```

testdata/baselines/reference/submoduleAccepted/compiler/uniqueSymbolJs.errors.txt.diff
Andarist: This depends on `t := c.getTypeFromTypeNode(typeNode)`. Corsa gets an unresolved any type (an error type) whereas Strada gets `unique symbol` type.

In what way, you'd like to unify this stuff?

testdata/baselines/reference/submoduleAccepted/conformance/callbackTag2.types.diff
Andarist: > this diff is because the type annotation is now a contextual type, not a type annotation that attaches to the arrow

Is this a desired change? 🤔

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefault.js.diff
Andarist: This pach would work, please let me know if you think this is OK and if I should include it in this PR:
```diff
diff --git a/internal/transformers/declarations/transform.go b/internal/transformers/declarations/transform.go
index 9b07cd78a..6a4c5152d 100644
--- a/internal/transformers/declarations/transform.go
+++ b/internal/transformers/declarations/transform.go
@@ -52,6 +52,7 @@ type DeclarationTransformer struct {
 	needsScopeFixMarker              bool
 	resultHasScopeMarker             bool
 	enclosingDeclaration             *ast.Node
+	generatedDefaultExportIdentifier *ast.Node
 	resultHasExternalModuleIndicator bool
 	suppressNewDiagnosticContexts    bool
 	lateStatementReplacementMap      map[ast.NodeId]*ast.Node
@@ -912,7 +913,7 @@ func (tx *DeclarationTransformer) visitDeclarationStatements(input *ast.Node) *a
 			return input
 		}
 		// expression is non-identifier, create _default typed variable to reference
-		newId := tx.Factory().NewUniqueNameEx("_default", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic})
+		newId := tx.getGeneratedDefaultExportIdentifier()
 		tx.state.getSymbolAccessibilityDiagnostic = func(_ printer.SymbolAccessibilityResult) *SymbolAccessibilityDiagnostic {
 			return &SymbolAccessibilityDiagnostic{
 				diagnosticMessage: diagnostics.Default_export_of_the_module_has_or_is_using_private_name_0,
@@ -1114,10 +1115,14 @@ func (tx *DeclarationTransformer) transformTopLevelDeclaration(input *ast.Node)
 
 func (tx *DeclarationTransformer) transformTypeAliasDeclaration(input *ast.TypeAliasDeclaration) *ast.Node {
 	tx.needsDeclare = false
+	name := input.Name()
+	if ast.IsSourceFile(input.Parent) && ast.IsIdentifier(name) && name.Text() == "default" {
+		name = tx.getGeneratedDefaultExportIdentifier()
+	}
 	return tx.Factory().UpdateTypeAliasDeclaration(
 		input,
 		tx.ensureModifiers(input.AsNode()),
-		input.Name(),
+		name,
 		tx.Visitor().VisitNodes(input.TypeParameters),
 		tx.Visitor().Visit(input.Type),
 	)
@@ -1767,3 +1772,12 @@ func (tx *DeclarationTransformer) transformJSDocOptionalType(input *ast.JSDocOpt
 	tx.EmitContext().SetOriginal(replacement, input.AsNode())
 	return replacement
 }
+
+func (tx *DeclarationTransformer) getGeneratedDefaultExportIdentifier() *ast.Node {
+	name := tx.generatedDefaultExportIdentifier
+	if name == nil {
+		name = tx.Factory().NewUniqueNameEx("_default", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic})
+		tx.generatedDefaultExportIdentifier = name
+	}
+	return name
+}

```

jakebailey: Feels suspect but maybe; why doesn't Strada need this?

Andarist: In the case above, Strada generated incorrect dts ([TS playground](https://www.typescriptlang.org/play/?target=1&noEmit=true&outDir=.%2Fout&ts=5.9.0-dev.20250623&filetype=js#code/PTAEAEEMBtoewO4CkDOAuUAXATgVwKYBQIEAxgBb6kDWqGOBxY4mk2A5vphvigKxMIcXJgAiAS2wYAdMGGZB4ACZVobSJnFwAdvTxFC+AB4AHONkyhSOlJYD6KgGaRc0SwF4ralClABvAF9DU3NLJxc3UABGACYAbmIAKkTCUESITABPE3wnf1tscW12UAAfUG1cAFsAI3xsANBw1wU04CA)):
```ts
export const _default: {
    new (): {};
};
type _default = string | number;
declare const _default: 12;
export default _default;
```

My goal was to fix this in Corsa while implementing the change

jakebailey: Ah, might as well send it there then.

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefaultsErr.js
```diff
@@ -60,14 +60,14 @@ declare class Cls {
...
 export default class C {
 }
-export type default = string | number;
+export type _default = string | number;
 //// [index3.d.ts]
 // merge type alias and variable (behavior is borked, see #32366)
 declare const x = 12;
 export { x as default };
-export type default = string | number;
+export type _default = string | number;
```
jakebailey: Hm, but this is weird, no?

internal/parser/reparser.go
sandersn: yes, that

testdata/baselines/reference/submoduleAccepted/conformance/callbackTag2.types.diff
sandersn: It's intentional, at least. I like it and @jakebailey doesn't =)

testdata/baselines/reference/submoduleAccepted/compiler/uniqueSymbolJs.errors.txt.diff
sandersn: ah, well, in that case the Strada type is more correct. I dn't think there's anything to fix for this PR, just a later fix--it sounds as if this might even be a difference between Corsa and Strada in TS as well.

testdata/baselines/reference/submoduleAccepted/conformance/callbackTag2.types.diff
jakebailey: 😡

internal/astnav/tokens.go
sandersn: I don't know either but I suspect that it's the cause of the difference in astnav baselines

testdata/baselines/reference/submodule/conformance/jsDeclarationsDefaultsErr.js
Andarist: it is, the comment above this says the behavior is borked - and the type alias declaration here is the same in Strada:
https://github.com/microsoft/TypeScript/blob/78c16795cdee70b9d9f0f248b6dbb6ba50994a59/tests/baselines/reference/jsDeclarationsDefaultsErr.js#L81-L84

internal/parser/reparser.go
Andarist: All of those work alright-ish right now:
```ts
// @checkJs: true
// @allowJs: true

// @filename: /foo.js
/**
 * @import {A} from 
 */

/**
 * @import from 
 */

/**
 * @import
 */
```

So I wonder, if the proposed change is actually needed. The module specifier end up being an "empty identifier" (with an error flag set on it) in all three of those cases

sandersn: Yes, if it works without the new code, please delete it.

jakebailey:
Seems like it needs more baseline updates.

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsUniqueSymbolUsage.types.diff
```diff
@@ -1,15 +1,13 @@
...
+ >b : (value: import("./a").WithSymbol) => import("./a").WithSymbol
 ->value : import("a").WithSymbol
-+>b : (value: any) => any
-+>value : any
++>value : import("./a").WithSymbol
 
      return value;
 ->value : import("a").WithSymbol
-+>value : any
++>value : import("./a").WithSymbol
```
Andarist, line 12: interestingly, this seems buggy in Strada - but this is already in the `submoduleAccepted` here so 🤷 

https://github.com/microsoft/TypeScript/blob/e6a50a78619632bba7af4f23b2e7e4940b9cf788/tests/baselines/reference/jsDeclarationsUniqueSymbolUsage.types#L18-L31

jakebailey:
Still a test failure?

```
+--- old.partiallyNamedTuples2.types
++++ new.partiallyNamedTuples2.types
+@@= skipped -37, +37 lines =@@
+ const matches = x.get(id1);
+ >matches : Iterable<[id2: string, object]>
+ >x.get(id1) : Iterable<[id2: string, object]>
+->x.get : <Key extends [id1: string, id2: string] | [id1: string] | []>(...key: Key) => GetResult<[id1: string, id2: string], Key, object>
++>x.get : <Key extends [] | [id1: string] | [id1: string, id2: string]>(...key: Key) => GetResult<[id1: string, id2: string], Key, object>
+ >x : MultiKeyMap<[id1: string, id2: string], object>
+->get : <Key extends [id1: string, id2: string] | [id1: string] | []>(...key: Key) => GetResult<[id1: string, id2: string], Key, object>
++>get : <Key extends [] | [id1: string] | [id1: string, id2: string]>(...key: Key) => GetResult<[id1: string, id2: string], Key, object>
+ >id1 : string
```

Seems to have been a bad merge with main?

Andarist:
weird, maybe I have forgotten to git add this one single baseline after syncing with main. But either way, this is still blocked on the `GetTouchingPropertyName` difference - I hope to fix this before Friday but we'll see how it goes

internal/transformers/declarations/transform.go
```diff
@@ -1767,3 +1772,12 @@ func (tx *DeclarationTransformer) transformJSDocOptionalType(input *ast.JSDocOpt
 	tx.EmitContext().SetOriginal(replacement, input.AsNode())
 	return replacement
 }
+
+func (tx *DeclarationTransformer) getGeneratedDefaultExportIdentifier() *ast.Node {
```
jakebailey: This function feels wrong to me; this means we're putting the same node into the tree multiple times, which is a no-no.

jakebailey: (I'm not 100% sure this fix is required for the rest of the PR, so would probably be worth a follow-up as a d.ts improvement?)

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsUniqueSymbolUsage.types.diff
jakebailey: This is a known difference across a lot of these baselines, yeah.

Note that "accepted" applies to basically all JS code right now given we're claiming it's "not complete", so it was helpful to ignore these diffs in other PRs just fixing type checking issues.

testdata/baselines/reference/astnav/GetTouchingPropertyName.mapCode.ts.baseline.txt
andrewbranch: I think the diff generator has a subtle bug where it’s not showing the actual positions where this is occurring, but all of these (729, 730, 780–783, etc.) are whitespace between top-level statements where Strada returns the SourceFile (seems right enough) and Corsa is returning the EOF token (seems wrong, since its text range is nowhere near the position being triggered).

internal/transformers/declarations/transform.go
Andarist: Ah, ye - I meant to ask if I should clone this before returning and forgot. 

> (I'm not 100% sure this fix is required for the rest of the PR, so would probably be worth a follow-up as a d.ts improvement?)

yes, it's a little bit out of scope of this PR. This has just popped up in the diffs after EOF got included. I'll remove it from this PR

jakebailey (commented):
The core of this PR is fine, but I think we're still waiting on the removal of the extra unrelated bits?

Andarist:
@jakebailey I think I addressed everything that had to be addressed here and it should be good to go now

jakebailey:
What about:

- https://github.com/microsoft/typescript-go/pull/1257#discussion_r2163949008
- https://github.com/microsoft/typescript-go/pull/1257#discussion_r2167141443

Andarist:
1. this is outdated - I have not included the asked change because all tests pass without it. The change I've made is required by the tests in the codebase
2. this has slipped my mind. I removed this now

jakebailey:
Thanks!

internal/transformers/declarations/transform.go
```diff
@@ -53,6 +53,7 @@ type DeclarationTransformer struct {
 	needsScopeFixMarker              bool
 	resultHasScopeMarker             bool
 	enclosingDeclaration             *ast.Node
+	generatedDefaultExportIdentifier *ast.Node
```
jakebailey:
```suggestion
```

jakebailey (approved):

PR #1260, opened 2025-06-21
Port TS PR 60566 (@import * as namespace cannot be used with @implements)

jakebailey:
Ports https://github.com/microsoft/TypeScript/pull/60566/files


internal/ast/utilities.go
```diff
@@ -3033,7 +3033,7 @@ func IsJSDocSingleCommentNode(node *Node) bool {
 }
 
 func IsValidTypeOnlyAliasUseSite(useSite *Node) bool {
-	return useSite.Flags&NodeFlagsAmbient != 0 ||
+	return useSite.Flags&(NodeFlagsAmbient|NodeFlagsJSDoc) != 0 ||
```
DanielRosenwasser, line 3036: Kind of funny that you didn't do this in the other PR.

DanielRosenwasser (approved):

internal/ast/utilities.go
jakebailey: Other PR?

PR #1262, opened 2025-06-22
Port lib replacement and --libReplacement

jakebailey:
Ports lib replacement itself along with https://github.com/microsoft/TypeScript/pull/60829.

The default in this PR is libReplacement=false, unlike Strada, per the deprecation plan. But, easy to flip back if we think it's a problem.


internal/compiler/fileloader.go
```diff
@@ -463,6 +480,52 @@ func (p *fileLoader) createSyntheticImport(text string, file *ast.SourceFile) *a
...
 
+func (p *fileLoader) pathForLibFile(name string) string {
+	if cached, ok := p.pathForLibFileCache.Load(name); ok {
+		return cached
+	}
+
+	path := tspath.CombinePaths(p.defaultLibraryPath, name)
+	if p.opts.Config.CompilerOptions().LibReplacement.IsTrue() {
+		libraryName := getLibraryNameFromLibFileName(name)
+		resolveFrom := getInferredLibraryNameResolveFrom(p.opts.Config.CompilerOptions(), p.opts.Host.GetCurrentDirectory(), name)
```
sheetalkamat, line 497: i think you want to store these resolutions so we can watch these locations ?

jakebailey: Ah, yes!

jakebailey: I did something, though I'm not sure if it's too hacky. Probably not?

internal/compiler/fileloader.go
```diff
@@ -162,6 +180,12 @@ func processAllProgramFiles(
 
 	allFiles := append(libFiles, files...)
 
+	if loader.pathForLibFileResolutions.Size() > 0 {
+		// Hack: place lib file resolutions into the resolvedModules map as if they were
+		// resolved by the config file itself, letting the Project code watch these files.
+		resolvedModules[loader.toPath(compilerOptions.ConfigFilePath)] = loader.pathForLibFileResolutions.ToMap()
```
sheetalkamat: May be use this instead : `resolveFrom := getInferredLibraryNameResolveFrom(p.opts.Config.CompilerOptions(), p.opts.Host.GetCurrentDirectory(), name) ` since configFilePath might be ""

jakebailey: It also works if it's empty, as far as I can tell; the watcher code never looks at the actual paths.

But, you're suggesting I use the fake file path instead?

sheetalkamat: yeah thats what we do to resolve auto type ref and add resolutions in typeResolutions so makes similar

jakebailey: I'll try it, though it makes the atomics a little annoying 😄

jakebailey: Okay, done.

sheetalkamat (approved):

PR #1264, opened 2025-06-22
Default expected `sortText` to `LocationPriority` when verifying completions

Andarist:


internal/fourslash/fourslash.go
```diff
@@ -466,14 +466,20 @@ func verifyCompletionsAreExactly(t *testing.T, prefix string, actual []*lsproto.
...
+				return false
+			}
 		},
 		cmp.Ignore(),
 	)
 	assertDeepEqual(t, actual, expected, prefix, ignoreKind)
 	if expected.Kind != nil {
 		assertDeepEqual(t, actual.Kind, expected.Kind, prefix+" Kind mismatch")
 	}
+	assertDeepEqual(t, actual.SortText, core.OrElse(expected.SortText, ptrTo(string(ls.SortTextLocationPriority))), prefix+" SortText mismatch")
```
Andarist, line 482: from Strada: https://github.com/microsoft/TypeScript/blob/479285d0ac293c35a926508d17f0bb5eca7e0303/src/harness/fourslashImpl.ts#L1223

gabritto:
I wonder if instead of doing this, it wouldn't be better to just have the conversion script add the default sort text explicitly when it's missing in the original Strada tests.

Andarist:
@gabritto I considered this but figured out it will be annoying to include this explicitly in all tests once u start adding new tests manually to Corsa. If you prefer that though - let me know, I can certainly switch this PR to do that instead

gabritto:
Yeah, I think defaulting to something is more convenient for writing the tests, but might be more confusing for reading them.

I've been trying to make the fourslash behavior more consistent in terms of which properties we ignore, which we default etc, but I think that's impossible, sadly.
So we might as well make it more convenient to write these tests in the common case.

gabritto (approved):

PR #1266, opened 2025-06-23
Handle `blockedStringType`

Andarist:


jakebailey:
Seems like there might be some nondeterministic behavior based on the test failures?

gabritto:
> Seems like there might be some nondeterministic behavior based on the test failures?

Yep, which is really surprising to me... @Andarist if you want, I can investigate this when I'm done with my current batch of completions/fourslash fixes.

Andarist:
@jakebailey pointed me to the cause of non-determinism (thanks!) so it's fixed now

internal/checker/services.go
```diff
@@ -585,28 +585,27 @@ func (c *Checker) getResolvedSignatureWorker(node *ast.Node, checkMode CheckMode
 }
 
 func (c *Checker) GetCandidateSignaturesForStringLiteralCompletions(call *ast.CallLikeExpression, editingArgument *ast.Node) []*Signature {
-	candidatesSet := collections.Set[*Signature]{}
```
gabritto, line 588: I think I thought I was using `collections.OrderedSet` here, oops.

gabritto (approved):

internal/checker/services.go
```diff
@@ -585,28 +585,27 @@ func (c *Checker) getResolvedSignatureWorker(node *ast.Node, checkMode CheckMode
...
-
 	// first, get candidates when inference is blocked from the source node.
 	candidates := runWithInferenceBlockedFromSourceNode(c, editingArgument, func() []*Signature {
 		_, blockedInferenceCandidates := c.getResolvedSignatureWorker(call, CheckModeNormal, 0)
 		return blockedInferenceCandidates
 	})
-	for _, candidate := range candidates {
-		candidatesSet.Add(candidate)
-	}
+	candidatesSet := collections.NewSetFromItems(candidates...)
```
jakebailey, line 593: Could just use an `OrderedSet`, right?

internal/checker/services.go
```diff
@@ -585,28 +585,27 @@ func (c *Checker) getResolvedSignatureWorker(node *ast.Node, checkMode CheckMode
...
 
 	// next, get candidates where the source node is considered for inference.
-	candidates = runWithoutResolvedSignatureCaching(c, editingArgument, func() []*Signature {
+	otherCandidates := runWithoutResolvedSignatureCaching(c, editingArgument, func() []*Signature {
 		_, inferenceCandidates := c.getResolvedSignatureWorker(call, CheckModeNormal, 0)
 		return inferenceCandidates
 	})
 
-	for _, candidate := range candidates {
-		candidatesSet.Add(candidate)
```
jakebailey, line 606: I'm confused, isn't the new behavior different from the old behavior? The old code added both candidates to the same set, but the new one filters the second candidate list based on the first.

jakebailey: Oh, you're tacking stuff onto the first candidate list. I see.

jakebailey (approved):

PR #1272, opened 2025-06-23
Add a basic `copilot-instructions.md`.

DanielRosenwasser:


RyanCavanaugh (approved):

.github/copilot-instructions.md
```diff
@@ -0,0 +1,84 @@
+This is the codebase for a native port of the TypeScript compiler and language server.
+The source directories of interest that we have are:
+
+- `internal` - Contains the compiler and language server code.
+- `_extension` - Contains a preview VS Code extension code that integrates with the language server.
```
andrewbranch, line 5: Note the existence of `_submodules/TypeScript`; I saw a porting PR fail to notice that

.github/copilot-instructions.md
```diff
@@ -0,0 +1,84 @@
...
+hereby format # Format the code
+hereby lint   # Run linters
+```
+
+Always make sure code is formatted, linted, and tested before sending a pull request.
+
+## Compiler Features, Fixes, and Tests
+
+
+When fixing a bug or implementing a new feature, at least one minimal test case should always be added in advance to verify the fix.
```
andrewbranch, line 25: I don’t think this is true at this stage of development. Many things are adequately tested but have existing diffs that should be deleted. Even for fixes outside porting PRs, I would rather not add new tests that are wholly duplicative.

DanielRosenwasser: How would I phrase this exactly?

> Don't add a test case, unless nothing changes, in which case, undo all your work, add a failing test case, and re-apply the changes you've made and verify that they fix the issue.

I kind of doubt that a bot is going to do the right thing, so I'd rather just say "delete the new test and accept baselines". But honestly, I'm flexible here.

andrewbranch: I guess it’s not bad if the effect is a new correct test that we can take or leave, but I think the diff system does need to be explained, as deleting diffs is a more reliable sign that the implementation is right than the AI’s judgment of new baselines that go with an AI-written test case.

PR #1274, opened 2025-06-23
fourslash baselines and `VerifyBaselinefindAllReferences`

iisaduan:
Ports most of the baselining for fourslash (the functions that format FS content with range and location markers). There are features I did not port because they were not used in the baselines that I was porting/the functions I know that are planned on be ported pretty directly. I'm leaving those features unimplemented for now in case they are used differently/no longer used 

Also adds `VerifyBaselineFindAllReferences` as an example of how to use the new baseline. These tests aren't being run against the submodule yet, they are just a standard "has the baseline changed" test. Not sure if we will eventually make these run against the submodule, since in corsa we currently only compute the locations of references so most of the stuff that is in the strada baseline (the json structures) is no longer computed. There are also some known bugs in what's currently implemented in FAR (see https://github.com/microsoft/typescript-go/issues/1342)


internal/ls/linemap.go
```diff
@@ -48,3 +50,20 @@ func ComputeLineStarts(text string) *LineMap {
...
+func (lm *LineMap) ComputeIndexOfLineStart(targetPos core.TextPos) int {
+	// port of computeLineOfPosition(lineStarts: readonly number[], position: number, lowerBound?: number): number {
+	lineNumber, ok := slices.BinarySearchFunc(lm.LineStarts, targetPos, func(p, t core.TextPos) int {
+		return cmp.Compare(int(p), int(t))
+	})
+	if !ok && lineNumber > 0 {
+		// If the actual position was not found, the binary search returns where the target line start would be inserted
+		// if the target was in the slice.
+		// e.g. if the line starts at [5, 10, 23, 80] and the position requested was 20
+		// then the search will return -3.
```
gabritto: This comment probably needs updating, since the Corsa version will return `(3, false)` instead of `-3`.

internal/fourslash/fourslash.go
```diff
@@ -317,6 +319,14 @@ func (f *FourslashTest) goToPosition(t *testing.T, position lsproto.Position) {
 	f.selectionEnd = nil
 }
 
+func (f *FourslashTest) GoToPosAndFile(t *testing.T, pos lsproto.Position, fileName string) {
+	// GoToRangeStart
+	f.ensureActiveFile(t, fileName)
+	f.currentCaretPosition = pos
+	f.activeFilename = fileName
+	// !!! this.selectionEnd = -1
```
gabritto: I recently added `selectionEnd` in https://github.com/microsoft/typescript-go/pull/1310/ and `GoToRangeStart` in https://github.com/microsoft/typescript-go/pull/1357/, so this probably needs updating.

internal/fourslash/fourslash.go
```diff
@@ -435,13 +454,8 @@ func (f *FourslashTest) verifyCompletionsAtMarker(t *testing.T, markerName strin
...
 func (f *FourslashTest) verifyCompletionsWorker(t *testing.T, expected *CompletionsExpectedList) {
 	params := &lsproto.CompletionParams{
-		TextDocumentPositionParams: lsproto.TextDocumentPositionParams{
-			TextDocument: lsproto.TextDocumentIdentifier{
-				Uri: ls.FileNameToDocumentURI(f.activeFilename),
-			},
-			Position: f.currentCaretPosition,
-		},
-		Context: &lsproto.CompletionContext{},
+		TextDocumentPositionParams: f.currentTextDocumentPositionParams(),
```
gabritto, line 541: :)

internal/fourslash/fourslash.go
```diff
@@ -652,6 +666,67 @@ func assertDeepEqual(t *testing.T, actual any, expected any, prefix string, opts
...
+) {
+	// if there are no markers specified, use all ranges
+	var referenceLocations []MarkerOrRange
+	if len(markers) == 0 {
+		referenceLocations = core.Map(f.testData.Ranges, func(r *RangeMarker) MarkerOrRange { return r })
+	} else {
+		referenceLocations = core.Map(markers, func(markerName string) MarkerOrRange {
+			marker, ok := f.testData.MarkerPositions[markerName]
+			if !ok {
+				t.Fatalf("Marker %s not found", markerName)
```
gabritto: nit:
```suggestion
				t.Fatalf("Marker '%s' not found", markerName)
```

internal/fourslash/test_parser.go
```diff
@@ -75,8 +105,12 @@ func ParseTestData(t *testing.T, contents string, fileName string) TestData {
 		markers = append(markers, file.markers...)
 		ranges = append(ranges, file.ranges...)
 		for _, marker := range file.markers {
-			if _, ok := markerPositions[marker.Name]; ok {
-				t.Fatalf("Duplicate marker name: %s", marker.Name)
+			if marker.Name == "" && marker.Data != nil {
```
gabritto: I fixed this in https://github.com/microsoft/typescript-go/pull/1357/ by having a marker name be `*string` instead of `string`, since in Strada, so probably no need for this change now except for the improved error message.

internal/fourslash/baselineutil.go
```diff
@@ -0,0 +1,459 @@
...
+	additionalSpan                *lsproto.Location
+}
+
+func (f *FourslashTest) getBaselineForLocationsWithFileContents(spans []*lsproto.Location, options baselineFourslashLocationsOptions) string {
+	return f.getBaselineForGroupedLocationsWithFileContents(collections.GroupBy(spans, func(span *lsproto.Location) lsproto.DocumentUri { return span.Uri }), options)
+}
+
+func (f *FourslashTest) getBaselineForGroupedLocationsWithFileContents(groupedLocations *collections.MultiMap[lsproto.DocumentUri, *lsproto.Location], options baselineFourslashLocationsOptions) string {
+	baselineEntries := []string{}
+	err := f.server.FS().WalkDir("/", func(path string, d vfs.DirEntry, e error) error {
```
gabritto:
```suggestion
	err := f.vfs.WalkDir("/", func(path string, d vfs.DirEntry, e error) error {
```

internal/fourslash/baselineutil.go
```diff
@@ -0,0 +1,459 @@
...
+	return t.content
+}
+
+func newTextWithContext(fileName string, content string) *textWithContext {
+	t := &textWithContext{
+		nLinesContext: 4,
+
+		readableContents: &strings.Builder{},
+
+		isLibFile:  regexp.MustCompile(`/lib.*\.d\.ts$/`).MatchString(fileName),
```
gabritto: Should this be ``regexp.MustCompile(`lib.*\.d\.ts$`)``?

testdata/baselines/reference/fourslash/findAllRef/AmbientShorthandFindAllRefs.baseline.jsonc
```diff
@@ -0,0 +1,12 @@
+// === findAllReferences ===
```
gabritto, line 1: This looks different from the original baseline file in Strada, right? Where in the code are we accounting for those differences such that tests don't fail?

gabritto (changes requested):
Some minor comments (mostly about recent updates I made that may conflict, sorry) and a question, but in general looks good.

testdata/baselines/reference/fourslash/findAllRef/AmbientShorthandFindAllRefs.baseline.jsonc
iisaduan: The findAllReferences tests aren't being run against the submodule yet, they are just a standard "has the baseline changed" test. For FAR specifically, in corsa we currently only compute the locations of references--so most of the stuff that is in corsa's baseline (the json structures) is just info that is no longer computed.

gabritto (approved):

PR #1276, opened 2025-06-24
Fix race in isFromExternalLibrary

jakebailey:
I don't know why I thought mutating the task would be okay. Multiple tasks can be walking the same task via the subtasks, so we'll race when marking in the critical section, but reading at the top.

Just mark them as external in a second pass during collection.

### DIFF
--- a/internal/compiler/fileloadertask.go
+++ b/internal/compiler/fileloadertask.go
@@ -29,9 +29,10 @@ type fileLoaderWorker[K fileLoaderWorkerTask[K]] struct {
 }
 
 type queuedTask[K fileLoaderWorkerTask[K]] struct {
-	task        K
-	mu          sync.Mutex
-	lowestDepth int
+	task                K
+	mu                  sync.Mutex
+	lowestDepth         int
+	fromExternalLibrary bool
 }
 
 func (w *fileLoaderWorker[K]) runAndWait(loader *fileLoader, tasks []K) {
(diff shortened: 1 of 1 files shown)


internal/compiler/fileloadertask.go
```diff
@@ -83,13 +85,20 @@ func (w *fileLoaderWorker[K]) start(loader *fileLoader, tasks []K, depth int, is
...
-				w.start(loader, task.getSubTasks(), loadedTask.lowestDepth, task.isFromExternalLibrary())
+				w.start(loader, task.getSubTasks(), loadedTask.lowestDepth, loadedTask.fromExternalLibrary)
 			}
 		})
 	}
 }
 
 func (w *fileLoaderWorker[K]) collect(loader *fileLoader, tasks []K, iterate func(K, []tspath.Path)) []tspath.Path {
+	// Mark all tasks we saw as external after the fact.
+	w.tasksByFileName.Range(func(key string, value *queuedTask[K]) bool {
```
sheetalkamat, line 96: Not quite what we do in strada right ? 

if the task is root it isnt marked as from external library - so dont you need that check ?

Also doing this in collect has possibility that it wont be marked as external if it was already collected? - though not sure if thats what we want or not

jakebailey: > if the task is root it isnt marked as from external library - so dont you need that check ?

The way that this code works, a file can only be marked external if it was loaded via another task (via `addSubTask`). Root files don't get the marker, and later if we ever re-encounter them via another import path that _is_ external, the `isRoot` check above prevents us from forcing it to be external.

> Also doing this in collect has possibility that it wont be marked as external if it was already collected? - though not sure if thats what we want or not

This is the top level `collect`, not `collectWorker`, so is guaranteed to have been called first (and also only once).

sheetalkamat: Confused it to collectWorker.

sheetalkamat (approved):

PR #1277, opened 2025-06-24
Rename transform-relevant subtree facts with what they actually should track

weswigham:
Following up my comments in #1270, this PR adjusts the terminology used in the subtree facts enum to make plain it's about optimizing transformations. The old names are left in as aliases for reference when porting code, but you should really use (one of) what they alias. This renaming already called out some nodes which set `ContainsESXXXX` flags but didn't have corresponding transform logic, causing us to needlessly perform traversals on ASTs containing those node kinds (namely, `bigint`s and `import.meta` were big offenders that triggered a no-op `es2020` transform, and an `async` modifier would guarantee a function hits both the `es2018` and `es2017` transforms, even though the containing function set the more specific individual transform flag already).

### DIFF
--- a/internal/ast/subtreefacts.go
+++ b/internal/ast/subtreefacts.go
@@ -8,18 +8,22 @@ type SubtreeFacts int32
 
 const (
 	// Facts
-	// - Flags used to indicate that a node or subtree contains syntax specific to a particular ECMAScript variant.
+	// - Flags used to indicate that a node or subtree contains syntax relevant to a specific transform
 
 	SubtreeContainsTypeScript SubtreeFacts = 1 << iota
 	SubtreeContainsJsx
-	SubtreeContainsESNext
-	SubtreeContainsES2022
-	SubtreeContainsES2021
-	SubtreeContainsES2020
-	SubtreeContainsES2019
-	SubtreeContainsES2018
-	SubtreeContainsES2017
-	SubtreeContainsES2016
+	SubtreeContainsESDecorators
+	SubtreeContainsUsing
+	SubtreeContainsClassStaticBlocks
+	SubtreeContainsESClassFields
+	SubtreeContainsLogicalAssignments
+	SubtreeContainsNullishCoalescing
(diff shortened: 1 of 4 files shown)


internal/ast/subtreefacts.go
```diff
@@ -8,18 +8,22 @@ type SubtreeFacts int32
...
-	SubtreeContainsES2022
-	SubtreeContainsES2021
-	SubtreeContainsES2020
-	SubtreeContainsES2019
-	SubtreeContainsES2018
-	SubtreeContainsES2017
-	SubtreeContainsES2016
+	SubtreeContainsESDecorators
+	SubtreeContainsUsing
+	SubtreeContainsClassStaticBlocks
```
jakebailey, line 17: Is this list exhaustive all the way down to ES6?

weswigham: I believe so. I double checked what our transforms cover against the list of accepted proposals from each year. The only thing I'm still a bit undecided on split-wise is if I really want to split class static blocks out from class fields - they're combined into one transform (along with the legacy class fields transform) in strada. Some decorator things compile to class static blocks, though, so it has the potential to save some node traversal on classes-which-use-decorators-but-not-fields by being split out. They're also separate tc39 proposals, even though they got accepted in the same year... eh. I'll finalize that one way or another when I put up the PR for it - this split is fine for now. Class fields and `async` transforms are probably going to be the last two I post, given their size - honestly I forgot core `async` wasn't es6 and so was still in the wheelhouse of needing transformation logic when I started working on these.

jakebailey: I'm just surprised that it's such a short list.

weswigham: Eh, some end up in the `module` transforms (like import attributes) and some just don't get downlevel support due to complexity (like `bigint`), but most new es features end up just being lib updates and not new syntax. 🤷‍♂️

PR #1251, opened 2025-06-20
Set parents during parse, not bind

jakebailey:
When loading files, Program construction needs to look a file's imports and resolve them. This requires parent pointers as determining what to do depends on context like the resolution mode, if it's in a require, etc.

This is a problem because binding can happen concurrently on the same AST as the same time as file loading, since ASTs can be reused between Programs.

We've worked around this by setting parents early in the parser just on the imports (was done in strada too), yet that still isn't enough to make the code not race (#970, #1031, #1224).

A simple way to solve this problem is to just make the parser the one in charge of setting pointers, rather than doing it during bind. This is slower, since the binder is already walking the AST and so usually has thing cached. But, it is a lot simpler to reason about.

Below are benchmarks which show the change; the real time itself is the critical number, showing that while in the checker, we save a millisecond in bind, we pay some 4ms for it during parse, likely due to caching / walking overhead.

If this seems undesirable, there are other fixes that we could try:

- Instead of collecting specifiers during parse to later resolve, instead collect the imports themselves, then walk down the tree during file loading to get the specifier, never using parent pointers. That avoids the race, but is a more extensive change to get the helpers to not use parents.
- Force binding to happen _immediately_ after file parse before looking at imports. Realistically, we need every file to have been bound anyway, so this would be "no slower", but would make it a challenge to track how much time was spent in bind.

```
                                                     │   old.txt   │               new2.txt               │
                                                     │   sec/op    │    sec/op     vs base                │
Bind/empty.ts-20                                       301.2n ± 6%   278.1n ± 32%        ~ (p=0.123 n=10)
Bind/checker.ts-20                                     19.52m ± 2%   18.60m ±  1%   -4.72% (p=0.000 n=10)
Bind/dom.generated.d.ts-20                             8.208m ± 0%   7.142m ±  1%  -12.99% (p=0.000 n=10)
Bind/Herebyfile.mjs-20                                 330.2µ ± 1%   310.5µ ±  0%   -5.95% (p=0.000 n=10)
Bind/jsxComplexSignatureHasApplicabilityError.tsx-20   152.6µ ± 1%   137.3µ ±  1%  -10.03% (p=0.000 n=10)
geomean                                                300.1µ        275.1µ         -8.32%
```

```
                                                             │   old.txt   │              new2.txt               │
                                                             │   sec/op    │   sec/op     vs base                │
Parse/empty.ts/tsc-20                                          482.3n ± 1%   518.2n ± 1%   +7.44% (p=0.000 n=10)
[...]

PR #1279, opened 2025-06-24
Set parent pointers as nodes are ready in parser, rather than all at once as parse is complete

weswigham:
This is, imo, an improvement on https://github.com/microsoft/typescript-go/pull/1251. It's _almost_ very straightforward to just set parents in `finishNode` (which would be ideal as it's when all the relevant pointers are in cache)... except for all the cases where we reparse a node and need to forcibly overwrite parent pointers we already set (which can't be the default behavior, as last-in-wins parent setting by default breaks the jsdoc reparser which currently only ends up setting parents on new nodes, and not copied or moved ones).

This also exposed an interesting issue with the `@satisfies` jsdoc transform and ended up reducing the diffs for it - by overriding the `.Parent` explicitly on the transformed result, as is needed to ensure one is always set in this manner (in current `main`, the parent of the synthetic `satisfies` expression/type only gets set if it hasn't been set by the main file walk, which is to say basically never), we get correct context lookup behavior in the checker.


internal/parser/parser.go
```diff
@@ -5888,6 +5915,16 @@ func (p *Parser) finishNodeWithEnd(node *ast.Node, pos int, end int) {
 		node.Flags |= ast.NodeFlagsThisNodeHasError
 		p.hasParseError = false
 	}
+	p.currentParent = node
+	node.ForEachChild(p.setParent)
```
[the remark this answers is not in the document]
weswigham: Unneeded in this case - it's not recursive and only used to smuggle context for the `setParent` method.

jakebailey:
This is an improvement I was hoping we'd "eventually" send, so I'm happy to see that "eventually" could be "right now" 😄 

The JSX case was the one case I knew we did something funky with. Hopefully there aren't any more? I know @andrewbranch mentioned that we may not know the right parents for the reparsed nodes early enough.

Is there a way we could in testing verify that the parents are set properly, like potentially in the harness's `GetSourceFile` function `ForEachChild` like `SetParentsInChildren` and check that the parent is the caller?

internal/parser/reparser.go
```diff
@@ -5,6 +5,11 @@ import (
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
+func (p *Parser) finishReparsedNode(node *ast.Node) {
```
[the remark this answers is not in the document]
weswigham: We keep this as-is so it's in the same style as `finishNode` in the main parser codepath, which it is analogous to. Honestly, it should probably _also_ handle `Loc` setting so it's not as ad-hoc throughout the reparser, but that's a refactor for another day.

weswigham:
> Is there a way we could in testing verify that the parents are set properly, like potentially in the harness's GetSourceFile function ForEachChild like SetParentsInChildren and check that the parent is the caller?

Probably. I think we used to have tests that did this in Strada to check that `.parent` got fixed up correctly after incremental reparsing, actually. We should probably just resurrect those assertions.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag1.errors.txt.diff
```diff
@@ -1,51 +1,24 @@
...
--     * @typedef {Object} T4
--     * @property {string} a
--     */
--    
--    const t1 = /** @satisfies {T1} */ ({ a: 1 });
--    const t2 = /** @satisfies {T1} */ ({ a: 1, b: 1 });
--                                               ~
--!!! error TS2353: Object literal may only specify known properties, and 'b' does not exist in type 'T1'.
--    const t3 = /** @satisfies {T1} */ ({});
++/a.js(22,38): error TS2741: Property 'a' is missing in type '{}' but required in type 'T1'.
```
jakebailey, line 7: Nice, this is actually a desired change IIRC, some new error stack collapsing

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag10.errors.txt.diff
```diff
@@ -2,19 +2,16 @@
...
-     
-     const p = /** @satisfies {Partial<Record<Keys, unknown>>} */ ({
-         a: 0,
+@@= skipped -9, +9 lines =@@
          b: "hello",
          x: 8 // Should error, 'x' isn't in 'Keys'
--        ~
+         ~
 -!!! error TS2353: Object literal may only specify known properties, and 'x' does not exist in type 'Partial<Record<Keys, unknown>>'.
++!!! error TS2353: Object literal may only specify known properties, and 'x' does not exist in type 'Partial<Record<"a" | "b" | "c" | "d", unknown>>'.
```
jakebailey: Weird, where did `a b c d` come from?

weswigham: Expansion of `Keys` - presumably a difference in type node printback behavior to do with not currently having most input node reuse logic in place right now.

jakebailey:
FWIW I could have sworn the intention was that any reparsed nodes were copies and their parents set in a "normal" way, so I'd be interested in fixing that stuff too at some point. The lest special cases the better...

weswigham:
> The lest special cases the better...

Yeah.... the `await` reparser and `js` reparser also behave differently when it comes to updating parent nodes - the js reparser mutates nodes in-place, while the `await` reparser returns a new parent node. Both have differing tradeoffs when it comes to setting parent pointers (the await reparser has to mutate parents on all the retained old nodes, the js reparser has to ensure parents are fixed up on nodes moved via mutation), but I'd appreciate it if we were at least consistent, so we could use a consistent style of parent-(re)setting.

weswigham:
> like potentially in the harness's GetSourceFile function ForEachChild like SetParentsInChildren and check that the parent is the caller?

I think we could assert non-`nil` parents right now (or rather, we definitely should), but the js reparser's laissez faire attitude with moving nodes around and keeping the same node accessible in multiple places (?!¿) makes me think we can't currently rely totally on that matching visitor-ordered traversal parents. Not in files where it's been invoked, anyway.

jakebailey:
> the js reparser's laissez faire attitude with moving nodes around and keeping the same node accessible in multiple places

If this is still happening, I'm pretty sure it's a bug. I know I had found one place in the reparser that was missing a clone, but perhaps there are more.

jakebailey:
When I test main vs this branch+main, the perf is a lot worse:

```
goos: linux
goarch: amd64
Parse/empty.ts/tsc-20                                          495.3n ± 1%   522.9n ± 1%   +5.58% (p=0.000 n=10)
Parse/empty.ts/server-20                                       495.1n ± 1%   525.1n ± 1%   +6.07% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        44.96m ± 2%   53.98m ± 2%  +20.06% (p=0.000 n=10)
Parse/checker.ts/server-20                                     46.18m ± 1%   55.07m ± 2%  +19.23% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                15.07m ± 2%   18.54m ± 1%  +23.07% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             21.94m ± 1%   25.63m ± 1%  +16.85% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    826.6µ ± 1%   945.9µ ± 0%  +14.43% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 830.2µ ± 1%   945.8µ ± 1%  +13.93% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      251.0µ ± 2%   297.5µ ± 1%  +18.51% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   382.5µ ± 1%   434.5µ ± 2%  +13.58% (p=0.000 n=10)
geomean                                                        637.5µ        733.1µ       +15.00%
```
```
                                                             │     B/op     │     B/op       vs base                │
Parse/empty.ts/tsc-20                                            881.0 ± 0%      912.0 ± 0%   +3.52% (p=0.000 n=10)
Parse/empty.ts/server-20                                         881.0 ± 0%      912.0 ± 0%   +3.52% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        25.69Mi ± 0%    30.22Mi ± 0%  +17.63% (p=0.000 n=10)
Parse/checker.ts/server-20                                     25.87Mi ± 0%    30.40Mi ± 0%  +17.53% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                9.245Mi ± 0%   10.789Mi ± 0%  +16.70% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             11.31Mi ± 0%    13.04Mi ± 0%  +15.28% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    450.7Ki ± 0%    510.0Ki ± 0%  +13.16% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 450.7Ki ± 0%    510.0Ki ± 0%  +13.16% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      164.9Ki ± 0%    187.6Ki ± 0%  +13.75% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   244.8Ki ± 0%    272.7Ki ± 0%  +11.41% (p=0.000 n=10)
geomean                                                        464.1Ki         521.9Ki       +12.45%
```
```
                                                             │   old.txt   │                new.txt                 │
                                                             │  allocs/op  │  allocs/op    vs base                  │
Parse/empty.ts/tsc-20                                           4.000 ± 0%     6.000 ± 0%    +50.00% (p=0.000 n=10)
Parse/empty.ts/server-20                                        4.000 ± 0%     6.000 ± 0%    +50.00% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        18.09k ± 0%   314.28k ± 0%  +1637.05% (p=0.000 n=10)
Parse/checker.ts/server-20                                     18.41k ± 0%   315.45k ± 0%  +1613.36% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                11.49k ± 0%   112.69k ± 0%   +881.13% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             13.50k ± 0%   126.93k ± 0%   +840.41% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    1.825k ± 0%    5.634k ± 0%   +208.71% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 1.825k ± 0%    5.634k ± 0%   +208.71% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20       278.0 ± 0%    1734.0 ± 0%   +523.74% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20    404.0 ± 0%    2200.0 ± 0%   +444.55% (p=0.000 n=10)
geomean                                                         889.2         4.776k        +437.16%
```

This seems to be due to the closure thing again:

image

weswigham:
*sigh* always the method closure overhead. I'm willing to bet if I write an actual small inline closure instead of a method reference it'll actually get inlined.

In other news, though, the jsdoc reparser reparses
```js
/** @type {string} */
module.exports = 0;
```
into a JS export declaration with a `.Type` set to the `string` type node, but _also_ reparses the same type node to set a `.Type` on the original binary expression - this is a big no-no - the same node in two traversable parts of the AST, but with different logical parent nodes. It's impossible to satisfy a traversal-order-parenting assertion with this logic in place (and I'm sure there's other downstream invariants that get broken by it).

The jsdoc reparser only gets away with the mutations it does normally because the `.jsDoc` member the source nodes come from isn't part of the child preorder traversal set. It really should be deep cloning any nodes it moves, for each destination location, for exactly this reason.

internal/parser/reparser.go
```diff
@@ -6,8 +6,10 @@ import (
 )
 
 func (p *Parser) finishReparsedNode(node *ast.Node) {
-	p.currentParent = node
-	node.ForEachChild(p.setParent)
+	node.ForEachChild(func(n *ast.Node) bool {
```
jakebailey: ForEachChild is an indirect call, so is not inlinable and escapes its argument; this is why I had to do the goofy parent setting thing in `SetParentsInChildren`. I would suspect that this wouldn't work.

weswigham:
> I know I had found one place in the reparser that was missing a clone, but perhaps there are more.

You sure someone didn't walk this back? The text "Clone" doesn't even appear inside the reparser.

jakebailey:
I found it, but it was never committed as we just undid the parent code in the binder instead. https://github.com/microsoft/typescript-go/pull/1187#discussion_r2146132042 is actually one such case.

andrewbranch:
> this is a big no-no

That’s what I said too, but apparently that’s the design that @ahejlsberg and @sandersn discussed. The invariant that was _intended_ to hold is weaker: that real nodes have real parents. So a real node is allowed to appear inside a reparsed node, but its parent should point into the real AST. In other words, if you visit the children of a reparsed subtree, you may not be able to get back up to the same place by parent pointer traversal. I don’t particularly like this, but that was the intended design, and many things broke when I tried adding consistent cloning.

weswigham:
> that real nodes have real parents

That's true, but reparsed nodes are, for all intents, real nodes. Like an `await` reparsed source file or the handful of places where we parse out an `ExpressionWithTypeArguments` but then discard it and pull the members of it into a different kinded node (all of which get called out in this PR as places which need exceptions to the normal parent setting behavior of `finishNode`). The js reparser is _not_ the first time we "reparse" things. They need real parents - and more importantly, they can't be shared into multiple (traversable) places in the AST - that's the real problem. Anything about the fake-ness of these nodes the reparser makes should be about their _positions_ (since, ofc, reparsed nodes with cloned positions violate the formatter's "all nodes contain the spans of all child nodes" invariant, so the formatter needs to know not to include them in its' formatting logic, but that's a bug for another day), not the data objects themselves. We can't go making our abstract tree into an abstract graph - some of the algorithms don't hold up, and running _those_ down is a lot harder than ensuring we keep to the tree abstraction in the first place.

> and many things broke when I tried adding consistent cloning.

Same here where swapping parents between multiple "correct" options causes changes! And looking through those breaks, they're places where "alternate but correct" parent types don't seem to be handled, like `JSExportDeclaration`s, since which parent is set today between, eg, binary expression and js export declaration, only comes down to traversal order in the parent setting walk.

sandersn:
To me, it sounds like we need to revisit the no-clone real-parent design again in a realtime discussion. I can't make the design meeting today but maybe on Friday.

The current design really doesn't turn the tree into a graph, though. It adds out-of-band secondary parents that are only checked in a few places. (Which are then treated inconsistently on commonjs exports to avoid adding a secondary parent field to BinaryExpression, as you noticed.) What algorithms are broken by it, if they skip nodes flagged as reparsed?

weswigham:
> It adds out-of-band secondary parents that are only checked in a few place

It does that, too, but it _also_, in a roundabout way, sets the `.Parent` in reparsed nodes simply by making reparsed nodes children of "real" nodes (as they then get hit by the parent node setting traversal in `main`); because they're in the tree twice in some cases, which position's `.Parent` wins is ??? (either the first or last topographically, depending on how the setter is written and when it's invoked) - this is the core logical problem that causes the `.Parent` setting data races we've been contending with, I imagine, as it prevents us from setting parents on partial trees and having a singular canonical output.

jakebailey:
I think we're likely to merge #1241 soon, so hopefully that isn't too hard to merge into this PR.

weswigham:
@jakebailey I had to except the cases where the reparser reuses nodes, but otherwise the tests show the assertion holding now. There's some baseline diffs as in some of those cases, a different parent is selected now; but the baselines show mostly improvements (or at least reduced diffs) from the current post-hoc parent selection method. I'm pretty disappointed on the manual indirections needed to iterate node children and set parent pointers efficiently though. :(

jakebailey:
Here's the current perf diff. Memory is unchanged, but runtime gets slower:

```
Parse/empty.ts/tsc-20                                          493.9n ± 1%   510.5n ± 1%   +3.35% (p=0.000 n=10)
Parse/empty.ts/server-20                                       493.9n ± 1%   511.0n ± 1%   +3.45% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        45.59m ± 3%   50.64m ± 1%  +11.07% (p=0.000 n=10)
Parse/checker.ts/server-20                                     46.75m ± 1%   51.99m ± 2%  +11.21% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                15.16m ± 2%   17.38m ± 1%  +14.67% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             22.29m ± 2%   24.30m ± 1%   +9.02% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    834.9µ ± 1%   898.8µ ± 1%   +7.66% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 838.0µ ± 1%   896.5µ ± 1%   +6.97% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      254.6µ ± 2%   280.6µ ± 1%  +10.19% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   386.4µ ± 1%   414.4µ ± 1%   +7.25% (p=0.000 n=10)
geomean                                                        643.1µ        697.3µ        +8.43%
```

The profile is a little unclear, but it does imply that 10% of the time is spent in OverrideParentInImmediateChildren, whereas the old recursive setting was only 5% of the parse time.

Half of that is just messing with the pool; my guess is that while the old code would only get something out of the pool _once_ per AST, now we're doing it for every node on finish.

image

jakebailey:
Probably this just means we should be sticking this on the parser as a cached function pointer instead of a global?

internal/parser/parser.go
```diff
@@ -94,6 +103,7 @@ func getParser() *Parser {
 
 func putParser(p *Parser) {
 	*p = Parser{scanner: p.scanner}
+	p.initializeClosures()
```
```suggestion
	*p = Parser{scanner: p.scanner, setParentFromContext: p.setParentFromContext}
```

Otherwise this will just reclose each time the parser is put

jakebailey:
With the suggestion above, versus main:

```
Parse/empty.ts/tsc-20                                          495.7n ± 1%   475.2n ± 1%  -4.14% (p=0.000 n=10)
Parse/empty.ts/server-20                                       494.8n ± 0%   477.5n ± 1%  -3.49% (p=0.000 n=10)
Parse/checker.ts/tsc-20                                        44.85m ± 3%   43.94m ± 2%  -2.03% (p=0.000 n=10)
Parse/checker.ts/server-20                                     46.12m ± 1%   45.15m ± 1%  -2.12% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                15.11m ± 4%   15.05m ± 1%       ~ (p=0.684 n=10)
Parse/dom.generated.d.ts/server-20                             22.14m ± 1%   21.75m ± 1%  -1.73% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    843.9µ ± 1%   831.4µ ± 2%  -1.48% (p=0.004 n=10)
Parse/Herebyfile.mjs/server-20                                 847.0µ ± 1%   828.8µ ± 1%  -2.15% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      255.5µ ± 1%   249.6µ ± 1%  -2.32% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   390.4µ ± 1%   377.7µ ± 2%  -3.24% (p=0.000 n=10)
geomean                                                        643.1µ        628.2µ       -2.31%
```

jakebailey:
Main without parent setting in the parser, vs main vs this PR:

```
                                                             │ oldold.txt  │              old.txt               │              new.txt               │
                                                             │   sec/op    │   sec/op     vs base               │   sec/op     vs base               │
Parse/empty.ts/tsc-20                                          464.6n ± 1%   495.7n ± 1%  +6.71% (p=0.001 n=10)   475.2n ± 1%  +2.29% (p=0.001 n=10)
Parse/empty.ts/server-20                                       468.6n ± 2%   494.8n ± 0%  +5.59% (p=0.000 n=10)   477.5n ± 1%  +1.91% (p=0.005 n=10)
Parse/checker.ts/tsc-20                                        42.32m ± 2%   44.85m ± 3%  +5.96% (p=0.000 n=10)   43.94m ± 2%  +3.82% (p=0.004 n=10)
Parse/checker.ts/server-20                                     43.20m ± 3%   46.12m ± 1%  +6.76% (p=0.000 n=10)   45.15m ± 1%  +4.50% (p=0.000 n=10)
Parse/dom.generated.d.ts/tsc-20                                14.13m ± 1%   15.11m ± 4%  +6.91% (p=0.000 n=10)   15.05m ± 1%  +6.48% (p=0.000 n=10)
Parse/dom.generated.d.ts/server-20                             20.74m ± 2%   22.14m ± 1%  +6.73% (p=0.000 n=10)   21.75m ± 1%  +4.88% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    807.1µ ± 2%   843.9µ ± 1%  +4.56% (p=0.000 n=10)   831.4µ ± 2%  +3.01% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20                                 787.9µ ± 1%   847.0µ ± 1%  +7.50% (p=0.000 n=10)   828.8µ ± 1%  +5.19% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      236.1µ ± 2%   255.5µ ± 1%  +8.22% (p=0.000 n=10)   249.6µ ± 1%  +5.71% (p=0.000 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   363.8µ ± 1%   390.4µ ± 1%  +7.31% (p=0.000 n=10)   377.7µ ± 2%  +3.82% (p=0.000 n=10)
geomean                                                        603.2µ        643.1µ       +6.62%                  628.2µ       +4.15%
```

Not getting back as much as I thought...

jakebailey:
However, I think the numbers do correspond to the time _removed_ from the binder in #1251, so that may all be expected.

weswigham:
Yeah, as I mentioned offline, I think the only way we could get more efficient than this would be if we built setting parent pointers into the node factory functions to avoid the virtual dispatch on `ForEachChild` (which, as you said, would need to be toggleable functionality so the transformers don't override parse tree parents on children of transformed nodes), which'd need a fair amount of toil to add that functionality to the constructors (or a reasonably capable codegen - still wish most of `ast.go` was codegen!), then additional review to ensure anything mutating nodes after construction (eck) manually sets up parent pointers.

In any case, this gets us back to where we were-but-racing but without-a-race, so seems like a good place to stop for now.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag4.errors.txt.diff
```diff
@@ -3,24 +3,21 @@
...
--     * @typedef {Object} Foo
--     * @property {number} a
--     */
--    
--    export default /** @satisfies {Foo} */ ({ a: 1 });
-+<no content>
\ No newline at end of file
++                                              ~
++!!! error TS2741: Property 'a' is missing in type '{}' but required in type 'Foo'.
++!!! related TS2728 /a.js:3:23: 'a' is declared here.
```
jakebailey, line 20: Not for this PR but seems like we're not setting the range of the satisfies expression properly? That or we need to do some special handling to choose the JSDoc node to report the satisfies error on?

testdata/baselines/reference/submoduleAccepted/conformance/jsdocTemplateTag8.errors.txt.diff
```diff
@@ -2,85 +2,43 @@
...
-+==== a.js (12 errors) ====
++==== a.js (8 errors) ====
      /**
       * @template out T
 +                 ~~~
 +!!! error TS1274: 'out' modifier can only appear on a type parameter of a class, interface or type alias
       * @typedef {Object} Covariant
       * @property {T} x
-+                  ~
-+!!! error TS2304: Cannot find name 'T'.
```
jakebailey, line 37: Nice to fix this one.

jakebailey (commented):
Baseline improvements all seem good too.

internal/testrunner/compiler_runner.go
```diff
@@ -506,6 +508,44 @@ func (c *compilerTest) verifyUnionOrdering(t *testing.T) {
 	})
 }
 
+func (c *compilerTest) verifyParentPointers(t *testing.T) {
```
jakebailey, line 516: This will end up repeating the same tests given we share source files so much.

Maybe this should just be an assertion in the test harness's GetSourceFile implementation? Or, we `SyncSet[NodeId]` this?

Then again, it's only 4% of the total runtime.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocSatisfiesTag4.errors.txt.diff
weswigham: The location on the synthetic `satisfies` expression is set to the `satisfies` tag, but  `binder.GetErrorRangeForNode` isn't set up to handle reparsed nodes for this node kind, so this error span is accidentally close, at best. It's "the token range at the end of the trivia after the expression position", which for the `@satisfies` tag is the semicolon trivia we see underlined here.

internal/testrunner/compiler_runner.go
weswigham: We leave it as-is for now so someone can go and make tests 5% faster by removing a "low value test" after we're mostly done porting and nobody remembers them catching anything recently, like we did with these parent assert tests in strada. 😂

jakebailey: Yeah I simply cannot judge, my union order consistency checks take more time than the parent checks

testdata/baselines/reference/submoduleAccepted/conformance/jsdocTemplateTag8.errors.txt.diff
weswigham: Entirely accidental upside to going through the reparser and adding calls to manually set parents. Node just gets a `.Parent` set to a node in the real expression tree now, where it did not before (it got a jsdoc parent).

internal/checker/checker.go
```diff
@@ -30197,7 +30197,7 @@ func (c *Checker) getSymbolOfNameOrPropertyAccessExpression(name *ast.Node) *ast
 		return c.getSymbolOfNode(name.Parent)
 	}
 
-	if name.Parent.Kind == ast.KindExportAssignment && ast.IsEntityNameExpression(name) {
+	if (name.Parent.Kind == ast.KindExportAssignment || name.Parent.Kind == ast.KindJSExportAssignment) && ast.IsEntityNameExpression(name) {
```
sandersn, line 30200: if this is code ported since the introduction of JSExportAssignment, I think that means I'll need to make another pass through the source looking for places where all the synthetic kinds need to be handled alongside their non-synthetic peers.

internal/parser/jsdoc.go
```diff
@@ -977,6 +978,9 @@ func (p *Parser) parseTypedefTag(start int, tagName *ast.IdentifierNode, indent
 
 	typedefTag := p.factory.NewJSDocTypedefTag(tagName, typeExpression, fullName, comment)
 	p.finishNodeWithEnd(typedefTag, start, end)
+	if typeExpression != nil {
```
sandersn, line 981: nested `@typedef` pushes the whole syntax from tragedy to loop back around to comedy.

testdata/baselines/reference/submoduleAccepted/conformance/moduleExportWithExportPropertyAssignment3.types.diff
```diff
@@ -75,7 +75,7 @@
 +>module.exports : { justExport: number; bothBefore: number; bothAfter: number; }
 +>module : { "export=": { justExport: number; bothBefore: number; bothAfter: number; }; }
 +>exports : { justExport: number; bothBefore: number; bothAfter: number; }
-+>bothBefore : number
++>bothBefore : "string"
```
sandersn, line 78: this was `number | "string"` before, and switched types, so I *think* that's an effect of our stricter rules combined with different visit order. Or something like that?

testdata/baselines/reference/submoduleAccepted/conformance/moduleExportsElementAccessAssignment.types.diff
```diff
@@ -37,30 +37,35 @@
  >exports["b"] = { x: "x" } : { x: string; }
  >exports["b"] : { x: string; }
 ->exports : typeof import("mod1")
+->"b" : "b"
 +>exports : typeof import("./mod1")
- >"b" : "b"
++>"b" : { x: string; }
```
sandersn, line 42: this looks like a change in how the baseliner is interpreting `"b"`, from a string literal to a property name. The latter is defensible but definitely different from before.

jakebailey:
Turning off auto merge for now, given we were wanting to talk about this tomorrow and any approvals here will make it go in (it would have been merged just now if there weren't a merge conflict)

sandersn:
thanks, didn't notice that

jakebailey:
That being said, I'm pro-this-change because yay perf, just wanted to make sure we understood the model we're going for with this new reparsed AST.

testdata/baselines/reference/submoduleAccepted/conformance/moduleExportsElementAccessAssignment.types.diff
weswigham: Yeah, for sure a difference in the symbol and therefore type associated with the node.

weswigham:
> That being said, I'm pro-this-change because yay perf, just wanted to make sure we understood the model we're going for with this new reparsed AST.

Haven't changed anything here about that, this just tosses exceptions into the test you asked for in the cases that highlight the issues with the current model - we do need to talk about it, but that's followup work. I think we can take this as-is (sans merge conflict).

PR #1282, opened 2025-06-24
Fix typedef alias lookup

sandersn:
Just another missing call to GetEffectiveTypeParent. Cleans up baselines nicely.

Now with bonus fix to correct the host of typedef constraints.


testdata/baselines/reference/submoduleAccepted/compiler/checkJsTypeDefNoUnusedLocalMarked.types.diff
```diff
@@ -8,11 +8,10 @@
 ->module.exports : (foo: Foo) => string
 ->module : { exports: (foo: Foo) => string; }
 ->exports : (foo: Foo) => string
-->(void 0) : FooFun
-+>module.exports = /** @type {FooFun} */(void 0) : (foo: typeof import("./file")) => string
-+>module.exports : (foo: typeof import("./file")) => string
-+>module : { "export=": (foo: typeof import("./file")) => string; }
-+>exports : (foo: typeof import("./file")) => string
-+>(void 0) : (foo: typeof import("./file")) => string
++>module.exports = /** @type {FooFun} */(void 0) : FooFun
```
sandersn, line 11: This cast now correctly uses the alias--which is better than Strada.

testdata/baselines/reference/submoduleAccepted/compiler/exportDefaultWithJSDoc1.types.diff
```diff
@@ -11,11 +11,11 @@
...
 ->A : import("a").NumberLike[]
-+>A : (string | number)[]
++>A : import("./a").NumberLike[]
 
  A[0]
 ->A[0] : import("a").NumberLike
 ->A : import("a").NumberLike[]
-+>A[0] : string | number
-+>A : (string | number)[]
++>A[0] : import("./a").NumberLike
```
sandersn, line 19: we still print filenames differently =(

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsFunctionClassesCjsExportAssignment.types.diff
```diff
@@ -35,31 +35,28 @@
...
 ->module.exports : typeof Hook
 ->module : { exports: typeof Hook; }
 ->exports : typeof Hook
 ->Hook : typeof Hook
-+>module.exports = Hook : (handle: (arg: any) => void) => void
-+>module.exports : (handle: (arg: any) => void) => void
-+>module : { Hook(handle: (arg: any) => void): void; }
-+>exports : (handle: (arg: any) => void) => void
-+>Hook : (handle: (arg: any) => void) => void
++>module.exports = Hook : (handle: HookHandler) => void
```
sandersn, line 55: module.exports types still print differently. It's due to the different symbol flags, and the different (still fake) way that module.exports references are checked, so maybe not worth fixing.

sandersn (commented):
some diff explanations

testdata/baselines/reference/submoduleAccepted/conformance/typedefMultipleTypeParameters.errors.txt.diff
```diff
@@ -3,7 +3,7 @@
 @@= skipped -0, +0 lines =@@
 -a.js(12,23): error TS2344: Type '{ a: number; }' does not satisfy the constraint '{ a: number; b: string; }'.
 -  Property 'b' is missing in type '{ a: number; }' but required in type '{ a: number; b: string; }'.
-+a.js(12,23): error TS2741: Property 'b' is missing in type '{ a: number; }' but required in type '{ a: number; b: string; }'.
++a.js(12,23): error TS2741: Property 'b' is missing in type '{ a: number; }' but required in type 'Everything<{ a: number; }, undefined, { c: 1; d: 1; }, number, string>'.
```
gabritto: This new message seems wrong to me, but maybe I'm misunderstanding something? I thought the error was supposed to be actually between types `{ a: number }` and the constraint of the corresponding type parameter, `{ a: number, b: string }`. To me it seems that now we're somehow comparing the type argument to the whole instantiated type, which doesn't look right.

sandersn: Definitely wrong--well, it was previously not what Strada had, but even wronger than before.

sandersn: good catch; this fix exposed the fact that the host of typedef constraints was the typedef, not the type parameter it constrained

gabritto (approved):

PR #1286, opened 2025-06-25
Pool commentState allocations

jakebailey:
These comment states appear to be one of the main allocators when emitting the VS Code codebase. Pool them within a print run.

Reduces the VS Code alloc count from `36015276` to `28140625` (20% reduction).


jakebailey:
`emitCommentsAfterNode` returns early on `nil`; I tried using a zero value but it breaks things because a zero value is actually seemingly valid. ☹️

weswigham:
> I tried using a zero value but it breaks things because a zero value is actually seemingly valid. ☹️

A sentinel value with `-1` positions and `-1` text range positions should work instead, I think.

PR #1287, opened 2025-06-25
Pool EmitContext

jakebailey:
These objects can be reused, which is good because we create one of them per file to emit.


DanielRosenwasser (approved):
Probably a good idea - any before/after numbers?

jakebailey:
For the test runner, it accounts for 5% of the allocations (6179460), and with this PR just 0.1% (99838).

PR #1289, opened 2025-06-25
Avoid holding on to NodeBuilder in checker

jakebailey:
Holding onto the builder, the emit context, and therefore NodeFactory, holds onto a load of old nodes permanently. Create these fresh each time and let them get GC'd rather than try and manually clear everything. ~As an optimization, reuse the same builder while checking the same file.~

For a perf problem found in #988.


DanielRosenwasser:
Somewhat related question - do node builders use pooled allocators the same way the parser does?

jakebailey:
Yes, they do.

jakebailey:
Yeah, I previously tried that in the PR but it was unwieldy trying to get it right. But, I can retry.

jakebailey:
I think I have the change, but I'm going to wait until I get access to the problematic codebase before actually committing to it.

jakebailey:
Going to take this for now; I have access to the linked codebase and there are other problems than just this one.

PR #1290, opened 2025-06-25
Add completion helpers and port more completion fourslash tests

gabritto:
This PR mainly adds completion helpers like `completionGlobals` and `completionGlobalsPlus` to Corsa. Those helpers are in `internalfourslash/tests/util_test.go` (and copied to `internalfourslash/tests/gen/util_test.go`).
It also ports Strada tests that use those helpers.

Another main change is that, to make the `completionGlobals` helper work, this PR also makes a change to completion such that whenever the old code would set an `optionalReplacementSpan`, we set the `editRange` in `CompletionList.ItemDefaults`. Before, we were using the optional replacement span to set a `TextEdit` on each completion item, so we couldn't reliably use `completionGlobals` because depending on the completion position, each global completion item would have a `TextEdit` (or not).

There are, like always, a number of bug fixes needed to get the tests to at least not crash. This PR ports 68 new completion tests, 38 of which are failing tests added to the failing test list, and removes 13 tests from the failing list. A lot of the new failing tests are JS/JSDoc stuff that I think isn't properly implemented yet.


internal/ls/completions.go
```diff
@@ -4025,22 +4087,6 @@ func (l *LanguageService) createLSPCompletionItem(
 				Range:   *replacementSpan,
 			},
 		}
-	} else {
```
gabritto, line 4028: This moved to `setItemDefaults`.

internal/parser/parser.go
```diff
@@ -6452,7 +6452,7 @@ func extractPragmas(commentRange ast.CommentRange, text string) []ast.Pragma {
 		}
 	}
 	if commentRange.Kind == ast.KindMultiLineCommentTrivia {
-		text = text[:len(text)-2]
+		text = strings.TrimSuffix(text, "*/")
```
jakebailey, line 6455: When would the old code have failed?

gabritto: If you have an unclosed multiline comment, like `/* ...`, then it is wrong to unconditionally drop the last two characters, because they won't be `*/`, and if the comment is empty, e.g. `/*`, this later causes an out of bounds panic in the call to `skipTo`, since we start at `pos := 2`.

jakebailey (approved):

PR #1291, opened 2025-06-25
Replace no-default-lib with IsLibFile on Program

jakebailey:
`no-default-lib` in a file no longer has any effect. Instead, the file loader keeps track of which files are libs, which is then used later for `skipDefaultLibCheck`.

This creates a load of errors in `1.0lib-noErrors.ts`, which is our test for `no-default-lib`.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -539,6 +539,7 @@ type Program interface {
 	GetImportHelpersImportSpecifier(path tspath.Path) *ast.Node
 	SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmit bool) bool
 	IsSourceFromProjectReference(path tspath.Path) bool
+	IsLibFile(path tspath.Path) bool
 	GetSourceAndProjectReference(path tspath.Path) *tsoptions.SourceAndProjectReference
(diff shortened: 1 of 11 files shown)


internal/checker/checker.go
```diff
@@ -539,6 +539,7 @@ type Program interface {
 	GetImportHelpersImportSpecifier(path tspath.Path) *ast.Node
 	SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmit bool) bool
 	IsSourceFromProjectReference(path tspath.Path) bool
+	IsLibFile(path tspath.Path) bool
```
sheetalkamat, line 542: btw strada has `isSourceFileDefaultLibrary ` instead

jakebailey: Yeah, though that name is pretty redundant; the only files we have a are source files, and this function reports true even on lib replaced files.

jakebailey: Though, the option is still called `skipDefaultLibCheck`.

jakebailey: Renamed it to `IsSourceFileDefaultLibrary`.

PR #1297, opened 2025-06-26
Enable the noCheck emit consistency test

weswigham:
Present, but commented out, these tests are important to have on as I port emit features, as they point out places where the `EmitResolver` fails to lazily calculate things emit needs. I'm also removing the eager-diagnostics behavior from emit resolver construction entirely. With emit often happening on many separate threads, it is basically never a good idea to require an emit resolver generate diagnostics - right now we lock on a single checker across emit threads, but we can and probably should swap to a pooled checker/emitresolver model and remove all those locks, just like we do for LS requests. Any questions an emit resolver answers really do just have to be lazy-calculation-safe under a concurrent model.

### DIFF
--- a/internal/testutil/tsbaseline/js_emit_baseline.go
+++ b/internal/testutil/tsbaseline/js_emit_baseline.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/parser"
 	"github.com/microsoft/typescript-go/internal/testutil/baseline"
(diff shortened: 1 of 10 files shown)


internal/testutil/tsbaseline/js_emit_baseline.go
```diff
@@ -98,44 +99,40 @@ func DoJSEmitBaseline(
...
-	////			}
-	////		}
-	////	}
-	////	compareResultFileSets(withoutChecking.DTS, result.DTS)
-	////	compareResultFileSets(withoutChecking.JS, result.JS)
-	////}
+	if !options.NoCheck.IsTrue() && !options.NoEmit.IsTrue() {
+		testConfig := make(map[string]string)
+		testConfig["noCheck"] = "true"
+		withoutChecking := result.Repeat(testConfig)
```
jakebailey, line 105: Do you know if this repeat is sufficient to test this? I'm reading the profile (since it of course makes tests longer), and I'm noticing that we pull on declaration diagnostics in these repeats, which I would think would cause a bunch to be eagerly resolved. But, maybe this was already happening in Strada?

weswigham: It's definitely the same as in strada, at least - the emit resolver doesn't share many entrypoints between declaration emit and normal emit, though, so it's still pretty likely to highlight differences in emit between pre-checked files and unchecked files. I'll probably improve how decoupled it is as part of trying to wrest these locks out of the `EmitResolver` (by fetching diagnostics in parallel with emit on a separate thread+checker).

gabritto (approved):
For my future reference: why did we force eager diagnostics in Strada in the emitter?

weswigham:
A legacy of pre-`noCheck` support - the `EmitResolver`'s APIs had no lazy fallbacks and looked up data directly in checker caches.

PR #1301, opened 2025-06-26
Enable pre/post emit diagnostic consistency tests

weswigham:
These assertions are here to ensure that the emit process doesn't create diagnostics that the main checker diagnostic pass doesn't, and doesn't block any diagnostics from being produced after emit. 

Unfortunately, unlike the `noCheck` emit consistency checks, these have a bunch of new failures that have snuck in while this check was off that need triage first.


weswigham:
Superseded by #4526

PR #1304, opened 2025-06-27
File loader diags

jakebailey:


jakebailey:
This PR was a noop without #1302, so I merged it into that one.

PR #1307, opened 2025-06-27
Port TS PR 61764 (for (using of = is incorrectly parsed) 

a-tarasyuk:
Ports in https://github.com/microsoft/TypeScript/pull/61764


jakebailey:
This isn't in 5.8, but theoretically we could fix this now?

a-tarasyuk:
@jakebailey, it depends — the change is simply an update to follow the spec, so it can be fixed now or deferred to the next release. -1 PR for the port. :)

jakebailey (approved):

Issue #1265, opened 2025-06-23
visual studio extension doesn't (quickly) detect updated packages


lukpsaxo:
I had a @types/lodash package installed that caused errors, then switched branch to one with a previous version and did yarn install.

at 07:08:21 the yarn command had finished and the editor showed errors.

I started raising this bug and seemingly without interaction at 07:11:01 the errors disappeared (the extension noticed that node_modules had been updated)

I've seen this behaviour when switching branches and running yarn before so I think it should be easily reproducible.

The logs looked like this:

````
Info 142  [07:08:21.025] createGlobMapper took 80.1279ms to create 11 globs for 153405 failed lookups
Info 143  [07:08:21.025] createGlobMapper took 0s to create 6 globs for 349 failed lookups
Info 144  [07:08:21.025] Finishing updateGraph: Project: c:/Projects/x/y/src/frontend/tsconfig.json version: 5 in 1.3436852s
Info 145  [07:11:01.206] Starting updateGraph: Project: c:/Projects/x/y/src/frontend/tsconfig.json
Info 146  [07:11:01.206] Program 6 used 3 checker(s)
Info 147  [07:11:01.206] Starting updateGraph: Project: c:/Projects/x/y/src/toolkit/tsconfig.json
Info 148  [07:11:01.206] Program 6 used 2 checker(s)
Info 149  [07:11:01.206] Starting updateGraph: Project: c:/Projects/x/y/src/ui/tsconfig.json
Info 150  [07:11:01.206] Program 4 used 2 checker(s)
Info 151  [07:11:01.334] 

````

**Comment by jakebailey** [comment]
It'd be interesting if you could look at the raw LSP logs (the other output window) and see when the watch events came in.

All the new server does is ask VS Code to watch the files, so we'd need to know where the delay is.

**Comment by lukpsaxo** [comment]
I re-did the test. Again it eventually gets the right types after 3 minutes.

The LSP server sends a notification whilst yarn is installing - so you see one notification and then a 2nd with the rest of the files.

```

[Trace - 10:38:40 AM] Sending notification 'workspace/didChangeWatchedFiles'.
Params: {
    "changes": [
        {
            "uri": "file:///c%3A/Projects/x/y/node_modules/react-stately/node_modules/%40internationalized/date/dist/BuddhistCalendar.main.js",
            "type": 3
        },
        {
[...]

PR #1308, opened 2025-06-27
Deduplicate watched file events, avoid MatchesFileName

jakebailey:
These are two tactics which I think will help with #1265.

I've noticed that VS Code can send duplicate events; deduplicate them before handling.

Also, we can avoid the expensive `MatchesFileName` call if the config file entry is already pending, hopefully saving even more time.

### DIFF
--- a/internal/project/service.go
+++ b/internal/project/service.go
@@ -302,7 +302,7 @@ func (s *Service) OnWatchedFilesChanged(ctx context.Context, changes []*lsproto.
 	s.projectsMu.RLock()
 	defer s.projectsMu.RUnlock()
 	for _, change := range changes {
-		if seen.AddIfAbsent(*change) {
+		if !seen.AddIfAbsent(*change) {
 			continue
 		}
 


internal/project/service.go
```diff
@@ -296,6 +298,17 @@ func (s *Service) Close() {
...
 func (s *Service) OnWatchedFilesChanged(ctx context.Context, changes []*lsproto.FileEvent) error {
+	{
+		deduped := slices.Clone(changes)
+		slices.SortFunc(deduped, func(a, b *lsproto.FileEvent) int {
+			if c := strings.Compare(string(a.Uri), string(b.Uri)); c != 0 {
+				return c
+			}
+			return cmp.Compare(a.Type, b.Type)
+		})
+		changes = slices.CompactFunc(deduped, func(a, b *lsproto.FileEvent) bool { return *a == *b })
```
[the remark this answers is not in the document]
jakebailey: Terrible code suggestion, but not wrong

internal/project/service.go
```diff
@@ -296,6 +297,11 @@ func (s *Service) Close() {
 }
 
 func (s *Service) OnWatchedFilesChanged(ctx context.Context, changes []*lsproto.FileEvent) error {
+	{
```
sheetalkamat: insterad of duplicating like this, may be create set to add if absent and continue otherwise

jakebailey: I'm not sure what you mean, isn't that what I have?

sheetalkamat: i meant move inside the processing loop instead of having it separately

jakebailey: Ah, gotcha; done.

sheetalkamat (approved):

PR #1310, opened 2025-06-27
Add `edit` functionality to fourslash

gabritto:
This PR adds functions like `edit.insert(...)` to fourslash, and enables the test conversion script to convert Strada tests that use those functions.

Note that the fourslash test object now has to keep track of file changes, so it has a map of `scriptInfo`s. It also has to convert between absolute positions and LSP positions.

This PR adds 41 new converted tests, 24 of which are skipped due to failing.


jakebailey:
The git diff is showing stuff I think has already been merged in main; bad merge maybe?

gabritto:
> The git diff is showing stuff I think has already been merged in main; bad merge maybe?

No, I'm just rebasing and re-generating stuff right now, you just looked before I'm done with it 😅

internal/fourslash/fourslash.go
```diff
@@ -28,13 +29,47 @@ type FourslashTest struct {
 	in     *lspWriter
 	out    *lspReader
 	id     int32
+	vfs    vfs.FS
 
-	testData *TestData
+	testData *TestData // !!! consolidate test files from test data and script info
```
gabritto, line 34: Weirdly some requests in Strada's fourslash use the original files instead of the fourslash object's script info with the updated content, so I'm not sure how things are gonna go and have kept the whole test data here for now.

PR #1320, opened 2025-06-30
Pool sourceMapState tool

jakebailey:
Realized that I missed this in #1286.


jakebailey:
I'll start with this and send the value change separately.

PR #1322, opened 2025-06-30
Incremental port

sheetalkamat:


internal/ast/diagnostic.go
```diff
@@ -21,6 +21,7 @@ type Diagnostic struct {
 	relatedInformation []*Diagnostic
 	reportsUnnecessary bool
 	reportsDeprecated  bool
+	skippedOnNoEmit    bool
```
jakebailey, line 24: Do the other functions that compare/do equality need to be updated to check this?

internal/collections/set.go
```diff
@@ -48,6 +50,32 @@ func (s *Set[T]) AddIfAbsent(key T) bool {
...
+		return false
+	}
+	if s.Len() != other.Len() {
+		return false
+	}
+	for key := range s.M {
+		if !other.Has(key) {
+			return false
+		}
+	}
```
jakebailey: I think you could probably just use `maps.Equal(s.M, other.M)`.

internal/collections/syncset.go
```diff
@@ -16,3 +16,19 @@ func (s *SyncSet[T]) Add(key T) {
...
 	s.m.Delete(key)
 }
+
+func (s *SyncSet[T]) Range(fn func(key T) bool) {
+	s.m.Range(func(key T, value struct{}) bool {
+		return fn(key)
+	})
+}
+
+func (s *SyncSet[T]) ToArray() []T {
```
jakebailey: I'd say `ToSlice()` here, just since it's not actually a Go array.

internal/compiler/emitter.go
```diff
@@ -462,3 +469,93 @@ func getDeclarationDiagnostics(host EmitHost, file *ast.SourceFile) []*ast.Diagn
 	transform.TransformSourceFile(file)
 	return transform.GetDiagnostics()
 }
+
+type AnyProgram interface {
```
jakebailey: Is there a better name for this than "AnyProgram"?

internal/compiler/host.go
```diff
@@ -72,6 +77,9 @@ func (h *compilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast.Sourc
 }
 
 func (h *compilerHost) GetResolvedProjectReference(fileName string, path tspath.Path) *tsoptions.ParsedCommandLine {
-	commandLine, _ := tsoptions.GetParsedCommandLineOfConfigFilePath(fileName, path, nil, h, nil)
+	if h.extendedConfigCache == nil {
+		h.extendedConfigCache = &collections.SyncMap[tspath.Path, *tsoptions.ExtendedConfigCacheEntry]{}
```
jakebailey: This seems like it is concurrency unsafe. But, the coverage never says this happens, so maybe this should just panic?

internal/execute/tsc.go
```diff
@@ -41,25 +41,20 @@ func applyBulkEdits(text string, edits []core.TextChange) string {
 	return b.String()
 }
 
-func CommandLine(sys System, cb cbType, commandLineArgs []string) ExitStatus {
+func CommandLine(sys System, commandLineArgs []string, testing bool) (ExitStatus, *incremental.Program, *Watcher) {
```
jakebailey: Is it time to start returning a struct here? I assume that this is all just here because testing needs more info out of the CommandLine function?

internal/compiler/emitter.go
```diff
@@ -462,3 +469,93 @@ func getDeclarationDiagnostics(host EmitHost, file *ast.SourceFile) []*ast.Diagn
...
+	configFileParsingDiagnosticsLength := len(allDiagnostics)
+
+	allDiagnostics = append(allDiagnostics, program.GetSyntacticDiagnostics(ctx, file)...)
+	allDiagnostics = append(allDiagnostics, program.GetProgramDiagnostics()...)
+
+	if len(allDiagnostics) == configFileParsingDiagnosticsLength {
+		// Options diagnostics include global diagnostics (even though we collect them separately),
+		// and global diagnostics create checkers, which then bind all of the files. Do this binding
+		// early so we can track the time.
+		bindStart := recordTime("bind", true, time.Time{})
```
jakebailey: I don't know how I feel about this pattern... Feels a bit weird for the emitter file to be the one with this time tracking code, and then it be indiredcted through callbacks and so on, when it was quite direct before. What was wrong with the old way? Can this just return a function closure that is used to start/stop the range?

internal/execute/tsc.go
```diff
@@ -241,7 +264,28 @@ func performCompilation(
...
+		reportDiagnostic,
+		configTime,
+		parseTime,
+	)
+}
 
+func emitAndReportStatistics(
+	sys System,
+	program compiler.AnyProgram,
+	getCoreProgram func() *compiler.Program,
```
jakebailey: Is this indirection needed? Can't just be immediately invoked? Or just use `program`?

internal/incremental/emitfileshandler.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (h *emitFilesHandler) updateSnapshot() []*compiler.EmitResult {
+	h.signatures.Range(func(file tspath.Path, signature string) bool {
+		info := h.program.snapshot.fileInfos[file]
+		info.signature = signature
+		if h.program.updatedSignatureKinds != nil {
+			h.program.updatedSignatureKinds.Store(file, SignatureUpdateKindStoredAtEmit)
+		}
+		h.program.snapshot.buildInfoEmitPending = true
```
jakebailey: I haven't gotten there yet, but it strikes me as odd that we are atomically mutating one thing at the same time as we are not atomically mutating another thing, when both are sort of "on" the same Program... I can't really put my finger on it but it gives me the bad feeling of the model not being consistent and free of logical races.

internal/tsoptions/enummaps.go
```diff
@@ -217,6 +217,11 @@ var targetToLibMap = map[core.ScriptTarget]string{
 	core.ScriptTargetES2015: "lib.es6.d.ts", // We don't use lib.es2015.full.d.ts due to breaking change.
 }
 
+// Used for test to add the files
```
jakebailey: I would just delete this; I initially read this as it saying that a test was _modifying_ the map, when it's just adding fake files.

jakebailey: That being said, why do we need this when the harness already knows how to do this without this mapping?

jakebailey (commented):
Seemingly still logical races or something; note the CI failure:

```
Using D:\a\typescript-go\typescript-go\Herebyfile.mjs to run baseline-accept
Starting baseline-accept
Finished baseline-accept in 19ms
Completed baseline-accept in 20ms
diff --git a/testdata/baselines/reference/tsc/incremental/when-global-file-is-added,-the-signatures-are-updated.js b/testdata/baselines/reference/tsc/incremental/when-global-file-is-added,-the-signatures-are-updated.js
index e3612043..f4adbdf5 100644
--- a/testdata/baselines/reference/tsc/incremental/when-global-file-is-added,-the-signatures-are-updated.js
+++ b/testdata/baselines/reference/tsc/incremental/when-global-file-is-added,-the-signatures-are-updated.js
@@ -551,7 +551,6 @@ declare function something2(): number;
 //// [/home/src/workspaces/project/src/fileNotFound.js] *new* 
 function something2() { return 20; }
 
-//// [/home/src/workspaces/project/src/main.js] *modified time*
 //// [/home/src/workspaces/project/tsconfig.tsbuildinfo] *modified* 
 {"version":"FakeTSVersion","fileNames":["../../tslibs/TS/Lib/lib.d.ts","./src/filePresent.ts","./src/fileNotFound.ts","./src/anotherFileWithSameReferenes.ts","./src/newFile.ts","./src/main.ts"],"fileInfos":[{"version":"7dee939514de4bde7a51760a39e2b3bfa068bfc4a2939e1dbad2bfdf2dc4662e-/// \u003creference no-default-lib=\"true\"/\u003e\ninterface Boolean {}\ninterface Function {}\ninterface CallableFunction {}\ninterface NewableFunction {}\ninterface IArguments {}\ninterface Number { toExponential: any; }\ninterface Object {}\ninterface RegExp {}\ninterface String { charAt: any; }\ninterface Array\u003cT\u003e { length: number; [n: number]: T; }\ninterface ReadonlyArray\u003cT\u003e {}\ninterface SymbolConstructor {\n    (desc?: string | number): symbol;\n    for(name: string): symbol;\n    readonly toStringTag: symbol;\n}\ndeclare var Symbol: SymbolConstructor;\ninterface Symbol {\n    readonly [Symbol.toStringTag]: string;\n}\ndeclare const console: { log(msg: any): void; };","affectsGlobalScope":true,"impliedNodeFormat":1},{"version":"8f597c315d3bba69a79551601042fdcfe05d35f763762db4908b255c7f17c7d5-function something() { return 10; }","signature":"4f3eeb0c183707d474ecb20d55e49f78d8a3fa3ac388d3b7a318d603ad8478c2-declare function something(): number;\n","affectsGlobalScope":true,"impliedNodeFormat":1},{"version":"6c5b229cbf53b2c6867ab14b139eeac37ed3ec0c1564ba561f7faa869aaba32c-function something2() { return 20; }","signature":"14ba6a62cd6d3e47b343358b2c3e1b7e34b488b489f0d9b915c796cd2e61bbad-declare function something2(): number;\n","affectsGlobalScope":true,"impliedNodeFormat":1},{"version":"552b902790cfbb88a7ed6a7b04b24bb18c58a7f52bcfe7808912e126359e258d-/// \u003creference path=\"./filePresent.ts\"/\u003e\n/// \u003creference path=\"./fileNotFound.ts\"/\u003e\nfunction anotherFileWithSameReferenes() { }","signature":"d4dbe375786b0b36d5425a70f140bbb3d377883027d2fa29aa022a2bd446fbda-declare function anotherFileWithSameReferenes(): void;\n","affectsGlobalScope":true,"impliedNodeFormat":1},{"version":"a4c88e3619994da0f5e4da2dc210f6038e710b9bb831003767da68c882137fb1-function foo() { return 20; }","signature":"f0d67d5e01f8fff5f52028627fc0fb5a78b24df03e482ddac513fa1f873934ee-declare function foo(): number;\n","affectsGlobalScope":true,"impliedNodeFormat":1},{"version":"930f3c7023d87506648db3c352e86f7a2baf38e91f743773fddf4a520aff393a-/// \u003creference path=\"./newFile.ts\"/\u003e\n/// \u003creference path=\"./filePresent.ts\"/\u003e\n/// \u003creference path=\"./fileNotFound.ts\"/\u003e\nfunction main() { }something();something();foo();","signature":"ed4b087ea2a2e4a58647864cf512c7534210bfc2f9d236a2f9ed5245cf7a0896-declare function main(): void;\n","affectsGlobalScope":true,"impliedNodeFormat":1}],"fileIdsList":[[2,3],[2,3,5]],"options":{"composite":true},"referencedMap":[[4,1],[6,2]],"latestChangedDtsFile":"./src/fileNotFound.d.ts"}
 //// [/home/src/workspaces/project/tsconfig.tsbuildinfo.readable.baseline.txt] *modified* 
```

internal/ast/diagnostic.go
sheetalkamat: not really this is additional information where the error is filtered out if "noEmit" is true. thjis ensures that the error is cached in buildInfo without having to make "noEmit" as semantic diagnostic flag. In strada this use to be ` skippedOn?: keyof CompilerOptions;` but because its only noEmit that needs to handle this, i have simplified that here

jakebailey: Right, but is it possible to issue the same error with this bool set in both directions? Since we dedupe diags, they could both be in the same list and then be nondeterministically ordered.

internal/compiler/emitter.go
sheetalkamat: ProgramOrContainer ?

sheetalkamat: ProgramOrIncrementalProgram

sheetalkamat: ProgramLike

internal/incremental/emitfileshandler.go
sheetalkamat: This was left from moving stuff around

internal/tsoptions/enummaps.go
sheetalkamat: Harness usually parsed tsconfig so it has info on which default lib is needed.. also all the libs are bundled and available.i want to fake presence of all libs with the content so that the output is determinitstic and not giant for tsc tests

internal/ast/diagnostic.go
sheetalkamat: I can add, but i dont think we ever report same error in same place with this flag on and off.. so it should not be issue. Its the errors about module etc: eg https://github.com/microsoft/typescript-go/pull/1322/files#diff-8e542d9de5c6b06221168642e2d9dd2df1f2d42d9c79ab60dc28fa6eac85fe39R10081

jakebailey:
The CI is still racing, unfortunately; is there any way we could just not do the concurrency stuff and just make it slower but definitely correct?

sheetalkamat:
i am looking at the race and checking whats the issue. I cant repro it on my windows laptop but looking at the log to see whats going on

internal/incremental/incremental.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"encoding/json"
+
+	"github.com/microsoft/typescript-go/internal/compiler"
+	"github.com/microsoft/typescript-go/internal/outputpaths"
+	"github.com/microsoft/typescript-go/internal/tsoptions"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type BuildInfoReader interface {
```
jakebailey, line 12: There's only one implementation of this; do we need this to be an interface? Seems like it'd be fine as a struct.

internal/incremental/snapshot.go
```diff
@@ -0,0 +1,566 @@
...
+		diagnosticToStringBuilder(chain, file, builder)
+	}
+	for _, info := range diagnostic.RelatedInformation() {
+		diagnosticToStringBuilder(info, file, builder)
+	}
+	return builder.String()
+}
+
+func (s *snapshot) computeHash(text string) string {
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(text)))
```
jakebailey: This is the first time we're introducing `crypto` into the repo; I was hoping to avoid this given it means we now need to think about FIPS and all that stuff. Does this actually need to be a SHA256 sum, or would https://pkg.go.dev/hash/fnv#New128a suffice?

internal/compiler/program.go
```diff
@@ -1350,7 +1485,7 @@ func (p *Program) GetImportHelpersImportSpecifier(path tspath.Path) *ast.Node {
 }
 
 func (p *Program) SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmit bool) bool {
-	return sourceFileMayBeEmitted(sourceFile, &emitHost{program: p}, forceDtsEmit)
+	return sourceFileMayBeEmitted(sourceFile, p, forceDtsEmit)
```
jakebailey, line 1488: I missed one, thanks!

internal/incremental/incremental.go
sheetalkamat: we will need this with tsc --build which will implement this such that it caches the data. so i kept it like that

internal/incremental/snapshot.go
```diff
@@ -0,0 +1,568 @@
...
+	for _, info := range diagnostic.RelatedInformation() {
+		diagnosticToStringBuilder(info, file, builder)
+	}
+	return builder.String()
+}
+
+func (s *snapshot) computeHash(text string) string {
+	hasher := fnv.New128a()
+	hasher.Write([]byte(text))
+	hash := fmt.Sprintf("%x", hasher.Sum(nil))
```
jakebailey:
```suggestion
	hasher := fnv.New128a()
	hasher.Write([]byte(text))
	hash := hex.EncodeToString(hasher.Sum(nil))
```

internal/incremental/snapshot.go
```diff
@@ -0,0 +1,569 @@
...
+		}
+		if file != fileOfDecl {
+			referencedFiles.Add(fileOfDecl.Path())
+		}
+	}
+}
+
+// Get the module source file and all augmenting files from the import name node from file
+func addReferencedFilesFromImportLiteral(file *ast.SourceFile, referencedFiles *collections.Set[tspath.Path], checker *checker.Checker, importName *ast.LiteralLikeNode) {
+	symbol := checker.GetSymbolAtLocation(importName)
```
jakebailey, line 514: Pretty shocked that this is how it used to work; was not aware this function was ever used outside services, and yet the old code did do this.

Is it actually necessary to collect info at this level of granularity? Would it be sufficient to just in the Program remember info from file loading to know file references or something?

sheetalkamat: Since this is a port, i would like to keep this as is and then we can revisit this part for better perf and understanding the scnearios. Its constrained to single function and should be easy to swap out to better logic later

jakebailey: Certainly, yes; I'm at least glad that this doesn't cause races.

internal/incremental/snapshot.go
```diff
@@ -0,0 +1,569 @@
...
+func (s *snapshot) computeSignatureWithDiagnostics(file *ast.SourceFile, text string, data *compiler.WriteFileData) string {
+	var builder strings.Builder
+	builder.WriteString(getTextHandlingSourceMapForSignature(text, data))
+	for _, diag := range data.Diagnostics {
+		diagnosticToStringBuilder(diag, file, &builder)
+	}
+	return s.computeHash(builder.String())
+}
+
+func diagnosticToStringBuilder(diagnostic *ast.Diagnostic, file *ast.SourceFile, builder *strings.Builder) string {
```
jakebailey: Just noticed that this function returns a string, but the return is never used. Probably it could just be a void function.

sheetalkamat:
```suggestion
func diagnosticToStringBuilder(diagnostic *ast.Diagnostic, file *ast.SourceFile, builder *strings.Builder) {
```

jakebailey: Just the edit isn't enough, since the returns are still present

jakebailey:
Neat

image

Compare to tsc

image

- tsc first build, 8.8s
- tsc second build, 1.4s
- tsgo first build, 1.4s
- tsgo second build, 0.17s

So tsgo can quite literally do full incremental build skipping in the time it takes nodejs to start tsc.

jakebailey (approved):

PR #1326, opened 2025-06-30
Folding Ranges implementation

navya9singh:
This PR implements folding ranges in the lsp. tsserver also had hint spans while lsp doesn't so I have combined those tests into folding tests.


internal/ast/utilities.go
```diff
@@ -1265,7 +1308,7 @@ func IsDeclaration(node *Node) bool {
 	if node.Kind == KindTypeParameter {
 		return node.Parent != nil
 	}
-	return IsDeclarationNode(node)
+	return isDeclarationKind(node.Kind)
```
jakebailey: The old code should have been sufficient; when did it fail? JSDoc?

internal/lsp/server.go
```diff
@@ -573,6 +576,13 @@ func (s *Server) handleInitialize(req *lsproto.RequestMessage) {
 				FirstTriggerCharacter: "{",
 				MoreTriggerCharacter:  &[]string{"}", ";", "\n"},
 			},
+			FoldingRangeProvider: &lsproto.BooleanOrFoldingRangeOptionsOrFoldingRangeRegistrationOptions{
+				FoldingRangeOptions: &lsproto.FoldingRangeOptions{
+					WorkDoneProgressOptions: lsproto.WorkDoneProgressOptions{
+						WorkDoneProgress: ptrTo(true),
```
jakebailey: Are you actually using work done progress?

internal/ast/utilities.go
navya9singh: Yes

internal/lsp/server.go
jakebailey: (I don't think so, in which case this could just be `FoldingRangeProvider: &...{Boolean: ptrTo(true)}`)

internal/ast/utilities.go
jakebailey: Hm, I'm guessing we're going to have to redo JSDoc anyway, because I think this change is likely to make us end up crashing? Can we defer JSDoc stuff for now?

internal/lsp/server.go
navya9singh: You're right, I didn't check earlier what WorkDoneProgress meant 👍

internal/ls/folding_test.go
```diff
@@ -1101,12 +1099,6 @@ const[| {
 		const foo = null;
 
 		function Foo()[| {
-		  [|/**
-		    * Description
-		    *
-		    * @param {string} param
-		    * @returns
-		    */|]
```
jakebailey: Does this crash, and that's why it's being removed? Can we leave the comment in and make sure we don't crash?

If we do crash, the editor will pop up an error message.

internal/ast/utilities.go
jakebailey: To be clear, by "defer" I mean "don't try and do folding ranges in JSDoc" (whatever that would mean), but we need to make sure we don't crash on it either, since this is a request that just "happens" as the user is working.

internal/ls/folding_test.go
navya9singh: It doesn't crash, the range is just returned twice because of the `IsDeclarationNode(node)` change. So, in Strada this range would've returned once because we checked for `isDeclarationKind(node.Kind)` instead of  `IsDeclarationNode(node)` returning true here.  The behavior still remains the same in the editor even if it is returned twice, I tested it.

jakebailey: Gotcha; that seems fine, though it seems like `visitNode` could just skip any JSDoc nodes too?

navya9singh: Same reason for the other change in that same commit. 
Btw, why is it that we're testing `IsDeclarationNode(node)` and not `isDeclarationKind(node.Kind)`?

jakebailey: With the new reparsed JS AST, JSDoc nodes are not actually declaration nodes anymore, as we should be inserting into the tree real nodes instead. But, that all is in flux.

navya9singh: I compared the debugging here with Strada, it's not exactly JSDoc, what's happening is that when the node is a `BinaryExpression`, then it returns true for `IsDeclarationNode(node)` which did not use to be the case in Strada for `isDeclarationKind(node.Kind)`. This is causing the duplication of that range.

jakebailey: Oh, bleh, because BinaryExpressions are currently Declarations for JSDoc reasons. Can you easily special case that for now until that is resolved?

internal/ls/utilities.go
```diff
@@ -390,7 +390,7 @@ func isInRightSideOfInternalImportEqualsDeclaration(node *ast.Node) bool {
 }
 
 func (l *LanguageService) createLspRangeFromNode(node *ast.Node, file *ast.SourceFile) *lsproto.Range {
-	return l.createLspRangeFromBounds(node.Pos(), node.End(), file)
+	return l.createLspRangeFromBounds(astnav.GetStartOfNode(node, file, false /*includeJSDoc*/), node.End(), file)
```
jakebailey: This seems weird; do we actually want to do that for _all_ nodes? What happens if we need to make a range out of a JSDoc range itself? (Should your calls be using the other helper below?)

internal/ls/folding.go
```diff
@@ -0,0 +1,543 @@
...
+	statements := sourceFile.Statements
+	// Includes the EOF Token so that comments which aren't attached to statements are included
+	var curr *ast.Node
+	currentTokenEnd := 0
+	if statements != nil && statements.Nodes != nil {
+		curr = statements.Nodes[len(statements.Nodes)-1]
+		currentTokenEnd = curr.End()
+	}
+	scanner := scanner.GetScannerForSourceFile(sourceFile, currentTokenEnd)
+	statements.Nodes = append(statements.Nodes, sourceFile.GetOrCreateToken(scanner.Token(), scanner.TokenFullStart(), scanner.TokenEnd(), curr))
```
DanielRosenwasser: Doesn't this mutate the SourceFile's contents? Did this not cause issues when running the language server for you?

In any case, I don't think allocating an entirely new array was ever necessary. Just call `visitNode` at the end on the EOF. I just sent a change in Strada to do this. https://github.com/microsoft/TypeScript/pull/61987

jakebailey: Yeah, I'm very suspicious of any logic using this outside astnav itself, we have some that I think are wrong already by trying to emulate get children

jakebailey: Oh, well this is just explicitly the append bug

internal/ls/folding_test.go
```diff
@@ -0,0 +1,1445 @@
+package ls_test
```
gabritto: Any reason why those are unit tests and not fourslash tests? I think eventually we want to drop the services unit tests in favor of fourslash tests, so it would be nice to have them either in this PR or in a follow-up. I think all of the existing fourslash basics are in place for implementing `verify.outliningSpansInCurrentFile`, but if there's something missing, let me know.

internal/ls/utilities.go
gabritto: I think this is actually correct. In Strada this is equivalent to `createTextSpanFromNode`, which uses `node.getStart()` as the start of the span.

internal/ls/folding_test.go
navya9singh: Right, I'll send that as a follow-up PR

internal/ls/utilities.go
navya9singh: Yeah, that's why I changed it here.

internal/ls/folding.go
```diff
@@ -0,0 +1,543 @@
...
+	// Includes the EOF Token so that comments which aren't attached to statements are included
+	var curr *ast.Node
+	currentTokenEnd := 0
+	if statements != nil && statements.Nodes != nil {
+		curr = statements.Nodes[len(statements.Nodes)-1]
+		currentTokenEnd = curr.End()
+	}
+	scanner := scanner.GetScannerForSourceFile(sourceFile, currentTokenEnd)
+	foldingRange = append(foldingRange, visitNode(sourceFile.GetOrCreateToken(scanner.Token(), scanner.TokenFullStart(), scanner.TokenEnd(), curr), depthRemaining, sourceFile, l)...)
+	return foldingRange
```
DanielRosenwasser: Thanks to https://github.com/microsoft/typescript-go/pull/1257 from @Andarist, you can now just write:

```suggestion
	// Visit the EOF Token so that comments which aren't attached to statements are included.
	foldingRange = append(foldingRange, visitNode(sourceFile.EndOfFileToken, depthRemaining, sourceFile, l)...)

	return foldingRange
```

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+	sort.Slice(res, func(i, j int) bool {
+		if res[i].StartLine != res[j].StartLine {
+			return res[i].StartLine < res[j].StartLine
+		}
+		if res[i].EndLine != res[j].EndLine {
+			return res[i].EndLine < res[j].EndLine
+		} else if res[i].StartCharacter != nil && res[j].StartCharacter != nil {
+			return *res[i].StartCharacter < *res[j].StartCharacter
+		}
+		return *res[i].EndCharacter < *res[j].EndCharacter
```
DanielRosenwasser: Shouldn't you need to check that `EndCharacter` not also possibly `nil`? And only check this if ~`StartCharacter`~ `EndLine` differs?

DanielRosenwasser: On top of that, the sorting here is off in a different way.

The original code sorts only by span starts. We don't sort by ends/length (and I guess we have no tests that are affected by that - so I don't have a strong opinion about sorting by the `End`s at all).

But the main issue - we don't have span starts here. Instead we have only `StartLine` and `StartCharacter`. Those two are the primary things we need to sort by, *before* anything about the `End`.

We likely don't have any examples that will be affected - but we should get the sorting right here.

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+			foldingRangeKindRegion := lsproto.FoldingRangeKindRegion
+			collapsedTest := "#region"
+			if result.name != "" {
+				collapsedTest = result.name
+			}
+			regions = append(regions, &lsproto.FoldingRange{
+				StartLine:      span.Line,
+				StartCharacter: &span.Character,
+				Kind:           &foldingRangeKindRegion,
+				CollapsedText:  &collapsedTest,
```
DanielRosenwasser:
```suggestion
			collapsedText := "#region"
			if result.name != "" {
				collapsedText = result.name
			}
			regions = append(regions, &lsproto.FoldingRange{
				StartLine:      span.Line,
				StartCharacter: &span.Character,
				Kind:           &foldingRangeKindRegion,
				CollapsedText:  &collapsedText,
```

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+			})
+		} else {
+			if len(regions) > 0 {
+				region := regions[len(regions)-1]
+				regions = regions[:len(regions)-1]
+				if region != nil {
+					if out == nil {
+						out = []*lsproto.FoldingRange{}
+					}
```
DanielRosenwasser: I don't think you need to do this, `append` will just handle this.


```suggestion
```

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+	var out []*lsproto.FoldingRange
+	lineStarts := scanner.GetLineStarts(sourceFile)
+	for _, currentLineStart := range lineStarts {
+		lineEnd := scanner.GetLineEndOfPosition(sourceFile, int(currentLineStart))
+		lineText := sourceFile.Text()[currentLineStart:lineEnd]
+		result := parseRegionDelimiter(lineText)
+		if result == nil || isInComment(sourceFile, int(currentLineStart), nil) != nil {
+			continue
+		}
+		if result.isStart {
```
DanielRosenwasser, line 81:
```suggestion

		if result.isStart {
```

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+				StartLine:      span.Line,
+				StartCharacter: &span.Character,
+				Kind:           &foldingRangeKindRegion,
+				CollapsedText:  &collapsedTest,
+			})
+		} else {
+			if len(regions) > 0 {
+				region := regions[len(regions)-1]
+				regions = regions[:len(regions)-1]
+				if region != nil {
```
DanielRosenwasser: You don't need the `nil` check here, the code would have crashed if you indexed out-of-bounds but you had the `len(regions) > 0` check above.

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+	lineStarts := scanner.GetLineStarts(sourceFile)
+	for _, currentLineStart := range lineStarts {
+		lineEnd := scanner.GetLineEndOfPosition(sourceFile, int(currentLineStart))
+		lineText := sourceFile.Text()[currentLineStart:lineEnd]
+		result := parseRegionDelimiter(lineText)
+		if result == nil || isInComment(sourceFile, int(currentLineStart), nil) != nil {
+			continue
+		}
+		if result.isStart {
+			span := l.createLspPosition(strings.Index(sourceFile.Text()[currentLineStart:lineEnd], "//")+int(currentLineStart), sourceFile)
```
DanielRosenwasser: This isn't a `span` though, this is a position. I would call it `commentStart` or something like that

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+			foldingRangeKindRegion := lsproto.FoldingRangeKindRegion
+			collapsedTest := "#region"
+			if result.name != "" {
+				collapsedTest = result.name
+			}
+			regions = append(regions, &lsproto.FoldingRange{
```
DanielRosenwasser, line 91: Okay, there is a subtlety both here and in the original code about how the end position isn't filled in yet.


```suggestion
			// Our spans start out with some initial data.
			// On every `#endregion`, we'll come back to these `FoldingRange`s
			// and fill in their EndLine/EndCharacter.
			regions = append(regions, &lsproto.FoldingRange{
```

DanielRosenwasser (changes requested):

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/printer"
+	"github.com/microsoft/typescript-go/internal/scanner"
+)
+
+func (l *LanguageService) ProvideFoldingRange(ctx context.Context, documentURI lsproto.DocumentUri) []*lsproto.FoldingRange {
+	_, sourceFile := l.getProgramAndFile(documentURI)
+	res := l.addNodeOutliningSpans(sourceFile)
+	res = append(res, l.addRegionOutliningSpans(sourceFile)...)
+	sort.Slice(res, func(i, j int) bool {
```
jakebailey: This should be `slices.SortFunc`; `sort.Slice` is old-school

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+
+type regionDelimiterResult struct {
+	isStart bool
+	name    string
+}
+
+func parseRegionDelimiter(lineText string) *regionDelimiterResult {
+	// We trim the leading whitespace and // without the regex since the
+	// multiple potential whitespace matches can make for some gnarly backtracking behavior
+	lineText = strings.TrimLeft(lineText, " \t")
```
jakebailey: The original code said `lineText = lineText.trimStart()`, so this code should say `strings.TrimLeftFunc(lineText, unicode.IsSpace)`.

internal/ls/folding.go
```diff
@@ -0,0 +1,536 @@
...
+	if !strings.HasPrefix(lineText, "//") {
+		return nil
+	}
+	lineText = strings.TrimLeft(lineText[2:], " \t")
```
jakebailey: The original code trimmed both sides, so should be `strings.TrimSpace(lineText[2:])`

internal/ls/utilities.go
jakebailey: If that's how it was, then that seems fine, and maybe we can close the other PR?

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+		if a.StartCharacter != nil && b.StartCharacter != nil {
+			return cmp.Compare(*a.StartCharacter, *b.StartCharacter)
+		}
+		if a.EndLine != b.EndLine {
+			return cmp.Compare(a.EndLine, b.EndLine)
+		}
+		if a.EndCharacter != nil && b.EndCharacter != nil {
+			return cmp.Compare(*a.EndCharacter, *b.EndCharacter)
+		}
+		return 0
```
DanielRosenwasser: If we're sorting by the ends at all, you can't immediately return `cmp.Compare(...)` on `StartCharacter` - you have to check that they're actually different like you do in every other branch.

One way you could do this:

```suggestion
		if a.StartLine != b.StartLine {
			return cmp.Compare(a.StartLine, b.StartLine)
		}
		if a.StartCharacter != nil && b.StartCharacter != nil && *a.StartCharacter != *b.StartCharacter {
			return cmp.Compare(*a.StartCharacter, *b.StartCharacter)
		}
		if a.EndLine != b.EndLine {
			return cmp.Compare(a.EndLine, b.EndLine)
		}
		if a.EndCharacter != nil && b.EndCharacter != nil && *a.EndCharacter != *b.EndCharacter {
			return cmp.Compare(*a.EndCharacter , *b.EndCharacter)
		}
		return 0
```

Another way you could do this:

```suggestion
		if r := cmp.Compare(a.StartLine, b.StartLine); r != 0 {
			return r
		}
		if a.StartCharacter != nil && b.StartCharacter != nil {
			if r := cmp.Compare(*a.StartCharacter, *b.StartCharacter); r != 0 {
				return r
			}
		}
		if r := cmp.Compare(a.EndLine, b.EndLine); r != 0 {
			return r
		}
		if a.EndCharacter != nil && b.EndCharacter != nil {
			if r := cmp.Compare(*a.EndCharacter, *b.EndCharacter); r != 0 {
				return r
			}
		}
		return 0
```

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+	return res
+}
+
+func (l *LanguageService) addNodeOutliningSpans(sourceFile *ast.SourceFile) []*lsproto.FoldingRange {
+	depthRemaining := 40
+	current := 0
+
+	statements := sourceFile.Statements
+	n := len(statements.Nodes)
+	var foldingRange []*lsproto.FoldingRange
```
DanielRosenwasser: I actually think starting out with a nil slice is probably not ideal. Folding ranges are never cached on our side, so maybe it's best to just start out with some reasonable initial capacity (I dunno, I'm just going to throw out a nice round number like 40?).

This isn't a blocker for this PR.

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+	// We trim the leading whitespace and // without the regex since the
+	// multiple potential whitespace matches can make for some gnarly backtracking behavior
+	lineText = strings.TrimLeftFunc(lineText, unicode.IsSpace)
+	if !strings.HasPrefix(lineText, "//") {
+		return nil
+	}
+	lineText = strings.TrimSpace(lineText[2:])
+	result := regionDelimiterRegExp.FindStringSubmatch(lineText)
+	if result != nil {
+		return &regionDelimiterResult{
```
DanielRosenwasser: @jakebailey it's minor, but any idea if the allocation can be eliminated here? I might just make `parseRegionDelimiter` return `result, ok`, but I get wanting to preserve the original code.

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+	case ast.KindArrayLiteralExpression:
+		return spanForNode(n, ast.KindOpenBracketToken, !ast.IsArrayLiteralExpression(n.Parent) && !ast.IsCallExpression(n.Parent) /*useFullStart */, sourceFile, l)
+	case ast.KindJsxElement, ast.KindJsxFragment:
+		return spanForJSXElement(n, sourceFile, l)
+	case ast.KindJsxSelfClosingElement, ast.KindJsxOpeningElement:
+		return spanForJSXAttributes(n, sourceFile, l)
+	case ast.KindTemplateExpression, ast.KindNoSubstitutionTemplateLiteral:
+		return spanForTemplateLiteral(n, sourceFile, l)
+	case ast.KindArrayBindingPattern:
+		return spanForNode(n, ast.KindOpenBracketToken, !ast.IsBindingElement(n.Parent) /*useFullStart */, sourceFile, l)
```
DanielRosenwasser: A lot of these `useFullStart` comments have a trailing space on the end. Can you clean that up?

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+	combineAndAddMultipleSingleLineComments := func() *lsproto.FoldingRange {
+		// Only outline spans of two or more consecutive single line comments
+		if singleLineCommentCount > 1 {
+			return createFoldingRangeFromBounds(firstSingleLineCommentStart, lastSingleLineCommentEnd, foldingRangeKindComment, sourceFile, l)
+		}
+		return nil
+	}
+
+	sourceText := sourceFile.Text()
+	comments(func(comment ast.CommentRange) bool {
```
DanielRosenwasser: Not that you can't, but why was this written this way instead of `for comment := range GetLeadingCommentRanges`?

internal/ls/folding.go
```diff
@@ -0,0 +1,541 @@
...
+		commentPos := comment.Pos()
+		commentEnd := comment.End()
+		// cancellationToken.throwIfCancellationRequested();
```
DanielRosenwasser: Should these be rewritten in terms of `ctx.Err()`? I see other places where this has been written, so I'm okay with doing that as a separate PR.

DanielRosenwasser (changes requested):

internal/ls/folding.go
navya9singh: Yes, it's used in a lot of places so I fix that in a separate PR.
 Is that supposed to replace `cancellationToken`?

DanielRosenwasser: I *think* so.

internal/ls/format.go
```diff
@@ -137,6 +137,9 @@ func getRangeOfEnclosingComment(
 	precedingToken *ast.Node,
 	tokenAtPosition *ast.Node,
 ) *ast.CommentRange {
+	if tokenAtPosition == nil {
+		tokenAtPosition = astnav.GetTokenAtPosition(file, position)
+	}
```
DanielRosenwasser: Pretty sure @gabritto omitted this intentionally in a recent PR (might be mixing details up).

The comment above says 

```
// Unlike the TS implementation, this function *will not* compute default values for
// `precedingToken` and `tokenAtPosition`.
```

so I think you should explicitly pass in a `tokenAtPosition`.

DanielRosenwasser (changes requested):

internal/ls/format.go
gabritto: Ah yes, thanks for catching that, I didn't realize this PR was using this function.

DanielRosenwasser:
If you merge from the `foldingRangesUpdates` branch, that should have adjustments based off of `main`, and should get you in sync.

The feedback I have in this PR still applies though. Also, I tried the branch out locally and wrote the following (notice the leading newline)

```ts

let x = {

}
```

and hit the following crash:

```
panic handling request textDocument/foldingRange runtime error: index out of range [-1] goroutine 189 [running]:
runtime/debug.Stack()
	/home/daniel/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25rc1.linux-arm64/src/runtime/debug/stack.go:26 +0x74
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1.1()
	/home/daniel/typescript-go/internal/lsp/server.go:357 +0x58
panic({0x8fb700?, 0x4004e11008?})
	/home/daniel/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25rc1.linux-arm64/src/runtime/panic.go:783 +0xec
github.com/microsoft/typescript-go/internal/scanner.GetLineEndOfPosition({0xaf8590, 0x4000532008}, 0x0)
	/home/daniel/typescript-go/internal/scanner/scanner.go:2379 +0x1d4
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).addRegionOutliningSpans(0x40064fb1a0, 0x4000532008)
	/home/daniel/typescript-go/internal/ls/folding.go:83 +0x104
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideFoldingRange(0x40064fb1a0, {0xafa3f0, 0x40001552c0}, {0x4002939a40, 0x13})
	/home/daniel/typescript-go/internal/ls/folding.go:21 +0x90
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleFoldingRange(0x40001b8a20, {0xafa3f0, 0x40001552c0}, 0x4002448ed0)
	/home/daniel/typescript-go/internal/lsp/server.go:697 +0x10c
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x40001b8a20, {0xafa3f0, 0x40001552c0}, 0x4002448ed0)
	/home/daniel/typescript-go/internal/lsp/server.go:504 +0x7d8
	/home/daniel/typescript-go/internal/lsp/server.go:370 +0xa8
created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 34
	/home/daniel/typescript-go/internal/lsp/server.go:388 +0x414
```

In fact, all you need is a blank file with a newline at the beginning.

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -274,14 +277,26 @@ function parseEditStatement(funcName: string, args: readonly ts.Expression[]): E
...
+                        kind: "edit",
+                        goStatement: `f.Backspace(t, ${arg0.text})`,
+                    };
+                }
+                // Handle test.marker("").position
+                if (ts.isPropertyAccessExpression(arg) && arg.name.text === "position") {
+                    const expr = arg.expression;
+                    if (ts.isCallExpression(expr) && expr.expression.getText() === "test.marker") {
+                        if (expr.arguments.length === 1 && ts.isStringLiteralLike(expr.arguments[0])) {
+                            return {
```
gabritto: That looks like it should produce something like `f.Backspace(t, f.MarkerByName(...).Position)`.

gabritto: If there's only a few instances of tests with non-numeric arguments to `edit.backspace`, it might be better to just manually port them instead of handling every possible variation here.

internal/fourslash/_scripts/manualTests.txt
```diff
@@ -13,3 +13,4 @@ quickInfoForOverloadOnConst1
 renameDefaultKeyword
 renameForDefaultExport01
 tsxCompletion12
+OutliningHintSpansForFunction
```
gabritto: I suspect using a non-canonical casing for the test name here might cause problems since Linux is case sensitive.
```suggestion
outliningHintSpansForFunction
```

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+
+func (l *LanguageService) ProvideFoldingRange(ctx context.Context, documentURI lsproto.DocumentUri) (lsproto.FoldingRangeResponse, error) {
+	_, sourceFile := l.getProgramAndFile(documentURI)
+	res := l.addNodeOutliningSpans(sourceFile)
+	res = append(res, l.addRegionOutliningSpans(sourceFile)...)
+	slices.SortFunc(res, func(a, b *lsproto.FoldingRange) int {
+		if a.StartLine != b.StartLine {
+			return cmp.Compare(a.StartLine, b.StartLine)
+		}
+		if a.StartCharacter != nil && b.StartCharacter != nil && *a.StartCharacter != *b.StartCharacter {
```
gabritto: A couple questions:
- Is the sorting order important for correctness?
- Do we ever leave `StartCharacter` as `nil`? If we don't, we should probably just assert so and unconditionally compare the start characters. If we do, for completeness, we need to consider the case where one start character is `nil` and the other isn't. I see the spec says that if this is `nil`, then `StartCharacter` is taken to be the length of the start line, so we'd need to consider that when sorting.

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+		}
+	}
+
+	// Visit the EOF Token so that comments which aren't attached to statements are included.
+	foldingRange = append(foldingRange, visitNode(sourceFile.EndOfFileToken, depthRemaining, sourceFile, l)...)
+	return foldingRange
+}
+
+func (l *LanguageService) addRegionOutliningSpans(sourceFile *ast.SourceFile) []*lsproto.FoldingRange {
+	regions := make([]*lsproto.FoldingRange, 0, 40)
```
gabritto, line 70: Out of curiosity, where does the 40 initial capacity come from?

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+		}
+	}
+	return out
+}
+
+func visitNode(n *ast.Node, depthRemaining int, sourceFile *ast.SourceFile, l *LanguageService) []*lsproto.FoldingRange {
+	if depthRemaining == 0 {
+		return nil
+	}
+	// cancellationToken.throwIfCancellationRequested();
```
gabritto: We're already able to do cancellation in Corsa. This should be something like `if ctx.Err() != nil { return nil }`.

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+	foldingRange := make([]*lsproto.FoldingRange, 0, 40)
+	// !!! remove !ast.IsBinaryExpression(n) after JSDoc implementation
```
gabritto: Outdated comment?

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+		case ast.KindMultiLineCommentTrivia:
+			comments := combineAndAddMultipleSingleLineComments()
+			if comments != nil {
+				foldingRange = append(foldingRange, comments)
+			}
+			foldingRange = append(foldingRange, createFoldingRangeFromBounds(commentPos, commentEnd, foldingRangeKindComment, sourceFile, l))
+			singleLineCommentCount = 0
+			break
+		default:
+			// Debug.assertNever(kind);
```
gabritto: Same as for cancellation, we have `Debug.AssertNever` now.

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+func addOutliningForLeadingCommentsForNode(n *ast.Node, sourceFile *ast.SourceFile, l *LanguageService) []*lsproto.FoldingRange {
+	if ast.IsJsxText(n) {
+		return nil
+	}
+	return addOutliningForLeadingCommentsForPos(n.Pos(), sourceFile, l)
+}
+
+func addOutliningForLeadingCommentsForPos(pos int, sourceFile *ast.SourceFile, l *LanguageService) []*lsproto.FoldingRange {
+	p := &printer.EmitContext{}
+	foldingRange := make([]*lsproto.FoldingRange, 0, 40)
```
gabritto, line 215: Do we expect to typically see around 40 ranges for a comment position?

internal/ls/folding.go
```diff
@@ -0,0 +1,540 @@
...
+		}
+	}
+	addedComments := combineAndAddMultipleSingleLineComments()
+	if addedComments != nil {
+		foldingRange = append(foldingRange, addedComments)
+	}
+	return foldingRange
+}
+
+var regionDelimiterRegExp = regexp.MustCompile(`^#(end)?region(.*)\r?$`)
```
gabritto: IIRC we have been avoiding the use of RegExp in Corsa since they're slow in Go (@jakebailey can confirm). Could this be a function instead?

internal/fourslash/fourslash.go
```diff
@@ -1444,6 +1444,55 @@ func (f *FourslashTest) VerifyBaselineWorkspaceSymbol(t *testing.T, query string
...
+		},
+	}
+	result := sendRequest(t, f, lsproto.TextDocumentFoldingRangeInfo, params)
+	if result.FoldingRanges == nil {
+		t.Fatalf("Nil response received for folding range request")
+	}
+
+	// Extract actual folding ranges from the result and filter by kind if specified
+	var actualRanges []*lsproto.FoldingRange
+	if result.FoldingRanges != nil {
```
gabritto: Nit: this was already checked a few lines above.

internal/fourslash/fourslash.go
```diff
@@ -1444,6 +1444,55 @@ func (f *FourslashTest) VerifyBaselineWorkspaceSymbol(t *testing.T, query string
...
+	})
+
+	for i, expectedRange := range f.Ranges() {
+		actualRange := actualRanges[i]
+		startPos := lsproto.Position{Line: actualRange.StartLine, Character: *actualRange.StartCharacter}
+		endPos := lsproto.Position{Line: actualRange.EndLine, Character: *actualRange.EndCharacter}
+
+		if lsproto.ComparePositions(startPos, expectedRange.LSRange.Start) != 0 ||
+			lsproto.ComparePositions(endPos, expectedRange.LSRange.End) != 0 {
+			t.Fatalf("verifyOutliningSpans failed - span %d has invalid positions: start (%d,%d), end (%d,%d)",
```
gabritto: Would it be helpful to also print the expected range here?

internal/ls/folding.go
jakebailey: Yes, this would be a good one to change, given this will run in every line. This is just a prefix check if anything

internal/ls/folding.go
jakebailey: We always set StartCharacter.

Folding ranges don't overlap so this can actually just be:

```go
if c := cmp.Compare(a.StartLine, b.StartLine); c != 0 {
    return c
}
return cmp.Compare(*a.StartCharacter, *b.StartCharacter)
```

internal/ls/folding.go
```diff
@@ -0,0 +1,552 @@
...
+				region.EndLine = endingPosition.Line
+				region.EndCharacter = &endingPosition.Character
+				out = append(out, region)
+			}
+		}
+	}
+	return out
+}
+
+func visitNode(ctx context.Context, n *ast.Node, depthRemaining int, sourceFile *ast.SourceFile, l *LanguageService) []*lsproto.FoldingRange {
```
jakebailey, line 111: I think thins might turn out better if all of this code were instead methods in on a struct, like `foldingRangeCollector`, then you don't need to pass everything down so much, the deth can just be there, you can append into a slice there, etc

internal/fourslash/tests/gen/correuptedTryExpressionsDontCrashGettingOutlineSpans_test.go
```diff
@@ -0,0 +1,23 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestCorreuptedTryExpressionsDontCrashGettingOutlineSpans(t *testing.T) {
```
jakebailey, line 10: correupted 😄

love us

internal/ls/folding.go
jakebailey: 40 is the max depth that strada used, but I don't think that we actually need to prealloc these lists.

navya9singh:
> If you merge from the `foldingRangesUpdates` branch, that should have adjustments based off of `main`, and should get you in sync.
> 
> The feedback I have in this PR still applies though. Also, I tried the branch out locally and wrote the following (notice the leading newline)
> 
> ```ts
> let x = {
> 
> }
> ```
> 
> and hit the following crash:
> 
> ```
> panic handling request textDocument/foldingRange runtime error: index out of range [-1] goroutine 189 [running]:
> runtime/debug.Stack()
> 	/home/daniel/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25rc1.linux-arm64/src/runtime/debug/stack.go:26 +0x74
> github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1.1()
> 	/home/daniel/typescript-go/internal/lsp/server.go:357 +0x58
> panic({0x8fb700?, 0x4004e11008?})
> 	/home/daniel/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25rc1.linux-arm64/src/runtime/panic.go:783 +0xec
> github.com/microsoft/typescript-go/internal/scanner.GetLineEndOfPosition({0xaf8590, 0x4000532008}, 0x0)
> 	/home/daniel/typescript-go/internal/scanner/scanner.go:2379 +0x1d4
> github.com/microsoft/typescript-go/internal/ls.(*LanguageService).addRegionOutliningSpans(0x40064fb1a0, 0x4000532008)
> 	/home/daniel/typescript-go/internal/ls/folding.go:83 +0x104
> github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideFoldingRange(0x40064fb1a0, {0xafa3f0, 0x40001552c0}, {0x4002939a40, 0x13})
> 	/home/daniel/typescript-go/internal/ls/folding.go:21 +0x90
> github.com/microsoft/typescript-go/internal/lsp.(*Server).handleFoldingRange(0x40001b8a20, {0xafa3f0, 0x40001552c0}, 0x4002448ed0)
> 	/home/daniel/typescript-go/internal/lsp/server.go:697 +0x10c
> github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x40001b8a20, {0xafa3f0, 0x40001552c0}, 0x4002448ed0)
> 	/home/daniel/typescript-go/internal/lsp/server.go:504 +0x7d8
> github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1()
> 	/home/daniel/typescript-go/internal/lsp/server.go:370 +0xa8
> created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 34
> 	/home/daniel/typescript-go/internal/lsp/server.go:388 +0x414
> ```
> 
> In fact, all you need is a blank file with a newline at the beginning.

This doesn't panic anymore.

jakebailey:
This unfortunately crashed on me when I opened into `checker.ts`:

```

[Error - 10:32:43 AM] Request textDocument/foldingRange failed.
  Message: InternalError: panic handling request textDocument/foldingRange: Token cache mismatch: parent. Expected parent of kind KindTryStatement, got KindFinallyKeyword
  Code: -32603 
panic handling request textDocument/foldingRange Token cache mismatch: parent. Expected parent of kind KindTryStatement, got KindFinallyKeyword goroutine 978 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:26 +0x8e
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0xc0001c2008, 0xc0108f2060)
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:848 +0x53
panic({0xfad5a0?, 0xc011e3c7b0?})
	/usr/local/go/src/runtime/panic.go:783 +0x136
github.com/microsoft/typescript-go/internal/ast.(*SourceFile).GetOrCreateToken(0xc0074e2608, 0x61, 0x10b3d, 0x10b52, 0xc00dafea80)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:10957 +0x505
github.com/microsoft/typescript-go/internal/ls.findChildOfKind(0xc00dafea80, 0x12, 0xc0074e2608)
	/home/jabaile/work/TypeScript-go/internal/ls/utilities.go:217 +0x297
github.com/microsoft/typescript-go/internal/ls.spanForNode(0xc00dafea80, 0x12, 0x1, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:478 +0x65
github.com/microsoft/typescript-go/internal/ls.getOutliningSpanForNode(0xc0011156c8, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:330 +0x5e5
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc0011156c8, 0x1e, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:151 +0xd4b
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc0011156c8)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00ce95840, 0xc0011156c8)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*TryStatement).ForEachChild(0xc0009ae7e0, 0xc00ce95840)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3484 +0x9c
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc0009ae7e0, 0xc00ce95840)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc0009ae7e0, 0x1f, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc0009ae7e0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visitNodes(0xc00ce94c40, {0xc0015ad7d8, 0x4, 0x4})
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:28 +0x78
github.com/microsoft/typescript-go/internal/ast.visitNodeList(0xc00ce94c40, 0xc0017c9b28)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:37 +0x51
github.com/microsoft/typescript-go/internal/ast.(*Block).ForEachChild(0xc001115728, 0xc00ce94c40)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3679 +0x32
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc001115728, 0xc00ce94c40)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc001115728, 0x20, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc001115728)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00ce94980, 0xc001115728)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*ArrowFunction).ForEachChild(0xc002e16960, 0xc00ce94980)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:6461 +0x174
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc002e16960, 0xc00ce94980)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc002e16960, 0x21, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc002e16960)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00ce948c0, 0xc002e16960)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*PropertyAssignment).ForEachChild(0xc002e030b0, 0xc00ce948c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:7406 +0x106
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc002e030b0, 0xc00ce948c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc002e030b0, 0x22, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc002e030b0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visitNodes(0xc00c3873c0, {0xc002e3e908, 0xae, 0xae})
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:28 +0x78
github.com/microsoft/typescript-go/internal/ast.visitNodeList(0xc00c3873c0, 0xc0017c9be8)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:37 +0x51
github.com/microsoft/typescript-go/internal/ast.(*ObjectLiteralExpression).ForEachChild(0xc000d485a0, 0xc00c3873c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:7308 +0x32
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc000d485a0, 0xc00c3873c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc000d485a0, 0x23, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc000d485a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00c387200, 0xc000d485a0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*VariableDeclaration).ForEachChild(0xc0026fab68, 0xc00c387200)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3776 +0xd1
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc0026fab68, 0xc00c387200)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc0026fab68, 0x24, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc0026fab68)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visitNodes(0xc00c387180, {0xc002e3ee78, 0x1, 0x1})
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:28 +0x78
github.com/microsoft/typescript-go/internal/ast.visitNodeList(0xc00c387180, 0xc0017c9c08)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:37 +0x51
github.com/microsoft/typescript-go/internal/ast.(*VariableDeclarationList).ForEachChild(0xc00204e108, 0xc00c387180)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3826 +0x32
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc00204e108, 0xc00c387180)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc00204e108, 0x25, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc00204e108)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00c387140, 0xc00204e108)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*VariableStatement).ForEachChild(0xc002079118, 0xc00c387140)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3722 +0x5e
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc002079118, 0xc00c387140)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc002079118, 0x26, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc002079118)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visitNodes(0xc00c0192c0, {0xc00b4c6008, 0x921, 0x921})
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:28 +0x78
github.com/microsoft/typescript-go/internal/ast.visitNodeList(0xc00c0192c0, 0xc005678608)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:37 +0x51
github.com/microsoft/typescript-go/internal/ast.(*Block).ForEachChild(0xc00b1e1f08, 0xc00c0192c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:3679 +0x32
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc00b1e1f08, 0xc00c0192c0)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc00b1e1f08, 0x27, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.visitNode.func1(0xc00b1e1f08)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:194 +0x8f
github.com/microsoft/typescript-go/internal/ast.visit(0xc00c018f80, 0xc00b1e1f08)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:21 +0x32
github.com/microsoft/typescript-go/internal/ast.(*FunctionDeclaration).ForEachChild(0xc00a0c4e50, 0xc00c018f80)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:4104 +0x1b4
github.com/microsoft/typescript-go/internal/ast.(*Node).ForEachChild(0xc00a0c4e50, 0xc00c018f80)
	/home/jabaile/work/TypeScript-go/internal/ast/ast.go:242 +0x2f
github.com/microsoft/typescript-go/internal/ls.visitNode({0x14144f8, 0xc00cffd140}, 0xc00a0c4e50, 0x28, 0xc0074e2608, 0xc00cffd1a0)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:200 +0x178f
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).addNodeOutliningSpans(0xc00cffd1a0, {0x14144f8, 0xc00cffd140}, 0xc0074e2608)
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:40 +0x34e
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideFoldingRange(0xc00cffd1a0, {0x14144f8, 0xc00cffd140}, {0xc00892e100, 0x3c})
	/home/jabaile/work/TypeScript-go/internal/ls/folding.go:20 +0xe5
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleFoldingRange(0xc0001c2008, {0x14144f8, 0xc00cffd140}, 0xc00cffd1a0, 0xc000454020)
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:1144 +0x96
github.com/microsoft/typescript-go/internal/lsp.registerLanguageServiceDocumentRequestHandler[...].func1({0x14144f8, 0xc00cffd140}, 0xc0108f2060)
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:597 +0x28e
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0xc0001c2008, {0x14144f8, 0xc00cffd140}, 0xc0108f2060)
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:478 +0xf3
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:381 +0x65
	/home/jabaile/work/TypeScript-go/internal/lsp/server.go:401 +0x5c5
```

jakebailey:
Since I pushed, I can't review anymore. Can someone else look?

mthines:
On version `7.0.0-dev.20251203.1` I still get the `Request textDocument/foldingRange failed.`  error when opening one of my files in VSCode

I can provide you file, but not in a public thread :)

jakebailey:
Please file an issue with at least the panic stack trace.

PR #1327, opened 2025-06-30
Implement the Go to Symbol in Workspace (Ctrl-T) command

ahejlsberg:
This PR implements the LSP "workspace/symbol" method, which surfaces as the Go to Symbol in Workspace (Ctrl-T) command in VS Code. Requests are handled by returning the top 256 symbols across all active projects that match the given query string. A symbol name matches a query string when it contains each of the characters in the query string in the same order. Upper case characters in the query must be matched exactly, whereas lower case characters can be matched in either case. Matches are ranked by how few unmatched characters remain in the symbol name (so exact matches always sort first).

### DIFF
--- a/internal/ls/symbols.go
+++ b/internal/ls/symbols.go
@@ -1,0 +1,152 @@
+package ls
+
+import (
+	"context"
+	"slices"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/microsoft/typescript-go/internal/ast"
+	"github.com/microsoft/typescript-go/internal/collections"
+	"github.com/microsoft/typescript-go/internal/compiler"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/scanner"
+)
+
+type DeclarationInfo struct {
+	name        string
+	declaration *ast.Node
+	matchScore  int
+}
+
+func ProvideWorkspaceSymbols(ctx context.Context, programs []*compiler.Program, converters *Converters, query string) ([]lsproto.SymbolInformation, error) {
+	// Obtain set of non-declaration source files from all active programs.
+	var sourceFiles collections.Set[*ast.SourceFile]
+	for _, program := range programs {
+		for _, sourceFile := range program.SourceFiles() {
+			if !sourceFile.IsDeclarationFile {
+				sourceFiles.Add(sourceFile)
+			}
+		}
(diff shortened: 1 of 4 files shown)


internal/ls/symbols.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	s1 := ast.GetSourceFileOfNode(d1.declaration)
+	s2 := ast.GetSourceFileOfNode(d2.declaration)
+	if s1 != s2 {
+		return strings.Compare(string(s1.Path()), string(s2.Path()))
+	}
+	return d1.declaration.Pos() - d2.declaration.Pos()
+}
+
+func compareStringsCaseInsensitive(s1, s2 string) int {
```
jakebailey: This function should exist already as `stringutil.CompareStringsCaseInsensitive`.

ahejlsberg: Ah, I missed that one. But we should go with my implementation which is allocation free.

jakebailey: Yeah, feel free to overwrite that other one with your better version.

jakebailey:
Though, I don't know why `__sortedArrayBrand` is even showing up as an option; it's a property? (I can't recall if these were skipped in Strada.)

ahejlsberg:
It's a property named `" __sortedArrayBrand"` (notice the blank). I was thinking about eliminating symbols that start with a blank, but the old language service includes them. 🤷

jakebailey:
Yeah, I just didn't realize this feature could be used to navigate to properties in Strada; my mental model of the feature is that it really only is for finding declarations. Which, I know it's a property declaration, but I mainly mean "actual code declarations" that would exist at runtime.

ahejlsberg:
It pretty much includes any kind of declaration (notably excepting parameters). Strada even included expando property "declarations", which are really just assignments of the form `obj.foo = xxx`. I'm currently excluding those because we need binding information to determine if they actually resulted in a declared symbol. Strada just always included them, which is wrong. Anyway, something to discuss when we have a chance.

jakebailey:
Likely we can just force a bind on the file before doing this then have at the symbols; I think it'd be a lot simpler here than Strada.

But the syntax-only version is great to start.

jakebailey:
Also, document symbols is likely very easy to do now with what's in this PR?

ahejlsberg:
Yeah, document symbols should be fairly easy, though they are returned a hierarchy and not a flat list. BTW, regarding the speed of the requests, it's even faster the second time around because we cache the declaration maps in the `SourceFile` objects.

PR #1330, opened 2025-07-01
Fix out of bounds error in `recreateBindingPattern`

AlCalzone:
If the length of the array is 1, the first item is at index 0, not index 1.


jakebailey (approved):
Seems obviously correct, but how did you find this? Is there a test case?

AlCalzone:
I tried fixing the other issue I reported and caused a panic on a lot of tests.

Andarist:
> Is there a test case?

Yes, but one of them was actually hit. I'm fixing this in https://github.com/microsoft/typescript-go/pull/1336

PR #1336, opened 2025-07-01
Fixed top-level binding element declaration emit

Andarist:


jakebailey:
Is this for a specific issue by chance? Or just a general improvement?

Andarist:
https://github.com/microsoft/typescript-go/pull/1330 prompted me to investigate this. I dont know if there is any reported issue related to this

internal/transformers/declarations/transform.go
```diff
@@ -673,9 +673,6 @@ func (tx *DeclarationTransformer) recreateBindingPattern(input *ast.BindingPatte
 }
 
 func (tx *DeclarationTransformer) recreateBindingElement(e *ast.BindingElement) *ast.Node {
-	if e.Kind == ast.KindBindingElement {
-		return nil
-	}
```
jakebailey, line 678: Should this have been OmittedExpression, based on what's in `recreateBindingElement` in Strada?

jakebailey: Confusingly, it doesn't seem to do anything when added back...

jakebailey: Ah, since `Name` will return nil for that kind.

jakebailey (approved):

internal/transformers/declarations/transform.go
Andarist: ye, that's the Corsa's equivalent of `OmittedExpression` since `OmttedExpression` is no longer a thing

jakebailey: `OmittedExpression` is a thing, though; we use it in many places.

Andarist: From `CHANGES.md`:
> 4. Empty binding elements no longer have a separate kind for OmittedExpression. Instead they have Kind=BindingElement with a nil Initialiser, Name and DotDotDotToken.

I have not double-checked if this still holds true - but in many places in the code where `OmittedExpression` is referenced, there are mentions of those potentially having to be adjusted because it's not a thing anymore.

jakebailey: Yeah, I'm not sure what the status is either. We definitely have two places we create omitted expressions, but they seem to be for arrays with missing elements.

jakebailey: (I am not convinced CHANGES is fully accurate anymore but, it may be.)

jakebailey: But, reading `convertBindingElementToArrayAssignmentElement` it's clear that the binding case should be a nil check, so this PR is okay.

weswigham: Yeah, the parser doesn't make `OmittedExpression`s anymore, but a bunch of code was ported to still reference them. 🤷‍♂️

jakebailey: > Yeah, the parser doesn't make `OmittedExpression`s anymore, but a bunch of code was ported to still reference them. 🤷‍♂️

It does, at least for `[1, 2, , , , ]`, I think

PR #1354, opened 2025-07-03
Clean up host related code

jakebailey:
This removes the need to have options as part of the CompilerHost entirely, and is a cleanup we were thinking about ages ago.

The only notable change is that we now just use `\n` for the output newline on Windows, which I think is just fine.

Also clean up some other unused code in the host.


jakebailey:
Sorry, I missed one more removal, which I caught only as I was doing a final pass in the diff.

andrewbranch (approved):

PR #1355, opened 2025-07-03
fix(1342): adjust node ranges to ignore trivia

### LINKED ISSUE #1342: Find-all-references includes all trivia
```ts
export interface A {
    /** Comment */
    method(): string;
    /** Comment */
    method(format: string): string;
}
```

Image

a-tarasyuk:
Fixes #1342

---

https://github.com/microsoft/TypeScript/blob/833a8d492c728d606454865e8c0fee84842f9f10/src/services/findAllReferences.ts#L745

https://github.com/microsoft/TypeScript/blob/833a8d492c728d606454865e8c0fee84842f9f10/src/services/findAllReferences.ts#L872-L884

---

https://github.com/user-attachments/assets/810cfb6c-d78d-49b4-ad50-b16360d59631

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -419,7 +419,7 @@ func (l *LanguageService) convertSymbolAndEntryToLocation(s *SymbolAndEntries) [
 	for _, ref := range s.references {
 		if ref.textRange == nil {
 			sourceFile := ast.GetSourceFileOfNode(ref.node)
-			ref.textRange = l.createLspRangeFromNode(ref.node, ast.GetSourceFileOfNode(ref.node))
+			ref.textRange = l.getRangeOfNode(ref.node, sourceFile, nil /*endNode*/)
 			ref.fileName = sourceFile.FileName()
 		}
 		locations = append(locations, &lsproto.Location{
--- a/internal/ls/findallreferences_test.go
+++ b/internal/ls/findallreferences_test.go
@@ -241,6 +241,19 @@ class C extends [|/*3*/Base|] { }`,
 				"3": collections.NewSetFromItems("2", "3"),
 			},
 		},
+		{
+			title: "findAllRefsTrivia",
+			input: `export interface A {
+    /** Comment */
+    [|/*m1*/method|](): string;
+    /** Comment */
+    [|/*m2*/method|](format: string): string;
+}`,
+			expectedLocations: map[string]*collections.Set[string]{
+				"m1": collections.NewSetFromItems("m1", "m2"),
+				"m2": collections.NewSetFromItems("m1", "m2"),
+			},
+		},
 	}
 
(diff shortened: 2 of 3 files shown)


internal/ls/findallreferences.go
```diff
@@ -419,7 +419,7 @@ func (l *LanguageService) convertSymbolAndEntryToLocation(s *SymbolAndEntries) [
 	for _, ref := range s.references {
 		if ref.textRange == nil {
 			sourceFile := ast.GetSourceFileOfNode(ref.node)
-			ref.textRange = l.createLspRangeFromNode(ref.node, ast.GetSourceFileOfNode(ref.node))
```
gabritto, line 422: I can't tell if this is the right fix (`createLspRangeFromNode` -> `getRangeOfNode`) or if this is the right fix: https://github.com/microsoft/typescript-go/pull/1326/files#diff-4cd7fcdded6dbd561658f34fe1abb4030739555b30d90785a6ba6f08aec0d523R393

a-tarasyuk: The proposed fix is to apply the same normalization used by ts in `toReferenceEntry` for non-span entries. 

```ts
function getTextSpan(node: Node, sourceFile: SourceFile, endNode?: Node): TextSpan {
    let start = node.getStart(sourceFile);
    let end = (endNode || node).getEnd();
    if (isStringLiteralLike(node) && (end - start) > 2) {
        Debug.assert(endNode === undefined);
        start += 1;
        end -= 1;
    }
    if (endNode?.kind === SyntaxKind.CaseBlock) {
        end = endNode.getFullStart();
    }
    return createTextSpanFromBounds(start, end);
}
```

While `tsgo` has a similar utility, it wasn't used in the `find-all-references` results.

https://github.com/microsoft/typescript-go/blob/15def0236938e25ec7822755444f4d58c68045be/internal/ls/findallreferences.go#L257-L274

It would be helpful to see cases involving `CaseBlock` or `StringLiteralLike`...

gabritto (approved):
Thanks. This just needs a test baseline update and should be good to merge.

jakebailey:
Is this not redundant if we do https://github.com/microsoft/typescript-go/pull/1326/files#diff-4cd7fcdded6dbd561658f34fe1abb4030739555b30d90785a6ba6f08aec0d523R393? Or have I been misunderstanding?

gabritto:
> Is this not redundant if we do https://github.com/microsoft/typescript-go/pull/1326/files#diff-4cd7fcdded6dbd561658f34fe1abb4030739555b30d90785a6ba6f08aec0d523R393? Or have I been misunderstanding?

AFAICT it's not exactly redundant because Strada's `getTextSpan`/Corsa's `getRangeOfNode` also special cases string literals and case blocks.

PR #1356, opened 2025-07-03
Fixed some completions close to end of files

Andarist:


internal/fourslash/_scripts/failingTests.txt
Andarist, line 1: sorry for the noisy diff here - I sorted this list to help with diffing when comparing this list against the list of all current failing test cases. Upon reques, I can revert that change and just remove the newly passing tests surgically

Andarist: my process was a little bit manual when it comes to this, maybe there is a better one to learn what can be unskipped? cc @gabritto

gabritto: I usually just run a command in the terminal to sort/deduplicate this as part of updating. What I really need to do is to add a script or something to update the failing tests list. I'll do that soon, but meanwhile you can do whatever is easiest.

gabritto (approved):

PR #1357, opened 2025-07-03
Add `goTo....` functions to fourslash

gabritto:
This PR adds functions like `goTo.file(...)`, `goTo.position(...)`, to fourslash. I think I added most of the `goTo` functions that showed up in Strada's fourslash tests. The conversion script has also been updated to translate those functions, with the exception of the ones that take a function argument, e.g. `goTo.eachMarker(...)`. It's possible tests that use this function will have to be manually migrated.

This PR also has a signature help bug fix due to missing `findContainingList` function, and a bug fix in fourslash test parser where previously a marker's name had type `string`, but now has type `*string` to reflect that a `nil` name and an empty string name are different: anonymous markers originating from object markers will have a `nil` name.


internal/fourslash/fourslash.go
```diff
@@ -426,12 +512,13 @@ func (f *FourslashTest) VerifyCompletions(t *testing.T, markerInput MarkerInput,
 	}
 }
 
-func (f *FourslashTest) verifyCompletionsAtMarker(t *testing.T, markerName string, expected *CompletionsExpectedList) {
```
gabritto, line 431: This got inlined.

internal/fourslash/fourslash.go
```diff
@@ -428,12 +514,13 @@ func (f *FourslashTest) VerifyCompletions(t *testing.T, markerInput MarkerInput,
...
-	f.verifyCompletionsWorker(t, expected)
-}
-
 func (f *FourslashTest) verifyCompletionsWorker(t *testing.T, expected *CompletionsExpectedList) {
+	var prefix string
+	if f.lastKnownMarkerName != nil {
+		prefix = fmt.Sprintf("At marker '%s': ", *f.lastKnownMarkerName)
+	} else {
+		prefix = fmt.Sprintf("At position (Ln %d, Col %d): ", f.currentCaretPosition.Line, f.currentCaretPosition.Character)
+	}
```
DanielRosenwasser, line 523: Maybe I'm overthinking this, but it might be nice to extract this to a helper method since these string prefixes probably get constructed on every completions test - regardless of whether an error occurs.

internal/fourslash/fourslash.go
```diff
@@ -312,11 +317,89 @@ func (f *FourslashTest) GoToEOF(t *testing.T) {
...
+	LSPPos := f.converters.PositionToLineAndCharacter(script, core.TextPos(position))
+	f.goToPosition(t, LSPPos)
+}
+
 func (f *FourslashTest) goToPosition(t *testing.T, position lsproto.Position) {
 	f.currentCaretPosition = position
 	f.selectionEnd = nil
 }
 
+func (f *FourslashTest) GoToEachMarker(t *testing.T, markerNames []string, action func(t *testing.T, marker *Marker, index int)) {
```
DanielRosenwasser, line 335: Out of curiosity, why is `t` passed through to the callback? The only test I see using this closes over the outer `FourslashTest` anyway (though who knows if that gets inlined).

DanielRosenwasser (commented):
Looks good apart from the minor things. Would be helpful to get a second pair of eyes though.

weswigham (approved):

internal/fourslash/fourslash.go
gabritto: I forgot those comments, but yeah, I don't remember exactly what scenario I had in mind for passing the test, but I think you're right and it's probably not necessary. I'll fix that and the other comment in my next PR.

PR #1358, opened 2025-07-03
Port TS PR 61505 - Cache mapper instantiations

jakebailey:
This is an early port of https://github.com/microsoft/TypeScript/pull/61505, because I like to go fast

Calculating the key increases allocs by a bit (really wish we had the yet-to-exist Go custom map implementation), but for https://github.com/AnswerOverflow/AnswerOverflow/tree/1ab687793a598789a0c86b4ca5f17ee6f3a256aa/apps/dashboard:

TypeScript 5.8 vs 5.9:

```
Files:              3351
Lines:            573983
Identifiers:      646173
Symbols:          652920
Types:            139674
Instantiations: 13672403
Memory used:     912429K
I/O read:          0.13s
I/O write:         0.00s
Parse time:        1.91s
Bind time:         0.60s
Check time:        7.67s
Emit time:         0.12s
Total time:       10.30s
```

```
Files:             3353
Lines:           575420
Identifiers:     646578
Symbols:         651905
Types:           139189
Instantiations:  797184
Memory used:    898938K
I/O read:         0.10s
I/O write:        0.00s
Parse time:       1.84s
Bind time:        0.61s
Check time:       3.79s
Emit time:        0.11s
Total time:       6.35s
```

tsgo main vs this PR;

```
Files:              3501
Lines:            601735
Identifiers:      661699
Symbols:         1219580
Types:            420852
Instantiations: 25441996
Memory used:     862438K
Memory allocs:  63315601
Config time:      0.002s
Parse time:       0.170s
Bind time:        0.044s
Check time:       2.034s
Emit time:        0.000s
Total time:       2.257s
```

```
Files:              3501
Lines:            601735
Identifiers:      661699
Symbols:         1219580
Types:            419356
Instantiations:  2178836
Memory used:     861708K
Memory allocs:  12761985
Config time:      0.003s
Parse time:       0.163s
Bind time:        0.048s
Check time:       1.107s
Emit time:        0.000s
Total time:       1.327s
```

(Instantiations look higher than 5.8 or 5.9 because tsgo's metrics sum up instantiations between all concurrent checkers)

```
Benchmark 1: ~/work/TypeScript-go/built/local-old/tsgo
  Time (mean ± σ):      2.270 s ±  0.078 s    [User: 13.092 s, System: 1.112 s]
  Range (min … max):    2.164 s …  2.392 s    10 runs

Benchmark 2: ~/work/TypeScript-go/built/local/tsgo
  Time (mean ± σ):      1.376 s ±  0.069 s    [User: 7.626 s, System: 1.028 s]
  Range (min … max):    1.293 s …  1.531 s    10 runs

Summary
  ~/work/TypeScript-go/built/local/tsgo ran
    1.65 ± 0.10 times faster than ~/work/TypeScript-go/built/local-old/tsgo
```

```
Benchmark 1 (3 runs): /home/jabaile/work/TypeScript-go/built/local-old/tsgo
  wall_time          2.29s  ± 90.2ms    2.23s  … 2.39s           0 ( 0%)        0%
  peak_rss           1.46GB ± 92.0MB    1.36GB … 1.55GB          0 ( 0%)        0%
  cpu_cycles         59.7G  ± 1.84G     57.7G  … 61.1G           0 ( 0%)        0%
  instructions       85.3G  ± 2.30G     82.7G  … 87.2G           0 ( 0%)        0%
  cache_references   1.19G  ± 21.0M     1.17G  … 1.22G           0 ( 0%)        0%
  cache_misses        251M  ± 6.31M      244M  …  256M           0 ( 0%)        0%
  branch_misses       177M  ± 1.79M      175M  …  179M           0 ( 0%)        0%
Benchmark 2 (4 runs): /home/jabaile/work/TypeScript-go/built/local/tsgo
  wall_time          1.37s  ± 80.5ms    1.30s  … 1.47s           0 ( 0%)        ⚡- 40.1% ±  7.3%
  peak_rss           1.13GB ± 7.47MB    1.12GB … 1.14GB          0 ( 0%)        ⚡- 22.8% ±  7.9%
  cpu_cycles         34.1G  ± 1.02G     33.3G  … 35.5G           0 ( 0%)        ⚡- 42.9% ±  4.6%
  instructions       40.8G  ±  912M     39.9G  … 41.7G           0 ( 0%)        ⚡- 52.2% ±  3.7%
  cache_references    979M  ± 17.1M      963M  … 1.00G           0 ( 0%)        ⚡- 18.0% ±  3.1%
  cache_misses        154M  ± 8.52M      146M  …  165M           0 ( 0%)        ⚡- 38.5% ±  6.0%
  branch_misses       109M  ± 2.03M      107M  …  112M           0 ( 0%)        ⚡- 38.8% ±  2.1%
```

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -21206,6 +21206,7 @@ func (c *Checker) pushActiveMapper(mapper *TypeMapper) {
 
 	lastIndex := len(c.activeTypeMappersCaches)
 	if cap(c.activeTypeMappersCaches) > lastIndex {
+		// The cap may contain an empty map from popActiveMapper; reuse it.
 		c.activeTypeMappersCaches = c.activeTypeMappersCaches[:lastIndex+1]
 		if c.activeTypeMappersCaches[lastIndex] == nil {
 			c.activeTypeMappersCaches[lastIndex] = make(map[string]*Type, 1)


Andarist:
Okay. I approve.

internal/checker/checker.go
```diff
@@ -21174,14 +21176,64 @@ func (c *Checker) instantiateTypeWithAlias(t *Type, m *TypeMapper, alias *TypeAl
...
+}
+
+func (c *Checker) pushActiveMapper(mapper *TypeMapper) {
+	c.activeMappers = append(c.activeMappers, mapper)
+
+	c.activeTypeMappersCaches = append(c.activeTypeMappersCaches, activeMapperCache.Get().(map[string]*Type))
+}
+
+func (c *Checker) popActiveMapper() {
+	c.activeMappers[len(c.activeMappers)-1] = nil
```
jakebailey, line 21217: FWIW this nil-ing out is _not_ done in Strada, but should be, because the stacks in Strada are always held onto. I'm going to send a PR to fix that in Strada.

Andarist: if I understand the concern correctly, this stack was modeled after the one that deals with contextual types - `popContextualType` doesn't clear up stale entries either. My assumption was the claim is that those slots will (likely) eventually be replaced by other entries anyway so it's not worth clearing the stack entry itself. But with those mappers the situation might be slightly more delicate as the mapper cache might statisictally hold onto more objects than a contextual type

jakebailey: Ah, you're right, and those are wrong in the Go code too.

internal/checker/checker.go
```diff
@@ -856,6 +856,8 @@ type Checker struct {
 	skipDirectInferenceNodes                    collections.Set[*ast.Node]
 	ctx                                         context.Context
 	packagesMap                                 map[string]bool
+	activeMappers                               []*TypeMapper
+	activeTypeMappersCaches                     []map[string]*Type
```
jakebailey, line 860: The other "global" caches don't use two separate lists in the Go port, instead scoping them together, but I'm keeping them separate since I'm trying to reuse the maps.

internal/checker/checker.go
jakebailey: https://github.com/microsoft/TypeScript/pull/62016

internal/checker/checker.go
```diff
@@ -21174,14 +21176,64 @@ func (c *Checker) instantiateTypeWithAlias(t *Type, m *TypeMapper, alias *TypeAl
...
+	c.activeTypeMappersCaches = append(c.activeTypeMappersCaches, activeMapperCache.Get().(map[string]*Type))
+}
+
+func (c *Checker) popActiveMapper() {
+	c.activeMappers[len(c.activeMappers)-1] = nil
+	c.activeMappers = c.activeMappers[:len(c.activeMappers)-1]
+
+	cache := c.activeTypeMappersCaches[len(c.activeTypeMappersCaches)-1]
+	c.activeTypeMappersCaches[len(c.activeTypeMappersCaches)-1] = nil
+	clear(cache)
```
DanielRosenwasser: Something I'm not clear on - why do you need to clear the cache in addition to setting the old map to `nil`? Can you just `nil` it out?

internal/checker/checker.go
```diff
@@ -21174,14 +21176,64 @@ func (c *Checker) instantiateTypeWithAlias(t *Type, m *TypeMapper, alias *TypeAl
...
+	c.activeTypeMappersCaches = c.activeTypeMappersCaches[:len(c.activeTypeMappersCaches)-1]
+}
+
+func (c *Checker) findActiveMapper(mapper *TypeMapper) int {
+	return core.FindLastIndex(c.activeMappers, func(m *TypeMapper) bool { return m == mapper })
+}
+
+func (c *Checker) clearActiveMapperCaches() {
+	for _, cache := range c.activeTypeMappersCaches {
+		clear(cache)
```
DanielRosenwasser, line 21233: Why does the old code not just set them all to `undefined` instead of clearing all of them and keeping them in place?

DanielRosenwasser (approved):

internal/checker/checker.go
jakebailey: It's cleared before I `Put` it back in the pool.

internal/checker/checker.go
jakebailey: Since that would mean call stacks which have pushed onto them will then have an undefined map. Well, they wouldn't, but they'd have a reference to a cache that would never be used again.

internal/checker/checker.go
jakebailey: That being said, I should take an approach more like https://github.com/microsoft/TypeScript/pull/62016 and just keep the maps in the slice but empty.

internal/checker/checker.go
```diff
@@ -21174,14 +21176,65 @@ func (c *Checker) instantiateTypeWithAlias(t *Type, m *TypeMapper, alias *TypeAl
...
 }
 
+func (c *Checker) pushActiveMapper(mapper *TypeMapper) {
+	c.activeMappers = append(c.activeMappers, mapper)
+
+	lastIndex := len(c.activeTypeMappersCaches)
+	if cap(c.activeTypeMappersCaches) > lastIndex {
+		c.activeTypeMappersCaches = c.activeTypeMappersCaches[:lastIndex+1]
+		if c.activeTypeMappersCaches[lastIndex] == nil {
+			c.activeTypeMappersCaches[lastIndex] = make(map[string]*Type, 1)
```
DanielRosenwasser: Trying to reuse an empty map is already  subtle - I see the comment in `popActiveMapper`, but leave a comment here as well since this code is also involved.

PR #1361, opened 2025-07-04
Add script to update list of failing generated fourslash tests

gabritto:
Now you can do `npm run updatefailing` and the script will delete the current list of failing generated fourslash tests, re-generate the tests again (so that none is skipped), run the tests, and write the new list of failing tests.


gabritto:
@Andarist would this help with the process of fixing generated fourslash tests or do you need something else?

internal/fourslash/_scripts/updateFailing.mts
Andarist, line 1: This is great 👍 it's a similar flow I've used manually myself

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -2,22 +2,26 @@ import * as cp from "child_process";
...
-const failingTests = new Set(failingTestsList);
 const helperFilePath = path.join(import.meta.dirname, "../", "tests", "util_test.go");
 
 const outputDir = path.join(import.meta.dirname, "../", "tests", "gen");
 
 const unparsedFiles: string[] = [];
 
-function main() {
+function getFailingTests(): Set<string> {
+    const failingTestsList = fs.readFileSync(failingTestsPath, "utf-8").split("\n").map(line => line.trim().substring(4)).filter(line => line.length > 0);
```
jakebailey, line 20: Does this need to be sorted? Some OSs return these out of order

gabritto: No, we just use this list to construct the set of failing tests. Only the script that updates the list should keep it sorted, and even then just for ease of diff/review.

gabritto: Ok now I'm curious, what exactly is returned out of order in some OSs?

jakebailey: Oops. I misread this as readdirSync

jakebailey (approved):

PR #1362, opened 2025-07-06
Use absolute paths in error messages for CI environments

tmm1:
This solves the same need as [tsc-absolute](https://github.com/microsoft/TypeScript/issues/7238) or [this patch](https://github.com/microsoft/TypeScript/issues/36221#issuecomment-2106656302)

Errors reported by `tsc` are normally automatically surfaced as annotations in GitHub PRs, using Problem Matchers

However this fails to work in monorepos where multiple typescript projects exist. See [this blog post](https://christopher.xyz/2023/04/04/github-actions-problem-matchers.html) for more details.

cc https://github.com/microsoft/TypeScript/issues/7238 https://github.com/microsoft/TypeScript/issues/36221

(Opening for visibility; I understand divergence from Strada is being avoided at this point)

### DIFF
--- a/internal/diagnosticwriter/diagnosticwriter.go
+++ b/internal/diagnosticwriter/diagnosticwriter.go
@@ -17,7 +17,8 @@ import (
 
 type FormattingOptions struct {
 	tspath.ComparePathsOptions
-	NewLine string
+	NewLine       string
+	AbsolutePaths bool
 }
 
 const (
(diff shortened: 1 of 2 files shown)


tmm1:
Actually, this isn't right because I changed the pretty output paths, and the [default matcher](https://github.com/actions/setup-node/blob/main/.github/tsc.json) only works for the `--pretty false` path.

The matchers [don't do well with color](https://github.com/actions/runner/issues/2341) anyway, but some people have resorted to putting [color codes into regex](https://github.com/actions/runner/issues/2341#issuecomment-2981938438).

PR #1368, opened 2025-07-07
Temporarily work around GetLineMap CPU usage

jakebailey:
Something I'm experiencing, even more after #1363, is that a load of CPU time is spent in `ComputeLineStarts` because when I'm typing, the versions changed and the mutable-ness of the script infos and projects means we skip reusing the line map.

This leads to a catastrophic CPU usage where we never cache _any_ line starts at all once the version is wrong, so requests in files like `checker.ts` for document symbols just keep going forever, recomputing the same info.

I believe that this all should go away in the new snapshot model, but for now this trick seems to contain the problem enough to make the editor usable again while typing, including making completions in `checker.ts` basically instant again.

### DIFF
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -42,6 +42,7 @@ type snapshot struct {
 	project          *Project
 	positionEncoding lsproto.PositionEncodingKind
 	program          *compiler.Program
+	lineMaps         collections.SyncMap[*ast.SourceFile, *ls.LineMap]
 }
 
 // GetLineMap implements ls.Host.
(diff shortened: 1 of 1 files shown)


internal/project/project.go
```diff
@@ -51,7 +52,13 @@ func (s *snapshot) GetLineMap(fileName string) *ls.LineMap {
 	if s.project.getFileVersion(file) == scriptInfo.Version() {
```
jakebailey, line 52: I still don't really understand why we're doing this at all; it seems to me that we should not cache line maps on script infos at all, instead do it per-source-file, and then just always use that. But I seem to remember someone saying that was wrong for some reason, so, this hack seems better than nothing?

internal/project/project.go
```diff
@@ -42,6 +42,7 @@ type snapshot struct {
 	project          *Project
 	positionEncoding lsproto.PositionEncodingKind
 	program          *compiler.Program
+	lineMaps         collections.SyncMap[*ast.SourceFile, *ls.LineMap]
 }
```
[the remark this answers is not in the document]
jakebailey: No, these snapshots are request-scoped

gabritto (approved):
Seems ok for a temporary fix, but I also don't get why a version change would be more frequent than a change to a source file object...

jakebailey:
I'm pretty sure this is a race condition while typing, when the ScriptInfo has been updated with new text, but the request is still running on the old Program and therefore has a file with a different version.

internal/project/project.go
DanielRosenwasser: I recall a conversation around line maps differing between LS and CLI

jakebailey: That's a different thing but yeah they do.

PR #1369, opened 2025-07-07
Clear out checker-level stacks on pop

jakebailey:
See https://github.com/microsoft/TypeScript/pull/62016, https://github.com/microsoft/typescript-go/pull/1358#discussion_r2191080650

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -17879,6 +17879,7 @@ func (c *Checker) pushTypeResolution(target TypeSystemEntity, propertyName TypeS
 func (c *Checker) popTypeResolution() bool {
 	lastIndex := len(c.typeResolutions) - 1
 	result := c.typeResolutions[lastIndex].result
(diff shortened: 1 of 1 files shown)


internal/checker/checker.go
```diff
@@ -29354,7 +29355,9 @@ func (c *Checker) pushContextualType(node *ast.Node, t *Type, isCache bool) {
 }
 
 func (c *Checker) popContextualType() {
-	c.contextualInfos = c.contextualInfos[:len(c.contextualInfos)-1]
+	lastIndex := len(c.contextualInfos) - 1
+	c.contextualInfos[lastIndex] = ContextualInfo{}
+	c.contextualInfos = c.contextualInfos[:lastIndex]
```
jakebailey, line 29360: This is honestly better written `slices.Delete`, which also nils things out, but, whatever

DanielRosenwasser (approved):

jakebailey:
Long term we should just have a `Stack` type. We probably do somewhere else, but I'd rather futz with that in a separate PR.

PR #1371, opened 2025-07-08
Port ES2016 transform

weswigham:
Which is just the exponentiation operators - `**` and `**=`


jakebailey (approved):
I didn't even know JS had this operator

weswigham:
> I didn't even know JS had this operator

It has funky rules because coming to consensus is hard. Most things need parens on the LHS to work, by choice, even though it's not required to parse it. Most notably, `-2 ** 2` is just invalid code, because nobody wanted to dictate if it was equivalent to `(-2) ** 2` or `-(2 ** 2)`.

PR #1373, opened 2025-07-08
Port some more truncation check related code

jakebailey:
These are things I found while trying to work through #988.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -67,6 +67,7 @@ type NodeFactory struct {
 	importSpecifierPool               core.Pool[ImportSpecifier]
 	indexedAccessTypeNodePool         core.Pool[IndexedAccessTypeNode]
 	interfaceDeclarationPool          core.Pool[InterfaceDeclaration]
+	intersectionTypeNodePool          core.Pool[IntersectionTypeNode]
 	jsdocDeprecatedTagPool            core.Pool[JSDocDeprecatedTag]
 	jsdocParameterOrPropertyTagPool   core.Pool[JSDocParameterOrPropertyTag]
 	jsdocPool                         core.Pool[JSDoc]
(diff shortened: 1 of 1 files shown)


internal/checker/nodebuilderimpl.go
```diff
@@ -212,12 +213,121 @@ func (b *nodeBuilderImpl) createElidedInformationPlaceholder() *ast.TypeNode {
 	return b.f.NewKeywordTypeNode(ast.KindAnyKeyword)
 }
 
-func (b *nodeBuilderImpl) mapToTypeNodes(list []*Type) *ast.NodeList {
+func (b *nodeBuilderImpl) mapToTypeNodes(list []*Type, isBareList bool) *ast.NodeList {
```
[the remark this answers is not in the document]
jakebailey: if we had a time machine, sure

internal/checker/nodebuilderimpl.go
```diff
@@ -212,12 +213,121 @@ func (b *nodeBuilderImpl) createElidedInformationPlaceholder() *ast.TypeNode {
...
 	if len(list) == 0 {
 		return nil
 	}
-	contents := core.Map(list, b.typeToTypeNode)
-	return b.f.NewNodeList(contents)
+
+	if b.checkTruncationLength() {
+		if !isBareList {
+			var node *ast.Node
```
jakebailey, line 223: This entire block is unused per codecov, so not sure what the deal is there; it might not even be used in Strada

weswigham (approved):
Yep. `mapToTypeNodes` was still a shorthand from the initial printer. Probably shoulda had a `// !!!` on it as a reminder that it still needed work.

internal/checker/nodebuilderimpl.go
weswigham: I think it's triggered by some LS tests in Strada? I vaguely remember adding this truncation check (along with the hard cut at the top level) in response to an LS request freezing the editor with a 50MB hover.

PR #1377, opened 2025-07-09
Better formatting for generated fourslash tests

gabritto:
I added some line breaks to the generated tests and moved the position of some properties to make them more readable.


internal/fourslash/_scripts/failingTests.txt
jakebailey, line 1: It looks like the ordering here keeps changing; somewhere there's instability in the sort order.

gabritto: I think we just don't check that the file is ordered in CI, and I think https://github.com/microsoft/typescript-go/pull/1274  merged without using the updating script I added. I'm about to replace the CI check to use the updating script.

gabritto: Waiting for https://github.com/microsoft/typescript-go/pull/1378 so the failing test diff goes away.

jakebailey (approved):

PR #1280, opened 2025-06-24
Do less allocation in JSDoc parsing

jakebailey:
One of the largest allocators when parsing code for the editor is JSDoc parsing. By deferring JSDoc part joining until `Text()` is called, we can save a bunch of memory allocations, and make parsing out the DOM types ~5% faster.

```
goos: linux
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/parser
cpu: Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz
                                                             │   old.txt   │              new.txt               │
                                                             │   sec/op    │   sec/op     vs base               │
Parse/empty.ts/tsc-20                                          527.5n ± 4%   523.9n ± 1%       ~ (p=0.072 n=10)
Parse/empty.ts/server-20                                       524.1n ± 1%   527.3n ± 1%       ~ (p=0.138 n=10)
Parse/checker.ts/tsc-20                                        48.66m ± 2%   47.70m ± 1%  -1.97% (p=0.002 n=10)
Parse/checker.ts/server-20                                     48.88m ± 4%   49.15m ± 1%       ~ (p=0.912 n=10)
Parse/dom.generated.d.ts/tsc-20                                16.39m ± 2%   16.11m ± 1%  -1.69% (p=0.023 n=10)
Parse/dom.generated.d.ts/server-20                             24.56m ± 2%   23.54m ± 3%  -4.13% (p=0.000 n=10)
Parse/Herebyfile.mjs/tsc-20                                    884.3µ ± 1%   881.8µ ± 1%       ~ (p=0.143 n=10)
Parse/Herebyfile.mjs/server-20                                 886.8µ ± 3%   880.6µ ± 2%       ~ (p=0.165 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/tsc-20      267.4µ ± 2%   265.2µ ± 2%  -0.80% (p=0.015 n=10)
Parse/jsxComplexSignatureHasApplicabilityError.tsx/server-20   419.1µ ± 2%   403.0µ ± 3%  -3.83% (p=0.000 n=10)
geomean                                                        686.5µ        677.5µ       -1.31%
```
```
                                                             │   old.txt    │                new.txt                │
                                                             │     B/op     │     B/op      vs base                 │
Parse/empty.ts/tsc-20                                            881.0 ± 0%     881.0 ± 0%       ~ (p=1.000 n=10) ¹
[...]

PR #1384, opened 2025-07-10
Handle `completionItem/resolve`

gabritto:
This PR implements LSP's `completionItem/resolve` method in the language server.
This roughly corresponds to Strada's `completionEntryDetails`.

`completionItem/resolve` method sends a completion item to the server, and expects the same completion item with lazily-computed properties filled in. Right now vscode supports lazy computation of `documentation`, `detail`, and `additionalTextEdits`.
This PR implements computing `detail` (Strada's `text`) and `documentation` (Strada's `documentation`). `additionalTextEdits` will be used once we have auto-imports.
Note that for now, if a client doesn't support lazy computation of `detail` and `documentation`, we simply won't fill in those properties.

The completion code is the same as it was in Strada, with one exception. To be able to compute the properties for a completion item resolve call, we effectively need to do much of the same work we do for the original completion request, e.g. we need to collect symbols available at the requested file and location. To reconstruct the original completion request, need to know the file and position of the original request, among other things. That information is now passed in the completion item `data` property in the original completion response, which is always preserved by the client. In Strada, this information was passed by the client in the `completionEntryDetails` request.

Another note is that many of the tests ported in this PR fail because quickinfo produces a different string. Hopefully that will be fixed later when we have a more complete implementation of quickinfo, but in any case I'll start fixing the failing tests after this PR.


internal/ls/completions.go
```diff
@@ -4543,3 +4552,289 @@ func getArgumentInfoForCompletions(node *ast.Node, position int, file *ast.Sourc
...
+	program, file := l.tryGetProgramAndFile(data.FileName)
+	if file == nil {
+		return nil, fmt.Errorf("file not found: %s", data.FileName)
+	}
+
+	return l.getCompletionItemDetails(ctx, program, data.Position, file, item, data, clientOptions, preferences), nil
+}
+
+func GetCompletionItemData(item *lsproto.CompletionItem) (*itemData, error) {
+	bytes, err := json.Marshal(item.Data)
```
jakebailey, line 4611: To avoid this, we should probably special case some of these incoming `any`s as `json.RawMessage`, maybe?

internal/ls/completions.go
```diff
@@ -4543,3 +4552,289 @@ func getArgumentInfoForCompletions(node *ast.Node, position int, file *ast.Sourc
...
+	item *lsproto.CompletionItem,
+	data *itemData,
+	clientOptions *lsproto.CompletionClientCapabilities,
+	preferences *UserPreferences,
+) (*lsproto.CompletionItem, error) {
+	if data == nil {
+		return nil, errors.New("completion item data is nil")
+	}
+
+	program, file := l.tryGetProgramAndFile(data.FileName)
```
jakebailey, line 4602: In the near future, I sort of wonder how resolve will work in the snapshot model where things can change. Should we include some sort of snapshot number that indicates where a completion came from and reject it if the snapshot has been dropped?

Or, I guess in this code, is it possible for this code to fail in an unhand able way just because the user is typing while this is happening?

internal/parser/jsdoc.go
```diff
@@ -285,6 +286,7 @@ loop:
 	}
 
 	if len(comments) > 0 {
+		comments[len(comments)-1] = strings.TrimRightFunc(comments[len(comments)-1], unicode.IsSpace)
```
jakebailey, line 289: Ugh, so there was a place where this mattered? can we instead post-process those away rather than trimming just one side of the comments or something?

internal/ls/completions.go
gabritto: Yes, that would be nice. I can try and add this to the lsp generator in a separate PR.

jakebailey: The annoying bit is that it would probably break things where we are passing the objects as-is, I think

internal/ls/completions.go
```diff
@@ -4193,7 +4202,7 @@ func (l *LanguageService) createLSPCompletionItem(
 		InsertTextFormat: insertTextFormat,
 		TextEdit:         textEdit,
 		CommitCharacters: commitCharacters,
-		Data:             nil, // !!! auto-imports
+		Data:             &data,
```
jakebailey, line 4205: This will eventually come to bite us, but one thing we are not doing at all in the LS at the moment is respecting the _client's_ capabilities. All of these things should actually be things where we check the client's caps and say "do they support resolve?", and if not,  include all of this information like docs up front, or similar.

Not really critical for this PR, but something to think about.

internal/ls/completions.go
gabritto: It's possible we can't resolve an item because something in the program changed, yes. But it was already possible in Strada, I think. If this fails though, the worse that can happen is we try to request completions in a file/position that doesn't exist or that produces different symbols, and then we simply won't fill in the lazy properties.
Also, I don't know if it's as simple as "this file changed, we can no longer resolve the completion item", because I can imagine the common scenario being a user starts typing, the client request completions and gets back a list, and then the user continues typing the same identifier, let's say. The file has changed, but if you request completions at the old position you'd get the same list, and so we'll still be able to resolve the items. And in this scenario I think the client won't re-request completions with every new character, but will instead re-use the old list for a bit and filter it down based on the new characters. So I don't know, maybe it's fine as it is, or maybe we can do better and re-use the old snapshot once we have that?

jakebailey:
yesss

image

internal/parser/jsdoc.go
gabritto: This matters for quickinfo and everything that depends on it, and possibly other places that handle jsdoc, we just didn't have tests for any of this before. Do you mean post-processing in quickinfo itself? I added this code here in the parser because that's exactly what Strada does, and I don't know if it'll be enough to just do it on quickinfo.

internal/ls/completions.go
gabritto: Yep, that's what I meant to note in the description:
>Note that for now, if a client doesn't support lazy computation of detail and documentation, we simply won't fill in those properties.

Especially for auto-imports, this could be detrimental to performance.

internal/parser/jsdoc.go
jakebailey: I don't know if what's added here is exactly like Strada; see also #1280 where I basically undid a bunch of this in hopes it wouldn't affect anything, with the assumption that any caller who cared would instead do that processing themself or something.

But, maybe it's `.Text()` which needs to do that.

internal/ls/completions.go
jakebailey: Ah, missed it, great.

internal/ls/completions.go
gabritto: Right, because then the type of `CompletionItem.Data` would be bytes and we'd have to marshall that as we construct the object.

internal/parser/jsdoc.go
gabritto: Well, it's the closest equivalent to Strada's, since that part of the code changed a bit in Corsa. But I'm open to ideas here, I'm just not very familiar with jsdoc handling.
The thing I find tricky with changing `.Text()` is that we want to trim the last comment text on a list of comments, right? since the intermediary ones should keep their trailing spaces.

jakebailey: Eh, what you've done is fine, if it really is just the last one that must be trimmed.

PR #1385, opened 2025-07-10
Port `?.` emit

weswigham:
Which is the remainder of the `es2020` transform.

There are some discrepancies in parenthesis (we over-parenthesize in some cases) in the emit, seemingly caused by changes to how we parenthesize in general.


jakebailey:
> There are some discrepancies in parenthesis (we over-parenthesize in some cases) in the emit, seemingly caused by changes to how we parenthesize in general.

Yeah, that all is WIP and not done :(

Related is #1055 #1123

testdata/baselines/reference/submodule/compiler/assertionFunctionsCanNarrowByDiscriminant.js.diff
```diff
@@ -7,8 +7,4 @@
 -"use strict";
```
jakebailey, line 7: I wonder if this very common diff is an easy fix somewhere...

testdata/baselines/reference/submodule/compiler/optionalChainWithInstantiationExpression2(target=es2019).js.diff
```diff
@@ -1,11 +1,8 @@
...
-@@= skipped -15, +15 lines =@@
-
-
+@@= skipped -17, +17 lines =@@
  //// [optionalChainWithInstantiationExpression2.js]
--var _a;
--a === null || a === void 0 ? void 0 : a();
+ var _a;
+ a === null || a === void 0 ? void 0 : a();
 -(_a = (a)) === null || _a === void 0 ? void 0 : _a();
```
jakebailey, line 7: Wow, a place where Strada over-parenthesized 😄

testdata/baselines/reference/submodule/conformance/elementAccessChain.3.js.diff
```diff
@@ -16,55 +16,31 @@
...
--    ;
--for (obj === null || obj === void 0 ? void 0 : obj.a["b"] in {})
--    ;
--for (obj === null || obj === void 0 ? void 0 : obj["a"] of [])
--    ;
--for (obj === null || obj === void 0 ? void 0 : obj.a["b"] of [])
--    ;
--({ a: obj === null || obj === void 0 ? void 0 : obj["a"] } = { a: 1 });
--({ a: obj === null || obj === void 0 ? void 0 : obj.a["b"] } = { a: 1 });
++(obj === null || obj === void 0 ? void 0 : obj["a"]) = 1;
```
jakebailey, line 30: This seems to be incorrect, right? This is trying to assign 1 to the parenthesized expression?

testdata/baselines/reference/submodule/conformance/optionalChainingInParameterBindingPattern(target=es5).js.diff
```diff
@@ -5,4 +5,4 @@
  // https://github.com/microsoft/TypeScript/issues/36295
  const a = () => undefined;
 -(({ [(_a = a()) === null || _a === void 0 ? void 0 : _a.d]: c = "" }) => { var _a; })();
-+(({ [a()?.d]: c = "" }) => { })();
\ No newline at end of file
++((_a) => { var _b; var { [(_b = a()) === null || _b === void 0 ? void 0 : _b.d]: c = "" } = _a; })();
```
jakebailey, line 8: This seems funky. (But maybe still equivalent?)

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+		return ch.visitDeleteExpression(node.AsDeleteExpression())
+	default:
+		return ch.Visitor().VisitEachChild(node)
+	}
+}
+
+func (ch *optionalChainTransformer) visitCallExpression(node *ast.CallExpression, captureThisArg bool) *ast.Node {
+	if node.Flags&ast.NodeFlagsOptionalChain != 0 {
+		// If `node` is an optional chain, then it is the outermost chain of an optional expression.
+		return ch.visitOptionalExpression(node.AsNode(), captureThisArg, false)
```
DanielRosenwasser, line 36: Can you restore the `isDelete` comments?

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+	return ch.Factory().UpdateParenthesizedExpression(node, expr)
+}
+
+func (ch *optionalChainTransformer) visitPropertyOrElementAccessExpression(node *ast.Expression, captureThisArg bool, isDelete bool) *ast.Expression {
+	if node.Flags&ast.NodeFlagsOptionalChain != 0 {
+		// If `node` is an optional chain, then it is the outermost chain of an optional expression.
+		return ch.visitOptionalExpression(node.AsNode(), captureThisArg, isDelete)
+	}
+	expression := ch.Visitor().VisitNode(node.Expression())
+	// Debug.assertNotNode(expression, isSyntheticReference); // !!!
```
DanielRosenwasser, line 75: Does anything go wrong if you panic here?

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+	}
+	return expression
+}
+
+func (ch *optionalChainTransformer) visitDeleteExpression(node *ast.DeleteExpression) *ast.Node {
+	unwrapped := ast.SkipParentheses(node.Expression)
+	if unwrapped.Flags&ast.NodeFlagsOptionalChain != 0 {
+		return ch.visitNonOptionalExpression(node.Expression, false, true)
+	}
+	return ch.Visitor().VisitEachChild(node.AsNode())
```
DanielRosenwasser, line 109: Any reason you do this instead of the original-ish

```suggestion
	deletedExpr := ch.Visitor().VisitNode(node.Expression)
	return ch.Factory().UpdateDeleteExpression(node, deletedExpr)
```

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+}
+
+func flattenChain(chain *ast.Node) flattenResult {
+	// Debug.assertNotNode(chain, isNonNullChain); // !!!
+	links := []*ast.Node{chain}
+	for !ast.IsTaggedTemplateExpression(chain) && chain.QuestionDotToken() == nil {
+		chain = ast.SkipPartiallyEmittedExpressions(chain.Expression())
+		// Debug.assertNotNode(chain, isNonNullChain); // !!!
+		links = append([]*ast.Node{chain}, links...)
+	}
```
DanielRosenwasser, line 137: I know that this was the original but it seems so not-ideal - why didn't we just keep appending and reverse?


```suggestion
		links = append(links, chain)
	}
    slices.Reverse(links)
```

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+func (ch *optionalChainTransformer) visitOptionalExpression(node *ast.Node, captureThisArg bool, isDelete bool) *ast.Node {
+	r := flattenChain(node)
+	expression := r.expression
+	chain := r.chain
+	left := ch.visitNonOptionalExpression(ast.SkipPartiallyEmittedExpressions(expression), isCallChain(chain[0]), false)
+	var leftThisArg *ast.Expression
+	capturedLeft := left
+	if ast.IsSyntheticReferenceExpression(left) {
+		leftThisArg = left.AsSyntheticReferenceExpression().ThisArg
+		capturedLeft = left.AsSyntheticReferenceExpression().Expression
```
DanielRosenwasser, line 154:
```suggestion
	if ast.IsSyntheticReferenceExpression(left) {
        	left := left.AsSyntheticReferenceExpression()
		leftThisArg = left.ThisArg
		capturedLeft = left.Expression
```

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+			if i == len(chain)-1 && captureThisArg {
+				if !transformers.IsSimpleCopiableExpression(rightExpression) {
+					thisArg = ch.Factory().NewTempVariable()
+					ch.EmitContext().AddVariableDeclaration(thisArg)
+					rightExpression = ch.Factory().NewAssignmentExpression(thisArg, rightExpression)
+				} else {
+					thisArg = rightExpression
+				}
+			}
+			if segment.Kind == ast.KindElementAccessExpression {
```
DanielRosenwasser, line 177: nit - the check is inverted compared to the original code

internal/transformers/estransforms/optionalchain.go
```diff
@@ -11,7 +11,222 @@ type optionalChainTransformer struct {
...
+	} else {
+		target = ch.Factory().NewConditionalExpression(
+			createNotNullCondition(ch.EmitContext(), leftExpression, capturedLeft, true),
+			ch.Factory().NewToken(ast.KindQuestionToken),
+			ch.Factory().NewVoidZeroExpression(),
+			ch.Factory().NewToken(ast.KindColonToken),
+			rightExpression,
+		)
+	}
+	target.Loc = node.Loc
```
DanielRosenwasser, line 224: Any reason you have to do this? I see you run `SetOriginal` farther below. The ordering is a little bit different here though. I would have thought you'd run `SetOriginal` here based on the original code.

testdata/baselines/reference/submodule/conformance/optionalChainingInParameterBindingPattern(target=es5).js.diff
weswigham: This is just the >es5 output, since there's no es5 transform. In <=es5, we used to just emit a `var _a` in the body following the parameter list, which, pre-es6, was valid.

internal/transformers/estransforms/optionalchain.go
weswigham: These asserts were to satisfy the TS typechecker and remove a type from a union type. Technically, they're not actually required anymore, but _if_ they're added back, they should be in the same systematic debug-build-specific style, not as ad-hoc panics.

internal/transformers/estransforms/optionalchain.go
weswigham: Because they're equivalent (as best I can tell). Why manually pull out and `update` the child of the node, when `visitEachChild` does that already.

internal/transformers/estransforms/optionalchain.go
weswigham: `.Loc = .Loc` is equivalent to `SetTextRange`, which is independent from `SetOriginalNode` (text ranges control comment and source map spans, original nodes control emit resolver API nodes and in *some* cases fallback location spans) - and the original nodes are set in new places now because we now have an assert if we try and set original multiple times (to different things), so all these codepaths now *reliably* set originals when they're done constructing the new node, rather than sometimes doing it and sometimes letting the caller handle it.

internal/transformers/estransforms/optionalchain.go
weswigham: As are the bodies in the branches of the check. `E` comes before `P`, so `E` comes first in the `case` clause list, so handling the first case explicitly is a bit more natural looking.

internal/transformers/estransforms/optionalchain.go
weswigham: I can... but the go formatting for it is weird af, placing the comment before the preceding comma. I've been avoiding most of those comments since I noticed it. With only two callsites that aren't named variables (both of which pass `false`), I'm not too worried about this one being too confusing.

testdata/baselines/reference/submodule/conformance/elementAccessChain.3.js.diff
weswigham: This is the structure we make in strada for these inputs, too, just without the parenthesis. The strada output does not roundtrip - the parens are actually required. And yes, they do make it clear that the output is... questionable. But that's fine, these all have a `The left-hand side of an assignment expression may not be an optional expression` error, which'd be why it's not handled sensibly in the transform - it's completely invalid.

testdata/baselines/reference/submodule/compiler/assertionFunctionsCanNarrowByDiscriminant.js.diff
weswigham: This is a script file by strada logic. (no exports/imports, nothing forcing module mode parsing.) In corsa, the module transforms add `use strict` if needed, but in strada, the `ts` transform does (via `visitLexicalEnvironment` in `visitSourceFile`) so it's done for script files (and I think the module transforms, redundantly, also do). Obviously the module transforms don't touch script files, so they don't add the prologue here.

testdata/baselines/reference/submodule/compiler/optionalChainWithInstantiationExpression2(target=es2019).js.diff
weswigham: Yeah, strada went "oh, this is a cast that became a partially emitted expression, gotta add some parens as we build it", but in corsa, with parens added on print, we're like "oh, we're in an assignment and this is a partially emitted expression of an identifier? Nah, no need for parens".

internal/transformers/estransforms/optionalchain.go
weswigham: More often than not, these flattened chains are single element - it's not often people do a `a?.b?.c?.d` deal. Given that, I think the original is fine for the single or even two element cases. It's not something I'd think needs optimization, especially since reversing a slice isn't great. If anything, you'd do a quick decent to count the chain length, preallocate the slice (probably from a pool, if this was actually hot code), and _then_ insert directly at target position as you descend.

internal/transformers/estransforms/optionalchain.go
weswigham: I don't think that kind of shadowing is a great idea (`left` the node being shadowed by `left` the node data) - don't we have a lint rule forbidding it? Two uses like this kinda aren't enough to bother, anyway, since it's not clarifying or simplifying much of anything. Unless someone shows me the casts having an outsized impact on perf, which is going to make me scream, since old casts were free-and-erased so never really needed perf scrutiny.

jakebailey: All of the calls and casts have cost, I'm afraid

jakebailey: (but really, don't think about it)

internal/transformers/estransforms/optionalchain.go
DanielRosenwasser: True, though it's technically doing a bit more work under the hood.

internal/transformers/estransforms/optionalchain.go
DanielRosenwasser: It just gets a bit unwieldy, and I end up thinking about it :D

DanielRosenwasser (approved):
I haven't looked closely at the test cases, but the porting looks precise enough apart from comments I left.

PR #1395, opened 2025-07-11
Undo my CI caching changes

jakebailey:
My idea failed, for two reasons:

- The go build cache is of course never cleaned, so in the course of a given copilot PR, it would just grow and grow and grow.
- Even though the cache was restored per the CI logs, it didn't make building/testing any faster! Something must be contributing to go build's cache key internally that invalidates all of the saved info.

So, it's moot. Revert it all back in shame. Will have to try again some day when I can think harder about it.

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -8,9 +8,6 @@ inputs:
   cache-name:
     description: Name of scoped cache for this set up.
     default: 'cache'
-  cache-from-previous-run:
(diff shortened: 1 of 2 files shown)


DoctorGester:
Caching in github actions is often quite slow, to the point that restore/save cache overhead can be higher than just not caching anything.

jakebailey:
Normally I'd agree, but if it did work, it would have saved minutes. But, `go build` just didn't use it anyway.

PR #1397, opened 2025-07-11
Port `isInComment`

gabritto:
Port `isInComment`/`getRangeOfEnclosingComment`.
This fixes some completion tests, but some tests now fail because JSDoc completions are unimplemented, and we always return `nil` for completions in a comment position.


internal/ls/format.go
```diff
@@ -123,3 +126,55 @@ func (l *LanguageService) getFormattingEditsAfterKeystroke(
 	}
 	return nil
 }
+
+// Unlike the TS implementation, this function *will not* compute default values for
+// `precedingToken` and `tokenAtPosition`.
+// It is the caller's responsibility to call `astnav.GetTokenAtPosition` to compute a default `tokenAtPosition`,
+// or `astnav.FindPrecedingToken` to compute a default `precedingToken`.
+func getRangeOfEnclosingComment(
```
gabritto, line 134: This function in Strada used to differentiate between a `null` `precedingToken` (which would be left undefined) and an `undefined` `precedingToken` (which would get replaced with a default computed value). We can't easily do this in Go, so, I'm simply not doing anything to the parameters here.

DanielRosenwasser: Maybe a separate `shouldFindPrecedingToken bool`?

internal/ls/format.go
```diff
@@ -123,3 +126,55 @@ func (l *LanguageService) getFormattingEditsAfterKeystroke(
...
+	file *ast.SourceFile,
+	position int,
+	precedingToken *ast.Node,
+	tokenAtPosition *ast.Node,
+) *ast.CommentRange {
+	jsdoc := ast.FindAncestor(tokenAtPosition, func(n *ast.Node) bool {
+		return n.Kind == ast.KindJSDoc
```
DanielRosenwasser: We're probably missing an `IsJSDoc` method, which you could just pass here as (*ast.Node).IsJSDoc`

internal/ls/utilities.go
```diff
@@ -1628,3 +1628,10 @@ func findContainingList(node *ast.Node, file *ast.SourceFile) *ast.NodeList {
 	astnav.VisitEachChildAndJSDoc(node.Parent, file, nodeVisitor)
 	return list
 }
+
+func getLeadingCommentRangesOfNode(node *ast.Node, file *ast.SourceFile) iter.Seq[ast.CommentRange] {
+	if node.Kind == ast.KindJsxText {
+		return nil
+	}
+	return scanner.GetLeadingCommentRanges(&ast.NodeFactory{}, file.Text(), node.Pos())
```
DanielRosenwasser, line 1636: It's so weird that this needs a factory because we only ever really need to look at a single comment range at a time.

internal/fourslash/tests/gen/completionsJSDocNoCrash3_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 func TestCompletionsJSDocNoCrash3(t *testing.T) {
 	t.Parallel()
-
+	t.Skip()
```
DanielRosenwasser, line 14: do we have an easy way to eyeball these `Skip` tests for the future? Do we just assume every `Skip` in fourslash is a bug?

DanielRosenwasser (approved):
Looks good, but I had a few nits. I am curious to hear what others say about needing an allocator for that method though.

internal/ls/format.go
jakebailey: I could have sworn we had that already...

internal/fourslash/tests/gen/completionsJSDocNoCrash3_test.go
gabritto: Some are bugs, some are missing features, e.g. JSDoc tests. The way to find out why they fail is to re-generate the tests without `Skip` and look at the error messages.

internal/ls/format.go
gabritto: I thought we could be avoiding adding it on purpose, to avoid confusion with `isJSDocNode`, but I'll add it.

internal/ls/format.go
gabritto: I kinda dislike the original behavior, and there was only one place which needed it in Strada, so I'd rather leave it like this.

internal/ls/format.go
```diff
@@ -123,3 +126,53 @@ func (l *LanguageService) getFormattingEditsAfterKeystroke(
...
+	// or leading on the latter (and none are in both lists).
+	var trailingRangesOfPreviousToken iter.Seq[ast.CommentRange]
+	if precedingToken != nil {
+		trailingRangesOfPreviousToken = scanner.GetTrailingCommentRanges(&ast.NodeFactory{}, file.Text(), position)
+	}
+	leadingRangesOfNextToken := getLeadingCommentRangesOfNode(tokenAtPosition, file)
+	commentRanges := core.ConcatenateSeq(trailingRangesOfPreviousToken, leadingRangesOfNextToken)
+	for commentRange := range commentRanges {
+		// The end marker of a single-line comment does not include the newline character.
+		// With caret at `^`, in the following case, we are inside a comment (^ denotes the cursor position):
```
DanielRosenwasser:
```suggestion
		// In the following case where the cursor is at `^`, we are inside a comment:
```

internal/ls/format.go
```diff
@@ -123,3 +126,53 @@ func (l *LanguageService) getFormattingEditsAfterKeystroke(
...
+		//
+		//    // asdf   ^\n
+		//
+		// But for closed multi-line comments, we don't want to be inside the comment in the following case:
+		//
+		//    /* asdf */^
+		//
+		// However, unterminated multi-line comments *do* contain their end.
+		//
+		// Internally, we represent the end of the comment at the newline and closing '/', respectively.
```
DanielRosenwasser:
```suggestion
		// Internally, we represent the end of the comment prior to the newline and at the '/', respectively.
		//
		// However, unterminated multi-line comments lack a `/`, end at the end of the file, and *do* contain their end.
```

gabritto:
Hopefully this will be the last reapproval needed. 🙁

DanielRosenwasser (approved):

PR #1398, opened 2025-07-12
lsp: Check diagnostic refresh capabilities against nil values & support more clients

### LINKED ISSUE #1396: lsp(zed): UpdateDiagnostics can end up panicking due to null pointer dereference
Hi,
Thanks for your work on a better TS development experience. In Zed we sometimes see tsgo crashing due to a null pointer dereference on the following line:
https://github.com/microsoft/typescript-go/blob/b0107afcd582b1cb8602790017beee381f998f77/internal/lsp/server.go#L236

While Zed does not fill out this capability at all(or rather it does under certain conditions), the crash reproduces when editing .ts files even with this field always in place. This makes me believe the capabilities as a whole are not available. 
A stack trace (prolly the same as the one over from https://github.com/zed-extensions/tsgo/issues/6): 
```
unknown method workspace/didChangeConfiguration
panic handling request workspace/didChangeWatchedFiles runtime error: invalid memory address or nil pointer dereference goroutine 34 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x64
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1.1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:351 +0x54
panic({0x103143c20?, 0x1037e6ed0?})
	runtime/panic.go:792 +0x124
github.com/microsoft/typescript-go/internal/lsp.(*Server).RefreshDiagnostics(0x14000372000?, {0x1032ea630?, 0x14000098000?})
	github.com/microsoft/typescript-go/internal/lsp/server.go:236 +0x30
github.com/microsoft/typescript-go/internal/project.(*Service).OnWatchedFilesChanged(0x14000370000, {0x1032ea630, 0x14000098000}, {0x1404e786a60, 0x1, 0x100000109a3cb04?})
	github.com/microsoft/typescript-go/internal/project/service.go:340 +0x3bc
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleDidChangeWatchedFiles(...)
	github.com/microsoft/typescript-go/internal/lsp/server.go:634
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x500000010?, {0x1032ea630?, 0x14000098000?}, 0x30?)
	github.com/microsoft/typescript-go/internal/lsp/server.go:472 +0x280
	github.com/microsoft/typescript-go/internal/lsp/server.go:364 +0x70
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop(0x14000000b40, {0x1032ea630?, 0x14000218000?})
	github.com/microsoft/typescript-go/internal/lsp/server.go:380 +0x478
	github.com/microsoft/typescript-go/internal/lsp/server.go:249 +0x28
golang.org/x/sync/errgroup.(*Group).add.func1()
	golang.org/x/sync@v0.15.0/errgroup/errgroup.go:128 +0x88
created by golang.org/x/sync/errgroup.(*Group).add in goroutine 1
	golang.org/x/sync@v0.15.0/errgroup/errgroup.go:96 +0x80
```
My hunch right now is that Zed issues `didChangeConfiguration`, which might somehow mess with capabilities setup. Will try to come up with a more tangible repro.

osiewicz:
Fixes #1396

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -233,14 +233,17 @@ func (s *Server) UnwatchFiles(ctx context.Context, handle project.WatcherHandle)
 
 // RefreshDiagnostics implements project.Client.
 func (s *Server) RefreshDiagnostics(ctx context.Context) error {
-	if s.initializeParams == nil || s.initializeParams.Capabilities == nil || s.initializeParams.Capabilities.Workspace == nil || s.initializeParams.Capabilities.Workspace.Diagnostics == nil {
+	if s.initializeParams.Capabilities == nil ||
+		s.initializeParams.Capabilities.Workspace == nil ||
+		s.initializeParams.Capabilities.Workspace.Diagnostics == nil ||
+		!ptrIsTrue(s.initializeParams.Capabilities.Workspace.Diagnostics.RefreshSupport) {
 		return nil
 	}
-	if ptrIsTrue(s.initializeParams.Capabilities.Workspace.Diagnostics.RefreshSupport) {
-		if _, err := s.sendRequest(ctx, lsproto.MethodWorkspaceDiagnosticRefresh, nil); err != nil {
-			return fmt.Errorf("failed to refresh diagnostics: %w", err)
-		}
+
+	if _, err := s.sendRequest(ctx, lsproto.MethodWorkspaceDiagnosticRefresh, nil); err != nil {
+		return fmt.Errorf("failed to refresh diagnostics: %w", err)
 	}
+
 	return nil
 }
 
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -233,6 +233,9 @@ func (s *Server) UnwatchFiles(ctx context.Context, handle project.WatcherHandle)
 
 // RefreshDiagnostics implements project.Client.
 func (s *Server) RefreshDiagnostics(ctx context.Context) error {
+	if s.initializeParams == nil || s.initializeParams.Capabilities == nil || s.initializeParams.Capabilities.Workspace == nil || s.initializeParams.Capabilities.Workspace.Diagnostics == nil {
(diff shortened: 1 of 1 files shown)


jakebailey:
Is this patching required? You said in the issue that the protocol was fine, so I was expecting a change that just did a nil check.

osiewicz:
No, it's not. I opened this PR before I wrote the comment in question. I'll fix up this PR in a bit.

internal/lsp/server.go
```diff
@@ -233,6 +233,9 @@ func (s *Server) UnwatchFiles(ctx context.Context, handle project.WatcherHandle)
 
 // RefreshDiagnostics implements project.Client.
 func (s *Server) RefreshDiagnostics(ctx context.Context) error {
+	if s.initializeParams == nil || s.initializeParams.Capabilities == nil || s.initializeParams.Capabilities.Workspace == nil || s.initializeParams.Capabilities.Workspace.Diagnostics == nil {
```
jakebailey: I would break this check into multiple lines and then also put the !ptrIsTrue here, making this func have just the one early return.

jakebailey: That and the top level params themselves will never be nil.

osiewicz: Don't we still need an extra early return for:
https://github.com/microsoft/typescript-go/pull/1398/files#diff-0077cc8440363b2137f8e3e405549e59083242b4281d50b08702a6b7153492a8R241
? I guess we could also store `err` aside in a variable and conditionally overwrite it, but eh - is that really better?

jakebailey: I think what you have in the PR now is what I meant.

PR #1399, opened 2025-07-13
fix(1374): support declaration emit for expando functions

### LINKED ISSUE #1374: implicit namespace creation omited from emitted types
tsconfig.json
```
{
    "compilerOptions": {
        "outDir": "./dist",
        "declaration": true,
        "moduleResolution": "node",
        "strict": true,
        "skipLibCheck": true
    }
}
```

a.ts
```
export function Foo() {
    return "Foo";
}

export enum Bar {
    bar
}

Foo.Bar = Bar;
```

tsgo emit output
```
➜ tsgo
➜ cat dist/a.d.ts
export declare function Foo(): string;
export declare enum Bar {
    bar = 0
}
```

tsc emit output
```
➜ tsc
➜ cat dist/a.d.ts
export declare function Foo(): string;
export declare namespace Foo {
    var Bar: typeof import("./a").Bar;
}
export declare enum Bar {
    bar = 0
}
```

a-tarasyuk:
Fixes #1374


internal/checker/emitresolver.go
```diff
@@ -605,8 +605,40 @@ func (r *emitResolver) IsLiteralConstDeclaration(node *ast.Node) bool {
 }
 
 func (r *emitResolver) IsExpandoFunctionDeclaration(node *ast.Node) bool {
-	// node = r.emitContext.ParseNode(node)
-	// !!! TODO: expando function support
+	if !ast.IsParseTreeNode(node) {
```
jakebailey: These newly implemented methods I believe are going to require locking, like the other "public" methods

testdata/baselines/reference/submodule/compiler/declarationEmitDefaultExportWithStaticAssignment.js.diff
```diff
@@ -18,37 +18,40 @@
...
  }
  //// [index1.d.ts]
--declare function Example(): void;
+ declare function Example(): void;
 -declare namespace Example {
--    var Foo: typeof import("./foo").Foo;
--}
--export default Example;
-+export default function Example(): void;
++declare module Example {
```
jakebailey: Seems to be generating the wrong keyword here. Should be namespace.

testdata/baselines/reference/submodule/compiler/declarationEmitLateBoundAssignments.js.diff
```diff
@@ -1,10 +1,7 @@
...
--export declare namespace foo {
--    var bar: number;
--    var strMemName: string;
--}
\ No newline at end of file
+@@= skipped -39, +39 lines =@@
+     var bar: number;
+     var strMemName: string;
+ }
++export {};
```
jakebailey: Hm. Wonder why these are appearing so frequently in this change.

a-tarasyuk: @jakebailey Looks like these empty exports are related to this typo

https://github.com/microsoft/typescript-go/blob/2b82831a05b6b99da279d12fecbcbc460574a85b/internal/transformers/declarations/transform.go#L268-L270

```go
ast.IsExternalModuleIndicator(elem)
```

testdata/baselines/reference/submodule/compiler/declarationEmitFunctionKeywordProp.js.diff
```diff
@@ -1,21 +1,12 @@
...
--    var normal: boolean;
--}
+@@= skipped -35, +35 lines =@@
+ }
  declare function baz(): void;
--declare namespace baz {
+ declare namespace baz {
 -    var _a: boolean;
--    export var normal: boolean;
++    var _b: boolean;
```
jakebailey: It's interesting that this temp variable turns into `_b` and not `_a`; it's not incorrect but I wonder if something isn't being cleared between expandos or something.

(I am reading this on my phone so have not actually read the code yet)

testdata/baselines/reference/submodule/compiler/declarationEmitLateBoundJSAssignments.js.diff
```diff
@@ -8,5 +8,8 @@
 -export namespace foo {
 -    let bar: number;
 -    let strMemName: string;
--}
-+export declare function foo(): void;
\ No newline at end of file
++export declare function foo(): void;
```
jakebailey, line 11: Maybe not for this PR but I wonder what the leftover code difference is that's adding "declare" to these.

testdata/baselines/reference/submodule/compiler/jsxDeclarationsWithEsModuleInteropNoCrash.js.diff
```diff
@@ -21,20 +22,39 @@
 -    export { bar_1 as bar };
 -}
 -import PropTypes from 'prop-types';
++    var propTypes: {
```
jakebailey, line 24: Seems to be missing the exports here.

weswigham (commented):
This looks like a pretty straight port of the logic from strada - @sandersn should weigh in on weather he thinks that's right. I do not recall if these is one of the constructs that changed meaningfully in the js reparser rewrite and, thus, should also have it's declaration emit logic adjusted, too.

internal/ast/utilities.go
```diff
@@ -3633,3 +3637,79 @@ func GetSemanticJsxChildren(children []*JsxChild) []*JsxChild {
...
+		if initializer.OperatorToken.Kind == KindBarBarToken || initializer.OperatorToken.Kind == KindQuestionQuestionToken {
+			if node.Name() != nil && IsEntityNameExpressionEx(node.Name(), IsInJSFile(node)) && IsSameEntityName(node.Name(), initializer.Left) {
+				return initializer.Right
+			}
+		}
+	}
+	return node.Initializer()
+}
+
+func GetDeclaredExpandoInitializer(node *Node) *Expression {
```
weswigham: In particular, this suddenly needing to be implemented is suspect. It used to be used all over in the checker in the old expando checking paths in checker, but now it isn't at all and was only added for the declaration emit - it implies we're doing something wrong and ignoring the new parser output for expandos.

sandersn:
Expandos didn't fundamentally change, but they are losing a lot of their complicated features. So the core changes to emit might be right-ish, but you won't need to reintroduce any code in the parser or binder, I think.

internal/ast/utilities.go
```diff
@@ -3633,3 +3637,79 @@ func GetSemanticJsxChildren(children []*JsxChild) []*JsxChild {
 		}
 	})
 }
+
+func IsExpandoPropertyDeclaration(node *Node) bool {
```
sandersn: the binder already has to detect these patterns, so you shouldn't need to port anything new.

edit: your explanation of why these are needed makes sense, but much of it can be simplified given how much less Corsa supports.

a-tarasyuk:
@sandersn Thanks for the clarification.

internal/ast/utilities.go
```diff
@@ -3633,3 +3637,79 @@ func GetSemanticJsxChildren(children []*JsxChild) []*JsxChild {
...
+	if node == nil {
+		return false
+	}
+	return IsPropertyAccessExpression(node) || IsElementAccessExpression(node) || IsBinaryExpression(node)
```
sandersn: expando declarations should only be binary expressions now.

```js
/** @type {string} */
f.p
```

is no longer supported by Corsa, and access expressions don't have a Symbol field.

internal/ast/utilities.go
```diff
@@ -3633,3 +3637,79 @@ func GetSemanticJsxChildren(children []*JsxChild) []*JsxChild {
...
+}
+
+func GetExpandoInitializer(initializer *Node, isPrototypeAssignment bool) *Node {
```
sandersn: prototype assignments aren't supported in Corsa 
(`C.prototype = { ... }`)

internal/ast/utilities.go
```diff
@@ -3633,3 +3637,79 @@ func GetSemanticJsxChildren(children []*JsxChild) []*JsxChild {
...
+		return initializer
+	}
+
+	return nil
+}
+
+func GetEffectiveInitializer(node *Node) *Expression {
+	if IsInJSFile(node) && node.Initializer() != nil && IsBinaryExpression(node.Initializer()) {
+		initializer := node.Initializer().AsBinaryExpression()
+		if initializer.OperatorToken.Kind == KindBarBarToken || initializer.OperatorToken.Kind == KindQuestionQuestionToken {
```
sandersn: defaulting assignments with `x.p = x.p || function() {}` are not supported in Corsa.

jakebailey:
This needs a merge from main again since one of the baselines changed via another PR.

a-tarasyuk:
@jakebailey Thanks for pointing it out. I’ve merged main

testdata/baselines/reference/submodule/compiler/declarationEmitExpandoPropertyPrivateName.js.diff
```diff
@@ -14,4 +14,7 @@
  export declare function f(): I;
  export {};
 +//// [b.d.ts]
```
jakebailey, line 16: It's super odd that the original code did not emit a declaration file here, not sure why (the new code looks correct).

weswigham (commented):
This is still pretty semantic in how it operates - the binder is *much* more syntactic now, and I think we can be more syntactic in the declaration emitter now, too, because of it. Specifically, I'd like to avoid needing a `GetPropertiesOfContainerFunction` emit resolver function. I'd rather we visited expression statements, looking for assignment expressions that are expando assignments (`ast.GetAssignmentDeclarationKind` result of `Property` - I think the js reparser even sets a `.Type` on these now!), and translate them to the equivalent expansion, instead of seeing the function declaration and semantically checking for extra properties to patch on. So, for example, we see the AST for
```js
export function foo() { return 1 }
/** @type {"ok"} */
foo.bar = "ok"
```
we visit `export function foo() { return 1 }` and emit `export function foo(): number` and then we visit
```js
/** @type {"ok"} */
foo.bar = "ok"
```
and emit 
```ts
export namespace foo { export const bar: "ok" }
```
Notably, multiple expandos would emit as multiple namespaces:
```ts
export namespace foo { export const bar: "ok" }
export namespace foo { export const baz: "yes" }
```
by default, and then you can have a "collection" pass at the end of a scope that merges all the newly added namespaces together (or into an explicit one, if it's there and the names all work out). The only awkward thing I can see here is that the `export` modifier-ness of the namespace declaration relies on nonlocal information (if the host function is exported). This could be alleviated by _never_ exporting a function with expando properties (which is still a nonlocal semantic check, unfortunately...) via an `export` modifier, and instead adding a `export { foo }` declaration - then all the declarations can always be non-exported (which is roughly how the old js declaration emitter in the node builder would have done it, and then had an extra pass to "inline" the export declarations as modifiers where possible).

a-tarasyuk:
@weswigham, thanks for the feedback. I've started implementing this approach and ran into a few edge cases that are worth considering:

1. Default exports. When a function with expandos is a default export:

```ts
export default function foo() {}
foo.meta = 123;
```

We can't emit:

```ts
export default function foo(): void;
export namespace foo { export const meta: 123; }
```

As part of this, I’ve added a temporary solution using _`pendingDefaultExport`_, which defers the _default export_ until after all expandos are emitted. However, this slightly changes the behavior for _export default function_ declarations without expandos — they're now emitted in split form rather than as a single _default-exported_ function.

2. Non-exported hosts

For:

```ts
function foo() {}
foo.extra = true;
```

We cannot emit:

```ts
export namespace foo { const extra: true; }
```

At this stage, it's necessary to determine whether the host is exported, to correctly apply or omit the _export_ modifier from the merged namespace declaration

3. Type resolution scope 

To correctly serialize the type of an expando property and avoid self-references,
```ts
declare namespace Example {
    const Bar: typeof Bar;
}
```

The type resolver typically relies on the `enclosingDeclaration` to determine symbol visibility and to avoid self-references during serialization. This means that information about the surrounding namespace scope should be available and correctly set during type serialization to ensure accurate and consistent type output.

@weswigham WDYT?

weswigham:
> However, this slightly changes the behavior for export default function declarations without expandos — they're now emitted in split form rather than as a single default-exported function.

In the old JS nodebuilder logic (unported) we had a late pass that looked to see if we had a `export default x` where `x` was something(s) we could just slam a `default` modifier onto to elide the statement - I think we should be able to manage the same here. This introduces a change in the other direction (an explicit `export default x` can get elided in favor of a default modifier on `x`) unless we're careful with how the logic's invoked.

> Non-exported hosts

Yeah, this is something I think we could best handle with a two-pass system, too. One pass translates the core of everything and notes what names had at least one `export` modifier, and another which adds the missing `export` modifiers on all the matching names. Again, this is basically what the old JS nodebuilder would always do to handle weird merges like this. Actually, handling the `default` modifier in the same way is probably smart, rather than doing the extra-statement-and-possible-inline I proposed earlier; but really anything that does a bit of cross-declaration collation probably works. You can probably "cheat" a bit and use `resolveName` to get the export/default status of the enclosing namespace from the symbol, too, if you really wanted, since the binder makes that information available to locally merged symbols like this. While I'm trying to avoid poking holes into the checker's API, the `referenceResolver` is already split from the checker, and node binding is now eager instead of delayed - so binder data should be considered generally available where needed.

> This means that information about the surrounding namespace scope should be available and correctly set during type serialization to ensure accurate and consistent type output.

Yes. Historically, we've done this by manufacturing a block/namespace declaration, attaching the relevant symbols/symbol tables/parent pointers to it, then using it as the enclosing declaration we pass to the serializer. Presumably you're wondering how we should collect the list of symbols to attach as members to that namespace, as `getPropertiesOfContainerFunction` is how we managed that before, which is... fair. In an ideal world, just attaching the symbol for the member we're serializing as an export would be enough, but unfortunately with how namespace scope merging works (nonlexically!) it isn't. Buuuuuuut! With how the binder works now, all the "expando" exports should (eagerly) be bound into the original symbol, like any other member (vs late merged in the checker) - so simple name resolution is enough, and you aughta just be able to copy the symbol's exports to the namespace's table of locals.

Do note, all this is basically predicated on passing in the `referenceResolver` to the transform to actually resolve the symbol names within the file via `resolveName`.

This looks on its' face similar to what was done before, which is good, but avoids ever invoking `getTypeOfSymbol` and all the work that can chain to - it just looks at the binder's output, which should be better for parallel/no-check emit performance, and favors any existing annotation nodes (and declaration/comment order) vs always manufacturing them whole-cloth or relying on nonlocal node copying in the node builder (which isn't currently implemented).

testdata/baselines/reference/submodule/compiler/declarationEmitExpandoWithGenericConstraint.js.diff
```diff
@@ -7,4 +7,47 @@
...
++!!! error TS2395: Individual declarations in merged declaration 'Point' must be all exported or all local.
++                         ~~~~~
++!!! error TS2451: Cannot redeclare block-scoped variable 'Point'.
++        (x: number, y: number): Point;
++        zero: () => Point;
++    };
++    export declare const Rect: <p extends Point>(a: p, b: p) => Rect<p>;
++    declare namespace Point {
++                      ~~~~~
++!!! error TS2395: Individual declarations in merged declaration 'Point' must be all exported or all local.
```
jakebailey: Seems like this namespace is missing an export.

testdata/baselines/reference/submodule/compiler/declarationEmitFunctionDuplicateNamespace.js.diff
```diff
@@ -1,9 +1,9 @@
...
- //// [declarationEmitFunctionDuplicateNamespace.d.ts]
+@@= skipped -20, +20 lines =@@
  declare function f(a: 0): 0;
  declare function f(a: 1): 1;
--declare namespace f {
+ declare namespace f {
 -    var x: number;
--}
\ No newline at end of file
++    const x: 2;
```
jakebailey: Hm, wonder why this widening behavior is wrong?

testdata/baselines/reference/submodule/compiler/declarationEmitFunctionKeywordProp.js.diff
```diff
@@ -1,21 +1,31 @@
...
+@@= skipped -25, +25 lines =@@
  //// [declarationEmitFunctionKeywordProp.d.ts]
  declare function foo(): void;
--declare namespace foo {
+ declare namespace foo {
 -    var _a: boolean;
 -    export { _a as null };
--}
++    const null_1: true;
++    export { null_1 as null };
```
jakebailey: Interesting that the temp name changed here. Not that it much matters.

testdata/baselines/reference/submodule/compiler/declarationEmitLateBoundAssignments.js.diff
```diff
@@ -1,10 +1,10 @@
...
+@@= skipped -36, +36 lines =@@
  //// [declarationEmitLateBoundAssignments.d.ts]
  export declare function foo(): void;
--export declare namespace foo {
+ export declare namespace foo {
 -    var bar: number;
 -    var strMemName: string;
--}
\ No newline at end of file
++    const bar: 12;
```
jakebailey: Hm, weird widening change here too...

jakebailey: (not to mention, `const` vs `var`)

testdata/baselines/reference/submodule/compiler/declarationEmitExpandoPropertyPrivateName.js.diff
weswigham: Certain kinds of errors blocked declaration emit before (notably: declaration emit errors). That doesn't seem to be the case anymore.

testdata/baselines/reference/submodule/compiler/isolatedDeclarationErrorsExpandoFunctions.js
```diff
@@ -25,3 +25,61 @@ foo.length = 10;
...
+    export declare namespace foo {
+        const toString: () => void;
+    }
+    export declare namespace foo {
+        const length: number;
+              ~~~~~~
+!!! error TS2451: Cannot redeclare block-scoped variable 'length'.
+    }
+    export declare namespace foo {
+        const length: number;
```
weswigham: Hm. This would benefit from being `var` instead of `const`... @sandersn thoughts? Multiple assignments produce multiple declarations - those only merge if they're `var`.

a-tarasyuk: @weswigham, thanks for the feedback. I've changed to use `var`

PR #1404, opened 2025-07-15
Completions fixes

gabritto:
In this PR:
- Fix for the test conversion script when `verify.completions` does not set any expected completion result.
- Fix for a crash when getting the text of a JSX tag name expression.
- Fix for a crash in `setMemberDeclaredBySpreadAssignment` due to missing `nil` check.
- Fix for parsing test options (specifically `allowNonTSExtension` which wasn't being set correctly in fourslash tests).


internal/testutil/harnessutil/harnessutil.go
```diff
@@ -290,19 +290,35 @@ func setOptionsFromTestConfig(t *testing.T, testConfig TestConfiguration, compil
...
 		}
 
 		t.Fatalf("Unknown compiler option '%s'.", name)
 	}
 }
 
-var harnessCommandLineOptions = []*tsoptions.CommandLineOption{
-	{
+var compilerOptions = append(
+	tsoptions.OptionsDeclarations,
```
jakebailey, line 302: This should be a concat, since append may mutate

PR #1405, opened 2025-07-15
Improve go-to-definition and implement go-to-type-definition

ahejlsberg:
This PR implements

* Go-to-def on `override` keyword selects the overridden method declaration.
* Go-to-def on label in `break` or `continue` statement selects the label definition.
* Go-to-def on `case` or `default` keyword selects the containing `switch` statement.
* Go-to-def on `return`, `yield` or `await` keyword selects the containing function declaration.
* Go-to-def on a call expression selects the called function or method following overload resolution.
* Go-to-def on a property in an object literal with a contextual type selects the declaration in the contextual type.
* Go-to-def on a property with no explicit declaration selects the index signature(s) for the property.
* Go-to-type-definition on a variable or property selects the associated type(s).

### DIFF
--- a/internal/checker/services.go
+++ b/internal/checker/services.go
@@ -618,3 +618,67 @@ func (c *Checker) GetTypeParameterAtPosition(s *Signature, pos int) *Type {
 	}
 	return t
 }
+
+func (c *Checker) GetContextualDeclarationsForObjectLiteralElement(objectLiteral *ast.Node, name string) []*ast.Node {
+	var result []*ast.Node
+	if t := c.getApparentTypeOfContextualType(objectLiteral, ContextFlagsNone); t != nil {
+		for _, t := range t.Distributed() {
+			prop := c.getPropertyOfType(t, name)
+			if prop != nil {
+				for _, declaration := range prop.Declarations {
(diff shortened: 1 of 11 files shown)


internal/checker/services.go
```diff
@@ -618,3 +618,67 @@ func (c *Checker) GetTypeParameterAtPosition(s *Signature, pos int) *Type {
...
+	"WeakSet":          {},
+	"ReadonlySet":      {},
+	"Map":              {},
+	"WeakMap":          {},
+	"ReadonlyMap":      {},
+	"Partial":          {},
+	"Required":         {},
+	"Readonly":         {},
+	"Pick":             {},
+	"Omit":             {},
```
RyanCavanaugh, line 662:
```suggestion
	"Omit":             {},
	"Exclude":          {},
```

ahejlsberg: The types in this list generally operate on object types for which we record source locations. `Exclude` and `Extract` are almost always used with (unions of) literal types, and we don't record source locations for those.

ahejlsberg: However, we should include `NonNullable<T>`.

RyanCavanaugh (approved):

PR #1412, opened 2025-07-17
Deep clone JSDoc

sandersn:
This PR switches the approach of JSDoc reparsing from copying pointers plus adding alternate parents to deep-cloning nodes and keeping a map of original nodes to clones. However, this PR does not include the map yet; it's not needed in the checker and the jsdoc support in the language service doesn't work correctly either way (as measured by our tests at least).

Notably, I couldn't figure out how to set parent pointers during the deep clone itself, due to the visitor structure. I think the cost of two passes is likely to be small given the number of jsdoc nodes that will need to be cloned. But suggestions are welcome for how to incorporate parent-setting into the deep clone.

The code is slightly more verbose than it could be, with individual DeepCloneNode calls for all children. I could have instead cloned the entire node at the end of the process, but this is more efficient for hosted tags that would be cloning all children of, say, a function, which would end up cloning the entire body: a possibly-large subtree. I also think that the explicit child clones convey the intent more obviously.


internal/testutil/tsbaseline/type_symbol_baseline.go
```diff
@@ -327,8 +327,9 @@ func forEachASTNode(node *ast.Node) []*ast.Node {
 	for len(work) > 0 {
 		elem := work[len(work)-1]
 		work = work[:len(work)-1]
-		if elem.Flags&ast.NodeFlagsReparsed == 0 || elem.Kind == ast.KindAsExpression || elem.Kind == ast.KindSatisfiesExpression {
-			if elem.Flags&ast.NodeFlagsReparsed == 0 {
+		if elem.Flags&ast.NodeFlagsReparsed == 0 || elem.Kind == ast.KindAsExpression || elem.Kind == ast.KindSatisfiesExpression ||
```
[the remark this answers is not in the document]
sandersn: I guess, but it's also in test code, and is an ad-hoc condition that will only ever appear here. The mess makes that obvious; extracting to a function would make readers think that we might want to use it elsewhere.

I'm on the fence. Any humans want to weigh in?

internal/parser/reparser.go
```diff
@@ -5,7 +5,9 @@ import (
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
-func (p *Parser) finishReparsedNode(node *ast.Node) {
+func (p *Parser) finishReparsedNode(node *ast.Node, locationNode *ast.Node) {
+	node.Flags = p.contextFlags | ast.NodeFlagsReparsed
```
sandersn, line 9: This change is responsible for a lot of the diff, but that's because it dedupes a lot of code.

internal/parser/reparser.go
```diff
@@ -203,26 +196,23 @@ func (p *Parser) reparseJSDocTypeLiteral(t *ast.TypeNode) *ast.Node {
...
-		p.finishReparsedNode(t)
 	}
-	return t
+	return p.factory.DeepCloneReparse(t)
 }
 
-func (p *Parser) gatherTypeParameters(j *ast.Node, tagWithTypeParameters *ast.Node, host *ast.Node) *ast.NodeList {
-	typeParameters := p.nodeSlicePool.NewSlice(0)
+func (p *Parser) gatherTypeParameters(j *ast.Node, tagWithTypeParameters *ast.Node) *ast.NodeList {
+	var typeParameters []*ast.Node
```
sandersn, line 220: I don't know why I didn't lazy-initialise this before, because the common case is to have no parameters, which previously called `NewSlice(0)` each time anyway.

internal/parser/reparser.go
```diff
@@ -284,103 +276,118 @@ func (p *Parser) reparseHosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Node)
 		case ast.KindVariableStatement:
 			if parent.AsVariableStatement().DeclarationList != nil {
 				for _, declaration := range parent.AsVariableStatement().DeclarationList.AsVariableDeclarationList().Declarations.Nodes {
-					if declaration.AsVariableDeclaration().Type == nil {
-						declaration.AsVariableDeclaration().Type = setHost(tag.AsJSDocTypeTag().TypeExpression, declaration)
-						p.finishReparsedNode(declaration)
+					if declaration.AsVariableDeclaration().Type == nil && tag.AsJSDocTypeTag().TypeExpression != nil {
+						declaration.AsVariableDeclaration().Type = p.factory.DeepCloneReparse(tag.AsJSDocTypeTag().TypeExpression.Type())
+						p.overrideParentInImmediateChildren(declaration)
```
sandersn: Since finishReparsedNode now sets location and flags, this just (re)sets parent pointers of the existing node's children. I think this makes more sense since the variable declaration is not a reparsed node, just its type annotation.

internal/parser/reparser.go
```diff
@@ -562,31 +566,16 @@ func (p *Parser) makeNewCast(t *ast.TypeNode, e *ast.Node, isAssertion bool) *as
...
-	}
-	t := typeExpression.Type()
-	t.Flags |= ast.NodeFlagsReparsed
-	return t
-}
-
 func getClassLikeData(parent *ast.Node) *ast.ClassLikeBase {
 	var class *ast.ClassLikeBase
-	if parent.Kind == ast.KindClassDeclaration {
+	switch parent.Kind {
```
sandersn, line 579: gopls or some tool wouldn't be quiet until I accepted this change. I guess it's better.

internal/parser/reparser.go
```diff
@@ -460,26 +467,23 @@ func (p *Parser) reparseHosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Node)
 				if implementsClause := core.Find(class.HeritageClauses.Nodes, func(node *ast.Node) bool {
 					return node.AsHeritageClause().Token == ast.KindImplementsKeyword
 				}); implementsClause != nil {
-					implementsClause.AsHeritageClause().Types.Nodes = append(implementsClause.AsHeritageClause().Types.Nodes, implementsTag.ClassName)
-					p.finishReparsedNode(implementsClause)
+					implementsClause.AsHeritageClause().Types.Nodes = append(implementsClause.AsHeritageClause().Types.Nodes, p.factory.DeepCloneReparse(implementsTag.ClassName))
+					p.overrideParentInImmediateChildren(implementsClause)
 					return
 				}
 			}
```
sandersn, line 478: setting `implementsTag.ClassName.Flags |= ast.NodeFlagsReparsed` was a simple mistake as far as I can tell.

testdata/baselines/reference/fourslash/findAllRef/FindAllReferencesJsRequireDestructuring.baseline.jsonc
```diff
@@ -2,10 +2,3 @@
 // === /bar.js ===
 
 // const { /*FIND ALL REFS*/[|foo|]: bar } = require('./foo');
```
sandersn, line 4: fourslash results are already very spotty and incomplete; commonjs and typedef `@property` tags both seem to worse, however. These were likely working as well as they did because the shared tree structure meant that a symbol would be available on the jsdoc-embedded node. With this PR, no JSDoc node has a symbol; they're all on the clones.

testdata/baselines/reference/submodule/compiler/jsExportAssignmentNonMutableLocation.symbols
```diff
@@ -18,6 +18,5 @@ module.exports = {
 
 exports.customSymbol2 = Symbol("custom");
 >exports : Symbol("file", Decl(file.js, 0, 0))
```
sandersn, line 20: conflicting `module.exports=` and individual `module.exports.p=` declarations now result in no binding for the individual exports. The commonjs one wins completely.

testdata/baselines/reference/submoduleAccepted/compiler/jsExportAssignmentNonMutableLocation.types.diff
```diff
@@ -27,7 +27,7 @@
 +>exports.customSymbol2 = Symbol("custom") : symbol
 +>exports.customSymbol2 : any
 +>exports : typeof import("./file")
-+>customSymbol2 : symbol
++>customSymbol2 : any
```
sandersn, line 30: no binding for these conflicting symbols also means no types.

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsExportAssignedClassExpression.types.diff
```diff
@@ -10,12 +10,12 @@
 ->exports : typeof import("index")
 ->class Thing {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import("index")
 ->Thing : typeof import("index")
-+>module.exports = class Thing {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import(".")
-+>module.exports : typeof import(".")
-+>module : { Thing: typeof import("."); }
-+>exports : typeof import(".")
-+>class Thing {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import(".")
-+>Thing : typeof import(".")
++>module.exports = class Thing {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof Thing
```
sandersn, line 13: I'm not sure why this type printing changed, but it seems like an improvement.

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsExportAssignedClassExpressionAnonymous.types.diff
```diff
@@ -9,11 +9,11 @@
 ->module : { exports: typeof import("index"); }
 ->exports : typeof import("index")
 ->class {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import("index")
-+>module.exports = class {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import(".")
-+>module.exports : typeof import(".")
-+>module : { "\uFFFDclass": typeof import("."); }
-+>exports : typeof import(".")
-+>class {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof import(".")
++>module.exports = class {    /**     * @param {number} p     */    constructor(p) {        this.t = 12 + p;    }} : typeof exports
++>module.exports : typeof (Anonymous class)
```
sandersn, line 13: same, except arguably `import(".")` is better than `(Anonymous class)` since it's unique.

testdata/baselines/reference/submoduleAccepted/conformance/jsdocTypeReferenceToImportOfClassExpression.types.diff
```diff
@@ -24,18 +24,18 @@
     */
    constructor(compiler) {
 ->compiler : import("MC")
-+>compiler : import("./MC")
++>compiler : MC
```
sandersn, line 27: this is similar, except here `MC` is the name of the type alias and `import("./MC")` is its value.

testdata/baselines/reference/submoduleAccepted/conformance/moduleExportWithExportPropertyAssignment3.types.diff
```diff
@@ -75,7 +75,7 @@
 +>module.exports : { justExport: number; bothBefore: number; bothAfter: number; }
 +>module : { "export=": { justExport: number; bothBefore: number; bothAfter: number; }; }
 +>exports : { justExport: number; bothBefore: number; bothAfter: number; }
-+>bothBefore : "string"
++>bothBefore : number
```
sandersn, line 78: Corsa dropped the combination of `module.exports.both=1` with `module.exports={ both: "string" }` -- now that the `module.exports=` always wins, it always provides the type. Before, the property would sometimes win (when it comes first?).

testdata/baselines/reference/submoduleAccepted/conformance/moduleExportsElementAccessAssignment.types.diff
```diff
@@ -80,8 +75,7 @@
 +>module["exports"] : typeof import("./mod1")
 +>module : { "\"mod1\"": typeof import("./mod1"); }
  >"exports" : "exports"
-->"d" : "d"
-+>"d" : {}
+ >"d" : "d"
```
sandersn, line 78: I don't understand this change, but it's the way Strada works. Maybe it's checking the literal itself instead of the property with the literal name.

internal/parser/reparser.go
```diff
@@ -407,14 +414,14 @@ func (p *Parser) reparseHosted(tag *ast.Node, parent *ast.Node, jsDoc *ast.Node)
 				}
 
 				fun.FunctionLikeData().Parameters = p.newNodeList(thisParam.Loc, newParams)
-				p.finishReparsedNode(fun)
+				p.overrideParentInImmediateChildren(fun)
```
weswigham: nit: maybe rather than using `overrideParentInImmediateChildren` directly on these node mutations, it'd be more semantically clear to call something like `finishMutatedNode`, which in turn calls this. That way it's a bit easier to go "oh, this result hasn't been finished properly in all code paths" at a glance.

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsExportAssignedClassExpression.types.diff
weswigham: `Thing` is in scope at the return position now because the `.Parent` chain isn't inside the jsdoc comment anymore, so it's resolving at the actual export now. I think.

Or we're mistakenly binding `Thing` into the local scope. One of those.

weswigham (approved):

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsExportAssignedClassExpression.types.diff
sandersn: Sounds like the first then, because the only weirdness with binding `module.exports=` compared to `export=` is that the former always binds with the source file as a container, even if it's inside some other container. Which doesn't apply here.

jakebailey (approved):

testdata/baselines/reference/submoduleAccepted/conformance/jsDeclarationsExportAssignedClassExpression.types.diff
sandersn: I checked this test by changing it to use `export=`, and that result is the same as the new result, so this is an improvement inasmuch as it makes JS more like TS.

PR #1414, opened 2025-07-17
Simplify `filterText` computation + fixes

gabritto:
This PR simplifies the computation of a completion item's `filterText` which was ported from the VSCode TS extension. There were originally two pieces of code where `filterText` was computed, and I think one of those was mostly unnecessary, so I removed it and simplified the rest of the code.

The code I removed used to set `filterText` for all dot completions, which wasn't needed, so the Corsa tests have been updated to reflect that, and a bunch of tests ported from Strada now pass as well because we no longer set `filterText` in most cases.

I also added tests for all the cases where I think `filterText` is necessary, since this piece of code is new and so not covered by Strada tests.

This PR also fixes two porting bugs, one in the parser and one bug in the scanner, and some minor completion bugs.


internal/ls/completions.go
```diff
@@ -4137,40 +4157,10 @@ func (l *LanguageService) createLSPCompletionItem(
 	// Filter text
 
 	// Ported from vscode ts extension.
-	wordRange, wordStart := getWordRange(file, position)
+	wordSize, wordStart := getWordRange(file, position)
+	dotAccessor := getDotAccessor(file, position-wordSize)
 	if filterText == "" {
-		filterText = getFilterText(file, position, insertText, name, isMemberCompletion, isSnippet, wordStart)
-	}
-	if isMemberCompletion && !isSnippet {
```
gabritto, line 4144: This code here was removed because it mostly overlapped with `getFilterText` above.

testdata/baselines/reference/submodule/conformance/privateNameHashCharName.js
```diff
@@ -13,8 +13,10 @@ class C {
 
 
 //// [privateNameHashCharName.js]
+#;
 class C {
+    #;
     m() {
-        this.;
```
gabritto, line 18: This results from fixing the scanner bug.

internal/ls/completions.go
```diff
@@ -2052,15 +2052,15 @@ func isRecommendedCompletionMatch(localSymbol *ast.Symbol, recommendedCompletion
...
 var wordSeparators = collections.NewSetFromItems(
-	'`', '~', '!', '@', '#', '$', '%', '^', '&', '*', '(', ')', '-', '=', '+', '[', '{', ']', '}', '\\', '|',
+	'`', '~', '!', '@', '%', '^', '&', '*', '(', ')', '-', '=', '+', '[', '{', ']', '}', '\\', '|',
 	';', ':', '\'', '"', ',', '.', '<', '>', '/', '?',
 )
 
 // Finds the range of the word that ends at the given position.
 // e.g. for "abc def.ghi|jkl", the word range is "ghi" and the word start is 'g'.
-func getWordRange(sourceFile *ast.SourceFile, position int) (wordRange *core.TextRange, wordStart rune) {
+func getWordRange(sourceFile *ast.SourceFile, position int) (wordSize int, wordStart rune) {
```
DanielRosenwasser: The naming and comment still talks about "ranges". I might also use "length" instead of "size".

internal/ls/completions.go
```diff
@@ -2127,41 +2143,45 @@ func getFilterText(
 	// In which case we want to insert a bracket accessor but should use `.abc` as the filter text instead of
```
DanielRosenwasser, line 2143: This is sort of an aside, but I don't understand why conceptually the "access operator" (the `.` or `?.`) is part of the filter text at all.

internal/scanner/scanner.go
```diff
@@ -863,12 +863,11 @@ func (s *Scanner) Scan() ast.Kind {
...
 			}
-			if s.scanIdentifier(1) {
-				s.token = ast.KindPrivateIdentifier
-			} else {
+			if !s.scanIdentifier(1) {
 				s.errorAt(diagnostics.Invalid_character, s.pos-1, 1)
-				s.token = ast.KindUnknown
+				s.tokenValue = "#"
 			}
+			s.token = ast.KindPrivateIdentifier
```
DanielRosenwasser, line 870: Seems fine, but why did we need this? Or, why is Strada able to get away without this?

DanielRosenwasser (commented):
Seems okay, just wanted to clarify a few things before approving.

internal/ls/completions.go
gabritto: The client will rank the completion items by comparing the filter text with some suffix of the existing test at the completions position. The suffix is usually the word touching the position, but may be a custom range when the item specifies a text edit.
Given that, we want to use the filter text to avoid the client overly de-ranking or omitting completions that have to modify some of the text that the user already typed.
For instance, if we have this:
```ts
declare const obj: { "ab c": string };
obj.a|
```
The completion item for `ab c` will need to insert `["ab c"]`, and it will specify a range in the text edit covering `.a`, but we want it to be ranked by the client the same as a regular property should, so the client should compare `.a` (text at current position that the item will replace) with `.ab c`, instead of `["ab c"]`, because `["ab c"]` doesn't really match `.a`.

The access operator is there because in the situations where we use it in the filter text, it is part of the text that will be replaced if you accept the completion item, so it is part of what will be compared with the filter text.
That said, it's possible there are some cases where we currently set the filter text that could be optimized away.

internal/scanner/scanner.go
gabritto: The new code does what Strada does. The code before this PR seems like a porting bug.

internal/ls/completions.go
DanielRosenwasser: I see, so this is entirely because the filter text is matched up against text in the replacement range. If there was a distinct way to send back what we calculated as the "effective search text", we wouldn't have to hack around with the operator itself.

Funny enough, you can see how this sort of breaks down if you insert a newline or comment after a dot.

image

PR #1426, opened 2025-07-21
Accept and document jsdoc diffs, round 1

sandersn:
I started adding precise differences for JS semantics to CHANGES.md. As you can see, I accepted 25 tests--and looked at around 40--before I got bored and switched to compiling well-known JSDoc codebases (webpack and svelte). The changes include a couple of diffs from Svelte as well.

I'm not 100% sure that this is worth generating. At some point we may instead want to declare legacy test bankruptcy, if people appear to be using JS features with no complaint. However, the first 2.5% of accepted tests probably has 25% of the semantics changes, so this list is at least worth reviewing by the team for surprises.

A good number of the diffs are due to the reparsing strategy, which gives JS *exactly* the semantics of TS. The rest are due to cut features, both whole features (eg type annotations on functions) and partial (eg use of `@class` to mark an otherwise-normal function as a constructor).


CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+}
+```
+
+will now have `x?: number` not `x?: number | undefined` with `strictNullChecks` off.
+Regardless of strictness, it still makes parameters optional when used in a `@param` tag.
+
+
+### JSDoc Tags
+
+1. `@type` tags no longer apply to function declarations, and now contextually type function expressions instead of applying directly. So this annotation no longer does anything:
```
jakebailey: I still really really think we should reconsider this and come up with some sort of way to do it

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
 
-1. `@class`
+1. `@class`/`@constructor`
 2. `@throws`
 3. `@author`
 4. `@enum`
+
+## Checker
+
+1. When `"strict": false`, Corsa no longer allows omitting arguments for parameters with type `undefined`, `unknown`, or `any`:
```
jakebailey: Can you use `###` sections for each of these instead of numbers? I don't think these render as lists or anything.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+```
+
+Must now be written as
+
+```js
+/** @param {...number} ns */
+function sum(...ns) {}
+```
+
+3. The postfix `=` type no longer adds `undefined` even when `strictNullChecks` is off:
```
jakebailey: Isn't this one just visual? Like, in a parameter position, missing vs undefined is not a real thing that matters.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+/** @type {AssertFunc} */
+const assert = check => {
+    if (!check) throw new Error();
+}
+```
+
+A number of things change slightly because of differences between type annotation and contextual typing.
+
+2. `asserts` annotation for an arrow function must be on the declaring variable, not on the arrow itself. This no longer works:
```
jakebailey: Why this restriction?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+/**
+ * @type {(a: A) => asserts a is B}
+ */
+const foo = (a) => {
+    if (/** @type { B } */ (a).y !== 0) throw TypeError();
+    return undefined;
+};
+```
+
+This is identical to the Typescript rule.
```
jakebailey: I'm not sure I understand this; one can't separate the types from the declaration in TS, but in JS you have to do so.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+3. Error messages on async functions that incorrectly return non-Promises now use the same error as TS.
+
+4. `@typedef` and `@callback` in a class body are no longer accessible outside the class. 
+They must be moved outside the class to use them outside the class.
+
+5. `@class` or `@constructor` does not make a function into a constructor function.
+
+Corsa ignores `@class` and `@constructor`.
```
jakebailey, line 249: For some of these, I hope we can emit a suggestion diagnostic or something at some point to point out that the JSDoc isn't taking effect.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+If they're in a place where they could apply to multiple functions, they apply only to the first one.
+If you have `"strict": true`, you will see a noImplicitAny error on the now-untyped parameters.
+
+```js
+/** @param {number} x */
+var f = x => x, g = x => x;
+```
+
+7. Optional marking on parameter names now makes the parameter both optional and undefined:
```
jakebailey: Isn't this the opposite of the change made to `=`? Do we have to have `| undefined` at all?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+```js
+/** @param {number} [x] */
+function f(x) {
+    return x;
+}
+```
+
+This behaves the same as Typescript's `x?: number` syntax. 
+Strada makes the parameter optional but does not add `undefined` to the type.
```
jakebailey: Hm, but this seems like the opposite of the section's description? This seems fine, but the description (i commented on above) seems weird.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+    }
+}
+```
+
+In Strada, `cu` incorrectly narrows to `C` inside the `if` block, unlike with TS assertion syntax.
+In Corsa, the behaviour is the same between TS and JS.
+
+### Expandos
+
+1. Expando assignments of `void 0` are no longer ignored as a special case:
```
jakebailey: Why?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+```js
+var o = {}
+o.y = void 0
+```
+
+creates a property `y: undefined` on `o` (which will widen to `y: any` if strictNullChecks is off).
+
+2. A this-property expression with a type annotation in the constructor no longer creates a property:
```
jakebailey: Is this just a cruft cleanup? Or a technical limitation?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+Foo.prototype.bar = function(x) {
+    return x;
+};
+```
+
+Although classes are a much better way to write this code.
+
+### CommonJS
+
+1. Chained exports no longer work:
```
jakebailey: Is this one a cleanup? Or just impossible?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+```
+
+Now only exports `x`, not `y` as well.
+
+2. Exporting `void 0` is no longer ignored as a special case:
+
+```js
+exports.x = void 0
+// several lines later...
+exports.x = theRealExport
```
jakebailey: This seems bad; this syntax is used by transpiled CJS (even our own emit) to ensure that an export "exists" in the case of a cycle, not to actually make it undefined.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+```
+
+This exports `x: undefined` not `x: typeof theRealExport`.
+
+3. Type info for `module` shows a property with name of the export instead of `exports`:
```
jakebailey: I'm not sure what this one means. Are you talking about hover? or something else?

`module`, and `exports` are different symbols entirely, right?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+```js
+module.exports = singleIdentifier
+```
+
+results in `module: { singleIdentifier: any }`
+
+4. Property access on `require` no longer imports a single property from a module:
```
jakebailey: Why are we disallowing this one? I've definitely seen this before...

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+const x = require("y").x
+```
+
+If you can't configure your package to use ESM syntax, you can use destructuring instead:
+
+```js
+const { x } = require("y")
+```
+
+5. `Object.defineProperty` on `exports` no longer creates an export:
```
jakebailey: I have a feeling this one will bite us...

CHANGES.md
sandersn: Pretty much. It shows up in the type baselines so I noted it.

CHANGES.md
sandersn: The right way is to make a WholeType property for functions and check that if the function's Type and parameter types are nil, then only set WholeType in JS. It's possible but I'd put it even below constructor functions and maybe even Object.defineProperty expandos in importance, since contextual typing occurs in nearly all uses of it.

CHANGES.md
sandersn: I don't know, it's just how Typescript works, and Javascript didn't work that way by mistake.

CHANGES.md
sandersn: It's the difference in TS between

```ts
const bad = (a: A): asserts a is B => ...
const good: (a: A) => asserts a is B = a => ...
```

That is what the above jsdoc now reparses to.

CHANGES.md
sandersn: It's extra code, and an extra thing to explain, for a pattern that isn't used anymore.

CHANGES.md
sandersn: I should add an explicit Corsa line to this.

CHANGES.md
sandersn: The difference is that `number=` now has exactly `number | undefined` semantics, including erasure of undefined when strict null checks is off. `[x]` now has exactly `x?` semantics, where before it mistakenly left off `undefined`.

The biggest distinction is that postfix-= is a type, brackets and question mark are parameter syntax. They were both buggy in different ways in Strada.

CHANGES.md
sandersn: It's Closure cleanup, and obsoleted by ES standard syntax. It is basically unused as far as I remember from my survey.

CHANGES.md
sandersn: Simplification -- it's possible, but adds a bunch of code.

CHANGES.md
sandersn: It's for hover, yes. module and exports are different symbols.

CHANGES.md
sandersn: It was extremely rare in the survey I did. And it would require significant hacking to the module exports resolution in the checker, too.

CHANGES.md
sandersn: I didn't know that, so maybe this one should come back.

CHANGES.md
jakebailey: Right; who is the target of this doc? I had figured this was for end users to read when they find something that doesn't work, so this particular one seemed not enough to mention.

CHANGES.md
jakebailey: I'm confused, but maybe I don't understand what this is saying. Are you meaning the difference between these?

```ts
const fn2: (x: any): asserts x = (x) => {};
const fn2 = (x: any): asserts x => {};
```

I don't think anyone is reasonably doing the latter in JS because it's so awkward.

CHANGES.md
jakebailey: Ah, I just said the above before reading this, okay, I get it!

CHANGES.md
jakebailey: Probably this is one we can avoid mentioning for end users. But the doc doesn't structure it like that of course so it's okay.

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+### CommonJS
+
+1. Chained exports no longer work:
+
+```js
+exports.x = exports.y = 12
+```
+
+Now only exports `x`, not `y` as well.
```
weswigham, line 358: Q: Does that mean there's an error on the `exports.y` access?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+/** @param {C | undefined} cu */
+function f(cu) {
+    if (/** @type {any} */ (cu).undeclaredProperty) {
+        cu // still has type C | undefined
+    }
+}
+```
+
+In Strada, `cu` incorrectly narrows to `C` inside the `if` block, unlike with TS assertion syntax.
+In Corsa, the behaviour is the same between TS and JS.
```
weswigham, line 287: ...should we change this TS behavior? Seems like a bug, honestly. The presence of the cast doesn't make the truthiness check not present.

CHANGES.md
weswigham: We'd want it to emit the "convert to class declaration" quickfix on, anyway, right?

CHANGES.md
```diff
@@ -24,7 +24,293 @@ JSDoc types are parsed in normal type annotation position but show a grammar err
...
+
+```js
+/** @param {number[]} ns */
+function sum(...ns) {}
+```
+
+They have no other semantics.
+
+
+2. A variadic type on a parameter no longer makes it a rest parameter. The parameter must use standard rest syntax.
```
weswigham: This one might need some changes in declaration emit? Or not? `...number` still get changed to `number[]`, which is.... still right, I think? Ah, whatever, mental note to check on it.

weswigham (approved):
This looks like a useful summary of where we are right now. A question, though:

> A good number of the diffs are due to the reparsing strategy, which gives JS exactly the semantics of TS.

Should we be applying changes to https://github.com/microsoft/TypeScript/tree/tsgo-port to have a strada version with matching semantics for generating matching baselines? Seems like it'd be useful to have a JS version of the compiler that can report the same diagnostics (and maybe provide upgrade fixes? 6.0?).

CHANGES.md
DanielRosenwasser: Should this be "even when ... is on"

sandersn: No, it is supposed to add undefined when it is on. The Strada bug is that it adds it even when strict is off. I will make this note explain it better.

CHANGES.md
sandersn: yes, it says "property 'y' does not exist on type 'typeof import("/path/to/current/file.js")'"

CHANGES.md
sandersn: I'm not sure; the emit changes for `sum` but unless you ported the special-case check already, the correct emit is `function sum(ns: number[])` and that's what the checker should give you if you're generating from a type.

CHANGES.md
sandersn: I have trouble keeping the precise semantics of narrowing in my head, but `if (cu.undeclaredProperty)` doesn't seem like a truthiness check compared to `if (cu?.undeclaredProperty)`. I guess implicitly it is--"we are here without a crash, so `cu` must exist". Also casts are often an escape hatch for when the compiler goes wrong, so maybe that's why that's here.

CHANGES.md
sandersn: (keeping it in here for now so that it's documented)

CHANGES.md
sandersn: I checked and declaration emit does the right thing already.

CHANGES.md
jakebailey: Related https://github.com/sveltejs/svelte/pull/16485#discussion_r2226227316

CHANGES.md
```diff
@@ -1,5 +1,85 @@
 CHANGES.md lists intentional changes between the Strada (Typescript) and Corsa (Go) compilers.
```
jakebailey:
```suggestion
CHANGES.md lists intentional changes between the Strada (TypeScript) and Corsa (Go) compilers.
```

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
 
+# Feature-Level Description
```
jakebailey: If we are going to extend this file past JS changes, should we name the headers differently?

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
+
+At a high level, Javascript support in Corsa is intended to expose Typescript features in a .js file, working exactly as they do in Typescript with different syntax.
```
jakebailey:
```suggestion
At a high level, JavaScript support in Corsa is intended to expose TypeScript features in a .js file, working exactly as they do in Typescript with different syntax.
```

i don't want to comment on this on every line but it's everywhere 😄

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
+This differs from Strada, which has many Javascript features that do not exist in Typescript at all, and quite a few differences in features that overlap.
+For example, Corsa uses the same rule for checking calls in both Typescript and Javascript; Strada lets you skip parameters with type `any`.
+And because Corsa uses the same rule for optional parameters, it fixes subtle Strada bugs with `"strict": true` in Javascript.
+
+We primarily want to support people writing modern Javascript, using things like ES modules, classes, destructuring, etc.
+Not CommonJS modules and constructor functions, although those do still work.
+However, we have trimmed a lot of unused or underused features.
```
jakebailey, line 12: Can we say that we did so because it makes the implementation simpler / more robust? (We of course aren't cutting for no reason)

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
+
+## JSDoc Tags and Types
+
+| Name | Example | Substitute | Note  |
+| ---- | ------- | ---------- | ----- |
+| UnknownType  | `?`              | `any`               |                                     |
+| NamepathType | `Module:file~id` | `import("file").id` | TS has never had semantics for this |
+| `@class`     | `/** @class \*/`<br/>`function C() {`<br/>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`this.p = 1`<br/>`}` |  | Only inference from `this.p=` or `C.prototype.m=` is supported. |
+|`@throws`     |`/** @throws {E} */` | Keep the same    | TS never had semantics for this     |
+|`@enum` | `/** @enum {number} */`<br/> `const E = { A: 1, B: 2 }`  |`/** @typedef {number} E */`<br/>`/** @type {Record<string, E>}`<br/>`const E = { A: 1, B: 2 }` | Closure feature.  |
```
jakebailey: For the multi-line ones, can you do `<pre><code>a</code><br><code>b</code></pre>` to make them multiline again? rather than single lines and breaks

(this is a nit, sorry)

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
+| `@author`    | `/** @author Finn <finn@treehouse.com> */` | Keep the same  | `@treehouse` parses as a new tag in Corsa. |
```
jakebailey: Dumb question, but I thought tags were lines, and therefore only `@author` would get parsed, right?

CHANGES.md
```diff
@@ -1,5 +1,85 @@
...
+| Ignored empty module.exports assignment      | `module.exports = {}` | Delete this line   | People used to write in this in case module.exports was not defined. |
+| `this` alias for `module.exports`            | `this.p = 1`          | `exports.p = 1`    | When used at the top level of a CommonJS module. |
+| Multiple assignments narrow with control flow| `if (isWindows) {`<br/>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`exports.platform = 'win32'`<br/>`} else {`<br/>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`exports.platform = 'posix'`<br/>`}` | Keep the same in most cases | This now unions instead; most uses have the same type in both branches. |
+| Single-property access `require`             | `var readFile = require('fs').readFile`    | `var { readFile } = require('fs')` | |
+| Aliasing of `module.exports`                 | `var mod = module.exports`<br/>`mod.x = 1` | `module.exports.x = 1` | |
+
+## Features yet to be implemented
+
+Object.defineProperty for CommonJS exports and expandos. The compiler treats this as an alternate to the usual assignment syntax:
```
jakebailey:
```suggestion
`Object.defineProperty` for CommonJS exports and expandos. The compiler treats this as an alternate to the usual assignment syntax:
```

CHANGES.md
sandersn: Yeah...a good answer doesn't spring to mind but I'll think about it.

CHANGES.md
sandersn: I'll do a global replace.

sandersn: I need a prettier-like tool that rewrites typescript to TypeScript on save.

CHANGES.md
sandersn: Yes, that is the official jsdoc.app rule. But the jsdoc lexer/parser is a straight port of Strada, so it also parses tags in the middle of a line.

In order to make the port feasible, I decided to focus on implementations that could either be (1) removed entirely (2) removed and replaced with a small bit of code (3) ported as-is. The jsdoc parser would be a partial rewrite, or else a complete rewrite to something about as complex.  

I wish, though!

Also, figuring out which formatting features people rely on is harder than figuring out what syntax or semantics they use because I can't use the syntax tree or the checker to help me.

CHANGES.md
sandersn: I went with JavaScript Support and Component-Level Changes.

CHANGES.md
sandersn: seems like it works ok, it's just a matter of making macros to do it. Or I could try AI I guess.

jakebailey (approved):

PR #1430, opened 2025-07-21
Implement Go To Implementations command

ahejlsberg:
This PR implements the LSP `textDocument/implementation` request which enables the Go To Implementations command in VS Code.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -605,15 +605,13 @@ func (n *Node) Type() *Node {
 		return n.AsCommonJSExport().Type
 	case KindBinaryExpression:
 		return n.AsBinaryExpression().Type
-	case KindEnumMember, KindBindingElement:
-		return nil
 	default:
 		funcLike := n.FunctionLikeData()
 		if funcLike != nil {
 			return funcLike.Type
 		}
 	}
-	panic("Unhandled case in Node.Type: " + n.Kind.String())
+	return nil
 }
 
 func (n *Node) Initializer() *Node {
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -128,14 +128,10 @@ func newNodeEntryWithKind(node *ast.Node, kind entryKind) *referenceEntry {
 
 func newNodeEntry(node *ast.Node) *referenceEntry {
 	// creates nodeEntry with `kind == entryKindNode`
-	n := node
-	if node != nil && node.Name() != nil {
-		n = node.Name()
-	}
 	return &referenceEntry{
 		kind:    entryKindNode,
-		node:    node,
-		context: getContextNodeForNodeEntry(n),
+		node:    core.OrElse(node.Name(), node),
+		context: getContextNodeForNodeEntry(node),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


internal/ls/findallreferences.go
```diff
@@ -128,14 +128,10 @@ func newNodeEntryWithKind(node *ast.Node, kind entryKind) *referenceEntry {
...
-	n := node
-	if node != nil && node.Name() != nil {
-		n = node.Name()
-	}
 	return &referenceEntry{
 		kind:    entryKindNode,
-		node:    node,
-		context: getContextNodeForNodeEntry(n),
+		node:    core.OrElse(node.Name(), node),
+		context: getContextNodeForNodeEntry(node),
```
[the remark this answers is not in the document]
ahejlsberg: The original logic was incorrectly ported.

internal/ls/findallreferences.go
```diff
@@ -410,22 +406,53 @@ func (l *LanguageService) ProvideReferences(params *lsproto.ReferenceParams) []*
...
+	for i, entry := range entries {
+		if entry.textRange == nil {
+			sourceFile := ast.GetSourceFileOfNode(entry.node)
+			entry.textRange = l.getRangeOfNode(entry.node, sourceFile, nil /*endNode*/)
+			entry.fileName = sourceFile.FileName()
+		}
+		locations[i] = &lsproto.Location{
+			Uri:   FileNameToDocumentURI(entry.fileName),
+			Range: *entry.textRange,
 		}
```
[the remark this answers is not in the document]
ahejlsberg: No need for extra `nil` check, entries aren't supposed to be `nil`.

internal/ast/ast.go
```diff
@@ -605,15 +605,13 @@ func (n *Node) Type() *Node {
...
-	case KindEnumMember, KindBindingElement:
-		return nil
 	default:
 		funcLike := n.FunctionLikeData()
 		if funcLike != nil {
 			return funcLike.Type
 		}
 	}
-	panic("Unhandled case in Node.Type: " + n.Kind.String())
+	return nil
```
jakebailey, line 614: This seems like it's going to end up biting us at some point when we forget to add a case here; is this required?

DanielRosenwasser: Alternatively, `Type()` (panics) and `TryGetType()`/`TypeOrNil()` (returns `nil`)

jakebailey: Well, other than that some nil returning is actually intentional IIRC

internal/ls/findallreferences.go
```diff
@@ -410,22 +406,53 @@ func (l *LanguageService) ProvideReferences(params *lsproto.ReferenceParams) []*
...
+}
+
+func (l *LanguageService) ProvideImplementations(params *lsproto.ImplementationParams) []*lsproto.Location {
+	program, sourceFile := l.getProgramAndFile(params.TextDocument.Uri)
+	position := int(l.converters.LineAndCharacterToPosition(sourceFile, params.Position))
+	node := astnav.GetTouchingPropertyName(sourceFile, position)
+
+	var seenNodes collections.Set[*ast.Node]
+	var entries []*referenceEntry
+	queue := l.getImplementationReferenceEntries(program, node, position)
```
DanielRosenwasser, line 419: The original had something like the following check before processing as a queue

```
    if (
        node.parent.kind === SyntaxKind.PropertyAccessExpression
        || node.parent.kind === SyntaxKind.BindingElement
        || node.parent.kind === SyntaxKind.ElementAccessExpression
        || node.kind === SyntaxKind.SuperKeyword
    ) {
        referenceEntries = entries && [...entries];
    }
```

and idea what that was intended for? I think there was some special handling for not "cascading" beyond the direct superclass

internal/ast/ast.go
ahejlsberg: `Type()` is already optional and we already have logic that always returns `nil` for some node kinds. Seems fine to just allow it for any node kind. `Name()` is similar already.

internal/ls/findallreferences.go
ahejlsberg: Yeah, I think it is so that Go To Implementation on say a `super` keyword only goes to that type and not other types that implement it. Which seems a little odd. Not really sure why you'd want to Go To Implementation on `super`, as opposed to Go To Definition or Go To Type Definition.

jakebailey: I feel like that makes sense, though, since super is basically "do the parent class thing" and so shouldn't jump anywhere else?

ahejlsberg: Well, but what does Go To Implementation on something that isn't a type really mean to begin with? I would think of it as go-to-implementation-of-the-type-of-the-thing-i-am-on. And if the type of `super` is `Base`, then implementations of `Base` also includes `Derived`.

jakebailey: My mental model of it is "the implementation of `super` is code in a parent class's constructor".

ahejlsberg: In Strada, Go To Definition on a `super()` call goes to the constructor code. But Go To Implementation goes to the base class declaration.

DanielRosenwasser: Yeah I looked at the test cases and was surprised. Both go-to-definition and go-to-implementation should probably go to the constructor, but instead we go to the base class for go-to-implementation which feels like it is the less-intuitive thing.

jakebailey: weirdddd

ahejlsberg: I suppose the meaning we ascribe to "implementation" in go-to-implementations is something that implements that type. As opposed to the implementation of a method or constructor.

jakebailey (approved):

PR #1432, opened 2025-07-22
Stop getFlowTypeOfProperty overwriting with base class property

sandersn:
In a bad translation from Typescript, the call to
getTypeOfPropertyInBaseClass always overwrites `initialType`, even when there's no base class or property in the base class.

I think the reason this only affects JS is that in TS all constructor control flow properties are auto-typed, so never take this branch.

This subtle bug would have been prevented with strict null checking.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -26219,8 +26219,13 @@ func (c *Checker) isAssignmentToReadonlyEntity(expr *ast.Node, symbol *ast.Symbo
 }
 
 func (c *Checker) isThisPropertyAccessInConstructor(node *ast.Node, prop *ast.Symbol) bool {
-	kind, _ := c.isConstructorDeclaredThisProperty(prop)
-	return (kind == thisAssignmentDeclarationConstructor || isThisProperty(node) && c.isAutoTypedProperty(prop)) && ast.GetThisContainer(node, true /*includeArrowFunctions*/, false /*includeClassComputedPropertyName*/) == c.getDeclaringConstructor(prop)
+	var constructor *ast.Node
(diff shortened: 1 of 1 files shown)


testdata/baselines/reference/submodule/compiler/jsDeclarationEmitDoesNotRenameImport.types
```diff
@@ -46,9 +46,9 @@ class X extends Test {
 
             this.test = new options.test();
 >this.test = new options.test() : import("./Test.js").default
->this.test : any
+>this.test : import("./Test.js").default | undefined
```
sandersn: I *think* this is better than Strada? I thought that the write type of an auto-typed was `any` but maybe that's changed in our baselines?

jakebailey: If the code is the same, I'd expect us to not have this diff at all, so theoretically there's some other bug somewhere...

sandersn: No, it's not any in the usual case -- I compared Strada vs Corsa for compiler/classAttributeInferenceTemplate. I think Strada is just wrong here.

sandersn: The code is not the same; `isThisPropertyAccessInConstructor` in Strada also checks for `isConstructorDeclaredProperty` as well as `isThisProperty && isAutoTypedProperty`.

jakebailey (approved):

weswigham (approved):
I miss null checking. And unions.

sandersn:
=(

PR #1433, opened 2025-07-22
Verify required properties in generated LSP code

### LINKED ISSUE #1425: Invalid `textEdit` range in `completionItem/resolve` response
`tsgo` returns an invalid `textEdit` range when resolving a completion item. According to [the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#completionItem) "the range of the edit must be a single line range and it must contain the position at which completion has been requested". However with the following code:

```ts
function foo() {
    console.log('foo');
}

page| // Accept completion for pageXOffset
```

Here's the return from the server for the resolve request:
```
[Trace - 10:01:39 PM] Received response 'completionItem/resolve - (26)' in 3ms.
Result: {
    "label": "pageXOffset",
    "kind": 6,
    "detail": "var pageXOffset: number",
    "documentation": {
        "kind": "markdown",
        "value": "[MDN Reference](https://developer.mozilla.org/docs/Web/API/Window/scrollX)"
    },
    "sortText": "15",
    "insertTextFormat": 1,
    "textEdit": {
        "range": {
            "start": {
                "line": 0,
                "character": 0
            },
            "end": {
                "line": 0,
                "character": 0
            }
        },
        "newText": "pageXOffset"
    },
    "commitCharacters": [
        ".",
        ",",
        ";"
    ],
    "data": {
        "fileName": "...",
        "name": "pageXOffset",
        "position": 46
    }
}
```

This causes the following behavior:

https://github.com/user-attachments/assets/0d9e4628-aff6-4694-beb6-5151a2ae9bcb

jakebailey:
Fixes #1425

Our union type decoding is broken because Go's JSON parsing has no mechanism to verify required properties. This meant that decoding into a union of structs is almost always guaranteed to succeed at choosing the first choice, except in the case where the types are incompatible.

This PR adds an unmarshal function to each structure that verifies that all required properties have been provided. This is not particularly efficient, but it does get the job done.

### DIFF
--- a/internal/lsp/lsproto/_generate/generate.mjs
+++ b/internal/lsp/lsproto/_generate/generate.mjs
@@ -361,43 +361,61 @@ function generateCode() {
     writeLine("// Structures\n");
 
     for (const structure of model.structures) {
-        write(formatDocumentation(structure.documentation));
-
-        writeLine(`type ${structure.name} struct {`); // Embed extended types and mixins
-        for (const e of structure.extends || []) {
-            if (e.kind !== "reference") {
-                throw new Error(`Unexpected extends kind: ${e.kind}`);
+        /**
+         * @param {string} name
+         * @param {boolean} includeDocumentation
+         */
+        function generateStructFields(name, includeDocumentation) {
+            if (includeDocumentation) {
+                write(formatDocumentation(structure.documentation));
             }
-            writeLine(`\t${e.name}`);
-        }
 
-        for (const m of structure.mixins || []) {
-            if (m.kind !== "reference") {
-                throw new Error(`Unexpected mixin kind: ${m.kind}`);
+            writeLine(`type ${name} struct {`);
+
+            // Embed extended types and mixins
+            for (const e of structure.extends || []) {
+                if (e.kind !== "reference") {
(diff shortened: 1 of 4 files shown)


internal/lsp/lsproto/_generate/generate.mjs
```diff
@@ -361,44 +361,87 @@ function generateCode() {
...
+
+            for (const prop of requiredProps) {
+                writeLine(`\tif _, ok := keys["${prop.name}"]; !ok {`);
+                writeLine(`\t\treturn fmt.Errorf("required key '${prop.name}' is missing")`);
+                writeLine(`\t}`);
+            }
+
+            writeLine(``);
+            writeLine(`\t// Redeclare the struct to prevent infinite recursion`);
+            generateStructFields("temp", false);
```
gabritto, line 448: Why do we need this? What's the infinite recursion this is preventing?

jakebailey: Inside `UnmarshalJSON`, we still need to unmarshal the data into the struct after validating the keys, except if we just do a plain `json.Unmarshal(data, v)`, then it'll just call `UnmarshalJSON` again.

gabritto (approved):

PR #1437, opened 2025-07-23
Port go-to-definition baseline tests

DanielRosenwasser:
A few noteworthy things:

- `goToDefinition` and `getDefinitionAtPosition` are slightly distinct in Strada, but we have no such distinction around "bound spans" in LSP that I'm aware of. I'm not sure if this is something "missing" in LSP that we should revisit.
- Baselines for find-all-references would previously drop files that contained the originating request. Now they are preserved.
- I noticed that baselines for fourslash are slightly different due to file ordering. Might need to look into that.


internal/fourslash/fourslash.go
```diff
@@ -832,9 +822,92 @@ func (f *FourslashTest) VerifyBaselineFindAllReferences(
...
+				resultAsLocations = core.Map(*resultAsLocOrLocations.Locations, func(loc lsproto.Location) *lsproto.Location {
+					return &loc
+				})
+			} else {
+				resultAsLocations = []*lsproto.Location{resultAsLocOrLocations.Location}
+			}
+
+			f.baseline.addResult("goToDefinition", f.getBaselineForLocationsWithFileContents(resultAsLocations, baselineFourslashLocationsOptions{
+				marker:     markerOrRange.GetMarker(),
+				markerName: "/*GO TO DEFINITION*/",
```
DanielRosenwasser, line 881: This marker is different from both of the ones we have in Strada. Open to adopting one over the other.

DanielRosenwasser:
A few things that I want to look into tomorrow:

1. Some of these tests are incorrectly generated because they can't support `...test.markerNames()`
2. The test suite isn't resilient to a `nil` result right now. The same might be true of find-all-references.
3. I think the old code specifically adds the triggering file at the end if it was never encountered.

I think that causes more tests to be considered failing than is really necessary.

gabritto:
> A few things that I want to look into tomorrow:
> 
> 1. Some of these tests are incorrectly generated because they can't support `...test.markerNames()`
> 2. The test suite isn't resilient to a `nil` result right now. The same might be true of find-all-references.
> 3. I think the old code specifically adds the triggering file at the end if it was never encountered.
> 
> I think that causes more tests to be considered failing than is really necessary.

1. I think we shouldn't have the script convert tests that we know cannot yet be converted properly. If tests that have `...test.markerNames()` won't work for now, we should just skip porting them, otherwise it's hard to keep track of what has been successfully ported and what hasn't.
2. This is just me being curious, but will we be able to compare baselines between Strada and Corsa in any way, or are the spans always too different?

DanielRosenwasser:
> The code seems fine but I don't know how reasonable it will be to actually review all half a million lines of baselines...

Yeah, for now, I would say don't really. Just eyeball the accepted baselines at each commit. Knowing what is currently failing will be a little bit helpful though.


> This is just me being curious, but will we be able to compare baselines between Strada and Corsa in any way, or are the spans always too different?

I don't know, but I would still like to see if diffing could help later.

> I think we shouldn't have the script convert tests that we know cannot yet be converted properly.

Probably something to try later. Right now they're incorrectly converted and skipped.

DanielRosenwasser:
> The same might be true of find-all-references.

Apparently not because it just comes back as a nil slice.

jakebailey (approved):

iisaduan (approved):
Looks good, and good catch on the trigger locations not being printed if references to the location aren't returned!

iisaduan:
Some comments on things mentioned above: 

> `...test.markerNames()`

can be translated to
```go
...core.MapNonNil(f.Markers(), func(m *fourslash.Marker) *string {return m.Name})
```
which will return a slice of all markers in test `f` with a name (skipping unnamed object and range markers, just like in strada)


> 2. This is just me being curious, but will we be able to compare baselines between Strada and Corsa in any way, or are the spans always too different?

I thought about this a lot and tried some different methods of comparing when I was writing the FAR baseline, but besides the different data being returned, other things like whitespace, lines skipped, file and per-marker-result ordering in the baseline weren't the same, so it made an overall string comparison hard.
I have some more ideas that I think would work, but in the interest of time for the original PR, I wasn't able to fully try them (and it would probably require something slightly different for every fourslash baseline function)



>2. The test suite isn't resilient to a nil result right now. The same might be true of find-all-references.

It would be nice to have better error handling in the LS (and fourslash) in general. The error handling in `baselineFAR` was pretty much added on as I ran into them, and doesn't really do anything to report why the error happened

Issue #1439, opened 2025-07-23
Support `LocationLink` in go-to-definition


DanielRosenwasser:
[From the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition):

> Goto Definition Request (:leftwards_arrow_with_hook:)
> The go to definition request is sent from the client to the server to resolve the definition location of a symbol at a given text document position.
> 
> The result type [`LocationLink`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink)`[]` got introduced with version 3.14.0 and depends on the corresponding client capability `textDocument.definition.linkSupport`

where `LocationLink` looks like

```ts
interface LocationLink {

	/**
	 * Span of the origin of this link.
	 *
	 * Used as the underlined span for mouse interaction. Defaults to the word
	 * range at the mouse position.
	 */
	originSelectionRange?: Range;

	/**
	 * The target resource identifier of this link.
	 */
	targetUri: DocumentUri;

	/**
	 * The full target range of this link. If the target for example is a symbol
	 * then target range is the range enclosing this symbol not including
	 * leading/trailing whitespace but everything else like comments. This
	 * information is typically used to highlight the range in the editor.
	 */
	targetRange: Range;

	/**
	 * The range that should be selected and revealed when this link is being
	 * followed, e.g the name of a function. Must be contained by the
	 * `targetRange`. See also `DocumentSymbol#range`
	 */
	targetSelectionRange: Range;
}
```

In Strada, we had the concept of a "definition with bound span". In Corsa we only provide `Location | Location[]` for go-to-definition. We probably should model the concept of a bound span via the `targetSelection`.

jakebailey:
One thing I'll want to do for this is fix the LSP generation script to notice that in this:

```
		{
			"method": "textDocument/definition",
			"typeName": "DefinitionRequest",
			"result": {
				"kind": "or",
				"items": [
					{
						"kind": "reference",
						"name": "Definition"
					},
					{
						"kind": "array",
						"element": {
							"kind": "reference",
							"name": "DefinitionLink"
						}
					},
					{
						"kind": "base",
						"name": "null"
					}
				]
			},
```

That `Definition` is an alias for `LocationOrLocations` which is a union, so this needs to be flattened. Probably this also means that we should just ditch aliases and use the full types to simplify this idea.

DanielRosenwasser:
[...]

PR #1441, opened 2025-07-23
Generate LSP response types, redo nullable, delete aliases, statically type LSP handlers

jakebailey:
This allows us to work on #1439.


DanielRosenwasser (approved):

DanielRosenwasser (approved):

jakebailey:
Okay, well I went a little overboard. Now, the request/notification handlers are fully statically typed, something I've wanted to do for a while. I also eliminated `OrNull` for response types, since `nil` works just as well for those.

internal/lsp/server.go
```diff
@@ -459,71 +463,77 @@ func (s *Server) sendResponse(resp *lsproto.ResponseMessage) {
...
+	if handler := handlers[req.Method]; handler != nil {
+		return handler(s, ctx, req)
+	}
+	s.Log("unknown method", req.Method)
+	if req.ID != nil {
 		s.sendError(req.ID, lsproto.ErrInvalidRequest)
-		return nil
-	case *lsproto.InitializedParams:
-		return s.handleInitialized(ctx, req)
-	case *lsproto.DidOpenTextDocumentParams:
```
jakebailey, line 469: Changing how this works is actually a prereq for handling requests with the same params but different methods.

internal/fourslash/fourslash.go
```diff
@@ -544,17 +549,14 @@ func (f *FourslashTest) verifyCompletionsWorker(t *testing.T, expected *Completi
...
 	}
-	result := resMsg.AsResponse().Result
-	switch result := result.(type) {
-	case *lsproto.CompletionList:
-		f.verifyCompletionsResult(t, f.currentCaretPosition, result, expected, prefix)
-	default:
-		t.Fatalf(prefix+"Unexpected response type for completion request: %v", result)
+	if !resultOk {
+		t.Fatalf(prefix+"Unexpected response type for completion request: %T", resMsg.AsResponse().Result)
 	}
```
DanielRosenwasser, line 558: It'd be ideal if `sendRequest` (or a wrapper function) automatically errored based on `resMsg` and `resultOk` and used the passed `Method` for the errors.

internal/fourslash/fourslash.go
```diff
@@ -267,20 +267,25 @@ func getCapabilitiesWithDefaults(capabilities *lsproto.ClientCapabilities) *lspr
 	return &capabilitiesWithDefaults
 }
 
-func (f *FourslashTest) sendRequest(t *testing.T, method lsproto.Method, params any) *lsproto.Message {
+func sendRequest[Params, Resp any](t *testing.T, f *FourslashTest, mapping lsproto.RequestInfo[Params, Resp], params Params) (*lsproto.Message, Resp, bool) {
```
DanielRosenwasser: My nit is that I'd prefer fourslash still come first, but it's fine.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -546,6 +583,49 @@ function generateCode() {
...
+        if (Array.isArray(request.params)) {
+            throw new Error("Unexpected request params for " + methodName + ": " + JSON.stringify(request.params));
+        }
+
+        const paramType = request.params ? resolveType(request.params) : undefined;
+        const paramGoType = paramType ? (paramType.needsPointer ? `*${paramType.name}` : paramType.name) : "any";
+
+        writeLine(`// Type mapping info for \`${request.method}\``);
+        if (responseTypeName) {
+            writeLine(`var ${methodName}Info = RequestInfo[${paramGoType}, ${responseTypeName}]{Method: Method${methodName}}`);
```
DanielRosenwasser, line 684: If you don't like `Info`, we can call them `Mapping`s, `Pairing`s, `ReqRes`, or maybe some combo of those.

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -13286,269 +13153,913 @@ const (
...
+	}
+	var vStrings []string
+	if err := json.Unmarshal(data, &vStrings); err == nil {
+		o.Strings = &vStrings
+		return nil
+	}
+	return fmt.Errorf("invalid StringOrStrings: %s", data)
+}
+
+type TextDocumentContentChangePartialOrTextDocumentContentChangeWholeDocument struct {
```
DanielRosenwasser: Something difficult about this naming scheme is that it kind of repeats itself. Would it be too "cute" to find a way to name this `TextDocumentContentChangePartialOrWholeDocument`?

internal/lsp/lsproto/lsp.go
```diff
@@ -92,3 +75,16 @@ func (v *requiredProp) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
 	*v = true
 	return dec.SkipValue()
 }
+
+// Inspired by https://www.youtube.com/watch?v=dab3I-HcTVk
+
+type RequestInfo[Params, Resp any] struct {
```
DanielRosenwasser, line 81: Given the parameter in fourslash is called `mapping`, maybe it makes sense to at least call this a `ParameterResponseMapping` or something

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -13286,269 +13153,913 @@ const (
...
+var TextDocumentHoverInfo = RequestInfo[*HoverParams, HoverResponse]{Method: MethodTextDocumentHover}
+
+// Response type for `textDocument/signatureHelp`
+type SignatureHelpResponse = *SignatureHelp
+
+// Type mapping info for `textDocument/signatureHelp`
+var TextDocumentSignatureHelpInfo = RequestInfo[*SignatureHelpParams, SignatureHelpResponse]{Method: MethodTextDocumentSignatureHelp}
+
+// Response type for `textDocument/definition`
+type DefinitionResponse = *LocationOrLocationsOrDefinitionLinks
```
DanielRosenwasser, line 13447: It is a little tricky to remember this is a pointer internally, but I'll get used to it.

DanielRosenwasser (approved):

internal/fourslash/fourslash.go
jakebailey: If I do that, then lint rules that say "`t` comes first" will get mad.

internal/lsp/lsproto/_generate/generate.mts
jakebailey: I had it as "Mapping" before. ReqRes not so much because it has to also work for notifications.

internal/lsp/lsproto/lsp_generated.go
jakebailey: I could try and find a common prefix.

internal/lsp/lsproto/lsp.go
jakebailey: That's a mistake; I meant to rename that one too.

internal/fourslash/fourslash.go
jakebailey: I wanted to do this, but it means I have to also pass in `prefix` which felt strange...

DanielRosenwasser (approved):

PR #1444, opened 2025-07-24
Special-case JSDocVariadicType in JSDocSignature

sandersn:
Currently in Corsa, `...T` is always interpreted as `T[]`, which is considerably
simpler than in Strada. However, there is no
way to express rest parameters in a `@callback` or `@overload` tag. This
PR special-cases `...` in `@param` tags in JSDocSignature, which is how these tags parse,
such that instead of `...T --> T[]`,
it reparses as `...T --> ...paramName: T`.

Notes:
1. This is how it used in unifier, which is a heavily jsdoc'ed
codebase checked with Strada.
2. This PR *removes* the variadic type in reparsing, meaning that the
author needs to provide an array type, like `...[number, number]` for
example. This is still a subset of the Strada semantics.
3.  Code that relies on Strada treating `...` both as rest *and* as
array type in this position gets an error that an array type is
required. This should help other uses update to Corsa semantics.

The new semantics are inconsistent but they are still fairly simple and
should match with how people use `...T` in Strada. See how the new test
based on example code from unifiedjs uses `...T` to mean rest parameter
on `@overload` and to mean `T[]` on a `@param` of the implementation.


testdata/baselines/reference/submodule/conformance/callbackTagVariadicType.errors.txt
```diff
@@ -1,16 +1,20 @@
...
+callbackTagVariadicType.js(3,4): error TS2370: A rest parameter must be of an array type.
+callbackTagVariadicType.js(9,13): error TS2345: Argument of type '["a", "b"]' is not assignable to parameter of type 'string'.
 
 
-==== callbackTagVariadicType.js (1 errors) ====
+==== callbackTagVariadicType.js (2 errors) ====
     /**
      * @callback Foo
      * @param {...string} args
+       ~~~~~~~~~~~~~~~~~~~~~~~
```
sandersn, line 9: this shows the old, still unsupported semantics where the parameter `args` becomes a rest parameter *and* the type `string` becomes `string[]`. Only the first happens now, based on seeing how a real person (Titus Woormer) wrote overloads.

testdata/baselines/reference/submodule/conformance/callbackTagVariadicType.js.diff
```diff
@@ -23,7 +23,7 @@
 
 
  //// [callbackTagVariadicType.d.ts]
-+export type Foo = (args: string[]) => number;
++export type Foo = (...args: string) => number;
```
jakebailey, line 26: Nothing you did in this PR, but this looks like a declaration emit bug where `callback` is not supposed to be exported, right?

sandersn: well, no, `@callback` is like `@typedef` in that it's implicitly exported.

jakebailey: Except the original d.ts file didn't do so, no?

jakebailey: Oh, I can't read, it's just an ordering problem, I can see it below hidden in the github diff.

PR #1446, opened 2025-07-24
Adding manual tests list

navya9singh:
After tests are generated with `npm run convertfourslash`, if we need to edit them manually, then this script can be run by
 `npm run manualtest testname`. This creates a `manual` folder under `fourslash/tests/` and adds the name of the test to `manualTests.txt`. 
- All tests in `manualTests.txt` will ne skipped and not regenerated when convertfourslash is run again. 
- Once a `manual` folder is created, `testOutput = cp.execFileSync(go, ["test", "./internal/fourslash/tests/gen"],  ["test", "./internal/fourslash/tests/manual"],  { encoding: "utf-8" })` can be added to `updateFailing.mts` to run all the tests under `./manual` and update `failingtests.txt` accordingly.


gabritto (approved):

internal/fourslash/tests/manual/util_test.go
jakebailey, line 1: This one test is 14k lines; is this intentionally here? Is it generated somehow?

internal/fourslash/tests/manual/completionListInClosedFunction05_test.go
```diff
@@ -20,7 +20,7 @@ func TestCompletionListInClosedFunction05(t *testing.T) {
 	f.VerifyCompletions(t, "1", &fourslash.CompletionsExpectedList{
 		IsIncomplete: false,
 		ItemDefaults: &fourslash.CompletionsExpectedItemDefaults{
-			CommitCharacters: &[]string{},
+			CommitCharacters: &defaultCommitCharacters,
```
jakebailey: Maybe a silly question but why can't we generate this one automatically?

gabritto: There's no simplified way to compute what the commit characters should be other than the actual implementation. The porting script currently sets empty commit characters if `isNewIdentifierLocation` is true, which is an approximation that does not always hold, like in this test. An alternative would be to ignore commit characters for all ported tests, but I didn't want to do that and end up with no coverage of it.

jakebailey: Would it be simpler to just keep a list of these in the generator script?

internal/fourslash/tests/manual/util_test.go
gabritto: That's just the helper file with things like global completions lists. It's not generated, but it's a copy of the file with the same name in `internal/fourslash/tests`. It's annoying that folders are different packages and therefore cannot use the same utils file without importing. Eventually there'll be just one folder with a single copy of this file.

jakebailey: Is it worth it to move these to a package these all can import at this point?

internal/fourslash/tests/manual/completionListInClosedFunction05_test.go
gabritto: For this case, probably, but there are other tests that need other types of manual fixes, and tests that will need to be at least partially manually ported because they have things like for loops and functions.

internal/fourslash/tests/manual/util_test.go
gabritto: I think the reason why I didn't want to do that before is that it makes names longer in tests, and we'd eventually undo it, but it probably doesn't matter one way or the other.

jakebailey: Possibly we could just "sin" and use a `.` import, pulling everything into the local scope.

gabritto: We'll end up having to rename from lowercase to uppercase and then to lowercase again later when we finish porting tests. Should I go ahead and do it in a separate PR?

jakebailey: Yeah.

gabritto: #1473

jakebailey, line 1: I think this can be deleted now.

jakebailey (approved):

PR #1447, opened 2025-07-24
Write list of unparsed fourslash tests to file

jakebailey:
That way we know our progress.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,9 +251,7 @@ jobs:
       - run: node --experimental-strip-types ./internal/lsp/lsproto/_generate/generate.mts
 
       - name: Regenerate fourslash tests and update failing test list
-        run: |
-          npm run convertfourslash
-          npm run updatefailing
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -251,7 +251,9 @@ jobs:
       - run: node --experimental-strip-types ./internal/lsp/lsproto/_generate/generate.mts
 
       - name: Regenerate fourslash tests and update failing test list
-        run: npm run updatefailing
+        run: |
+          npm run convertfourslash
+          npm run updatefailing
```
gabritto: `npm run updatefailing` will run `convertFourslash` (twice), so, not sure why we'd need to add it here.

jakebailey: Yeah I was just trying to figure out why the test is failing even though I already committed the file...

gabritto: I see. Yeah, I don't get why it's failing. 🙁

PR #1448, opened 2025-07-24
Build with -trimpath for nicer stack traces

tmm1:


tmm1:
Ah missed that. What do you suggest for ppl doing their own internal releases?

Maybe I can make it conditional based on an env ?

jakebailey:
If you're doing your own thing, then you can modify the code on your end. We're not going to accept patches for customization here for that purpose.

tmm1:
Sounds good I can keep a git stash. I've been tracking main locally and wanted to share cleaner stack traces.

PR #1450, opened 2025-07-24
Implement `MarkerNames()` in fourslash

DanielRosenwasser:


testdata/baselines/reference/fourslash/findAllRef/FindReferencesAfterEdit.baseline.jsonc
DanielRosenwasser, line 1: Actually this change is sketchy - but I think it is possibly due to a bug in the baselining logic. My guess is that baselining is currently too eager. Any test that requests multiple baselines where an edit has occurred in between probably fails because any of the following might be true:

1. the baselines were never accepted so the test could never succeed
2. the second baseline call will always read the results of the first baseline, and is assumed to fail
3. the second baseline was accepted, so the first call will always fail

jakebailey (approved):

PR #1454, opened 2025-07-24
Support `@satisfies` on same nodes as `@type`

sandersn:
Unlike `@type`, there is no special field for these; they just create a satisfies expression around the expression or initialiser of the parent node.

The code is duplicative and AI-generated but I didn't see a way to reduce duplication, much like the `@type` code.

Compiling remark-link reminded me that I forgot to do this earlier, since it uses `@satisfies`.


jakebailey (commented):
Seems fine but the duplication is kinda yucky, yeah.

sandersn:
I moved a lot of the duplication over to ast.go as mutableNode methods, and I think it's an improvement. See if you agree.

sandersn:
The diff is completely unreadable now though. Pretty much have to read the new side only.

jakebailey:
I don't think I like the new thing more, but at least it has a default panic so we could figure out when we forgot it?

If you like the old style better then that's fine.

sandersn:
From our individual discussion: I like the current state slightly better than the original commit because it's easier for humans to read and notice the slight differences between cases. Otherwise it's pretty much the same.

PR #1456, opened 2025-07-24
Add locale-sensitive sorting + completion fixes

gabritto:
This adds locale-sensitive sorting, similar to what we had in Strada, but using the `collate` package.
Tests set `en-US` locale, the server gets its locale from the client during initialization, the API doesn't specify a locale yet.
An annoying thing is that the collator is not concurrency-safe, since calling `collator.CompareString` sets state in the collator.

I'm not sure what the best way to pass down the locale from the server to the language service methods is.

There are also some small completion fixes.


jakebailey:
> I'm not sure what the best way to pass down the locale from the server to the language service methods is.

This will end up being `context.Context` thing (see `GetRequestID`, `GetFormatCodeSettingsFromContext`).

internal/fourslash/fourslash.go
```diff
@@ -227,7 +227,11 @@ func (f *FourslashTest) nextID() int32 {
 }
 
 func (f *FourslashTest) initialize(t *testing.T, capabilities *lsproto.ClientCapabilities) {
-	params := &lsproto.InitializeParams{}
+	params := &lsproto.InitializeParams{
+		InitializeParamsBase: lsproto.InitializeParamsBase{
+			Locale: ptrTo("en-US"),
```
jakebailey: Probably we should still let this be unspecified, but then treat unspecified as `en-US` in the LS.

internal/ls/completions.go
```diff
@@ -2982,11 +2983,11 @@ func getCompletionsSymbolKind(kind ScriptElementKind) lsproto.CompletionItemKind
 // by the language service consistent with what TS Server does and what editors typically do. This also makes
 // completions tests make more sense. We used to sort only alphabetically and only in the server layer, but
 // this made tests really weird, since most fourslash tests don't use the server.
-func compareCompletionEntries(entryInSlice *lsproto.CompletionItem, entryToInsert *lsproto.CompletionItem) int {
-	// !!! use locale-aware comparison
-	result := stringutil.CompareStringsCaseSensitive(*entryInSlice.SortText, *entryToInsert.SortText)
+func (l *LanguageService) compareCompletionEntries(entryInSlice *lsproto.CompletionItem, entryToInsert *lsproto.CompletionItem) int {
+	compareStrings := collate.New(l.host.GetLocale()).CompareString
```
jakebailey: I suspect we'll want to avoid calling `New` often, since that's not a cheap operation.

Probably, we should make some sort of new `locale` package (or dump it in core) which has a struct that caches the collator and so on.

jakebailey: Actually reading their impl, I'm not sure collators are concurrency safe 😓

gabritto: They are not, as I noted in the description. My first attempt at this code had a shared collator, but it crashes. Also saw https://github.com/golang/go/issues/67286.

internal/fourslash/fourslash.go
gabritto: There's already an `undefined` locale, is there a reason why we'd want to default to `en-US` in general?

internal/ls/completions.go
jakebailey: Yeah, sorry, I missed the description and clicked right to the code in my email.

internal/fourslash/fourslash.go
jakebailey: IIRC Strada defaulted to en-US for localization, but I'm not sure how it worked otherwise.

jakebailey:
I'll note that including the `x/text` stuff increases the binary size by 1.8MB on my machine (at least, the non-release binary); not great but I don't know how it's avoidable.

internal/core/context.go
```diff
@@ -19,3 +26,17 @@ func GetRequestID(ctx context.Context) string {
 	}
 	return ""
 }
+
+func WithLocale(ctx context.Context, locale language.Tag) context.Context {
+	return context.WithValue(ctx, localeKey, locale)
+}
+
+func GetLocale(ctx context.Context) language.Tag {
+	if ctx == nil {
```
jakebailey: Context is never undefined; not sure how I didn't catch this in the GetRequestID PR.

internal/fourslash/fourslash.go
gabritto: I think Strada left `locale` as `undefined` unless provided, which, according to MDN, means "the runtime's default locale is used when undefined" for the collator. I tested and Node just gets the system's locale.
I don't know how this compares. I'd hope using the `undefined` locale would result in less opinionated choices somehow, but I don't know if that's even possible.

internal/core/context.go
gabritto: It seems to be undefined in the tests 😬

jakebailey: That's definitely a bug; nothing should ever pass in a nil context, only `Background` if it's the top of a call stack and not likely to get a context, or `TODO` if it will in the future.

jakebailey: Ah, `LanguageService` still has a context on it. That should be removed.

internal/ls/completions.go
```diff
@@ -2983,26 +2987,28 @@ func getCompletionsSymbolKind(kind ScriptElementKind) lsproto.CompletionItemKind
...
-	//         (entryToInsert.data as CompletionEntryDataResolved).moduleSpecifier,
-	//     );
-	// }
-	if result == stringutil.ComparisonEqual {
-		// Fall back to symbol order - if we return `EqualTo`, `insertSorted` will put later symbols first.
-		return stringutil.ComparisonLessThan
-	}
+func getCompareCompletionEntries(ctx context.Context) func(entryInSlice *lsproto.CompletionItem, entryToInsert *lsproto.CompletionItem) int {
+	return func(entryInSlice *lsproto.CompletionItem, entryToInsert *lsproto.CompletionItem) int {
+		compareStrings := collate.New(core.GetLocale(ctx)).CompareString
```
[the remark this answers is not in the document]
gabritto: Not entirely a bad suggestion, but let's see if we need to do this.

internal/core/context.go
jakebailey: https://github.com/microsoft/typescript-go/pull/1463

gabritto: ~I'll do that in a follow-up.~ Never mind, just saw #1463

jakebailey:
Yeah, this does increase the binary size by 1.4MB due to the collation tables.

Also, it turns out that the collation tables are very out of date: https://github.com/golang/go/issues/52485

I'm trying to figure out how to fix this upstream problem and get them up to date, but it makes me wonder if it's worth doing this at the moment given the Unicode versions don't really line up?

There are other unrelated fixes in this PR, can we separate those out and move the collation stuff into its own PR to think about? I'm curious which tests are specifically due to the collation and whether or not it's even worth trying to fix it. But, it may be inevitable that we need the the collation tables for import sorting...

gabritto:
A majority of the fixed tests in this PR is because of the sort order, IIRC. I could probably make the completion tests pass by using a case-insensitive sorting for now, but we will need this for organize imports soon. I just didn't really know that we would want to go and put in the effort to either implement our own locale-aware sorting thing, or improve the package, or something else, but if you think we're actually gonna try and do that, let me know and I'll undo the collator changes.

jakebailey:
I hate to say it but I think we should probably hold off on the collate thing temporarily, and just leave a TODO and do the insensitive ordering to make the tests feel better. I would like to see if we can get it up to date before we commit to having such large tables in the binary, at least for a little bit.

jakebailey (commented):
Nah, I changed my mind, it's okay.

What we do need in this PR is an update in the NOTICE file. The license is the same as the other Go libs so that can be copied and pasted with version updates.

Alternatively, I can do it tomorrow if that's when this is merged. I need to set up automation for that in general anyway.

jakebailey (approved):

PR #1460, opened 2025-07-25
Fix more LSP related issues

jakebailey:
The result of my refactor is even more bugs.

This PR:

- Verifies that the LSP messages serialize in fourslash, which would have caught the problem.
- Undoes the `OrNull` -> pointer trick, which broke cases where we return a nil response.
- Pushes the response types further down, which finds places we are mishandling things.

I'd like to undo the undo of `OrNull` at some point, but right now I just want to stop the crashing.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -1,6 +1,7 @@
 package fourslash
 
 import (
+	"encoding/json"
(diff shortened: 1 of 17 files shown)


jakebailey:
Copilot's reviews are bogus; a zero struct means `null`.

PR #1468, opened 2025-07-28
Full-signature types for JSDoc

sandersn:
This PR allows `@type` tags to apply as a whole to a function. This is a returning Strada feature, but it's no longer implemented as a fallback lookup for a type tag. Instead, `FunctionLikeBase` now has a property `WholeType` that gets set in the reparser. If we can come up with Typescript syntax for this feature, it will work immediately--with a couple of caveats described below.

In the checker, the implementation is basically the same as in Strada, except that it refers to a property on FunctionLikeBase instead of looking up a tag. I put it on FunctionLikeBase because it is more convenient to retrieve, even though it only applies to FunctionLikeWithBodyBase. Notably, there was one check that I skipped because it short-circuits contextual typing in favour of the whole type, which I don't believe is correct--or is perhaps unobservable. When I removed it no tests changed.

I didn't check that declaration emit works, but I think there may be some tests that exercise it. In any case, since the signature's types are pulled from the whole-type when needed, and parameter/return types are supposed to be nil when it's present, I expect that emit will generate types correctly from the signature.

Another difference is that the parameter names of the resulting signature come from the type annotation, not the function declaration. This is consistent with something like
```ts
const x: number = 'xx'
```
where `x: number` when used even though it is initialised with a string, which of course is a string.

Speaking of errors, I improved the wording of the existing error and added another that forbids using `@type` and `@param/@return`. Both errors are worded in terms of JSDoc tags even though we could implement this feature in TS. We can revisit this either after coming up with a good name for this feature, or after we implement it in TS, if ever.

## Open Questions
1. What should this feature be called? Whole-function type annotation? Function annotation? TypeTag type annotation? This leads naturally to:
2. What should the property on FunctionLikeBase be called?
3. Can the errors be improved further, either the wording or the checks?
4. (bonus) What would TS syntax for this look like?

## Open Tasks
1. Rename all the properties and functions to use the new terminology.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -2125,7 +2125,7 @@ type FunctionLikeBase struct {
 	TypeParameters *NodeList // NodeList[*TypeParameterDeclarationNode]. Optional
 	Parameters     *NodeList // NodeList[*ParameterDeclarationNode]
 	Type           *TypeNode // Optional
-	WholeType      *TypeNode // Type that applies to the whole function; should not be set if Type is set or if Parameters have types set.
+	FullSignature  *TypeNode // Type that applies to the whole function; should not be set if Type is set or if Parameters have types set.
 }
 
 func (node *FunctionLikeBase) LocalsContainerData() *LocalsContainerBase {
@@ -3704,7 +3704,7 @@ type FunctionDeclaration struct {
 	ReturnFlowNode *FlowNode
 }
 
-func (f *NodeFactory) NewFunctionDeclaration(modifiers *ModifierList, asteriskToken *TokenNode, name *IdentifierNode, typeParameters *NodeList, parameters *NodeList, returnType *TypeNode, wholeType *TypeNode, body *BlockNode) *Node {
+func (f *NodeFactory) NewFunctionDeclaration(modifiers *ModifierList, asteriskToken *TokenNode, name *IdentifierNode, typeParameters *NodeList, parameters *NodeList, returnType *TypeNode, fullSignature *TypeNode, body *BlockNode) *Node {
 	data := f.functionDeclarationPool.New()
 	data.modifiers = modifiers
 	data.AsteriskToken = asteriskToken
@@ -3712,29 +3712,29 @@ func (f *NodeFactory) NewFunctionDeclaration(modifiers *ModifierList, asteriskTo
 	data.TypeParameters = typeParameters
 	data.Parameters = parameters
 	data.Type = returnType
-	data.WholeType = wholeType
+	data.FullSignature = fullSignature
 	data.Body = body
 	return f.newNode(KindFunctionDeclaration, data)
 }
 
-func (f *NodeFactory) UpdateFunctionDeclaration(node *FunctionDeclaration, modifiers *ModifierList, asteriskToken *TokenNode, name *IdentifierNode, typeParameters *TypeParameterList, parameters *ParameterList, returnType *TypeNode, wholeType *TypeNode, body *BlockNode) *Node {
-	if modifiers != node.modifiers || asteriskToken != node.AsteriskToken || name != node.name || typeParameters != node.TypeParameters || parameters != node.Parameters || returnType != node.Type || wholeType != node.WholeType || body != node.Body {
-		return updateNode(f.NewFunctionDeclaration(modifiers, asteriskToken, name, typeParameters, parameters, returnType, wholeType, body), node.AsNode(), f.hooks)
+func (f *NodeFactory) UpdateFunctionDeclaration(node *FunctionDeclaration, modifiers *ModifierList, asteriskToken *TokenNode, name *IdentifierNode, typeParameters *TypeParameterList, parameters *ParameterList, returnType *TypeNode, fullSignature *TypeNode, body *BlockNode) *Node {
+	if modifiers != node.modifiers || asteriskToken != node.AsteriskToken || name != node.name || typeParameters != node.TypeParameters || parameters != node.Parameters || returnType != node.Type || fullSignature != node.FullSignature || body != node.Body {
+		return updateNode(f.NewFunctionDeclaration(modifiers, asteriskToken, name, typeParameters, parameters, returnType, fullSignature, body), node.AsNode(), f.hooks)
 	}
 	return node.AsNode()
 }
 
 func (node *FunctionDeclaration) ForEachChild(v Visitor) bool {
 	return visitModifiers(v, node.modifiers) || visit(v, node.AsteriskToken) || visit(v, node.name) || visitNodeList(v, node.TypeParameters) ||
-		visitNodeList(v, node.Parameters) || visit(v, node.Type) || visit(v, node.WholeType) || visit(v, node.Body)
+		visitNodeList(v, node.Parameters) || visit(v, node.Type) || visit(v, node.FullSignature) || visit(v, node.Body)
 }
 
 func (node *FunctionDeclaration) VisitEachChild(v *NodeVisitor) *Node {
-	return v.Factory.UpdateFunctionDeclaration(node, v.visitModifiers(node.modifiers), v.visitToken(node.AsteriskToken), v.visitNode(node.name), v.visitNodes(node.TypeParameters), v.visitParameters(node.Parameters), v.visitNode(node.Type), v.visitNode(node.WholeType), v.visitFunctionBody(node.Body))
+	return v.Factory.UpdateFunctionDeclaration(node, v.visitModifiers(node.modifiers), v.visitToken(node.AsteriskToken), v.visitNode(node.name), v.visitNodes(node.TypeParameters), v.visitParameters(node.Parameters), v.visitNode(node.Type), v.visitNode(node.FullSignature), v.visitFunctionBody(node.Body))
 }
 
 func (node *FunctionDeclaration) Clone(f NodeFactoryCoercible) *Node {
-	return cloneNode(f.AsNodeFactory().NewFunctionDeclaration(node.Modifiers(), node.AsteriskToken, node.Name(), node.TypeParameters, node.Parameters, node.Type, node.WholeType, node.Body), node.AsNode(), f.AsNodeFactory().hooks)
+	return cloneNode(f.AsNodeFactory().NewFunctionDeclaration(node.Modifiers(), node.AsteriskToken, node.Name(), node.TypeParameters, node.Parameters, node.Type, node.FullSignature, node.Body), node.AsNode(), f.AsNodeFactory().hooks)
 }
 
 func (node *FunctionDeclaration) Name() *DeclarationName {
@@ -3753,7 +3753,7 @@ func (node *FunctionDeclaration) computeSubtreeFacts() SubtreeFacts {
 			propagateEraseableSyntaxListSubtreeFacts(node.TypeParameters) |
 			propagateNodeListSubtreeFacts(node.Parameters, propagateSubtreeFacts) |
 			propagateEraseableSyntaxSubtreeFacts(node.Type) |
-			propagateEraseableSyntaxSubtreeFacts(node.WholeType) |
+			propagateEraseableSyntaxSubtreeFacts(node.FullSignature) |
 			propagateSubtreeFacts(node.Body) |
 			core.IfElse(isAsync && isGenerator, SubtreeContainsForAwaitOrAsyncGenerator, SubtreeFactsNone) |
 			core.IfElse(isAsync && !isGenerator, SubtreeContainsAnyAwait, SubtreeFactsNone)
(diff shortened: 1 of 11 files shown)


testdata/baselines/reference/submodule/conformance/typeTagOnFunctionReferencesGeneric.js.diff
```diff
@@ -25,4 +25,25 @@
...
++out/typeTagOnFunctionReferencesGeneric.d.ts(6,37): error TS2304: Cannot find name 'T'.
++
++
++==== out/typeTagOnFunctionReferencesGeneric.d.ts (2 errors) ====
++    export type IFn = <T>(m: T) => T;
++    /**
++     * @typedef {<T>(m : T) => T} IFn
++     */
++    /**@type {IFn}*/
++    export declare function inJs(l: T): T;
```
jakebailey: Seems like declaration emit isn't pulling the type params out.

testdata/baselines/reference/submoduleAccepted/compiler/contextuallyTypedParametersOptionalInJSDoc.errors.txt.diff
```diff
@@ -0,0 +1,33 @@
...
+@@= skipped -21, +21 lines =@@
+        */
+       function self(a, b) {
+         acceptNum(b); // error
+-                  ~
+-!!! error TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
+-!!! error TS2345:   Type 'undefined' is not assignable to type 'number'.
+         self("");
++        ~~~~
++!!! error TS2554: Expected 2 arguments, but got 1.
```
sandersn: you can no longer make a parameter from a whole-type annotation optional by tacking on a nested `/** @param [b] */`, sorry, it's a horrible idea that should never have worked.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
```diff
@@ -2,46 +2,60 @@
...
 -!!! error TS8030: The type of a function declaration must match the function's signature.
++!!! error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
      function funcWithMoreParameters(more) {} // error
      
      /** @type {() => void} */
-@@= skipped -19, +17 lines =@@
+     const variableWithMoreParameters = function (more) {}; // error
+-          ~~~~~~~~~~~~~~~~~~~~~~~~~~
+-!!! error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+-!!! error TS2322:   Target signature provides too few arguments. Expected 1 or more, but got 0.
```
jakebailey: Hm, are we missing some checking here?

testdata/baselines/reference/submoduleAccepted/compiler/moduleExportsTypeNoExcessPropertyCheckFromContainedLiteral.types.diff
```diff
@@ -14,8 +14,8 @@
  const tseslint = require('./typescript-eslint.js');
 ->tseslint : typeof tseslint
 ->require('./typescript-eslint.js') : typeof tseslint
-+>tseslint : { config: (...configs: any[]) => void; }
-+>require('./typescript-eslint.js') : { config: (...configs: any[]) => void; }
++>tseslint : { config: (...configs: import("./typescript-eslint.js").Config[]) => void; }
```
sandersn, line 17: we now dereference the imported type instead of leaving it as `typeof eslint`.

testdata/baselines/reference/submoduleAccepted/conformance/asyncFunctionDeclaration16_es5.errors.txt.diff
```diff
@@ -37,13 +35,9 @@
 -!!! error TS1065: The return type of an async function or method must be the global Promise<T> type.
 -!!! related TS1055 /a.js:4:14: Type 'string' is not a valid async function return type in ES5 because it does not refer to a Promise-compatible constructor value.
      const f2 = async str => {
```
sandersn: this looks wrong. T1 is string -> string but f2 is string -> Promise<string>

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.types.diff
```diff
@@ -55,14 +27,26 @@
...
+ >variableWithMoreParameters : () => void
+->function (more) {} : (more: any) => void
++>function (more) {} : () => void
+ >more : any
+
+ /** @type {() => void} */
+ const arrowWithMoreParameters = (more) => {}; // error
+ >arrowWithMoreParameters : () => void
+->(more) => {} : (more: any) => void
++>(more) => {} : () => void
```
jakebailey: This to me is a regression; I would not expect variable declarations to have changed.

testdata/baselines/reference/submoduleAccepted/conformance/jsdocFunction_missingReturn.errors.txt.diff
```diff
@@ -5,9 +5,18 @@
...
+     /** @type {function(): number} */
 -                           ~~~~~~
 -!!! error TS2355: A function whose declared type is neither 'undefined', 'void', nor 'any' must return a value.
--    function f() {}
--    
-+<no content>
\ No newline at end of file
++               ~~~~~~~~
++!!! error TS2552: Cannot find name 'function'. Did you mean 'Function'?
++!!! related TS2728 lib.es5.d.ts:--:--: 'Function' is declared here.
```
jakebailey, line 18: Is this syntax no longer supported?

testdata/baselines/reference/submoduleAccepted/conformance/callbackTagNamespace.errors.txt.diff
```diff
@@ -0,0 +1,25 @@
...
++     * @param {Object} space - spaaaaaaaaace
++     * @param {Object} peace - peaaaaaaaaace
++     * @return {string | number}
++     */
++    var x = 1;
++    /** @type {NS.Nested.Inner} */
++               ~~
++!!! error TS2702: 'NS' only refers to a type, but is being used as a namespace here.
++               ~~~~~~~~~~~~~~~
++!!! error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
```
sandersn, line 21: our fancy span for the first error hurts us because the useless second error would get deduped if it used the whole span of the name.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
```diff
@@ -16,32 +19,33 @@
...
 -!!! error TS8030: The type of a function declaration must match the function's signature.
++!!! error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
      function funcWithMoreParameters(more) {} // error
      
      /** @type {() => void} */
-@@= skipped -19, +17 lines =@@
+@@= skipped -18, +18 lines =@@
      ({
        /** @type {() => void} */
        methodWithMoreParameters(more) {}, // error
```
sandersn, line 52: hmph. Still no error here.

internal/diagnostics/diagnostics_generated.go
```diff
@@ -3466,7 +3466,7 @@ var JSDoc_may_only_appear_in_the_last_parameter_of_a_signature = &Message{code:
 
 var JSDoc_param_tag_has_name_0_but_there_is_no_parameter_with_that_name_It_would_match_arguments_if_it_had_an_array_type = &Message{code: 8029, category: CategoryError, key: "JSDoc_param_tag_has_name_0_but_there_is_no_parameter_with_that_name_It_would_match_arguments_if_it_h_8029", text: "JSDoc '@param' tag has name '{0}', but there is no parameter with that name. It would match 'arguments' if it had an array type."}
 
-var The_type_of_a_function_declaration_must_match_the_function_s_signature = &Message{code: 8030, category: CategoryError, key: "The_type_of_a_function_declaration_must_match_the_function_s_signature_8030", text: "The type of a function declaration must match the function's signature."}
+var A_JSDoc_type_tag_on_a_function_must_have_a_signature_with_the_correct_number_of_arguments = &Message{code: 8030, category: CategoryError, key: "A_JSDoc_type_tag_on_a_function_must_have_a_signature_with_the_correct_number_of_arguments_8030", text: "A JSDoc '@type' tag on a function must have a signature with the correct number of arguments."}
```
jakebailey, line 3469: Missing the json update to match this?

testdata/baselines/reference/submoduleAccepted/conformance/errorOnFunctionReturnType.errors.txt.diff
```diff
@@ -1,50 +1,47 @@
...
--        ~~~~~~
--!!! error TS2322: Type 'string' is not assignable to type 'number'.
+         ~~~~~~
+@@= skipped -9, +7 lines =@@
      }
      
      var testPromise3 = /** @type {FunctionReturningPromise} */ function() {
 -                                                               ~~~~~~~~
--!!! error TS2355: A function whose declared type is neither 'undefined', 'void', nor 'any' must return a value.
++        ~~~~~~~~~~~~
```
jakebailey: Hm, curious that this error span is on the var when the JSDoc is on the function?

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
jakebailey: I think there is, it's just hidden in the diff?

internal/diagnostics/diagnostics_generated.go
jakebailey: I hid it when reviewing, ignore me.

testdata/baselines/reference/submoduleAccepted/conformance/jsdocFunction_missingReturn.errors.txt.diff
sandersn: yes, it's closure syntax

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
sandersn: no, it's the "Type '(more: any) => void' is not assignable to type '() => void'." one that's marked as deleted below wrt Strada.

testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
sandersn: yes, fixed now, although it's not contextually typed anymore so it gets the less-informative new error.

testdata/baselines/reference/submoduleAccepted/conformance/errorOnFunctionReturnType.errors.txt.diff
sandersn: `errorNode := core.OrElse(fn.Type(), fn)`, probably should be `fn.Type() || fn.WholeType || fn`

testdata/baselines/reference/submodule/conformance/typeTagOnFunctionReferencesGeneric.js.diff
sandersn: The new commit fixes this although I don't know if the code is correct.

paoloricciuti:
I love this and i would love to have the same in TS. Could this be a potential syntax?

```ts
function my_function(){

} typed MyFunction;
```

this would require a new construct but it seems very legible

testdata/baselines/reference/submoduleAccepted/conformance/errorOnFunctionReturnType.errors.txt.diff
```diff
@@ -58,28 +65,36 @@
...
--!!! error TS2322: Type '"asd"' is not assignable to type 'never'.
+         ~~~~~~
+@@= skipped -17, +14 lines =@@
      }
      
      var testNever3 = /** @type {FunctionReturningNever} */ function() {
 -                                                           ~~~~~~~~
--!!! error TS2534: A function returning 'never' cannot have a reachable end point.
++                                ~~~~~~~~~~~~~~~~~~~~~~
+ !!! error TS2534: A function returning 'never' cannot have a reachable end point.
```
jakebailey, line 97: The change you made moved the error from the var to the type node (better), but I think I was trying to get the error to move to where it used to be, on the function keyword.

jakebailey:
Other than the one error positioning thing, I think this seems good.

The CHANGES.md file also needs an update, but I think that PR was never merged so that's not in this PR.

sandersn:
Yeah, I got sidetracked from updating CHANGES.md.
I'll take another look at the position tomorrow--yesterday I was convinced that TS would already work the new way in certain circumstances, but I need to double-check.

sandersn:
@paoloricciuti I'm afraid that would be ambiguous with existing expression syntax.

Really what we should do is start a new issue to track this idea, but it sounds like the team has tried to come up with ideas in the past and failed to find any that more than 1 person likes.

testdata/baselines/reference/submoduleAccepted/conformance/errorOnFunctionReturnType.errors.txt.diff
sandersn: I'd like it to be like TS, which prefers putting the span on the type when present. (it was `OrElse(fn.Type(), fn)` before this PR) Admittedly, that makes more sense for the other 3 errors than this one.

sandersn:
I renamed WholeType to FullSignature based on ideas from the design meeting yesterday, so I think this is ready for final review. I decided to keep the new error span since it's closer to the intent of Typescript's span. Also I didn't want to write any more special-case code for functions (which may be how the TS code was arrived at as well)

jakebailey:
> I decided to keep the new error span since it's closer to the intent of Typescript's span.

Can you explain what you mean by this?

I'm surprised by the current location just because the error goes on the type and not on the thing causing the error. It feel like if we had a function returning the wrong type, so put the error on the return type instead of the returned expression.

sandersn:
Yes. The Typescript code is `OrElse(fn.Type(), fn)`, and the errors show up on the return type annotation. I think the closest thing with here is the full signature type annotation. I agree that it's strange for this particular error but it's better for the other 3. I think it was OK by mistake in Strada but when I weighed the time needed to figure out the special case code to be equivalent versus following the intent of the (simpler) TS code path, I decided on the simpler way.

jakebailey:
I think you might misunderstand my example? I'm comparing this PR's behavior to if you wrote:

```ts
function foo(): number {
    return "Hello!";
}
```

And the error were placed on `number` and not on the return expression, where the bug is.

I suppose if this PR had "real syntax", like:


```ts
function foo() :: Hmm {
    return "Hello!";
}
```

And the func were missing a parameter `Hmm` asks for or something, it'd be questionable where the error goes, but I just find it odd to have:

```js
/** @type {Hmm} */
function foo() {
    return "Hello!";
}
```

And have the error appear on `Hmm`, when it's the function that would need to change?

jakebailey (approved):
Implementation wise, this seems fine, though.

The only other thing I was wondering is if we really need this on FunctionLikeData and not just a specific node or two (does this actually work on accessors? Is it tested?), but it's probably okay.

sandersn:
it is *supposed* to work on accessors; I don't know if we have test coverage for it.

PR #1473, opened 2025-07-29
Move fourslash test utils into imported package

gabritto:
This temporarily avoids us having to copy `util_test.go` file with all the fourslash constants and helper functions that the tests use into the different test folders we have while we are still porting the fourslash tests (see https://github.com/microsoft/typescript-go/pull/1446/files/ebe23baa79b6a2aa05366691905154bd1fca565c#r2237553353)
This can be undone once all fourslash tests are in the same package.


jakebailey:
> This can be undone once all fourslash tests are in the same package.

Is this planned? I didn't think it was such a big deal to keep things split.

jakebailey:
The PR crashes my browser, but a gotcha is that `PtrTo` being exported means it'll show up in completions for unrelated packages, so hopefully that's not too annoying :|

gabritto:
> > This can be undone once all fourslash tests are in the same package.
> 
> Is this planned? I didn't think it was such a big deal to keep things split.

It's not necessary, it just doesn't make a lot of sense to keep the separate folders on the long run, but we don't have to do it.

jakebailey (approved):

PR #1475, opened 2025-07-29
Compute the snapshot changes in parallel 

sheetalkamat:


jakebailey (approved):
Seems fine, though at some level I wonder if all of this atomic-ness is worth it or if just a couple of mutexes on a shared structure would be sufficient and less fiddly.

sheetalkamat:
The compute happens in parallel so we cant do mutex on that shared structure so used sync map

PR #1476, opened 2025-07-29
Port over some baselined hover/quick info tests

DanielRosenwasser:
This change ports over fourslash tests for quick info.

Things are definitely not apples-to-apples given that hovers don't use display parts like quick info did. I did not do a very thorough diff, but some notable differences I picked up on:

- No returned ranges for which the current hover continues to be valid
- Some missing trailing JSDoc tags, or cases where `@see`/`@link` doesn't process text the same.
- `@inheritdoc` doesn't work
- Hover on keywords like `this`, `class`, and `default` (like an `export default`) are both not working
- Changes in descriptors (we used to say "getter" or "setter" and now we say "accessor")

image

image

image

There are only a few new failing tests, that I believe fail with something like the following:

```
Unexpected response type for quick info request: <nil>
```


DanielRosenwasser:
Also, I think these baselines are going in different directories than before - I hope that's fine?

internal/fourslash/fourslash.go
```diff
@@ -903,6 +903,99 @@ func (f *FourslashTest) VerifyBaselineGoToDefinition(
...
+		var result []string
+
+		if item.Contents.MarkupContent != nil {
+			result = strings.Split(item.Contents.MarkupContent.Value, "\n")
+		}
+		if item.Contents.String != nil {
+			result = strings.Split(*item.Contents.String, "\n")
+		}
+		if item.Contents.MarkedStringWithLanguage != nil {
+			result = appendLinesForMarkedStringWithLanguage(result, item.Contents.MarkedStringWithLanguage)
```
gabritto, line 968: We probably don't need this, I think this has been deprecated, but I guess it's fine to have it.

gabritto (approved):

PR #1477, opened 2025-07-30
Switch RecursionId to any

jakebailey:
This code is subtly wrong because some of the IDs are `uint64` and are being converted to `uint32`. In long running programs, this would manifest as very strange behavior.

I don't think we actually need anything special here; `any` is enough to do the right thing and avoids making that same mistake in the same amount of space as it would take to expand the struct to have a `uint64` value + padding.


internal/checker/relater.go
```diff
@@ -86,17 +86,13 @@ type ErrorOutputContainer struct {
...
-	RecursionIdKindType
-)
-
 type RecursionId struct {
-	kind RecursionIdKind
-	id   uint32
+	value any
+}
+
+// This function exists to constraint the types of values that can be used as recursion IDs.
```
jakebailey:
```suggestion
// This function exists to constrain the types of values that can be used as recursion IDs.
```

PR #1481, opened 2025-07-30
Fix parser fuzzing crashes

jakebailey:
I added a Fuzz test a while ago but never sent any fixes; I let Agent mode go on these, and it fixed a good few crashes.


jakebailey:
Still WIP, I'll close this for now.

PR #1484, opened 2025-07-30
Build mode

sheetalkamat:


internal/compiler/fileloader.go
```diff
@@ -610,19 +613,19 @@ func (p *fileLoader) pathForLibFile(name string) *LibFile {
 
 	path := tspath.CombinePaths(p.defaultLibraryPath, name)
 	replaced := false
-	if p.opts.Config.CompilerOptions().LibReplacement.IsTrue() {
+	if p.opts.Config.CompilerOptions().LibReplacement.IsTrue() && name != "lib.d.ts" {
```
jakebailey, line 616: Huh, I must have missed this special case

internal/execute/outputs.go
```diff
@@ -32,17 +32,16 @@ func getFormatOptsOfSys(sys System) *diagnosticwriter.FormattingOptions {
 type diagnosticReporter = func(*ast.Diagnostic)
 
 func quietDiagnosticReporter(diagnostic *ast.Diagnostic) {}
-func createDiagnosticReporter(sys System, options *core.CompilerOptions) diagnosticReporter {
+func createDiagnosticReporter(sys System, w io.Writer, options *core.CompilerOptions) diagnosticReporter {
```
jakebailey: Seems suspect to take both System and a writer, but I do see why that changed here. It sort of makes me wonder if `System` should even have the writer on it.

internal/execute/tscnoemit_test.go
jakebailey, line 1: I feel like we're ending up with a lot of test files in this package; are they all long enough that concating a few would be impossible?

Maybe it's fine, just sort of surprising me. I know many projects with more files, though typically it's also because they have a lot of non-test files.

internal/incremental/program.go
```diff
@@ -281,18 +289,39 @@ func (p *Program) emitBuildInfo(ctx context.Context, options compiler.EmitOption
...
+		return
+	}
+	if len(program.GetConfigFileParsingDiagnostics()) > 0 ||
+		len(program.GetSyntacticDiagnostics(ctx, nil)) > 0 ||
+		len(program.GetProgramDiagnostics()) > 0 ||
+		len(program.GetBindDiagnostics(ctx, nil)) > 0 ||
+		len(program.GetOptionsDiagnostics(ctx)) > 0 ||
+		len(program.GetGlobalDiagnostics(ctx)) > 0 {
+		p.snapshot.hasErrors = core.TSTrue
+		// Dont need to encode semantic errors state!!
```
jakebailey: Is this a TODO?

internal/parser/parser.go
```diff
@@ -6390,13 +6390,13 @@ func (p *Parser) processPragmasIntoFields(context *ast.SourceFile) {
 				})
 			case libOk:
 				context.LibReferenceDirectives = append(context.LibReferenceDirectives, &ast.FileReference{
-					TextRange: types.TextRange,
+					TextRange: lib.TextRange,
```
jakebailey, line 6393: Oops

internal/testutil/incrementaltestutil/readablebuildinfo.go
```diff
@@ -174,6 +182,27 @@ type readableBuildInfoEmitSignature struct {
...
+	Root     string
+}
+
+func (b *readableBuildInfoResolvedRoot) MarshalJSON() ([]byte, error) {
+	return json.Marshal([2]string{b.Resolved, b.Root})
+}
+
+func (b *readableBuildInfoResolvedRoot) UnmarshalJSON(data []byte) error {
+	var resolvedAndRoot *[2]string
+	if err := json.Unmarshal(data, &resolvedAndRoot); err == nil {
```
jakebailey: Why is resolvedAndRoot a pointer here? It seems like it should just be `[2]string`, since you're basically just passing in `**[2]string` and then assuming it decodes. It seems like maybe it's intending to handle null, but then it unconditionally uses resolvedAndRoot anyway.

internal/testutil/incrementaltestutil/readablebuildinfo.go
```diff
@@ -174,6 +182,27 @@ type readableBuildInfoEmitSignature struct {
...
+		*b = readableBuildInfoResolvedRoot{
+			Resolved: resolvedAndRoot[0],
+			Root:     resolvedAndRoot[1],
+		}
+		return nil
+	}
+	return fmt.Errorf("invalid BuildInfoResolvedRoot: %s", data)
```
jakebailey: Nit, but typically this code is written the opposite way around, with the error check first.

internal/tsoptions/declscompiler.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var OptionsDeclarations = slices.Concat(commonOptionsWithBuild, optionsForCompiler)
 
-var optionsForCompiler = []*CommandLineOption{
+var commonOptionsWithBuild = []*CommandLineOption{
 	//******* commandOptionsWithoutBuild *******
```
jakebailey: Out of date comment maybe?

internal/execute/solutionbuilder.go
```diff
@@ -0,0 +1,690 @@
...
+		for _, downstream := range task.downStream {
+			downstream.status <- status
+		}
+
+		// Wait for previous build task to complete reporting status, errors etc
+		if task.previousTaskReporter != nil {
+			<-task.previousTaskReporter
+		}
+		taskReporter.report(s, path, buildResult)
+		task.reporter <- taskReporter
```
jakebailey: Would it be simpler to just make the reporter have a mutex, rather than trying to hand things around?

internal/execute/solutionbuilder.go
```diff
@@ -0,0 +1,690 @@
...
+		upStreamStatus := make([]*upToDateStatus, len(task.upStream))
+		for i, upstream := range task.upStream {
+			if upstream.status != nil {
+				upStreamStatus[i] = <-upstream.status
+			}
+		}
+
+		status, taskReporter := s.buildProject(config, path, task, upStreamStatus)
+		for _, downstream := range task.downStream {
+			downstream.status <- status
```
jakebailey: I feel like there must be a way to do this without explicitly having upstream/downstream channels for communication, passing statuses around, etc.

The way I'd probably do this is to instead have the builder hold all of the projects being built, then each project calls a `wait` method on each of the upstreams. The wait method can do `<-project.done`, and then return the status (since the project is now "done" and all values would be set). This saves having to explicitly send a status to everyone, and is a bit easier to reason about.

To mark the project as done, you can close the channel, which will then trigger all receives at once.

internal/execute/outputs.go
sheetalkamat: i think its ok as it finally goes onto `system.writer()` only , its just we create string buffer to get output and when its ready to stream to output is when we stream it to system.writer()

internal/execute/solutionbuilder.go
sheetalkamat: i can move around stuff but thats essentially what i am doing.

Order generator sets up upsteam and downstream pointers (handling circularity) 
BuildProjects -> starts projects ->

Each project build entails -> starting upstream projects , waiting on them to complete, then doing its own job and reporting errors and then unblocking downstream ones 

What i understand you mean is to encapsulate these things in task itself as part of `wait` and `done` methods ?

jakebailey: Well, if circularity wasn't a thing, what I'm suggesting is to basically just spawn all of the tasks at once, and they "await" all of their dependencies, and then it all finishes when the root tasks are done. If each project only exists once, there isn't duplicate work, and multiple dependants of the same project both become unblocked at the same time.

But if circular is possible (I thought we banned that), then yeah you need some other method to do it.

But otherwise, it's sort of like how hereby does tasks with deps; they're just funcs awaiting other funcs.

Maybe a more complicated way would work fine, but I somewhat suspect it'd be easier to do another way, and potentially allow us to say, limit the concurrency to N projects at once or something.

(I could write a quick example code if that's helpful; sort of hard to describe in words)

sheetalkamat: in my example i am spawning all builds at a time and multiple do get ready.. (look into `startProjectBuild` Thats why i have circularity dependency analysis through `analizeConfig` which sets up downstream and upstream based on circularity (we report error if "circular" is not set on the reference path)  after this startProject goes and spanws all the builds and each project waits on its upstream and then builds and unblocks all downstreams .. 

If you run this branch (and instead of writing status to taskReporter -> write to sys.writer) and use it toi build strada source you will see that order is not determinitistic because "compiler" unblocks mutiple projects 

Happy to reorganize this if there is better way but i thought chanels were best to comunicate the completion so did that. (may be we are saying same thing just organized it differenty)

sheetalkamat:
Refactoring further into two sub packages build and tsc so its easier to look at files

sheetalkamat:
Looking into race

andrewbranch (approved):
I feel like some explanatory comments in `execute` would be helpful, but I don't see anything concerning here. Tried it out on arethetypeswrong and it's 4.5x faster :)

jakebailey (approved):
I am happy to report straight up 10x on dt-tools 😄 

image

PR #1485, opened 2025-07-31
Add baseline tests from signature help

DanielRosenwasser:
This change ports over Strada's baselining signature help tests. There is only one new test failure which has the following problem:

```
fourslash.go:1040: At position (Ln 0, Col 0): Unexpected response type for signature help request: <nil>
```

Beyond that, signature help in LSP is surprisingly different in a few ways, and so reconstructing similar visualization is going to be weird. Stuff worth calling out:

- There is no section for tags. Everything just appears in docs. (expected)
- We currently don't return any docs! We should do that eventually.
- There is no notion of an "applicable span" so that the editor can avoid retriggers. Our testing infrastructure used to leverage this.
- Editors like VS Code bold out the signature help parameter based on the label of the parameter itself. There is a "tuple" version that theoretically avoids a string search, but we don't use the "tuple" version of parameter labels for signature help. This is sketchy for *reasons*...

  image

  I am currently testing things in a very lazy way where these are indistinguishable. I probably should do it a bit differently.


internal/fourslash/fourslash.go
```diff
@@ -998,6 +998,136 @@ func appendLinesForMarkedStringWithLanguage(result []string, ms *lsproto.MarkedS
...
+				// parameter text within the original string.
+				activeParamLabel := ""
+				if activeParam.Label.String != nil {
+					activeParamLabel = *activeParam.Label.String
+				} else if activeParam.Label.Tuple != nil {
+					activeParamLabel = signatureLine[(*activeParam.Label.Tuple)[0]:(*activeParam.Label.Tuple)[1]]
+				} else {
+					t.Fatal("Unsupported param label kind.")
+				}
+				signatureLine = strings.Replace(signatureLine, activeParamLabel, "**"+activeParamLabel+"**", 1)
```
DanielRosenwasser, line 1082: This should really only be done if we don't come back with the tuple representation.

internal/fourslash/fourslash.go
```diff
@@ -998,6 +998,136 @@ func appendLinesForMarkedStringWithLanguage(result []string, ms *lsproto.MarkedS
...
+			if activeParamIndex >= 0 && activeParamIndex < len(*sig.Parameters) {
+				activeParam := (*sig.Parameters)[activeParamIndex]
+
+				// Get the parameter label and bold the
+				// parameter text within the original string.
+				activeParamLabel := ""
+				if activeParam.Label.String != nil {
+					activeParamLabel = *activeParam.Label.String
+				} else if activeParam.Label.Tuple != nil {
+					activeParamLabel = signatureLine[(*activeParam.Label.Tuple)[0]:(*activeParam.Label.Tuple)[1]]
```
DanielRosenwasser, line 1078: This is slightly wrong because these are supposed to be UTF-16 offsets. But it's not really "testable' today because we don't even come back with this data.

jakebailey: You could utf16.Encode + slice + utf16.Decode

DanielRosenwasser:
The stuff I need to do to get signature offsets is absolutely gross, so I would like to get a review once I fix the conflicts.

jakebailey (approved):

PR #1486, opened 2025-07-31
Switch to jsonv2 wholesale

jakebailey:
jsonv2 is all around faster and has better defaults. It's not yet fully enabled in Go, but we've been using it in some places already. I think it's worth switching to already for the perf.

This introduces a new `github.com/microsoft/typescript-go/internal/json` package, which should be used to provide JSON features; it wraps `github.com/go-json-experiment/json`, ensuring we can audit all of the uses.

Many things had to change to make this all work, but the result is faster code all around.

Part of this is the full switching to `omitzero` instead of `omitempty`, which has a good result on LSP stuff.


testdata/baselines/reference/fourslash/hover/CompletionDetailsOfContextSensitiveParameterNoCrash.baseline
```diff
@@ -102,7 +102,7 @@
         "character": 60
       },
       "Name": "",
-      "Data": null
+      "Data": {}
```
jakebailey, line 105: This is correct; we don't want to encode an empty struct as null.

testdata/baselines/reference/fourslash/hover/DeprecatedInheritedJSDocOverload.baseline
```diff
@@ -41,12 +41,12 @@
...
       },
       "Name": "",
-      "Data": null
+      "Data": {}
     },
     "item": {
       "contents": {
         "kind": "markdown",
-        "value": "```tsx\n(method) ThingWithDeprecations.subscribe(observer?: PartialObserver\u003cvoid\u003e): Subscription\n```\n"
+        "value": "```tsx\n(method) ThingWithDeprecations.subscribe(observer?: PartialObserver<void>): Subscription\n```\n"
```
jakebailey, line 49: This has no effect on clients who decode this just fine, but the HTML escaping junk going away means our baseline looks nicer.

testdata/baselines/reference/tsc/incremental/change-to-modifier-of-class-expression-field-with-declaration-emit-enabled.js
```diff
@@ -91,24 +91,24 @@ export {};
...
+      "version": "b19bd7c47bb2cad085a470ada95e9d46-/// <reference no-default-lib=\"true\"/>\ninterface Boolean {}\ninterface Function {}\ninterface CallableFunction {}\ninterface NewableFunction {}\ninterface IArguments {}\ninterface Number { toExponential: any; }\ninterface Object {}\ninterface RegExp {}\ninterface String { charAt: any; }\ninterface Array<T> { length: number; [n: number]: T; }\ninterface ReadonlyArray<T> {}\ninterface SymbolConstructor {\n    (desc?: string | number): symbol;\n    for(name: string): symbol;\n    readonly toStringTag: symbol;\n}\ndeclare var Symbol: SymbolConstructor;\ninterface Symbol {\n    readonly [Symbol.toStringTag]: string;\n}\ndeclare const console: { log(msg: any): void; };\ntype ReturnType<T extends (...args: any) => any> = T extends (...args: any) => infer R ? R : any;\ntype InstanceType<T extends abstract new (...args: any) => any> = T extends abstract new (...args: any) => infer R ? R : any;",
```
jakebailey, line 94: Smaller buildinfos because the escaping goes away.

jakebailey:
FWIW I was going to use https://pkg.go.dev/github.com/go-json-experiment/jsonsplit, but then I realized it uses `expvar`, which then uses `net/http`, which then makes our binary not static, pulls in crypto, etc. So, just switching to it entirely.

go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/microsoft/typescript-go
 
-go 1.24.0
+go 1.24.2
+
+toolchain go1.24.5
```
jakebailey: Did not mean to leave this in

.golangci.yml
```diff
@@ -28,6 +29,7 @@ linters:
     - inamedparam
     - ineffassign
     - intrange
+    - importas
```
jakebailey: This is leftover

jakebailey:
```suggestion
```

Issue #774, opened 2025-04-08
Parser doesn't parse an EOF token for source files


sandersn:
An EOF token is needed to attach jsdoc at the end of the file to:

```js
/** @typedef {number} A */
var unrelated;
/** @typedef {number} B */
```

Currently, `B` is not parsed because there's no node to attach the jsdoc to.

sandersn:
Found while reviewing #610, thanks @jakebailey

PR #1491, opened 2025-07-31
fix: emit end of file comments

chengcyber:
### Summary

Fixes https://github.com/microsoft/typescript-go/issues/774

### Detail

The end of file comments won't be emitted correctly now. In this PR, excludes the length of the additional rephrased list from the end position of statements. This behavior now aligns with the TypeScript implementation in the TS version.

### Test

Added two test cases

### DIFF
--- a/testdata/tests/cases/compiler/emitEndOfFileJSDocComments.ts
+++ b/testdata/tests/cases/compiler/emitEndOfFileJSDocComments.ts
@@ -1,0 +1,8 @@
+// @allowJs: true
+// @checkJs: true
+// @outDir: ./out
+// @filename: emitEndOfFileJSDocComments.js
+
+/** @typedef {number} A */
+var unrelated;
+/** @typedef {number} B */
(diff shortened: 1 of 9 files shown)


testdata/tests/cases/compiler/emitEndOfFileJSDocComments.ts
```diff
@@ -0,0 +1,8 @@
+// @allowJs: true
+// @checkJs: true
+// @outDir: ./out
+// @filename: emitEndOfFileJSDocComments.js
+
+/** @typedef {number} A */
+var unrelated;
+/** @typedef {number} B */
```
sandersn, line 8: I think we have enough test coverage that this new test is not needed.

sandersn (changes requested):
Thanks for the fix. Please remove the new test, since I couldn't see that it adds anything to the existing tests.

testdata/tests/cases/compiler/emitEndOfFileJSDocComments.ts
jakebailey: I feel like it's worth having given the other tests are sort of random?

sandersn: Fair enough.

sandersn (approved):

PR #1502, opened 2025-08-01
Optimize LSP marshalling via jsonv2 methods

jakebailey:
- Replace `RawMessage` with `Value` (cleanup)
- Prevent double decode when checking required fields by using new `UnmarshalJSONFrom` and generated checks.
- Eliminate redundant unmarshal methods for enums; the defaults are fine as they are just values.
- Use `MarshalJSONTo` and `UnmarshalJSONFrom` in union/literal marshalling to avoid extra `[]byte` encoding; the former never produces an intermediary `[]byte`, and the latter slices out of the input buffer (zero copy).
- Use pointers for union type marshalling, now that it behaves consistently in v2.

### DIFF
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -831,9 +831,13 @@ function generateCode() {
         writeLine(`}`);
         writeLine("");
 
-        writeLine(`func (o *${name}) UnmarshalJSON(data []byte) error {`);
-        writeLine(`\tif string(data) != \`${jsonValue}\` {`);
-        writeLine(`\t\treturn fmt.Errorf("invalid ${name}: %s", data)`);
+        writeLine(`func (o *${name}) UnmarshalJSONFrom(dec *jsontext.Decoder) error {`);
+        writeLine(`\tv, err := dec.ReadValue();`);
+        writeLine(`\tif err != nil {`);
+        writeLine(`\t\treturn err`);
+        writeLine(`\t}`);
+        writeLine(`\tif string(v) != \`${jsonValue}\` {`);
+        writeLine(`\t\treturn fmt.Errorf("expected ${name} value %s, got %s", \`${jsonValue}\`, v)`);
         writeLine(`\t}`);
         writeLine(`\treturn nil`);
         writeLine(`}`);
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -754,10 +754,14 @@ function generateCode() {
 
         // Determine if this union contained null (check if any member has containedNull = true)
         const unionContainedNull = members.some(member => member.containedNull);
-        const assertionFunc = unionContainedNull ? "assertAtMostOne" : "assertOnlyOne";
+        if (unionContainedNull) {
+            write(`\tassertAtMostOne("more than one element of ${name} is set", `);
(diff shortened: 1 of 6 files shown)


internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -519,35 +520,62 @@ function generateCode() {
...
+                writeLine(`\t\tcase "${prop.name}":`);
+                if (!prop.optional) {
+                    writeLine(`\t\t\tseen${titleCase(prop.name)} = true`);
+                }
+                writeLine(`\t\t\tif err := json.UnmarshalDecode(dec, &s.${titleCase(prop.name)}); err != nil {`);
+                writeLine(`\t\t\t\treturn err`);
+                writeLine(`\t\t\t}`);
+            }
+
+            writeLine(`\t\t}`);
```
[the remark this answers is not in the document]
jakebailey: Not handling unknown props is intentional; they are always ignored. I will add a comment.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -808,14 +839,24 @@ function generateCode() {
...
-        writeLine(`\t\treturn fmt.Errorf("invalid ${name}: %s", data)`);
+        writeLine(`var _ json.UnmarshalerFrom = &${name}{}`);
+        writeLine("");
+
+        writeLine(`func (o *${name}) UnmarshalJSONFrom(dec *jsontext.Decoder) error {`);
+        writeLine(`\tv, err := dec.ReadValue();`);
+        writeLine(`\tif err != nil {`);
+        writeLine(`\t\treturn err`);
+        writeLine(`\t}`);
+        writeLine(`\tif string(v) != \`${jsonValue}\` {`);
```
jakebailey, line 858: This is wrong, since the value can contain whitespace; need to fix.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -771,13 +796,19 @@ function generateCode() {
...
 
-        // Handle null case only for unions that can contain null
+        writeLine("\tdata, err := dec.ReadValue()");
+        writeLine("\tif err != nil {");
+        writeLine("\t\treturn err");
+        writeLine("\t}");
+
         if (unionContainedNull) {
-            writeLine(`\t// Handle null case`);
             writeLine(`\tif string(data) == "null" {`);
```
jakebailey, line 812: this is also wrong, blah

internal/lsp/lsproto/_generate/generate.mts
jakebailey: I am wrong; ReadValue always trims. https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20250725192818-e39067aee2d2/jsontext#Decoder.ReadValue

internal/lsp/lsproto/_generate/generate.mts
jakebailey: I am wrong; ReadValue always trims. https://pkg.go.dev/github.com/go-json-experiment/json@v0.0.0-20250725192818-e39067aee2d2/jsontext#Decoder.ReadValue

jakebailey:
One last change, sorry; remembered that marshalling methods on pointers now behave well in v2, so I was able to eliminate some extra copying going on in union marshalling.

PR #1504, opened 2025-08-01
Use json v2 directly, move utils to new jsonutil package

jakebailey:
Originally the `json` package in our repo contained some shims to adapt behavior, but then I moved everything to just be v2. Now the package serves no purpose, except for a couple of pretty printing helpers.

~Those helpers needing to exist seems like a bug to me; probably worth reporting upstream.~


jakebailey:
Ah, the behavior difference is that setting Prefix or Indent also enables multiline printing in v2. It's actually a bug in our code that we are attempting to stringify with empty prefix and indent.

sheetalkamat (approved):

PR #1505, opened 2025-08-02
Snapshot LSP

andrewbranch:
> [!NOTE]
> Diagrams are generated by Copilot with Python; they’re pretty rough but get the idea across. I used them for an internal team presentation and figured they’re still better than nothing for this.

## Summary

This PR replaces the project system backing the LSP server with an immutable snapshot-based architecture.

## Problem Statement

The original LSP implementation here in typescript-go inherited microsoft/TypeScript's architecture, which relied on mutable data structures suitable for synchronous operation. However, as we started to parallelize request handling, this approach became unsustainable. We were starting to spend a lot of time [chasing down data races](https://github.com/microsoft/typescript-go/issues?q=is%3Aissue%20state%3Aclosed%20%20label%3A%22Data%20race%22), for which the solution was often tacking on yet another mutex for an individual component.

## Design Goals

The new architecture addresses these issues through immutable snapshots that can be safely read by multiple goroutines and efficiently cloned to create new state. This enables:

- Elimination of data races with less complex mutex coordination
- Support for multiple concurrent clients accessing the same workspace state (e.g. LSP, API, MCP)
- Simplified reasoning about state changes and their effects

### Core Concepts

The system centers around **snapshots** - immutable representations of LSP server state at a point in time. Snapshots can be read concurrently without coordination and cloned to produce new snapshots when state changes are needed.

While TypeScript programs were already immutable, additional components needed to be refactored for immutability:

- **File state**: Text content, line maps, and open/closed state from client
- **Project state**: Loaded projects, file associations, and program freshness tracking  
- **Config state**: tsconfig existence and freshness of parsed options and files/include/exclude evaluation

### System Structure

LSP Server Hierarchy

The architecture separates concerns between mutable session management and immutable snapshot state:

**Session Layer (Mutable)**
- File change and request handlers called by LSP server
- Live filesystem view and overlay state management
- Current snapshot reference and lifecycle management
- Ref-counting AST caches
- Side effect triggering after snapshot transition

**Snapshot Layer (Immutable)**  
- Cached filesystem state at snapshot creation time
- Open file state snapshot
- Complete project collection with programs and configurations

### State Transitions

Snapshot Creation Process

State changes occur through snapshot cloning:
1. Session accumulates pending changes from LSP events and file watching
2. Changes combine with current open file state to create new overlay state and summary of changes
3. Builders handle changes with copy-on-write strategy against previous snapshot state, and finalize into new snapshot
4. Session triggers logging, file watcher updates, ATA updates, and diagnostics refresh requests by comparing previous snapshot with newly adopted snapshot.

### Request Processing

LSP Request Flow

When an LSP request needs to invoke a language service operation for a document, we first check if there are any pending changes that need to be applied (e.g. file watchers have been triggered, open files have been changed, or an ATA request has completed). If so, the Session's current snapshot is cloned to incorporate the pending changes while simultaneously ensuring that the default project for the requested file is loaded and up-to-date. If there are no pending changes, we check if the Session's current snapshot is capable of serving that request (that is, if a default project can be found for that file and the project is up-to-date). If not, we clone the snapshot, in this case with no file changes, but a request to load or update the default project for the requested file. Finally, we return a language service that uses the project from the latest snapshot.

## Notable behavior changes

- When searching project references to find a default project for a file, the previous default project finder would race on returning a referenced project that contains the target file, but it would also continue loading every referenced project recursively, and keep these projects open until the next file open request. The new system guarantees that when multiple references contain the target file, the one returned is the one at the lowest index in the searched tsconfig.json's `references` array, and it terminates work as soon as higher priority matches have been ruled out. In order to support this early termination of work, we can no longer keep every project we loaded open, since that set is nondeterministic. Instead, we keep open only the projects that led us to the one that ultimately contained the target file. Example:
  - Target file: `/workspace/packages/foo/test.ts`
  - Load `/workspace/packages/foo/tsconfig.json`, doesn't contain the file, has two references:
    - `/workspace/packages/bar/tsconfig.json`
    - `/workspace/packages/baz/tsconfig.json`
  - Load the two references in parallel; neither contains the file
  - Load `/workspace/tsconfig.json`, doesn't contain the file, has 1,000 references
  - Load all 1,000 references in a parallel work group; none contains the file, but collects 100 additional referenced projects. One of the 1,000 projects called `tests.tsconfig.json` had a reference of `foo.test.tsconfig.json`
  - Kick off loading for all 100 references in a parallel work group; `foo.test.tsconfig.json` contains the file at index 50
  - Wait for these 100 jobs to be done; ones with index > 50 are skipped.
  - No other project contains the file, so `foo.test.tsconfig.json` is the default project.
  - Projects we keep open after this process:
    - `/workspace/foo.test.tsconfig.json` (the default project)
    - `/workspace/tests.tsconfig.json`
    - `/workspace/tsconfig.json`
    - `/workspace/packages/foo/tsconfig.json`
  - The last 3 will be closed after the next file open, assuming they don't contribute to finding a default project for another file in the same way again.
  - Projects we _used_ to keep open after this process: all 1,100+ we loaded
- There is only one inferred project. Strada had the capability to have multiple inferred projects for different workspace roots, and some of this infrastructure was ported into Corsa, but it was only used in tests, because LSP differs significantly from TS Server in how files get assigned to inferred projects. If multiple inferred projects are desired in Corsa, the feature needs to be redesigned from the ground up.

### DIFF
--- a/internal/project/ata/ata.go
+++ b/internal/project/ata/ata.go
@@ -191,6 +191,7 @@ func (ti *TypingsInstaller) installTypings(
 			typingFile := ti.typingToFileName(resolver, packageName)
 			if typingFile == "" {
 				logger.Log(fmt.Sprintf("ATA:: Failed to find typing file for package '%s'", packageName))
+				ti.missingTypingsSet.Store(packageName, true)
 				continue
 			}
 
@@ -433,9 +434,21 @@ func (ti *TypingsInstaller) processCacheLocation(projectID string, fs vfs.FS, lo
 				}
 				typingFile := ti.typingToFileName(resolver, packageName)
 				if typingFile == "" {
+					ti.missingTypingsSet.Store(packageName, true)
 					continue
 				}
-				newVersion := semver.MustParse(npmLockValue.Version)
+				if existingTypingFile, existingTypingsFilePresent := ti.packageNameToTypingLocation.Load(packageName); existingTypingsFilePresent {
+					if existingTypingFile.TypingsLocation == typingFile {
+						continue
+					}
+					logger.Log("ATA:: New typing for package " + packageName + " from " + typingFile + " conflicts with existing typing file " + existingTypingFile.TypingsLocation)
+				}
+				logger.Log("ATA:: Adding entry into typings cache: " + packageName + " => " + typingFile)
+				version := npmLockValue.Version
+				if version == "" {
+					continue
+				}
+				newVersion := semver.MustParse(version)
 				newTyping := &CachedTyping{TypingsLocation: typingFile, Version: &newVersion}
 				ti.packageNameToTypingLocation.Store(packageName, newTyping)
 			}


internal/compiler/program.go
```diff
@@ -255,13 +262,57 @@ func equalCheckJSDirectives(d1 *ast.CheckJsDirective, d2 *ast.CheckJsDirective)
...
+func (p *Program) CommandLine() *tsoptions.ParsedCommandLine { return p.opts.Config }
+func (p *Program) Host() CompilerHost                        { return p.opts.Host }
 func (p *Program) GetConfigFileParsingDiagnostics() []*ast.Diagnostic {
 	return slices.Clip(p.opts.Config.GetConfigFileParsingDiagnostics())
 }
 
+// ExtractUnresolvedImports returns the unresolved imports for this program.
+// The result is cached and computed only once.
+func (p *Program) ExtractUnresolvedImports() collections.Set[string] {
+	if p.unresolvedImports != nil {
```
jakebailey: This code will have to be moved below the Once (or removed) in order to be race-free. The perf will be fine; Once is inlined and will do a quick atomic check before locking to do the work.

internal/core/core.go
```diff
@@ -587,3 +588,37 @@ func ConcatenateSeq[T any](seqs ...iter.Seq[T]) iter.Seq[T] {
...
+	}
+
+	for k, v2 := range m2 {
+		if _, ok := m1[k]; !ok {
+			onAdded(k, v2)
+		}
+	}
+}
+
+func CopyMap[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](dst M1, src M2) map[K]V {
```
jakebailey: I feel like this might need to be called something else; was surprised to see it writing into a dest, but maybe Strada had a similar helper anyway?

(`CopyMapInto`?)

internal/core/workgroup.go
```diff
@@ -86,3 +89,37 @@ func (w *singleThreadedWorkGroup) pop() func() {
...
 }
+
+// ThrottleGroup is like errgroup.Group but with global concurrency limiting via a semaphore.
+type ThrottleGroup struct {
+	semaphore chan struct{}
+	group     *errgroup.Group
+}
+
+// NewThrottleGroup creates a new ThrottleGroup with the given context and semaphore for concurrency limiting.
+func NewThrottleGroup(ctx context.Context, semaphore chan struct{}) *ThrottleGroup {
```
jakebailey, line 100: I haven't gotten to where this is used yet, but I was surprised to see the channel be a parameter and not just created here.

internal/testutil/baseline/baseline.go
```diff
@@ -191,24 +189,16 @@ func writeComparison(t *testing.T, actualContent string, local, reference string
...
-		relReference = tspath.NormalizeSlashes(relReference)
-
-		relLocal, err := filepath.Rel(repo.RootPath, local)
-		assert.NilError(t, err)
-		relLocal = tspath.NormalizeSlashes(relLocal)
-
 		if _, err := os.Stat(reference); err != nil {
 			if comparingAgainstSubmodule {
-				t.Errorf("the baseline file %s does not exist in the TypeScript submodule", relReference)
+				t.Errorf("the baseline file %s does not exist in the TypeScript submodule", reference)
```
andrewbranch, line 195: I changed this so the path would always be linkified in VS Code output windows.

internal/ls/converters.go
```diff
@@ -83,46 +83,7 @@ func LanguageKindToScriptKind(languageID lsproto.LanguageKind) core.ScriptKind {
 }
 
 func DocumentURIToFileName(uri lsproto.DocumentUri) string {
```
jakebailey: I assume we'll inline this at some point?

testdata/baselines/reference/fourslash/findAllRef/FindAllRefsThisKeyword.baseline.jsonc
andrewbranch, line 1: I didn't mean to commit this

internal/lsp/server.go
```diff
@@ -762,14 +732,16 @@ func (s *Server) handleCompletion(ctx context.Context, params *lsproto.Completio
 		&ls.UserPreferences{})
 }
 
-func (s *Server) handleCompletionItemResolve(ctx context.Context, params *lsproto.CompletionItem) (lsproto.CompletionResolveResponse, error) {
+func (s *Server) handleCompletionItemResolve(ctx context.Context, params *lsproto.CompletionItem, recover func()) (lsproto.CompletionResolveResponse, error) {
```
jakebailey: I think it'd be a good idea to not name the param `recover`, since it shadows a builtin. (Surprised the linter does not get mad)

internal/project/ata/typesmap.go
```diff
@@ -1,4 +1,4 @@
-package project
+package ata
```
jakebailey, line 1: Thanks for moving this.

internal/project/dirty/box.go
```diff
@@ -0,0 +1,64 @@
+package dirty
```
jakebailey, line 1: Very strong naming, I approve

internal/project/dirty/interfaces.go
```diff
@@ -0,0 +1,21 @@
+package dirty
+
+type cloneable struct{}
```
jakebailey: Is this used anywhere? Or is it just for the type assertion?

internal/project/dirty/syncmap.go
```diff
@@ -0,0 +1,281 @@
...
+		entry.mu.Lock()
+		defer entry.mu.Unlock()
+	}
+	if !entry.dirty {
+		entry.value = entry.value.Clone()
+		entry.dirty = true
+	}
+	if loaded {
+		// !!! There are now two entries for the same key...
+		// for now just sync the values.
```
jakebailey: Is this something that needs to be fixed before the PR goes in?

internal/project/parsecache.go
```diff
@@ -0,0 +1,99 @@
...
+	defer entry.mu.Unlock()
+	if !loaded || entry.hash != fh.Hash() {
+		// Reparse the file if the hash has changed, or parse for the first time.
+		entry.sourceFile = parser.ParseSourceFile(opts, fh.Content(), scriptKind)
+		entry.hash = fh.Hash()
+	}
+	return entry.sourceFile
+}
+
+func (c *ParseCache) Ref(file *ast.SourceFile) {
```
jakebailey, line 62: I haven't seen where this is used, but I'm assuming this has ref counting because it's the cache for the actual server, and not the one used for the test harness that should cache foreverish?

internal/project/programcounter.go
```diff
@@ -0,0 +1,33 @@
...
+type programCounter struct {
+	refs collections.SyncMap[*compiler.Program, *atomic.Int32]
+}
+
+func (c *programCounter) Ref(program *compiler.Program) {
+	counter, _ := c.refs.LoadOrStore(program, &atomic.Int32{})
+	counter.Add(1)
+}
+
+func (c *programCounter) Deref(program *compiler.Program) bool {
```
jakebailey, line 19: The naming seems potentially inconsistent; Deref here, Release elsewhere, etc

internal/testutil/baseline/baseline.go
jakebailey: It was already linkified for me...

internal/compiler/program.go
andrewbranch: Hm, I see the race issue, but the reason I did this is behavior, not perf. `p.unresolvedImports` gets copied as part of `p.UpdateProgram`, and the cloned program should never run its `p.unresolvedImportsOnce` because the result is guaranteed to be the same.

internal/core/workgroup.go
andrewbranch: This would just be `errorgroup` if it created the limiter here. This is to limit `npm install` concurrency process-wide even if calling the typings installer concurrently.

internal/compiler/program.go
jakebailey: If you move the if into the Once, that will achieve the same thing, then.

internal/project/dirty/interfaces.go
andrewbranch: It’s just used for the assertion 😕 I couldn’t find another way to do this; any suggestions? I can just delete the assertions.

internal/core/workgroup.go
jakebailey: Ah, then could it just throttle `npm install` at the top instead? (But, this could be fine, just speculating.)

internal/project/dirty/interfaces.go
jakebailey: I thought there might be a way with just interfaces, but I think you do need a concrete type to do the self-referencing type param constraint trick. So, maybe just rename it to `testCloneable` or something to make it clear this value should never be used... I think when I read it I immediately thought it was being embedded somewhere else and was trying to find the bug without realizing it was never referenced.

jakebailey: clonableDoNotUseOrYouWillBeFired

internal/project/parsecache.go
andrewbranch: This replaces DocumentStore, yeah. And I ended up just making an option on it to disable deletion so it can be passed in for tests without implementing an interface.

internal/testutil/baseline/baseline.go
andrewbranch: Not on Windows 🥲

internal/compiler/program.go
jakebailey: I see the `cloned` thing, but I guess I would have assumed that it would be sufficient to do a nil check inside and just return it as-is? Is it possible for the method to not have yet been called on the program before it's cloned, and then it's never set up?

internal/core/workgroup.go
andrewbranch: Not sure exactly what you mean - I want the cancelation and error propagation behavior of errgroup, but with a global concurrency limit instead of a concurrency limit per group. You can't just share a top-level errgroup.Group between goroutines, because that would mix up the error propagation, and you're not allowed to `Add` to a group while `Wait` is running. I guess it doesn't provide that much utility to extract this to a type in core, as it's a pretty thin wrapper around errgroup, and I only use it here, so it could be inlined, if that's what you mean. But I think I haven't understood your suggestion at all.

internal/project/dirty/interfaces.go
andrewbranch: I'll just delete it; it was useful while authoring but now that all the usage is in place, it doesn't provide much value.

internal/core/workgroup.go
jakebailey: I was mainly suggesting that the throttling happen inside the implementor of the `NpmExecutor` interface, such that the callers downstream don't need to know that they are throttled (and `NpmInstall` accepts a context so it can be cancelled early, potentially).

andrewbranch: I don't know, it's kind of nice that `NpmExecutor` can be implemented without worrying about spawning a goroutine and checking cancelation.

jakebailey: I don't mind how it is now, the abstraction just surprised me. Either way, someone has to do that, so it felt better to do it on its own without involving channels in the API.

internal/compiler/program.go
```diff
@@ -255,13 +264,55 @@ func equalCheckJSDirectives(d1 *ast.CheckJsDirective, d2 *ast.CheckJsDirective)
...
+// The result is cached and computed only once.
+func (p *Program) ExtractUnresolvedImports() collections.Set[string] {
+	if p.cloned {
+		return *p.unresolvedImports
+	}
+	p.unresolvedImportsOnce.Do(func() {
+		p.unresolvedImports = p.extractUnresolvedImports()
+	})
+
+	return *p.unresolvedImports
```
jakebailey: I still feel like this would make more sense as:

```suggestion
	p.unresolvedImportsOnce.Do(func() {
		if p.unresolvedImports == nil {
			p.unresolvedImports = p.extractUnresolvedImports()
		}
	})

	return *p.unresolvedImports
```

And no `cloned` field needed.

jakebailey: (That and I'm not sure why this returns `collections.Set` instead of a pointer to it)

jakebailey (commented):
Ran the LSP and typed in the checker with race mode on, and got this race:

```
==================
WARNING: DATA RACE
Write at 0x00c02e47df00 by goroutine 49382:
  github.com/microsoft/typescript-go/internal/project.(*compilerHost).freeze()
      /home/jabaile/work/TypeScript-go/internal/project/compilerhost.go:50 +0xd6e
  github.com/microsoft/typescript-go/internal/project.(*Snapshot).Clone()
      /home/jabaile/work/TypeScript-go/internal/project/snapshot.go:189 +0xd03
  github.com/microsoft/typescript-go/internal/project.(*Session).UpdateSnapshot()
      /home/jabaile/work/TypeScript-go/internal/project/session.go:425 +0x232
  github.com/microsoft/typescript-go/internal/project.(*Session).GetLanguageService()
      /home/jabaile/work/TypeScript-go/internal/project/session.go:387 +0x417
  github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[go.shape.*uint8,go.shape.struct { FullDocumentDiagnosticReport *github.com/microsoft/typescript-go/internal/lsp/lsproto.RelatedFullDocumentDiagnosticReport; UnchangedDocumentDiagnosticReport *github.com/microsoft/typescript-go/internal/lsp/lsproto.RelatedUnchangedDocumentDiagnosticReport }].10()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:506 +0x15b
  github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:421 +0x1f8
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:326 +0x64

Previous read at 0x00c02e47df00 by goroutine 49385:
  github.com/microsoft/typescript-go/internal/project.(*compilerFS).UseCaseSensitiveFileNames()
      /home/jabaile/work/TypeScript-go/internal/project/compilerhost.go:155 +0x28
  github.com/microsoft/typescript-go/internal/compiler.(*Program).UseCaseSensitiveFileNames()
      /home/jabaile/work/TypeScript-go/internal/compiler/program.go:143 +0x43
  github.com/microsoft/typescript-go/internal/compiler.(*Program).toPath()
      /home/jabaile/work/TypeScript-go/internal/compiler/program.go:1491 +0x5c
  github.com/microsoft/typescript-go/internal/compiler.(*Program).GetSourceFile()
      /home/jabaile/work/TypeScript-go/internal/compiler/program.go:1495 +0x5e
  github.com/microsoft/typescript-go/internal/ls.(*LanguageService).tryGetProgramAndFile()
      /home/jabaile/work/TypeScript-go/internal/ls/languageservice.go:27 +0x4f
  github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getProgramAndFile()
      /home/jabaile/work/TypeScript-go/internal/ls/languageservice.go:33 +0x53
  github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideDocumentSymbols()
      /home/jabaile/work/TypeScript-go/internal/ls/symbols.go:21 +0x48
  github.com/microsoft/typescript-go/internal/lsp.(*Server).handleDocumentSymbol()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:788 +0x65
  github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[go.shape.*uint8,go.shape.struct { SymbolInformations *[]*github.com/microsoft/typescript-go/internal/lsp/lsproto.SymbolInformation; DocumentSymbols *[]*github.com/microsoft/typescript-go/internal/lsp/lsproto.DocumentSymbol }].21()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:511 +0x22e
  github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:421 +0x1f8
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:326 +0x64

Goroutine 49382 (running) created at:
  github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:346 +0x976
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:222 +0x46
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/jabaile/go/pkg/mod/golang.org/x/sync@v0.16.0/errgroup/errgroup.go:93 +0x86

Goroutine 49385 (running) created at:
  github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop()
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:346 +0x976
      /home/jabaile/work/TypeScript-go/internal/lsp/server.go:222 +0x46
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/jabaile/go/pkg/mod/golang.org/x/sync@v0.16.0/errgroup/errgroup.go:93 +0x86
==================
```

internal/compiler/program.go
andrewbranch: Oh yeah, that would work

internal/tsoptions/tsconfigparsing.go
```diff
@@ -909,56 +913,48 @@ func readJsonConfigFile(fileName string, path tspath.Path, readFile func(fileNam
...
-			}
+		return &ExtendedConfigCacheEntry{
+			extendedResult: extendedResult,
+			extendedConfig: extendedConfig,
 		}
 	}
+
+	var cacheEntry *ExtendedConfigCacheEntry
+	if extendedConfigCache != nil {
+		cacheEntry = extendedConfigCache.GetExtendedConfig(extendedConfigFileName, extendedConfigPath, parse)
```
sheetalkamat, line 958: Dont you have to get errors here and add just like what we do above as otherwise say "missing file" or some such errors will not be reported

internal/compiler/host.go
```diff
@@ -80,11 +76,6 @@ func (h *compilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast.Sourc
 }
 
 func (h *compilerHost) GetResolvedProjectReference(fileName string, path tspath.Path) *tsoptions.ParsedCommandLine {
-	h.extendedConfigCacheOnce.Do(func() {
```
sheetalkamat, line 90: I think you do want to create cache here esp because these project references can be using same extended file and you want to use cache

sheetalkamat (commented):
I am still reviewing but dont want my current comment to go away so submitting

internal/fourslash/fourslash.go
```diff
@@ -169,7 +147,11 @@ func NewFourslash(t *testing.T, capabilities *lsproto.ClientCapabilities, conten
 		FS:                 fs,
 		DefaultLibraryPath: bundled.LibPath(),
 
-		ParsedFileCache: &parsedFileCache{},
+		ParseCache: &project.ParseCache{
```
gabritto: I may be missing something, but this doesn't seem like it's doing what the previous setup was doing: before, the cache would be shared among *all fourslash tests*, and now it seems like we create a new cache for each test. I see that we're not deleting files from the cache, but, if we're creating a new parse cache for every test, then we're not sharing the `entries` which is where we access the cached files, right?

jakebailey: That does seem to be the case, and the new code spends about 68% of its time parsing/binding, versus 7% before.

That being said, on my machine that means `gen` packages' tests going from 3.8s to 7.7s, which isn't awful. But it does seem like we gave up some caching here.

andrewbranch: Good catch! This was a mistake, not intentional.

internal/execute/extendedconfigcache.go
```diff
@@ -0,0 +1,37 @@
...
+		e.mu.Unlock()
+		return entry
+	}
+	e.mu.Unlock()
+	entry := parse()
+	e.mu.Lock()
+	if e.m == nil {
+		e.m = make(map[tspath.Path]*tsoptions.ExtendedConfigCacheEntry)
+	}
+	e.m[path] = entry
```
sheetalkamat: This doesnt seem like it is concurrency safe as alreadt parsed entry that could be already set and in use is lost. 

Two things:
1. shouldnt lock whole cache - it should use sync map isntead as the entries are independent of each other
2.use LoadOrStore here so the already loaded one always wins if two or more threads did create it

internal/testutil/projecttestutil/projecttestutil.go
```diff
@@ -3,142 +3,114 @@ package projecttestutil
...
 }
 
-// Log implements project.ProjectServiceHost.
-func (p *ProjectServiceHost) Log(msg ...any) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-	fmt.Fprintln(&p.output, msg...)
-}
+func (h *SessionUtils) SetupNpmExecutorForTypingsInstaller() {
+	if h.testOptions == nil {
```
sheetalkamat: Lets no skip typing installer set up by default. we have few typings installer cases but the fact that it wasnt running by default on all the project tests had caught some issues with project updates, request etc in past. So we would want the typing installer to be setup by default.

andrewbranch: I'll have to think about how to handle this; the problem is that ATA executes asynchronously after a snapshot update, so if the test doesn't wait for background tasks, it could be nondeterministic whether it executes.

internal/execute/extendedconfigcache.go
jakebailey: I am not sure if it's helpful, but there is also https://github.com/microsoft/typescript-go/pull/1595

internal/compiler/host.go
andrewbranch: I guess it's a matter of taste, but since this is not currently public API, it seems simpler to rely on the creator of the host to provide a cache if they want one. We don't have to safeguard against every possible usage pattern like we did in Strada.

sheetalkamat: yeah ignore .  this was before i saw execute passing it along correctly

internal/testutil/baseline/baseline.go
jakebailey: Can we maybe conditionalize this on the OS?

(I'd be interested in an example, I've successfully used the test suite on windows so I'm curious what specifically breaks)

jakebailey (approved):
Is there anything left? I've been playing with this and can't break it, even with the race detector enabled.

andrewbranch:
I realized while gone that I forgot the file watchers for ATA locations. Working on that now.

internal/execute/tsc.go
```diff
@@ -169,12 +168,10 @@ func tscCompilation(sys System, commandLine *tsoptions.ParsedCommandLine, testin
...
 	compilerOptionsFromCommandLine := commandLine.CompilerOptions()
 	configForCompilation := commandLine
-	var extendedConfigCache collections.SyncMap[tspath.Path, *tsoptions.ExtendedConfigCacheEntry]
+	extendedConfigCache := &extendedConfigCache{}
 	var configTime time.Duration
 	if configFileName != "" {
-		configStart := sys.Now()
-		configParseResult, errors := tsoptions.GetParsedCommandLineOfConfigFile(configFileName, compilerOptionsFromCommandLine, sys, &extendedConfigCache)
-		configTime = sys.Now().Sub(configStart)
+		configParseResult, errors := tsoptions.GetParsedCommandLineOfConfigFile(configFileName, compilerOptionsFromCommandLine, sys, extendedConfigCache)
```
sheetalkamat, line 175: configTime reporting is missing

internal/project/ata/ata.go
```diff
@@ -0,0 +1,487 @@
...
+				// key is @types/<package name>
+				packageName := tspath.GetBaseFileName(key)
+				if packageName == "" {
+					continue
+				}
+				typingFile := ti.typingToFileName(resolver, packageName)
+				if typingFile == "" {
+					continue
+				}
+				newVersion := semver.MustParse(npmLockValue.Version)
```
sheetalkamat: There use to be code for version mismatch detection - dont we need it ?

internal/project/ata/ata_test.go
```diff
@@ -0,0 +1,310 @@
...
+		configFile := program.GetSourceFile("/user/username/projects/project/config.js")
+		assert.Assert(t, configFile != nil, "local config.js should be included")
+
+		// Verify that only types-registry was installed (no @types/config since it's a local module)
+		npmCalls := utils.NpmExecutor().NpmInstallCalls()
+		assert.Equal(t, len(npmCalls), 1)
+		assert.Equal(t, npmCalls[0].Args[2], "types-registry@latest")
+	})
+
+	t.Run("configured projects", func(t *testing.T) {
```
sheetalkamat, line 56: Can you please keep the skipped tests or keep them commented or something so they are tracking pending work 

Eg deduplicate from local @types packages

internal/project/ata/ata_test.go
```diff
@@ -0,0 +1,310 @@
...
+		t.Parallel()
+
+		files := map[string]any{
```
sheetalkamat, line 59: I seem some other tests that were not skipped but not copied. did you miss them or something else

eg: discover from node_modules empty types
discover from node_modules explicit types
discover from node_modules empty types has import
Malformed package.json should be watched
should redo resolution that resolved to '.js' file after typings are installed
expired cache entry (inferred project, should install typings)
non-expired cache entry (inferred project, should not install typings)
expired cache entry (inferred project, should install typings) lockfile3
non-expired cache entry (inferred project, should not install typings) lockfile3

internal/project/projectcollectionbuilder.go
```diff
@@ -0,0 +1,854 @@
...
+				}
+			},
+		)
+	}
+
+	// Recompute default projects for open files that now have different config file presence.
+	var hasChanges bool
+	for path := range configChangeResult.affectedFiles {
+		fileName := b.fs.overlays[path].FileName()
+		_ = b.ensureConfiguredProjectAndAncestorsForOpenFile(fileName, path, logger)
```
sheetalkamat, line 398: Computing this here seems like work that should be done when the next request comes in ? "you could get multiple watch events and its wasteful and expensive to do this right ?

internal/project/project.go
```diff
@@ -81,981 +49,315 @@ const (
...
-		// 	}
-		// }
-	}
-
-	if p.rootFileNames.Size() > len(rootFileNames) {
-		for root := range p.rootFileNames.Keys() {
-			if _, ok := newRootScriptInfos[root]; !ok {
-				p.rootFileNames.Delete(root)
-			}
+		if p.Program != nil && p.Program.HasSameFileNames(newProgram) {
```
sheetalkamat: nit change to program.HasSameFileNames to HasSameFiles ? as its checking that the files themselves are same too?

internal/compiler/program.go
```diff
@@ -206,20 +210,23 @@ func NewProgram(opts ProgramOptions) *Program {
...
 	// TODO: reverify compiler options when config has changed?
 	result := &Program{
-		opts:                        p.opts,
+		opts:                        newOpts,
 		comparePathsOptions:         p.comparePathsOptions,
 		processedFiles:              p.processedFiles,
 		usesUriStyleNodeCoreModules: p.usesUriStyleNodeCoreModules,
 		programDiagnostics:          p.programDiagnostics,
 		hasEmitBlockingDiagnostics:  p.hasEmitBlockingDiagnostics,
+		unresolvedImports:           p.unresolvedImports,
```
sheetalkamat, line 229: can copy unresolvedImports when program has same files as well into new program ?

sheetalkamat (changes requested):

internal/project/ata/ata_test.go
andrewbranch: Ugh, I asked Copilot to port all the tests 😑

internal/project/projectcollectionbuilder.go
andrewbranch: No, `affectedFiles` is only set when a file with basename `tsconfig.json` or `jsconfig.json` is created or deleted. Watch events are already debounced and consolidated to a degree, and even in a `git switch` or `npm install` scenario where a flood of them comes in over an extended time period, it's still unlikely that many of those events contain tsconfig creations/deletions. This fixes a bug in the old system I noticed where deleting the tsconfig.json file for an open file has no effect until a _different_ file is opened.

internal/project/project.go
andrewbranch: I think this isn't what I intended - I actually intended to check just that the file names are the same (I was thinking the map being compared was `map[tspath.Path]string`, not `map[tspath.Path]*ast.SourceFile`), i.e., Program was not cloned but the change didn't result in a different structure.

internal/project/ata/ata.go
andrewbranch: Another silent Copilot porting omission 😠

internal/execute/tsc.go
andrewbranch: Thanks, merge resolution error

internal/project/ata/ata.go
andrewbranch: A little spooky that this is untested, even after porting the missed tests 
image

internal/project/project.go
```diff
@@ -81,981 +49,316 @@ const (
...
+	installedTypingsInfo *ata.TypingsInfo
+	// typingsFiles are the root files added by the typings installer.
+	typingsFiles []string
 }
 
 func NewConfiguredProject(
 	configFileName string,
 	configFilePath tspath.Path,
-	host ProjectHost,
+	builder *projectCollectionBuilder,
```
camc314, line 92: `NewConfiguredProject` is exported, but it takes a `*projectCollectionBuilder`, which isn’t exported. That means code outside the package can’t actually call this function. Was the intent for the constructor to be internal (unexported), or for projectCollectionBuilder to be exported?

internal/project/snapshot.go
```diff
@@ -0,0 +1,250 @@
...
+	compilerOptionsForInferredProjects *core.CompilerOptions
+
+	builderLogs *logging.LogTree
+	apiError    error
+}
+
+// NewSnapshot
+func NewSnapshot(
+	id uint64,
+	fs *snapshotFS,
```
camc314, line 42: same comment here, `NewSnapshot` is public, but `snapshotFS` is not?

Same comment with `extendedConfigCache` on L45

internal/project/project.go
jakebailey: These functions aren't referenced outside the package, I think, so it's somewhat innocuous given the internalness of the whole API (we have worse examples elsewhere... but usually not in this part of the code). Are you expecting to need to call any of these in tsgolint and that's the interest in having these available?

jakebailey (approved):
🙏

internal/project/project.go
camc314: Two reasons really:
 - from tsgolint, we aren't expecting to be able to call these, but it breaks how we generate our shims (we can always work around this so feel free to tell me to do that 🙂 ), as the inconsistency means that we're trying to reference private types.
 - from a code pov, it doesn't really make sense for one to be public, and not, as it means `NewConfiguredProject` can't be called outside the package, despite it being exported?

That said, feel free to ignore these cmts and i can workaround in the tsgolint side.

andrewbranch:
@camc314 I'm merging now but happy to follow up with changes. To answer your questions broadly, the change most aligned with my intentions would be to make `NewSnapshot` and `NewProject` unexported; the only way you're supposed to get a reference to them is through `Session`. I would eventually like a way for API consumers to speculate on a snapshot change; i.e. call `snapshot.Clone` directly, or have a method through Session to help with that, but that wasn't in my first string of priorities. There is very little in here that is tailored to external API usage, but over time I anticipate needing to add more functionality to integrate with the IPC API to support LS plugins, maybe MCP usage, etc., and I think those kinds of changes will be similar to what tsgolint would like to see.

PR #1512, opened 2025-08-04
Regenerate NOTICE.txt file

jakebailey:
One day this will be automated...

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,10 +1,10 @@
 NOTICES AND INFORMATION
 Do Not Translate or Localize
 
-This software incorporates material from third parties. Microsoft makes certain
-open source code available at https://3rdpartysource.microsoft.com, or you may
-send a check or money order for US $5.00, including the product name, the open
-source component name, and version number, to:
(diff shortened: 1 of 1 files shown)


NOTICE.txt
```diff
@@ -200,132 +199,117 @@ TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
...
+
+(c) (c) A
+(c) AeHa (c)
+(c) EeAoq (c)
+(c) EAiE (c) A
+(c) lEEe (c) AE
+(c) (c) AAEE (c)
+(c) oav!A (c) AY
+(c) aA"AE (c) 1AE (c)
+(c) 1/4, ?i'U CY A$?A CO
```
jakebailey, line 278: Whatever system they use for this clearly has a bug

jakebailey:
Copilot is right but I'm not going to fix any of it, that's just how it's generated, and people always forget to actually fill in their licenses' name/year

sheetalkamat (approved):

PR #1514, opened 2025-08-04
Fix module cache

jakebailey:
I typo'd this, so no workflow is restoring the modules ☹️


jakebailey:
This is unfortunately a bit of a game of trial and error, but from what I am seeing even though it's brokenish without this PR, it's still a few minutes faster than before the new method.

PR #1516, opened 2025-08-05
Set UseAliasDefinedOutsideCurrentScope in hover

jakebailey:
Before:

image

After:

image

Confusingly, no existing fourslash test seems to test this? I was seeing it all the time.

This one flag is not the extent of the flags hover used to set; it also used to sometimes set `MultilineObjectLiterals` for type display parts, but doing that regressed the existing tests. `MultilineObjectLiterals`, `WriteTypeArgumentsOfSignature`, `OmitParameterModifiers` used to be set for signature display parts, but it wasn't immediately obvious to me where to set that.

Mainly, I just want to stop seeing the random import nodes which Strada does not show.


DanielRosenwasser:
I'm guessing we have tests that are not ported (non-baseline ones). In the meantime, can you add a non-generated test which previously exhibits the behavior? I feel like that should be pretty easy to repro.

jakebailey:
Rebased with a test that shows the difference.

DanielRosenwasser (approved):

PR #1525, opened 2025-08-06
Fix panic runtime error on ast.KindTupleTypes

### LINKED ISSUE #1408: panic: runtime error: slice bounds out of range [:52400] with length 524 when processing Knex module augmentations
tsgo panics when processing Knex module augmentations with complex generic constraints.

Reproduction

Minimal reproduction: https://github.com/vpontis/tsgo-panic

```bash
git clone https://github.com/vpontis/tsgo-panic
cd tsgo-panic
bun install
bun tsgo  # panics
bun tsc   # works fine
```

Stack Trace

```
panic: runtime error: slice bounds out of range [:52400] with length 524

goroutine 1 [running]:
github.com/microsoft/typescript-go/scanner.(*Scanner).GetLineAndCharacterOfPosition(...)
        /Users/victor/go/pkg/mod/github.com/microsoft/typescript-go@v0.0.0-20250116223143-00ad56e4d9d7/scanner/scanner.go:3156
github.com/microsoft/typescript-go/scanner.GetLineAndCharacterOfPosition(...)
        /Users/victor/go/pkg/mod/github.com/microsoft/typescript-go@v0.0.0-20250116223143-00ad56e4d9d7/scanner/utilities.go:1222
github.com/microsoft/typescript-go/transformer.(*transformer).getJSONTypeArgument(0xc002c12240, 0xc0053084c0)
        /Users/victor/go/pkg/mod/github.com/microsoft/typescript-go@v0.0.0-20250116223143-00ad56e4d9d7/transformer/declarations.go:2172 +0x125
```

File

```ts
import "knex";

// Minimal UnionPick type
type UnionPick<T, K extends keyof T> = T extends unknown ? Pick<T, K> : never;

// Minimal reproduction of the module augmentation that causes tsgo panic
declare module "knex" {
  namespace Knex {
    interface QueryInterface<TRecord extends {} = any> {
      firstOrFail<TKey extends keyof TRecord>(
        ...columns: TKey[]
      ): Knex.QueryBuilder<TRecord, UnionPick<TRecord, TKey>> &
        Promise<UnionPick<TRecord, TKey>>;

      firstOrFail(
        column: "*",
      ): Knex.QueryBuilder<TRecord, TRecord> & Promise<TRecord>;
    }
  }
}
```

### LINKED ISSUE #1501: panic: runtime error: slice bounds out of range [:59935] with length 568
Hello, I got a new error when running TSGO with latest version.

`"@typescript/native-preview": "^7.0.0-dev.20250801.1",`

## Stack trace

```
panic: runtime error: slice bounds out of range [:59935] with length 568

goroutine 147808 [running]:
github.com/microsoft/typescript-go/internal/scanner.GetLineAndCharacterOfPosition({0x1017a9130, 0x14017d86b08}, 0xea1f)
	github.com/microsoft/typescript-go/internal/scanner/scanner.go:2434 +0x158
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visitDeclarationSubtree(0x14083e9ea80, 0x14023b38f00)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:471 +0x808
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visit(0x1402a9d1408?, 0x10101dd6c?)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:145 +0x114
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitSlice(0x14060f25620, {0x14023b54318, 0x2, 0x2})
	github.com/microsoft/typescript-go/internal/ast/visitor.go:148 +0xa0
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitNodes(0x14060f25620, 0x14022a19a28)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:99 +0x40
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).visitNodes(0x1402a9d1508?, 0x1012bb3b8?)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:236 +0x44
github.com/microsoft/typescript-go/internal/ast.(*UnionTypeNode).VisitEachChild(0x14023b4c7c8, 0x14060f25620)
	github.com/microsoft/typescript-go/internal/ast/ast.go:7343 +0x30
github.com/microsoft/typescript-go/internal/ast.(*Node).VisitEachChild(...)
	github.com/microsoft/typescript-go/internal/ast/ast.go:245
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitEachChild(...)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:191
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visitDeclarationSubtree(0x14083e9ea80, 0x14023b4c7c8)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:494 +0xce8
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visit(0x10127d660?, 0x14083e84f08?)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:145 +0x114
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformTypeAliasDeclaration(0x14083e9ea80, 0x14023b2e9e0)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:1124 +0x78
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformTopLevelDeclaration(0x14083e9ea80, 0x14023b2e9e0)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:1092 +0x468
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visitDeclarationStatements(0x14083e9ea80, 0x14023b2e9e0)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:942 +0x4c
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visit(0x140123578c8?, 0x1012c1d04?)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:120 +0xec
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitSlice(0x14060f25620, {0x140245a3208, 0x106, 0x106})
	github.com/microsoft/typescript-go/internal/ast/visitor.go:148 +0xa0
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitNodes(0x14060f25620, 0x14024070de8)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:99 +0x40
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformModuleDeclaration(0x14083e9ea80, 0x14022a60bd0)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:1182 +0xc4
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformTopLevelDeclaration(0x14083e9ea80, 0x14022a60bd0)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:1098 +0x438
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformAndReplaceLatePaintedStatements(0x14083e9ea80, 0x14017d82a80)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:236 +0xf4
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).transformSourceFile(0x14083e9ea80, 0x14017d86b08)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:188 +0x40
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visitSourceFile(0x14083e9ea80, 0x14017d86b08)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:174 +0x174
github.com/microsoft/typescript-go/internal/transformers/declarations.(*DeclarationTransformer).visit(0x14035fd0dd0?, 0xd0?)
	github.com/microsoft/typescript-go/internal/transformers/declarations/transform.go:104 +0xc0
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitNode(0x14022717290?, 0x14083cea320?)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:51 +0x30
github.com/microsoft/typescript-go/internal/ast.(*NodeVisitor).VisitSourceFile(...)
	github.com/microsoft/typescript-go/internal/ast/visitor.go:36
github.com/microsoft/typescript-go/internal/transformers.(*Transformer).TransformSourceFile(...)
	github.com/microsoft/typescript-go/internal/transformers/transformer.go:40
github.com/microsoft/typescript-go/internal/compiler.(*emitter).emitDeclarationFile(0x14022717290, 0x14017d86b08, {0x14035fd0dd0, 0xd0}, {0x0, 0x0})
	github.com/microsoft/typescript-go/internal/compiler/emitter.go:179 +0x1b8
github.com/microsoft/typescript-go/internal/compiler.(*emitter).emit(0x14022717290)
	github.com/microsoft/typescript-go/internal/compiler/emitter.go:50 +0xb0
github.com/microsoft/typescript-go/internal/compiler.(*Program).Emit.func2()
	github.com/microsoft/typescript-go/internal/compiler/program.go:1278 +0x1ac
github.com/microsoft/typescript-go/internal/core.(*parallelWorkGroup).Queue.func1()
	github.com/microsoft/typescript-go/internal/core/workgroup.go:39 +0x5c
created by github.com/microsoft/typescript-go/internal/core.(*parallelWorkGroup).Queue in goroutine 147611
	github.com/microsoft/typescript-go/internal/core/workgroup.go:37 +0x84
```

## Steps to reproduce

1. tsgo --noEmit --skipLibCheck --project tsconfig.tsgo.json

hamidrezahanafi:
Fixes #1501
Fixes #1408

Root cause from my debugging?
The declaration transformer is processing AST nodes that originated from `@types/react/index.d.ts` (the large file), but it's trying to use those node positions with a completely different source file (types.ts) as the context.
This is what I suspected - cross-file node processing during incremental compilation. 
The incremental build system mixes nodes from different files, but the transformer was always using `tx.state.currentSourceFile` for position calculations instead of the node's actual source file.

What's the fix?
When handling TupleType nodes, we now compute line/character information against the node’s original parse location by using original := EmitContext().MostOriginal(input) and then the originalSourceFile derived from that node. If the original source file is available and the tuple spans a single line there, we set EFSingleLine. This ensures formatting and position calculations are based on the correct file, regardless of whether the node originated in the current file or a dependency, eliminating mis-formatting and position errors during declaration emit.

### DIFF
--- a/internal/transformers/declarations/transform.go
+++ b/internal/transformers/declarations/transform.go
@@ -10,7 +10,6 @@ import (
 	"github.com/microsoft/typescript-go/internal/modulespecifiers"
 	"github.com/microsoft/typescript-go/internal/nodebuilder"
 	"github.com/microsoft/typescript-go/internal/printer"
-	"github.com/microsoft/typescript-go/internal/scanner"
 	"github.com/microsoft/typescript-go/internal/transformers"
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
@@ -468,14 +467,8 @@ func (tx *DeclarationTransformer) visitDeclarationSubtree(input *ast.Node) *ast.
 	case ast.KindTupleType:
 		result = tx.Visitor().VisitEachChild(input)
 		if result != nil {
-			original := tx.EmitContext().MostOriginal(input)
-			originalSourceFile := ast.GetSourceFileOfNode(original)
-			if originalSourceFile != nil {
-				startLine, _ := scanner.GetLineAndCharacterOfPosition(originalSourceFile, original.Loc.Pos())
-				endLine, _ := scanner.GetLineAndCharacterOfPosition(originalSourceFile, original.Loc.End())
-				if startLine == endLine {
-					tx.EmitContext().AddEmitFlags(result, printer.EFSingleLine)
-				}
+			if transformers.IsOriginalNodeSingleLine(tx.EmitContext(), input) {
+				tx.EmitContext().AddEmitFlags(result, printer.EFSingleLine)
 			}
 		}
 	case ast.KindJSDocTypeExpression:
--- a/internal/transformers/utilities.go
+++ b/internal/transformers/utilities.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/printer"
+	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
 func IsGeneratedIdentifier(emitContext *printer.EmitContext, name *ast.IdentifierNode) bool {
@@ -244,4 +245,21 @@ func IsSimpleCopiableExpression(expression *ast.Expression) bool {
 		ast.IsNumericLiteral(expression) ||
 		ast.IsKeywordKind(expression.Kind) ||
 		ast.IsIdentifier(expression)
+}
+
+func IsOriginalNodeSingleLine(emitContext *printer.EmitContext, node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	original := emitContext.MostOriginal(node)
+	if original == nil {
+		return false
+	}
+	source := ast.GetSourceFileOfNode(original)
+	if source == nil {
+		return false
+	}
+	startLine, _ := scanner.GetLineAndCharacterOfPosition(source, original.Loc.Pos())
+	endLine, _ := scanner.GetLineAndCharacterOfPosition(source, original.Loc.End())
+	return startLine == endLine
 }


jakebailey:
Do you have a test which reproduces this for the PR? It seems like you know the conditions, so laying out a compiler test for it would be helpful.

hamidrezahanafi:
> Do you have a test which reproduces this for the PR? It seems like you know the conditions, so laying out a compiler test for it would be helpful.

I tried super hard to find a test that failing and many combinations, still no luck of a compiler test that would fail. 
Based on my observations, the failures are on the basically where we again didn't import React directly and use `React.` in the code. 
Also any chance that we can merge this fix without test to unblock?

jakebailey:
We typically don't do that, but we could.

What I'm not understanding is why this particular fix is needed; the original code says:

```ts
if (isTupleTypeNode(input) && (getLineAndCharacterOfPosition(currentSourceFile, input.pos).line === getLineAndCharacterOfPosition(currentSourceFile, input.end).line)) {
	setEmitFlags(input, EmitFlags.SingleLine);
}
```

Which is equivalent to the original code, and that sort of tells me the problem is elsewhere, e.g. maybe we have forgotten to set the text on the intermediate SourceFile?

hamidrezahanafi:
> We typically don't do that, but we could.
> 
> What I'm not understanding is why this particular fix is needed; the original code says:
> 
> ```ts
> if (isTupleTypeNode(input) && (getLineAndCharacterOfPosition(currentSourceFile, input.pos).line === getLineAndCharacterOfPosition(currentSourceFile, input.end).line)) {
> 	setEmitFlags(input, EmitFlags.SingleLine);
> }
> ```
> 
> Which is equivalent to the original code, and that sort of tells me the problem is elsewhere, e.g. maybe we have forgotten to set the text on the intermediate SourceFile?

Even the `tx.state.currentSourceFile` is different than `ast.GetSourceFileOfNode(input)` and this only happens when I don't import React from "react" and use `React.` on `slate/src/private/shared/types.ts `
If you could point me in code where we set this state, `currentSourceFile`, maybe I would be able to fix it there
```
CROSS-FILE TUPLE: pos=59935, nodeFile=.../node_modules/@types/react/index.d.ts, currentFile=.../slate/src/private/shared/types.ts, currentLen=1905
SLICE BOUNDS DANGER: pos=59935 > currentLen=1905
  - Node file: .../node_modules/@types/react/index.d.ts
  - Current file: .../slate/src/private/shared/types.ts  
  - Position exceeds current file length by 58030 characters
  - This would cause: runtime error: slice bounds out of range [:59935] with length 1905
```

jakebailey:
I understand that they are different, but they would be different in the old TS codebase too. It's odd that a 1:1 port is not working here.

Calling `GetSourceFileOfNode` on the input node is very suspect because the input may be transformed and not have a parent at all.

hamidrezahanafi:
> I understand that they are different, but they would be different in the old TS codebase too. It's odd that a 1:1 port is not working here.
> 
> Calling `GetSourceFileOfNode` on the input node is very suspect because the input may be transformed and not have a parent at all.

I tried asking AI to look for this case, failing and passing and it has trouble finding where the react types are being inlined when there is no explicit React import, if you have any ideas on where it happen, I can take further look
I wish I could create a repro

relsunkaev:
I have a test in [this pr](https://github.com/microsoft/typescript-go/pull/1539) that reproduces the bug

jakebailey:
Turned into a compiler test, that might work, but we usually do not add this sort of unit test since it's not clear that the unit tests is actually setting up things how the actual compiler does.

hamidrezahanafi:
@jakebailey GPT-5 found this a fix that actually works, please let me know if it makes sense, also I am not able to write a compiler test that fails

Queue late-painted statements only from the active file:
In internal/transformers/declarations/tracker.go, we now append to `lateMarkedStatements` only if `ast.GetSourceFileOfNode(ref) == s.state.currentSourceFile`. This mirrors TS’s per-file queue during declaration emit.

TS sets currentSourceFile at file entry for both single-file and bundle flows and assumes the late-painted queue is for the same file; it then uses currentSourceFile for position math during declaration emit:
File entry and per-file traversal: [declarations.ts L510–538](https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/declarations.ts#L510-L538), bundle path L466–499.
Tuple single-line check uses currentSourceFile: [declarations.ts L1260](https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/declarations.ts#L1260).

hamidrezahanafi:
> I can understand how this would fix the problem - this kind of check is pervasive in the node builder - but since strada doesn't have this check here (and doesn't seem to need to?) we'd _really_ want a test case demonstrating the crash, so we can see where, exactly, the divergence in behavior is. I'd much prefer to fix the difference in behavior at the root than at some symptom.

Tried many test cases on compiler tests and couldn't do the panic
On our codebase, whenever we remove `import React from "react"`, TS 5.9 works fine but TSGo panics, but it's not happening on new repo so size also matters maybe

weswigham:
Sure, but what I'm getting at is that the issue isn't that these paths don't check for containing file consistency, but rather that the built/marked nodes shouldn't be getting their `.original` set/value returned when the original containing file is inconsistent with the destination location for the node in the first place, and that's where the bug truly lies - and tracking where that is down without a test case is going to be challenging, as will preventing a regression, so a test case is probably the first order of business.

TL;DR: This is the wrong fix - it's masking a symptom, not fixing the cause.

hamidrezahanafi:
> Sure, but what I'm getting at is that the issue isn't that these paths don't check for containing file consistency, but rather that the built/marked nodes shouldn't be getting their `.original` set/value returned when the original containing file is inconsistent with the destination location for the node in the first place, and that's where the bug truly lies - and tracking where that is down without a test case is going to be challenging, as will preventing a regression, so a test case is probably the first order of business.
> 
> TL;DR: This is the wrong fix - it's masking a symptom, not fixing the cause.

Finally could create a reproduction 
https://github.com/hamidrezahanafi/nx-empty-baseurl/tree/hrh.triggertspanic

do `npm install` and `npm run tsc-test-app`

jakebailey:
I followed those steps and it did not panic.

hamidrezahanafi:
> hrh.triggertspanic

Are you on this branch on that repo? `hrh.triggertspanic`
I can reliably reproduce it with `npm run tsc-test-app`

jakebailey:
Yep:

image

hamidrezahanafi:
> Yep:
> 
> image

It needed the `tsconfig.tsbuildinfo` that I had to trigger the panic, pushed it to the branch, it should panic now 🤞

jakebailey:
Yes, I now see the crash. #1408 also reproduces this with probably less stuff in the repo, so it might be a better candidate for mimimization. Though neither of these are minimal.

jakebailey:
Surprisingly, this issue does seem to require a tsbuildinfo be present to work; same happens in #1408's.

jakebailey:
Hm, no, it doesn't.

jakebailey:
Here is a minimal repro:

```ts
// @strict: true
// @module: preserve
// @declaration: true

// @filename: node_modules/knex/index.d.ts

// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward
// A bunch of random text to move the positions forward

type ShouldJustBeAny = [any][0];

declare namespace knex {
  export { Knex };
}

declare namespace Knex {
  interface Interface {
    method(): ShouldJustBeAny;
  }
}

export = knex;

// @filename: index.ts

import "knex";
declare module "knex" {
  namespace Knex {
    function newFunc(): Knex.Interface;
  }
}
```

jakebailey:
Debugging, what I'm seeing is that in `transformAndReplaceLatePaintedStatements` we walk nodes in the `node_modules` source file (the wrong file), but I think that was already said.

jakebailey:
Yeah, so the code that we have now _is_ the same as Strada and behaves the same. We really do attempt to ask for position information out of the wrong file. This only succeeds because `GetLineAndCharacterOfPosition` allows it in `Strada`. But Corsa's implementation is different and _doesn't_.

I think the fix is either to make `GetLineAndCharacterOfPosition` behave more like Strada (here be UTF-8 dragons), or actually do get the original node for the input, then get its position info from its own source file if one exists.

hamidrezahanafi:
Updated based on your suggestions, let me know if the actual fix is something else

hamidrezahanafi:

jakebailey:
I would suspect that you need to do `MostOriginal` on `input` (that's what I meant by `original`, our specific terminology), just in case we ever pass in something that has been transformed. But I suspect that is never the case, so maybe not.

internal/transformers/declarations/transform.go
```diff
@@ -468,8 +468,12 @@ func (tx *DeclarationTransformer) visitDeclarationSubtree(input *ast.Node) *ast.
...
-			endLine, _ := scanner.GetLineAndCharacterOfPosition(tx.state.currentSourceFile, input.Loc.End())
+			sourceFileForLocation := tx.state.currentSourceFile
+			if nodeFile := ast.GetSourceFileOfNode(input); nodeFile != nil && nodeFile != sourceFileForLocation {
+				sourceFileForLocation = nodeFile
+			}
+			startLine, _ := scanner.GetLineAndCharacterOfPosition(sourceFileForLocation, input.Loc.Pos())
+			endLine, _ := scanner.GetLineAndCharacterOfPosition(sourceFileForLocation, input.Loc.End())
 			if startLine == endLine {
 				tx.EmitContext().AddEmitFlags(result, printer.EFSingleLine)
 			}
```
jakebailey:
```suggestion
			original := tx.EmitContext().MostOriginal(input)
			originalSourceFile := ast.GetSourceFileOfNode(original)
			if originalSourceFile != nil {
				startLine, _ := scanner.GetLineAndCharacterOfPosition(originalSourceFile, original.Loc.Pos())
				endLine, _ := scanner.GetLineAndCharacterOfPosition(originalSourceFile, original.Loc.End())
				if startLine == endLine {
					tx.EmitContext().AddEmitFlags(result, printer.EFSingleLine)
				}
			}
```

hamidrezahanafi: done

weswigham (approved):
Now this, I like. Honestly, the conditional logic should probably be extracted into a helper function `IsOriginalNodeSingleLine`, since this should probably get used elsewhere, too.

From what @jakebailey  said, dollars to donuts, this doesn't crash strada (no bounds checks on the slice), but does result in incorrect output single-lined-ness, and so this change aughta also get ported back to https://github.com/microsoft/TypeScript too.

hamidrezahanafi:
something like this on `transform.go`?
```
func (tx *DeclarationTransformer) isOriginalNodeSingleLine(node *ast.Node) bool {
	if node == nil {
		return false
	}
	original := tx.EmitContext().MostOriginal(node)
	if original == nil {
		return false
	}
	source := ast.GetSourceFileOfNode(original)
	if source == nil {
		return false
	}
	startLine, _ := scanner.GetLineAndCharacterOfPosition(source, original.Loc.Pos())
	endLine, _ := scanner.GetLineAndCharacterOfPosition(source, original.Loc.End())
	return startLine == endLine
}
```

jakebailey:
I would make it a top level function accepting `*EmitContext` put into `internal/transformers/utilities.go` (see the other helpers there).

jakebailey (approved):

jakebailey:
Thank you!

Issue #1415, opened 2025-07-18
Signature help crash on parenthesized `await`


DanielRosenwasser:
## Stack trace

```
panic handling request textDocument/signatureHelp runtime error: invalid memory address or nil pointer dereference goroutine 6021 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x64
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1.1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:357 +0x54
panic({0x7ff6a477cbc0?, 0x7ff6a4fa1390?})
	runtime/panic.go:792 +0x124
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).GetSignatureHelpItems(0x4011b55b18?, {0x7ff6a4a79d10?, 0x4016d0a1e0?}, 0x19, 0x7ff6a49256bc?, 0x4011e242c8, 0x4012747ea0, 0x40000059e0, 0x4011cf6380?)
	github.com/microsoft/typescript-go/internal/ls/signaturehelp.go:103 +0x10c
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideSignatureHelp(0x4012421290, {0x7ff6a4a79d10, 0x4016d0a1e0}, {0x40128dfef0?, 0x0?}, {0x0?, 0x0?}, 0x4012747ea0, 0x40000059e0, 0x400e983e70)
	github.com/microsoft/typescript-go/internal/ls/signaturehelp.go:47 +0x98
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleSignatureHelp(0x4000000b40, {0x7ff6a4a79d10, 0x4016d0a1e0}, 0x40129fd230)
	github.com/microsoft/typescript-go/internal/lsp/server.go:679 +0xe4
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x400e983f68?, {0x7ff6a4a79d10?, 0x4016d0a1e0?}, 0x7ff6a46138a0?)
	github.com/microsoft/typescript-go/internal/lsp/server.go:492 +0x2c0
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:370 +0x70
created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 9
[...]

PR #1527, opened 2025-08-06
Fix signature help crash

jakebailey:
No test caught this, but it's a trivially correct port. I am not sure how to make it happen, I just saw it while typing.


jakebailey:
Maybe this is #1415/#1421?

jakebailey:
Yeah, it totally is. Will try and make a test.

jakebailey:
Testing this requires `signatureHelpPresentForTriggerReason` to be ported in the fourslash test, because we only test sig help baselines without context and this bug requires context.

DanielRosenwasser (approved):

DanielRosenwasser (approved):

PR #1533, opened 2025-08-07
fix: make ResponseMessage.Result omitzero

skewb1k:
According to the JSON-RPC specification, a response object must include either a `result` field or an `error` field, but not both.

Without omitzero, if `Result` is nil, it is marshalled as 'null', causing both `result` and `error` fields to appear in the JSON object when `Error` is also present, which violates the specification.

### DIFF
--- a/internal/lsp/lsproto/jsonrpc.go
+++ b/internal/lsp/lsproto/jsonrpc.go
@@ -207,7 +207,7 @@ func (r *RequestMessage) UnmarshalJSON(data []byte) error {
 type ResponseMessage struct {
 	JSONRPC JSONRPCVersion `json:"jsonrpc"`
 	ID      *ID            `json:"id,omitzero"`
-	Result  any            `json:"result"`
+	Result  any            `json:"result,omitzero"`
 	Error   *ResponseError `json:"error,omitzero"`
 }
 


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -207,7 +207,7 @@ func (r *RequestMessage) UnmarshalJSON(data []byte) error {
 type ResponseMessage struct {
 	JSONRPC JSONRPCVersion `json:"jsonrpc"`
 	ID      *ID            `json:"id,omitzero"`
-	Result  any            `json:"result"`
+	Result  any            `json:"result,omitzero"`
```
jakebailey, line 210:
```suggestion
	Result  *any           `json:"result,omitzero"`
```

I suspect that this would be needed to be able to still encode an explicit null, which many requests can respond with.

jakebailey:
What client are you using that is so strict?

internal/lsp/lsproto/jsonrpc.go
arkriny: I initially thought the same, but it still works because for null fields, response structs have custom marshal implementations that emit `jsontext.Null` token directly. (At least for Hover, this is the case.)

jakebailey: Yeah, most are explicit like that, but I think some do need to explicitly return `null` like the `shutdown` request (the rest I think are server-to-client where we never send anything). Perhaps we could do your existing PR but just create some new explicit `null` value or something for that one case...

arkriny:
> What client are you using that is so strict?

Neovim on the master branch. This is actually caused by my pull request that added the capability to distinguish between `null` and `undefined` fields, so now it can detect responses like `{result: null, error: {}}`. I agree it might be too strict and should be ignored, relying only on the presence of the error field. These two flaws actually help catch each other :).
Still, I think it’s good practice for servers not to violate the JSON-RPC spec.

internal/lsp/lsproto/jsonrpc.go
jakebailey: I have a change which does the above (explicit null types for these cases); do you mind if I push it into this PR? Together I think the change would be correct.

arkriny: @jakebailey Of course, go ahead - that would be great.

jakebailey (approved):
LGTM but since I pushed to it last, I can no longer approve and merge it 😅

jakebailey:
Technically you could merge main and then I could merge the PR 😉

jakebailey (approved):
Thanks!

arkriny:
> Technically you could merge main and then I could merge the PR 😉

Wow, it feels like this shouldn’t work that way.

jakebailey:
Sorta kinda, I could have also sent code via GitHub code suggestions and you could have merged them in via the UI and it would still commit "as me" and do the same thing, so, oh well.

PR #1537, opened 2025-08-07
Port --traceResolutions

sheetalkamat:


testdata/baselines/reference/submodule/compiler/allowJsCrossMonorepoPackage.trace.json.diff
```diff
@@ -0,0 +1,50 @@
+--- old.allowJsCrossMonorepoPackage.trace.json
++++ new.allowJsCrossMonorepoPackage.trace.json
+@@= skipped -11, +11 lines =@@
+ File '/packages/main/node_modules/shared/index.d.ts' does not exist.
+ Directory '/packages/main/node_modules/@types' does not exist, skipping all lookups in it.
+ Directory '/packages/node_modules' does not exist, skipping all lookups in it.
++Directory '/packages/node_modules/@types' does not exist, skipping all lookups in it.
```
sheetalkamat: seems like we are adding these extra lookups

testdata/baselines/reference/submodule/compiler/allowJsCrossMonorepoPackage.trace.json.diff
```diff
@@ -0,0 +1,50 @@
...
+ File '/node_modules/shared.ts' does not exist.
+ File '/node_modules/shared.tsx' does not exist.
+ File '/node_modules/shared.d.ts' does not exist.
+ Directory '/node_modules/@types' does not exist, skipping all lookups in it.
+-Resolving real path for '/packages/main/node_modules/shared/index.js', result '/packages/shared/index.js'.
+-======== Module name 'shared' was successfully resolved to '/packages/shared/index.js' with Package ID 'shared/index.js@1.0.0'. ========
++======== Module name 'shared' was successfully resolved to '/packages/shared/index.js' with Package ID 'shared@1.0.0'. ========
```
sheetalkamat: package id seems to be different

testdata/baselines/reference/submodule/compiler/allowJsCrossMonorepoPackage.trace.json.diff
```diff
@@ -0,0 +1,50 @@
...
+ Directory '/packages/node_modules' does not exist, skipping all lookups in it.
++Directory '/packages/node_modules/@types' does not exist, skipping all lookups in it.
+ File '/node_modules/pkg/package.json' does not exist.
+ File '/node_modules/pkg.ts' does not exist.
+ File '/node_modules/pkg.tsx' does not exist.
+@@= skipped -10, +12 lines =@@
+ File '/node_modules/pkg/index.d.ts' exists - use it as a name resolution result.
+ Resolving real path for '/node_modules/pkg/index.d.ts', result '/node_modules/pkg/index.d.ts'.
+ ======== Module name 'pkg' was successfully resolved to '/node_modules/pkg/index.d.ts'. ========
+-File '/node_modules/pkg/package.json' does not exist according to earlier cached lookups.
```
sheetalkamat: i didnt add traces for package .json lookups for file's metadata that we use to in strada as part of this PR

testdata/baselines/reference/submodule/compiler/allowJsCrossMonorepoPackage.trace.json.diff
jakebailey: I suspect this comes down us not having ported classic/node10 resolution

testdata/baselines/reference/submodule/compiler/allowJsCrossMonorepoPackage.trace.json.diff
jakebailey: All of this caching was removed, because the FS itself caches. We just let the resolver go wild.

jakebailey:
I suspect diffing the resolutions won't be useful due to the things noted above; we don't have classic/node10 resolution and we removed all deduplication from the resolver because it was far faster to just let the FS cache things.

sheetalkamat:
> tion from the resolver because it was far faster to just let the FS cache things.

So do we not want to diff with old but baseline `trace.json`

jakebailey:
I am not sure; I think we had been concerned about the output even being consistent given it's now happening concurrently. It implies that the resolutions can't be streamed to an output without being interleaved.

sheetalkamat:
> I am not sure; I think we had been concerned about the output even being consistent given it's now happening concurrently. It implies that the resolutions can't be streamed to an output without being interleaved.

The way this is implemented it guaranetees the order of the trace
The output is always after the trace so shouldnt not interfere with that. 

i am not sure if diffing with old is useful or not and who is looking at it so may be i will just skip that part but baseline the trace so we can detect the changes

internal/testutil/baseline/baseline.go
```diff
@@ -42,7 +43,7 @@ func Run(t *testing.T, fileName string, actual string, opts Options) {
 		writeComparison(t, actual, localPath, referencePath, false)
 	}
 
-	if !opts.IsSubmodule {
+	if !opts.IsSubmodule || opts.SkipDiffWithOld {
```
jakebailey, line 46: Is this still needed? I think it's leftover now that the diffs were dropped.

internal/compiler/host.go
```diff
@@ -68,7 +72,7 @@ func (h *compilerHost) GetCurrentDirectory() string {
 }
 
 func (h *compilerHost) Trace(msg string) {
-	//!!! TODO: implement
+	h.trace(msg)
```
jakebailey, line 78: If this gains a nil check, you could simplify the other callers to just not bother.

internal/testutil/baseline/baseline.go
sheetalkamat: This is needed to ensure that it does not diff with old 
Should i be setting "submodule" as false instead but then the trace will go in different location compared to their emit and other stuff ?

jakebailey: Hm, you're right. I thought we had a way to do this before but I'm not seeing it anymore.

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -516,17 +516,39 @@ func (h *cachedCompilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast
...
 type tracer struct {
-	fs               vfs.FS
-	currentDirectory string
-	packageJsonCache map[tspath.Path]bool
-	builder          strings.Builder
+	fs                         vfs.FS
+	currentDirectory           string
+	packageJsonCache           map[tspath.Path]bool
+	builder                    strings.Builder
+	specifiTypesVersionMessage bool
```
jakebailey: Nit: typo here

internal/testutil/harnessutil/harnessutil.go
```diff
@@ -511,9 +515,97 @@ func (h *cachedCompilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast
...
+		!t.noTypesVersionMessage {
+		fmt.Fprintln(&t.builder, "'package.json' does not have a 'typesVersions' field.")
+	}
+	t.noTypesVersionMessage = false
+	t.specifiTypesVersionMessage = false
+	// Version
+	if str := strings.Replace(msg, "'"+core.Version()+"'", "'"+FakeTSVersion+"'", 1); str != msg {
+		return str
+	}
+	// caching of fs in trace to be replaces with non caching version
```
jakebailey, line 534: Just to be clear, this all is sort of a hack to work around the FS caching and produce consistent traces in tests?

sheetalkamat: Yes..if you last few commits i didnt change anything in code just made traces consistent and show "caching" results in the traces or show them consistently when we can

sheetalkamat:
there is one more race with the caching log, let me fix that

jakebailey (approved):

jakebailey:
Ah, right, merge conflict with that other PR 😄

jakebailey (approved):

Issue #137, opened 2024-12-06
Incorrect, stray hashes parse differently than tsc


sandersn:
See _submodules/TypeScript/tests/cases/conformance/classes/members/privateNames/privateNameHashCharName.ts

RyanCavanaugh:
Won't fix, not an invariant

PR #1542, opened 2025-08-08
Port class fields transformer

chengcyber:
### Summary

This PR ports the `classFieldsTransformer` from Strada into the current codebase.

### Details

The motivation for this change originates from this [issue](https://github.com/microsoft/typescript-go/issues/137). However, after digging into the implementation, it became clear that the scope of work is significantly more complex than initially expected.

At this stage, the PR fixes the transformation logic for the `privateNameHashCharName` test case. Please check https://github.com/microsoft/typescript-go/commit/78458b3079e2ecc7eed033f2beda5700e1f8ff8b

Before proceeding further, I'd like to confirm that this direction aligns with project goals and that this contribution is welcome. Once confirmed, I plan to continue addressing the remaining related test cases.

### Test

This PR reduces the test result difference between the Strada and Corsa, brining them closer to parity.


weswigham (commented):
It's ambitious to take this on - this is one of the largest transforms in the TS codebase. Correspondingly, it's kinda difficult to judge how "right" this is without all of it in place. I can say "great, one less diff" - but with tests failing with unaccepted diffs it's a 🤷‍♂️ if it's actually net-positive, even for just the currently claimed logic. Still, aside from the spelling errors copilot picked up on, I see a few other fixable issues:

1. The substitution pass is no more (intentionally). Anything referencing them needs to be reimplemented with direct transformations. Usually not a problem, but it's not a direct port of the substitution logic like I partially see here.
2. Closures. Closures need to be avoided, generally - we used them liberally in TS, but in the `go` codebase they've proven to generally have an outsized performance cost. Most, if not all closures, and auto-closing function/method references, need to be replaced with either static function refs or captured a single time on transform creation. Or just eliminated entirely.
3. This falls more into the "open questions" category - in the TS repo, we unconditionally execute a single class fields transform that has a bunch of internal switches for param properties,get/set semantics, auto accessors, private name support, static private name support, etc. While I did stub out a single empty transform, it might ultimately make more sense to have each of those sub-features all in separate transformers for the sake of being able to modularly compose them sensibly (moving the configuration mostly out of the transformer logic and into initial transformer selection). We've broken down other transforms by reasonably logical feature in this way already, and some functionality has already moved, as noted, to other transformers, like parameter properties. I'd probably prefer more subdivision here, but it's a bit of extra work. Doing so might make piecemeal implementations like this a bit easier to digest one bit at a time, though.

chengcyber:
Hey @weswigham ,thanks for the detailed feedback! As I mentioned earlier, I only realized during implementation just how complex this transformer is — far beyond my initial expectation. As the code size grew, I started to suspect I might be heading in the wrong direction.

My initial intent was to fix a single test case and then immediately pause, using this PR as a way to “test the waters” and invite feedback from existing contributors so we could adjust the approach early. Even if this PR doesn’t get merged, I completely understand.

I really appreciate the suggestions and review. I’ll take some time to get a deeper feel for the current design philosophy, and then explore whether it’s possible to narrow this down to a well-scoped subdivision so that any contribution stays aligned and ensures we’re on the same page.

chengcyber:
Hi @weswigham 

I’ve updated this PR based on the earlier review feedback.

The key change is introducing a dedicated `classPrivateFieldsTransformer` for handling private fields, rather than relying solely on `classFieldsTransformer`.

At this stage, I’ve only implemented part of the logic as a proof of concept (not all cases yet). Could you take another look and let me know if this seems to be heading in the right direction?

jakebailey:
Thanks for sending this; there was enough to change that I opted to just do it from scratch in #2956. So, closing this one.

PR #1550, opened 2025-08-10
Reduce string copies and allocations when emitting inlined sourcemaps

dzbarsky:


internal/sourcemap/generator.go
```diff
@@ -333,13 +335,26 @@ func (gen *Generator) RawSourceMap() *RawSourceMap {
...
+
+// Gets the string representation of the source map
+func (gen *Generator) String() string {
+	return string(gen.bytes())
+}
+
+func (gen *Generator) Base64DataURL() string {
+	const prefix = "data:application/json;base64,"
+	data := gen.bytes()
+	buf := make([]byte, len(prefix)+base64.StdEncoding.EncodedLen(len(data)))
```
dzbarsky: this can also be done via `strings.Builder` with `Grow` and `Write` if preferred, it's essentially the same `unsafe.String` trick under the hood

dzbarsky:
@microsoft-github-policy-service agre

dzbarsky:

internal/sourcemap/generator.go
jakebailey: Yes, avoiding unsafe is preferred.

internal/sourcemap/generator.go
```diff
@@ -1,11 +1,13 @@
 package sourcemap
 
 import (
+	"encoding/base64"
+	"encoding/json"
```
jakebailey: Not really sure why you swapped out the json import? (Must have been auto import's fault)

jakebailey:
I updated this PR, however, if you trace this back out, it goes to:

```go
e.writer.WriteComment("//# sourceMappingURL=" + sourceMappingURL)
```

So, it's not like we aren't doing random concats anyway.

internal/compiler/emitter.go
```diff
@@ -269,7 +267,8 @@ func (e *emitter) printSourceFile(jsFilePath string, sourceMapFilePath string, s
 				e.writer.RawWrite(core.IfElse(options.NewLine == core.NewLineKindCRLF, "\r\n", "\n"))
 			}
 			sourceMapUrlPos = e.writer.GetTextPos()
-			e.writer.WriteComment("//# sourceMappingURL=" + sourceMappingURL)
+			e.writer.WriteComment("//# sourceMappingURL=")
+			e.writer.WriteComment(sourceMappingURL)
```
jakebailey, line 271: This looks suspicious, but it's actually fine to do two calls here

PR #1551, opened 2025-08-10
Avoid copying content when writing files

dzbarsky:
A well-behaved FS implementation should not mutate the bytes to be written. (Also, it seems like many call sites already have an `[]byte` and convert to a string just to call WriteFile)

### DIFF
--- a/internal/vfs/iovfs/iofs.go
+++ b/internal/vfs/iovfs/iofs.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/fs"
 	"strings"
+	"unsafe"
 
 	"github.com/microsoft/typescript-go/internal/stringutil"
 	"github.com/microsoft/typescript-go/internal/tspath"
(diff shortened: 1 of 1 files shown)


jakebailey:
> (Also, it seems like many call sites already have an `[]byte` and convert to a string just to call WriteFile)

For fun, I went to try and refactor file writing to see if using `[]byte` were possible all the way, but it is a lot of changes. Reading as strings is pretty intentional so things kinda bubble from there. But maybe it's possible.

It'd be interesting to see your benchmarks for your PRs.

jakebailey:
@dzbarsky Did you have benchmarks for this, by chance? Not against the change, but I want any use of `unsafe` to be well-justified.

jakebailey:
It turns out that we can just use a different signature. I then did more reading and decided to go futher: #3008

dzbarsky:
Cool, thanks!

PR #1553, opened 2025-08-11
autoimport completions

iisaduan:
This PR implements autoimports, and allows symbols to be autoimported through completions. Also ports the basic functionalities of the changetracker (adds changes and calculates/formats the output string).

demo (note: the formatting has been fixed, see [test cases](https://github.com/microsoft/typescript-go/pull/1553/files#diff-d4159983ef63037437a36a07bad5a4600b60a10a8d07dd756fd07b467c50b27cR9))

https://github.com/user-attachments/assets/f28235f3-d25e-4e23-9015-42001e608b5d

Notes on implementation:
- Caching structures (like `ExportInfoMap`, `ImportMap`, etc) have been ported for functionality, but are not saved between server requests). After the ls is more fully implemented and perf tested, we will [revisit](https://github.com/microsoft/typescript-go/pull/1553#discussion_r2283257877) to see what caching we need to do 
- The `IncludeCompletionsForModuleExports` and `IncludeCompletionsForImportStatements` in the `UserPreferences` passed into any completion get/resolve request has been set to `true` to allow autoimports to work https://github.com/microsoft/typescript-go/pull/1553#discussion_r2279606193
- changetracker formatting workarounds: https://github.com/microsoft/typescript-go/pull/1553#discussion_r2283577277

Related features to be implemented after this PR: 
- autoimport user settings/format settings
- package json autoimport provider ([probably mostly a rewrite](https://github.com/microsoft/typescript-go/pull/1553#discussion_r2283257877))
- organize imports
- codefix provider (to be able to use autoimports without completions)


alexwork1611:
I'm gonna closely follow this one...

iisaduan:
bug squashing atm :). Hopefully it'll be ready soon

internal/ls/completions.go
```diff
@@ -95,8 +98,8 @@ type completionDataData struct {
 	isJsxIdentifierExpected   bool
 	isRightOfOpenTag          bool
 	isRightOfDotOrQuestionDot bool
-	importStatementCompletion any  // !!!
-	hasUnresolvedAutoImports  bool // !!!
+	importStatementCompletion *importStatementCompletionInfo // !!!
```
gabritto, line 104: You can drop the `// !!!` here now I think

internal/ls/types.go
```diff
@@ -18,15 +19,26 @@ const (
...
+	IncludeCompletionsWithSnippetText *bool
+
 	// Unless this option is `false`,  member completion lists triggered with `.` will include entries
 	// on potentially-null and potentially-undefined values, with insertion text to replace
 	// preceding `.` tokens with `?.`.
 	IncludeAutomaticOptionalChainCompletions *bool
 
+	// If enabled, the completion list will include completions with invalid identifier names.
+	// For those entries, The `insertText` and `replacementSpan` properties will be set to change from `.x` property access to `["x"]`.
+	IncludeCompletionsWithInsertText *bool
```
gabritto: We don't need this anymore, every LSP client should support this.

internal/ls/types.go
```diff
@@ -18,15 +19,26 @@ const (
...
+	// If enabled, TypeScript will search through all external modules' exports and add them to the completions list.
+	// This affects lone identifier completions but not completions on the right hand side of `obj.`.
+	IncludeCompletionsForModuleExports *bool
+
 	// Enables auto-import-style completions on partially-typed import statements. E.g., allows
 	// `import write|` to be completed to `import { writeFile } from "fs"`.
 	IncludeCompletionsForImportStatements *bool
 
+	// Allows completions to be formatted with snippet text, indicated by `CompletionItem["isSnippet"]`.
+	IncludeCompletionsWithSnippetText *bool
```
gabritto: We also don't need this as a user preference. Instead, we need to check completion client capabilities for `snippetSupport`.

internal/ls/completions.go
```diff
@@ -4855,8 +5399,226 @@ func createCompletionDetailsForSymbol(
...
+	moduleSymbol := symbolDetails.origin.moduleSymbol()
+
+	_, importCompletionAction := l.getImportCompletionAction(
+		ctx,
+		ch,
+		targetSymbol,
+		moduleSymbol,
+		file,
+		position,
+		"",          // entryId.data.exportMapKey
```
gabritto: What are these comments for? Is it the corresponding Strada code?

internal/ls/completions.go
```diff
@@ -992,13 +1073,200 @@ func getCompletionData(program *compiler.Program, typeChecker *checker.Checker,
...
+		if file.ExternalModuleIndicator != nil || file.CommonJSModuleIndicator != nil {
+			return true
+		}
+		// If module transpilation is enabled or we're targeting es6 or above, or not emitting, OK.
+		if compilerOptionsIndicateEsModules(program.Options()) {
+			return true
+		}
+
+		// If some file is using ES6 modules, assume that it's OK to add more.
+		return true || // !!! symlink program.getSymlinkCache?.().hasAnySymlinks() ||
```
gabritto: Shouldn't this be `false` while the symlink condition isn't implemented yet?

internal/fourslash/fourslash.go
```diff
@@ -731,7 +731,7 @@ func (f *FourslashTest) verifyCompletionsAreExactly(t *testing.T, prefix string,
 var completionIgnoreOpts = cmp.FilterPath(
 	func(p cmp.Path) bool {
 		switch p.Last().String() {
-		case ".Kind", ".SortText", ".Data":
```
gabritto: We have completion tests that test `filterText`, so we shouldn't always ignore it. If you need to ignore it for some tests, I'd suggest adding an explicit `ignore` option like we do for other properties (see `ExpectedCompletionEditRange`). I think ideally we'd do what we did in Strada where we checked if the property was defined, but since that's not a thing in Go it's trickier to allow for testing presence/absence of `filterText` when we want to but ignore it the rest of the time. If it's just for a few tests, we can probably manually fill in `filterText`, otherwise we can set up an explicit `ignore`. I've been meaning to do that for a while now, let me know if you want me to do that and what tests are failing without this change it.

internal/scanner/scanner.go
```diff
@@ -2428,6 +2428,10 @@ func GetLineStarts(sourceFile ast.SourceFileLike) []core.TextPos {
 	return sourceFile.LineMap()
 }
 
+func GetLineStartPositionForPosition(pos int, lineMap []core.TextPos) int {
```
gabritto: Not sure if it does what you need, but we have a function with the same name in `internal/format/util.go`.

internal/ls/completions.go
```diff
@@ -3402,6 +3399,49 @@ func getCompareCompletionEntries(ctx context.Context) func(entryInSlice *lsproto
...
+// False:
+// 'staet' in 'useState'
+// 'tate' in 'useState'
+// 'ment' in 'ENVIRONMENT_VARIABLE'
+func charactersFuzzyMatchInString(identifierString string, lowercaseCharacters string) bool {
+	if lowercaseCharacters == "" {
+		return true
+	}
+
+	var prevChar *rune
```
jakebailey: I suspect you might be able to get away with using `0` here, given an identifer can't contain `NUL`.

internal/ls/completions.go
iisaduan: I think I mistranslated as `&&` in my brain

internal/scanner/scanner.go
iisaduan: The new function is because the existing `format/util` function re-calculates the linemap every time

internal/format/indent.go
```diff
@@ -139,7 +139,7 @@ func getActualIndentationForListItem(node *ast.Node, sourceFile *ast.SourceFile,
 		// VariableDeclarationList has no wrapping tokens
 		return -1
 	}
-	containingList := getContainingList(node, sourceFile)
+	containingList := GetContainingList(node, sourceFile)
```
jakebailey, line 143: I didn't know this func existed; I wonder if it's general enough to use for sig help. (No action needed, just opining)

internal/format/indent.go
```diff
@@ -196,10 +196,10 @@ func deriveActualIndentationFromList(list *ast.NodeList, index int, sourceFile *
 
 func findColumnForFirstNonWhitespaceCharacterInLine(line int, char int, sourceFile *ast.SourceFile, options *FormatCodeSettings) int {
 	lineStart := scanner.GetPositionOfLineAndCharacter(sourceFile, line, 0)
-	return findFirstNonWhitespaceColumn(lineStart, lineStart+char, sourceFile, options)
+	return FindFirstNonWhitespaceColumn(lineStart, lineStart+char, sourceFile, options)
 }
 
-func findFirstNonWhitespaceColumn(startPos int, endPos int, sourceFile *ast.SourceFile, options *FormatCodeSettings) int {
+func FindFirstNonWhitespaceColumn(startPos int, endPos int, sourceFile *ast.SourceFile, options *FormatCodeSettings) int {
```
jakebailey, line 203: I am suspicious of this function, given it appears to return the column in terms of runes? I'm not sure it walks the chars correctly either, hmm...

internal/ls/changetrackerimpl.go
jakebailey, line 1: Seems like this doesn't need the `impl` suffix in the filename? It's just the one file.

internal/printer/printer.go
```diff
@@ -45,10 +45,10 @@ type PrinterOptions struct {
 	OmitBraceSourceMapPositions bool
 	// ExtendedDiagnostics           bool
 	OnlyPrintJSDocStyle bool
-	// NeverAsciiEscape              bool
+	NeverAsciiEscape    bool // !!!
```
jakebailey: I implemented this in main so this should merge in

internal/lsp/server.go
```diff
@@ -759,7 +759,10 @@ func (s *Server) handleCompletion(ctx context.Context, params *lsproto.Completio
 		params.Position,
 		params.Context,
 		getCompletionClientCapabilities(s.initializeParams),
-		&ls.UserPreferences{})
+		&ls.UserPreferences{
```
jakebailey, line 760: Long term I think we're going to want to pass this info down a different way but this is fine for now of course.

jakebailey:
The import in the demo is inserted in a non-pretty way; is that something being worked on?

iisaduan:
> The import in the demo is inserted in a non-pretty way; is that something being worked on?

Yes, this is a formatting bug/limitation. Strada formats with a `sourceFileLike`, and theres a lot of places in the formatter where synthesized nodes aren't handled

Edit: Fixed with help from @weswigham  New statements come out pretty now (haven't updated the demo vid yet tho)

internal/scanner/scanner.go
gabritto: Can the version in `format` be replaced by this one?

internal/format/indent.go
weswigham: This is what I was writing when I first went "wait, do we work in runes or in byte offsets for columns? Looks inconsistent right now". Both ways look wrong to some part of the stack around here because of that inconsistency.

internal/ast/utilities.go
```diff
@@ -2648,6 +2658,10 @@ func nodeContainsPosition(node *Node, position int) bool {
 	return node.Kind >= KindFirstNode && node.Pos() <= position && (position < node.End() || position == node.End() && node.Kind == KindEndOfFile)
 }
 
+func NodeRangeContainsPosition(node *Node, pos int) bool {
```
weswigham: All uses of this can just be `node.Loc.ContainsInclusive(pos)` - our text ranges have methods now.

internal/format/indent.go
jakebailey: Yeah, I think we need _everything_ to be UTF-8 byte offsets, until they hit the LS and get converted based on the client preferences.

The only exception might be symbol baselines (UTF-16 compat?), source maps, and maybe the API, and _maybe_ printed diagnostics? (That's a lot of exceptions...)

Related, #1578

weswigham (changes requested):

internal/ls/utilities.go
```diff
@@ -39,6 +41,13 @@ func CompareRanges(lsRange, other *lsproto.Range) int {
 	return ComparePositions(lsRange.End, other.End)
 }
 
+func applyTextChanges(text string, changes []core.TextChange) string {
+	for i := len(changes) - 1; i >= 0; i-- {
+		text = changes[i].ApplyTo(text)
```
weswigham: When there are many edits (like, say, 30000 to re-indent every line of `checker.ts`), this algorithm produces a _huge_ quantity of trash strings (gigabytes worth). There's an `applyBulkEdits` function in `tsc.go` that's already worked around the issue by avoiding individual `.ApplyTo` calls and using a string builder. You'll probably want to move it and use it in the LS, too.

internal/fourslash/tests/gen/completionsObjectLiteralModuleExports_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func TestCompletionsObjectLiteralModuleExports(t *testing.T) {
 	t.Parallel()
-
+	t.Skip()
```
gabritto: Do you know why this test now fails? We can always fix them later, but I'm curious.

internal/fourslash/tests/manual/autoImportCompletion_test.go
```diff
@@ -0,0 +1,145 @@
+package fourslash_test
```
gabritto: Technically the tests in `internal/fourslash/tests/manual/` are supposed to be generated tests that had to be manually modified, and new tests go into `internal/fourslash/tests/`, though that doesn't have practical differences I think other than our own organization of things.

internal/ls/autoimportfixes.go
```diff
@@ -0,0 +1,324 @@
...
+			return i.addAsTypeOnly == AddAsTypeOnlyNotAllowed
+		})
+
+		existingSpecifiers := []*ast.Node{} // []*ast.ImportSpecifier
+		if importClause.NamedBindings != nil && importClause.NamedBindings.Kind == ast.KindNamedImports {
+			existingSpecifiers = importClause.NamedBindings.Elements()
+		}
+
+		if defaultImport != nil {
+			// Debug.assert(!clause.name, "Cannot add a default import to an import clause that already has one");
```
gabritto: If #1593 gets merged before this PR, this can be updated. 🙂

internal/scanner/scanner.go
iisaduan: I was able to unify them. I don't know what I remembered wrong (maybe it was a different function?), but the linemap is not calculated every time

internal/fourslash/tests/gen/completionsObjectLiteralModuleExports_test.go
iisaduan: fixed the test case, it happened due to a missed `parent == nil` check that was added when this PR is combined with merge changes

internal/ast/utilities.go
```diff
@@ -3013,7 +3032,7 @@ func IsJSDocSingleCommentNodeList(nodeList *NodeList) bool {
 		return false
 	}
 	parent := nodeList.Nodes[0].Parent
-	return IsJSDocSingleCommentNode(parent) && nodeList == parent.CommentList()
+	return parent != nil && IsJSDocSingleCommentNode(parent) && nodeList == parent.CommentList()
```
gabritto: When does a node in a node list have no parents? 🤔

internal/ls/types.go
```diff
@@ -40,4 +39,19 @@ type UserPreferences struct {
 	IncludeCompletionsWithObjectLiteralMethodSnippets *bool
 
 	JsxAttributeCompletionStyle *JsxAttributeCompletionStyle
+
+	ImportModuleSpecifierPreference       modulespecifiers.ImportModuleSpecifierPreference
+	ImportModuleSpecifierEndingPreference modulespecifiers.ImportModuleSpecifierEndingPreference
+	PreferTypeOnlyAutoImports             *bool
+	AllowIncompleteCompletions            *bool
```
gabritto: This doesn't seem to be a user-facing config in vscode. Should this just be removed? I think every LSP client has to support incomplete completions.

internal/ls/completions.go
```diff
@@ -1641,6 +1908,92 @@ func keywordCompletionData(
...
+func (r *resolvingModuleSpecifiersForCompletions) tryResolve(ch *checker.Checker, exportInfo []*SymbolExportInfo, isFromAmbientModule bool) (*ImportFix, string) {
+	if isFromAmbientModule {
+		if result := r.resolver.getModuleSpecifierForBestExportInfo(ch, exportInfo, r.position, r.isValidTypeOnlyUseSite); result != nil {
+			r.ambientCount++
+			return result, ""
+		}
+		return nil, "failed"
+	}
+
+	allowIncompleteCompletions := true // !!! preferences ptrIsTrue(r.resolver.UserPreferences.AllowIncompleteCompletions)
```
gabritto: Like I commented above, this should probably not be a user preference (and so the comment needs updating/removing)

internal/ls/autoimportstypes.go
```diff
@@ -0,0 +1,229 @@
...
+		// 		isReExport          *bool
+		// exportInfo          *SymbolExportInfo // !!! | FutureSymbolExportInfo | undefined
+		// moduleSpecifierKind modulespecifiers.ResultKind
+		// moduleSpecifier     string
+		typeOnlyAliasDeclaration: typeOnlyAliasDeclaration,
+	}
+}
+
+/** Information needed to augment an existing import declaration. */
+// rename all fixes to say fix at end
```
gabritto: Should that be a `// !!!`?

internal/ast/utilities.go
iisaduan: If it is a synthesized node

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1642 @@
...
+	targetFlags       ast.SymbolFlags
+	isFromPackageJson *bool
+}
+
+type symbolExportEntry struct {
+	symbol       *ast.Symbol
+	moduleSymbol *ast.Symbol
+}
+
+type ExportMapInfoKey string
```
gabritto: Perhaps silly question: do we need this to be a `string`? Could it not be a struct and avoid the whole parsing/formatting thing?

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1642 @@
...
+	symbols          map[int]symbolExportEntry
+	exportInfoId     int
+	usableByFileName tspath.Path
+	packages         map[string]string
+
+	globalTypingsCacheLocation string
+
+	releaseSymbols func()
+	isEmpty        func() bool
+	/** @returns Whether the change resulted in the cache being cleared */
```
gabritto: Nit: no need to use JSDoc/`@returns`.

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1642 @@
...
+	var exportInfos []*SymbolExportInfo
+	if exportMapKey != "" {
+		// The new way: `exportMapKey` should be in the `data` of each auto-import completion entry and
+		// sent back when asking for details.
+		exportInfos = l.getExportInfoMap(ch, sourceFile, preferences).get(sourceFile.Path(), ch, exportMapKey)
+		if len(exportInfos) == 0 {
+			panic("Some exportInfo should match the specified exportMapKey")
+		}
+	} else {
+		// The old way, kept alive for super old editors that don't give us `data` back.
```
gabritto: Will we need this now that we're LSP? I think "super old editors" that are not following LSP (and therefore not providing `data` back) are not gonna work anyways.

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1642 @@
...
+func (l *LanguageService) getExportInfoMap(
+	ch *checker.Checker,
+	importingFile *ast.SourceFile,
+	preferences *UserPreferences,
+) *exportInfoMap {
+	// Pulling the AutoImportProvider project will trigger its updateGraph if pending,
+	// which will invalidate the export map cache if things change, so pull it before
+	// checking the cache.
+	// l.GetPackageJsonAutoImportProvider?.();
+	// cache := host.getCachedExportInfoMap()
```
gabritto: Do you have an idea of if/how this is going to work post-snapshots?

internal/ast/utilities.go
iisaduan: Did some more thinking, and it might be a bad/too general of a check. I'll see if i can do something to set the parents

internal/ls/autoimports.go
iisaduan: caching-- we will cache if needed after looking at perf once autoimports is pretty much all implemented. There's a chance we don't need to cache anything at all, just like with module resolution.
packageJsonAutoImportProvider--will probably be an entirely new implementation; haven't thought about the specifics too much yet. We do need to have this, otherwise files/modules not currently in/referenced in the program can't be autoimported from

internal/ls/types.go
jakebailey: Yeah, this can just be hardcoded true, if it's controlling `isIncomplete`: https://github.com/microsoft/vscode/blob/9f0d013656299028a056163a95f8ea1972b41a38/extensions/typescript-language-features/src/languageFeatures/fileConfigurationManager.ts#L204

internal/ls/changetrackerimpl.go
```diff
@@ -0,0 +1,413 @@
...
+	shouldRemoveSemicolons := options.Semicolons == format.SemicolonPreferenceRemove || shouldAutoDetectSemicolonPreference && !probablyUsesSemicolons(sourceFile)
+	if shouldRemoveSemicolons {
+		options.Semicolons = format.SemicolonPreferenceRemove
+	}
+
+	return options
+}
+
+/** Note: output node may be mutated input node. */
+func (ct *changeTracker) getNonformattedText(node *ast.Node, sourceFile *ast.SourceFile) (string, *ast.Node) {
```
iisaduan, line 125: The way changetracker creates the input into the formatter is a little hacky. The formatter doesn't work on synthesized nodes, and the `changetrackerwriter` tries to work around this by calculating parents and positions of synthesized nodes. This was done in strada by wrapping any node that needed to be formatted in a `sourceFileLike`, but we can't do the exact same thing here because it creates illegal AST structures and crashes during nodeWriting. However, nodeWriting+formatting output is different given whether the formatted node is or isn't a `sourceFile`, and so in order to mimic the strada formatting, I've had to wrap the nodes at different points, depending on what kind of node it is. Its possible more refinement will be needed later on, as we run into different cases.

internal/fourslash/tests/manual/autoImportCompletion_test.go
jakebailey: Yeah, I would suggest moving this file for clarity.

internal/ls/autoimportstypes.go
```diff
@@ -0,0 +1,228 @@
...
+
+const (
+	ExportKindNamed        ExportKind = 0
+	ExportKindDefault      ExportKind = 1
+	ExportKindExportEquals ExportKind = 2
+	ExportKindUMD          ExportKind = 3
+	ExportKindModule       ExportKind = 4
+)
+
+func (k ExportKind) String() string {
```
jakebailey: You could use `stringer` to generate this, if you do need the method.

jakebailey:
Tried it out, nice:

image

But, the import did appear with the wrong indent level:

image

(I am ignoring the ordering, since that's a different problem.)

internal/ls/changetrackerimpl.go
```diff
@@ -0,0 +1,413 @@
...
+/** Note: this may mutate `nodeIn`. */
+func (ct *changeTracker) getFormattedTextOfNode(nodeIn *ast.Node, targetSourceFile *ast.SourceFile, sourceFile *ast.SourceFile, pos int, options changeNodeOptions) string {
+	text, sourceFileLike := ct.getNonformattedText(nodeIn, targetSourceFile)
+	// !!! if (validate) validate(node, text);
+	formatOptions := getFormatCodeSettingsForWriting(ct.formatSettings, targetSourceFile)
+
+	var initialIndentation, delta int
+	if options.indentation == nil {
+		// !!! indentation for position
+		// initialIndentation = format.GetIndentationForPos(pos, sourceFile, formatOptions, options.prefix == ct.newLine || scanner.GetLineStartPositionForPosition(pos, targetFileLineMap) == pos);
```
iisaduan, line 99: @jakebailey That part of the formatter isn't ported yet (`getSmartIndent`) and involves the `FindFirstNonWhitespaceColumn` functions, so I didn't include it

internal/fourslash/fourslash.go
```diff
@@ -486,6 +486,7 @@ type CompletionsExpectedList struct {
 	IsIncomplete bool
 	ItemDefaults *CompletionsExpectedItemDefaults
 	Items        *CompletionsExpectedItems
+	*ls.UserPreferences
```
andrewbranch: Nit: seems odd to make this an embedding

internal/ls/changetrackerimpl.go
weswigham: I wouldn't assume anything else is gonna necessitate porting more parts of `getSmartIndent` - they're only nominally "in the formatter" - none of the `FormatX` entrypoints use anything as-yet unported. Anything else is basically just part of the quickfix pipeline that lives elsewhere, for some reason.

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1619 @@
...
+
+	symbol            *ast.Symbol // may be nil
+	moduleSymbol      *ast.Symbol // may be nil
+	moduleFileName    string      // may be ""
+	targetFlags       ast.SymbolFlags
+	exportKind        ExportKind
+	isFromPackageJson bool
+}
+
+type exportInfoMap struct {
```
andrewbranch, line 71: I would have liked to see this data structure rethought, for a few reasons:

- Storing symbols is problematic; that's why there's a lot of complexity around serializing and "rehydrating" symbols. Some symbols are checker-generated, which we have to be extra careful of now that we have multiple checkers.
- A lot of up-front checker work goes into building this data structure, which is super tailored to providing global completions with no prefix text, which was justified in Strada by caching it across most edits to the same file. Without that caching in place, or even in a future where we rethink the caching strategy, building this wastes a lot of work a lot of the time.
- The two points above make it unclear how to integrate package.json auto imports - it would be nice to be able to process those without a program/checker, but this data structure relies on having those.
- "Rehydrating" symbol info across requests has been a source of crashes in the past, so it might pay off to revisit from a clean slate.
- A checker-free search of files could be massively parallel; building this is going to be single-threaded over potentially thousands of files.

All that said, it seems like the overriding motivation is to get some form of auto imports implemented fast, so a direct port makes sense. But I see a strong potential that this goes the way of the initial project service port, with a relatively short lifetime before getting a proper rewrite.

internal/ls/autoimports.go
```diff
@@ -0,0 +1,1619 @@
...
+	exportMapKey ExportInfoMapKey,
+	symbolName string, // !!! needs *string ?
+	isJsxTagName bool,
+	// formatContext *formattingContext,
+	preferences *UserPreferences,
+) (string, codeAction) {
+	var exportInfos []*SymbolExportInfo
+	// The new way: `exportMapKey` should be in the `data` of each auto-import completion entry and
+	// sent back when asking for details.
+	exportInfos = l.getExportInfoMap(ch, sourceFile, preferences).get(sourceFile.Path(), ch, exportMapKey)
```
andrewbranch: Probably need to pass `ctx` and check for cancellation where Strada does.

internal/ls/completions.go
```diff
@@ -1641,6 +1908,91 @@ func keywordCompletionData(
...
+func (l *LanguageService) resolvingModuleSpecifiers(
+	logPrefix string,
+	resolver *importSpecifierResolverForCompletions,
+	position int,
+	isForImportStatementCompletion bool,
+	isValidTypeOnlyUseSite bool,
+	cb func(*resolvingModuleSpecifiersForCompletions) *ImportFix,
+) *ImportFix {
+	// !!! timestamp
+	// Under `--moduleResolution nodenext` or `bundler`, we have to resolve module specifiers up front, because
```
andrewbranch: A bunch of code needs to be deleted, because these are the only module resolution modes existing in Corsa / 6.0. All auto import completions should be "fully resolved," which should dramatically simplify the code in this file.

internal/ls/completions.go
```diff
@@ -4654,6 +5094,11 @@ func hasCompletionItem(clientOptions *lsproto.CompletionClientCapabilities) bool
 	return clientOptions != nil && clientOptions.CompletionItem != nil
 }
 
+// strada TODO: this function is, at best, poorly named. Use sites are pretty suspicious.
+func compilerOptionsIndicateEsModules(options *core.CompilerOptions) bool {
+	return options.Module == core.ModuleKindNone || options.GetEmitScriptTarget() >= core.ScriptTargetES2015 || options.NoEmit.IsTrue()
```
andrewbranch, line 4934: Always true in Corsa / 6.0

internal/ls/languageservice.go
```diff
@@ -23,6 +23,11 @@ func (l *LanguageService) GetProgram() *compiler.Program {
 	return l.host.GetProgram()
 }
 
+func (l *LanguageService) GetPackageJsonAutoImportProvider() *compiler.Program {
+	// !!! packageJsonAutoImportProvider
+	return l.GetProgram()
```
andrewbranch: I don't think this should be stubbed this way; couldn't this cause duplicate work?

internal/fourslash/fourslash.go
iisaduan: forgot to update as I cut this part out as userpreferences isn't working in fourslash yet

internal/ls/autoimports.go
iisaduan: Since now the exportInfoMap struct is used at most once per call, I've done a minor refactor to do less initial work in creating the struct. As discussed offline, further refactors and improvements to come as autoimports gets more implemented :)

internal/ls/completions.go
```diff
@@ -1643,6 +1833,57 @@ func keywordCompletionData(
...
+	position                       int
+	isForImportStatementCompletion bool
+	isValidTypeOnlyUseSite         bool
+
+	needsFullResolution    bool
+	skippedAny             bool
+	ambientCount           int
+	resolvedCount          int
+	resolvedFromCacheCount int
+	cacheAttemptCount      int
```
andrewbranch: Can we get rid of all this? We should just make `needsFullResolution` effectively always true and simplify all this. The limiter logic is highly dependent on incomplete completion lists, so that should go away too.

alexwork1611:
I can finally make the VS Code extension my daily driver. Thank you!

PR #1561, opened 2025-08-11
JSDoc completions

gabritto:


internal/ls/completions.go
```diff
@@ -4860,3 +4985,621 @@ func createCompletionDetailsForSymbol(
...
+			Kind:     ptrTo(lsproto.CompletionItemKindKeyword),
+			SortText: ptrTo(string(SortTextLocationPriority)),
+		}
+		items = append(items, item)
+	}
+	return items
+})
+
+func getJSDocTagNameCompletions() []*lsproto.CompletionItem {
+	return cloneItems(jsDocTagNameCompletionItems())
```
jakebailey, line 5171: Is cloning needed? Are they actually modified after the fact? (Just curious)

gabritto: Yes. They get always get `data` and can possibly get e.g. commit characters added.

gabritto: Yes. They get always get `data` and can possibly get e.g. commit characters added.

jakebailey:
> But code in services relies on that special behavior, so I added it directly to `astnav.VisitEachChildAndJSDoc` (in the right way this time, instead of the partial fix I had in place before).

Is there a chance we need the old non-special case way?

internal/ast/utilities.go
```diff
@@ -2104,6 +2104,9 @@ func TryGetTextOfPropertyName(name *Node) (string, bool) {
 }
 
 // True if node is of a JSDoc kind that may contain comment text.
+// NOTE: while this is a faithful port of Strada's implementation,
```
jakebailey: Should Comments just check this and return nil or something?

internal/format/util.go
```diff
@@ -75,7 +75,7 @@ func getCloseTokenForOpenToken(kind ast.Kind) ast.Kind {
 	return ast.KindUnknown
 }
 
-func getLineStartPositionForPosition(position int, sourceFile *ast.SourceFile) int {
+func GetLineStartPositionForPosition(position int, sourceFile *ast.SourceFile) int {
 	lineStarts := scanner.GetLineStarts(sourceFile)
```
jakebailey, line 79: Not for this PR probably but its very suspect how this asks for the line starts again, and it's probably not even the right line starts given this is LS code. We need to audit all of this again but doesn't have to be this PR obviously

internal/ls/completions.go
```diff
@@ -400,8 +444,79 @@ func getCompletionData(program *compiler.Program, typeChecker *checker.Checker,
 	insideJsDocImportTag := false
 	isInSnippetScope := false
 	if insideComment != nil {
-		// !!! jsdoc
-		return nil
+		if hasDocComment(file, position) {
+			if r, _ := utf8.DecodeRuneInString(file.Text()[position:]); r == '@' {
```
jakebailey: For ASCII it's actually totally fine to just directly access the string

internal/ls/completions.go
```diff
@@ -4860,3 +4985,625 @@ func createCompletionDetailsForSymbol(
...
+						nodebuilder.FlagsUseSingleQuotesForStringLiteralType,
+						nodebuilder.FlagsNone,
+					)
+					typeNode := typeChecker.TypeToTypeNode(inferredType, ast.FindAncestor(initializer, ast.IsFunctionLike), builderFlags)
+					if typeNode != nil {
+						emitContext := printer.NewEmitContext()
+						// !!! snippet p
+						p := printer.NewPrinter(printer.PrinterOptions{
+							RemoveComments: true,
+							// !!!
```
jakebailey, line 5360: Willing to bet none of this ever gets added to the printer

gabritto: Wesley said the plan was to have those ported, they are still commented out in the printer code... I'd rather have the reminder and not need it than the other way around.

internal/ast/utilities.go
gabritto: That's not the way it used to work in Strada. You can access `.comments` on the node even if `.comments` is a string, but you wouldn't visit them in Strada's `forEachChild`.

> Is there a chance we need the old non-special case way?

I think we might need a version that visits all things. e.g. in Corsa,`ForEachChild` is used in the binder and so probably needs to visit single comments in JSDoc. We may be able to get away with adding the special behavior to Corsa's `VisitEachChild`, but I'm not sure.

@sandersn might have a better idea.

internal/checker/printer.go
```diff
@@ -372,3 +372,8 @@ func (c *Checker) formatUnionTypes(types []*Type) []*Type {
 	}
 	return result
 }
+
+func (c *Checker) TypeToTypeNode(t *Type, enclosingDeclaration *ast.Node, flags nodebuilder.Flags) *ast.TypeNode {
```
jakebailey, line 376: This code appears to be uncovered by the tests; is there something that is missing that calls this?

gabritto: The Strada tests that cover it are baseline completions tests, which I haven't ported yet (because that's only 31 tests). I'll add a test to minimally cover this for now.

jakebailey (approved):

internal/ast/ast.go
```diff
@@ -1060,6 +1063,32 @@ func (n *Node) QuestionDotToken() *Node {
...
+	case KindJSDocTypedefTag:
+		return n.AsJSDocTypedefTag().TypeExpression
+	case KindJSDocSatisfiesTag:
+		return n.AsJSDocSatisfiesTag().TypeExpression
+	}
+	panic("Unhandled case in Node.TypeExpression: " + n.Kind.String())
+}
+
+func (n *Node) ClassName() *Node {
+	switch n.Kind {
```
sandersn, line 1083: I realise this is intended for JSDoc, but couldn't these entries be included in Name instead of having their own method?
and in general maybe their field should be `name` instead of `ClassName`, but maybe not worth changing--although I don't think we're constrained by public API compatibility yet, right?

internal/ast/utilities.go
sandersn: stupid suggestion because I just started thinking about this, but: JSDocTypeLiteral and JSDocTypeSignature are the bodies of a `typedef` and a `callback` respectively. The reason they don't have comments is that their parent has the comment instead. Why do they return true for this function? Is it because they contain children that could themselves contain comments?

This function is only called in two places, neither of which appear to deal with jsdoc comment text, but finding nodes from a position. My best guess is that the function needs to be renamed so that people don't think Comment() is safe to call on nodes it returns true from.

internal/fourslash/tests/gen/jsdocExtendsTagCompletion_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func TestJsdocExtendsTagCompletion(t *testing.T) {
 	t.Parallel()
-	t.Skip()
+
```
sandersn, line 13: the original test is weird -- we expect to have `A` as a completion for `A extends /**/` ? I guess it's the only class around, but it still doesn't make sense.

internal/ls/completions.go
```diff
@@ -400,8 +444,79 @@ func getCompletionData(program *compiler.Program, typeChecker *checker.Checker,
...
+						break
+					}
+				}
+				if noCommentPrefix {
+					return &completionDataJSDocTag{}
+				}
+			}
+		}
+
+		// Completion should work inside certain JsDoc tags. For example:
```
sandersn:
```suggestion
		// Completion should work inside certain JSDoc tags. For example:
```

internal/ls/completions.go
```diff
@@ -400,8 +444,79 @@ func getCompletionData(program *compiler.Program, typeChecker *checker.Checker,
...
+		//     /** @type {number | string} */
+		// Completion should work in the brackets
+		tag := getJSDocTagAtPosition(currentToken, position)
+		if tag != nil {
```
sandersn:
```suggestion
		if tag := getJSDocTagAtPosition(currentToken, position); tag != nil {
```

internal/ls/completions.go
```diff
@@ -400,8 +444,79 @@ func getCompletionData(program *compiler.Program, typeChecker *checker.Checker,
...
+			if tag.TagName().Pos() <= position && position <= tag.TagName().End() {
+				return &completionDataJSDocTagName{}
+			}
+			if ast.IsJSDocImportTag(tag) {
+				insideJsDocImportTag = true
+			} else {
+				typeExpression := tryGetTypeExpressionFromTag(tag)
+				if typeExpression != nil {
```
sandersn:
```suggestion
				if typeExpression := tryGetTypeExpressionFromTag(tag); typeExpression != nil {
```

internal/ls/completions.go
```diff
@@ -4679,14 +4794,24 @@ func (l *LanguageService) getCompletionItemDetails(
...
 	switch {
 	case symbolCompletion.request != nil:
-		request := symbolCompletion.request
-		// !!! JSDoc completions
-		if core.Some(request.keywordCompletions, func(c *lsproto.CompletionItem) bool {
-			return c.Label == itemData.Name
-		}) {
+		request := *symbolCompletion.request
+		switch request := request.(type) {
+		case *completionDataJSDocTagName:
```
sandersn, line 4797: can these three cases be collapsed? I don't know how type switches work.

internal/ls/completions.go
```diff
@@ -4860,3 +4985,625 @@ func createCompletionDetailsForSymbol(
...
+		paramName = getJSDocParamNameWithInitializer(paramName, initializer)
+	}
+	if isSnippet {
+		paramName = escapeSnippetText(paramName)
+	}
+	if isJS {
+		t := "*"
+		if isObject {
+			// !!! Debug.assert(!dotDotDotToken, `Cannot annotate a rest parameter with type 'Object'.`);
+			t = "Object"
```
sandersn: wait, what? `Object` is somehow the worst type for anything in either TS or JS. In Strada it means `any` if strict is off but in Corsa it always means `Object`. I think `object` makes more sense, or `*` or `any` if `object` is nonsensical somehow.

internal/ast/ast.go
sandersn: oh wait, tags already have a name `implements` or `extends` don't they?

sandersn (approved):
Minor suggestions and questions, but I think this is a faithful port of some slightly odd code.

internal/ls/completions.go
gabritto: Sadly, no. You can't collapse cases in type switches even if you don't rely on the type information inside it, I tried and the compiler complained.

internal/ls/completions.go
gabritto: Yeah, I can't really remember why I used `Object` here originally. Maybe that's what I more commonly saw in JS at the time? Anyways, I'll update to `object`.

internal/ls/completions.go
jakebailey: You can't, no, though you could make them conform to an interface and then switch on that.

internal/ast/utilities.go
gabritto: Oh, I got totally confused in my original answer, I thought Jake's comment was about a different function. But I agree we should rename it, I just don't know to what. In Strada this function even has a comment saying "True if node is of a kind that may contain comment text", so maybe this function was wrong in Strada too? I'm not sure. Maybe we should rename to `DoNotUseVeryBadIsJSDocCommentContainingNode` 😅

internal/ast/ast.go
gabritto: They have a `.TagName` already. I think it could be confusing if we decide to return `.ClassName` in `Name()`, since there are two possible candidates for a "name" here.

internal/ast/utilities.go
gabritto: I renamed it and also moved it to the astnav package, the only place where it's used. Couldn't think of a good name because I don't really know what the function is supposed to do.

internal/format/util.go
jakebailey: I filed https://github.com/microsoft/typescript-go/issues/1578

internal/ls/completions.go
sandersn: wait wait. I remembered. It's probably because it's preparing to emit nested object literal type notation:

```js
/** 
 * @param {Object} param0
 * @param {string} param0.a
 * @param {number} param0.b
 */
function f({ a, b }) {}
```

So it might make sense. 
It's also one more reason why `Object -> any` in Strada, even though I hate that and it's not coming back for Corsa.

jakebailey (approved):

internal/ls/completions.go
gabritto: As discussed offline, this works with `object` too:

https://www.typescriptlang.org/play/?ts=6.0.0-dev.20250814&filetype=js#code/PQKhAIChwgBAHAhgJ0QW3AbwPYCMBWApgMYAuAvuEqmgAzRzXpYDOpyAlgHYDmlTdAHSIG4BCmaYuAVzS5CyfhKG4GwSADNpXMh2xdwGgBSZwiADThc4ALzgAjOHIBKLNHAfi+ltgA2hQV9sHiNEZwBudw9kQlJpZANHAGorSPIgA

So keeping `object` until someone complains.

PR #1579, opened 2025-08-13
Set completion item's text edit if optional replacement span and insert text are present

### LINKED ISSUE #1568: Output from selecting optional property from auto complete modal incorrectly includes `?` in JS object
```
type T = {
  name?: string;
  age: number;
};

const obj: T = {};
```

When adding a value to `obj`, when typing the `name` key, the auto complete opens and includes `name?`, selecting this option will output `name?` in the object property identifier.

Image

```
const obj: T = {
  name?: "foo"
};
```

This does _not_ happen when using ^ space / `editor.action.triggerSuggest` and selecting the name option from the list.

Image

Image

gabritto:
Fixes #1568.

As per my comment on the issue, the problem happens because in LSP, if an `editRange` is present in the completion response (i.e. we set a default `editRange` to be used for all completion items), then an item's `insertText` is ignored. This is in line with the behavior when you specify `textEdit`s in individual items: `insertText` is also ignored in that case.
Ignoring `insertText` meant the client would use the item's label, e.g. `name?`, instead of the `insertText`'s `name`.

We may get rid of this problem in the future if instead of appending `?` in the item's label directly, we use the item's `labelDetails`.
But I want to fix the bug right away, and (1) not all clients support `labelDetails`, and (2) I'd like to have snippet completions implemented before figuring out the right `labelDetails` behavior, because snippet completions also use `labelDetails`.


internal/ls/completions_test.go
```diff
@@ -646,131 +646,6 @@ class bar5{ constructor(public /*constructorParameter1*/`,
 				},
 			},
 		},
-		{
```
gabritto, line 649: This test file will be deleted as part of the snapshopt PR, and we also have fourslash tests now, so I just deleted this test case instead of updating it.

jakebailey (approved):

PR #1582, opened 2025-08-14
Pool collators in LS

jakebailey:
Collators are not safe for concurrent use, so we create them fresh. But, pprof shows that `collate.New` is 15% of the time spent in the fourslash tests.

Stick them in a `sync.Pool` per language.

### DIFF
--- a/internal/ls/completions.go
+++ b/internal/ls/completions.go
@@ -24,6 +24,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 	"github.com/microsoft/typescript-go/internal/stringutil"
 	"golang.org/x/text/collate"
+	"golang.org/x/text/language"
 )
 
 func (l *LanguageService) ProvideCompletion(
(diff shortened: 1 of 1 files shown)


internal/ls/completions.go
```diff
@@ -2980,6 +2981,26 @@ func getCompletionsSymbolKind(kind ScriptElementKind) lsproto.CompletionItemKind
...
+			New: func() any {
+				return collate.New(tag)
+			},
+		})
+	}
+	return pool.Get().(*collate.Collator)
+}
+
+func putCollator(tag language.Tag, collator *collate.Collator) {
+	pool, _ := collatorCache.Load(tag)
```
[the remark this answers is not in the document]
jakebailey: Suggested code is wrong because it'll just waste loads and loads of allocs. If I wanted to solve this, I'd just return a `done` func from `get`.

internal/ls/completions.go
```diff
@@ -2980,6 +2981,25 @@ func getCompletionsSymbolKind(kind ScriptElementKind) lsproto.CompletionItemKind
 	}
 }
 
+var collatorCache collections.SyncMap[language.Tag, *sync.Pool]
+
+func getCollator(tag language.Tag) *collate.Collator {
+	pool, ok := collatorCache.Load(tag)
+	if !ok {
+		pool, _ = collatorCache.LoadOrStore(tag, &sync.Pool{
```
DanielRosenwasser, line 2989: Why is this a `LoadOrStore` instead of just a `Store`?

jakebailey: Otherwise two callers would update totally different maps

PR #1585, opened 2025-08-14
Use go.mod ignore directive instead of postinstall

jakebailey:
Before 1.25, we needed a postinstall to force Go to ignore `node_modules`. Now, we can just stick it in `go.mod`.

This also means we can now drop the `_` prefix on the dirs, but it's probably too late to change that.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,3 +28,12 @@ tool (
 	golang.org/x/tools/cmd/stringer
 	mvdan.cc/gofumpt
 )
+
+ignore (
+	node_modules
+	./_extension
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -28,3 +28,12 @@ tool (
 	golang.org/x/tools/cmd/stringer
 	mvdan.cc/gofumpt
 )
+
+ignore (
+	node_modules
```
jakebailey: When specified without `./`, this means "ignore `node_modules` at any level", which I think is correct in general.

PR #1589, opened 2025-08-15
Updatess to tsc tests

sheetalkamat:
These are first few commits from #1484 that only change tests framework

### DIFF
--- a/cmd/tsgo/sys.go
+++ b/cmd/tsgo/sys.go
@@ -46,11 +46,6 @@ func (s *osSys) Writer() io.Writer {
 	return s.writer
 }
 
-func (s *osSys) EndWrite() {
-	// do nothing, this is needed in the interface for testing
-	// todo: revisit if improving tsc/build/watch unittest baselines
-}
-
 func (s *osSys) WriteOutputIsTTY() bool {
 	return term.IsTerminal(int(os.Stdout.Fd()))
 }
--- a/internal/vfs/iovfs/iofs.go
+++ b/internal/vfs/iovfs/iofs.go
@@ -24,6 +24,11 @@ type WritableFS interface {
 	Remove(path string) error
 }
 
+type FsWithSys interface {
+	vfs.FS
+	FSys() fs.FS
+}
+
 // From creates a new FS from an [fs.FS].
 //
 // For paths like `c:/foo/bar`, fsys will be used as though it's rooted at `/` and the path is `/c:/foo/bar`.
@@ -33,7 +38,7 @@ type WritableFS interface {
 //
 // From does not actually handle case-insensitivity; ensure the passed in [fs.FS]
 // respects case-insensitive file names if needed. Consider using [vfstest.FromMap] for testing.
-func From(fsys fs.FS, useCaseSensitiveFileNames bool) vfs.FS {
+func From(fsys fs.FS, useCaseSensitiveFileNames bool) FsWithSys {
 	var realpath func(path string) (string, error)
 	if fsys, ok := fsys.(RealpathFS); ok {
 		realpath = func(path string) (string, error) {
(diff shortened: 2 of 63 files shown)


cmd/tsgo/sys.go
```diff
@@ -46,11 +46,6 @@ func (s *osSys) Writer() io.Writer {
 	return s.writer
 }
 
-func (s *osSys) EndWrite() {
```
jakebailey, line 49: Very happy to see `EndWrite` go

internal/vfs/iovfs/iofs.go
```diff
@@ -24,6 +24,11 @@ type WritableFS interface {
 	Remove(path string) error
 }
 
+type FsWithSys interface {
+	vfs.FS
+	FSys() fs.FS
```
jakebailey, line 29: I think I probably would have called this `Underlying()` but `FSys` is fine.

internal/vfs/vfstest/vfstest.go
```diff
@@ -482,20 +523,54 @@ func (m *mapFS) WriteFile(path string, data []byte, perm fs.FileMode) error {
...
+			return "/" + string(fileInfo.Data), true
+		}
+	}
+	return "", false
+}
+
+func (m *MapFS) Entries() iter.Seq2[string, *fstest.MapFile] {
+	return func(yield func(string, *fstest.MapFile) bool) {
+		m.mu.RLock()
+		defer m.mu.RUnlock()
```
jakebailey, line 557: I'm not sure if it much matters, but this will effectively mean that the whole MapFS is locked until iteration completes.

internal/vfs/vfstest/vfstest.go
```diff
@@ -17,7 +18,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/vfs/iovfs"
 )
 
-type mapFS struct {
+type MapFS struct {
```
jakebailey, line 21: If `MapFS` is exported, you could probably just make `FromMap` and `FromMapWithClock` export `*MapFS` directly. Optional, though.

internal/vfs/vfstest/vfstest.go
sheetalkamat: which is fine as we anyways arent operating on it.. we are just using fileInfo and baselining it

internal/vfs/vfstest/vfstest.go
sheetalkamat: It cant be though right. it returns `ioFS` which has MapFS

internal/vfs/iovfs/iofs.go
jakebailey: Eh, it's not really "wrapping" per se, it's _adapting_ one FS to the other. Ignore this 😄

cmd/tsgo/sys.go
sheetalkamat: Yeah that one revealed bug when writing diagnostics that dont have locations

internal/vfs/vfstest/vfstest.go
jakebailey: Ah, true. Mainly I was looking for a way that didn't involve basically this chain:

```go
s.fs.FS.(*testFs).FS.(iovfs.FsWithSys).FSys().(*vfstest.MapFS)
```

Which seemed pretty cast-y.

sheetalkamat: unfortunately the way its wrapped it becomes like that.

jakebailey: Yeah. At least it's all concreteish, it's just this one wrapping.

jakebailey (approved):

PR #1590, opened 2025-08-15
Fix during porting tsc --composite tests

sheetalkamat:


sheetalkamat:
This was cherry pick from #1484 but given there are more tests that were fixed as part of tsc -b work, closing this

PR #1593, opened 2025-08-15
Port strada `debug` namespace, enable many debug assertions

weswigham:
Almost every commented-out assertion in the codebase enables cleanly, except for three - two don't seem to apply anymore due to code structure changes and I've deleted them here, the third I've added a comment to and will enable as a follow-up alongside the fix to make it not immediately be triggered, since that'll have some baseline diffs accompanying it.

The _conditional_ `debug` namespace members are all empty (or pass-thru where appropriate) functions when compiled with the `release` tag. That should make the compiler elide them and their arguments as dead code in that mode, which should make conditional assertions free in terms of shipping-release-performance. _Unconditional_ members (`fail`, `failBadSyntaxKind`, `assertNever`) should be in-place both with and without the `release` tag.

Additionally, `hereby` now applies the `release` tag by default to `build` tasks. Tests aughta still get compiled without it, so should have full assertions. And if anyone builds from source or via the normal `go build ./...` command, they should also get assertions enabled by default. (The `noembed` tag we were already using already works like this, so this isn't exactly a new split).

There are a bunch of `panic`s that should probably get reverted back into `assert` and `fail` calls at some point so they have a unified debug breakpoint inside `Fail`, but that can be a future task.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -189,7 +189,7 @@ export const lib = task({
 function buildTsgo(opts) {
     opts ||= {};
     const out = opts.out ?? "./built/local/";
-    return $({ cancelSignal: opts.abortSignal, env: opts.env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${goBuildTags("noembed", "release")} -o ${out} ./cmd/tsgo`;
+    return $({ cancelSignal: opts.abortSignal, env: opts.env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${options.debug ? goBuildTags("noembed") : goBuildTags("noembed", "release")} -o ${out} ./cmd/tsgo`;
 }
 
 export const tsgoBuild = task({
--- a/internal/debug/debug.go
+++ b/internal/debug/debug.go
@@ -117,7 +117,7 @@ func AssertIsDefined(value any, message ...string) {
 	}
 }
 
-func CheckDefined(value any, message ...string) any {
+func CheckDefined[T any](value T, message ...string) T {
 	AssertIsDefined(value, message...)
 	return value
 }
--- a/internal/debug/shared.go
+++ b/internal/debug/shared.go
@@ -1,6 +1,9 @@
 package debug
 
-import "fmt"
+import (
+	"fmt"
+	"runtime"
+)
 
 func Fail(reason string) {
 	if len(reason) == 0 {
@@ -8,6 +11,7 @@ func Fail(reason string) {
 	} else {
 		reason = "Debug failure. " + reason
 	}
+	runtime.Breakpoint()
 	panic(reason)
 }
 
@@ -34,7 +38,7 @@ func AssertNever(member any, message ...string) {
 	} else if member, ok := member.(fmt.Stringer); ok {
 		detail = member.String()
 	} else {
-		detail = "<unformattable>"
+		detail = fmt.Sprintf("%v", member)
 	}
 	Fail(fmt.Sprintf("%s %s", msg, detail))
 }
--- a/internal/debug/release.go
+++ b/internal/debug/release.go
@@ -14,7 +14,7 @@ func AssertLessThanOrEqual(a int, b int, message ...string)                     
 func AssertGreaterThan(a int, b int, message ...string)                                       {}
 func AssertGreaterThanOrEqual(a int, b int, message ...string)                                {}
 func AssertIsDefined(value any, message ...string)                                            {}
-func CheckDefined(value any, message ...string) any                                           { return value }
+func CheckDefined[T any](value T, message ...string) T                                        { return value }
 func AssertEachIsDefined[TElem any](value []TElem, message ...string)                         {}
 func CheckEachIsDefined[TElem any](value []TElem, message ...string) []TElem                  { return value }
 func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, message ...string)  {}


internal/debug/debug.go
```diff
@@ -0,0 +1,198 @@
...
+	if test == nil {
+		return
+	}
+	if len(message) == 0 {
+		message = unexpectedNode
+	}
+	Assert(test(node), message...)
+}
+
+func AssertOptionalToken[TElem interface{ KindValue() int16 }](node TElem, kind int16, message ...string) {
```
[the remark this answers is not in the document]
weswigham: ...No, because that would prevent us from using `debug` helpers inside `ast`. `debug` stands alone - which means no wrapper types allowed.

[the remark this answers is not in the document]
jakebailey: You could actually do `KindValue() ~int16` (might need an extra type param) and skip the conversion in the `ast` package, but that's not really much different.

internal/transformers/declarations/transform.go
```diff
@@ -1525,8 +1526,6 @@ func (tx *DeclarationTransformer) ensureTypeParams(node *ast.Node, params *ast.T
 				Nodes: nodes,
 			}
 		}
```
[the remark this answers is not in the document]
weswigham: That's a question for @sandersn - near as I can tell, this branch was added alongside the `.FullSignature` support, and when enabled, this assertion is pretty immediately triggered in tests. I can't logically see what'd make it unreachable (it _is_ reachable in reality), so away it goes.

internal/debug/debug.go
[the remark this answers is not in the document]
weswigham: I'd still need to keep the `KindValue` method to access the field at all, and nothing else implements the method, so it's not like it changes much - may as well use the concrete type 🤷‍♂️

internal/ls/signaturehelp.go
```diff
@@ -617,7 +620,7 @@ func getContainingArgumentInfo(node *ast.Node, sourceFile *ast.SourceFile, check
 	for n := node; !ast.IsSourceFile(n) && (isManuallyInvoked || !ast.IsBlock(n)); n = n.Parent {
 		// If the node is not a subspan of its parent, this is a big problem.
 		// There have been crashes that might be caused by this violation.
-		// Debug.assert(rangeContainsRange(n.parent, n), "Not a subspan", () => `Child: ${Debug.formatSyntaxKind(n.kind)}, parent: ${Debug.formatSyntaxKind(n.parent.kind)}`);
+		debug.Assert(RangeContainsRange(n.Parent.Loc, n.Loc), fmt.Sprintf("Not a subspan. Child: %s, parent: %s", n.KindString(), n.Parent.KindString()))
```
jakebailey, line 623: Even with the build tag a lot of this will end up being executed eagerly; theoretically you could accept a func for these still and then they won't be called in the inlined empty functions so should be eliminated. Not clear how much that matters in this version of debugging though where we only get these in tests.

internal/debug/shared.go
```diff
@@ -0,0 +1,40 @@
+package debug
+
+import "fmt"
+
+func Fail(reason string) {
+	if len(reason) == 0 {
```
jakebailey, line 9: I don't know if this actually matters with delve but just like `Debug.fail` had `debugger;` you could do `runtime.Breakpoint()`.

internal/debug/debug.go
```diff
@@ -0,0 +1,198 @@
...
+		if len(message) == 0 {
+			msg = ""
+		} else {
+			msg = message[0]
+		}
+		Fail(msg)
+	}
+}
+
+func CheckDefined(value any, message ...string) any {
```
jakebailey:
```suggestion
func CheckDefined[T any](value T, message ...string) T {
```

internal/ls/signaturehelp.go
weswigham: Eh, I don't know why this one was a deferred message to begin with - it was just stringifying two `kind`s, nothing actually expensive. And it's the only assertion (thus far) using the deferred message details pattern, so I just eager'd it rather than adding a deferred-message-taking-overload. In an ideal world under a `release` tag, the whole `fmt.Sprintf` *should* get elided since its' result is unused, anyway. (Weather it actually does, I haven't checked in godbolt, but from what I've read on assertion libs in `go`, that's the expected codegen for something like this.)

jakebailey: It'll still get executed, as far as I know. If this were a pure string concat, that would probably get elided.

Herebyfile.mjs
```diff
@@ -189,7 +189,7 @@ export const lib = task({
 function buildTsgo(opts) {
     opts ||= {};
     const out = opts.out ?? "./built/local/";
-    return $({ cancelSignal: opts.abortSignal, env: opts.env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${goBuildTags("noembed")} -o ${out} ./cmd/tsgo`;
+    return $({ cancelSignal: opts.abortSignal, env: opts.env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${goBuildTags("noembed", "release")} -o ${out} ./cmd/tsgo`;
```
jakebailey: I think we should make this conditional on `--debug`, just so we can get the asserts when debugging the editor / if we want a debug build for local testing. `--debug` right now means "turn off optimization", so it's an odd combo, but I think not unreasonable.

internal/debug/shared.go
```diff
@@ -0,0 +1,40 @@
...
+	} else {
+		msg = message[0]
+	}
+	var detail string
+	if member, ok := member.(interface{ KindString() string }); ok {
+		detail = member.KindString()
+	} else if member, ok := member.(fmt.Stringer); ok {
+		detail = member.String()
+	} else {
+		detail = "<unformattable>"
```
jakebailey: Could just do `%v` or `%#v` and let anything in.

internal/fourslash/baselineutil.go
```diff
@@ -181,7 +182,8 @@ func (f *FourslashTest) getBaselineContentForFile(
 		if options.startMarkerPrefix != nil {
 			if startPrefix := options.startMarkerPrefix(span.Location); startPrefix != "" {
 				if fileName == options.marker.FileName() && span.Range.Start == options.marker.LSPosition {
-					// ts.Debug.assert(!detailPrefixes.has(details[0]), "Expected only single prefix at marker location");
+					_, ok := detailPrefixes[*details[0]]
+					debug.Assert(!ok, "Expected only single prefix at marker location")
```
jakebailey, line 186: I suspect the ones in this file should actually be `t.Error`; it used to be that in Strada we had a _couple_ debug asserts that should have been `chai` checks but were not.

weswigham: There's no `t` in scope here - it's deep in the test setup logic. It's awkward, at least.

jakebailey: Interesting, most `FourslashTest` take `t` as their first arg, but I guess we don't have that here.

I'm surprised we don't have it as a plain field at this point. Only one is ever allocated per test. Doing so would simplify all of the methods greatly, so long as we never call a failing T method from the wrong goroutine.

weswigham:
The hope, IMO, is that when you test something you get a repro that fails faster, closer to the root issue - we could also have assertions left on in nightly builds if we wanted, so there'd be a public artifact with assertions on.

PR #1598, opened 2025-08-18
Fix typo in isRequiredInitializedParameter

weswigham:
Followup to #1593 enabling another assertion, which points out a missing `!` in `isRequiredInitializedParameter`.


testdata/baselines/reference/submodule/compiler/declarationEmitCastReusesTypeNode1(strictnullchecks=true).js
```diff
@@ -80,4 +77,6 @@ declare const _default: {
 };
 export default _default;
 // allows `undefined` on the input side, thanks to the initializer
-export declare function fnWithPartialAnnotationOnDefaultparam(x: P, b: number): void;
+export declare function fnWithPartialAnnotationOnDefaultparam(x: {
+    name: string;
+} | undefined, b: number): void;
```
jakebailey, line 82: Interesting that we are losing `P` here; I assume this is a reuse problem?

weswigham: Yep - we just swapped the diff caused by lacking node reuse logic from the `strict: false` case to `strict: true`.

PR #1599, opened 2025-08-19
Const enum inlining and synthetic comment emit support

### LINKED ISSUE #1216: Unable to use `const enum` in third-party API
This is a followup to https://github.com/microsoft/typescript-go/issues/829

The library `@homebridge/ciao` has this definition:
https://github.com/homebridge/ciao/blob/11e5abcd8656ce627517275013cc44bbb1bf7769/src/index.ts#L37-L40
```ts
export const enum Protocol {
  TCP = "tcp",
  UDP = "udp",
}
```

Its API expects this enum to be used for some option parameters, see https://github.com/homebridge/ciao/blob/11e5abcd8656ce627517275013cc44bbb1bf7769/src/CiaoService.ts#L78

This used to work with TS 5.x, but `tsgo` no longer accepts the existing code. I understand the intention is to not support `const enum`s cross module, but the current behavior is problematic.

If I try to use the string directly, tsgo complains about a mismatch of types:
```
src/lib/server.ts:543:9 - error TS2322: Type '"tcp"' is not assignable to type 'Protocol | undefined'.

543         protocol: "tcp",
            ~~~~~~~~
```
**--> I would expect this to work**

If I use the imported `const enum` with `isolatedModules`, tsgo complains about the `const enum`:
```
src/lib/server.ts:543:19 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

543         protocol: Protocol.TCP,
                      ~~~~~~~~

Found 1 error in src/lib/server.ts:543
```
**--> This seems fine to me**

And if I use the imported `const enum` without `isolatedModules`, I end up with an undefined reference in the compiled JS output, without any warning that this might be a problem:

Image

**--> I would expect this to be a compile-time error**

weswigham:
The synthetic comment support should unblock a handful of TODOs in the node builder for services logic. There might be some helpers missing that are used elsewhere, but all the core printer support is in, since that's most of our const enum emit logic that isn't checker-driven.

Despite appearances, this works differently from how we did const enum inlining in strada - there, we enabled substitutions in the `ts` transform and did const enum inlining late during the substitution pass (a super late pass invoked by the printer just before printing the node), which is also how transforms used to handle renaming exports and a few other things. Like other substitution uses, that's had to change. Now it's a transform executed at the end of the very end of the js transform pipeline instead, so you'll actually see the changed nodes in the AST. Unfortunately, since it's still checker-based, only nodes derived from a checked node get inlined. (So, for example, if a transform manifests an `A.B` from nothing into the result tree, and `A.B` refers to a `const enum`, the inliner transform doesn't know to inline it since it only inlines nodes related to parsed nodes - in practice, this doesn't seem to be a problem, since transforms don't generally manufacture new references from nothing to const enums - only to globals and helpers and such.)

Fixes #1216


jakebailey:
Does this fix #1216?

internal/checker/checker.go
```diff
@@ -15087,7 +15087,7 @@ func (c *Checker) resolveEntityName(name *ast.Node, meaning ast.SymbolFlags, ign
 				name.Parent != nil && name.Parent.Kind == ast.KindJSExportAssignment) {
 			c.markSymbolOfAliasDeclarationIfTypeOnly(getAliasDeclarationFromName(name), symbol, nil /*finalTarget*/, true /*overwriteEmpty*/, nil, "")
 		}
-		if symbol.Flags&meaning == 0 && !dontResolveAlias {
+		if symbol.Flags&meaning == 0 && !dontResolveAlias && symbol.Flags&ast.SymbolFlagsAlias != 0 {
```
weswigham, line 15090: In strada, `checkExpressionCached` on `a.b.c.d` caches resolved symbols on `a.b.c` and `a.b` and `a`. `getConstantValue` uses these cached symbols to shortcut its' logic. In corsa, it does not, and, in so doing, reveals that here we try to call `resolveAlias` on non-alias (module) symbols when we consider if we want to inline `module.exports`. This fixes that.

internal/compiler/emitter.go
```diff
@@ -128,6 +129,11 @@ func getScriptTransformers(emitContext *printer.EmitContext, host printer.EmitHo
 
 	// transform module syntax
 	tx = append(tx, getModuleTransformer(&opts))
+
+	// inlining (formerly done via substitutions)
+	if !options.IsolatedModules.IsTrue() {
+		tx = append(tx, inliners.NewConstEnumInliningTransformer(&opts))
+	}
```
jakebailey: Is not inlining when isolatedModules is enabled a new behavior in Corsa?

internal/compiler/emitter.go
```diff
@@ -128,6 +129,11 @@ func getScriptTransformers(emitContext *printer.EmitContext, host printer.EmitHo
 
 	// transform module syntax
 	tx = append(tx, getModuleTransformer(&opts))
+
+	// inlining (formerly done via substitutions)
+	if !options.IsolatedModules.IsTrue() {
```
jakebailey:
```suggestion
	if !options.GetIsolatedModules() {
```

I think?

jakebailey: (and above?)

internal/transformers/inliners/constenum.go
jakebailey, line 1: Are there more inliners than just constenum?

internal/transformers/inliners/constenum.go
```diff
@@ -0,0 +1,88 @@
...
+	compilerOptions := opt.CompilerOptions
+	emitContext := opt.Context
+	if compilerOptions.IsolatedModules.IsTrue() {
+		debug.Fail("const enums are not inlined under isolated modules")
+	}
+	tx := &ConstEnumInliningTransformer{compilerOptions: compilerOptions, emitResolver: opt.EmitResolver}
+	return tx.NewTransformer(tx.visit, emitContext)
+}
+
+func (tx *ConstEnumInliningTransformer) visit(node *ast.Node) *ast.Node {
```
jakebailey, line 32: Wow, I love how short this is

testdata/baselines/reference/submodule/compiler/blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js.diff
```diff
@@ -0,0 +1,19 @@
+--- old.blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js
++++ new.blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js
+@@= skipped -28, +28 lines =@@
+     })(E || (E = {}));
+ }
+ function foo2() {
+-    return E.A;
++    return 0 /* E.A */;
```
jakebailey: VMS is on here, so this would be fixed by the `GetIsolatedModules` thing I believe.

testdata/baselines/reference/submodule/compiler/blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js.diff
```diff
@@ -0,0 +1,19 @@
...
+     let E;
+     (function (E) {
+         E[E["A"] = 0] = "A";
+     })(E || (E = {}));
+ }
+ const config = {
+-    a: AfterObject.A,
++    a: 2 /* AfterObject.A */,
```
weswigham: I'm about to push the fix for this one (should be using the calculated options `isolatedModules` flag and not the raw one), but this test seems to raise an important issue in VMS, imo - `const enum`s really just don't work under it under current rules, do they? Even within the same function...

jakebailey: The emit is bad, but IIRC this emits a checker error here anyway, so it's fair game.

testdata/baselines/reference/submodule/conformance/jsxCheckJsxNoTypeArgumentsAllowed.js.diff
```diff
@@ -10,5 +10,5 @@
 -</>;
 +const component_1 = require("./component");
 +const React = require("react");
-+let x = (<component_1.MyComp />, <component_1.Prop> a={10} b="hi" />; // error, no type arguments in js
++let x = (<component_1.MyComp />, <Prop> a={10} b="hi" />; // error, no type arguments in js
```
weswigham, line 13: This is the only diff that isn't a strict improvement to match strada, imo - seems like swapping from the binder-based import-only resolver to the full emit resolver has changed the behavior of the cjs transform on the `Prop` identifier here. Now, this is all syntactic gobbeldeygook because it's a `jsx` file with a jsx tag with a type argument, being _generously_ reinterpreted as a comma expression, so the emit here doesn't actually matter, but might indicate a more important bug somewhere in the checker. It'd be trivial to swap the behavior back (without troubleshooting the root cause) by having both the import-only resolver and the checker-resolver on hand and using the import-only one for the cjs transform, though.

testdata/baselines/reference/submodule/compiler/constEnums.js.diff
```diff
@@ -37,9 +37,9 @@
 +    // correct cases: reference to the enum member from different enum declaration
 +    Enum1["W1"] = A0;
 +    if (typeof Enum1.W1 !== "string") Enum1[Enum1.W1] = "W1";
-+    Enum1["W2"] = Enum1.A0;
++    Enum1["W2"] = 100 /* Enum1.A0 */;
```
jakebailey, line 40: Not related to your PR, but this is an interesting case, since I assume this is due to us not doing enum merging properly or something

internal/compiler/emitter.go
weswigham: No - I just lifted the check from inside the `ts` transform out to transform composition time.

internal/transformers/inliners/constenum.go
weswigham: As of right now, no, but we've toyed with adding minifer-like emit passes in the past, so it's a logical place to group them.

testdata/baselines/reference/submodule/conformance/jsxCheckJsxNoTypeArgumentsAllowed.js.diff
jakebailey, line 13: This seems to be regressing, though I am not sure why. (I think we recently got a PR that fixed this?)

internal/transformers/inliners/constenum.go
weswigham: Yeah, I inlined some indirections from strada.

internal/compiler/emitter.go
jakebailey: TIL.

weswigham:
And yea, this is logically the fix for #1216 since it reenables const enum inlining. I'll edit the OP to close it on merge.

testdata/baselines/reference/submodule/conformance/jsxCheckJsxNoTypeArgumentsAllowed.js.diff
weswigham: Yeah, see https://github.com/microsoft/typescript-go/pull/1599#discussion_r2283783227 - the input is syntactic garbage here, though, so I'm not too inclined to track down the difference between the binder-based resolver and the checker-based resolver just right now - none of the tests with valid input syntax were affected.

weswigham:
Looks like my one-liner change to `resolveEntityName` also fixed a bunch of fourslash tests - that's nice.

weswigham:
Or not? `npm run updatefailing` unskipped a bunch of tests like they pass now, but they all still crash...? What gives?

jakebailey:
It looks like this panic is unrecovered, so the test suite crashed, leading all tests after it to appear passing.

```
SIGTRAP: trace trap
PC=0x486381 m=5 sigcode=128

goroutine 20581 gp=0xc0005156c0 m=5 mp=0xc0006bc008 [running]:
runtime.breakpoint()
	/opt/hostedtoolcache/go/1.25.0/x64/src/runtime/asm_amd64.s:388 +0x1 fp=0xc000c21a68 sp=0xc000c21a60 pc=0x486381
runtime.Breakpoint(...)
	/opt/hostedtoolcache/go/1.25.0/x64/src/runtime/proc.go:5431
github.com/microsoft/typescript-go/internal/debug.Fail(...)
	/home/runner/work/typescript-go/typescript-go/internal/debug/shared.go:14
github.com/microsoft/typescript-go/internal/checker.(*Checker).resolveAlias(0xc000963420?, 0x217ff68?)
	/home/runner/work/typescript-go/typescript-go/internal/checker/checker.go:15501 +0x36e fp=0xc000c21ad8 sp=0xc000c21a68 pc=0x115d20e
github.com/microsoft/typescript-go/internal/checker.(*Checker).ResolveAlias(...)
	/home/runner/work/typescript-go/typescript-go/internal/checker/checker.go:15495
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getReferencedSymbolsForModuleIfDeclaredBySourceFile(0xc001c3e2e8, {0x217ff68, 0xc000ae3540}, 0xc000235c00?, 0xc001905a40, {0xc00029d720, 0x9, 0x9}, {0x0, 0x0, ...}, ...)
	/home/runner/work/typescript-go/typescript-go/internal/ls/findallreferences.go:631 +0x3c6 fp=0xc000c21bb0 sp=0xc000c21ad8 pc=0x1494aa6
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getReferencedSymbolsForNode(0xc001c3e2e8, {0x217ff68, 0xc000ae3540}, 0x13?, 0xc00250ac60, 0xc001905a40, {0xc00029d720, 0x9, 0x9}, {0x0, ...}, ...)
	/home/runner/work/typescript-go/typescript-go/internal/ls/findallreferences.go:599 +0x899 fp=0xc000c21d88 sp=0xc000c21bb0 pc=0x14940f9
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideReferences(0xc001c3e2e8, {0x217ff68, 0xc000ae3540}, 0xc00133ced0)
	/home/runner/work/typescript-go/typescript-go/internal/ls/findallreferences.go:409 +0x195 fp=0xc000c21e60 sp=0xc000c21d88 pc=0x14921b5
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleReferences(0xbfa970495a0e337?, {0x217ff68, 0xc000ae3540}, 0xc00133ced0)
	/home/runner/work/typescript-go/typescript-go/internal/lsp/server.go:740 +0xa5 fp=0xc000c21eb8 sp=0xc000c21e60 pc=0x153b925
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerRequestHandler[...].15({0x217ff68, 0xc000ae3540}, 0xc00133cf00)
	/home/runner/work/typescript-go/typescript-go/internal/lsp/server.go:535 +0x86 fp=0xc000c21ef8 sp=0xc000c21eb8 pc=0x1535346
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0xc001d963c0, {0x217ff68, 0xc000ae3540}, 0xc00133cf00)
	/home/runner/work/typescript-go/typescript-go/internal/lsp/server.go:471 +0x18c fp=0xc000c21f50 sp=0xc000c21ef8 pc=0x15399cc
	/home/runner/work/typescript-go/typescript-go/internal/lsp/server.go:376 +0x89 fp=0xc000c21fe0 sp=0xc000c21f50 pc=0x15385a9
runtime.goexit({})
	/opt/hostedtoolcache/go/1.25.0/x64/src/runtime/asm_amd64.s:1693 +0x1 fp=0xc000c21fe8 sp=0xc000c21fe0 pc=0x488301
created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 20499
	/home/runner/work/typescript-go/typescript-go/internal/lsp/server.go:396 +0x648
```

jakebailey:
Wait, does Breakpoint fall over if no debugger is attached? Do we need to remove that line?

weswigham:
Looks like it. I'll just remove the `Debug.fail` call from this PR to unblock it, since that was only a driveby swap to debug a test anyway.

jakebailey:
Sent #1601 too.

internal/transformers/inliners/constenum.go
jakebailey: I don't think that's going to happen, but 😄

internal/printer/emitcontext.go
```diff
@@ -917,3 +927,37 @@ func (c *EmitContext) VisitIterationBody(body *ast.Statement, visitor *ast.NodeV
 
 	return updated
 }
+
+func (c *EmitContext) SetSyntheticLeadingComments(node *ast.Node, comments []SynthesizedComment) *ast.Node {
```
jakebailey, line 931: Not this PR, but there are some callers of this I left commented out in the node builder (probably more than that though, just opining)

jakebailey (approved):

internal/printer/emitcontext.go
jakebailey: Sent #1607 before this task slipped out of my brain.

DoctorGester:
And I just got rid of const enums in our code to switch to tsgo because I thought they are not coming back 😭

PR #1601, opened 2025-08-19
Remove runtime.Breakpoint call

jakebailey:
I guess this is _not_ similar enough to `debugger;`, doing it without a debugger attach just does SIGTRAP. ☹️


jakebailey:
At best we could set a build tag that would apply when starting a launch task...

PR #1607, opened 2025-08-19
Add missing calls to AddSyntheticLeadingComment in node builder

jakebailey:
#1599 added these helpers; call them where we previously had them commented out.

I've also removed some commented out code that would have previously used the synthetic comment system, but are not longer needed due to our new JSDoc rewrite.


testdata/baselines/reference/submodule/compiler/emitClassExpressionInDeclarationFile.js.diff
```diff
@@ -17,16 +17,9 @@
...
-+        tags(c: any): any;
-     };
+@@= skipped -38, +38 lines =@@
      getTags(c: {
--        tags(c: /*elided*/ any): /*elided*/ any;
-+        tags(c: any): any;
+         tags(c: /*elided*/ any): /*elided*/ any;
      }): {
 -        tags(c: /*elided*/ any): /*elided*/ any;
 +        tags(c: any): any;
```
jakebailey, line 25: I am unsure why this is not also fixed.

jakebailey: Ah, I think I found it.

jakebailey: No, I didn't, I just switched back to `main` and forgo I did so and read code I already fixed in this PR 🙃

weswigham (approved):
Honestly, kinda surprised we even have this many tests that capture the elision text.

PR #1610, opened 2025-08-19
Always set synthetic position when deep cloning

gabritto:
Fixes an out of bounds crash found when compiling an internal project.

We weren't setting (-1, -1) as a position for non-leaf nodes when deep cloning them. As a result, we'd later use those positions to access a file's text that wasn't the file for the cloned node, so the positions were out of the file text's bounds.

The diff for some compiler tests increased because we now seem to omit a comma during emit, but I think this is a problem that can be addressed later.


weswigham (approved):
Do we need to copy the `hasTrailingComma` node list member?

gabritto:
> Do we need to copy the `hasTrailingComma` node list member?

Uh, right, that's what's causing the diffs. That's a method now in Corsa, and it skips synthetic nodes and checks positions to determine the result. We could set up positions when cloning to preserve this (maybe set `end` to -2 on the last node of the list if it had a trailing comma?), I think. I'll try doing that and see if it works.

weswigham (approved):

PR #1614, opened 2025-08-20
Don't error on referenced project's noEmit if program has empty files

gabritto:
Fixes what seems like a porting bug in checking config references. If a tsconfig has no files (i.e. it's a top-level tsconfig that refers every project entry point in your monorepo), it's ok for a project it references to have `noEmit` true (same for having `composite` false).

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -862,7 +862,7 @@ func (p *Program) verifyProjectReferences() {
 		}
 		refOptions := config.CompilerOptions()
 		if !refOptions.Composite.IsTrue() || refOptions.NoEmit.IsTrue() {
-			if len(config.FileNames()) > 0 {
+			if len(parent.FileNames()) > 0 {
 				if !refOptions.Composite.IsTrue() {
 					createDiagnosticForReference(parent, index, diagnostics.Referenced_project_0_must_have_setting_composite_Colon_true, ref.Path)
 				}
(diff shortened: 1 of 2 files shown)


internal/compiler/program.go
```diff
@@ -862,7 +862,7 @@ func (p *Program) verifyProjectReferences() {
 		}
 		refOptions := config.CompilerOptions()
 		if !refOptions.Composite.IsTrue() || refOptions.NoEmit.IsTrue() {
-			if len(config.FileNames()) > 0 {
+			if len(parent.FileNames()) > 0 {
```
jakebailey, line 865: The old code said:

```
const inputs = parent ? parent.commandLine.fileNames : rootNames;
if (inputs.length) {
```

I assume `parent` simply always exists?

gabritto: Seems to be the case in Corsa: this function starts with `ref := parent.ProjectReferences()[index]`, which would crash if `parent` is `nil`.

jakebailey (approved):

internal/compiler/program.go
sheetalkamat: thanks. yes with being able to use and store parsedCommandLine in program, we always have parent 
I had this in #1484  fixed incorrectly too (was always checking program's file names) so thank you for doing the right fix.

PR #1623, opened 2025-08-22
Release program after finishing program emit, Only cache dts and json source files

### LINKED ISSUE #1622: [bug] `tsgo --build` uses extreme amounts of memory
## Stack trace

Unfortunately I don't have one to share as it crashed my remote dev machine (used 116GB of RSS before it did).

Image

## Steps to reproduce

This is probably unique to our scale here at Canva. I attempted to build one of the TS projects in our monorepo that transitively depends upon 5k+ other TS projects. It built for ~5 mins before crashing out due to excess memory usage.

```
./built/local/tsgo --build ../canva/<omitted>/tsconfig.json
```

sheetalkamat:
Parity with Strada:
Release program after emit - (this releases program all together and snapshot for now but it will be retained for --w like in strada)
Cache only source files for d.ts and json files as those are the ones that are exepected to be reused

Fixes #1622

### DIFF
--- a/internal/execute/build/buildtask.go
+++ b/internal/execute/build/buildtask.go
@@ -70,8 +70,10 @@ func (t *buildTask) report(orchestrator *Orchestrator, configPath tspath.Path, b
 		buildResult.programStats = append(buildResult.programStats, t.statistics)
 	}
 	if t.program != nil {
+		if orchestrator.opts.Testing != nil {
+			orchestrator.opts.Testing.OnProgram(t.program)
+		}
 		t.program.MakeReadonly()
-		buildResult.result.IncrementalProgram = append(buildResult.result.IncrementalProgram, t.program)
 		buildResult.statistics.ProjectsBuilt++
 	}
 	if t.pseudoBuild {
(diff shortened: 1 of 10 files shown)


jakebailey (approved):

jakebailey:
> Cache only source files for d.ts and json files as those are the ones that are exepected to be reused

I'll be curious if in the future we can still cache input ts files, given people can use build mode to parse out the same set of files multiple times, e.g. for tests or for multi-emit-formats. But, from your wording, it sounds like you're saying that Strada did not do this?

sheetalkamat:
> > Cache only source files for d.ts and json files as those are the ones that are exepected to be reused
> 
> I'll be curious if in the future we can still cache input ts files, given people can use build mode to parse out the same set of files multiple times, e.g. for tests or for multi-emit-formats. But, from your wording, it sounds like you're saying that Strada did not do this?



yeah - with project references - we will reference output file - which is d.ts file for the common projects so its very rare to have `.ts` files hence we didnt cache them - also that increases foot print in memory which bigger problem in Strada so we have has some optimizations there.

Esp imaging programs with 60k+ files, it quickly adds up

PR #1626, opened 2025-08-22
Fix incorrect passing of compile times

sheetalkamat:


andrewbranch (approved):

sheetalkamat:
> These aren't modified concurrently by the build system, right?

No, they are generated per project and then get get aggregated at the end of the build for reporting aggregate statistics

PR #1627, opened 2025-08-22
[WIP] fix go to definition

rgoomar:
These changes made it so I could go to a proper definition. This isn't the cleanest approach although it gives an idea of what can work. 

Basically we have a situation where we have an import in a `.ts` file like
```
import { getLogger } from '@latticehr/o11y';
```
in a directory like `apps/app-one` in an `index.ts` file and the `@latticehr` is an alias that should reference `packages/o11y` and it does go to the `packages/o11y/dist/*.d.ts` file although not the source file implementation 

I'd love to get some feedback and modify this to fit standards that you all have.

### DIFF
--- a/internal/ls/definition.go
+++ b/internal/ls/definition.go
@@ -1,15 +1,21 @@
 package ls
 
 import (
+	"bufio"
 	"context"
+	"os"
+	"regexp"
 	"slices"
+	"strings"
 
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/astnav"
 	"github.com/microsoft/typescript-go/internal/checker"
+	"github.com/microsoft/typescript-go/internal/compiler"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
 	"github.com/microsoft/typescript-go/internal/scanner"
+	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
 func (l *LanguageService) ProvideDefinition(ctx context.Context, documentURI lsproto.DocumentUri, position lsproto.Position) (lsproto.DefinitionResponse, error) {
@@ -104,10 +110,58 @@ func (l *LanguageService) createLocationsFromDeclarations(declarations []*ast.No
 	for _, decl := range declarations {
 		file := ast.GetSourceFileOfNode(decl)
 		name := core.OrElse(ast.GetNameOfDeclaration(decl), decl)
-		locations = core.AppendIfUnique(locations, lsproto.Location{
-			Uri:   FileNameToDocumentURI(file.FileName()),
-			Range: *l.createLspRangeFromNode(name, file),
-		})
+		
+		// Try to find the corresponding source file if this is a declaration file
+		// Use the file path directly so it works even for external declaration files
+		sourceFile := l.tryFindSourceFileForDeclarationByPath(file.FileName())
+		
+		// If we found a source file via project references, use it
+		if sourceFile != nil {
+			// Try to find the corresponding declaration in the source file
+			sourceDecl := l.tryFindDeclarationInSourceFile(decl, sourceFile)
+			
+			if sourceDecl != nil {
+				// Use the found declaration in the source file
+				sourceName := core.OrElse(ast.GetNameOfDeclaration(sourceDecl), sourceDecl)
+				locations = core.AppendIfUnique(locations, lsproto.Location{
+					Uri:   FileNameToDocumentURI(sourceFile.FileName()),
+					Range: *l.createLspRangeFromNode(sourceName, sourceFile),
+				})
+			} else {
+				// Fallback: point to the beginning of the source file
+				// This is not perfect but ensures we point to the right file
+				locations = core.AppendIfUnique(locations, lsproto.Location{
+					Uri:   FileNameToDocumentURI(sourceFile.FileName()),
+					Range: *l.createLspRangeFromBounds(0, 0, sourceFile),
+				})
+			}
+		} else {
+			// Fallback: Try heuristic approach for monorepos without project references
+			if candidateSourcePath := l.tryFindSourceFilePathHeuristic(file.FileName()); candidateSourcePath != "" {
+				// Try to find the approximate location of the symbol in the source file
+				if symbolLocation := l.tryFindSymbolLocationInFile(decl, candidateSourcePath); symbolLocation != nil {
+					locations = core.AppendIfUnique(locations, lsproto.Location{
+						Uri:   FileNameToDocumentURI(candidateSourcePath),
+						Range: *symbolLocation,
+					})
+				} else {
+					// Fallback: point to the beginning of the source file
+					locations = core.AppendIfUnique(locations, lsproto.Location{
+						Uri:   FileNameToDocumentURI(candidateSourcePath),
+						Range: lsproto.Range{
+							Start: lsproto.Position{Line: 0, Character: 0},
+							End:   lsproto.Position{Line: 0, Character: 0},
+						},
+					})
+				}
+			} else {
+				// Use the original approach for non-declaration files or when no mapping found
+				locations = core.AppendIfUnique(locations, lsproto.Location{
+					Uri:   FileNameToDocumentURI(file.FileName()),
+					Range: *l.createLspRangeFromNode(name, file),
+				})
+			}
+		}
 	}
 	return lsproto.LocationOrLocationsOrDefinitionLinksOrNull{Locations: &locations}
 }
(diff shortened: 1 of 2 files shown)


rgoomar:

alexwork1611:
Keeping an eye on this...

internal/project/compilerhost.go
```diff
@@ -144,7 +144,8 @@ func (fs *compilerFS) ReadFile(path string) (contents string, ok bool) {
 	if fh := fs.source.GetFile(path); fh != nil {
 		return fh.Content(), true
 	}
-	return "", false
+	// Fallback to underlying FS for files not tracked by the program (like .d.ts.map)
+	return fs.source.FS().ReadFile(path)
```
iisaduan: This shouldn't be added onto the compilerhost since `d.ts.map` is only used by ls features. Can you refactor so only the calls to `readFile` in `definitionSourceMap` has this fallback for `d.ts.map` files? It should probably also read into the snapshot's `fs`, instead of the underlying file system

rgoomar: Makes sense, I'll adjust that

internal/ls/definition.go
```diff
@@ -123,7 +123,12 @@ func (l *LanguageService) createLocationFromFileAndRange(file *ast.SourceFile, t
 
 func getDeclarationsFromLocation(c *checker.Checker, node *ast.Node) []*ast.Node {
 	if ast.IsIdentifier(node) && ast.IsShorthandPropertyAssignment(node.Parent) {
-		return c.GetResolvedSymbol(node).Declarations
+		if symbol := c.GetResolvedSymbol(node); symbol != nil {
+			return symbol.Declarations
+		}
+		if symbol := c.GetSymbolAtLocation(node); symbol != nil {
+			return symbol.Declarations
```
iisaduan: What does adding `getSymbolAtLocation` here fix?

rgoomar: I forgot to remove this change actually. It isn't doing anything here.

internal/project/compilerhost.go
rgoomar: @iisaduan I made the change with an alternative method here. Turns out that `fs.source.FS()` is the snapshot FS

rgoomar:
This is working for multiple engineers internally for us 😄

internal/ls/definitionsourcemap.go
```diff
@@ -0,0 +1,285 @@
...
+	return &lsproto.Location{
+		Uri: FileNameToDocumentURI(sourceRange.FileName),
+		Range: lsproto.Range{
+			Start: lsproto.Position{Line: uint32(sourceStartLine), Character: uint32(sourceStartChar)},
+			End:   lsproto.Position{Line: uint32(sourceEndLine), Character: uint32(sourceEndChar)},
+		},
+	}
+}
+
+type SourceRange struct {
```
jakebailey: I think all of this source mapping stuff really belongs in the `sourcemap` package. When I read the old impl, we did not have source map stuff directly implemented anywhere else, as far as I can tell. We had `createDocumentPositionMapper`, which was then referenced by the LS source map code, yes, for sure (and I think I new see what you were porting).

But, the way it stands, what needs to happen is to be doing source mapping in the source map package far away from the LS code, because the LS code needs to operate with client positioning, whereas like the emitter source map code, it needs to deal with byte encoding (converted to the weird UTF-16 + weird extra newlines ECMAScript defines).

rgoomar: Updated the PR to separate this

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,212 @@
...
+		sourceEndPos = len(sourceContent)
+	}
+	
+	sourceEndLine, sourceEndChar := core.PositionToLineAndCharacter(sourceEndPos, sourceLineStarts)
+
+	return &lsproto.Location{
+		Uri: FileNameToDocumentURI(startResult.FileName),
+		Range: lsproto.Range{
+			Start: lsproto.Position{Line: uint32(sourceStartLine), Character: uint32(sourceStartChar)},
+			End:   lsproto.Position{Line: uint32(sourceEndLine), Character: uint32(sourceEndChar)},
```
jakebailey: The conversions in this function unfortunately get the positioning info incorrect. We cannot use the line map from `core` in the LS _unless_ it is explicitly just for decoding mappings done by source mapping for ECMAScript `.map` files.

What this code needs to do (what I tried to describe before) is to only use line mapping from core just for purposes of decoding the source map (if it's even needed), get a byte offset, and then use the position mappers built into the LS to convert that byte offset to an LSP position.

jakebailey: Perhaps you're most of the way there given this fnct takes lsproto positions, but you need to use the position mapper from the LS to get byte offsets and go from there.

jakebailey (changes requested):

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,190 @@
...
+
+func tryMapLocation(sourceMapper sourcemap.SourceMapper, host *sourcemapHost, location lsproto.Location, languageService *LanguageService) *lsproto.Location {
+	fileName := location.Uri.FileName()
+	
+	declContent, ok := host.readFileWithFallback(fileName)
+	if !ok {
+		return nil
+	}
+
+	declLineStarts := core.ComputeLineStarts(declContent)
```
jakebailey: This should still get removed; the input is a lsproto location, and the output is too, so I expect to not use this, only LS positions (client defined encoding), or byte offsets, probably the latter, then pass that into the source mapper to do "whatever" with the source map position encoding, then return back a byte offset again.

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,190 @@
...
+	host := &sourcemapHost{program: program}
+	sourceMapper := CreateSourceMapperForProgram(program)
+	
+	return tryMapLocation(sourceMapper, host, location, languageService)
+}
+
+func tryMapLocation(sourceMapper sourcemap.SourceMapper, host *sourcemapHost, location lsproto.Location, languageService *LanguageService) *lsproto.Location {
+	fileName := location.Uri.FileName()
+	
+	declContent, ok := host.readFileWithFallback(fileName)
```
jakebailey: This ends up reading the contents even for files that we have parsed; probably we should have this instead return a sourcefilelike thing so the LS position mapper can still reuse its own line map, or otherwise cache it some place else?

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,190 @@
...
+func MapSingleDefinitionLocation(program *compiler.Program, location lsproto.Location, languageService *LanguageService) *lsproto.Location {
+	fileName := location.Uri.FileName()
+
+	if !strings.HasSuffix(fileName, ".d.ts") {
+		return nil
+	}
+
+	if strings.HasPrefix(fileName, "^/bundled/") {
+		return nil
+	}
```
jakebailey: I'm still not sure we need to special case this anymore?

internal/lsp/server.go
```diff
@@ -710,7 +711,61 @@ func (s *Server) handleSignatureHelp(ctx context.Context, languageService *ls.La
...
+	program := requestingProject.GetProgram()
+	if program == nil {
+		return locations
+	}
+
+	for _, location := range locations {
+		// Skip bundled library URIs that have invalid URL encoding
+		// These URIs contain "%3A" which causes panics in FileName()
+		uriString := string(location.Uri)
+		if strings.Contains(uriString, "bundled%3A") {
```
jakebailey, line 749: This too.

internal/ls/sourcemaphost.go
rgoomar: I don't think so. I'll remove it.

internal/lsp/server.go
rgoomar: Turns out this is needed. Tests fail without it.

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,197 @@
...
+	
+	return tryMapLocation(sourceMapper, host, location, languageService)
+}
+
+func tryMapLocation(sourceMapper sourcemap.SourceMapper, host *sourcemapHost, location lsproto.Location, languageService *LanguageService) *lsproto.Location {
+	fileName := location.Uri.FileName()
+	program := host.program
+	
+	declFile := program.GetSourceFile(fileName)
+	var declStartPos, declEndPos int
```
jakebailey: It seems like things would be simpler if these types were not `int`, but instead `int32` or `core.TextPos`; the code below does a lot of converting back and forth.

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,197 @@
...
+	sourceEndPos := startResult.Pos + originalRangeLength
+	
+	sourceFile := program.GetSourceFile(startResult.FileName)
+	var sourceStartLSP lsproto.Position
+	var sourceEndLSP lsproto.Position
+	
+	if sourceFile != nil {
+		if sourceEndPos > len(sourceFile.Text()) {
+			sourceEndPos = len(sourceFile.Text())
+		}
```
jakebailey: When would this be out of bounds? (These can be simplified with a `min` call).

internal/ls/sourcemaphost.go
```diff
@@ -0,0 +1,197 @@
...
+			sourceEndPos = len(sourceContent)
+		}
+		
+		textScript := &textScript{
+			fileName: startResult.FileName,
+			text:     sourceContent,
+		}
+		
+		sourceStartLSP = languageService.converters.PositionToLineAndCharacter(textScript, core.TextPos(startResult.Pos))
+		sourceEndLSP = languageService.converters.PositionToLineAndCharacter(textScript, core.TextPos(sourceEndPos))
```
jakebailey: These two blocks (if and else are the same); I would instead declare `textScript` outside, and then conditionally assign it, then move these two converter lines into the struct below.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,200 @@
...
+
+import (
+	"strings"
+
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+// SourceMapper provides source mapping functionality for LSP operations
+type SourceMapper interface {
```
jakebailey, line 11: I don't think all of these interfaces are needed; surely we only have one implementation of most of these and could just use a concrete type? do we need multiple hosts, interfaces, then struct fields of the various kinds?

internal/sourcemap/mapper.go
```diff
@@ -0,0 +1,270 @@
...
+
+	"github.com/go-json-experiment/json"
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+// DocumentPosition represents a position in a document
+type DocumentPosition struct {
+	FileName string
+	Pos      int
```
jakebailey: Could just be `core.TextPos`, I think?

internal/sourcemap/mapper.go
```diff
@@ -0,0 +1,270 @@
...
+	SourceIndex       int
+	SourcePosition    int
+	SourceFileName    string // Resolved source file name
+}
+
+// CreateDocumentPositionMapper creates a document position mapper from a source map
+func CreateDocumentPositionMapper(host DocumentPositionMapperHost, mapContent string, mapPath string) DocumentPositionMapper {
+	var sourceMap RawSourceMap
+	if err := json.Unmarshal([]byte(mapContent), &sourceMap); err != nil {
+		host.Log("Failed to parse source map: " + err.Error())
```
jakebailey: The old code didn't do the logging, did we handle this previously? I don't think adding logging to the source map package is good in general, but I could be wrong.

internal/sourcemap/mapper.go
```diff
@@ -0,0 +1,270 @@
...
+	GetSourceFileLike(fileName string) SourceFileLike
+	GetCanonicalFileName(path string) string
+	Log(text string)
+}
+
+// SourceFileLike represents a file with text content and line information
+type SourceFileLike interface {
+	Text() string
+	LineStarts() []core.TextPos
+}
```
jakebailey: The package already has a `Source` interface.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,211 @@
...
+	return &SimpleSourceFile{
+		fileName: fileName,
+		text:     text,
+		lineMap:  core.ComputeLineStarts(text),
+	}
+}
+
+// Helper functions
+
+func isDeclarationFileName(fileName string) bool {
```
jakebailey, line 188: This already exists in the `tspath` package, and other places you've used `HasSuffix` for this can be replaced too.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,211 @@
...
+	return strings.HasSuffix(fileName, ".d.ts")
+}
+
+func createGetCanonicalFileName(useCaseSensitiveFileNames bool) func(string) string {
+	if useCaseSensitiveFileNames {
+		return func(path string) string { return path }
+	}
+	return strings.ToLower
+}
```
jakebailey, line 197: This isn't a correct implementation, unfortunately. If we need to compare paths like this, we use `tspath.ToPath`.

We also don't pass around "canonical filename functions" anymore, instead choosing to pass around Paths and the case sensitivity.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,211 @@
...
+
+func tryGetSourceMappingURL(content string) string {
```
jakebailey, line 199: This func already exists in `util.go`.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+// SourceMapper provides source mapping functionality for LSP operations
+type SourceMapper interface {
+	TryGetSourcePosition(info DocumentPosition) *DocumentPosition
+	TryGetGeneratedPosition(info DocumentPosition) *DocumentPosition
+	ClearCache()
```
jakebailey, line 14: This ClearCache infra is unused.

internal/sourcemap/sourcemapper.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// SourceMapperHost provides the necessary dependencies for source mapping
+type SourceMapperHost interface {
+	DocumentPositionMapperHost
+	UseCaseSensitiveFileNames() bool
+	GetCurrentDirectory() string
+	ReadFile(path string) (string, bool)
+	FileExists(path string) bool
```
jakebailey, line 23: This could really just be an `FS()` method

rgoomar:
Closing this out since it has some issues with the architecture

PR #1628, opened 2025-08-22
Handle processing diagnostics similar to Strada

gabritto:
In Corsa, `GetProgramDiagnostics` was including all diagnostics of the `includeProcessor`, which includes `/// <reference ...>` diagnostics and resolution diagnostics. Because we did that, an error e.g. in a bad `/// <reference type="non-existent"/>` in a lib file would be included in the output of a command line execution even if the project had `skipLibCheck: true`.

This PR changes this so that `GetProgramDiagnostics` now only includes *global* `includeProcessor` diagnostics.
This is basically what Strada did (see `program.getCombinedDiagnostics` there).
The file-specific `includeProcessor` diagnostics are now obtained via `getSemanticDiagnosticsForFile`. This is also what Strada did.
The result is that in `getSemanticDiagnosticsForFile`, we have a chance to check if that file should be skipped per `skipLibCheck`, and we'll also handle `// @ts-ignore` correctly as well for these diagnostics.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -424,7 +424,9 @@ func (p *Program) GetSuggestionDiagnostics(ctx context.Context, sourceFile *ast.
 }
 
 func (p *Program) GetProgramDiagnostics() []*ast.Diagnostic {
-	return SortAndDeduplicateDiagnostics(slices.Concat(p.programDiagnostics, p.includeProcessor.getDiagnostics(p).GetDiagnostics()))
+	return SortAndDeduplicateDiagnostics(slices.Concat(
+		p.programDiagnostics,
+		p.includeProcessor.getDiagnostics(p).GetGlobalDiagnostics()))
 }
 
 func (p *Program) getSourceFilesToEmit(targetSourceFile *ast.SourceFile, forceDtsEmit bool) []*ast.SourceFile {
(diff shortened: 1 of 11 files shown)


testdata/baselines/reference/tsbuild/demo/in-bad-ref-branch-reports-the-error-about-files-not-in-rootDir-at-the-import-location.js
```diff
@@ -513,14 +517,86 @@ Object.defineProperty(exports, "createDog", { enumerable: true, get: function ()
     ]
   },
   "semanticDiagnosticsPerFile": [
-    "lib.d.ts",
-    "../../animals/animal.ts",
-    "../../animals/dog.ts",
-    "../../animals/index.ts",
-    "../../core/utilities.ts"
+    [
```
gabritto, line 520: As far as I could tell, these new errors are correct and show up in the Strada baselines for these tests.

internal/compiler/program.go
```diff
@@ -424,7 +424,9 @@ func (p *Program) GetSuggestionDiagnostics(ctx context.Context, sourceFile *ast.
 }
 
 func (p *Program) GetProgramDiagnostics() []*ast.Diagnostic {
-	return SortAndDeduplicateDiagnostics(slices.Concat(p.programDiagnostics, p.includeProcessor.getDiagnostics(p).GetDiagnostics()))
+	return SortAndDeduplicateDiagnostics(slices.Concat(
+		p.programDiagnostics,
+		p.includeProcessor.getDiagnostics(p).GetGlobalDiagnostics()))
```
jakebailey, line 429:
```suggestion
		p.includeProcessor.getDiagnostics(p).GetGlobalDiagnostics(),
	))
```

Nit (sorry)

PR #1639, opened 2025-08-26
fix(1632): erase const enums after inlining

### LINKED ISSUE #1632: Const enums shouldn't be emitted after inlining
Follow up on #1599

## Steps to reproduce

```ts
const enum A {
    B = 1
}

const a = A.B
```

The enum is inlined and then the enum itself is not emitted:

```js
const a = 1 /* A.B */;
```

## Behavior with `tsgo`
The enum is still emitted 
```js
var A;
(function (A) {
    A[A["B"] = 1] = "B";
})(A || (A = {}));
const a = 1 /* A.B */;
```

a-tarasyuk:
Fixes #1632


internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -309,6 +309,10 @@ func (tx *RuntimeSyntaxTransformer) addVarForDeclaration(statements []*ast.State
 }
 
 func (tx *RuntimeSyntaxTransformer) visitEnumDeclaration(node *ast.EnumDeclaration) *ast.Node {
+	if ast.IsEnumConst(node.AsNode()) || tx.compilerOptions.PreserveConstEnums.IsFalse() {
```
JoostK: Shouldn't this be `&&`?

testdata/baselines/reference/submodule/compiler/blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js.diff
```diff
@@ -0,0 +1,18 @@
+--- old.blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js
++++ new.blockScopedEnumVariablesUseBeforeDef_verbatimModuleSyntax.js
+@@= skipped -29, +29 lines =@@
+ }
+ function foo2() {
+     return E.A;
+-    let E;
```
jakebailey: This looks like a regression to investigate

a-tarasyuk: I think this is related to `ShouldPreserveConstEnums`, which doesn’t account for `verbatimModuleSyntax`, while that case is handled in

https://github.com/microsoft/TypeScript/blob/3eb7b6a1794a6d2cde7948a3016c57e628b104b9/src/compiler/utilities.ts#L9124-L9129

https://github.com/microsoft/TypeScript/blob/3eb7b6a1794a6d2cde7948a3016c57e628b104b9/src/compiler/utilities.ts#L9029-L9034

internal/core/compileroptions.go
```diff
@@ -277,7 +277,7 @@ func (options *CompilerOptions) GetResolveJsonModule() bool {
 }
 
 func (options *CompilerOptions) ShouldPreserveConstEnums() bool {
-	return options.PreserveConstEnums == TSTrue || options.IsolatedModules == TSTrue
+	return options.PreserveConstEnums == TSTrue || options.IsolatedModules == TSTrue || options.VerbatimModuleSyntax == TSTrue
```
jakebailey: Should probably just call GetIsolatedModules here.

a-tarasyuk: @jakebailey Yes, I've pushed changes

testdata/baselines/reference/submodule/compiler/constEnumErrors.js.diff
```diff
@@ -1,45 +1,10 @@
 --- old.constEnumErrors.js
 +++ new.constEnumErrors.js
-@@= skipped -48, +48 lines =@@
+@@= skipped -46, +46 lines =@@
+ }
+
  //// [constEnumErrors.js]
- var E;
+-var E;
```
jakebailey: This looks like it might have regressed; now there's no E and the code will crash at runtime.

testdata/baselines/reference/submodule/conformance/constEnum4.js.diff
```diff
@@ -1,27 +1,11 @@
...
 
  //// [constEnum4.js]
--if (1)
--    ;
--else if (2)
+ if (1)
 -    ;
--else
+ else if (2)
 -    ;
```
jakebailey: I think this is a regression; now we are emitting JS that won't parse, because the tree looks like:

```
if (1)
else if (2)
else
```

missing any braces.

testdata/baselines/reference/submodule/conformance/constEnum4.js
```diff
@@ -10,18 +10,6 @@ else
...
-    })(B || (B = {}));
-}
-else {
-    var C;
-    (function (C) {
-    })(C || (C = {}));
-}
+if (1)
+else if (2)
+else
```
jakebailey: I commented on the diff (I usually only look at those), but this is the problematic transform I was pointing out.

I suspect this comes down to the emit for if/else/similar needing to ensure that bodies are still printed out if missing?

a-tarasyuk: @jakebailey I’m currently investigating this case

testdata/baselines/reference/submodule/compiler/elidedEmbeddedStatementsReplacedWithSemicolon.js.diff
```diff
@@ -1,64 +1,19 @@
...
-@@= skipped -26, +26 lines =@@
-     const enum H {}
-
- //// [elidedEmbeddedStatementsReplacedWithSemicolon.js]
--if (1)
+@@= skipped -31, +31 lines =@@
+ else
+     ;
+ do
 -    ;
```
jakebailey: Hm, I wonder how much more it'd take to make this one fixed too. Maybe that's out of scope.

testdata/baselines/reference/submodule/conformance/constEnum4.js
a-tarasyuk: It seems `EmbeddedStatement` doesn’t treat `NotEmittedStatement` as an empty statement the way Strada does.

https://github.com/microsoft/TypeScript/blob/7956c00166df552a1d6da6c2f996b558e5dd94ff/src/compiler/factory/nodeFactory.ts#L3928-L3929

https://github.com/microsoft/TypeScript/blob/7956c00166df552a1d6da6c2f996b558e5dd94ff/src/compiler/factory/nodeFactory.ts#L7163-L7167

As a result, when updating a statement (for example, an `if`), a `NotEmittedStatement` is returned and the `;` is skipped.

https://github.com/microsoft/typescript-go/blob/52f2ee4f6e31ef7072dc4014bfc7e73649ad6a92/internal/printer/printer.go#L3895-L3899

internal/printer/emitcontext.go
```diff
@@ -961,3 +962,14 @@ func (c *EmitContext) GetSyntheticTrailingComments(node *ast.Node) []Synthesized
 	}
 	return nil
 }
+
+func (c *EmitContext) VisitEmbeddedStatement(node *ast.Statement, visitor *ast.NodeVisitor) *ast.Statement {
+	updated := visitor.VisitEmbeddedStatement(node)
```
jakebailey: I'm sort of surprised by this, given I thought the default was to do this already...

testdata/baselines/reference/submodule/compiler/elidedEmbeddedStatementsReplacedWithSemicolon.js.diff
a-tarasyuk: This issue is fixed now

testdata/baselines/reference/submodule/conformance/constEnum4.js
jakebailey: So, does that mean the fix should be elsewhere?

internal/printer/emitcontext.go
a-tarasyuk: It looks like it doesn’t by default

https://github.com/microsoft/typescript-go/blob/52f2ee4f6e31ef7072dc4014bfc7e73649ad6a92/internal/ast/visitor.go#L74-L84

https://github.com/microsoft/typescript-go/blob/52f2ee4f6e31ef7072dc4014bfc7e73649ad6a92/internal/ast/visitor.go#L201-L209

I’m not convinced the visitor is the right place for this - WDYT? Also, the change can be implemented via a transformer-level hook or in the emitter.

a-tarasyuk:
@jakebailey ok, thanks

internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -309,6 +309,10 @@ func (tx *RuntimeSyntaxTransformer) addVarForDeclaration(statements []*ast.State
 }
 
 func (tx *RuntimeSyntaxTransformer) visitEnumDeclaration(node *ast.EnumDeclaration) *ast.Node {
+	if !tx.shouldEmitEnumDeclaration(node) {
+		return tx.Factory().NewNotEmittedStatement()
```
weswigham: You want a helper to make these, like in the old logic, that sets the `original` node pointer and copies the `.Loc`. `tx.EmitContext().NewNotEmittedStatement(node)` would be what I expect.

internal/printer/emitcontext.go
```diff
@@ -961,3 +962,14 @@ func (c *EmitContext) GetSyntheticTrailingComments(node *ast.Node) []Synthesized
...
+	if updated == nil {
+		return nil
+	}
+	if ast.IsNotEmittedStatement(updated) {
+		return visitor.Factory.NewEmptyStatement()
```
weswigham: This needs to copy position/comment emit node information from `updated` - `NotEmittedStatement`s are injected into the tree explicitly to preserve that kind of thing, so failing to copy it here unintentionally drops it.

Though, why is this needed at all, actually? Is it just so follow-on transforms can see an `EmptyStatement` instead of a `NotEmittedStatement` and react appropriately? It might be better to not perform a remapping on transform like this, and instead use a `ast.IsEmptyStatementLike` helper that includes both `EmptyStatement` and `NotEmittedStatement`. Either one of the two strategies to preserve comment information should be fine, though (either copying that info here or dropping the remapping entirely and updating downstream transforms/printers to handle `NotEmittedStatement`s like `EmptyStatement`s).

internal/transformers/tstransforms/runtimesyntax.go
a-tarasyuk: @weswigham, thanks for the feedback. I've added `NewNotEmittedStatement` to `EmitContext `

internal/printer/emitcontext.go
weswigham: ^^ this comment still needs to be addressed - this remapping drops comments and location information.

a-tarasyuk: @weswigham Thanks for pointing that out. The main issue is using _remapping_ to handle cases related to embedded statements

```ts
// ts
if (true)
  const enum E1 { A = 1 }
else 
  const enum E2 { A = 1 }

// js
if (true)
    ;
else
    ;
```

https://github.com/microsoft/typescript-go/blob/93473660c085f4156c3a1b9a038ca78098a137fb/internal/printer/printer.go#L3163-L3174

`Strada` uses this helper to wrap all embedded statements

https://github.com/microsoft/TypeScript/blob/7956c00166df552a1d6da6c2f996b558e5dd94ff/src/compiler/factory/nodeFactory.ts#L7163-L7167

It might be an option to extend

https://github.com/microsoft/typescript-go/blob/93473660c085f4156c3a1b9a038ca78098a137fb/internal/printer/printer.go#L3430-L3432

for cases involving embedded statements.

weswigham: Yeah, you can use the transform hook to automatically do the thing the embedded statement helper is doing like you are here (probably - we're moving the remapping from node construction later a step to the generic visitor base, so it's possible a transform could witness the difference before returning; but transforms don't usually inspect what the node factory returns anyway), you definitely just need the corsa equivalents of the setTextRange+setOriginal calls.

a-tarasyuk: @weswigham  I was thinking about the following options:

- Extend `emitEmbeddedStatement` to emit a `;` for `NotEmittedStatement`, since it currently feels more like an alias used only at runtime

https://github.com/microsoft/typescript-go/blob/93473660c085f4156c3a1b9a038ca78098a137fb/internal/printer/printer.go#L3895-L3899

- Add handling logic to runtimesyntax so that when a `NotEmittedStatement` appears inside an `EmbeddedStatement`, an `EmptyStatement` is emitted instead.

a-tarasyuk: @weswigham I’ve made the changes to address https://github.com/microsoft/typescript-go/pull/1639#discussion_r2323131817

weswigham (approved):

Issue #1520, opened 2025-08-05
--incremental builds are a bit slower after the 20250805 release


joshcartme:
Parts of the incremental build (like changes compute time) are now significantly faster, but overall incremental builds are a bit slower.

memprofiles for the warm builds shown below:
- [20250804](https://github.com/user-attachments/files/21606380/55223-memprofile.pb.gz)
- [20250805](https://github.com/user-attachments/files/21606385/55922-memprofile.pb.gz)

### Diagnostic output:

2025-08-04 cold:
```
Files:                    48433
Lines:                  6937671
Identifiers:            8358012
Symbols:               19135744
Types:                  9232629
Instantiations:        33855859
Memory used:          15656224K
Memory allocs:        274036051
Config time:             1.981s
BuildInfo read time:     0.384s
Parse time:              2.853s
Bind time:               0.001s
Check time:             25.959s
Emit time:               3.792s
Changes compute time:    6.985s
Total time:             41.958s
```
2025-08-04 warm:
```
Files:                   48433
Lines:                 6937671
Identifiers:           8358012
Symbols:               4472788
Types:                     348
Instantiations:              0
Memory used:          6647971K
Memory allocs:        36869491
Config time:            1.989s
BuildInfo read time:    4.331s
Parse time:             2.999s
Bind time:              0.001s
Check time:             0.007s
Emit time:              0.006s
Changes compute time:   8.441s
Total time:            17.776s
```

2025-08-05 cold:
```
Files:                    48433
Lines:                  6937671
Identifiers:            8358012
Symbols:               19135734
Types:                  9232433
Instantiations:        33855859
Memory used:          17073228K
Memory allocs:        349832901
Config time:             2.014s
BuildInfo read time:     0.393s
Parse time:              2.970s
Bind time:               0.001s
Check time:             27.795s
Emit time:               4.138s
Changes compute time:    2.127s
Total time:             39.440s
```
2025-08-05 warm (substantially more memory allocs vs the 2025-08-04 warm build):
```
Files:                    48433
Lines:                  6937671
Identifiers:            8358012
Symbols:                4472788
Types:                      348
Instantiations:               0
Memory used:           8064926K
Memory allocs:        223219055
Config time:             2.028s
BuildInfo read time:     9.537s
Parse time:              4.549s
Bind time:               0.001s
Check time:              0.045s
Emit time:               2.976s
Changes compute time:    3.566s
Total time:             22.705s
```

jakebailey:
[...]

PR #1640, opened 2025-08-27
Speedup GetSpellingSuggestion

dzbarsky:
Noticed we weren't using buffers while trying the [Kibana repro](https://github.com/microsoft/typescript-go/issues/1520#issuecomment-3169391766).

Before:
```
➜  kibana git:(main) ✗ time ../typescript-go/tsgo -p tsconfig.json -pprofDir /tmp/pprof                    
tsconfig.json:3:3 - error TS5090: Non-relative paths are not allowed. Did you forget a leading './'?

3   "compilerOptions": {
    ~~~~~~~~~~~~~~~~~
tsconfig.json:3:3 - error TS5102: Option 'baseUrl' has been removed. Please remove it from your configuration.
  Use '"paths": {"*": ["./*"]}' instead.

3   "compilerOptions": {
    ~~~~~~~~~~~~~~~~~

Found 2 errors in the same file, starting at: tsconfig.json:3

CPU profile: /tmp/pprof/92848-cpuprofile.pb.gz
Memory profile: /tmp/pprof/92848-memprofile.pb.gz
../typescript-go/tsgo -p tsconfig.json -pprofDir /tmp/pprof  273.49s user 45.88s system 299% cpu 1:46.77 total
```
image

After:
```
../typescript-go/tsgo -p tsconfig.json -pprofDir /tmp/pprof  237.25s user 40.39s system 286% cpu 1:36.83 total
```
image

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -452,6 +452,10 @@ func GetScriptKindFromFileName(fileName string) ScriptKind {
 	return ScriptKindUnknown
 }
 
+type SpellingSuggestionBuffers struct {
+	previous, current []float64
+}
+
 // Given a name and a list of names that are *not* equal to the name, return a spelling suggestion if there is one that is close enough.
 // Names less than length 3 only check for case-insensitive equality.
 //
@@ -465,7 +469,7 @@ func GetScriptKindFromFileName(fileName string) ScriptKind {
 //	     and 1 insertion/deletion at 3 characters)
 //
 // @internal
-func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) string) T {
+func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) string, buffers *SpellingSuggestionBuffers) T {
 	maximumLengthDifference := max(2, int(float64(len(name))*0.34))
 	bestDistance := math.Floor(float64(len(name))*0.4) + 1 // If the best result is worse than this, don't bother.
 	runeName := []rune(name)
@@ -483,7 +487,7 @@ func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) s
 			if len(candidateName) < 3 && !strings.EqualFold(candidateName, name) {
 				continue
 			}
-			distance := levenshteinWithMax(runeName, []rune(candidateName), bestDistance-0.1)
+			distance := levenshteinWithMax(runeName, []rune(candidateName), bestDistance-0.1, buffers)
 			if distance < 0 {
 				continue
 			}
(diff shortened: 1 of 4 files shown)


jakebailey:
This should only matter in erroring code. Are you silencing a whole bunch of errors?

dzbarsky:
I'm not sure? The output above shows the only 2 errors that are reported.

jakebailey:
If possible I'd print something out each time he function is called and count how often it is, or look around for ts-ignores/ts-expect-error.

I think we can totally accept an optimization like this, but we certainly aren't expecting the diagnostic code paths to be hot.

dzbarsky:
There are around 3.1K ts-expect-error but I see GetSpellingSuggestion called around 750K times. I checked how many options were considered for each call (i.e. how many calls we executed to compute levenshtein distance, not just how many were provided):
```
count=751877 min=0 max=1060 avg=41.9747
```

A majority of calls only execute up to 2-3 computations, but for some reason there's 122K calls that consider exactly 175 possibilities...
```
Freq.   # of distances computed
325284 0
129899 1
122505 175
59133 2
26471 3
17585 26
14454 4
7951 211
7901 7
6482 47
5141 5
4515 87
4057 6
1774 375
1191 8
 935 538
 920 555
 890 9
 870 584
 832 514
 819 486
 807 431
 790 551
 790 522
```

jakebailey:
That many ignored errors is terrifying; I would strongly suggest converting those to casts or similar.

But, would be good to improve nevertheless...

RyanCavanaugh:
> There are around 3.1K ts-expect-error

Can you characterize these for us? Why so many? We only compute spelling suggestions we can't find an identifier or property with that name -- situations that should be very, very uncommon in theory (and usually have straightforward type-side fixes available, e.g. `declare missingVar: any`).

internal/core/core.go
```diff
@@ -495,9 +499,24 @@ func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) s
 	return bestCandidate
 }
 
-func levenshteinWithMax(s1 []rune, s2 []rune, maxValue float64) float64 {
-	previous := make([]float64, len(s2)+1)
-	current := make([]float64, len(s2)+1)
+func ensureSize(slice []float64, desired int) []float64 {
```
jakebailey, line 502: This can just use `slices.Grow`

internal/core/core.go
```diff
@@ -495,9 +499,24 @@ func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) s
...
+	return slice[:desired]
+}
+
+func levenshteinWithMax(s1 []rune, s2 []rune, maxValue float64, buffers *SpellingSuggestionBuffers) float64 {
+	if buffers == nil {
+		buffers = &SpellingSuggestionBuffers{}
+	}
+
+	buffers.previous = ensureSize(buffers.previous, len(s2)+1)
+	previous := buffers.previous
```
jakebailey, line 515: This would honestly benefit from simply using a `sync.Pool` to store the buffers globally, rather than scoping the object. Means less plumbing and callers don't have to know anything about it.

PR #1641, opened 2025-08-28
Snapshots: more logging, fix memory leak

andrewbranch:
I forgot to clean up the diskFile cache, so it was growing infinitely. The new logging exposed the problem.

### DIFF
--- a/internal/project/ata/ata_test.go
+++ b/internal/project/ata/ata_test.go
@@ -382,6 +382,8 @@ func TestATA(t *testing.T) {
 			Type: lsproto.FileChangeTypeChanged,
 			Uri:  lsproto.DocumentUri("file:///user/username/projects/project/package.json"),
 		}})
+		// diagnostics refresh triggered - simulate by getting the language service
+		_, _ = session.GetLanguageService(context.Background(), uri)
 		session.WaitForBackgroundTasks()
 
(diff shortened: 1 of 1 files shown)


internal/project/session.go
```diff
@@ -91,14 +103,6 @@ type Session struct {
 	pendingATAChanges   map[tspath.Path]*ATAStateChange
 	pendingATAChangesMu sync.Mutex
 
-	// snapshotUpdateCancel is the cancelation function for a scheduled
-	// snapshot update. Snapshot updates are debounced after file watch
-	// changes since many watch events can occur in quick succession
-	// during `npm install` or git operations.
-	// !!! This can probably be replaced by ScheduleDiagnosticsRefresh()
-	snapshotUpdateCancel context.CancelFunc
-	snapshotUpdateMu     sync.Mutex
```
andrewbranch, line 100: Removing this and the related code was a bit of a drive-by cleanup. There’s really no reason to trigger a snapshot change internally; the only time we were using this we _also_ scheduled a diagnostics refresh, and the subsequent diagnostics request would have updated the snapshot as needed. It's simpler and cleaner to let snapshot updates be purely reactive to a client request/notification.

jakebailey (approved):

PR #1643, opened 2025-08-28
Do not store include processing diagnsotics in buildInfo

sheetalkamat:
#1628 started adding program diagnostics to semantic diagnostic reporting like Strada
This handles incremental scenarios like Strada where we don't store program diagnostics to file because we don't know if they change or not if program structure changes and semantic diagnostics from cache are used

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -1005,10 +1005,6 @@ func (p *Program) getBindDiagnosticsForFile(ctx context.Context, sourceFile *ast
 }
 
 func FilterNoEmitSemanticDiagnostics(diagnostics []*ast.Diagnostic, options *core.CompilerOptions) []*ast.Diagnostic {
-	if diagnostics == nil {
-		return diagnostics
-	}
-
 	if !options.NoEmit.IsTrue() {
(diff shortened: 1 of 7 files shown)


gabritto:
I thought I had left a review yesterday but apparently I forgot to submit it, sorry. But one thing I think may be missing is that `getSemanticDiagnostics` also filters out diagnostics according to `// @ts-ignore`, and in Strada that included the processing diagnostics, whereas I think in this PR it now doesn't.

jakebailey:
I actually thought that this change preserved that, given it doesn't seem like a related test changed, but maybe I've misread the PR?

gabritto:
> I actually thought that this change preserved that, given it doesn't seem like a related test changed, but maybe I've misread the PR?

Do we have a test for this?

jakebailey:
Maybe I'm hallucinating that your PR had one 😅

gabritto:
> Maybe I'm hallucinating that your PR had one 😅

No 🙁

sheetalkamat:
Sorry that i missed this part when porting this. #1649 adds the filtering

PR #1644, opened 2025-08-28
Fix panic recovery in server and fail test updater on unrecovered panics

gabritto:
According to the Go language spec:
> The return value of recover is nil when the goroutine is not panicking or recover was not called directly by a deferred function.

So the panic recovery in `server.go` wasn't working, because the deferred function wasn't directly calling `recover()` (instead, it would call `s.recover()` which would call `recover()`, which returned `nil` even if a panic happened).
I got rid of the indirection when calling `recover()`. A downside to it is that we now have to pass a request message to the `handle...` methods, because the recovery code accesses request message properties.

This PR also makes the fourslash test failure updater script throw when it detects an unrecovered panic, because that means `go test` will have stopped as soon as the panic happens, and since it didn't run all tests, we can't know which tests fail and which don't.

### DIFF
--- a/internal/fourslash/_scripts/updateFailing.mts
+++ b/internal/fourslash/_scripts/updateFailing.mts
@@ -7,6 +7,7 @@ import { main as convertFourslash } from "./convertFourslash.mts";
 const failingTestsPath = path.join(import.meta.dirname, "failingTests.txt");
 
 function main() {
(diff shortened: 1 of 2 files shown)


jakebailey (approved):

iisaduan (approved):
Looks good, I merged it to autoimports and it works. Thanks!

alexwork1611:
Thank you for your work!

PR #1648, opened 2025-08-29
Fix BFS loop during config lookup

gabritto:
We were doing an infinite BFS in the project collection builder while looking for a tsconfig for a given file. The loop happened because the BFS node type included a pointer to a `logger`, which changed every time we computed the neighbors of a node, and so no node was ever considered as already visited, even if we had already visited its corresponding tsconfig.

### DIFF
--- a/internal/core/bfs.go
+++ b/internal/core/bfs.go
@@ -8,88 +8,88 @@ import (
 	"github.com/microsoft/typescript-go/internal/collections"
 )
 
-type BreadthFirstSearchResult[N comparable] struct {
+type BreadthFirstSearchResult[N any] struct {
 	Stopped bool
 	Path    []N
 }
 
-type breadthFirstSearchJob[N comparable] struct {
+type breadthFirstSearchJob[N any] struct {
 	node   N
 	parent *breadthFirstSearchJob[N]
 }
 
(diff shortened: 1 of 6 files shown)


internal/core/bfs.go
```diff
@@ -8,88 +8,88 @@ import (
...
 
-type breadthFirstSearchJob[N comparable] struct {
+type breadthFirstSearchJob[N any] struct {
 	node   N
 	parent *breadthFirstSearchJob[N]
 }
 
-type BreadthFirstSearchLevel[N comparable] struct {
-	jobs *collections.OrderedMap[N, *breadthFirstSearchJob[N]]
+type BreadthFirstSearchLevel[K comparable, N interface{ Key() K }] struct {
```
gabritto: Two things:
1. I don't know if having this kind of constraint here is the best way to do this. It would also be possible to pass `Key` as a function, I think, or maybe some other setup I haven't thought of.

2. I'm not sure if `Key` is the best name for this, so open to suggestions.

andrewbranch: I wonder if `key()` as a function (only in the `Ex` version) would be less disruptive. The change to the unit test seemed a little annoying to work with.

gabritto: I'll try that and see how it goes.

PR #1649, opened 2025-08-29
Fix error filtering by comment directives for program diagnostics

sheetalkamat:
Fixes missing filtering mentioned at https://github.com/microsoft/typescript-go/pull/1643#issuecomment-3237360263

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -433,7 +433,8 @@ func (p *Program) GetIncludeProcessorDiagnostics(sourceFile *ast.SourceFile) []*
 	if checker.SkipTypeChecking(sourceFile, p.Options(), p, false) {
 		return nil
 	}
-	return p.includeProcessor.getDiagnostics(p).GetDiagnosticsForFile(sourceFile.FileName())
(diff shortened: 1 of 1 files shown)


internal/compiler/program.go
```diff
@@ -1058,8 +1059,20 @@ func (p *Program) getSemanticDiagnosticsForFileNotFilter(ctx context.Context, so
 		})
 	}
 
+	filtered, directivesByLine := p.getDiagnosticsWithPrecedingDirectives(sourceFile, diags)
+	for _, directive := range directivesByLine {
```
gabritto, line 1063: I'm slightly confused about why this got moved here instead of in `getDiagnosticsWithPrecedingDirectives`. Would this mean that if a file has a `@ts-expect-error` and the next line has an `includeProcessing` error, then we will error for it here? Though this scenario is pretty convoluted.

Also, would be nice to add a test to make sure this is working and continues to work, since apparently we don't have tests for this from Strada (or maybe we haven't ported them yet).

sheetalkamat: This was how it was in Strada where unused expectations are not reported for program diagnostics

gabritto (approved):

PR #1653, opened 2025-08-30
Replace mutex with atomic in queue.go

arturmelanchyk:
Tiny optimization to avoid unneeded locking

### DIFF
--- a/internal/project/background/queue.go
+++ b/internal/project/background/queue.go
@@ -3,13 +3,13 @@ package background
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 )
 
 // Queue manages background tasks execution
 type Queue struct {
 	wg     sync.WaitGroup
-	mu     sync.RWMutex
-	closed bool
+	closed atomic.Bool
 }
 
 // NewQueue creates a new background queue for managing background tasks execution.
@@ -18,12 +18,9 @@ func NewQueue() *Queue {
 }
 
 func (q *Queue) Enqueue(ctx context.Context, fn func(context.Context)) {
-	q.mu.RLock()
-	if q.closed {
-		q.mu.RUnlock()
+	if q.closed.Load() {
 		return
 	}
-	q.mu.RUnlock()
 
 	q.wg.Add(1)
 	go func() {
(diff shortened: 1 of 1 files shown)


jakebailey:
Are you actually seeing a case where this matters?

arturmelanchyk:
Hi @jakebailey 
This looks cleaner and comes with less overhead

LukeAbby:
I was looking into this code since it smelled fishy. It helped me uncover that `Wait` actually isn't okay (even without this PR). `sync.WaitGroup` really buries the lede and isn't always super explicit but it can actually panic in a number of cases. In this case it's at least documented and says this under `(*sync.WaitGroup).Add`:
> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait.

And if you look in the implementation you can see this, confirming that it's a panic:
```go
	if w != 0 && delta > 0 && v == int32(delta) {
		panic("sync: WaitGroup misuse: Add called concurrently with Wait")
	}
```
(`w` is the number of workers and delta equals 1 when beginning a task or -1 when ending a task)

So when `(*Queue).Wait` says "It does not prevent new tasks from being enqueued while waiting." this is untrue if `Wait` is called while 0 workers are running (say they simply finished quickly) as if `Enqueue` is run in parallel at that point it will panic.

Now granted `(*Queue).Wait` is currently only called in `WaitForBackgroundTasks` which says "This is intended to be used only for testing purposes." but the occasional panic in test still seems bad. To make `Wait` truly safe it needs to be holding the write end of the `*sync.RWMutex`. It can't be safely done by making `Wait` close the queue with the atomic boolean since this sequence of events could occur:

Thread A calls Enqueue -> Thread A loads `q.close` and observes it as false ->
Thread B calls Wait -> Thread B sets `q.close` to true (expecting to prevent any other thread from being able to enqueue anything) -> Thread B calls `q.wg.Wait` ->
Thread A calls `q.wg.Add` thus causing a panic.

jakebailey:
@arturmelanchyk That doesn't really answer my question. While this is better, it's best to focus on perf and correctness issues seen in actual codebases...

@LukeAbby I think you're right here, the saving grace is that we never actually call `Close` until the test has already completed, as part of deferred/cleanup work.

PR #1654, opened 2025-08-30
Add size arg in make func

arturmelanchyk:


internal/ls/findallreferences.go
```diff
@@ -464,7 +464,7 @@ func (l *LanguageService) ProvideRename(ctx context.Context, params *lsproto.Ren
 	options := refOptions{use: referenceUseRename, useAliasesForRename: true}
 	symbolsAndEntries := l.getReferencedSymbolsForNode(ctx, position, node, program, program.GetSourceFiles(), options, nil)
 	entries := core.FlatMap(symbolsAndEntries, func(s *SymbolAndEntries) []*referenceEntry { return s.references })
-	changes := make(map[lsproto.DocumentUri][]*lsproto.TextEdit)
+	changes := make(map[lsproto.DocumentUri][]*lsproto.TextEdit, len(entries))
```
jakebailey, line 467: This seems wrong, no? The loop goes over the entries, yes, but it's building up the map based on the unique URIs, which I would think would be smaller. As in, you could be finding all ref on a symbol which appears solely in a single file 1000 times, but this code would be assuming that there are 1000 potential files.

jakebailey (changes requested):

PR #1655, opened 2025-08-30
Use collections.Set instead of map of bools for cachedArgumentsReferenced

arturmelanchyk:


jakebailey (changes requested):
This is incorrect; we need `map[*ast.Node]bool` because we both need to know the result _and_ if it's been calculated. If we were to take this PR, we'd repeatedly recalculate the result for every node where the answer is "false", because nothing is ever added to the set.

PR #1668, opened 2025-09-02
Don't use locale-aware sorting in completions

gabritto:
Haven't removed the collator package because presumably we'll need it for organize imports.

[Collator sorting is slow](#1582), and using non-locale aware sorting doesn't seem like it would noticeably impact the user experience. For instance, vscode sorts its completion items by using regular string comparison on labels: https://github.com/microsoft/vscode/blob/8ee3a001179ba5153d152a3256c10158e18e42f9/src/vs/editor/contrib/suggest/browser/suggest.ts#L338

### DIFF
--- a/internal/stringutil/compare.go
+++ b/internal/stringutil/compare.go
@@ -90,3 +90,11 @@ func HasSuffix(s string, suffix string, caseSensitive bool) bool {
 	}
 	return strings.EqualFold(s[len(s)-len(suffix):], suffix)
 }
+
+func CompareStringsCaseInsensitiveThenSensitive(a, b string) Comparison {
+	cmp := CompareStringsCaseInsensitive(a, b)
+	if cmp != ComparisonEqual {
+		return cmp
+	}
+	return CompareStringsCaseSensitive(a, b)
+}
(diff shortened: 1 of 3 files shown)


internal/stringutil/compare.go
```diff
@@ -90,3 +90,11 @@ func HasSuffix(s string, suffix string, caseSensitive bool) bool {
 	}
 	return strings.EqualFold(s[len(s)-len(suffix):], suffix)
 }
+
+func CompareStringsCaseInsensitiveThenSensitive(a, b string) Comparison {
```
jakebailey, line 94: Why do this instead of just insensitive entirely? Is insensitive not a total order or something?

gabritto: Yes. "Ab" and "ab" are going to be equal under case insensitive comparison, but we still want one of them to be consistently sorted before the other for the tests.

jakebailey: Hm, it's concerning that our case insensitive sort isn't correct, then

gabritto: I think it really depends on what you're using it for. If you're using it to, say, dedupe paths in a case insensitive FS, then "Ab" and "ab" should be equal.

Issue #1549, opened 2025-08-08
Crash in incremental due to inconsistent casing


gabritto:
Tracking the crash mentioned in #1535.

## Stack trace

```
panic: ManyToManySet.Set: key already exists: c:/users/gabrielaa/code/.../removable.d.ts

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/collections.(*ManyToManySet[...]).Set(0xa65e80, {0xc002fc0700, 0x5d}, 0xc000c07800)
        C:/Users/gabrielaa/code/typescript-go/internal/collections/manytomanyset.go:31 +0x33a
github.com/microsoft/typescript-go/internal/incremental.newSnapshotForProgram(0xc0000036c0, 0x0, 0x0)
        C:/Users/gabrielaa/code/typescript-go/internal/incremental/snapshot.go:373 +0x572
github.com/microsoft/typescript-go/internal/incremental.NewProgram(0xc0000036c0, 0x0, 0x0)
        C:/Users/gabrielaa/code/typescript-go/internal/incremental/program.go:36 +0x25
github.com/microsoft/typescript-go/internal/execute.performIncrementalCompilation({0xa54370, 0xc0000ceba0}, 0xc0000e61e0, 0xc0001f4e00, 0xc000037d10, 0x36d6ac, 0x0)
        C:/Users/gabrielaa/code/typescript-go/internal/execute/tsc.go:247 +0x377
github.com/microsoft/typescript-go/internal/execute.tscCompilation({0xa54370, 0xc0000ceba0}, 0xc0000e60f0, 0x0)
        C:/Users/gabrielaa/code/typescript-go/internal/execute/tsc.go:194 +0xb53
github.com/microsoft/typescript-go/internal/execute.CommandLine({0xa54370, 0xc0000ceba0}, {0xc00008a150, 0x2, 0x3}, 0x0)
        C:/Users/gabrielaa/code/typescript-go/internal/execute/tsc.go:63 +0x17e
main.runMain()
        C:/Users/gabrielaa/code/typescript-go/cmd/tsgo/main.go:23 +0x109
main.main()
        C:/Users/gabrielaa/code/typescript-go/cmd/tsgo/main.go:10 +0x13
```

## Steps to reproduce

1. From #1535, you can repro it with the following test in `tsc_test.go`:
```go
{
			subScenario:     "Compile incremental with case insensitive file names",
			commandLineArgs: []string{"-p", "."},
			files: FileMap{
[...]

PR #1675, opened 2025-09-03
Fix out of bounds panic when tsconfig "files" contains duplicates

andrewbranch:
Discovered this while investigating something else

### DIFF
--- a/internal/tsoptions/parsedcommandline.go
+++ b/internal/tsoptions/parsedcommandline.go
@@ -37,6 +37,7 @@ type ParsedCommandLine struct {
 	resolvedProjectReferencePaths     []string
 	resolvedProjectReferencePathsOnce sync.Once
 
+	literalFileNamesLen int
 	fileNamesByPath     map[tspath.Path]string // maps file names to their paths, used for quick lookups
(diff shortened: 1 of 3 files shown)


andrewbranch:
Wait this isn't quite right

andrewbranch:
Yes it is. Test modification proves it

sheetalkamat (commented):
I think fix should be at program and hence duplicate of https://github.com/microsoft/typescript-go/issues/1549

The main blocking here is file casing and module resolutions that happen and discussion for that fix here https://github.com/microsoft/typescript-go/pull/1535

andrewbranch:
I don’t think this is a duplicate; I wasn’t targeting a fix to programs at all. The test case added here, using only config file parsing and its methods, panics without the fix. `LiteralFileNames` is used by the LSP infrastructure without a program being involved.

sheetalkamat:
> I don’t think this is a duplicate; I wasn’t targeting a fix to programs at all. The test case added here, using only config file parsing and its methods, panics without the fix. `LiteralFileNames` is used by the LSP infrastructure without a program being involved.

Oh i see.. I forgot that we use Literal file names stuff for watching and matching..

PR #1678, opened 2025-09-04
test(jsnum): add binary/oct/hex BigInt cases including underscores and uppercase prefixes

Harsh1925:
Adds unit tests for ParsePseudoBigInt across non-decimal bases, responding to the
inline TODO ("tests for other bases") in internal/jsnum/pseudobigint_test.go.

- Binary (0b/0B), Octal (0o/0O), Hex (0x/0X)
- With/without underscore separators
- Small-number cases (0, 1, 10, 255)
No behavior changes; tests pass with the current implementation.

### DIFF
--- a/internal/jsnum/pseudobigint_test.go
+++ b/internal/jsnum/pseudobigint_test.go
@@ -33,6 +33,42 @@ func TestParsePseudoBigInt(t *testing.T) {
 
 	// TODO(jakebailey): tests for other bases
 
+		t.Run("parse non-decimal bases (small numbers)", func(t *testing.T) {
+		t.Parallel()
+
+		type tc struct {
+			lit string
+			out string
+		}
+		cases := []tc{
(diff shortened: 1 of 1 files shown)


Harsh1925:
I’ve rebased, formatted with dprint, and CI is now green. Could you take another look when you have a moment?

internal/jsnum/pseudobigint_test.go
```diff
@@ -33,6 +33,42 @@ func TestParsePseudoBigInt(t *testing.T) {
 
 	// TODO(jakebailey): tests for other bases
```
jakebailey: I think this TODO can be removed now

Harsh1925:
Removed the resolved TODO and ran dprint fmt on the test file. Thanks!

jakebailey (approved):

PR #1680, opened 2025-09-04
Add type definitions for `Uint8Array` to/from base64 methods

dontwanttothink:
Fixes https://github.com/microsoft/TypeScript/issues/61695.

For more information, see https://github.com/microsoft/TypeScript/issues/61695.


jakebailey:
These files are generated from the submodule and must not be updated by hand. This change will need to flow in through the submodule as we pick up more changes from the main TS repo.

dontwanttothink:
@jakebailey Sorry about that. An equivalent PR exists in the other repository, but hasn't been reviewed since it was submitted in May: https://github.com/microsoft/TypeScript/pull/61696. I believe this feature is now stable across all major browsers. Unfortunately, TypeScript doesn't have type definitions for it. Could you take a look, or maybe help route it to the right person? I would personally be grateful.

PR #1684, opened 2025-09-05
Porting of tsc -b -w

sheetalkamat:

PR #1686, opened 2025-09-05
Use package.json locations to determine uptodate ness

sheetalkamat:
This is port of https://github.com/microsoft/TypeScript/pull/62405

And on top of --watch mode PR https://github.com/microsoft/typescript-go/pull/1684

### DIFF
--- a/internal/collections/syncmap.go
+++ b/internal/collections/syncmap.go
@@ -16,11 +16,7 @@ func (s *SyncMap[K, V]) Load(key K) (value V, ok bool) {
 	if !ok {
 		return value, ok
 	}
-	if value, ok := val.(V); ok {
-		return value, true
-	}
-	var zero V
-	return zero, true
+	return val.(V), true
 }
 
 func (s *SyncMap[K, V]) Store(key K, value V) {
--- a/internal/collections/syncmap.go
+++ b/internal/collections/syncmap.go
@@ -16,7 +16,11 @@ func (s *SyncMap[K, V]) Load(key K) (value V, ok bool) {
 	if !ok {
 		return value, ok
 	}
-	return val.(V), true
+	if value, ok := val.(V); ok {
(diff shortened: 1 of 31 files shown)


internal/collections/syncmap.go
```diff
@@ -14,7 +14,11 @@ func (s *SyncMap[K, V]) Load(key K) (value V, ok bool) {
 	if !ok {
 		return
 	}
-	return val.(V), true
+	if value, ok := val.(V); ok {
+		return value, true
+	}
+	var zero V
+	return zero, true
```
jakebailey: This type assertion failing should be impossible; the only way to store a value into the map is to have passed in a `V`, which is enforced by the generics, no?

The only way that I can see this failing is if we are accidentally converting one SyncMap to another, which, I accidentally think is possible because I forgot to  mention `K` and `V` in the struct...

sheetalkamat: its happening because stored stat = nil for missing file = say using FileExists.. and we query stats for same file to get its modified time

jakebailey: Are you sure the issue isn't somewhere else where we shouldn't be boxing a nil pointer into an interface? I'm just not seeing how it's possible for this to crash at runtime at all

sheetalkamat: We are calling file exists on package .json which caches nil as stat for that file .. then to watch that location we get stats and record its mTime which gets you the cached stat that’s nil

jakebailey: Sorry, I'm not understanding... The only way to get a value into `m` is via `Store` or `LoadOrStore`, both of which only accept a `V`. How can the underlying value possibly be anything other than `V`?

sheetalkamat: Hmm not sure what fixed this but i dont need this change any more.

PR #1688, opened 2025-09-06
fix: refine export assignment diagnostics to match strada

a-tarasyuk:


testdata/baselines/reference/submodule/conformance/jsDeclarationsFunctionPrototypeStatic.js.diff
```diff
@@ -31,4 +31,30 @@
 + *
 + * @callback DoneCB
 + * @param {number} failures - Number of failures that occurred.
-+ */ 
\ No newline at end of file
++ */ 
++
++
++//// [DtsFileErrors]
```
jakebailey: This one's a regression.

jakebailey:
@sandersn @weswigham Is this the thing we were discussing about relaxing this for even TS code and not erroring at all? I can't remember where we left that.

testdata/baselines/reference/submodule/conformance/jsDeclarationsFunctionPrototypeStatic.js.diff
a-tarasyuk: @jakebailey Thanks for pointing this out. I’ll check — seems `IsInJSFile` is false in the declaration emit...

a-tarasyuk: @jakebailey This may not be directly related to the change; it seems the issue is that the jsdoc callback is emitted as exported in the `.d.ts` files.

https://github.com/microsoft/typescript-go/blob/d2c442de2766b7b99af66049f57cce6a6e856701/internal/parser/reparser.go#L73


```
export = MyClass;
....
export type DoneCB = (failures: number) ;
```

while `strada` handles this case as 

```
export = MyClass;
....
type DoneCB = (failures: number) => any;
```
I think the way to address this regression is to elide the export modifier. WDYT?

jakebailey: That sounds plausible. Thankfully I believe `ExternalModuleIndicator` is set before reparsing.

a-tarasyuk: It looks like `ExternalModuleIndicator` is set only after parsing top-level statements

https://github.com/microsoft/typescript-go/blob/c0aa663075218bac7bb9e42625d169e0b1231f59/internal/parser/parser.go#L350

https://github.com/microsoft/typescript-go/blob/c0aa663075218bac7bb9e42625d169e0b1231f59/internal/parser/parser.go#L379

weswigham:
We're at the "just do it like strada" stage because trying to improve it just moved where the jank in our code is - better the jank we know.

testdata/baselines/reference/submodule/conformance/jsDeclarationsFunctionPrototypeStatic.js.diff
weswigham: Yeah, the declaration emitter or reparser definitely needs to be adjusted to handle this. Probably the reparser, in this case - it probably shouldn't add an `export` modifier if there's an export assignment in the file. Truth be told, though, this is one of those cases where we can't accurately represent the input as a declaration file, and no matter what we do we're probably going to change behavior from strada.

testdata/baselines/reference/submodule/conformance/jsDeclarationsFunctionPrototypeStatic.js
```diff
@@ -34,10 +34,33 @@ MyClass.staticProperty = 123;
...
+
+//// [DtsFileErrors]
+
+
+out/source.d.ts(1,10): error TS2304: Cannot find name 'MyClass'.
+out/source.d.ts(2,34): error TS1005: '=>' expected.
+
+
+==== out/source.d.ts (2 errors) ====
+    export = MyClass;
```
weswigham: Hm. This is because we don't actually support constructor functions yet, right @sandersn ?

weswigham (approved):
Definitely an improvement. I'm sure we'll have to keep adjusting the JS declaration emit for export assignments, though, since they're still only partially supported.

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypeAliases.js.diff
```diff
@@ -67,8 +67,11 @@
 + * @returns {T}
 + */
  //// [mixed.d.ts]
- export type SomeType = {
+-export type SomeType = {
++type SomeType = {
```
jakebailey: This one is regressing; maybe a bad interplay with the new code?

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypedefAndImportTypes.js.diff
```diff
@@ -41,9 +40,11 @@
 -declare namespace Conn {
 -    export { Whatever };
 -}
--type Whatever = string | number;
+ type Whatever = string | number;
++export = Conn;
  //// [usage.d.ts]
- export type Conn = import("./conn");
+-export type Conn = import("./conn");
++type Conn = import("./conn");
```
jakebailey: This one too? (I hate JSDoc typedef rules)

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypedefPropertyAndExportAssignment.js.diff
```diff
@@ -28,8 +28,9 @@
 
  //// [module.d.ts]
 -export type TaskGroupIds = "parseHTML" | "styleLayout";
-+export type TaskGroupIds = 'parseHTML' | 'styleLayout';
- export type TaskGroup = {
+-export type TaskGroup = {
++type TaskGroupIds = 'parseHTML' | 'styleLayout';
++type TaskGroup = {
```
jakebailey: This file's also regressing in a similar way, it seems...

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypeAliases.js.diff
a-tarasyuk: This is somewhat of a regression, related to how Corsa handles the `module.exports = {}` export assignment, for instance 

```js
/**
 * @typedef {{x: string} | number | LocalThing | ExportedThing} SomeType
 */
/**
 * @param {number} x
 * @returns {SomeType}
 */
function doTheThing(x) {
    return {x: ""+x};
}
class ExportedThing {
    z = "ok"
}
module.exports = {
    doTheThing,
    ExportedThing,
};
class LocalThing {
    y = "ok"
}
```

Strada

```js
export type SomeType = {
    x: string;
} | number | LocalThing | ExportedThing;
/**
 * @typedef {{x: string} | number | LocalThing | ExportedThing} SomeType
 */
/**
 * @param {number} x
 * @returns {SomeType}
 */
declare function doTheThing(x: number): SomeType;
declare class ExportedThing {
    z: string;
}
declare class LocalThing {
    y: string;
}
export {};
```

Corsa
```js
//// [mixed.d.ts]
type SomeType = {
    x: string;
} | number | LocalThing | ExportedThing;
/**
 * @typedef {{x: string} | number | LocalThing | ExportedThing} SomeType
 */
/**
 * @param {number} x
 * @returns {SomeType}
 */
declare function doTheThing(x: number): SomeType;
declare class ExportedThing {
    z: string;
}
declare const _default: {
    doTheThing: typeof doTheThing;
    ExportedThing: typeof ExportedThing;
};
export = _default;
declare class LocalThing {
    y: string;
}
```

https://github.com/microsoft/typescript-go/blob/0522ac5164677f505673f8757c7d90e48175f572/internal/transformers/declarations/transform.go#L916-L939

Since `export = _default;` is present, all other exports should be elided to avoid mismatches related to this error handling. 

I think fixing this regression requires changing how export assignments are handled, so cases like

```js
module.exports = {
  doTheThing,
  ExportedThing,
};
```

are transformed without emitting them as a _default export_. @jakebailey WDYT?

jakebailey: That sounds about right, I think? (Not personally an expert in this crossover)

a-tarasyuk: @weswigham @jakebailey I’m wondering if, instead of replicating Strada’s JavaScript declaration emitter, which produces 

```js
// @input: a.js
/**
 * @typedef {string | number} T
 */

class A { }
module.exports = {
    A,
}

// @output: a.d.ts
export type T = string | number;
/**
 * @typedef {string | number} T
 */
export class A {
}
```

it might make more sense to align more closely with what TypeScript’s transformer does, and adjust the output to something like this...

```ts
// @output: a.d.ts
declare class A {}

declare const _default: {
  A: typeof A;
};
export = _default;

declare namespace _default {
  type T = string | number;
}
```

jakebailey: Wouldn't the impact of that be that `A` is inaccessible from type space?

jakebailey:
Why the deletion? ☹️

PR #1692, opened 2025-09-08
Limit number of projects that are built concurrently

sheetalkamat:
Open to option name suggestions

Name of the flag?
Default value?


ghiscoding:
I'm not sure if that helps or not, but in Lerna/Lerna-Lite when running `lerna run` (or `lerna exec`) in a monorepo project, you could specify a custom `--concurrency` option in the CLI and its default value is:
```ts
import { cpus } from 'node:os';
const DEFAULT_CONCURRENCY = cpus().length;
const concurrency = Math.max(1, +concurrency || DEFAULT_CONCURRENCY);
```

jakebailey:
There's an equivalent in Go, but I think we need to come up with some larger plan on how to limit resource usage; this may not be the right way to do things given there's parallelism inside projects such that we might need to instead be limiting the number of concurrent checks, or some other gate.

PR #1694, opened 2025-09-09
Fix last character of JSX text nodes being cut off

bradchristensen:
Minimal repro, in test case format:

```tsx
// @strict: true
// @target: esnext
// @noEmit: false
// @jsx: react-jsx

/// <reference path="/.lib/react16.d.ts" />

<div>
    Foo
</div>;
```

```
import { jsx as _jsx } from "react/jsx-runtime";
/// <reference path="react16.d.ts" />
_jsx("div", { children: "Foo" });
```

## Behavior with `tsgo`

"Foo" is truncated to "Fo":

```
import { jsx as _jsx } from "react/jsx-runtime";
/// <reference path="react16.d.ts" />
_jsx("div", { children: "Fo" });
```


bradchristensen:

internal/transformers/jsxtransforms/jsx.go
```diff
@@ -794,15 +794,15 @@ func fixupWhitespaceAndDecodeEntities(text string) string {
...
+	// firstNonWhitespace = 0 to indicate that we want leading whitespace,
 	// but lastNonWhitespace = -1 as a special flag to indicate that we *don't* include the line if it's all whitespace.
 	for i := 0; i < len(text); i++ {
 		c, size := utf8.DecodeRuneInString(text[i:])
 		if stringutil.IsLineBreak(c) {
 			// If we've seen any non-whitespace characters on this line, add the 'trim' of the line.
 			// (lastNonWhitespace === -1 is a special flag to detect whether the first line is all whitespace.)
 			if firstNonWhitespace != -1 && lastNonWhitespace != -1 {
-				addLineOfJsxText(acc, text[firstNonWhitespace:lastNonWhitespace], initial)
+				addLineOfJsxText(acc, text[firstNonWhitespace:lastNonWhitespace+1], initial)
```
jakebailey, line 805: Yep, original code was:

```
acc = addLineOfJsxText(acc, text.substr(firstNonWhitespace, lastNonWhitespace - firstNonWhitespace + 1));
```

PR #1695, opened 2025-09-09
Fix HTML entity handling in JSX transformer

bradchristensen:
This PR fixes the JSX transformer incorrectly emitting HTML entities as-is (as plaintext) when they're supposed to be converted into the characters they encode (e.g. `&amp;` should be replaced by `&`, `&apos;` should be replaced by `'` and so on).


bradchristensen:

gabritto (approved):

PR #1699, opened 2025-09-09
Document highlights

johnfav03:


johnfav03:
> @johnfav03 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

internal/fourslash/_scripts/failingTests.txt
andrewbranch, line 1: Is there a pattern among the failing tests, i.e. something they’re blocked on or a specific follow-up task?

internal/fourslash/tests/documentHighlight_test.go
andrewbranch, line 1: Did you meant to leave this one in (is it not redundant with the generated ones)?

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,718 @@
...
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+)
+
+func (l *LanguageService) ProvideDocumentHighlights(ctx context.Context, documentUri lsproto.DocumentUri, documentPosition lsproto.Position) (lsproto.DocumentHighlightResponse, error) {
+	program, sourceFile := l.getProgramAndFile(documentUri)
+	position := int(l.converters.LineAndCharacterToPosition(sourceFile, documentPosition))
+	node := astnav.GetTouchingPropertyName(sourceFile, position)
+
+	if node.Parent.Kind == ast.KindJsxOpeningElement || (node.Parent.Kind == ast.KindJsxClosingElement && node.Parent.TagName() == node) {
+		documentHighlights := []*lsproto.DocumentHighlight{}
```
andrewbranch: Tiny nit: we typically use `var documentHighlights []*lsproto.DocumentHighlight` if we don’t know the length/capacity up front, or `make(...)` if we do. (In other words, empty slices are rare; we usually have a nil slice to represent no items.)

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,718 @@
...
+				Kind:  &kind,
+			})
+		}
+		return lsproto.DocumentHighlightsOrNull{
+			DocumentHighlights: &documentHighlights,
+		}, nil
+	}
+
+	documentHighlights := l.getSemanticDocumentHighlights(ctx, position, node, program, sourceFile)
+	if documentHighlights == nil {
```
andrewbranch: In a similar vein, checking slice quality to `nil` is unusual; we’d usually use `len(documentHighlights) == 0`. If you actually needed to differentiate between “no results but return now” and “no results and look for syntactic highlights,” you should use multiple return values (e.g. `([]*lsproto.DocumentHighlight, bool)` for that instead of the nil/empty distinction—however, glancing at `getSemanticDocumentHighlights`, it doesn’t look like you’re trying to draw that distinction, so I think you can just convert this to a length check with no other changes. It’s just a way of reassuring future readers that you’re _not_ trying to draw a spooky distinction between nil and empty slice.

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,718 @@
...
+	kind := lsproto.DocumentHighlightKindText
+	if entry.kind == entryKindRange {
+		return entry.fileName, &lsproto.DocumentHighlight{
+			Range: *entry.textRange,
+			Kind:  &kind,
+		}
+	}
+
+	// Determine write access for node references.
+	if checker.IsWriteAccess(entry.node) {
```
andrewbranch: I think `IsWriteAccess` and its related types and functions probably should move to the `ast` package now that they’re exported. They’re not really checker-related.

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,718 @@
...
+				Kind:  &kind,
+			})
+		}
+	}
+	return highlights
+}
+
+func (l *LanguageService) getFromAllDeclarations(nodeTest func(*ast.Node) bool, keywords []ast.Kind, node *ast.Node, sourceFile *ast.SourceFile) []*lsproto.DocumentHighlight {
+	return l.useParent(node.Parent, nodeTest, func(decl *ast.Node, sf *ast.SourceFile) []*ast.Node {
+		symbolDecls := []*ast.Node{}
```
andrewbranch: Same nit about slice initialization here

andrewbranch (commented):
This is a great first PR 👏

internal/fourslash/tests/documentHighlight_test.go
johnfav03: Great catch, I'll remove this one!

internal/fourslash/_scripts/failingTests.txt
johnfav03: These tests are giving an incorrect output due to a missing case in `getReferencedSymbolsForNode()` in `findAllReferences` - I'll update the PR description with cases that could be updated as a follow-up task

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -778,6 +780,90 @@ function parseBaselineFindAllReferencesArgs(args: readonly ts.Expression[]): [Ve
...
+    }
+
+    return [{
+        kind: "verifyBaselineDocumentHighlights",
+        args: newArgs,
+        preferences: preferences ? preferences : "nil /*preferences*/",
+    }];
+}
+
+function parseBaselineDocumentHighlightsArg(arg: ts.Expression): string | undefined {
```
iisaduan: With the exception of the error message, this function is identical to `parseBaselineRenameArg`, and essentially parses parameters of type `markerOrRange?: ArrayOrSingle<MarkerOrNameOrRange>`, which happens a lot in strada fourslash. We should refactor so that we only have a single function that parses parameters of that type/kind, because we'll definitely be using it more in other functions

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -778,6 +780,90 @@ function parseBaselineFindAllReferencesArgs(args: readonly ts.Expression[]): [Ve
...
+            for (const elem of strArg.elements) {
+                const newArg = parseBaselineDocumentHighlightsArg(elem);
+                if (!newArg) {
+                    return undefined;
+                }
+                newArgs.push(newArg);
+            }
+        }
+        else if (ts.isObjectLiteralExpression(arg)) {
+            // User preferences case, but multiple files isn't implemented in corsa yet
```
iisaduan: What do you mean by "multiple files isn't implemented in corsa yet"? Does this fourslash function need different options per file?

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+)
+
+func (l *LanguageService) ProvideDocumentHighlights(ctx context.Context, documentUri lsproto.DocumentUri, documentPosition lsproto.Position) (lsproto.DocumentHighlightResponse, error) {
+	program, sourceFile := l.getProgramAndFile(documentUri)
+	position := int(l.converters.LineAndCharacterToPosition(sourceFile, documentPosition))
+	node := astnav.GetTouchingPropertyName(sourceFile, position)
+
+	if node.Parent.Kind == ast.KindJsxOpeningElement || (node.Parent.Kind == ast.KindJsxClosingElement && node.Parent.TagName() == node) {
```
iisaduan: The `node.Parent != nil` check is still needed because this will panic when `GetTouchingPropertyName` returns `ast.SourceFile` (can happen when document highlights is requested in an empty file, at EOF and possibly other situations as well)

iisaduan: Also, is the change intentional? the new code isn't identical to the logic in strada
```suggestion
	if node.Parent.Kind == ast.KindJsxClosingElement || (node.Parent.Kind == ast.KindJsxOpeningElement && node.Parent.TagName() == node) {
```

iisaduan: Thought about it a little more, and yeah this needs to be changed to match with strada because we don't want to do this jsx-specific behavior if we're in the opening tag but _not_ in the name of the tag

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+				Range: *l.createLspRangeFromNode(node.Parent, sourceFile),
+				Kind:  &kind,
+			})
+		}
+		if node.Parent.Kind == ast.KindJsxClosingElement {
+			documentHighlights = append(documentHighlights, &lsproto.DocumentHighlight{
+				Range: *l.createLspRangeFromNode(node.Parent, sourceFile),
+				Kind:  &kind,
+			})
+		}
```
iisaduan, line 40: This block isn't quite right--a `node` can only ever have a single `Kind`, so this code would only ever add one highlight to the list of highlights. I like the change in refactor to not use `.map`, but the easiest way to get both opening and closing element nodes is way done in strada

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	for _, entry := range referenceEntries {
+		for _, ref := range entry.references {
+			if ref.node != nil {
+				fileName, highlight := l.toDocumentHighlight(ref)
+				if fileName == sourceFile.FileName() {
+					highlights = append(highlights, highlight)
+				}
+			}
```
iisaduan, line 67:
```suggestion
			if ref.fileName == sourceFile.FileName() && ref.node != nil {
    			highlights = append(highlights, l.toDocumentHighlight(ref))
			}
```
`resolveEntry` can do a lot of work, so we should filter on `filename` earlier, since it's already accessible through `ref`, and remove the `fileName` return from `toDocumentHighlight`

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	kind := lsproto.DocumentHighlightKindText
+	if entry.kind == entryKindRange {
+		return entry.fileName, &lsproto.DocumentHighlight{
+			Range: *entry.textRange,
+			Kind:  &kind,
+		}
+	}
+
+	// Determine write access for node references.
+	if ast.IsWriteAccess(entry.node) {
```
iisaduan: What's the difference in the change from `IsWriteAccessForReference` to `IsWriteAccess`?

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+			}
+		}
+	}
+	return highlights
+}
+
+func (l *LanguageService) toDocumentHighlight(entry *referenceEntry) (string, *lsproto.DocumentHighlight) {
+	entry = l.resolveEntry(entry)
+
+	// If this is a plain range (Span), always treat it as a reference.
```
iisaduan: what does this mean?

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	kind := lsproto.DocumentHighlightKindText
```
iisaduan: Looks like the lsp conversion means `HighlightSpanKind.reference` becomes `lsproto.DocumentHighlightKindRead` https://github.com/microsoft/vscode/blob/07da241ca964f5bb4e618c68c828fcba815e1045/extensions/typescript-language-features/src/languageFeatures/documentHighlight.ts#L72
```suggestion
	kind := lsproto.DocumentHighlightKindRead
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	if ast.IsWriteAccess(entry.node) {
+		kind = lsproto.DocumentHighlightKindWrite
+	}
+
+	dh := &lsproto.DocumentHighlight{
+		Range: *entry.textRange,
+		Kind:  &kind,
+	}
+
+	return entry.fileName, dh
```
iisaduan, line 94:
```suggestion
	return &lsproto.DocumentHighlight{
		Range: *entry.textRange,
		Kind:  &kind,
	}
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+}
+
+func (l *LanguageService) getSyntacticDocumentHighlights(node *ast.Node, sourceFile *ast.SourceFile) []*lsproto.DocumentHighlight {
+	switch node.Kind {
+	case ast.KindIfKeyword, ast.KindElseKeyword:
+		if ast.IsIfStatement(node.Parent) {
+			result := l.getIfElseOccurrences(node.Parent.AsIfStatement(), sourceFile)
+			return result
```
iisaduan:
```suggestion
			return l.getIfElseOccurrences(node.Parent.AsIfStatement(), sourceFile)
```
nit

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+								if c.Kind == k {
+									symbolDecls = append(symbolDecls, c)
+									break outer
+								}
+							}
+						}
+					}
+				}
+			}
+		}
```
iisaduan, line 192: simplification of this block; also removes redundant check
```suggestion
		if ast.CanHaveSymbol(decl) {
			if symbol := decl.Symbol() && symbol != nil {
				for _, d := range symbol.Declarations {
					if nodeTest(d) {
					outer:
						for _, c := range getChildrenFromNonJSDocNode(d, sourceFile) {
							for _, k := range keywords {
								if c.Kind == k {
									symbolDecls = append(symbolDecls, c)
									break outer
								}
							}
						}
					}
				}
			}
		}
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+					outer:
+						for _, c := range getChildrenFromNonJSDocNode(d, sourceFile) {
+							for _, k := range keywords {
+								if c.Kind == k {
+									symbolDecls = append(symbolDecls, c)
+									break outer
+								}
+							}
+						}
+					}
```
iisaduan, line 189: can this be done with `findChildOfKind`? That's usually what we do to replace calls of `find(getChildren(...))`

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+	var keywords []*ast.Node
+	modifiers := parentFunc.Modifiers().Nodes
+	if len(modifiers) != 0 {
+		for _, modifier := range modifiers {
+			if modifier.Kind == ast.KindAsyncKeyword {
+				keywords = append(keywords, modifier)
+			}
+		}
+	}
```
iisaduan, line 567: can `parentFunc.Modifiers()` be nil? its safe to `range` over nil lists, so the check should actually be this:
```suggestion
	var keywords []*ast.Node
	modifiers := parentFunc.Modifiers()
	if modifiers != nil {
		for _, modifier := range modifiers.Nodes {
			if modifier.Kind == ast.KindAsyncKeyword {
				keywords = append(keywords, modifier)
			}
		}
	}
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	cb(node)
+	if !ast.IsFunctionLike(node) && !ast.IsClassLike(node) && !ast.IsInterfaceDeclaration(node) && !ast.IsModuleDeclaration(node) && !ast.IsTypeAliasDeclaration(node) && !ast.IsTypeNode(node) {
+		node.ForEachChild(func(child *ast.Node) bool {
+			traverseWithoutCrossingFunction(child, sourceFile, cb)
+			return false // continue traversal
+		})
+	}
+}
+
+func getModifierOccurrences(kind ast.Kind, node *ast.Node, sourceFile *ast.SourceFile) []*ast.Node { /* TODO */
```
iisaduan:
```suggestion
func getModifierOccurrences(kind ast.Kind, node *ast.Node, sourceFile *ast.SourceFile) []*ast.Node {
```

old comment?

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+		}
+	default:
+		// Syntactically invalid positions or unsupported containers
+		return nil
+	}
+
+	return result
+}
+
+func modifierToFlag(kind ast.Kind) ast.ModifierFlags {
```
iisaduan: duplicate function, already exists in `ast/utilities.go`
```suggestion
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+			result = append(result, modifier)
+		}
+	}
+	return result
+}
+
+func getNodesToSearchForModifier(declaration *ast.Node, modifierFlag ast.ModifierFlags) []*ast.Node {
+	var result []*ast.Node
+
+	container := declaration.Parent
```
iisaduan, line 633: I think copying the comments over from the previous code would be nice, so that every case is documented now

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+	switch container.Kind {
+	case ast.KindModuleBlock, ast.KindSourceFile, ast.KindBlock, ast.KindCaseClause, ast.KindDefaultClause:
+		// If abstract modifier and class declaration, include members and the declaration itself
+		if (modifierFlag&ast.ModifierFlagsAbstract) != 0 && ast.IsClassDeclaration(declaration) {
+			result = append(result, declaration)
+			result = append(result, declaration.Members()...)
+		} else {
+			result = append(result, container.Statements()...)
+		}
```
iisaduan, line 646:
```suggestion
		if (modifierFlag&ast.ModifierFlagsAbstract) != 0 && ast.IsClassDeclaration(declaration) {
			return append(append(result, declaration.Members()...), declaration)
		} else {
			return container.Statements()
		}
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+				result = append(result, constructor.Parameters()...)
+			} else {
+				result = append(result, nodes...)
+			}
+		} else if (modifierFlag & ast.ModifierFlagsAbstract) != 0 {
+			result = append(result, nodes...)
+			result = append(result, container)
+		} else {
+			result = append(result, nodes...)
+		}
```
iisaduan, line 673:
```suggestion
		nodes := container.Members()
        result = append(result, nodes...)
		if (modifierFlag & (ast.ModifierFlagsAccessibilityModifier | ast.ModifierFlagsReadonly)) != 0 {
			var constructor *ast.Node
			for _, member := range nodes {
				if ast.IsConstructorDeclaration(member) {
					constructor = member
					break
				}
			}
			if constructor != nil {
				result = append(result, constructor.Parameters()...)
			}
		} else if (modifierFlag & ast.ModifierFlagsAbstract) != 0 {
			result = append(result, container)
		}
        return nodes
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+func findModifier(node *ast.Node, kind ast.Kind) *ast.Node {
+	if ast.CanHaveModifiers(node) && node.Modifiers() != nil {
+		for _, modifier := range node.Modifiers().Nodes {
+			if modifier.Kind == kind {
+				return modifier
+			}
+		}
+	}
+	return nil
```
iisaduan: nit: suggestion due to change in how `.Modifiers()` is implmeneted
```suggestion
	if modifiers := node.Modifiers(); modifiers != nil {
		for _, modifier := range modifiers.Nodes {
			if modifier.Kind == kind {
				return modifier
			}
		}
	}
	return nil
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+		if owner.Label().Text() == labelName {
+			return ast.FindAncestorTrue
+		}
+		return ast.FindAncestorFalse
+	}) != nil
+}
+
+func getBreakOrContinueStatementOccurrences(node *ast.Node, sourceFile *ast.SourceFile) []*ast.Node {
+	owner := getBreakOrContinueOwner(node)
+	if owner != nil {
```
iisaduan:
```suggestion
	if owner := getBreakOrContinueOwner(node); owner != nil {
```
nit/go convention

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+func getBreakOrContinueOwner(statement *ast.Node) *ast.Node {
+	// Walk up ancestors to find the owner node.
+	return ast.FindAncestorOrQuit(statement, func(node *ast.Node) ast.FindAncestorResult {
+		switch node.Kind {
+		case ast.KindSwitchStatement:
+			if statement.Kind == ast.KindContinueStatement {
+				return ast.FindAncestorFalse
+			}
+			// falls through
```
iisaduan:
```suggestion
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+			// Don't cross function boundaries.
+			if ast.IsFunctionLike(node) {
+				return ast.FindAncestorQuit
+			}
+			return ast.FindAncestorFalse
+		}
+	})
+}
+
+// Helper function to check if a node is labeled by a given label name.
```
iisaduan:
```suggestion
// Whether or not a 'node' is preceded by a label of the given string.
// Note: 'node' cannot be a SourceFile.
```
bring over old comments where possible/if still applicable

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+func ownsBreakOrContinueStatement(owner *ast.Node, statement *ast.Node) bool {
+	actualOwner := getBreakOrContinueOwner(statement)
+	if actualOwner == nil {
+		return false
+	}
+	return actualOwner == owner
+}
+
+func getBreakOrContinueOwner(statement *ast.Node) *ast.Node {
+	// Walk up ancestors to find the owner node.
```
iisaduan:
```suggestion
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	// Aggregate all throw statements "owned" by this owner.
+	throwStatements := aggregateOwnedThrowStatements(owner, sourceFile)
+	for _, throw := range throwStatements {
+		keyword := findChildOfKind(throw, ast.KindThrowKeyword, sourceFile)
+		if keyword != nil {
+			keywords = append(keywords, keyword)
+		}
+	}
+
+	// If the owner is a function block, also include return keywords.
```
iisaduan:
```suggestion
    // If the "owner" is a function, then we equate 'return' and 'throw' statements in their
    // ability to "jump out" of the function, and include occurrences for both.
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+			}
+			return false // continue traversal
+		})
+	}
+
+	return keywords
+}
+
+func getThrowStatementOwner(throwStatement *ast.Node) *ast.Node {
+	child := throwStatement
```
iisaduan, line 377:
```suggestion

// For lack of a better name, this function takes a throw statement and returns the
// nearest ancestor that is a try-block (whose try statement has a catch clause),
// function-block, or source file.
func getThrowStatementOwner(throwStatement *ast.Node) *ast.Node {
	child := throwStatement
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+
+	return keywords
+}
+
+func aggregateAllBreakAndContinueStatements(node *ast.Node, sourceFile *ast.SourceFile) []*ast.Node {
+	if ast.IsBreakOrContinueStatement(node) {
+		return []*ast.Node{node}
+	}
+	if ast.IsFunctionLike(node) {
+		return []*ast.Node{}
```
iisaduan:
```suggestion
        return nil
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	clauses := switchStatement.CaseBlock.AsCaseBlock().Clauses
+	for _, clause := range clauses.Nodes {
+		clauseToken := lsutil.GetFirstToken(clause.AsNode(), sourceFile)
+		if clauseToken.Kind == ast.KindCaseKeyword || clauseToken.Kind == ast.KindDefaultKeyword {
+			keywords = append(keywords, clauseToken)
+		}
+
+		breakAndContinueStatements := aggregateAllBreakAndContinueStatements(clause, sourceFile)
+		for _, statement := range breakAndContinueStatements {
+			if ownsBreakOrContinueStatement(switchStatement.AsNode(), statement) && statement.Kind == ast.KindBreakStatement {
```
iisaduan: to avoid unneded calls to `findAncestor`
```suggestion
			if statement.Kind == ast.KindBreakStatement && ownsBreakOrContinueStatement(switchStatement.AsNode(), statement) {
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	if ast.IsTryStatement(node) {
+		var result []*ast.Node
+		// Exceptions thrown within a try block lacking a catch clause are "owned" in the current context.
+		statement := node.AsTryStatement()
+		tryBlock := statement.TryBlock
+		catchClause := statement.CatchClause
+		finallyBlock := statement.FinallyBlock
+
+		if catchClause != nil {
+			result = append(result, aggregateOwnedThrowStatements(catchClause, sourceFile)...)
```
iisaduan: could do this to all the cases so that we don't need to do the work of copying the slice again
```suggestion
			result = aggregateOwnedThrowStatements(catchClause, sourceFile)
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+		} else if tryBlock != nil {
+			result = append(result, aggregateOwnedThrowStatements(tryBlock, sourceFile)...)
```
iisaduan:
```suggestion
			result = aggregateOwnedThrowStatements(tryBlock, sourceFile)
```

internal/ls/documenthighlights.go
```diff
@@ -0,0 +1,717 @@
...
+	}
+	return keywords
+}
+
+func aggregateOwnedThrowStatements(node *ast.Node, sourceFile *ast.SourceFile) []*ast.Node {
+	if ast.IsThrowStatement(node) {
+		return []*ast.Node{node}
+	}
+	if ast.IsTryStatement(node) {
+		var result []*ast.Node
```
iisaduan: I'd move this var declaration (and other similar declarations) to right before the block where it's first written to (this case, the `if catchClause != nil`)

iisaduan (changes requested):
Great work! Seems like you've quickly gotten a good hang of the codebase :D 

Requesting changes because I noticed an incorrect port of some logic, and I also left comments on some refactors and code/comment clean up

internal/fourslash/_scripts/convertFourslash.mts
johnfav03: Document highlights has the option search multiple files for references, but Jake told me to ignore this feature; this else if branch is searching for an input parameter that looks like `{filesToSearch: ...}`

jakebailey: Yeah, LSP does not have the ability to return document highlights outside the file where the request occurred.

jakebailey: (I am not sure the old code to do so even worked in VS Code.)

internal/ls/documenthighlights.go
johnfav03: I don't know if this change is possible; many references from `getReferencedSymbolsForNode` are missing a fileName attribute, so resolving is a necessary step for comparison

iisaduan: Ah, makes sense. This + we don't support multiple files anyways means that we expect to resolve every entry anyways

internal/fourslash/_scripts/convertFourslash.mts
iisaduan: yeah, vscode doesnt request/handle symbols across multiple files ([old vscode extension](https://github.com/microsoft/vscode/blob/07da241ca964f5bb4e618c68c828fcba815e1045/extensions/typescript-language-features/src/languageFeatures/documentHighlight.ts#L61)) and is unused in any strada test, so totally removing/not porting it is probably the way to go

jakebailey: Well, https://github.com/microsoft/vscode/blob/07da241ca964f5bb4e618c68c828fcba815e1045/extensions/typescript-language-features/src/languageFeatures/documentHighlight.ts#L17 did actually do it, but that's not in the LSP. We could still do this in the future. I'm not sure it's worth it at the moment to do multi-file highlights, but we may want it for script tag reasons and for embedded languages and so on. But, needs to go into the LSP.

iisaduan: Ah, makes sense. I wasn't ever able to get it to trigger in vscode so I assumeed it was never fully implemented. We can leave it as a `!!! todo when in lsp` in the ls implementation, and later update to match the format in `vscode.proposed`, but I don't think we need to support it in fourslash conversion if we don't have tests that use it

internal/ls/documenthighlights.go
johnfav03: See other comment  - https://github.com/microsoft/typescript-go/pull/1699#discussion_r2364778236

internal/ls/documenthighlights.go
johnfav03: I'm not sure that this is possible, as this logic finds the first child matching any `Kind` in keywords, whereas `findChildOfKind` only accepts one `Kind` parameter

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -787,17 +787,17 @@ function parseBaselineDocumentHighlightsArgs(args: readonly ts.Expression[]): [V
...
+                const newArg = parseBaselineMarkerOrRangeArg(elem);
                 if (!newArg) {
                     return undefined;
                 }
                 newArgs.push(newArg);
             }
         }
         else if (ts.isObjectLiteralExpression(arg)) {
-            // User preferences case, but multiple files isn't implemented in corsa yet
+            // !! todo when multiple files supported in lsp
```
iisaduan:
```suggestion
            // !!! todo when multiple files supported in lsp/editor

```

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -201,6 +209,14 @@ TestGetJavaScriptQuickInfo6
...
 TestGetJavaScriptQuickInfo8
 TestGetJavaScriptSyntacticDiagnostics24
+TestGetOccurrencesAsyncAwait3
+TestGetOccurrencesClassExpressionConstructor
+TestGetOccurrencesConstructor
+TestGetOccurrencesConstructor2
+TestGetOccurrencesIfElseBroken
+TestGetOccurrencesOfAnonymousFunction
+TestGetOccurrencesStringLiteralTypes
+TestGetOccurrencesStringLiterals
```
jakebailey: Curious, what kinds of failures are these? Crashes? If they're crashes, I would definitely want to double check these because document highlight is something that happens automatically on cursor move, which means we could be spamming panics.

jakebailey: Ah, these must be https://github.com/microsoft/typescript-go/pull/1699/files#r2360590494, which I didn't notice since it was attached to an old rev.

iisaduan: They are because of unimplemented cases in FAR, so the test fails when no highlights are returned (which is fine in editor)

iisaduan: String references and a lot of the keyword references aren't implemented

jakebailey:
I think this looks good, but in `checker.ts` it does look like sometimes things take more time than expected, though it's not clear if that's inherent or just because it was doing a hover or something at the same time:

image

I filtered the LSP output to `Received response 'textDocument/documentHighlight` and then moved my cursor / mouse around.

To be clear, it usually isn't this way, it's just sometimes randomly it takes 100ms, or is cancelled, which is also fine too.

It does make me wonder if there's an implementation of this that could avoid FAR and the checker entirely. Or a FAR modification to say "no checker use" that is mostly accurate. Especially because we are scoping to one file.

jakebailey (approved):

internal/fourslash/_scripts/failingTests.txt
johnfav03: Only two tests fail, and it's due to unexpected responses in `fourslash.go` from modules that arent `documentHighlights`. The tests are `FindAllRefsForModule` and `TestJsdocTypedefTagServices`, which return an unexpected response from `FindAllReferences` and `QuickInfo` respectively.

iisaduan (approved):

PR #1702, opened 2025-09-10
Full JSDoc support in LSP

ahejlsberg:
This PR implements full support for JSDoc in the language service such that hover, find-all-references, go-to-definition, and rename all work properly with JSDoc annotated code (be that in .ts or .js files).


internal/astnav/tokens.go
```diff
@@ -9,9 +9,27 @@ import (
...
-	return getTokenAtPosition(sourceFile, position, false /*allowPositionInLeadingTrivia*/, func(node *ast.Node) bool {
+	return getReparsedNodeForNode(getTokenAtPosition(sourceFile, position, false /*allowPositionInLeadingTrivia*/, func(node *ast.Node) bool {
 		return ast.IsPropertyNameLiteral(node) || ast.IsKeywordKind(node.Kind) || ast.IsPrivateIdentifier(node)
-	})
+	}))
+}
+
+// If the given node is a declaration name node in a JSDoc comment that is subject to reparsing, return the declaration name node
+// for the corresponding reparsed construct. Otherwise, just return the node.
+func getReparsedNodeForNode(node *ast.Node) *ast.Node {
```
jakebailey, line 19: Is this a replacement for the bidirectional mapping plan we had originally conceived?

internal/checker/checker.go
```diff
@@ -30230,6 +30227,14 @@ func (c *Checker) getSymbolAtLocation(node *ast.Node, ignoreErrors bool) *ast.Sy
 			}
 			// no other meta properties are valid syntax, thus no others should have symbols
 			return nil
+		} else if ast.IsJSDocParameterTag(parent) && parent.Name() == node {
```
jakebailey, line 30230: Is it odd to have this one special case here in the checker? I don't think we have any cases like this elsewhere after the reparser change.

internal/checker/checker.go
```diff
@@ -30418,28 +30423,32 @@ func (c *Checker) getSymbolOfNameOrPropertyAccessExpression(name *ast.Node) *ast
 			// Missing entity name.
 			return nil
 		}
-
-		if name.Kind == ast.KindIdentifier {
+		isJSDoc := ast.IsJSDocNameReferenceContext(name)
+		meaning := core.IfElse(isJSDoc, ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace, ast.SymbolFlagsValue)
```
jakebailey, line 30427: This seems like it could be moved into the if body, right?

internal/astnav/tokens.go
ahejlsberg: Yeah, this is reasonably performant and not used in anything time critical. I haven't really seen a need for anything else.

internal/checker/checker.go
ahejlsberg: Yeah.

internal/checker/checker.go
ahejlsberg: We do have a few. We used to have logic in the binder that would bind declaration names in JSDoc nodes, but we don't do that anymore (which is good). The logic also can't be in `getReparsedNodeForNode` because that function is supposed to preserve the location of the node (it basically just returns the cloned copy of an identifier created by re-parse). So, I think it's in the right place.

internal/checker/checker.go
ahejlsberg: Specifically, the `meaning` variable can be moved, but not `isJSDoc`.

PR #1703, opened 2025-09-10
Fix compare LSP positions

gabritto:


gabritto:
Forgot to run `updateFailing` 😬

PR #1705, opened 2025-09-10
Port inlay hints

gabritto:
Waiting on user preferences handling to move forward with this.

### DIFF
--- a/internal/checker/printer.go
+++ b/internal/checker/printer.go
@@ -377,3 +377,8 @@ func (c *Checker) TypeToTypeNode(t *Type, enclosingDeclaration *ast.Node, flags 
 	nodeBuilder := c.getNodeBuilder()
 	return nodeBuilder.TypeToTypeNode(t, enclosingDeclaration, flags, nodebuilder.InternalFlagsNone, nil)
 }
+
+func (c *Checker) TypePredicateToTypePredicateNode(t *TypePredicate, enclosingDeclaration *ast.Node, flags nodebuilder.Flags) *ast.TypePredicateNodeNode {
+	nodeBuilder := c.getNodeBuilder()
+	return nodeBuilder.TypePredicateToTypePredicateNode(t, enclosingDeclaration, flags, nodebuilder.InternalFlagsNone, nil)
+}
--- a/internal/checker/printer.go
+++ b/internal/checker/printer.go
@@ -12,7 +12,7 @@ func createPrinterWithDefaults(emitContext *printer.EmitContext) *printer.Printe
 	return printer.NewPrinter(printer.PrinterOptions{}, printer.PrintHandlers{}, emitContext)
 }
 
-func createPrinterWithRemoveComments(emitContext *printer.EmitContext) *printer.Printer {
+func CreatePrinterWithRemoveComments(emitContext *printer.EmitContext) *printer.Printer {
 	return printer.NewPrinter(printer.PrinterOptions{RemoveComments: true}, printer.PrintHandlers{}, emitContext)
 }
 
@@ -199,7 +199,7 @@ func (c *Checker) typeToStringEx(t *Type, enclosingDeclaration *ast.Node, flags 
 	if t == c.unresolvedType {
 		printer = createPrinterWithDefaults(nodeBuilder.EmitContext())
 	} else {
-		printer = createPrinterWithRemoveComments(nodeBuilder.EmitContext())
+		printer = CreatePrinterWithRemoveComments(nodeBuilder.EmitContext())
 	}
 	var sourceFile *ast.SourceFile
 	if enclosingDeclaration != nil {
@@ -262,7 +262,7 @@ func (c *Checker) symbolToStringEx(symbol *ast.Symbol, enclosingDeclaration *ast
 	if enclosingDeclaration != nil && enclosingDeclaration.Kind == ast.KindSourceFile {
 		printer_ = createPrinterWithRemoveCommentsNeverAsciiEscape(nodeBuilder.EmitContext())
 	} else {
-		printer_ = createPrinterWithRemoveComments(nodeBuilder.EmitContext())
+		printer_ = CreatePrinterWithRemoveComments(nodeBuilder.EmitContext())
 	}
 
 	var builder func(symbol *ast.Symbol, meaning ast.SymbolFlags, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node
@@ -325,7 +325,7 @@ func (c *Checker) typePredicateToStringEx(typePredicate *TypePredicate, enclosin
 	nodeBuilder := c.getNodeBuilder()
 	combinedFlags := toNodeBuilderFlags(flags) | nodebuilder.FlagsIgnoreErrors | nodebuilder.FlagsWriteTypeParametersInQualifiedName
 	predicate := nodeBuilder.TypePredicateToTypePredicateNode(typePredicate, enclosingDeclaration, combinedFlags, nodebuilder.InternalFlagsNone, nil) // TODO: GH#18217
-	printer_ := createPrinterWithRemoveComments(nodeBuilder.EmitContext())
+	printer_ := CreatePrinterWithRemoveComments(nodeBuilder.EmitContext())
 	var sourceFile *ast.SourceFile
 	if enclosingDeclaration != nil {
 		sourceFile = ast.GetSourceFileOfNode(enclosingDeclaration)
--- a/internal/ls/inlay_hints.go
+++ b/internal/ls/inlay_hints.go
@@ -1,0 +1,855 @@
+package ls
+
+import (
+	"context"
+	"slices"
+	"strings"
+	"unicode"
+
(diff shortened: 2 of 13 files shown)


internal/checker/nodebuilderimpl.go
```diff
@@ -3066,13 +3066,13 @@ func (b *NodeBuilderImpl) typeToTypeNode(t *Type) *ast.TypeNode {
 	if t.flags&TypeFlagsTemplateLiteral != 0 {
 		texts := t.AsTemplateLiteralType().texts
 		types := t.AsTemplateLiteralType().types
-		templateHead := b.f.NewTemplateHead(texts[0], texts[0], ast.TokenFlagsNone)
+		templateHead := b.f.NewTemplateHead(texts[0], "", ast.TokenFlagsNone)
```
jakebailey, line 3067: This is fun, maybe we should just quick pull this into main?

internal/ls/inlay_hints.go
```diff
@@ -0,0 +1,935 @@
...
+		if ast.IsLiteralExpression(node) {
+			parts = append(parts, &lsproto.InlayHintLabelPart{Value: s.getLiteralText(node)})
+			return
+		}
+
+		switch node.Kind {
+		case ast.KindIdentifier:
+			identifierText := node.Text()
+			var name *ast.Node
+			// !!! This won't work in Corsa since we don't store symbols on identifiers. We need another strategy for it.
```
gabritto, line 441: The effect of this is that you can't hover over an identifier in an inlay hint in the editor. But I'm leaving this to be fixed later, because it seems more important to have inlay hints working in the editor now, and it could take some non-trivial work to fix this, e.g. constructing a map of identifiers to symbols in the node builder.

testdata/submoduleAccepted.txt
```diff
@@ -1,3 +1,25 @@
 # Diff files to instead write to submoduleAccepted as "accepted" changes.
 conformance/node10Alternateresult_noTypes.errors.txt.diff
-conformance/node10AlternateResult_noResolution.errors.txt.diff
\ No newline at end of file
+conformance/node10AlternateResult_noResolution.errors.txt.diff
+fourslash/inlayHints/inlayHintsInteractiveParameterNamesWithComments.baseline.diff
```
gabritto, line 4: This diff is because I fixed a bug from Strada where we didn't fully detect if a comment had the parameter name.

testdata/submoduleAccepted.txt
```diff
@@ -1,3 +1,25 @@
...
+fourslash/inlayHints/inlayHintsImportType2.baseline.diff
+fourslash/inlayHints/inlayHintsInteractiveImportType2.baseline.diff
```
gabritto, line 6: These two are due to a good change in the checker when we check JavaScript expandos: they used to be checked in a special way, but are now checked the same as TS expressions.

testdata/submoduleAccepted.txt
```diff
@@ -1,3 +1,25 @@
...
+fourslash/inlayHints/inlayHintsFunctionParameterTypes1.baseline.diff
```
gabritto, line 7: Diffs from here on are due to removing the `interactiveInlayHints` option, which is now always true in Corsa.

internal/ls/inlay_hints.go
```diff
@@ -27,6 +27,9 @@ func (l *LanguageService) ProvideInlayHint(
 ) (lsproto.InlayHintResponse, error) {
 	program, file := l.getProgramAndFile(params.TextDocument.Uri)
 	quotePreference := getQuotePreference(file, l.UserPreferences())
+	if !isAnyInlayHintEnabled(l.UserPreferences()) {
+		return lsproto.InlayHintsOrNull{InlayHints: nil}, nil
+	}
```
jakebailey, line 30: This could probably get moved a couple of lines up.

(setting the field to nil is also redundant, but that's ok)

internal/checker/printer.go
```diff
@@ -12,7 +12,7 @@ func createPrinterWithDefaults(emitContext *printer.EmitContext) *printer.Printe
 	return printer.NewPrinter(printer.PrinterOptions{}, printer.PrintHandlers{}, emitContext)
 }
 
-func createPrinterWithRemoveComments(emitContext *printer.EmitContext) *printer.Printer {
+func CreatePrinterWithRemoveComments(emitContext *printer.EmitContext) *printer.Printer {
```
jakebailey: Probably these helpers are in the wrong package, then?

testdata/baselines/reference/submoduleAccepted/fourslash/inlayHints/inlayHintsFunctionParameterTypes1.baseline.diff
```diff
@@ -0,0 +1,230 @@
+--- old.inlayHintsFunctionParameterTypes1.baseline
++++ new.inlayHintsFunctionParameterTypes1.baseline
+@@= skipped -1, +1 lines =@@
+ const f1: F1 = (a, b) => { }
+                  ^
+ {
+-  "label": ": string",
++  "label": [
```
jakebailey, line 8: For baselining purposes, should we convert these to plain strings?

testdata/baselines/reference/submoduleAccepted/fourslash/inlayHints/inlayHintsJsDocParameterNames.baseline.diff
```diff
@@ -0,0 +1,36 @@
...
++++ new.inlayHintsJsDocParameterNames.baseline
+@@= skipped -1, +1 lines =@@
+ y.foo(1, 2)
+       ^
+ {
+-  "label": "a:",
++  "label": [
++    {
++      "value": "a",
++      "location": {},
```
jakebailey, line 11: Location is empty in baselines; are we missing some code that can baseline these?

jakebailey: (Given we didn't baseline them before, I do wonder if we should just not bother look at this in baselines to have fewer diffs)

gabritto: We did baseline them before, the thing is in Strada we use `position` + `file`, and in Corsa it's LSP location, and I didn't want to write code to convert one into the other, I figured for now the important part is to notice if we do have a location or not, so I'm just emptying locations in the diff fixup functions.

testdata/baselines/reference/submoduleAccepted/fourslash/inlayHints/inlayHintsFunctionParameterTypes1.baseline.diff
gabritto: For the baselines or for diffing the baselines?

testdata/baselines/reference/submoduleAccepted/fourslash/inlayHints/inlayHintsJsDocParameterNames.baseline.diff
jakebailey: Hm, but there's only a string here, not something with a location? Maybe I'm missing something

gabritto: This label stands for the identifier `a`, so its value is `"a"`, and the location should point to the original parameter identifier, but location fields are being omitted when computing the diffs. They do show up in the `.baseline` file. If there's something I can do to make this all less confusing, let me know, I haven't put a ton of effort into the diff fixing code because we're eventually getting rid of it.

jakebailey: I think the thing I was missing was that the Strada baseline says:

```json
{
  "text": "a:",
  "position": 88,
  "kind": "Parameter",
  "whitespaceAfter": true
}
```

Whereas the new baseline says:

```json
{
  "position": {
    "line": 6,
    "character": 6
  },
  "label": [
    {
      "value": "a",
      "location": {
        "uri": "file:///a.js",
        "range": {
          "start": {
            "line": 3,
            "character": 17
          },
          "end": {
            "line": 3,
            "character": 18
          }
        }
      }
    },
    {
      "value": ":"
    }
  ],
  "kind": 2,
  "paddingRight": true
}
```

So with that context, it makes sense what is happening here.

jakebailey (approved):
Skimming, I think this seems good. All of the baseline diffs I read look good, and even though VS Code enables it by default, the defaults are such that they won't actually turn on until the user turns on one of the `typescript.inlayHints.` settings, so for checker.ts this costs nothing.

For the people who turn this on, we're going to use increased resources, but I think this is not a bad situation given that VS code doesn't enable these.

testdata/baselines/reference/submoduleAccepted/fourslash/inlayHints/inlayHintsJsDocParameterNames.baseline.diff
gabritto: Ah yes, that's because "interactive inlay hints" is always true in Corsa now, so we always return inlay hint label parts.

PR #1710, opened 2025-09-11
feat(ls): fix go to definition

rgoomar:
## Overview

These changes made it so I could go to a proper definition. This approach aligns with the way the TypeScript version of the compiler does it using VFS only and VLQ decoding.

Basically we have a situation where we have an import in a `.ts` file like
```
import { getLogger } from '@latticehr/o11y';
```
in a directory like `apps/app-one` in an `index.ts` file and the `@latticehr` is an alias that should reference `packages/o11y` and it does go to the `packages/o11y/dist/*.d.ts` file although not the source file implementation 

This gets to the correct file and line position using VLQ decoding like the TS version does. 

This is an alternative approach to https://github.com/microsoft/typescript-go/pull/1627 in a simpler way reusing existing functionality.

### DIFF
--- a/internal/ls/definition.go
+++ b/internal/ls/definition.go
@@ -12,6 +12,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/checker"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/parser"
 	"github.com/microsoft/typescript-go/internal/scanner"
 	"github.com/microsoft/typescript-go/internal/sourcemap"
 	"github.com/microsoft/typescript-go/internal/tspath"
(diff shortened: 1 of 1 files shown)


internal/project/compilerhost.go
```diff
@@ -144,7 +144,7 @@ func (fs *compilerFS) ReadFile(path string) (contents string, ok bool) {
 	if fh := fs.source.GetFile(path); fh != nil {
 		return fh.Content(), true
 	}
-	return "", false
+	return fs.source.FS().ReadFile(path)
```
rgoomar, line 175: This fixes issues with being able to get the source map files. It conforms to the fallback mechanism that TypeScript had which was going to the disk - https://github.com/microsoft/TypeScript/blob/release-5.9/src/services/sourcemaps.ts#L147-L151

internal/project/projectcollectionbuilder.go
```diff
@@ -535,6 +535,11 @@ func (b *projectCollectionBuilder) findOrCreateDefaultConfiguredProjectWorker(
 			}
 
 			project := b.findOrCreateProject(node.configFileName, configFilePath, node.loadKind, node.logger)
+			if project == nil {
+				node.logger.Log("Project not found")
+				return false, false
+			}
```
rgoomar: This came out of a weird segfault I was running into. I think it's a separate issue although this fixes it. Just a simple check that was missing.

internal/ls/definition_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/bundled"
+	"github.com/microsoft/typescript-go/internal/ls"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/testutil/projecttestutil"
+	"gotest.tools/v3/assert"
+)
+
+func TestInternalAliasGoToDefinition(t *testing.T) {
```
rgoomar, line 14: I specifically added this to have an automated test for functionality that wasn't tested by the baselines.

internal/ls/definition.go
```diff
@@ -235,3 +249,173 @@ func getDeclarationsFromType(t *checker.Type) []*ast.Node {
...
+
+func (s *sourceFileScript) Text() string {
+	return s.text
+}
+
+func (s *sourceFileScript) LineMap() []core.TextPos {
+	return s.lineMap
+}
+
+func findSymbolNearPosition(text, symbolName string, targetPos int) int {
```
rgoomar: This is similar to the TypeScript implementation

internal/ls/definition.go
```diff
@@ -2,14 +2,20 @@ package ls
 
 import (
 	"context"
+	"math"
+	"path/filepath"
```
jakebailey: The `filepath` package should never be used outside low level OS-using packages. `tspath` must be used instead.

internal/ls/definition.go
```diff
@@ -104,8 +110,16 @@ func (l *LanguageService) createLocationsFromDeclarations(declarations []*ast.No
 	for _, decl := range declarations {
 		file := ast.GetSourceFileOfNode(decl)
 		name := core.OrElse(ast.GetNameOfDeclaration(decl), decl)
+
+		fileName := file.FileName()
+		if strings.HasSuffix(fileName, ".d.ts") {
```
jakebailey: This is just `tspath.IsDeclarationFileName`.

internal/ls/definition.go
jakebailey: Similar to what function? I haven't found `findSymbolNearPosition`. I think we might already have this via some other touching property func, maybe

rgoomar: @jakebailey It's the `getPossibleSymbolReferenceNodes` function in TypeScript

jakebailey: Can you just call the func that? We do have `getTouchingPropertyName` too.

internal/ls/definition.go
rgoomar: Adjusted it to use `tspath`

rgoomar:
Closing this out
Superseded by https://github.com/microsoft/typescript-go/pull/1767 😄

PR #1711, opened 2025-09-11
Remove the minimum node version from @typescript/native-preview-* packages

Retsam:
The platform-specific packages are currently being published with:
```json
"engines": {
   "node": ">=20.6.0"
}
```
This will prevent the correct architecture's package from automatically installing when running an older version of node.  (At least, this happened with `pnpm` - not sure if other bundlers use the `engines` field when checking `optionalDependencies`)

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -1077,7 +1077,7 @@ export const buildNativePreviewPackages = task({
         const inputPackageJson = JSON.parse(fs.readFileSync(path.join(inputDir, "package.json"), "utf8"));
         inputPackageJson.version = getVersion();
         delete inputPackageJson.private;
-        delete inputPackageJson.engines.node;
+        delete inputPackageJson.engines;
 
         const { stdout: gitHead } = await $pipe`git rev-parse HEAD`;
         inputPackageJson.gitHead = gitHead;


Herebyfile.mjs
```diff
@@ -1077,6 +1077,7 @@ export const buildNativePreviewPackages = task({
         const inputPackageJson = JSON.parse(fs.readFileSync(path.join(inputDir, "package.json"), "utf8"));
         inputPackageJson.version = getVersion();
         delete inputPackageJson.private;
+        delete inputPackageJson.engines.node;
```
jakebailey: Can you just delete `engines` entirely instead?

jakebailey:
```suggestion
        delete inputPackageJson.engines;
```

Retsam: Yeah, if you'd rather.  I went with this version out of paranoia that there'd be some other engine requirement that would actually be enforced in the future

jakebailey: I can't imagine any. Package-manager based engine restrictions are practically unheard of.

jakebailey:
I think this is fine to do to fix the "package manager doesn't install anything" problem; I did not mean to put the engines in the other packages (they have no code). But note that this doesn't mean the package is going to work outside the engines range.

PR #1714, opened 2025-09-11
Port rename test + baseline fixes

gabritto:
This PR ports the fourslash rename tests.
Note that unlike the existing baseline tests, we also diff the rename baseline tests vs the Strada baselines. Once we have user preferences working, we can go look at the remaining diffs to look for bugs/missing features in rename. 

The PR also fixes some bugs and differences in fourslash baselines in Corsa compared to Strada, fixes some problems in the conversion script, and removes some code from baseline generation that were unused and I think won't be needed in Corsa (e.g. context spans `<| ... |>`). If it turns out we need them in the future, I'll add it back.

One key difference in the baselines is that in Corsa, each LS command gets its own baseline file. In Strada, all baselines for a test were written to the same file. I'm keeping that difference here hoping it will make it easier to look at baselines/diffs to see if things are correct.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -1077,6 +1077,7 @@ export const buildNativePreviewPackages = task({
         const inputPackageJson = JSON.parse(fs.readFileSync(path.join(inputDir, "package.json"), "utf8"));
         inputPackageJson.version = getVersion();
         delete inputPackageJson.private;
+        delete inputPackageJson.engines;
 
         const { stdout: gitHead } = await $pipe`git rev-parse HEAD`;
         inputPackageJson.gitHead = gitHead;
--- a/internal/compiler/fileloader.go
+++ b/internal/compiler/fileloader.go
@@ -356,7 +356,7 @@ func (p *fileLoader) addProjectReferenceTasks(singleThreaded bool) {
 					})
 				}
 			} else {
-				for outputDts := range resolved.GetOutputDeclarationFileNames() {
+				for outputDts := range resolved.GetOutputDeclarationAndSourceFileNames() {
 					if outputDts != "" {
 						p.rootTasks = append(p.rootTasks, &parseTask{
 							normalizedFilePath: outputDts,
--- a/internal/execute/tsctests/tsc_test.go
+++ b/internal/execute/tsctests/tsc_test.go
@@ -2550,6 +2550,62 @@ func TestTscModuleResolution(t *testing.T) {
 				},
 			},
 		},
+		{
+			subScenario: "resolution from d.ts of referenced project",
+			files: FileMap{
+				"/home/src/workspaces/project/common.d.ts": "export type OnValue = (value: number) => void",
+				"/home/src/workspaces/project/producer/index.ts": stringtestutil.Dedent(`
+                    export { ValueProducerDeclaration } from "./in-js"
+                    import { OnValue } from "@common"
+                    export interface ValueProducerFromTs {
+                        onValue: OnValue;
+                    }
+                `),
+				"/home/src/workspaces/project/producer/in-js.d.ts": stringtestutil.Dedent(`
+                    import { OnValue } from "@common"
+                    export interface ValueProducerDeclaration {
+                        onValue: OnValue;
+                    }
+                `),
+				"/home/src/workspaces/project/producer/tsconfig.json": stringtestutil.Dedent(`
+				{
+                    "compilerOptions": {
+                        "strict": true,
+                        "composite": true,
+                        "module": "nodenext",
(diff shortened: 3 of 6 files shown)


internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -97,7 +97,7 @@ function parseFileContent(filename: string, content: string): GoTest | undefined
 }
 
 function getTestInput(content: string): string {
-    const lines = content.split("\n");
+    const lines = content.split("\n").map(line => line.trimEnd());
```
jakebailey: Should this just be trimming `\r` and not all trailing spaces?

jakebailey:
Some of the output filenames have spaces in them; is that intentional? (I can't comment on them directly, github.dev won't let me)

gabritto:
> Some of the output filenames have spaces in them; is that intentional? (I can't comment on them directly, github.dev won't let me)

Ah yes. I think that's because I'm reusing command names (e.g. "Auto Imports") in the baseline file path. Is that likely to cause problems? I can normalize those if so.

jakebailey:
Well, it's sort of annoying to write those names out in the terminal, but it's not the worst. So, it's okay.

iisaduan:
If not difficult and especially if it's just `Auto Imports`, I'd prefer the file/folder names normalized for consistency. It's always bothered me that it's one of the only ones (if not the only one) with a space in it when the command is printed in the baseline.

jakebailey:
Test failure seems to be a logical race in build mode.

```diff
diff --git a/testdata/baselines/reference/tsbuild/sample/explainFiles.js b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
index fa3112b1..ff348ab6 100644
--- a/testdata/baselines/reference/tsbuild/sample/explainFiles.js
+++ b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
@@ -527,7 +527,7 @@ core/index.d.ts
    Imported via '../core/index' from file 'tests/index.ts'
    File is output of project reference source 'core/index.ts'
 core/anotherModule.d.ts
-   Imported via '../core/anotherModule' from file 'tests/index.ts'
+   Imported via '../core/anotherModule' from file 'logic/index.d.ts'
    File is output of project reference source 'core/anotherModule.ts'
 logic/index.d.ts
    Imported via '../logic/index' from file 'tests/index.ts'
```

gabritto:
> Test failure seems to be a logical race in build mode.
> 
> ```diff
> diff --git a/testdata/baselines/reference/tsbuild/sample/explainFiles.js b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> index fa3112b1..ff348ab6 100644
> --- a/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> +++ b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> @@ -527,7 +527,7 @@ core/index.d.ts
>     Imported via '../core/index' from file 'tests/index.ts'
>     File is output of project reference source 'core/index.ts'
>  core/anotherModule.d.ts
> -   Imported via '../core/anotherModule' from file 'tests/index.ts'
> +   Imported via '../core/anotherModule' from file 'logic/index.d.ts'
>     File is output of project reference source 'core/anotherModule.ts'
>  logic/index.d.ts
>     Imported via '../logic/index' from file 'tests/index.ts'
> ```

@sheetalkamat have you seen this problem before?

gabritto:
> Test failure seems to be a logical race in build mode.
> 
> ```diff
> diff --git a/testdata/baselines/reference/tsbuild/sample/explainFiles.js b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> index fa3112b1..ff348ab6 100644
> --- a/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> +++ b/testdata/baselines/reference/tsbuild/sample/explainFiles.js
> @@ -527,7 +527,7 @@ core/index.d.ts
>     Imported via '../core/index' from file 'tests/index.ts'
>     File is output of project reference source 'core/index.ts'
>  core/anotherModule.d.ts
> -   Imported via '../core/anotherModule' from file 'tests/index.ts'
> +   Imported via '../core/anotherModule' from file 'logic/index.d.ts'
>     File is output of project reference source 'core/anotherModule.ts'
>  logic/index.d.ts
>     Imported via '../logic/index' from file 'tests/index.ts'
> ```

@jakebailey I don't think this has anything to do with this PR, right? Since it only touches fourslash test code.

sheetalkamat:
@gabritto  sorry seeing this and noticing it now .. will look into it. its unrelated to your changes. it should have shown both entries, you can ignore this failure and i will work on fix for it.

PR #1720, opened 2025-09-15
Fix isolatedDeclarations emit of computed property names in interfaces

andrewbranch:
This was uncovered while trying to migrate an internal team onto `tsgo --build`.


testdata/baselines/reference/submodule/compiler/isolatedDeclarationErrorsClasses.js.diff
andrewbranch, line 1: I think this diff might be bogus. The submodule’s copy of this baseline doesn‘t include declaration emit output at all, so it must have been meddled with in some way during generation.

weswigham (approved):

PR #1721, opened 2025-09-16
Add `@types/node` to extension

mjbvz:
This is needed for `path`

### DIFF
--- a/_extension/package.json
+++ b/_extension/package.json
@@ -105,6 +105,7 @@
         "vscode-languageclient": "^9.0.1"
     },
     "devDependencies": {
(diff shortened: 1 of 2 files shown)


_extension/package.json
```diff
@@ -105,6 +105,7 @@
         "vscode-languageclient": "^9.0.1"
     },
     "devDependencies": {
+        "@types/node": "^24.5.0",
```
jakebailey, line 108: Probably this should be `^22` given that's what VS Code uses.

jakebailey:
Is this PR still needed? The repo root has `@types/node` already.

PR #1722, opened 2025-09-16
Don't require VSCode ext host restart on change

mjbvz:
Fixes https://github.com/microsoft/vscode/issues/266087

The VSCode TS extension should now dynamically register/unregister language features so a restart isn't required for current insiders builds

### DIFF
--- a/_extension/src/client.ts
+++ b/_extension/src/client.ts
@@ -81,12 +81,12 @@ export class Client {
         };
     }
 
-    async initialize(context: vscode.ExtensionContext): Promise<void> {
+    async initialize(context: vscode.ExtensionContext): Promise<vscode.Disposable> {
         const exe = await getExe(context);
-        this.start(context, exe);
+        return this.start(context, exe);
     }
 
-    async start(context: vscode.ExtensionContext, exe: { path: string; version: string; }): Promise<void> {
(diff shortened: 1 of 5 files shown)


jakebailey:
I tried testing out the latest insiders; I have tsgo enabled on load, then disable it, tsserver does start but tsgo stays running (duplicating everything). If I enable tsgo again, tsserver's errors do not disappear.

So, there might be more changes needed on both sides to better disable things. In our extension, we probably need to hook up disposasls and so on, and on the VS Code side, push diagnostics need to be cleared out?

andrewbranch:
Our disposables _are_ hooked up for deactivation, but a settings change to of `useTsgo` does not constitute a deactivation. We need to collect the subset of disposables that are supposed to be disabled when `useTsgo` is false separately. It’s not all of them; the command to enable the native preview itself is contributed by this extension, so that, at least, needs to be kept alive.

andrewbranch:
The latest commit seems to do the trick, though is not perfect since `context.subscriptions` grows every time you enable/disable. I also didn’t see any issue with diagnostics sticking around after either extension disables, but I didn’t click around a ton.

mjbvz:
Thanks @andrewbranch! Changes look good to me. Let's get this merged so we can do more self hosting

andrewbranch (approved):

jakebailey (approved):

PR #1723, opened 2025-09-16
Implement `getJSSyntacticDiagnosticsForFile`

### LINKED ISSUE #1386: Port getJSSyntacticDiagnosticsForFile
We're missing an implementation of `getJSSyntacticDiagnosticsForFile` and its calling code in `getSyntacticDiagnosticsForFile`. The existing implementation lives in `_submodules/TypeScript/src/compiler/program.ts`, but needs to be ported to Go, then all tests run and the baselines accepted, which should show diffs deleted for places we've forgotten these errors.

ahejlsberg:
In this PR:

* Port `getJSSyntacticDiagnosticsForFile` which generates diagnostics for TypeScript-only constructs in .js files.
* Remove extra blank lines printed between diagnostics in `-pretty false` mode.
* In language service, include semantic diagnostics even when there are syntax errors in a file (as was the case in Strada).

Note that `getJSSyntacticDiagnosticsForFile` is now called by `ParseSourceFile` such that it becomes part of the concurrent parsing of source files. In Strada it was deferred and called upon the first request to obtain syntactic diagnostics for a file. The function effective constitutes an extra pass over every parsed .js file. Longer term we could consider integrating the logic with the parser itself.

This PR supersedes #1387.

Fixes #1386.


internal/ls/diagnostics.go
```diff
@@ -13,17 +13,14 @@ import (
...
-		diagnostics = append(diagnostics, syntaxDiagnostics)
-	} else {
-		diagnostics = append(diagnostics, program.GetSemanticDiagnostics(ctx, file))
-		// !!! user preference for suggestion diagnostics; keep only unnecessary/deprecated?
-		// See: https://github.com/microsoft/vscode/blob/3dbc74129aaae102e5cb485b958fa5360e8d3e7a/extensions/typescript-language-features/src/languageFeatures/diagnostics.ts#L114
-		diagnostics = append(diagnostics, program.GetSuggestionDiagnostics(ctx, file))
-		if program.Options().GetEmitDeclarations() {
-			diagnostics = append(diagnostics, program.GetDeclarationDiagnostics(ctx, file))
-		}
+	diagnostics := make([][]*ast.Diagnostic, 0, 4)
```
jakebailey, line 16: I had thought this code was correct originally; I guess you're changing it to match that VS Code on the client side requested syntax and semantic in sequence? (I've been trying to find equivalent code but I think this is just somewhere that differs in general with Strada given the fine-grained-ness of tsserver.)

testdata/baselines/reference/submoduleAccepted/conformance/plainJSGrammarErrors.errors.txt.diff
```diff
@@ -1,69 +1,33 @@
...
- plainJSGrammarErrors.js(22,11): error TS1248: A class member cannot have the 'const' keyword.
--plainJSGrammarErrors.js(23,5): error TS8009: The 'const' modifier can only be used in TypeScript files.
- plainJSGrammarErrors.js(23,11): error TS1248: A class member cannot have the 'const' keyword.
+@@= skipped -11, +11 lines =@@
  plainJSGrammarErrors.js(26,11): error TS1030: 'async' modifier already seen.
  plainJSGrammarErrors.js(28,11): error TS1029: 'static' modifier must precede 'async' modifier.
  plainJSGrammarErrors.js(29,5): error TS1031: 'export' modifier cannot appear on class elements of this kind.
 -plainJSGrammarErrors.js(29,5): error TS8009: The 'export' modifier can only be used in TypeScript files.
  plainJSGrammarErrors.js(30,5): error TS1031: 'export' modifier cannot appear on class elements of this kind.
++plainJSGrammarErrors.js(34,9): error TS8017: Signature declarations can only be used in TypeScript files.
```
jakebailey, line 9: Seemingly a difference here.

testdata/baselines/reference/submoduleAccepted/conformance/plainJSGrammarErrors.errors.txt.diff
```diff
@@ -1,69 +1,33 @@
...
- plainJSGrammarErrors.js(58,26): error TS1090: 'export' modifier cannot appear on a parameter.
--plainJSGrammarErrors.js(58,26): error TS8012: Parameter modifiers can only be used in TypeScript files.
- plainJSGrammarErrors.js(59,25): error TS1090: 'async' modifier cannot appear on a parameter.
--plainJSGrammarErrors.js(59,25): error TS8012: Parameter modifiers can only be used in TypeScript files.
+ plainJSGrammarErrors.js(36,9): error TS1049: A 'set' accessor must have exactly one parameter.
+@@= skipped -26, +26 lines =@@
  plainJSGrammarErrors.js(60,7): error TS1030: 'async' modifier already seen.
  plainJSGrammarErrors.js(61,1): error TS1042: 'async' modifier cannot be used here.
  plainJSGrammarErrors.js(62,5): error TS1042: 'async' modifier cannot be used here.
 -plainJSGrammarErrors.js(62,5): error TS8009: The 'async' modifier can only be used in TypeScript files.
```
jakebailey, line 17: This error seems to still be missing? But, probably "good" because the old error was duplicated for some reason.

jakebailey (approved):
Seems fine to me, though it does slow down parsing of JS files to do the second pass:

```
goos: linux
goarch: amd64
                               │   sec/op    │    sec/op     vs base               │
Parse/Herebyfile.mjs/tsc-20      941.2µ ± 1%   1032.6µ ± 1%  +9.71% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20   956.7µ ± 3%   1020.3µ ± 4%  +6.65% (p=0.000 n=10)
geomean                          948.9µ         1.026m       +8.17%

Parse/Herebyfile.mjs/tsc-20      461.8Ki ± 0%   462.0Ki ± 0%  +0.04% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20   461.8Ki ± 0%   462.0Ki ± 0%  +0.04% (p=0.000 n=10)
geomean                          461.8Ki        462.0Ki       +0.04%

                               │  allocs/op  │  allocs/op   vs base               │
Parse/Herebyfile.mjs/tsc-20      1.761k ± 0%   1.766k ± 0%  +0.28% (p=0.000 n=10)
Parse/Herebyfile.mjs/server-20   1.761k ± 0%   1.766k ± 0%  +0.28% (p=0.000 n=10)
geomean                          1.761k        1.766k       +0.28%
```

This is nothing "new", the old code had this problem.

I don't see a way around this unless we move this code into the parser itself (which, I was surprised to learn it wasn't).

testdata/baselines/reference/submoduleAccepted/conformance/plainJSGrammarErrors.errors.txt.diff
jakebailey: Probably inconsequential, given it's a missing body error in JS. Which, sort of funny that we can detect that there's a syntax error for JS, as that implies that this check could be elsewhere.

internal/ls/diagnostics.go
ahejlsberg: I simply noticed a behavioral difference, with semantic error red squiggles appearing and disappearing depending on whether there are syntax errors as you're typing. LSP doesn't distinguish between syntactic/semantic so we need to return all of them to get consistent behavior.

testdata/baselines/reference/submoduleAccepted/conformance/plainJSGrammarErrors.errors.txt.diff
ahejlsberg: We already have grammar checks for all modifiers, so the check for `export` in Strada is duplicative (as evident from the diagnostics). Also, it's not even correct, `export` _is_ permitted in JS.

testdata/baselines/reference/submoduleAccepted/conformance/plainJSGrammarErrors.errors.txt.diff
ahejlsberg: Another case of duplicate checks.

PR #1728, opened 2025-09-17
Port late bound class property declaration emit

andrewbranch:
I ported this thinking it was related to #1720, but it ended up being something else.


testdata/baselines/reference/submodule/conformance/overrideDynamicName1(noimplicitoverride=false).js.diff
```diff
@@ -1,27 +1,26 @@
 --- old.overrideDynamicName1(noimplicitoverride=false).js
 +++ new.overrideDynamicName1(noimplicitoverride=false).js
-@@= skipped -30, +30 lines =@@
- //// [overrideDynamicName1.d.ts]
+@@= skipped -31, +31 lines =@@
  declare let prop: string;
  declare class Base1 {
--    [prop]: () => void;
+     [prop]: () => void;
 +    [x: string]: () => void;
```
andrewbranch: So this PR fixes the missing `[prop]: () => void;` but there was already an extra `[x: string]: () => void;` which is now superfluous.

weswigham: Yeah, the extra `[x: string]: () => void;` is being added because there's a missing `continue` in `CreateLateBoundIndexSignatures` in the `.components` branch that just got implemented (at the end of the `allComponentComputedNamesSerializable` if) to skip it from falling back to the index signature generation case.

PR #1729, opened 2025-09-17
fourslash userpreferences (and fetch inlay hints userpreferences from vscode) 

iisaduan:
Goal is to unblock https://github.com/microsoft/typescript-go/pull/1705

- fourslash supports all userpreferences
- only inlay hints preferences are parsed from vscode's client config responses. 

A full implementation of LSP compliant config getting/updating requires more work in the extension (currently in progress)

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -38,6 +38,7 @@ type FourslashTest struct {
 	testData     *TestData // !!! consolidate test files from test data and script info
 	baselines    map[string]*strings.Builder
 	rangesByText *collections.MultiMap[string, *RangeMarker]
+	config       *ls.UserPreferences
 
 	scriptInfos map[string]*scriptInfo
 	converters  *ls.Converters
@@ -271,6 +272,29 @@ func sendRequest[Params, Resp any](t *testing.T, f *FourslashTest, info lsproto.
 	if resp == nil {
 		return nil, *new(Resp), false
 	}
+
+	// currently, the only request that may be sent by the server during a client request is one `config` request
+	// !!! remove if `config` is handled in initialization and there are no other server-initiated requests
+	if resp.Kind == lsproto.MessageKindRequest {
+		req := resp.AsRequest()
+		switch req.Method {
+		case lsproto.MethodWorkspaceConfiguration:
+			req := lsproto.ResponseMessage{
+				ID:      req.ID,
+				JSONRPC: req.JSONRPC,
+				Result:  []any{&f.config},
+			}
+			f.writeMsg(t, req.Message())
+			resp = f.readMsg(t)
+		default:
+			// other types of responses not yet used in fourslash; implement them if needed
+			t.Fatalf("Unexpected request received: %s", req)
+		}
+	}
+
+	if resp == nil {
+		return nil, *new(Resp), false
+	}
 	result, ok := resp.AsResponse().Result.(Resp)
 	return resp, result, ok
 }
@@ -298,6 +322,13 @@ func (f *FourslashTest) readMsg(t *testing.T) *lsproto.Message {
 	}
 	assert.NilError(t, json.MarshalWrite(io.Discard, msg), "failed to encode message as JSON")
 	return msg
+}
+
+func (f *FourslashTest) configure(t *testing.T, config *ls.UserPreferences) {
+	f.config = config
+	sendNotification(t, f, lsproto.WorkspaceDidChangeConfigurationInfo, &lsproto.DidChangeConfigurationParams{
+		Settings: config,
+	})
 }
 
 func (f *FourslashTest) GoToMarkerOrRange(t *testing.T, markerOrRange MarkerOrRange) {
@@ -540,6 +571,11 @@ func (f *FourslashTest) verifyCompletionsWorker(t *testing.T, expected *Completi
 		},
 		Position: f.currentCaretPosition,
 		Context:  &lsproto.CompletionContext{},
+	}
+	if expected == nil {
+		f.configure(t, nil)
+	} else {
+		f.configure(t, expected.UserPreferences)
 	}
 	resMsg, result, resultOk := sendRequest(t, f, lsproto.TextDocumentCompletionInfo, params)
 	if resMsg == nil {
--- a/internal/ls/languageservice.go
+++ b/internal/ls/languageservice.go
@@ -7,19 +7,29 @@ import (
 )
 
 type LanguageService struct {
-	host       Host
-	converters *Converters
+	host            Host
+	converters      *Converters
+	userPreferences *UserPreferences
 }
 
-func NewLanguageService(host Host, converters *Converters) *LanguageService {
+func NewLanguageService(host Host, converters *Converters, preferences *UserPreferences) *LanguageService {
 	return &LanguageService{
-		host:       host,
-		converters: converters,
+		host:            host,
+		converters:      converters,
+		userPreferences: preferences,
 	}
 }
 
 func (l *LanguageService) GetProgram() *compiler.Program {
 	return l.host.GetProgram()
+}
+
+func (l *LanguageService) UpdateUserPreferences(preferences *UserPreferences) {
+	l.userPreferences = preferences
+}
+
+func (l *LanguageService) UserPreferences() *UserPreferences {
+	return l.userPreferences
 }
 
 func (l *LanguageService) tryGetProgramAndFile(fileName string) (*compiler.Program, *ast.SourceFile) {
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -217,6 +217,101 @@ func (s *Server) RefreshDiagnostics(ctx context.Context) error {
 	return nil
 }
 
+func (s *Server) Configure(ctx context.Context) (*ls.UserPreferences, error) {
+	result, err := s.sendRequest(ctx, lsproto.MethodWorkspaceConfiguration, &lsproto.ConfigurationParams{
+		Items: []*lsproto.ConfigurationItem{
+			{
+				Section: ptrTo("typescript"),
+			},
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("configure request failed: %w", err)
+	}
+	configs := result.([]any)
+	userPreferences := &ls.UserPreferences{}
+	for _, config := range configs {
+		if config == nil {
+			continue
+		}
+		if item, ok := config.(map[string]any); ok {
+			for name, values := range item {
+				switch name {
+				case "inlayHints":
+					inlayHintsPreferences := values.(map[string]any)
+					if v, ok := inlayHintsPreferences["parameterNames"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							if enabledStr, ok := enabled.(string); ok {
+								userPreferences.IncludeInlayParameterNameHints = enabledStr
+							} else {
+								userPreferences.IncludeInlayParameterNameHints = ""
+							}
+						}
+						if supressWhenArgumentMatchesName, ok := v["suppressWhenArgumentMatchesName"]; ok {
+							userPreferences.IncludeInlayParameterNameHintsWhenArgumentMatchesName = ptrTo(!supressWhenArgumentMatchesName.(bool))
+						}
+					}
+					if v, ok := inlayHintsPreferences["parameterTypes"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							userPreferences.IncludeInlayFunctionParameterTypeHints = ptrTo(enabled.(bool))
+						}
+					}
+					if v, ok := inlayHintsPreferences["variableTypes"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							userPreferences.IncludeInlayVariableTypeHints = ptrTo(enabled.(bool))
+						}
+						if supressWhenTypeMatchesName, ok := v["suppressWhenTypeMatchesName"]; ok {
+							userPreferences.IncludeInlayVariableTypeHintsWhenTypeMatchesName = ptrTo(!supressWhenTypeMatchesName.(bool))
+						}
+					}
+					if v, ok := inlayHintsPreferences["propertyDeclarationTypes"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							userPreferences.IncludeInlayPropertyDeclarationTypeHints = ptrTo(enabled.(bool))
+						}
+					}
+					if v, ok := inlayHintsPreferences["functionLikeReturnTypes"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							userPreferences.IncludeInlayFunctionLikeReturnTypeHints = ptrTo(enabled.(bool))
+						}
+					}
+					if v, ok := inlayHintsPreferences["enumMemberValues"].(map[string]any); ok && v != nil {
+						if enabled, ok := v["enabled"]; ok {
+							userPreferences.IncludeInlayEnumMemberValueHints = ptrTo(enabled.(bool))
+						}
+					}
+					userPreferences.InteractiveInlayHints = ptrTo(true)
+				case "tsserver":
+					// !!!
+				case "unstable":
+					// !!!
+				case "tsc":
+					// !!!
+				case "updateImportsOnFileMove":
+					// !!! moveToFile
+				case "preferences":
+					// !!!
+				case "experimental":
+					// !!!
+				case "organizeImports":
+					// !!!
+				case "importModuleSpecifierEnding":
+					// !!!
+				}
+			}
+			continue
+		}
+		if item, ok := config.(ls.UserPreferences); ok {
+			// case for fourslash
+			userPreferences = &item
+			break
+		}
+	}
+	// !!! set defaults for services, remove after extension is updated
+	userPreferences.IncludeCompletionsForModuleExports = ptrTo(true)
+	userPreferences.IncludeCompletionsForImportStatements = ptrTo(true)
+	return userPreferences, nil
+}
+
 func (s *Server) Run() error {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
@@ -440,6 +535,7 @@ var handlers = sync.OnceValue(func() handlerMap {
 	registerRequestHandler(handlers, lsproto.ShutdownInfo, (*Server).handleShutdown)
 	registerNotificationHandler(handlers, lsproto.ExitInfo, (*Server).handleExit)
 
+	registerNotificationHandler(handlers, lsproto.WorkspaceDidChangeConfigurationInfo, (*Server).handleDidChangeWorkspaceConfiguration)
 	registerNotificationHandler(handlers, lsproto.TextDocumentDidOpenInfo, (*Server).handleDidOpen)
 	registerNotificationHandler(handlers, lsproto.TextDocumentDidChangeInfo, (*Server).handleDidChange)
 	registerNotificationHandler(handlers, lsproto.TextDocumentDidSaveInfo, (*Server).handleDidSave)
@@ -638,7 +734,6 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 	if shouldEnableWatch(s.initializeParams) {
 		s.watchEnabled = true
 	}
-
 	s.session = project.NewSession(&project.SessionInit{
 		Options: &project.SessionOptions{
 			CurrentDirectory:   s.cwd,
@@ -655,6 +750,11 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 		NpmExecutor: s,
 		ParseCache:  s.parseCache,
 	})
+	userPreferences, err := s.Configure(ctx)
+	if err != nil {
+		return err
+	}
+	s.session.Configure(userPreferences)
 	// !!! temporary; remove when we have `handleDidChangeConfiguration`/implicit project config support
 	if s.compilerOptionsForInferredProjects != nil {
 		s.session.DidChangeCompilerOptionsForInferredProjects(ctx, s.compilerOptionsForInferredProjects)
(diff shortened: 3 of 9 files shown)


jakebailey:
> A full implementation of LSP compliant config getting/updating requires more work in the extension (currently in progress)

What is needed? I would think we would just ask for the setting blocks we are interested in and watch them, from the server side (no extension code needed).

iisaduan:
Our `userPreferences` struct doesn't exist at all in vscode--the previous extension did a lot of processing of the way that vscode's settings.json is represented before it passed a `userPreferences` to tsserver. We could do this processing on our side, but the code is so much more complicated in go (hence why I didn't implement every preference), and would mean that other clients (if they want to support userPreferences) have to convert their preferences into vscode's preferences, which then our server would convert again. Also, LSP spec says that the client should do this processing.

The other extension changes include to make configuration happen on initialization

jakebailey:
I feel confused. The LSP config spec is exactly the same as asking VS Code for a setting out of JSON. Putting more code in the client is just going to make that same code repeat across editors...

internal/lsp/server.go
```diff
@@ -217,6 +217,101 @@ func (s *Server) RefreshDiagnostics(ctx context.Context) error {
...
 }
 
+func (s *Server) Configure(ctx context.Context) (*ls.UserPreferences, error) {
+	result, err := s.sendRequest(ctx, lsproto.MethodWorkspaceConfiguration, &lsproto.ConfigurationParams{
+		Items: []*lsproto.ConfigurationItem{
+			{
+				Section: ptrTo("typescript"),
+			},
+		},
+	})
```
iisaduan, line 233: I think I see what you mean, are you asking if instead of requesting the config for one entry, `"typescript"`, we request a list of preferences, one for every preference area that we expect from vscode: 
```suggestion
	result, err := s.sendRequest(ctx, lsproto.MethodWorkspaceConfiguration, &lsproto.ConfigurationParams{
		Items: []*lsproto.ConfigurationItem{
			{
				Section: ptrTo("typescript.inlayHints"),
			},
			{
				Section: ptrTo("typescript.preferences"),
			},
			{
				Section: ptrTo("typescript.unstable"),
			},
           .... etc
		},
	})
```
 and process like that?

iisaduan: or even ask for every individual setting like `"typescript.inlayHints.parameterNames.enabled"`, "typescript.inlayHints.supressWhenTypeMatchesName.enabled", etc

internal/fourslash/fourslash.go
```diff
@@ -38,6 +38,7 @@ type FourslashTest struct {
 	testData     *TestData // !!! consolidate test files from test data and script info
 	baselines    map[string]*strings.Builder
 	rangesByText *collections.MultiMap[string, *RangeMarker]
+	config       *ls.UserPreferences
```
gabritto: nit: I think it would be easier to follow things if we stick to calling this "preferences", since that's the name of the type

internal/lsp/server.go
jakebailey: Yes, though I think we have to also request `javascript` too, and maybe even `js/ts`?

Then best case, we register a config watch on them and then keep them up to date.

internal/ls/languageservice.go
```diff
@@ -7,21 +7,31 @@ import (
...
+		converters:      converters,
+		userPreferences: preferences,
 	}
 }
 
 func (l *LanguageService) GetProgram() *compiler.Program {
 	return l.host.GetProgram()
 }
 
+func (l *LanguageService) UpdateUserPreferences(preferences *UserPreferences) {
```
gabritto: Where/why are we using this? Since this comes from the snapshot, shouldn't it be immutable?

internal/lsp/server.go
```diff
@@ -217,6 +217,101 @@ func (s *Server) RefreshDiagnostics(ctx context.Context) error {
...
+	configs := result.([]any)
+	userPreferences := &ls.UserPreferences{}
+	for _, config := range configs {
+		if config == nil {
+			continue
+		}
+		if item, ok := config.(map[string]any); ok {
+			for name, values := range item {
+				switch name {
+				case "inlayHints":
```
gabritto: I wonder if we could generate this code, maybe from the `package.json` in the extension when we have that.

internal/lsp/server.go
```diff
@@ -672,6 +772,16 @@ func (s *Server) handleExit(ctx context.Context, params any) error {
 	return io.EOF
 }
 
+func (s *Server) handleDidChangeWorkspaceConfiguration(ctx context.Context, params *lsproto.DidChangeConfigurationParams) error {
+	// !!! update user preferences
+	// !!! only usable by fourslash
+	if item, ok := params.Settings.(*ls.UserPreferences); ok {
```
gabritto: Can't we do the same kind of parsing here that we're doing in `Configure` above?

internal/lsp/server.go
```diff
@@ -655,6 +750,11 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 		NpmExecutor: s,
 		ParseCache:  s.parseCache,
 	})
+	userPreferences, err := s.Configure(ctx)
+	if err != nil {
+		return err
+	}
+	s.session.Configure(userPreferences)
```
gabritto: Don't we also need to register with the client here to be able to actually get config change notifications? Just like we do in `WatchFiles`.

internal/fourslash/fourslash.go
```diff
@@ -541,6 +572,11 @@ func (f *FourslashTest) verifyCompletionsWorker(t *testing.T, expected *Completi
 		Position: f.currentCaretPosition,
 		Context:  &lsproto.CompletionContext{},
 	}
+	if expected == nil {
```
gabritto: I was hoping we'd have a mechanism in fourslash to scope the preferences changes to a single `verify...` method call (maybe using `defer`?). Otherwise, I think we can accidentally change preferences for later methods.

internal/fourslash/fourslash.go
```diff
@@ -268,6 +269,29 @@ func sendRequest[Params, Resp any](t *testing.T, f *FourslashTest, info lsproto.
 	)
 	f.writeMsg(t, req.Message())
 	resp := f.readMsg(t)
+	if resp == nil {
+		return nil, *new(Resp), false
+	}
+
+	// currently, the only request that may be sent by the server during a client request is one `config` request
+	// !!! remove if `config` is handled in initialization and there are no other server-initiated requests
+	if resp.Kind == lsproto.MessageKindRequest {
```
gabritto, line 286: I do think we should move this handling to initialization. Or are going to need to handle other requests as well?

internal/lsp/server.go
iisaduan: Yeah, re: the conversation I had with jake above, I'm doing some thinking about other ways we could do this. This function as-is definitely isn't permanent

internal/fourslash/fourslash.go
iisaduan: Good idea, I'll implement that

internal/fourslash/fourslash.go
iisaduan: I'm not sure what we will need in the future

internal/lsp/server.go
iisaduan: Re: this + the comment above ([link](https://github.com/microsoft/typescript-go/pull/1729#discussion_r2360659453)), I didn't fully implement this function because I wanted to figure out what we're going to process in the extension first and I wasn't sure the range of how much info that the client can pass to the server (will it always pass the entire new config upon changes? or only the differences?)

internal/fourslash/fourslash.go
andrewbranch: I think you could imagine the fourslash client handling diagnostics refresh requests and potentially even watch requests.

internal/project/snapshot.go
```diff
@@ -234,6 +241,7 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
 		session.extendedConfigCache,
 		nil,
 		compilerOptionsForInferredProjects,
+		session.userPreferences,
```
andrewbranch: This is a race—you only protect the boolean `session.pendingChanges` with the mutex while computing whether you need to do a snapshot update. By the time you get here, `session.userPreferences` could have changed since that boolean was accessed. You need to lock and extract the actual new value of the user preferences, and then make that part of `SnapshotChange`.

andrewbranch (dismissed):

internal/fourslash/fourslash.go
iisaduan: If we're going to port the fourslash/server tests, then those probably will have to be handled?

internal/ls/userpreferences.go
```diff
@@ -199,7 +309,35 @@ const (
...
+	return QuotePreferenceUnknown
+}
+
+func (p *UserPreferences) Copy() *UserPreferences {
+	if p == nil {
+		return nil
+	}
+	prefCopy := *p
+	copy(prefCopy.AutoImportSpecifierExcludeRegexes, p.AutoImportSpecifierExcludeRegexes)
+	copy(prefCopy.AutoImportFileExcludePatterns, p.AutoImportFileExcludePatterns)
```
jakebailey: This kind of copy won't work; prefCopy is a shallow copy, so both args are the same.

Probably you meant `slices.Clone`?

internal/ls/userpreferences.go
```diff
@@ -209,3 +347,287 @@ func (p *UserPreferences) ModuleSpecifierPreferences() modulespecifiers.UserPref
...
+
+// returns non-nil if should break loop
+func (p *UserPreferences) Parse(item any) *UserPreferences {
+	if item == nil {
+		// continue
+	} else if config, ok := item.(map[string]any); ok {
+		p.parseWorker(config)
+	} else if item, ok := item.(*UserPreferences); ok {
+		// case for fourslash
+		return item.CopyOrDefault()
```
jakebailey, line 362: I think we should avoid this, because it means we are not testing the parsing end to end at all. But, I assume the alternative means somehow converting UserPreferences _back_ into LSP style objects?

internal/ls/userpreferences.go
```diff
@@ -209,3 +347,287 @@ func (p *UserPreferences) ModuleSpecifierPreferences() modulespecifiers.UserPref
...
+func (p *UserPreferences) parseInlayHints(prefs any) {
+	inlayHintsPreferences, ok := prefs.(map[string]any)
+	if !ok {
+		return
+	}
+	for name, value := range inlayHintsPreferences {
+		if v, ok := value.(map[string]any); ok {
+			// vscode's inlay hints settings are nested objects with "enabled" and other properties
+			switch name {
+			case "parameterNames":
```
jakebailey, line 416: We are case sensitive in some places, but not in others?

internal/lsp/server.go
```diff
@@ -686,6 +709,18 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
...
+
+	// request userPreferences if not provided at initialization
+	if s.initializeParams.InitializationOptions == nil {
+		userPreferences, err := s.RequestConfiguration(ctx)
+		if err != nil {
+			return err
+		}
+		s.session.Configure(userPreferences)
+	} else {
+		// !!! handle userPreferences from initialization
```
jakebailey: When will we ever have these?

internal/lsp/server.go
```diff
@@ -686,6 +709,18 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 		NpmExecutor: s,
 		ParseCache:  s.parseCache,
 	})
+
+	// request userPreferences if not provided at initialization
+	if s.initializeParams.InitializationOptions == nil {
+		userPreferences, err := s.RequestConfiguration(ctx)
+		if err != nil {
+			return err
+		}
```
jakebailey, line 725: We need to be checking a client capability before requesting this. Clients aren't required to implement config reading.

internal/lsp/server.go
```diff
@@ -703,6 +738,15 @@ func (s *Server) handleExit(ctx context.Context, params any) error {
 	return io.EOF
 }
 
+func (s *Server) handleDidChangeWorkspaceConfiguration(ctx context.Context, params *lsproto.DidChangeConfigurationParams) error {
+	userPreferences := s.session.UserPreferences().CopyOrDefault()
+	if parsed := userPreferences.Parse(params.Settings); parsed != nil {
```
jakebailey: I find this pattern odd; shouldn't Parse be a top level func that takes in the params and then returns user prefs, rather than having to have an existing one, copy it, then parse over top of it?

internal/ls/userpreferences.go
iisaduan: yeah. I did it this way because it feels easier to write the convert fourslash script to output `UserPreferences` struct, but it's also easy to add a case sending over `map[string]any` to do parsing testing

internal/ls/userpreferences.go
iisaduan: case sensitive is for vscode specifically (they have specific cases that aren't in the typescript-language-server api, so i don't think we should support them elsewhere)

internal/ls/userpreferences.go
```diff
@@ -199,7 +309,35 @@ const (
...
+	copy(prefCopy.AutoImportFileExcludePatterns, p.AutoImportFileExcludePatterns)
+	return &prefCopy
+}
+
+func (p *UserPreferences) CopyOrDefault() *UserPreferences {
+	if p == nil {
+		return NewDefaultUserPreferences()
+	}
+	return p.Copy()
 }
```
jakebailey, line 342: I think we should try to make sure user prefs are always non-nil, and just have the session's initialized default user pref be the default pref, rather than having this method called often.

internal/lsp/server.go
iisaduan: requires changes in the extension, which I haven't done yet

internal/lsp/server.go
```diff
@@ -788,10 +832,8 @@ func (s *Server) handleCompletion(ctx context.Context, languageService *ls.Langu
 		params.Position,
 		params.Context,
 		getCompletionClientCapabilities(s.initializeParams),
-		&ls.UserPreferences{
-			IncludeCompletionsForModuleExports:    core.TSTrue,
-			IncludeCompletionsForImportStatements: core.TSTrue,
-		})
+		languageService.UserPreferences(),
```
jakebailey: We should probably drop these params, as they are already on `languageService` methods, so the method can already ask for `UserPreferences`.

internal/lsp/server.go
jakebailey: What extension? Us? The VS Code extension will always be able to request via the normal means, so I don't think we would include that info anywhere.

internal/ls/userpreferences.go
jakebailey: Using UserPreferences in the tests is fine, we can keep this code for sure, but I do think we should eventually write code that can take UserPreferences and output an actual config object for us to start roundtripping, rather than relying on manually testing the editor to make sure the parsing code works.

internal/lsp/server.go
```diff
@@ -441,6 +463,7 @@ var handlers = sync.OnceValue(func() handlerMap {
 	registerRequestHandler(handlers, lsproto.ShutdownInfo, (*Server).handleShutdown)
 	registerNotificationHandler(handlers, lsproto.ExitInfo, (*Server).handleExit)
 
+	registerNotificationHandler(handlers, lsproto.WorkspaceDidChangeConfigurationInfo, (*Server).handleDidChangeWorkspaceConfiguration)
```
jakebailey, line 468: This handler is set up, but I don't think the PR sets up the watch capability yet.

internal/lsp/server.go
iisaduan: I did it this way because for both initalization/config requests and `didChangeConfig` notifications, we don't expect most of the preferences to be specified in the vast majority of cases. This means that most of the time, we would be merging a couple of parsed preferences into either an existing or a default preference, and doing "parsing over" makes it a lot easier to handle defaults/only change specified preferences in `didChangeConfig`

internal/lsp/server.go
iisaduan: theoretically, it would be nice to have userpreferences (and other preferences/config info) passed over in the actual `initialization` request through [`initializeParams.initializationOptions`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams), so we don't have to wait for a `configuration` request to come back. It doesn't seem to be included by default, so we'd have to fix the extension to give extra info in `initializationOptions`

jakebailey: I suppose so, but I've never seen a server which did that, honestly... It just means copying the same "get all of the options" code into the client another time, and then it gets copied into other clients, to save one roundtrip.

iisaduan: oh huh it looks like the typescript-language-server already allows this https://github.com/typescript-language-server/typescript-language-server/blob/master/docs/configuration.md/#initializationoptions

jakebailey: Yes, iirc because tsserver only really accepts prefs at init time, or supplied along with a request, as tsserver has no ability to call back to the client.

internal/lsp/server.go
```diff
@@ -218,6 +218,28 @@ func (s *Server) RefreshDiagnostics(ctx context.Context) error {
 	return nil
 }
 
+func (s *Server) RequestConfiguration(ctx context.Context) (*ls.UserPreferences, error) {
+	result, err := s.sendRequest(ctx, lsproto.MethodWorkspaceConfiguration, &lsproto.ConfigurationParams{
+		Items: []*lsproto.ConfigurationItem{
+			{
+				Section: ptrTo("typescript"),
```
jakebailey, line 230: Unfortunately there is also `javascript` and `js/ts` scopes we might need.

I think we're going to have to make user prefs have both variants and then return one or the other depending on the situation...

internal/project/session.go
```diff
@@ -177,11 +182,22 @@ func (s *Session) GetCurrentDirectory() string {
 	return s.options.CurrentDirectory
 }
 
+func (s *Session) UserPreferences() *ls.UserPreferences {
+	return s.userPreferences
```
andrewbranch: This should be protected by a mutex (you can reuse `s.pendingConfigChangesMu`, maybe with a rename and change to a RWMutex)

`s.compilerOptionsForInferredProjects` would be in the same boat except that it’s literally never read 🤔

internal/project/snapshot.go
```diff
@@ -126,6 +133,7 @@ type SnapshotChange struct {
 	// It should only be set the value in the next snapshot should be changed. If nil, the
 	// value from the previous snapshot will be copied to the new snapshot.
 	compilerOptionsForInferredProjects *core.CompilerOptions
+	newConfig                          *ls.UserPreferences
```
andrewbranch: I don’t love the mix of terminology. Can we either standardize everything to use LSP terminology (workspace configuration) or say, once the “workspace configuration” has been filtered and parsed into stuff the **language service** cares about, we refer to that bit consistently as user preferences?

internal/ls/userpreferences.go
jakebailey: This is not blocking for this PR; going to resolve it.

internal/lsp/server.go
jakebailey: Not blocking for this PR, but needs to be done eventually.

internal/project/snapshot.go
jakebailey: I would prefer the latter, as in "try to resolve settings to a clearer set of shared options and use that".

jakebailey:
This sort of blocks fixing #1914; would be really good to get some form of this in so we can start improving other stuff in the meantime, I think.

iisaduan:
Im currently fixing merge conflicts with this, and I should be able to address a lot of the formatting issues immediately after this is in. 

Before I left I was fixing a server crash that happened with `didChangeNotification`. I fixed that as well, but I'm not sure it's 100% registered correctly at initialization. We need `didChangeNotification` (or some other back door) for fourslash, but we didn't support it previously, especially not for `userPreferences`

internal/lsp/server.go
jakebailey: Are you planning to do this now, or in the future?

internal/ls/userpreferences.go
```diff
@@ -209,3 +348,289 @@ func (p *UserPreferences) ModuleSpecifierPreferences() modulespecifiers.UserPref
...
+		case "unstable":
+			continue
+		case "inlayHints":
+			p.parseInlayHints(values)
+		case "suggest":
+			p.parseSuggest(values)
+		case "preferences":
+			p.parsePreferences(values)
+		case "format":
+			// !!!
```
jakebailey, line 384: Is this challenging to fill out? This is where we are wanting it next, I believe.

internal/lsp/server.go
iisaduan: Future

internal/ls/userpreferences.go
iisaduan: I'm working on it at the moment, I didn't want to bog this PR down farther

iisaduan: it's not an insignificant add, I'll need to do pretty much the handling in the same places as i did for userpreferences-- pretty much do another version of everything except the actual server/client request, and I would like to get this PR in to finalize the server/client request and the general `userPreferences` lifecycle handling in snapshot/session at least. It'll be a lot easier to add on other kinds of settings (like `format`) if I'm not worried about needing to change those things

internal/lsp/server.go
```diff
@@ -683,6 +711,21 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
...
 	})
+
+	if s.initializeParams != nil && s.initializeParams.InitializationOptions != nil && *s.initializeParams.InitializationOptions != nil {
+		// handle userPreferences from initializationOptions
+		userPreferences := s.session.NewUserPreferences()
+		userPreferences.Parse(*s.initializeParams.InitializationOptions)
+		s.session.InitializeWithConfig(userPreferences)
+	} else {
+		// request userPreferences if not provided at initialization
+		userPreferences, err := s.RequestConfiguration(ctx)
```
jakebailey, line 722: One blocking thing here; we need to check the client capabilities to ensure that the client can even handle this request. As in `ClientCapabilities.Workspace.Configuration`.

jakebailey: Ah, sorry, just noticed you did this in `RequestConfiguration` instead.

iisaduan: that's checked in `RequestConfiguration()`

PR #1732, opened 2025-09-19
Integrate JS syntax checks with parser and remove separate pass

ahejlsberg:
This is a follow-up to #1723 that integrates JS syntax checks with the parser and removes the separate pass that, for some reason, was used to implement the checks previously. This eliminates any additional parsing overhead for parsing JS files (the separate pass added ~10% to parse time).

The PR also fixes a few oddities in the JS syntax checks. In particular, we now include full text ranges of signatures to make it clearer that the entire construct is disallowed. I have reviewed the baseline changes and they're all intended.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -1690,11 +1690,7 @@ func (p *Parser) parseHeritageClause() *ast.Node {
 	kind := p.token
 	p.nextToken()
 	types := p.parseDelimitedList(PCHeritageClauseElement, (*Parser).parseExpressionWithTypeArguments)
-	result := p.finishNode(p.factory.NewHeritageClause(kind, types), pos)
-	if result.Flags&ast.NodeFlagsJavaScriptFile != 0 && kind == ast.KindImplementsKeyword {
-		p.jsErrorAtRange(result.Loc, diagnostics.X_implements_clauses_can_only_be_used_in_TypeScript_files)
-	}
-	return result
+	return p.checkJSSyntax(p.finishNode(p.factory.NewHeritageClause(kind, types), pos))
 }
 
 func (p *Parser) parseExpressionWithTypeArguments() *ast.Node {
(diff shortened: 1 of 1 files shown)


internal/parser/parser.go
```diff
@@ -6437,111 +6531,58 @@ func parseResolutionMode(mode string, pos int, end int /*reportDiagnostic: Pragm
...
+
+func (p *Parser) checkNoJSSignature(node *ast.Node) *ast.Node {
+	if node.Flags&ast.NodeFlagsJavaScriptFile != 0 && node.Flags&ast.NodeFlagsJSDoc == 0 && node.Body() == nil {
+		p.jsErrorAtRange(core.OrElse(node.Name(), node).Loc, diagnostics.Signature_declarations_can_only_be_used_in_TypeScript_files)
+	}
+	return node
+}
+
+func (p *Parser) jsErrorAtRange(loc core.TextRange, message *diagnostics.Message, args ...any) {
+	p.jsDiagnostics = append(p.jsDiagnostics, ast.NewDiagnostic(nil, core.NewTextRange(scanner.SkipTrivia(p.sourceText, loc.Pos()), loc.End()), message, args...))
```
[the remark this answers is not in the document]
isaac30609: fiquei curioso  nesse código, esse sistema seria usado em qual função

internal/parser/parser.go
```diff
@@ -358,7 +358,7 @@ func (p *Parser) parseSourceFileWorker() *ast.SourceFile {
 	}
 	collectExternalModuleReferences(result)
 	if ast.IsInJSFile(node) {
-		result.SetJSDiagnostics(getJSSyntacticDiagnosticsForFile(result))
+		result.SetJSDiagnostics(attachFileToDiagnostics(p.jsDiagnostics, result))
```
jakebailey, line 361: Do we even need the concept of JS diagnostics anymore? Could these just be parse errors?

ahejlsberg: The distinction only matters for whether we report grammar errors: If there are syntactic errors, we suppress generation of grammar errors. But JS diagnostics don't count towards this (which is reasonable given that we didn't have an issue with the actual syntax, it's just that it isn't allowed). None of this matters with the command-line compiler because we only report grammar and semantic errors when there are no syntax errors of any kind. But it does make a difference in the IDE and in tests.

ahejlsberg: BTW, I'm not sure if any of the above was ever part of a well considered plan, or just happened to be that way over time.

jakebailey: It seems weird that the effect of this is that people can use the API to parse JS files and not actually get any diagnostics. But it's not a regression so it's probably ok.

ahejlsberg: Oh, they'll get diagnostics. The `Program.GetSyntacticDiagnostics` function always returns a concatenation of regular diagnostics and JS diagnostics. It's just that grammar checks are guarded by `Checker.hasParseDiagnostics` which only checks if there are regular syntactic diagnostics (and ignores JS diagnostics). It's not particularly intuitive, but that's how it was, so that's how it is.

Issue #1646, opened 2025-08-29
LSP: changes to files not open in editor are sometimes not picked up


lukpsaxo:
Essentially after switching branch, sometimes, particularly when alot of file changes are made, tsgo seems to have an old version of project files *not* open in the editor, cached.

Issue started occurring after snapshot LSP was merged.

## Steps to reproduce

I was re-testing https://github.com/microsoft/typescript-go/issues/1265 but I have seen this just in day-2-day too.

1. switch branch - have a file open that uses something from a file edited on another branch

logs do detect file changes:

```
======== Cloning snapshot 92 ========
[08:23:47.134] DidChangeFiles
	[08:23:47.134] Checking for changes affecting config files
		[08:23:47.134] Summarizing file changes
		[08:23:47.134] Checking if any changed files are config files
[08:23:47.134] DidRequestFile
	[08:23:47.134] Acquiring config for project
	[08:23:47.134] CompilerHost
	[08:23:47.965] Program update for c:/Projects/x/y/src/frontend/tsconfig.json completed in 831.249ms
[08:23:47.965] Finished cloning snapshot 92 into snapshot 93 in 831.249ms

[08:23:48.033] 
Project 'c:/Projects/x/y/src/frontend/tsconfig.json'
	Files (15029)
-----------------------------------------------

[08:23:48.053] Updated watch: affecting locations for c:/Projects/x/y/src/frontend/tsconfig.json watcher 200
[08:23:48.053] 	C:/Projects/x/y/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[08:23:48.053] 	C:/Projects/x/y/src/openapi-clientlib/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[08:23:48.053] 	C:/Projects/x/y/src/toolkit/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[08:23:48.053] 	C:/Projects/x/y/src/ui/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[08:23:48.053] 	C:/Projects/x/y/tools/sharedtypes/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[08:23:48.053] 	c:/Projects/x/y/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[...]

PR #1483, opened 2025-07-30
Don't use a regex in IsImplicitGlob

jakebailey:
This code is a hotspot in the critical loading path. In Strada, this was all regexes, which is great in JS where regexes are fast. In Go, not so much, even moreso when we use `regexp2` in ECMAScript compatibility mode. We've always wanted to rewrite this, it's just taken a bit.

This PR moves the old implementation to a different file, then I had agent mode write tests and a new implementation, verifying against those tests.

The result so far:

```
                                          │   old.txt    │               new.txt               │
                                          │    sec/op    │   sec/op     vs base                │
MatchFiles/CommonPattern-20                 186.58µ ± 1%   25.47µ ± 2%  -86.35% (p=0.000 n=10)
MatchFiles/SimpleInclude-20                  77.13µ ± 1%   11.87µ ± 1%  -84.61% (p=0.000 n=10)
MatchFiles/EmptyIncludes-20                  29.00µ ± 2%   11.16µ ± 1%  -61.52% (p=0.000 n=10)
MatchFiles/HiddenDirectories-20              55.31µ ± 1%   13.39µ ± 1%  -75.79% (p=0.000 n=10)
MatchFiles/NodeModulesSearch-20             219.55µ ± 1%   24.57µ ± 1%  -88.81% (p=0.000 n=10)
MatchFiles/LargeFileSystem-20               23.402m ± 1%   2.043m ± 1%  -91.27% (p=0.000 n=10)
MatchFilesLarge/AllFiles-20                 20.920m ± 1%   1.831m ± 1%  -91.25% (p=0.000 n=10)
MatchFilesLarge/Components-20               2308.1µ ± 1%   677.5µ ± 1%  -70.65% (p=0.000 n=10)
MatchFilesLarge/TestFiles-20                2763.0µ ± 1%   338.4µ ± 1%  -87.75% (p=0.000 n=10)
MatchFilesLarge/NestedUtilsWithPattern-20   1212.2µ ± 2%   363.5µ ± 1%  -70.01% (p=0.000 n=10)
geomean                                      673.4µ        113.2µ       -83.19%

                                          │    old.txt    │                new.txt                 │
                                          │     B/op      │     B/op       vs base                 │
MatchFiles/CommonPattern-20                  15.83Ki ± 0%    11.05Ki ± 0%   -30.16% (p=0.000 n=10)
MatchFiles/SimpleInclude-20                  6.797Ki ± 0%    8.117Ki ± 0%   +19.42% (p=0.000 n=10)
[...]

PR #1733, opened 2025-09-19
Simplify and consolidate LSP watcher registrations

andrewbranch:
In a large internal monorepo, this cut the number of individual globs we ask the LSP client to watch from 7667 (duplicated and narrower) globs to 7 (deduped and broader) globs, and cut the time it took to compute and register those globs by 80%. This is the basis for two follow-up fixes/improvements:

1. Change our registered globs to be extensionless in order to work around https://github.com/microsoft/vscode/issues/109754
2. Clear the file cache manually when a large number of files change in a short period of time, as some events are dropped by VS Code (#1646)

I want to get reviews on this first, as those fixes will build on this simplification.


internal/project/projectcollectionbuilder.go
```diff
@@ -535,7 +535,7 @@ func (b *projectCollectionBuilder) findOrCreateDefaultConfiguredProjectWorker(
 				// For composite projects, we can get an early negative result.
 				// !!! what about declaration files in node_modules? wouldn't it be better to
 				//     check project inclusion if the project is already loaded?
-				if !config.MatchesFileName(fileName) {
+				if _, ok := config.FileNamesByPath()[path]; !ok {
```
andrewbranch, line 538: This was a drive-by fix, but `MatchesFileName` is a more expensive function that speculates about files that may not exist; the usage here was for a file that already exists and is either properly included by the config or not.

internal/project/session_test.go
```diff
@@ -548,6 +552,67 @@ func TestSession(t *testing.T) {
 			assert.Check(t, lsAfter.GetProgram() != programBefore)
 		})
 
+		t.Run("change program file not in tsconfig root files", func(t *testing.T) {
```
andrewbranch, line 555: This test case exercises logic that was missing from the pre-snapshot LSP port.

internal/project/watch.go
andrewbranch, line 1: The removals in this file represent two strategy changes:

- Since we rely on the LSP client to handle the watchers at the system level, remove our logic for avoiding installing watchers too close to the file system root. VS Code installs a watcher at the workspace level no matter that directory’s location, so we have nothing to gain by not using it. Any decision to refuse installing a watcher at a given location belongs to the client; they can simply return an error when we request to watch that location.
- Prefer watching the workspace directory over anything else. If a file is in the workspace, don't register any additional watcher.

internal/tsoptions/parsedcommandline.go
```diff
@@ -285,48 +313,30 @@ func (p *ParsedCommandLine) GetConfigFileParsingDiagnostics() []*ast.Diagnostic
 	return p.Errors
 }
 
-// Porting reference: ProjectService.isMatchedByConfig
-func (p *ParsedCommandLine) MatchesFileName(fileName string) bool {
```
andrewbranch, line 289: MatchesFileName was no longer being used except for the incorrect usage I pointed out above. When I initially implemented it for file watching, it was a bad performance bottleneck. PossiblyMatchesFileName is less precise but faster. It may be worth revisiting this if we invest into perf improvements of our include/exclude globbing (#1483)

internal/lsp/server.go
```diff
@@ -651,9 +640,27 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 		s.watchEnabled = true
 	}
 
+	cwd := s.cwd
+	if s.initializeParams.Capabilities != nil &&
```
jakebailey, line 656: Boy do I want to add protobuf-style safe accessors for capabilities

jakebailey:
I think this seems fine, but are we likely to want some of these heuristics back when we eventually do "real" watch mode for the CLI using `fsnotify`?

jakebailey:
I guess the other concern I have is the part about watching all workspace files regardless of extension. I get that to VS Code, there's no difference, but we are still going to be sent all of those events...

I am sort of wondering if we should think about accelerating our trials of fsnotify to see if it works better than VS Code / the LSP client...

andrewbranch:
I think it’ll be fine. We already handle batches of 30k+ changes during big npm installs fine, (and we’re going to end up ignoring huge batches like that and reloading everything in a future change anyway); I doubt opening things up to other file extensions will change things that much in projects that are already predominantly TS/JS.

internal/lsp/server.go
```diff
@@ -639,9 +640,27 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
...
+		s.initializeParams.Capabilities.Workspace != nil &&
+		s.initializeParams.Capabilities.Workspace.WorkspaceFolders != nil &&
+		ptrIsTrue(s.initializeParams.Capabilities.Workspace.WorkspaceFolders) &&
+		s.initializeParams.WorkspaceFolders != nil &&
+		s.initializeParams.WorkspaceFolders.WorkspaceFolders != nil &&
+		len(*s.initializeParams.WorkspaceFolders.WorkspaceFolders) == 1 {
+		cwd = lsproto.DocumentUri((*s.initializeParams.WorkspaceFolders.WorkspaceFolders)[0].Uri).FileName()
```
sheetalkamat, line 663: i think we should use all the workspace folders here? 
https://github.com/microsoft/typescript-go/pull/1238 where i had the change to handle this where we could  infer project and its setting based on what path it is.

sheetalkamat: And also need to handle : `lsproto.DidChangeWorkspaceFoldersParams: `

sheetalkamat: it would also help with not looking past certain directory when finding tsconfig etc

andrewbranch: Yeah, this is definitely incomplete but I want to handle multi-root workspace support all at once in a future PR. In theory, this shouldn’t break watching in other workspace folders, because we always check if the directories we need to watch are inside this directory, and fall back to other locations if not. But I’ve literally never tried tsgo in a multi-root workspace so I have no idea if anything works at all, at the moment.

sheetalkamat: what i mean is instead of using "cwd" as the deciding factor for watching strategy it would be nice to instead have something else handle that. I would rather now change cwd as part of this and keep that in play when nothing else is there? You can handle single workspace for now (and add multiple places to consider later) off of that option rather than changing cwd.

andrewbranch: I do think this is kind of overloading the term "cwd" in an incorrect way, so I can rename/add a field to use instead. However, note that the actual process CWD was always `/` due to the way vscode-languageclient spawns the process. I thought about adding a CWD in the spawn options there, but I didn’t think there was an option that made sense, and we never use it for anything since every path that gets communicated over LSP is an absolute URI.

jakebailey: Yeah, the _actual_ cwd of the process should never be used for any reason.

jakebailey:
I went to try this out quick and was surprised by these watches:

```
[14:04:33.047] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.1
[14:04:33.047] 	/home/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[14:04:33.047] 
[14:04:33.049] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.2
[14:04:33.049] 	/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[14:04:33.049] 
```

That first one is saying to watch all files in all user dirs... As far as I know, editors will totally attempt to watch these and probably exhaust limits. Are we sure we don't want to cut off certain dirs or something?

andrewbranch:
Hm, either there’s a bug in the logic that gets the common parents of the failed lookup locations, or lookups occurred in multiple users’ home directories. I should probably make the minimum depth stricter for directories that fall outside of the workspace directory, but do you think you could grab the list of failed lookups that generated that watch so I can see if the logic is right?

jakebailey:
https://gist.github.com/jakebailey/84d5310acdb3397056009e0bbf998070 The big list of paths is me logging all of the files from `createResolutionLookupGlobMapper` (sorted).

andrewbranch:
@jakebailey could you try the latest? I walked back some of the deletions while keeping it still much simpler than before.

jakebailey:
Seems better:

```
[13:46:59.337] Added new watch: root files for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 9.0
[13:46:59.337] 	/home/jabaile/work/TypeScript/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.337] 
[13:46:59.344] Added new watch: non-root program files for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 10.1
[13:46:59.344] 	/home/jabaile/work/TypeScript-go/built/local/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.344] 
[13:46:59.352] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.0
[13:46:59.352] 	/home/jabaile/work/TypeScript/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.352] 
[13:46:59.353] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.1
[13:46:59.353] 	/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.353] 
[13:46:59.354] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.2
[13:46:59.355] 	/home/jabaile/work/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.355] 
[13:46:59.356] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.3
[13:46:59.356] 	/home/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.356] 
[13:46:59.357] Added new watch: failed lookups for /home/jabaile/work/TypeScript/src/compiler/tsconfig.json watcher 12.4
[13:46:59.357] 	/home/jabaile/node_modules/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[13:46:59.357] 
[13:46:59.357] Updated watches in 62.427381ms
```

Those are all definitely safe paths to watch.

jakebailey:
Though, the workspace ones seem potentially redundant? (Depending on how the editor does things)

andrewbranch:
It's not reflected in the logs here, but they appear duplicated because the failed lookups one is a create-only watcher, while the root files for the tsconfig one is a create/change/delete watcher. In theory, one of those could be independently removed if you start opening and closing other projects. We could make a special rule that says “any watcher on the workspace directory is always create/change/delete” but I’m not sure how worthwhile that is.

jakebailey:
This is a weird one; I opened DT tools to `packages/mergebot/src/pr-info.ts`, then opened `packages/publisher/src/lib/secrets.ts` and got this new watch:

```
[14:01:36.107] Added new watch: failed lookups for /home/jabaile/work/DefinitelyTyped-tools/packages/publisher/tsconfig.json watcher 37.5
[14:01:36.107] 	/home/jabail/typescript/**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}
[14:01:36.107] 
[14:01:36.108] Updated watches in 13.083169ms
```

But, that path doesn't exist and isn't mentioned anywhere else in the logs. This is a failed lookup so maybe there's a bad import somewhere that's not resolving, but I can't seem to find it.

jakebailey:
Just realized, `/home/jabail/typescript/` doesn't make sense because my username is `jabaile`. So definitely some string weirdness.

andrewbranch:
It’s happening to me too, but it’s not in a bug in the watcher code; that’s an actual module resolution lookup 😮 

image

andrewbranch:
It looks like the same bug exists in Strada. I prevented the bogus path lookup for now but I’m not sure if a different strategy should be implemented to find the proper peer dep lookup location, or if it doesn’t matter for monorepo packages like this. (@sheetalkamat you might want to look at the last commit)

jakebailey (approved):
I can't find any weird issues when I try this out and read the logs. Worth a shot!

jakebailey:
I'll note that in the TS repo, if I delete `node_modules`, nothing changes; there's no event for this. But, that happens on main, too. Which I believe is that known VS Code issue.

andrewbranch:
Yeah, that’s https://github.com/microsoft/vscode/issues/109754, and will be handled in a follow-up PR to remove the extensions from our glob patterns.

andrewbranch:
I merged main because license/cla was MIA, but I have not super high hopes that a new commit is going to unstick it...

jakebailey:
A close reopen usually fixes it for me.

PR #1753, opened 2025-09-26
fix: replace tsc with tsgo

Bashamega:
I think this is a typing mistake. It should be tsgo, not tsc

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ Additional tasks are a work in progress.
 
 ### Running `tsgo`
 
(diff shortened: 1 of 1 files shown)


JoostK:
`tsc` is intended here; `tsgo` is a work-in-progress binary that replicates the workings of `tsc`, hence the comment that it behaves similarly.

Bashamega:
> `tsc` is intended here; `tsgo` is a work-in-progress binary that replicates the workings of `tsc`, hence the comment that it behaves similarly.

I don't think so, since this is for running tsgo

jakebailey:
No, JoostK is correct

PR #1756, opened 2025-09-27
fix: add bounds checks to prevent slice bounds out of range panic

arash-mosavi:


arash-mosavi:

internal/ls/completions.go
iisaduan, line 1: The added checks in this file shouldn't be necessary since they happen at the very end of any completions requests. At that point, if `position` is out of bounds here, completions should have caught it and returned early, or else it means there is a deeper bug (in the ls infrastructure for example) that makes something else out of sync. Overall though, I'm not against keeping them for safety (cc @gabritto)

arash-mosavi: Thanks for the review! I’ve removed the completions changes and kept this PR scoped to **autoimports** only, per feedback.
Scope is now minimal and fixes the crash without masking any deeper invariants.

Kindly requesting another look

jakebailey:
These fixes could be plausible, but need fourslash tests to verify.

andrewbranch:
Superseded by #2390

PR #1757, opened 2025-09-27
Fix translation error that disabled union discrimination optimization

### LINKED ISSUE #1727: tsc vs tsgo: inference mismatch of parameter in passed callback
## Steps to reproduce

```
mkdir tsgo-vs-tsc
cd tsgo-vs-tsc
npm init -y
npm pkg set type=module
npm i typescript @typescript/native-preview css-tree @types/css-tree -D
npx tsc --init --module nodenext --moduleResolution nodenext
printf "import { parse, walk } from 'css-tree';\nconst declarationsText = '...';\nconst ast = parse(declarationsText, { context: 'declarationList', positions: true });\n\nwalk(ast, {\n    visit: 'Declaration',\n    enter(node) {\n        console.log(node);\n    },\n});" > main.ts
npx tsc --noEmit
npx tsgo --noEmit
```
In case your terminal doesn't support `printf`, the contents of `main.ts`:
```ts
import { parse, walk } from 'css-tree';
const declarationsText = '...';
const ast = parse(declarationsText, { context: 'declarationList', positions: true });

walk(ast, {
    visit: 'Declaration',
    enter(node) {
        console.log(node);
    },
});
```
No type errors. `node` is inferred as `Declaration`.

## Behavior with `tsgo`
```
main.ts:7:11 - error TS7006: Parameter 'node' implicitly has an 'any' type.

7     enter(node) {
            ~~~~


Found 1 error in main.ts:7
```

ahejlsberg:
Implements what I describe [here](https://github.com/microsoft/typescript-go/issues/1727#issuecomment-3341831052).

Fixes #1727.

### DIFF
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -1167,23 +1167,21 @@ func (c *Checker) mapTypesByKeyProperty(types []*Type, keyPropertyName string) m
 	for _, t := range types {
 		if t.flags&(TypeFlagsObject|TypeFlagsIntersection|TypeFlagsInstantiableNonPrimitive) != 0 {
 			discriminant := c.getTypeOfPropertyOfType(t, keyPropertyName)
-			if discriminant != nil {
-				if !isLiteralType(discriminant) {
-					return nil
+			if discriminant == nil || !isLiteralType(discriminant) {
+				return nil
+			}
+			duplicate := false
+			for _, d := range discriminant.Distributed() {
+				key := c.getRegularTypeOfLiteralType(d)
+				if existing := typesByKey[key]; existing == nil {
+					typesByKey[key] = t
+				} else if existing != c.unknownType {
+					typesByKey[key] = c.unknownType
+					duplicate = true
 				}
-				duplicate := false
-				for _, d := range discriminant.Distributed() {
-					key := c.getRegularTypeOfLiteralType(d)
-					if existing := typesByKey[key]; existing == nil {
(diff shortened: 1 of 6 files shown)


internal/checker/relater.go
```diff
@@ -1167,24 +1167,22 @@ func (c *Checker) mapTypesByKeyProperty(types []*Type, keyPropertyName string) m
...
-					if existing := typesByKey[key]; existing == nil {
-						typesByKey[key] = t
-					} else if existing != c.unknownType {
-						typesByKey[key] = c.unknownType
-						duplicate = true
-					}
-				}
-				if !duplicate {
-					count++
+			if discriminant == nil || !isLiteralType(discriminant) {
```
jakebailey, line 1170: Doesn't this behave differently from the original code? The original code would still continue the loop of the discriminate was undefined.

ahejlsberg: Yes, but the exit is intended. We only want to build a map when all constituents have a discriminant. When some constituents lack a discriminant, we want to skip the map optimization and go through the regular `discriminateTypeByDiscriminableItems` code path.

jakebailey (approved):

jakebailey:
Hm, https://github.com/microsoft/TypeScript/pull/62501#issuecomment-3343468049 shows a break in DT

jakebailey (approved):

PR #1765, opened 2025-09-29
fix: code style in emitHost.go file

KeoFoxy:
Now all functions in `emitHost.go` file have the same code style


jakebailey:
I don't think we need to do this; we do both of these sometimes (see ast.go).

PR #1767, opened 2025-09-29
Port source map position mapping + go to definition source mapping

gabritto:
In this PR:
- Port of `DocumentPositionMapper` from Strada
- Make go to definition follow source maps
- LS/snapshot changes to make it all work

The most notable change is that now snapshots are allowed to read a file from the file system. That file is then cached, such that all reads of that file return the same result. This is needed because we need to read source map files, as those are not part of a `Program`. Document position mappers are cached in the language service so we don't have to rebuild them multiple times per LS. In Strada, this all used to happen in the project or project service. We used to cache source map files, and therefore we also had to watch them, which greatly complicates things. By allowing snapshot/LS to read a file and cache information about it just for the duration of the snapshot/LS, we don't need to watch those files.

This PR supersedes #1710 (and is inspired by it).

### DIFF
--- a/internal/project/snapshot.go
+++ b/internal/project/snapshot.go
@@ -109,10 +109,6 @@ func (s *Snapshot) ReadFile(fileName string) (string, bool) {
 	return handle.Content(), true
 }
 
-func (s *Snapshot) FileExists(fileName string) bool {
-	return s.fs.fs.FileExists(fileName)
-}
-
 type APISnapshotRequest struct {
 	OpenProjects   *collections.Set[string]
 	CloseProjects  *collections.Set[tspath.Path]
--- a/internal/ls/host.go
+++ b/internal/ls/host.go
@@ -5,7 +5,6 @@ import "github.com/microsoft/typescript-go/internal/sourcemap"
 type Host interface {
 	UseCaseSensitiveFileNames() bool
 	ReadFile(path string) (contents string, ok bool)
-	FileExists(path string) bool
 	Converters() *Converters
 	GetECMALineInfo(fileName string) *sourcemap.ECMALineInfo
 }
--- a/internal/ls/source_map.go
+++ b/internal/ls/source_map.go
@@ -54,7 +54,7 @@ func (l *LanguageService) tryGetSourcePosition(
 ) *sourcemap.DocumentPosition {
 	newPos := l.tryGetSourcePositionWorker(fileName, position)
 	if newPos != nil {
-		if !l.host.FileExists(newPos.FileName) {
+		if _, ok := l.ReadFile(newPos.FileName); !ok { // File doesn't exist
 			return nil
 		}
 	}


internal/ls/languageservice.go
```diff
@@ -4,17 +4,32 @@ import (
...
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/sourcemap"
 )
 
 type LanguageService struct {
-	host       Host
-	converters *Converters
+	host                      Host
+	converters                *Converters
+	documentPositionMappers   map[string]*sourcemap.DocumentPositionMapper
```
gabritto: I think this doesn't need to be a `syncMap` for now as we create an LS per request, but that can be modified later if we need to share an LS.

internal/ls/languageservice.go
```diff
@@ -4,17 +4,32 @@ import (
...
+	useCaseSensitiveFileNames bool
+	readFile                  func(path string) (contents string, ok bool)
+	fileExists                func(path string) bool
 }
 
-func NewLanguageService(host Host, converters *Converters) *LanguageService {
+func NewLanguageService(
+	host Host,
+	converters *Converters,
+	readFile func(path string) (contents string, ok bool),
```
gabritto: One clunky thing about this PR: ideally `readFile`, `fileExists` and `useCaseSensitiveFileNames` would be part of the LS `Host`. But today we pass a `project` as host, and those FS functions don't come from a project, but instead from a snapshot.

internal/scanner/scanner.go
```diff
@@ -2451,31 +2452,42 @@ func GetPositionOfLineAndCharacter(sourceFile *ast.SourceFile, line int, charact
...
+}
+
+func ComputePositionOfLineAndCharacterEx(lineStarts []core.TextPos, line int, character int, text *string, allowEdits bool) int {
 	if line < 0 || line >= len(lineStarts) {
-		// if (allowEdits) {
-		//     // Clamp line to nearest allowable value
-		//     line = line < 0 ? 0 : line >= lineStarts.length ? lineStarts.length - 1 : line;
-		// }
-		panic(fmt.Sprintf("Bad line number. Line: %d, lineStarts.length: %d.", line, len(lineStarts)))
+		if allowEdits {
```
gabritto, line 2460: This is a direct port of Strada and allows cases where the source file and source map are out of sync, see test [declarationMapsOutOfDateMapping](https://github.com/microsoft/typescript-go/pull/1767/files#diff-f0b11a863f9b518f7f0e62c4c22b6d252316c5f8e86834d58cb4ac3b8346ba3c).

internal/ls/languageservice.go
jakebailey: This definitely feels clunky... why doesn't Project have these methods, if it was created from a snapshot?

(I think we should at least make a host interface to accept or something)

gabritto: At the time we're creating a project, we have a snapshot builder, not a snapshot, so it also feels weird to me to add the methods to project.

internal/sourcemap/source_mapper.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func tryGetSourceMappingURL(host Host, fileName string) string {
+	lineInfo := host.GetLineInfo(fileName)
+	return TryGetSourceMappingURL(lineInfo)
+}
+
+// Originally: /^data:(?:application\/json;charset=[uU][tT][fF]-8;base64,([A-Za-z0-9+/=]+)$)?/
+// Should have been /^data:(?:application\/json;(?:charset=[uU][tT][fF]-8;)?base64,([A-Za-z0-9+/=]+)$)?/
+func tryParseBase46Url(url string) (parseableUrl string, isBase64Url bool) {
```
gabritto: I think the Strada regex that this function is replacing had a problem. The regex `base64UrlRegExp` as written required a `charset=utf-8` to always be specified for us to decode an inline source map. But we don't even specify a `charset` when we write source maps in the emitter, so this means we wouldn't parse our own source map urls.

internal/sourcemap/source_mapper.go
```diff
@@ -0,0 +1,314 @@
+package sourcemap
```
gabritto, line 1: This file is also basically a close port of Strada.

internal/ls/languageservice.go
gabritto: Following Andrew's suggestion, I refactored this a bit and added those FS methods to Host, and now the LS takes in a program separately. So the project provides the LS with a program, and the snapshot implements a host with everything else.

internal/ls/languageservice.go
```diff
@@ -36,3 +44,32 @@ func (l *LanguageService) getProgramAndFile(documentURI lsproto.DocumentUri) (*c
...
+func (l *LanguageService) GetLineInfo(fileName string) *sourcemap.LineInfo {
+	text, ok := l.ReadFile(fileName)
+	if !ok {
+		return nil
+	}
+	lineMap := l.converters.getLineMap(fileName)
+	if lineMap == nil {
+		return nil
+	}
+	return sourcemap.CreateLineInfo(text, lineMap.LineStarts)
```
jakebailey: This is being provided to the source map package, but the line map being handed to it is derived from the LSP line map, which will be different. I'm pretty sure the line maps given to source mapping need to be the same ones used by the scanner / are on SourceFile.

There's going to be a lot of position conversion to try and get the edge cases working.

internal/sourcemap/source_mapper.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/debug"
+	"github.com/microsoft/typescript-go/internal/scanner"
+	"github.com/microsoft/typescript-go/internal/stringutil"
+	"github.com/microsoft/typescript-go/internal/tspath"
+)
+
+type Host interface {
+	UseCaseSensitiveFileNames() bool
+	GetLineInfo(fileName string) *LineInfo
```
jakebailey: I don't know if this abstraction will work; the line map here needs to be the mapping provided by `core`.

I think it might be time to sit down and get these two implementations renamed or something to make it clearer when to use / not use them.

jakebailey: (It could be fine, but the implementor needs to provide the correct one, since source maps are based on the line positions by ecmascript)

internal/project/snapshotfs.go
```diff
@@ -37,6 +43,15 @@ func (s *snapshotFS) GetFile(fileName string) FileHandle {
 	if file, ok := s.diskFiles[s.toPath(fileName)]; ok {
 		return file
 	}
+	newEntry := memoizedDiskFile(sync.OnceValue(func() *diskFile {
```
andrewbranch, line 46: This is probably fine, but I had imagined a change to `cachedvfs.FS` instead. Was there a reason not to put the cache there? I’m not sure why there wasn’t a cache there to begin with—presumably the result of every ReadFile gets stored somewhere in memory, and my memory is that the underlying memory for the string will be shared when it’s assigned from the ReadFile result to the cache to the diskFile to the SourceFile and so on.

gabritto: There were other usages of `cachedvfs.FS` where I wasn't sure caching file reads would make sense or how it would affect those usages, and Sheetal seemed to think we shouldn't cache files at that level.

sheetalkamat: yeah caching at cached.vfs level means you need to clear it on "writeFile" otherwise tsc -b will run into issues if we dont handle that well - which would add to cost right? and normally in those scenarios the text gets cached in "sourceFile" so another cache is generally not needed

gabritto: Thanks for clarifying!

andrewbranch: I guess that makes sense; there are already some weird inconsistencies with cachedvfs (FileExists can be out of sync with Stat can be out of sync with ReadFile) but I guess we can avoid growing those any more.

internal/project/snapshot.go
```diff
@@ -89,6 +89,22 @@ func (s *Snapshot) ID() uint64 {
...
+func (s *Snapshot) ReadFile(fileName string) (string, bool) {
+	handle := s.GetFile(fileName)
+	if handle == nil {
+		return "", false
+	}
+	return handle.Content(), true
+}
+
+func (s *Snapshot) FileExists(fileName string) bool {
+	return s.fs.fs.FileExists(fileName)
```
andrewbranch: I think this should check the snapshotfs diskFile cache first. It seems like there could be cases where something in the snapshot's file cache gets deleted from disk, and the underlying cachedfvfs doesn’t have a FileExists cache entry for it, so you get a `false` when there’s actually a file in the snapshot. (Or vice versa)

internal/sourcemap/source_mapper.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	var possibleMapLocations []string
+	if mapFileName != "" {
+		possibleMapLocations = append(possibleMapLocations, mapFileName)
+	}
+	possibleMapLocations = append(possibleMapLocations, generatedFileName+".map")
+	for _, location := range possibleMapLocations {
+		mapFileName := tspath.GetNormalizedAbsolutePath(location, tspath.GetDirectoryPath(generatedFileName))
+		if mapFileContents, ok := host.ReadFile(mapFileName); ok {
```
sheetalkamat, line 250: problem with having these in LS and not project layer and watching these files is that "navigation": leads to discarding these files and having to reparse source maps etc and use to be bottle neck early on when it lived in LS. having it life of scriptInfo its associated with or till things change, meant we could avoid having to parse things again and again and that really helped with perf

gabritto: How could this work with the new snapshot architecture? I couldn't figure out exactly where this should live in Corsa. It used to live in `ProjectService` in Strada, and we don't have that anymore.

sheetalkamat: it would live in session as thats kind of projectService now

andrewbranch: Currently, the file cache just gets passed from Snapshot to Snapshot during cloning. At the beginning of each clone, the list of files invalidated by a watch change gets refreshed in the new iteration of the cache, and at the end of clones where the set of files for any project was reevaluated, we drop files that are no longer referenced (i.e., they were seen by a ReadFile) by any project. If we want to persist map files, it would be best to have them work roughly the same way:

- they should be read in during snapshot building
- they should be passed to the next snapshot during cloning
- they need to be watched so changed files can be refreshed during that clone
- we need a way of determining when they're no longer needed so they can be dropped from the file cache

Practically, I think this means making these changes:

- undo the `readFiles` change in snapshotfs
- instead, the caller of LS functionality that might need source mapping needs to check the result, see if mapping is needed, and go back to the session and ask for a new snapshot with the requested map files present, and then do the mapping on top of that snapshot; pseudocode:
   ```go
   ls, err := session.GetLanguageService(goToDefRequestURI)
   result := ls.ProvideDefinitions(goToDefRequestURI, /*...*/)
   if dtsFileNamesToMap := getDtsFileNamesToMap(result); len(dtsFileNamesToMap) > 0 {
     mapper := session.GetSourceMapper(dtsFileNamesToMap)
     // Session will check if the necessary map files are already
     // in the current snapshot, and return a source mapper backed
     // by that snapshot if so. Otherwise, it will clone the snapshot
     // in order to include those files, and return a source mapper
     // backed by the new snapshot.
     return mapMappablePositions(result, mapper)
   }
   return result
   ```
   (Technically, this should capture the snapshot backing `ls` and use/clone _that_ one to back the mapper, not whichever one happens to be current in the session at the end of `ProvideDefinitions`.)
- During the snapshot clone that reads in new map files, the `*diskFile` that represents its input source file should also get a field written with the map file’s path (like ScriptInfo in Strada used to have). When the input file gets cleared from the snapshotfs file cache, its map file should get cleared with it.
- Watches may need to be updated when adding/removing maps (wait for #1733)

That’s a pretty big lift, so I would not block this PR on it, personally.

jakebailey: I am happy with a simpler solution for now even if it's slower. So if this PR looks good, then that's good in my eyes

jakebailey (approved):

gabritto:
I'm gonna merge this so we have something working for users while I try to get the reusable cache version working, because I don't know how long it's gonna take.

PR #1770, opened 2025-09-30
Organize imports for TypeScript and JavaScript

### LINKED ISSUE #1724: "Sort Imports" & "Remove Unused Imports" commands are missing
## Steps to reproduce
Open any TS file in vscode and search to the Sort Import command in the command palette. 

Image

## Behavior with `tsgo`
Command missing in vscode

saixiaoxi:
In this PR:

- Port organize imports functionality for TypeScript and JavaScript
- Preserve blank line groupings to respect user's logical organization
- Support all import types: side-effect, type-only, namespace, default, named, and require

Imports are sorted by:
1. Module type: absolute imports before relative imports
2. Import kind: side-effect → type-only → namespace → default → named → require
3. Alphabetically within each category

The feature is exposed via LSP `workspace/executeCommand` with command `typescript-go.organizeImports` and integrated into the VS Code extension as the "Sort Imports" command.

Fixes #1724

### DIFF
--- a/_extension/package.json
+++ b/_extension/package.json
@@ -85,6 +85,12 @@
                 "title": "Show LSP Trace",
                 "enablement": "typescript.native-preview.serverRunning",
                 "category": "TypeScript Native Preview"
+            },
+            {
+                "command": "typescript.native-preview.sortImports",
+                "title": "Sort Imports",
+                "enablement": "typescript.native-preview.serverRunning && editorLangId =~ /^(javascript|typescript|javascriptreact|typescriptreact)$/",
+                "category": "TypeScript Native Preview"
             }
         ]
     },
--- a/_extension/src/client.ts
+++ b/_extension/src/client.ts
@@ -158,4 +158,14 @@ export class Client {
         this.client.restart();
         return new vscode.Disposable(() => {});
     }
+
+    async executeCommand(command: string, ...args: any[]): Promise<any> {
+        if (!this.client) {
+            throw new Error("Language client is not initialized");
+        }
+        return this.client.sendRequest("workspace/executeCommand", {
+            command,
+            arguments: args,
+        });
+    }
 }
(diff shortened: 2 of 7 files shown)


saixiaoxi:

jakebailey:
I don't think this is the right approach. The language server protocol already has organize imports. We should not require anything custom to make this work. Adding anything to the client implies we are going outside the LSP spec.

saixiaoxi:
> I don't think this is the right approach. The language server protocol already has organize imports. We should not require anything custom to make this work. Adding anything to the client implies we are going outside the LSP spec.

I want to stick to the LSP spec too, but I'm a little confused - the `typescript.sortImports` command that I'm trying to implement isn't actually part of the LSP specification. The official TypeScript extension implements it as a custom workspace command (see their [package.json](https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/package.json#L200)). The only standard LSP code action for imports is `source.organizeImports`, which removes unused imports. So I don't know how to provide the non-destructive "sort only" behavior without going outside the LSP spec?

jakebailey:
That's fair, thanks. 

I'm not totally sure how to do this nicely when the same command is registered by the main TS extension...

jakebailey:
It's been a while, sorry. You probably want to merge with main and then see if anything you've done here has already been done on main. I see a bunch of helpers that I think we might already have declared elsewhere.

As far as I know, we can actually just register the same command name as the other TS extension and use it, but otherwise we can probably just do what you have here with the `when` clause.

mkantor:
If I remember to check again once this lands I'll open a proper issue, but in case I forget: does microsoft/TypeScript#60431 still happen with this implementation? I'm guessing it does since this is a port.

jakebailey:
I think #2331 supersedes this?

jakebailey:
This is an old PR; if anything, all that's left is a simple command, e.g. #3305.

Issue #1578, opened 2025-08-13
Fix GetLineAndCharacterOfPosition and audit all users


jakebailey:
Things are very weird with `GetLineAndCharacterOfPosition` and its users.

- `GetLineAndCharacterOfPosition` returns col position in terms of _rune count_. This could be fine, but some callers assume that it's a byte offset and then break on non-ASCII code. Most callers don't notice this because they actually only want the line number.
- `GetLineAndCharacterOfPosition` uses `GetLineStarts`, but that uses line separators as defined by ECMAScript. This can be fine, but should really only be done for things like parsing and source maps. But, we have users of this in the language server, and the language server should be using the LSP definition of newlines (`\n`, `\r\n`, `\r`).

We should redo these, likely renaming funcs or something to make it very obvious what each of them are actually doing.

See also:

- https://github.com/microsoft/typescript-go/pull/1561/files#r2271932196
- #1525

PR #1786, opened 2025-10-01
Rename line maps with ECMA/LSP prefix

gabritto:
Renaming line map-related functions to make the distinction clear between ECMA and LSP lines.
ECMA line maps/line starts consider characters other than `\n` and `\r\n` as line breaks (see `stringutil.IsLineBreak`), unlike LSP.
This should help avoid future incorrect usages of those functions, and should make it easier to do #1578.

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -10724,10 +10724,10 @@ type SourceFile struct {
 	ClassifiableNames         collections.Set[string]
 	PatternAmbientModules     []*PatternAmbientModule
 
-	// Fields set by LineMap
+	// Fields set by ECMALineMap
 
-	lineMapMu sync.RWMutex
-	lineMap   []core.TextPos
+	ecmaLineMapMu sync.RWMutex
+	ecmaLineMap   []core.TextPos
 
 	// Fields set by language service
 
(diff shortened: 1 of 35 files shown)


internal/format/indent.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func GetIndentationForNode(n *ast.Node, ignoreActualIndentationRange *core.TextRange, sourceFile *ast.SourceFile, options *FormatCodeSettings) int {
-	startline, startpos := scanner.GetLineAndCharacterOfPosition(sourceFile, scanner.GetTokenPosOfNode(n, sourceFile, false))
+	startline, startpos := scanner.GetECMALineAndCharacterOfPosition(sourceFile, scanner.GetTokenPosOfNode(n, sourceFile, false))
```
jakebailey, line 16: The formatter using these instead of the client encoding is definitely spooky

internal/ast/ast.go
```diff
@@ -10724,10 +10724,10 @@ type SourceFile struct {
 	ClassifiableNames         collections.Set[string]
 	PatternAmbientModules     []*PatternAmbientModule
 
-	// Fields set by LineMap
+	// Fields set by ECMALineMap
 
-	lineMapMu sync.RWMutex
-	lineMap   []core.TextPos
+	ecmaLineMapMu sync.RWMutex
+	ecmaLineMap   []core.TextPos
```
jakebailey, line 10730: Should we give `[]core.TextPos` a name like `ECMALineMap` or does a change like that cause problems elsewhere?

PR #1787, opened 2025-10-01
Permit expando property function with return type annotation as assertion

### LINKED ISSUE #1774: `tsgo` produces `TS2775` for function with assertion function assigned to a property but `tsc` does not
## Steps to reproduce

Create a new TS project:

```
npm init -y
npm add @typescript/native-preview
npm add typescript
echo '{}' > tsconfig.json
```

Create a file, `demo.ts`:

```ts
function example() {}

example.isFoo = function isFoo(value: string): asserts value is 'foo' {
  if (value !== 'foo') {
    throw new Error('Not foo');
  }
};

example.isFoo('test');
```

Run `node_modules/.bin/tsc` — no errors.

## Behavior with `tsgo`

Run `node_modules/.bin/tsgo` — see:

```
demo.ts:9:1 - error TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.

9 example.isFoo('test');
  ~~~~~~~~~~~~~

  demo.ts:3:1 - 'isFoo' needs an explicit type annotation.
      3 example.isFoo = function isFoo(value: string): asserts value is 'foo' {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      4   if (value !== 'foo') {
        ~~~~~~~~~~~~~~~~~~~~~~~~
    ...
      6   }
        ~~~
      7 };
        ~


Found 1 error in demo.ts:9
```

ahejlsberg:
Fixes #1774.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -8053,7 +8053,6 @@ func (c *Checker) checkCallExpression(node *ast.Node, checkMode CheckMode) *Type
 		if !ast.IsDottedName(node.Expression()) {
 			c.error(node.Expression(), diagnostics.Assertions_require_the_call_target_to_be_an_identifier_or_qualified_name)
 		} else if c.getEffectsSignature(node) == nil {
-			c.error(node.Expression(), diagnostics.Assertions_require_every_name_in_the_call_target_to_be_declared_with_an_explicit_type_annotation)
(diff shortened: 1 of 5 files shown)


wincent:
Thanks for the quick fix! 🙇🏻

PR #1793, opened 2025-10-02
added userpref gating for package.json auto imports

johnfav03:


internal/ls/autoimportsexportinfo.go
```diff
@@ -25,6 +25,8 @@ func (l *LanguageService) getExportInfoMap(
 	forEachExternalModuleToImportFrom(
 		ch,
 		l.GetProgram(),
+		l.UserPreferences(),
+		l.UseCaseSensitiveFileNames(),
```
jakebailey: Does this need to be passed explicitly? I figure that Program already has this.

internal/lsp/server.go
```diff
@@ -939,8 +939,8 @@ func (s *Server) handleDocumentSymbol(ctx context.Context, ls *ls.LanguageServic
 	return ls.ProvideDocumentSymbols(ctx, params.TextDocument.Uri)
 }
 
-func (s *Server) handleRename(ctx context.Context, ls *ls.LanguageService, params *lsproto.RenameParams) (lsproto.RenameResponse, error) {
-	return ls.ProvideRename(ctx, params)
+func (s *Server) handleRename(ctx context.Context, languageServer *ls.LanguageService, params *lsproto.RenameParams) (lsproto.RenameResponse, error) {
+	return languageServer.ProvideRename(ctx, params, lsutil.NewDefaultUserPreferences())
```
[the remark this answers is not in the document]
jakebailey: Yeah, should this just use the regular prefs?

(Eventually, this will I think be plumbed via ctx)

internal/modulespecifiers/specifiers.go
```diff
@@ -461,6 +461,7 @@ func getLocalModuleSpecifier(
 	preferences ModuleSpecifierPreferences,
 	pathsOnly bool,
 ) string {
+	baseUrl := compilerOptions.BaseUrl
```
jakebailey: baseUrl has been removed and should never be used for anything. Are you seeing this in a test that isn't skipped?

internal/modulespecifiers/types.go
```diff
@@ -88,6 +88,7 @@ const (
 type UserPreferences struct {
 	ImportModuleSpecifierPreference   ImportModuleSpecifierPreference
 	ImportModuleSpecifierEnding       ImportModuleSpecifierEndingPreference
+	allowRenameOfImportPath           core.Tristate
```
jakebailey: Unexported?

internal/modulespecifiers/specifiers.go
johnfav03: I used baseUrl for testing the `project-relative` option for `importModuleSpecifierPreference`. Is this an option I should ignore, or how would you approach this?

jakebailey: IIRC `project-relative` also works for import maps, like in `autoImportPackageJsonImportsPreference2.ts`. `baseUrl` shouldn't be required for it, I don't think?

If the test you're looking at used `baseUrl` before, then you can probably ignore it. If it's a new test, I'd be surprised given I had figured we got all of them in convertFourslash, but maybe not everything is converting such that you can test it with existing tests?

johnfav03: I wrote the test myself - I'll look into using an import map instead of a baseUrl

andrewbranch: `paths` should work to test it too—I think a realistic scenario would be multiple projects extend a shared tsconfig with something like

```json
{
  "compilerOptions": {
    "paths": {
      "@app/*": ["./packages/app/src/*"],
      "@lib/*": ["./packages/lib/src/*"],
      "@server/*": ["./packages/server/src/*"]
    }
  }
}
```

Then the assertion is that from any file within `packages/app/src`, you should never have an import of `"@app/anything"`, but you should always import from other projects like `"@lib/foo"` or `"@server/bar"`. For files within the `app` project, you should get imports like `./util/foo.js`.

An import map may work too, I just don't recall how that plays with the rest of the logic.

jakebailey:
Is the test generator just missing these options such that all tests are new handwritten ones?

internal/lsp/server.go
```diff
@@ -949,8 +949,8 @@ func (s *Server) handleDocumentSymbol(ctx context.Context, ls *ls.LanguageServic
 	return ls.ProvideDocumentSymbols(ctx, params.TextDocument.Uri)
 }
 
-func (s *Server) handleRename(ctx context.Context, ls *ls.LanguageService, params *lsproto.RenameParams) (lsproto.RenameResponse, error) {
-	return ls.ProvideRename(ctx, params)
+func (s *Server) handleRename(ctx context.Context, languageServer *ls.LanguageService, params *lsproto.RenameParams) (lsproto.RenameResponse, error) {
```
jakebailey: this seems to be a random rename

internal/ls/lsutil/userpreferences.go
```diff
@@ -603,7 +603,7 @@ func (p *UserPreferences) set(name string, value any) {
 	case "usealiasesforrename", "provideprefixandsuffixtextforrename":
 		p.UseAliasesForRename = tsoptions.ParseTristate(value)
 	case "allowrenameofimportpath":
-		p.AllowRenameOfImportPath = parseBoolWithDefault(value, true)
+		p.AllowRenameOfImportPath = tsoptions.ParseTristate(value)
```
jakebailey, line 681: I assume dropping the default here is fine because we handle `unknown` in the tristate elsewhere similarly to `undefined`?

internal/ls/autoimports.go
```diff
@@ -1415,27 +1420,78 @@ func forEachExternalModuleToImportFrom(
 	// }
 }
 
+func getIsExcludedPatterns(preferences *lsutil.UserPreferences, useCaseSensitiveFileNames bool) []*regexp2.Regexp {
+	if preferences.AutoImportFileExcludePatterns == nil {
+		return nil
+	}
+	var patterns []*regexp2.Regexp
```
jakebailey, line 1421: I so wish this feature had not been Regexp based

internal/ls/findallreferences.go
```diff
@@ -466,11 +467,11 @@ func (l *LanguageService) getImplementationReferenceEntries(ctx context.Context,
 	return core.FlatMap(symbolsAndEntries, func(s *SymbolAndEntries) []*ReferenceEntry { return s.references })
 }
 
-func (l *LanguageService) ProvideRename(ctx context.Context, params *lsproto.RenameParams) (lsproto.WorkspaceEditOrNull, error) {
+func (l *LanguageService) ProvideRename(ctx context.Context, params *lsproto.RenameParams, prefs *lsutil.UserPreferences) (lsproto.WorkspaceEditOrNull, error) {
 	program, sourceFile := l.getProgramAndFile(params.TextDocument.Uri)
 	position := int(l.converters.LineAndCharacterToPosition(sourceFile, params.Position))
 	node := astnav.GetTouchingPropertyName(sourceFile, position)
-	if node.Kind != ast.KindIdentifier {
+	if node.Kind != ast.KindIdentifier && !(ast.IsStringLiteralLike(node) && tryGetImportFromModuleSpecifier(node) != nil && prefs.AllowRenameOfImportPath == core.TSTrue) {
```
jakebailey: AllowRenameOfImportPath handling seems off, maybe? I'm noticing that the old code only checks it once. Maybe the structure really is that different

johnfav03:
> Is the test generator just missing these options such that all tests are new handwritten ones?

Isabel mentioned that I'd need to handwrite tests using `BaselineAutoImportsCompletions` as `codefixProvider` hasn't been ported yet

andrewbranch:
Codefixes were ported in #2053

internal/core/core.go
```diff
@@ -688,9 +688,9 @@ func DeduplicateSorted[T any](slice []T, isEqual func(a, b T) bool) []T {
 // CompareBooleans treats true as greater than false.
 func CompareBooleans(a, b bool) int {
 	if a && !b {
-		return -1
-	} else if !a && b {
 		return 1
+	} else if !a && b {
+		return -1
 	}
 	return 0
```
jakebailey, line 695: I noticed this a long time ago and never bothered to fix it 😄

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -3,19 +3,23 @@ TestAmbientShorthandGotoDefinition
 TestArgumentsAreAvailableAfterEditsAtEndOfFunction
 TestAugmentedTypesClass1
 TestAugmentedTypesClass3Fourslash
+TestAutoImportAllowImportingTsExtensionsPackageJsonImports1
```
jakebailey, line 6: Are all tests here being added new tests that fail?

internal/ls/autoimports.go
```diff
@@ -1409,27 +1414,78 @@ func forEachExternalModuleToImportFrom(
...
+	for _, spec := range preferences.AutoImportFileExcludePatterns {
+		// The client is expected to send rooted path specs since we don't know
+		// what directory a relative path is relative to.
```
jakebailey: We get these prefs raw now, so I'm not sure these are actually rooted from the client. We might need to pull on some VS Code code to root the paths like VS Code used to when sending prefs (unless that's in this PR and I just haven't gotten there yet).

internal/ls/autoimports.go
```diff
@@ -1409,27 +1414,78 @@ func forEachExternalModuleToImportFrom(
...
+		pattern := vfs.GetSubPatternFromSpec(spec, "", "exclude", vfs.WildcardMatcher{})
```
jakebailey: Can you export and use `vfs.UsageExclude` instead of copying this constant?

internal/ls/autoimports.go
```diff
@@ -1409,27 +1414,78 @@ func forEachExternalModuleToImportFrom(
...
+	// !!! SymlinkCache
+	// const realpathsWithSymlinks = host.getSymlinkCache?.().getSymlinkedDirectoriesByRealpath();
+	return func(sourceFile *ast.SourceFile) bool {
+		fileName := sourceFile.FileName()
+		for _, p := range excludePatterns {
+			if matched, _ := p.MatchString(fileName); matched {
+				return true
+			}
+		}
+		// !! SymlinkCache
```
jakebailey, line 1468: The symlink cache should now be available.

internal/modulespecifiers/util.go
```diff
@@ -36,18 +36,55 @@ func PathIsBareSpecifier(path string) bool {
...
+		match, _ := re.MatchString(moduleSpecifier)
 		if match {
 			return true
 		}
 	}
 	return false
 }
 
+func stringToRegex(pattern string) *regexp2.Regexp {
+	options := regexp2.None
```
jakebailey: This should also have `| regexp2.ECMAScript`.

internal/modulespecifiers/util.go
```diff
@@ -36,18 +36,55 @@ func PathIsBareSpecifier(path string) bool {
 
 func isExcludedByRegex(moduleSpecifier string, excludes []string) bool {
 	for _, pattern := range excludes {
-		compiled, err := regexp2.Compile(pattern, regexp2.None)
-		if err != nil {
+		re := stringToRegex(pattern)
```
jakebailey, line 50: This will recompile it every time; this isn't new in your code but we should probably have a cache for this.

internal/fourslash/_scripts/failingTests.txt
johnfav03: Yes, they're all tests that can be generated now with importFixModuleSpecifiers; this one specifically fails because it uses external as a target for codeFix, but the diagnostic generated is `external is deprecated` instead of `cannot find name external`

internal/modulespecifiers/util.go
```diff
@@ -36,18 +47,57 @@ func PathIsBareSpecifier(path string) bool {
...
+		match, _ := re.MatchString(moduleSpecifier)
 		if match {
 			return true
 		}
 	}
 	return false
 }
 
+func stringToRegex(pattern string) *regexp2.Regexp {
+	options := regexp2.None
```
jakebailey:
```suggestion
	options := regexp2.ECMAScript
```

This is what I meant; every regex we get over the wire is an ECMAScript one.

internal/modulespecifiers/util.go
```diff
@@ -36,18 +47,57 @@ func PathIsBareSpecifier(path string) bool {
...
+				pattern = pattern[1:lastSlash]
+
+				for _, flag := range flags {
+					switch flag {
+					case 'i':
+						options |= regexp2.IgnoreCase
+					case 'u':
+						options |= regexp2.Unicode
+					case 's':
+						options |= regexp2.ECMAScript
```
jakebailey:
```suggestion
```

internal/modulespecifiers/util.go
johnfav03: Ah, I see! Thanks for the pointer

jakebailey: Er, actually, `regexp2.RegexOptions(regexp2.ECMAScript)`, because this library miswrote their consts

internal/ls/autoimports.go
andrewbranch: Don't bother changing anything in this file more than you need to

jakebailey: alright, ignore me, I guess this is going away soon 😄

johnfav03: Sounds good, should be ready for final review then

PR #1794, opened 2025-10-02
refactor userpreferences

iisaduan:
I'd like to get this finalized before finalizing the parsing/config request handling. Preceeds #1729 

Ports the entire `UserPreferences` struct and marks the preferences that aren't fully implemented with `!!!`. I've turned `*bool` into `core.Tristate`, which is what `CompilerOptions` does, and simplified many preferences where it had type `undefined | bool` in strada and the behavior of `undefined` was the same as `false` (aka the default behavior was as-if it was set false). These changes simplify parsing and managing of the lifecycle of preferences structs in snapshots

Also renames the file from `ls/types.go`->`ls/userpreferences.go`


iisaduan:
I guess its nice that I can now ramble in my PR descriptions, and copilot will write a decent summary 😅

PR #1795, opened 2025-10-02
Port TS PR 61668 - Fix type variable leaks and cache inconsistencies

jakebailey:
Ports https://github.com/microsoft/TypeScript/pull/61668.

I am not 100% sure this is correct. Might need @ahejlsberg to attempt it.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -16431,7 +16431,7 @@ func (c *Checker) hasNonCircularBaseConstraint(t *Type) bool {
 
 // This is a worker function. Use getConstraintOfTypeParameter which guards against circular constraints
 func (c *Checker) getConstraintFromTypeParameter(t *Type) *Type {
-	if t.Flags()&TypeFlagsTypeParameter == 0 {
+	if t.flags&TypeFlagsTypeParameter == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


internal/checker/checker.go
```diff
@@ -16316,6 +16321,9 @@ func (c *Checker) getConstraintOfType(t *Type) *Type {
 }
 
 func (c *Checker) getConstraintOfTypeParameter(typeParameter *Type) *Type {
+	if typeParameter.flags&TypeFlagsTypeParameter == 0 {
+		return nil
+	}
```
jakebailey: Excitingly, the code actually passes in intrinsic types through this function because `inferReverseMappedTypeWorker` in Strada incorrectly casted the result of `getIndexedAccessType` to `TypeParameter` as the `typeParameter` of an inference context.

internal/checker/checker.go
```diff
@@ -16328,6 +16333,10 @@ func (c *Checker) hasNonCircularBaseConstraint(t *Type) bool {
 
 // This is a worker function. Use getConstraintOfTypeParameter which guards against circular constraints
 func (c *Checker) getConstraintFromTypeParameter(t *Type) *Type {
+	if t.Flags()&TypeFlagsTypeParameter == 0 {
```
ahejlsberg: Just use `t.flags` as we do elsewhere.

internal/checker/checker.go
```diff
@@ -24161,7 +24160,7 @@ func (c *Checker) newObjectType(objectFlags ObjectFlags, symbol *ast.Symbol) *Ty
 	case objectFlags&ObjectFlagsInstantiationExpressionType != 0:
 		data = &InstantiationExpressionType{}
 	case objectFlags&ObjectFlagsSingleSignatureType != 0:
```
ahejlsberg: I think you can delete this `case`. Single-signature types are always anonymous types handled by the next case.

PR #1797, opened 2025-10-03
Port PR 61805 - module=node20

jakebailey:
Ports https://github.com/microsoft/TypeScript/pull/61805


jakebailey:
Hm, data race...

internal/checker/checker.go
```diff
@@ -13919,14 +13925,42 @@ func (c *Checker) getTargetOfImportClause(node *ast.Node, dontResolveAlias bool)
...
 	var exportDefaultSymbol *ast.Symbol
+	var exportModuleDotExportsSymbol *ast.Symbol
 	if isShorthandAmbientModuleSymbol(moduleSymbol) {
 		exportDefaultSymbol = moduleSymbol
-	} else {
+	} else if file != nil && specifier != nil &&
+		core.ModuleKindNode20 <= c.moduleKind && c.moduleKind <= core.ModuleKindNodeNext &&
+		c.getEmitSyntaxForModuleSpecifierExpression(specifier) == core.ModuleKindCommonJS &&
+		c.program.GetImpliedNodeFormatForEmit(file.AsSourceFile()) == core.ModuleKindESNext {
+		exportModuleDotExportsSymbol = c.resolveExportByName(moduleSymbol, "module.exports", node, dontResolveAlias)
```
jakebailey: ast.InternalSymbolNameModuleExports

internal/checker/checker.go
```diff
@@ -14920,29 +14958,47 @@ func (c *Checker) resolveESModuleSymbol(moduleSymbol *ast.Symbol, referencingLoc
...
-			// }
+
+			targetFile := core.Find(moduleSymbol.Declarations, ast.IsSourceFile)
+			usageMode := c.getEmitSyntaxForModuleSpecifierExpression(reference)
+			var exportModuleDotExportsSymbol *ast.Symbol
+			if namespaceImport != nil && targetFile != nil &&
+				core.ModuleKindNode20 <= c.moduleKind && c.moduleKind <= core.ModuleKindNodeNext &&
+				usageMode == core.ModuleKindCommonJS &&
+				c.program.GetImpliedNodeFormatForEmit(targetFile.AsSourceFile()) == core.ModuleKindESNext {
+				exportModuleDotExportsSymbol = c.getExportOfModule(symbol, "module.exports", namespaceImport, dontResolveAlias)
```
jakebailey: ast.InternalSymbolNameModuleExports

jakebailey:
I can't figure out how to deal with the circularity errors when compiling concurrently. These are the kinds of errors that we have been silencing by skipping the tests (`concurrentSkippedErrorBaselines`), which would defeat the purpose of these checks. I am not sure if this means the tests are "bad" themselves for introducing those circularities, or if there's a bug in my (or someone else's) ported code that's causing the issue.

jakebailey:
The symbol baselines doing something like:

image

Implies that we are somehow choosing the wrong import when compiling concurrently, so I hope it's just that kind of bug, but the race detector does not detect anything.

internal/checker/checker.go
```diff
@@ -14960,29 +14997,51 @@ func (c *Checker) resolveESModuleSymbol(moduleSymbol *ast.Symbol, referencingLoc
...
+				if c.hasSignatures(typ) || c.getPropertyOfTypeEx(typ, ast.InternalSymbolNameDefault, true /*skipObjectFunctionPropertyAugment*/, false /*includeTypeOnlyMembers*/) != nil || isEsmCjsRef {
+					var moduleType *Type
+					if typ.Flags()&TypeFlagsStructuredType != 0 {
+						moduleType = c.getTypeWithSyntheticDefaultImportType(typ, symbol, moduleSymbol, reference)
+					} else {
+						moduleType = c.createDefaultPropertyWrapperForModule(symbol, symbol.Parent, nil)
+					}
+					return c.cloneTypeAsModuleType(symbol, moduleType, referenceParent)
+				}
+			}
```
jakebailey, line 15032: This bit is the code that's causing the circularity. Perhaps that's why it was commented out originally?

andrewbranch:
The circularities are real. I believe what’s going on here is this: suppose you have a big circle of imports: a.ts → b.ts → c.ts → d.ts → e.ts → a.ts. In single-threaded mode, we always start with a.ts because of file order / source order, and we go off down the chain:


- resolve import of b.ts
  - resolve import of c.ts
    - resolve import of d.ts
      - resolve import of e.ts
        - resolve import of a.ts - already in progress! set import of a.ts to any
        - return any
      - return any
    - return any
  - return any
- return any
- import of b.ts got set to any while resolving import of c.ts, so a circularity occurred
- report an error on the import of b.ts

With this algorithm, the circle can be arbitrarily long, but we only ever report an error on the place where we start, which is semi-arbitrary but deterministic by file order, source order, etc. But if we divvy up all the files between multiple independent checkers, there’s a strong chance multiple of them will start resolving the chain in different places, resulting in the same error being reported in a different file.

I’m not sure what we want to actually do about this.

- It would be easy to issue the circularity error on every file instead of an arbitrary point, but that would probably be annoying
- Otherwise, we would have to save the resolution stack in order to apply some arbitrary but deterministic way of picking where to issue the error regardless of where we started. (We don’t currently maintain that stack at all; we just notice a mutation happened and know a circularity occurred, but we have no breadcrumbs for it. It shouldn’t be hard to add, though.)

jakebailey:
That's what I thought, unfortunately.

As a non-fix, should the tests be fixed to not use a circular reference? It seems like those which break aren't intentionally testing that behavior.

andrewbranch:
Yeah, I think that would be fine for now.

jakebailey:
https://github.com/microsoft/TypeScript/pull/62568 does that, and it indeed makes this PR pass tests when applied to the submodule.

testdata/baselines/reference/submodule/compiler/esModuleInteropPrettyErrorRelatedInformation.errors.txt.diff
```diff
@@ -1,31 +1,14 @@
...
--[7m [0m [91m       ~~~[0m
--
+@@= skipped -3, +3 lines =@@
+ [7m3[0m invoke(foo);
+ [7m [0m [91m       ~~~[0m
+
 -  [96mindex.ts[0m:[93m1[0m:[93m1[0m
--    [7m1[0m import * as foo from "./foo";
--    [7m [0m [96m~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[0m
++  [96mindex.ts[0m:[93m1[0m:[93m1[0m - Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```
jakebailey, line 8: Curious difference here; I am not sure if this is something I did? Must have?

jakebailey: The code that reports this looks a bit different than Strada, so probably not worth blocking this PR on.

PR #1801, opened 2025-10-03
Ensure `Parameters` property of arrow function is never nil

ahejlsberg:
Fixes a crash in the language service:

```ts
const f = () => 42  // Delete () and LSP panics
```


jakebailey (approved):
Hopefully all of the other signature like things all have this set too (it was required in the old code)

ahejlsberg:
> Hopefully all of the other signature like things all have this set too (it was required in the old code)

I check all the code paths for parsing signatures, and this should be the only one that could cause a `nil`.

PR #1803, opened 2025-10-03
Port "Provide string completions for `in` keyword checks"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/60272

I still can't enable the related tests because in Corsa there is a pre-existing issue causing a label mismatch on the expected completion entry


jakebailey (approved):
Would be nice to have a test but it looks trivially correct to me

PR #1805, opened 2025-10-03
Port "Account for right operands & fix a weird error message for leftmost nullish literals in checkNullishCoalesceOperands"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/59569


internal/checker/checker.go
```diff
@@ -12406,17 +12399,32 @@ func (c *Checker) checkNullishCoalesceOperands(left *ast.Node, right *ast.Node)
...
-				c.error(leftTarget, diagnostics.Right_operand_of_is_unreachable_because_the_left_operand_is_never_nullish)
-			}
+			c.error(leftTarget, diagnostics.Right_operand_of_is_unreachable_because_the_left_operand_is_never_nullish)
+		}
+	}
+}
+
+func (c *Checker) checkNullishCoalesceOperandRight(right *ast.Node) {
+	binaryExpression := right.Parent
+	if binaryExpression.Parent != nil && ast.IsBinaryExpression(binaryExpression.Parent) && binaryExpression.Parent.AsBinaryExpression().OperatorToken.Kind == ast.KindQuestionQuestionToken {
```
jakebailey, line 12420: Is retaining `isNotWithinNullishCoalesceExpression` too annoying?

Andarist: it's not like super annoying, but it's a rather straightforward check and it was only used in this single place so I decided to inline it

PR #1807, opened 2025-10-03
(not intending to merge) Add option for overriding GOMEMLIMIT

maschwenk:
Just a conversation starter. On a very large monorepo by setting this pretty low we can get very good performance with 25% less memory usage.

### DIFF
--- a/_extension/src/extension.ts
+++ b/_extension/src/extension.ts
@@ -39,6 +39,9 @@ export function activate(context: vscode.ExtensionContext) {
         },
     };
 
+    const config = vscode.workspace.getConfiguration("typescript-go");
+    const memoryLimit = config.get<string>("memoryLimit");
+
     const clientOptions: LanguageClientOptions = {
         documentSelector: [
             { scheme: "file", language: "typescript" },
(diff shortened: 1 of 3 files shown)


_extension/src/extension.ts
```diff
@@ -1,75 +1,143 @@
...
+            { scheme: "file", language: "javascript" },
+            { scheme: "file", language: "javascriptreact" },
+            { scheme: "untitled", language: "typescript" },
+            { scheme: "untitled", language: "typescriptreact" },
+            { scheme: "untitled", language: "javascript" },
+            { scheme: "untitled", language: "javascriptreact" },
+        ],
+        outputChannel: output,
+        traceOutputChannel: traceOutput,
+        ...(memoryLimit ? { initializationOptions: { memoryLimit: memoryLimit } } : {}),
```
jakebailey, line 58: This doesn't need to be an init option for the LS, you can just set `GOMEMLIMIT` as an env var for the process

maschwenk: @jakebailey do you mean for VSCode itself? because that's how I was testing this, like

```
GOMEMLIMIT=2048MiB code .
```

but then obviously that sets it for anything you run within the integrated VSCode terminal etc

maschwenk: See

- https://github.com/microsoft/typescript-go/pull/1808

PR #1808, opened 2025-10-03
Add option for overriding GOMEMLIMIT (and maybe other options? let's discuss)

maschwenk:
- https://github.com/microsoft/typescript-go/pull/1807

But less silly

### DIFF
--- a/_extension/package.json
+++ b/_extension/package.json
@@ -51,6 +51,11 @@
                         "type": "string",
                         "description": "Path to the @typescript/native-preview package or tsgo binary directory. If not specified, the extension will look for it in the default location.",
                         "tags": ["experimental"]
+                    },
+                    "typescript.native-preview.goMemLimit": {
+                        "type": "string",
+                        "description": "Set GOMEMLIMIT for the language server (e.g., '2048MiB', '4GiB'). See https://pkg.go.dev/runtime#hdr-Environment_Variables for more information.",
+                        "tags": ["experimental"]
                     }
                 }
             }
--- a/_extension/src/client.ts
+++ b/_extension/src/client.ts
@@ -95,16 +95,25 @@ export class Client {
         const pprofDir = config.get<string>("pprofDir");
         const pprofArgs = pprofDir ? ["--pprofDir", pprofDir] : [];
 
+        // Get goMemLimit
+        const goMemLimit = config.get<string>("goMemLimit");
+        const env = goMemLimit ? { ...process.env, GOMEMLIMIT: goMemLimit } : process.env;
+        if (goMemLimit) {
+            this.outputChannel.appendLine(`Setting GOMEMLIMIT=${goMemLimit}`);
+        }
+
         const serverOptions: ServerOptions = {
             run: {
                 command: this.exe.path,
                 args: ["--lsp", ...pprofArgs],
                 transport: TransportKind.stdio,
+                options: { env },
             },
             debug: {
                 command: this.exe.path,
                 args: ["--lsp", ...pprofArgs],
                 transport: TransportKind.stdio,
+                options: { env },
             },
         };
 


_extension/package.json
```diff
@@ -51,6 +51,11 @@
                         "type": "string",
                         "description": "Path to the @typescript/native-preview package or tsgo binary directory. If not specified, the extension will look for it in the default location.",
                         "tags": ["experimental"]
+                    },
+                    "typescript.native-preview.goMemLimit": {
```
maschwenk, line 61: Y'all have to manage backwards compatible names a lot more than I do, so maybe you have better ideas on how to make this more generically useful going forward...i.e. some way to set any of the various Go flags. I also totally understand if you do not want to invite this added complexity right now!

jakebailey:
@maschwenk I think I want to try to get this in, just so we can have people try this and see how it behaves.

The current code looks fine, though it requires a merge from main.

maschwenk:
@jakebailey on it!

_extension/src/client.ts
```diff
@@ -103,16 +103,25 @@ export class Client {
         const pprofDir = config.get<string>("pprofDir");
         const pprofArgs = pprofDir ? ["--pprofDir", pprofDir] : [];
 
+        // Get goMemLimit
+        const goMemLimit = config.get<string>("goMemLimit");
+        const env = goMemLimit ? { ...process.env, GOMEMLIMIT: goMemLimit } : process.env;
```
maschwenk: Idk if we want to throw if someone provides something that is not of type

```ts
`${number}${'KiB' | 'MiB' | 'GiB'}`;
```

or if Go will barf loud enough to make it obvious

jakebailey: Go will do:

```console
$ GOMEMLIMIT=lol go version
GOMEMLIMIT=lol
fatal error: malformed GOMEMLIMIT; see `go doc runtime/debug.SetMemoryLimit`

runtime stack:
runtime.throw({0xbe5093?, 0xa?})
        runtime/panic.go:1094 +0x48 fp=0x7ffc5cac39f0 sp=0x7ffc5cac39c0 pc=0x47d488
runtime.readGOMEMLIMIT()
        runtime/mgcpacer.go:1365 +0xaf fp=0x7ffc5cac3a20 sp=0x7ffc5cac39f0 pc=0x42edaf
runtime.gcinit()
        runtime/mgc.go:189 +0x2e fp=0x7ffc5cac3a50 sp=0x7ffc5cac3a20 pc=0x4258ee
runtime.schedinit()
        runtime/proc.go:898 +0x21e fp=0x7ffc5cac3af0 sp=0x7ffc5cac3a50 pc=0x448abe
runtime.rt0_go()
        runtime/asm_amd64.s:349 +0x11c fp=0x7ffc5cac3af8 sp=0x7ffc5cac3af0 pc=0x48351c
```

jakebailey: I can imagine it being worth validating, but that sounds like complicated code

maschwenk: 🫡  Do nothing instead of something 🫡 

I am always on board

jakebailey: I would just validate it against the regex documented in Go: `^[0-9]+(([KMGT]i)?B)?$`

https://github.com/golang/go/blob/63fced531c0d702554fcfda41f9016701a6859c6/src/runtime/string.go#L406C4-L406C26

Then, log an error with `.error` and don't set pass it down.

jakebailey: Oh, actually, you can also add that regex to the package.json too. https://code.visualstudio.com/api/references/contribution-points#contributes.configuration:~:text=pattern%20for%20restricting%20strings%20to%20a%20given%20regular%20expression

maschwenk: Sounds good, on it

_extension/package.json
```diff
@@ -57,6 +57,15 @@
                         "tags": [
                             "experimental"
                         ]
+                    },
+                    "typescript.native-preview.goMemLimit": {
+                        "type": "string",
+                        "description": "Set GOMEMLIMIT for the language server (e.g., '2048MiB', '4GiB'). See https://pkg.go.dev/runtime#hdr-Environment_Variables for more information.",
+                        "pattern": "^[0-9]+[KMGT]iB$",
```
maschwenk: God help us if the `T` ever gets used 😆

jakebailey: This isn't the same regex as what I pasted; this won't support just B for bytes. You should be able to copy it exactly.

_extension/src/client.ts
jakebailey: You'll also want to revalidate the value, or at least hand test to verify that `.get` won't grab a bad value

_extension/package.json
maschwenk: Grr sorry a bad autocomplete

fixed regex in package.json
[5d58a90](https://github.com/microsoft/typescript-go/pull/1808/commits/5d58a9071fa37e7a758bf617d9deac418fb19ea1)

Retesting the pattern in the code as well, assuming that's what you meant by "revalidate the value"
[697ec86](https://github.com/microsoft/typescript-go/pull/1808/commits/697ec8633bc70498761433bb08120d2d64b3c05e)

jakebailey: I didn't mean a throw, I meant to output to the log with `.error()` and just pretend that the memlimit was never set.

_extension/src/client.ts
```diff
@@ -103,16 +103,29 @@ export class Client {
         const pprofDir = config.get<string>("pprofDir");
         const pprofArgs = pprofDir ? ["--pprofDir", pprofDir] : [];
 
+        const goMemLimit = config.get<string>("goMemLimit");
+        if (goMemLimit) {
+            // Keep this regex aligned with the pattern in package.json.
+            if (!/^[0-9]+(([KMGT]i)?B)?$/.test(goMemLimit)) {
+                this.outputChannel.error(`Invalid goMemLimit: ${goMemLimit}. Must be a valid memory limit (e.g., '2048MiB', '4GiB'). Not overriding GOMEMLIMIT.`);
```
jakebailey: This doesn't unset goMemLimit

_extension/package.json
maschwenk: e0f1bb756c9cea364b88fd2f14ac25f26cc86707

Not seeing other examples of an error log but this looks correct-ish

_extension/src/client.ts
maschwenk: d62cbcddda9c4f95941a5f66e964685aff145479

pulled out into a little helper to avoid having to do a bunch of weird nested conditional stuff and then don't need to "unset" anything, just set it once it's validated.

also figure at some point there might other env mutations that need to go on for other Go env vars, and might be easier to reason about if all that env configuration is done in one method

_extension/src/client.ts
```diff
@@ -103,16 +116,21 @@ export class Client {
         const pprofDir = config.get<string>("pprofDir");
         const pprofArgs = pprofDir ? ["--pprofDir", pprofDir] : [];
 
+        const goMemLimit = config.get<string>("goMemLimit");
```
jakebailey, line 106: I think this is a bit too abstracted now.

I would just spread process.env to copy it, then check the config, and if is valid, set GOMEMLIMIT on the copy.

maschwenk:
> This seems to work, though it does not integrate with the restart command because it's resolved before the client is created and then `.restart()` provide no opportunity to update it.
> 
> It may be that we should instead move this code into a user preference in the server itself and then configure it there on the fly.
> 
> Unfortunately, in my own testing, I felt like the option probably hurt more than it helped. If I set it to something like `2GiB`, then opening the main TS codebase would use 1.4G, when without the option it would use 600MB. Not sure what to make of that.

@jakebailey sort of similar to my experience, hence why I left this open for a while. Still, maybe over time this will be a useful dial to turn

PR #1809, opened 2025-10-03
Port "Deemphasize old JSX transform"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/61586


jakebailey:
Not sure why the CLA check is mad but hopefully reopening fixes it.

PR #1810, opened 2025-10-03
Port "Fix #61098"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/61113


internal/checker/relater.go
```diff
@@ -3298,6 +3298,22 @@ func (r *Relater) structuredTypeRelatedToWorker(source *Type, target *Type, repo
 					return result
 				}
 			}
+		case source.flags&TypeFlagsTemplateLiteral != 0:
+			if slices.Equal(source.AsTemplateLiteralType().texts, target.AsTemplateLiteralType().texts) {
```
jakebailey, line 3302: No changes needed, but `slices.Equal` doesn't do a quick check for slice equality (which in go would be a little more complicated) whereas `arrayIsEqualTo` did. Wonder how much perf we are losing due to the (handful) of calls that were swapped.

jakebailey (approved):

PR #1820, opened 2025-10-04
Port "Fixed crash in `hasVisibleDeclarations` related to binding elements"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/61352

It seems this isn't strictly needed right now. Likely, reparsing fixed the original issue already but I think it might still be a good idea to pick up this change for the sake of code parity between Strada and Corsa

### DIFF
--- a/internal/checker/emitresolver.go
+++ b/internal/checker/emitresolver.go
@@ -409,7 +409,14 @@ func (r *emitResolver) hasVisibleDeclarations(symbol *ast.Symbol, shouldComputeA
 					continue
 				}
 				if symbol.Flags&ast.SymbolFlagsBlockScopedVariable != 0 {
-					variableStatement := ast.FindAncestor(declaration, ast.IsVariableStatement)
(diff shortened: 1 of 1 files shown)


internal/checker/emitresolver.go
```diff
@@ -409,7 +409,14 @@ func (r *emitResolver) hasVisibleDeclarations(symbol *ast.Symbol, shouldComputeA
 					continue
 				}
 				if symbol.Flags&ast.SymbolFlagsBlockScopedVariable != 0 {
-					variableStatement := ast.FindAncestor(declaration, ast.IsVariableStatement)
+					rootDeclaration := ast.WalkUpBindingElementsAndPatterns(declaration)
+					if ast.IsParameter(rootDeclaration) {
+						return nil
+					}
+					variableStatement := rootDeclaration.Parent.Parent;
```
Andarist:
```suggestion
					variableStatement := rootDeclaration.Parent.Parent
```

jakebailey (approved):

PR #1823, opened 2025-10-04
Speed up levenshteinWithMax by reusing buffers

jakebailey:
Closes #1640

This turns out to be a top allocator in tests too.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -477,6 +477,8 @@ func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) s
 	maximumLengthDifference := max(2, int(float64(len(name))*0.34))
 	bestDistance := math.Floor(float64(len(name))*0.4) + 1 // If the best result is worse than this, don't bother.
 	runeName := []rune(name)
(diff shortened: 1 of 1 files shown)


internal/core/core.go
```diff
@@ -502,9 +503,28 @@ func GetSpellingSuggestion[T any](name string, candidates []T, getName func(T) s
...
+var levenshteinBuffersPool = sync.Pool{
+	New: func() any {
+		return &levenshteinBuffers{}
+	},
+}
+
 func levenshteinWithMax(s1 []rune, s2 []rune, maxValue float64) float64 {
-	previous := make([]float64, len(s2)+1)
-	current := make([]float64, len(s2)+1)
+	buffers := levenshteinBuffersPool.Get().(*levenshteinBuffers)
```
dzbarsky: Sync.Pool traffic is cheap but not free. I understand you want to keep the interface simple but perhaps the Get call can be lifted to the caller (since it's still package-private) to get back a bit of efficiency without polluting the public interface?

jakebailey: I suppose so.

PR #1824, opened 2025-10-05
Port "Fixed nullish coalesce operator's precedence"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/61372


internal/ast/precedence.go
```diff
@@ -181,6 +172,15 @@ const (
...
 	OperatorPrecedenceDisallowComma = OperatorPrecedenceYield
+	// ShortCircuitExpression:
+	//     LogicalORExpression
+	//     CoalesceExpression
+	// CoalesceExpression:
+	//     CoalesceExpressionHead `??` BitwiseORExpression
+	// CoalesceExpressionHead:
+	//     CoalesceExpression
+	//     BitwiseORExpression
+	OperatorPrecedenceCoalesce = OperatorPrecedenceLogicalOR
```
Andarist, line 183: Alternatively, I could make the `OperatorPrecedenceCoalesce` lower than `OperatorPrecedenceLogicalOR` and then handle that change from Strada in `parseBinaryExpressionRest` and its computation of `consumeCurrentOperator` (and maybe in some other places too). cc @weswigham

internal/printer/printer.go
```diff
@@ -2716,14 +2716,6 @@ func (p *Printer) getBinaryExpressionPrecedence(node *ast.BinaryExpression) (lef
 	case ast.OperatorPrecedenceAssignment:
 		// assignment is right-associative
 		leftPrec = ast.OperatorPrecedenceLeftHandSide
-	case ast.OperatorPrecedenceCoalesce:
```
Andarist, line 2719: I had to remove it because it would now be a duplicate case - but it seems that with the current logic I don't even have to move this code anywhere as nothing has regressed

internal/ast/precedence.go
jakebailey: Wouldn't that match Strada better? Or is there a downside?

Andarist: Strada defines those 2 with the same value so the current approach in this PR matches Strada rather closely. The alternative I mentioned would require changes that would diverge from Strada (code-wise)

jakebailey: Agh, my bad, read the other PR wrong and forgot to expand some code.

weswigham (approved):
Eh, this is probably the better change. It already matches the strada change because logicalOR's precedence is already in the newer position in corsa, so doesn't need to move like it did in the strada change, and it's probably better to need to touch less places anyway - precedence adjustments affect AST shape and API consumers, so it's best to be minimal when fixing them (re: the original strada change).

PR #1826, opened 2025-10-05
Port "Add support for `import defer` proposal" 

Andarist:
ports https://github.com/microsoft/TypeScript/pull/60757
ports https://github.com/microsoft/TypeScript/pull/61837


internal/ast/kind.go
```diff
@@ -422,7 +423,7 @@ const (
 	KindFirstJSDocTagNode       = KindJSDocTag
 	KindLastJSDocTagNode        = KindJSDocImportTag
 	KindFirstContextualKeyword  = KindAbstractKeyword
-	KindLastContextualKeyword   = KindOfKeyword
+	KindLastContextualKeyword   = KindDeferKeyword
```
jakebailey, line 426:
```suggestion
	KindLastContextualKeyword   = KindLastKeyword
```

internal/ast/utilities.go
```diff
@@ -2840,11 +2847,11 @@ func GetLeftmostAccessExpression(expr *Node) *Node {
 func IsTypeOnlyImportDeclaration(node *Node) bool {
 	switch node.Kind {
 	case KindImportSpecifier:
-		return node.AsImportSpecifier().IsTypeOnly || node.Parent.Parent.AsImportClause().IsTypeOnly
+		return node.AsImportSpecifier().IsTypeOnly || node.Parent.Parent.AsImportClause().PhaseModifier == KindTypeKeyword
```
jakebailey: Can you just do `IsTypeOnly()` for these cases?

internal/ls/autoimportfixes.go
```diff
@@ -89,7 +89,7 @@ func (ct *changeTracker) doAddExistingFix(
 					identifier = ct.NodeFactory.NewIdentifier(namedImport.propertyName).AsIdentifier().AsNode()
 				}
 				return ct.NodeFactory.NewImportSpecifier(
-					(!importClause.IsTypeOnly || promoteFromTypeOnly) && shouldUseTypeOnly(namedImport.addAsTypeOnly, preferences),
+					(importClause.PhaseModifier != ast.KindTypeKeyword || promoteFromTypeOnly) && shouldUseTypeOnly(namedImport.addAsTypeOnly, preferences),
```
jakebailey: Can you just do `IsTypeOnly()`?

internal/parser/parser.go
```diff
@@ -2100,22 +2100,38 @@ func (p *Parser) parseImportDeclarationOrImportEqualsDeclaration(pos int, hasJSD
...
 		identifier = nil
 		if p.isIdentifier() {
 			identifier = p.parseIdentifier()
 		}
+	} else if identifier != nil && identifier.AsIdentifier().Text == "defer" {
+		var shouldParseAsDeferModifier bool
+		if p.token == ast.KindFromKeyword {
+			shouldParseAsDeferModifier = !p.lookAhead(func(p *Parser) bool {
+				return p.nextTokenIsTokenStringLiteral()
+			})
```
```suggestion
			shouldParseAsDeferModifier = !p.lookAhead((*Parser).nextTokenIsTokenStringLiteral)
```

_packages/api/src/node.ts
```diff
@@ -544,7 +544,6 @@ export class RemoteNode extends RemoteNodeBase implements Node {
     get isTypeOnly(): boolean | undefined {
         switch (this.kind) {
             case SyntaxKind.ImportSpecifier:
-            case SyntaxKind.ImportClause:
```
jakebailey, line 547: Should this defer to `phaseModifier` in some way?

Andarist: I have added that at first but then concluded this should closely match up the defined AST shape and `ImportClause` doesn't have `IsTypeOnly` property anymore (in Strada it's deprecated). It could be handled here for convenience though. I'll defer (hehe) the decision to you and @andrewbranch

jakebailey: That's true, though that should come with an update to the `ast` package as well.

Andarist:
@jakebailey I pushed out a change but I'm not sure if this has to match Strada 100% or if this package can make *minor* breaking changes. Notably:
- I removed `ImportClause['isTypeOnly']` (Strada has it as deprecated)
- I made `ImportClause['phaseModifier']` **required** and the lack of the phase modifier is represented by `SyntaxKind.Unknown`. In Strada `undefined` would be used instead

jakebailey:
Looks correct to me. It's just a prototype so okay to make changes like this, and it won't be the final version anyway.

testdata/baselines/reference/submodule/conformance/importDeferComments.errors.txt.diff
```diff
@@ -2,29 +2,14 @@
...
-+                                                    ~~~
-+!!! error TS1005: ';' expected.
-+                                                    ~~~
-+!!! error TS2304: Cannot find name 'aNs'.
-+                                                              ~~~~
-+!!! error TS1434: Unexpected keyword or identifier.
-+                                                              ~~~~
-+!!! error TS2304: Cannot find name 'from'.
++                                                                         ~~~
++!!! error TS2307: Cannot find module 'a' or its corresponding type declarations.
```
jakebailey, line 14: This is an interesting diff. Are we not walking this import when collecting module references? That could explain why the checker can't find something to resolve to.

jakebailey: Or, maybe this test is "bad" as it relied on baseUrl behavior?

Andarist: > Or, maybe this test is "bad" as it relied on baseUrl behavior?

It might be the case - was there a default `baseUrl` set in Strada's tests? Either way, I have confirmed before this is unrelated to this PR as the same case without `defer` results in the same error on `main`

jakebailey: Indeed, thanks.

jakebailey (approved):

PR #1827, opened 2025-10-05
Remove JSDoc handling from the binder

Andarist:
Those lines can be removed given the used approach with the reparser


jakebailey (approved):
I think this seems fine.

PR #1828, opened 2025-10-05
Port "`arguments` should not be allowed in class static block"

Andarist:
ports https://github.com/microsoft/TypeScript/pull/48172


testdata/baselines/reference/submodule/compiler/argumentsUsedInClassFieldInitializerOrStaticInitializationBlock.types.diff
Andarist, line 1: this has some seemingly unwanted diffs but I think this is basically desired and I opened a PR to Strada to close this gap: https://github.com/microsoft/TypeScript/pull/62549

jakebailey: Yeah, I think this was just missed. Yay optional parameters

PR #1831, opened 2025-10-06
Fix translation bug causing `unknown` infer-extends constraint to be emitted

Andarist:


internal/checker/nodebuilderimpl.go
```diff
@@ -1440,7 +1440,7 @@ func (b *nodeBuilderImpl) createMappedTypeNodeFromType(t *Type) *ast.TypeNode {
 		originalConstraint := b.ch.instantiateType(rawConstraintTypeFromDeclaration, mapped.mapper)
 
 		var originalConstraintNode *ast.Node
-		if originalConstraint.flags&TypeFlagsUnknown != 0 {
+		if originalConstraint.flags&TypeFlagsUnknown == 0 {
```
Andarist, line 1443: the original is:
```ts
originalConstraint.flags & TypeFlags.Unknown ? undefined : typeToTypeNodeHelper(originalConstraint, context)
```

PR #1833, opened 2025-10-06
Fixed an issue with type params being renamed incorrectly

Andarist:
This PR removes 5 .diffs, most (if not all) of the other ones that were changed here are a result of the lack of return type nodes reuse ([here](https://github.com/microsoft/typescript-go/blob/912f193b712481c4a7be8b93d56d1e6b9f4d0d45/internal/checker/nodebuilderimpl.go#L1931-L1940))


internal/checker/nodebuilderscopes.go
```diff
@@ -24,12 +24,12 @@ func cloneNodeBuilderContext(context *NodeBuilderContext) func() {
...
 	oldMustCreateTypeParameterSymbolList := context.hasCreatedTypeParameterSymbolList
 	oldMustCreateTypeParametersNamesLookups := context.hasCreatedTypeParametersNamesLookups
-	context.hasCreatedTypeParameterSymbolList = false
-	context.hasCreatedTypeParametersNamesLookups = false
 	oldTypeParameterNames := context.typeParameterNames
 	oldTypeParameterNamesByText := context.typeParameterNamesByText
 	oldTypeParameterNamesByTextNextNameCount := context.typeParameterNamesByTextNextNameCount
 	oldTypeParameterSymbolList := context.typeParameterSymbolList
+	context.hasCreatedTypeParameterSymbolList = oldTypeParameterSymbolList != nil
+	context.hasCreatedTypeParametersNamesLookups = oldTypeParameterNames != nil
```
Andarist, line 32: why does it matter? because we don't want to override the saved **clones** of the data below with empty data here:
https://github.com/microsoft/typescript-go/blob/912f193b712481c4a7be8b93d56d1e6b9f4d0d45/internal/checker/nodebuilderimpl.go#L1308-L1313

testdata/baselines/reference/submodule/compiler/declarationEmitShadowingInferNotRenamed.js.diff
```diff
@@ -9,5 +9,5 @@
      foo: number;
  };
 -export declare const createClient: <D extends (new (...args: any[]) => Client) | Record<string, new (...args: any[]) => Client>>(clientDef: D) => D extends new (...args: any[]) => infer C ? UpdatedClient<C> : { [K in keyof D]: D[K] extends new (...args: any[]) => infer C ? UpdatedClient<C> : never; };
-+export declare const createClient: <D extends Record<string, new (...args: any[]) => string> | (new (...args: any[]) => string)>(clientDef: D) => D extends new (...args: any[]) => infer C ? UpdatedClient<C> : { [K in keyof D]: D[K] extends new (...args: any[]) => infer C ? UpdatedClient<C> : never; };
++export declare const createClient: <D extends Record<string, new (...args: any[]) => string> | (new (...args: any[]) => string)>(clientDef: D) => D extends new (...args: any[]) => infer C ? UpdatedClient<C> : { [K in keyof D]: D[K] extends new (...args: any[]) => infer C_1 ? UpdatedClient<C_1> : never; };
```
jakebailey, line 12: This seems like a regression?

testdata/baselines/reference/submodule/compiler/mappedTypeGenericInstantiationPreservesHomomorphism.types.diff
```diff
@@ -16,6 +16,6 @@
 ->usePrivateType(...args) : T[any] extends infer T_1 ? { [K in keyof T_1]: T_1[K]; } : never
 ->usePrivateType : <T_1 extends unknown[]>(...args: T_1) => T_1[any] extends infer T_2 ? { [K in keyof T_2]: T_2[K]; } : never
 +>usePrivateType(...args) : T_1[any] extends infer T ? { [K in keyof T]: T_1[any][K]; } : never
-+>usePrivateType : <T_1 extends unknown[]>(...args: T_1) => T_3[any] extends infer T_2 ? { [K in keyof T_2]: T_3[any][K]; } : never
++>usePrivateType : <T_1 extends unknown[]>(...args: T_1) => T_1[any] extends infer T_2 ? { [K in keyof T_2]: T_1[any][K]; } : never
```
jakebailey, line 19: These are so close to being correct. I wonder what tiny thing is causing this diff in general...

testdata/baselines/reference/submodule/compiler/declarationEmitShadowingInferNotRenamed.js.diff
Andarist: Based on my quick investigation into this, it will get fixed with return type nodes reuse

Andarist: I can confirm that Strada generates the same thing if I only comment out what is not implemented in Corsa yet [here](https://github.com/microsoft/typescript-go/blob/912f193b712481c4a7be8b93d56d1e6b9f4d0d45/internal/checker/nodebuilderimpl.go#L1931-L1940)

testdata/baselines/reference/submodule/compiler/mappedTypeGenericInstantiationPreservesHomomorphism.types.diff
Andarist: I'll look into this in a followup work

weswigham (approved):
So `hasCreatedTypeParameterSymbolList` and `hasCreatedTypeParametersNamesLookups` are inversions of the field in strada (which says `must` instead of `has`). This logically tracks, since in strada when we clone a context we init the field to `true`, (which, since inverted, here, `false`).

Honestly, dunno why I conditionalized it on the field content in the first place - maybe I was trying to be cheeky and reuse empty lists across contexts? Seems wrong in any case.

PR #1871, opened 2025-10-10
Persistently cached declaration maps

gabritto:
Follow up to #1767.

In that PR, we computed `documentPositionMapper`s based on declaration map files and cached them at the language service level. This can be rather inefficient because a language service is created per request, so the cached position mappers were recomputed for every LS request.

In this PR, we cache position mappers in the source file structures themselves, and those get persisted across snapshots and therefore across different language services. We achieve this in the following way:

1. When we get a request in the server, we get an LS from session that is based on the current snapshot. We use this LS to compute the response.
2. We collect all file paths present in the response which could be subjected to declaration mapping.
3. We ask session for a new LS that is based on the old snapshot but contains declaration maps for the files collected in the previous step.
4. We compute a new response by mapping old positions to new positions using this new LS with declaration map information.

To implement step 3, we now have a new operation on snapshots, `CloneWithSourceMaps`, which clones a snapshot with additional declaration map information. This is done by creating a `snapshotFSBuilder` based on the current snapshot, and having the snapshot FS builder read additional files and compute declaration map information.

To persist these newly read files and declaration map information across future snapshots, we also add these files and computed information to the session's current snapshot. This is implemented in `CloneWithDiskChanges`. We also update file watchers to include any newly read disk file that was added as part of those changes.

Note: a snapshot FS has two kinds of files cached, overlays and disk files. Overlays are files open in the client, and their contents may or may not match the corresponding disk files, if those exist. To simplify the implementation here, we only compute declaration map information based on disk files. That means if a .d.ts file or .d.ts.map file is open in the client, we won't use the overlay contents, and will instead read from disk. If the overlay content matches the disk content, there's no noticeable difference. If it doesn't, there will be a mismatch in the mapped positions. But in this scenario, something has gone wrong already: if the user edits a .d.ts file in the client (but doesn't save it to disk), and that .d.ts had a declaration map generated for it, the mapping of positions is already going to be wrong.

### DIFF
--- a/internal/project/snapshot.go
+++ b/internal/project/snapshot.go
@@ -425,7 +425,7 @@ func (s *Snapshot) CloneWithDiskChanges(changes map[tspath.Path]*dirty.Change[*d
 }
 
 func (s *Snapshot) cloneFromSnapshotFSChanges(
-	snapshotFS *snapshotFS,
+	snapshotFS *SnapshotFS,
 	changes map[tspath.Path]*dirty.Change[*diskFile],
 	session *Session,
 	logger *logging.LogTree,
@@ -439,6 +439,7 @@ func (s *Snapshot) cloneFromSnapshotFSChanges(
 		session.extendedConfigCache,
 		s.ConfigFileRegistry,
 		s.compilerOptionsForInferredProjects,
+		s.config,
 		s.toPath,
 	)
 
@@ -462,8 +463,8 @@ func (s *Snapshot) cloneFromSnapshotFSChanges(
 	return newSnapshot
 }
 
-func getComputeGlobPatterns(workspaceDirectory string, useCaseSensitiveFileNames bool) func(files map[tspath.Path]string) patternsAndIgnored {
-	return func(files map[tspath.Path]string) patternsAndIgnored {
+func getComputeGlobPatterns(workspaceDirectory string, useCaseSensitiveFileNames bool) func(files map[tspath.Path]string) PatternsAndIgnored {
+	return func(files map[tspath.Path]string) PatternsAndIgnored {
 		comparePathsOptions := tspath.ComparePathsOptions{
 			CurrentDirectory:          workspaceDirectory,
 			UseCaseSensitiveFileNames: useCaseSensitiveFileNames,
@@ -500,7 +501,7 @@ func getComputeGlobPatterns(workspaceDirectory string, useCaseSensitiveFileNames
 			}
 		}
 
-		return patternsAndIgnored{
+		return PatternsAndIgnored{
 			patterns: globs,
 			ignored:  ignored,
 		}
--- a/internal/ls/hover.go
+++ b/internal/ls/hover.go
@@ -10,6 +10,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/astnav"
 	"github.com/microsoft/typescript-go/internal/checker"
 	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/ls/lsconv"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
 )
 
@@ -491,7 +492,10 @@ func (l *LanguageService) writeJSDocLink(b *strings.Builder, c *checker.Checker,
 		declaration := declarations[0]
 		file := ast.GetSourceFileOfNode(declaration)
 		node := core.OrElse(ast.GetNameOfDeclaration(declaration), declaration)
-		loc := l.getMappedLocation(file.FileName(), createRangeFromNode(node, file))
+		loc := l.getMappedLocation(&lsproto.Location{
+			Uri:   lsconv.FileNameToDocumentURI(file.FileName()),
+			Range: *l.createLspRangeFromNode(node, file),
+		})
 		prefixLen := core.IfElse(strings.HasPrefix(text, "()"), 2, 0)
 		linkText := trimCommentPrefix(text[prefixLen:])
 		if linkText == "" {
--- a/internal/project/bulkcache_test.go
+++ b/internal/project/bulkcache_test.go
@@ -48,7 +48,7 @@ func TestBulkCacheInvalidation(t *testing.T) {
 			session.DidOpenFile(context.Background(), "file:///project/src/index.ts", 1, baseFiles["/project/src/index.ts"].(string), lsproto.LanguageKindTypeScript)
 
 			// Get initial snapshot and verify config
-			ls, err := session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
+			ls, _, err := session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
 			assert.NilError(t, err)
 			assert.Equal(t, ls.GetProgram().Options().Target, core.ScriptTargetES2015)
 
@@ -74,7 +74,7 @@ func TestBulkCacheInvalidation(t *testing.T) {
 			session.DidChangeWatchedFiles(context.Background(), fileEvents)
 
 			// Get language service again to trigger snapshot update
-			ls, err = session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
+			ls, _, err = session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
 			assert.NilError(t, err)
 
 			snapshotAfter, release := session.Snapshot()
@@ -127,7 +127,7 @@ func TestBulkCacheInvalidation(t *testing.T) {
 			session.DidOpenFile(context.Background(), "file:///project/src/index.ts", 1, baseFiles["/project/src/index.ts"].(string), lsproto.LanguageKindTypeScript)
 
 			// Get initial state
-			ls, err := session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
+			ls, _, err := session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
 			assert.NilError(t, err)
 			assert.Equal(t, ls.GetProgram().Options().Target, core.ScriptTargetES2015)
 
@@ -146,7 +146,7 @@ func TestBulkCacheInvalidation(t *testing.T) {
 			assert.NilError(t, err)
 
 			session.DidChangeWatchedFiles(context.Background(), fileEvents)
-			ls, err = session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
+			ls, _, err = session.GetLanguageService(context.Background(), "file:///project/src/index.ts")
 			assert.NilError(t, err)
 
 			if expectConfigReload {
@@ -189,7 +189,7 @@ func TestBulkCacheInvalidation(t *testing.T) {
 		assert.Equal(t, initialProject.Name(), "/project/tsconfig.json", "Should initially use root tsconfig")
 
 		// Get language service to verify initial strict mode
-		ls, err := session.GetLanguageService(context.Background(), "file:///project/src/utils/lib.ts")
+		ls, _, err := session.GetLanguageService(context.Background(), "file:///project/src/utils/lib.ts")
 		assert.NilError(t, err)
 		assert.Equal(t, ls.GetProgram().Options().Strict, core.TSTrue, "Should initially use strict mode from root config")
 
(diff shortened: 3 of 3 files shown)


internal/ls/lsconv/converters.go
```diff
@@ -101,6 +102,9 @@ var extraEscapeReplacer = strings.NewReplacer(
 )
 
 func FileNameToDocumentURI(fileName string) lsproto.DocumentUri {
+	if bundled.IsBundled(fileName) {
```
gabritto, line 105: This is needed to avoid crashes in the tests, since a file name for a bundled file doesn't round trip with the current implementations of `FileNameToDocumentURI` and `documentUri.FileName()`.

internal/lsp/server.go
```diff
@@ -537,6 +537,36 @@ func registerLanguageServiceDocumentRequestHandler[Req lsproto.HasTextDocumentUR
 	}
 }
 
+func registerDefinitionHandler(handlers handlerMap) {
```
gabritto, line 569: I'm planning on making this generic when we add position mapping for other requests.

internal/project/snapshotfs.go
```diff
@@ -100,12 +102,66 @@ func (s *snapshotFSBuilder) GetFileByPath(fileName string, path tspath.Path) Fil
...
+}
+
+// UseCaseSensitiveFileNames implements sourcemap.Host.
+func (s *snapshotFSBuilder) UseCaseSensitiveFileNames() bool {
+	return s.fs.UseCaseSensitiveFileNames()
+}
+
+// GetECMALineInfo implements sourcemap.Host.
+func (s *snapshotFSBuilder) GetECMALineInfo(fileName string) *sourcemap.ECMALineInfo {
+	if file := s.getDiskFile(fileName); file != nil {
```
gabritto, line 175: The argument for why it's safe to call `file.ECMALineInfo()` here is kind of subtle, but basically boils down to the fact that, during the process of computing a new snapshot with added map info, we only ever read files or add the computed map info to existing files. We never change a file's content.

internal/project/snapshotfs.go
```diff
@@ -122,20 +178,82 @@ func (s *snapshotFSBuilder) GetFileByPath(fileName string, path tspath.Path) Fil
...
+	}
+	// Source map information already computed
+	if file.sourceMapInfo != nil {
+		return
+	}
+	url, isInline := sourcemap.GetSourceMapURL(s, genFileName)
+	if isInline {
+		// Store document mapper directly in disk file for an inline source map
+		docMapper := sourcemap.ConvertDocumentToSourceMapper(s, url, genFileName)
+		entry.Change(func(file *diskFile) {
```
gabritto, line 242: This is not atomic or following any concurrency safe pattern because we don't need to: these are always computed sequentially for now.

internal/ls/definition.go
```diff
@@ -57,6 +57,49 @@ func (l *LanguageService) ProvideDefinition(ctx context.Context, documentURI lsp
...
+func (l *LanguageService) GetFilesToMapFromDefinition(response lsproto.DefinitionResponse) []string {
+	var files []string
+
+	if response.Location != nil {
+		files = append(files, response.Location.Uri.FileName())
+	}
+
+	if response.Locations != nil {
+		for _, location := range *response.Locations {
+			files = core.AppendIfUnique(files, location.Uri.FileName())
```
andrewbranch: Perf nit: AppendIfUnique is an order of magnitude slower than building a `map[string]struct{}` and converting to a slice by the time N=100.

andrewbranch: More importantly, I was expecting to see a `tspath.IsDeclarationFileName` condition somewhere around here. Is there ever a case where we would need to try to map a location except when it’s a declaration file?

internal/project/session.go
```diff
@@ -688,3 +692,42 @@ func (s *Session) triggerATAForUpdatedProjects(newSnapshot *Snapshot) {
...
+		return nil, fmt.Errorf("no project found for URI %s", uri)
+	}
+	snapshotWithFiles, changes := snapshot.CloneWithSourceMaps(files, s)
+	if s.options.LoggingEnabled {
+		s.logger.Write(snapshotWithFiles.builderLogs.String())
+		s.logger.Write("")
+	}
+	s.backgroundQueue.Enqueue(ctx, func(ctx context.Context) {
+		s.updateSnapshotWithDiskChanges(changes)
+	})
```
andrewbranch, line 756: I don’t think it saves a _ton_ of work, but couldn't you do

```ts
s.snapshotMu.Lock()
if snapshot == s.snapshot {
  s.snapshot = snapshotWithFiles
} else {
  s.backgroundQueue.Enqueue(...)
}
s.snapshotMu.Unlock()
```

? It seems like the _typical_ case will be that the session snapshot doesn’t change during source map acquisition, so you can just adopt the new one instead of applying the diff in another clone.

internal/project/snapshot.go
```diff
@@ -229,12 +240,37 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
...
+			// Files referenced by projects.
+			rootFiles := collections.Set[tspath.Path]{}
+			// Map of file to the file that references it, e.g. `foo.d.ts.map` -> `foo.d.ts`
+			referencedToFile := map[tspath.Path]tspath.Path{}
 			fs.diskFiles.Range(func(entry *dirty.SyncMapEntry[tspath.Path, *diskFile]) bool {
+				if sourceMapInfo := entry.Value().sourceMapInfo; sourceMapInfo != nil {
+					if sourceMapInfo.sourceMapPath != "" {
+						referencedToFile[s.toPath(sourceMapInfo.sourceMapPath)] = entry.Key()
+					} else if mapper := sourceMapInfo.documentMapper.m; mapper != nil {
+						for _, sourceFile := range mapper.GetSourceFiles() {
```
andrewbranch, line 290: How/when/why does a source mapper point to multiple files? I'm having trouble wrapping my head around what this means.

internal/project/snapshot.go
```diff
@@ -294,10 +330,153 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
 		}
 	}
 
+	newSnapshot.extraDiskFiles = maps.Clone(s.extraDiskFiles)
+	core.DiffMapsFunc(
```
andrewbranch: There can only be changes to source map files if they were deleted by the cleanup logic above, right? The cleanup logic is gated to only run on file open just so most snapshot clones never need to iterate every disk file, so I don’t think we want/need this here. (I bet if you look at snapshot cloning time on keystrokes in large projects, it gets noticeably slower with this.) Do we really need to diff the disk files at all here? It seems like you can just add something more direct in the cleanup logic where the file gets deleted.

internal/project/snapshot.go
```diff
@@ -294,10 +330,153 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
...
+		s.ConfigFileRegistry,
+		s.compilerOptionsForInferredProjects,
+		s.toPath,
+	)
+
+	newSnapshot.parentId = s.id
+	newSnapshot.ProjectCollection = s.ProjectCollection
+	newSnapshot.builderLogs = logger
+	newSnapshot.extraDiskFiles = maps.Clone(s.extraDiskFiles)
+	core.DiffMapsFunc(
```
andrewbranch: I think a similar thing applies here; `s.fs.diskFiles` can be very large, and I think we usually expect `changes` to be relatively small, right? It would probably be much more efficient to compute `extraDiskFiles` from `changes` rather than diffing the full file cache.

jakebailey:
Do we know in general how much faster this is compared to the more naive but simple approach? I was kind of hoping we could get away with "caching within a single request" rather than retaining this memory in later snapshots...

gabritto:
> Do we know in general how much faster this is compared to the more naive but simple approach? I was kind of hoping we could get away with "caching within a single request" rather than retaining this memory in later snapshots...

I'm working on a comparison, just happened to find a bug while doing this, so it's taking a while.

internal/project/snapshot.go
sheetalkamat: i think this use to be `--out` which we no longer support ?

gabritto:
> Do we know in general how much faster this is compared to the more naive but simple approach? I was kind of hoping we could get away with "caching within a single request" rather than retaining this memory in later snapshots...

Did a quick test with a library that had declaration maps which is used in the large repo I was investigating a while back.
Basically using this file:
```ts
import { CSSStyleDeclaration } from "happy-dom"
```
and going to definition on `CSSStyleDeclaration`.

With current per-request cache:
Time to compute the source map information we need for the request is ~17ms (i.e. total time of calling `GetDocumentPositionMapper`).
Definition response time was 20-22ms.

With persistent caches:
First definition request:
~17ms to clone the source maps with computed source map information.
Definition response time: 19ms.

Next definition requests (including after changing the file we're requesting definitions from):
Definition response time: 1ms.

Now, with go to definition, I don't think it's that big of a difference. However, we are going to soon add support for source mapping to find all references, and we are going to have cross-project find all references. So in this scenario, I can see these ~17ms per source map adding up to a noticeable difference.
Note the source map used for the test above is for a .d.ts file of ~1200 lines. I think this is on the larger side, but I could be wrong.

@sheetalkamat what do you think? Do you think we'll need this more efficient version of source map handling?

internal/ls/definition.go
gabritto: > Perf nit: AppendIfUnique is an order of magnitude slower than building a `map[string]struct{}` and converting to a slice by the time N=100.

So I should use `AppendIfUnique`, right? We shouldn't have anywhere close to 100 definition locations.

gabritto: > More importantly, I was expecting to see a `tspath.IsDeclarationFileName` condition somewhere around here. Is there ever a case where we would need to try to map a location except when it’s a declaration file?

Good catch, will fix.

andrewbranch: At N=1 (basically a no-op), `AppendIfUnique` was 2x faster (I guess the time to allocate a slice vs. a map and a slice), but I didn’t check where between 1 and 100 the crossover was. It feels to me like the existence of `AppendIfUnique` invites an antipattern, but it’s fine for very small collections.

internal/project/session.go
gabritto: I thought we didn't want to risk waiting for the snapshot mutex lock before returning the new language service object, since we want to respond to the LS request as fast as possible. But I can add that case as an optimization on the enqueued task, what do you think?

andrewbranch: Sure, I think that makes sense.

jakebailey:
> Now, with go to definition, I don't think it's that big of a difference. However, we are going to soon add support for source mapping to find all references, and we are going to have cross-project find all references. So in this scenario, I can see these ~17ms per source map adding up to a noticeable difference.

I feel like it has to be the case that the number of mappings we need in any given case has a low bound. When are we actually going to be in the situation where we are searching for say definitions and need to look past more than one .d.ts.map?

Find all refs I can understand, but only if source file redirection is disabled, right?

(I would also have sort of expected Lattice to have complained with the current implementation if it were slow but seemingly not. Maybe we need to test it in some internal repos that don't do references?)

gabritto:
> I feel like it has to be the case that the number of mappings we need in any given case has a low bound. When are we actually going to be in the situation where we are searching for say definitions and need to look past more than one .d.ts.map?
> Find all refs I can understand, but only if source file redirection is disabled, right?

Go to definition I don't think will be a problem, which is why I mentioned FAR, but maybe even that is not really a problem, I had forgotten about `disableSourceOfProjectReferenceRedirect` (and how many people even have `disableSourceOfProjectReferenceRedirect` on?). There's also navTo/workspace symbols which uses source mapping, though again, not sure if that could be problematic.

> (I would also have sort of expected Lattice to have complained with the current implementation if it were slow but seemingly not. Maybe we need to test it in some internal repos that don't do references?)

Any ideas which repos we could test on?

I'm ok to just leave this PR here in case we find we need it later. At least we know how to do something like this, if we need it for other things.

sheetalkamat:
I think when we added this "source of project reference" didnt exist so it was like everyone using this. And all scenarios use be affected. If we can keep source map per file - and keep it till the file doesnt change - that should be good enough no given we dont have to handle "--out" scenario? - this also allows sharing across projects (*there use to be some projects but i dont remember which that didnt use project references but had files shared across project - which i think is case when you use disableSourceOfProjectReference?)

The main problem we use to see then use to be those temporary file opens that would loose the source mapper if we stored it with program in LS ( again there use to be problem that these would change the program which we later fixed for more efficient way and i think is present in current snapshot one so we dont have to worry about )

sheetalkamat:
and yes FAR does do source map jumping

jakebailey:
> Go to definition I don't think will be a problem, which is why I mentioned FAR, but maybe even that is not really a problem

If we are caching things per request, then even multi-project FAR that may reread the same file would still work, since the cached thing would still be there (since this is happening a layer above). I feel like it's probable we could save a lot of memory / complexity by not doing any persistence.



> If we can keep source map per file - and keep it till the file doesnt change

I think we can do that, I just wonder if it's worth 20ms to store the mapping in memory potentially forever, given it's likely you will never rebuild the dependent project. Right?

PR #1876, opened 2025-10-11
support pnp resolver

gun-yu:
This PR adds support for PnP resolution.

https://github.com/microsoft/typescript-go/discussions/1875 in the previous discussion, I have internalized pnp-go into the repository.

Please note that the PR has become quite large due to the inclusion of the PnP source and test code. I apologize for the length and appreciate your understanding.

## what is pnp
Yarn Plug’n’Play (PnP) is a dependency resolution system that removes the need for a traditional node_modules folder.
describe at https://github.com/microsoft/typescript-go/issues/460 

### add pnp resolution config in host
```go
type host struct {
	orchestrator *Orchestrator
	host         compiler.CompilerHost
	// Caches that last only for build cycle and then cleared out
	extendedConfigCache tsc.ExtendedConfigCache
	sourceFiles         parseCache[ast.SourceFileParseOptions, *ast.SourceFile]
	configTimes         collections.SyncMap[tspath.Path, time.Duration]

	// caches that stay as long as they are needed
	resolvedReferences parseCache[tspath.Path, *tsoptions.ParsedCommandLine]
	mTimes             *collections.SyncMap[tspath.Path, time.Time]
	resolvedReferences  parseCache[tspath.Path, *tsoptions.ParsedCommandLine]
	mTimes              *collections.SyncMap[tspath.Path, time.Time]
	pnpResolutionConfig *pnp.ResolutionConfig
}
```
### add pnp branch in resolver.go
```go
// resolveTypeReferenceDirective
// skip typeRoots because PnP knows exactly where each @types is located.
	if r.resolver.pnpResolutionConfig != nil {
		if resolvedFromNearestNodeModulesDirectory := r.loadModuleFromNearestNodeModulesDirectory(true /*typesScopeOnly*/); !resolvedFromNearestNodeModulesDirectory.shouldContinueSearching() {
			return r.createResolvedTypeReferenceDirective(resolvedFromNearestNodeModulesDirectory, true /*primary*/)
		}
	} else {
		if len(typeRoots) > 0 {
			if r.tracer != nil {
				r.tracer.write(diagnostics.Resolving_with_primary_search_path_0.Format(strings.Join(typeRoots, ", ")))
			}
```

```go
// loadModuleFromNearestNodeModulesDirectory
if r.resolver.pnpResolutionConfig != nil {
			if result := r.loadModuleFromPNP(priorityExtensions, typesScopeOnly); !result.shouldContinueSearching() {
				return result
			}
		} else {
			if result := r.loadModuleFromNearestNodeModulesDirectoryWorker(priorityExtensions, mode, typesScopeOnly); !result.shouldContinueSearching() {
				return result
			}
		}
```


internal/project/ata/ata.go
```diff
@@ -186,7 +186,7 @@ func (ti *TypingsInstaller) installTypings(
 	if packageNames, ok := ti.installWorker(projectID, requestID, scopedTypings, logger); ok {
 		logger.Log(fmt.Sprintf("ATA:: Installed typings %v", packageNames))
 		var installedTypingFiles []string
-		resolver := module.NewResolver(ti.host, &core.CompilerOptions{ModuleResolution: core.ModuleResolutionKindNodeNext}, "", "")
+		resolver := module.NewResolver(ti.host, &core.CompilerOptions{ModuleResolution: core.ModuleResolutionKindNodeNext}, "", "", nil)
```
gun-yu, line 191: todo: add pnp support later

internal/module/resolver.go
```diff
@@ -947,17 +974,19 @@ func (r *resolutionState) loadModuleFromImmediateNodeModulesDirectory(extensions
...
 
 	return continueSearching()
 }
 
-func (r *resolutionState) loadModuleFromSpecificNodeModulesDirectory(ext extensions, moduleName string, nodeModulesDirectory string, nodeModulesDirectoryExists bool) *resolved {
-	candidate := tspath.NormalizePath(tspath.CombinePaths(nodeModulesDirectory, moduleName))
-	packageName, rest := ParsePackageName(moduleName)
-	packageDirectory := tspath.CombinePaths(nodeModulesDirectory, packageName)
-
+func (r *resolutionState) loadModuleFromSpecificNodeModulesDirectory(ext extensions, candidate string, packageDirectory string, rest string, nodeModulesDirectoryExists bool) *resolved {
```
gun-yu, line 989: I change this function's interface

internal/module/pnp/lib_test.go
```diff
@@ -0,0 +1,301 @@
...
+	if sub == nil || *sub != "a/b/c/index.js" {
+		if sub == nil {
+			t.Fatalf("expected subpath=%q, got nil", "a/b/c/index.js")
+		}
+		t.Fatalf("expected subpath=%q, got %q", "a/b/c/index.js", *sub)
+	}
+}
+
+// need fixtures to be yarn install and make global cache
+func TestGlobalCache(t *testing.T) {
```
gun-yu: These test cases were taken from [pnp-rs](https://github.com/yarnpkg/pnp-rs/tree/main/fixtures/global-cache), but since they are relatively complex to run and maintain.
I think it would be fine to remove them if they’re not considered necessary.
Please let me know your thoughts!

gun-yu:

internal/compiler/host.go
```diff
@@ -49,12 +53,20 @@ func NewCompilerHost(
 	if trace == nil {
 		trace = func(msg string) {}
 	}
+
+	pnpResolutionConfig := TryGetPnpResolutionConfig(currentDirectory)
+
+	if pnpResolutionConfig != nil {
+		fs = pnpvfs.From(fs)
```
gun-yu: On second thought, this way I wouldn’t be able to take advantage of the cached VFS, so I’m thinking of moving its location instead.

jakebailey:
You said some of this comes from pnp-rs. Is it a pure port? What is the license of that code? Do you own it? (You signed the CLA, but that may not be enough if it's not yours.)

gun-yu:
@jakebailey 
The original source code, [pnp-rs](https://github.com/yarnpkg/pnp-rs), is licensed under the BSD 2-Clause License. As the ported code is derived from that source, it should also fall under the BSD 2-Clause License — I missed that part, thank you for pointing it out.

As far as I understand, this means I need to include the original author’s license and copyright notice in the comments.

Would this cause any issues? If everything is fine, I’ll go ahead and add the appropriate BSD 2-Clause License notice.

jakebailey:
We could pull it, but would need to explicitly declare that dependence in some extra metadata to generate NOTICE.txt (something we have avoided entirely by ensuring all code was written by us from spec or tests, but it's not really different than other deps).

But you definitely need to declare that there are files derived from code under a different license.

gun-yu:
@jakebailey
Sorry for the late reply. I just want to confirm my understanding—would it be sufficient to directly add a notice for the BSD 2-Clause License of [pnp-rs](https://github.com/yarnpkg/pnp-rs/blob/main/LICENSE.md
) in the NOTICE.txt file?

jakebailey:
No, the files containing the code need to have headers with said license. For example, these tests: https://github.com/microsoft/typescript-go/blob/a05e479428cb43c6228f78dc04f8326bbd91317b/internal/jsnum/ryu_test.go#L3

gun-yu:
Thank you. I’ve added the license notice as suggested. @jakebailey 
[b368c1c](https://github.com/microsoft/typescript-go/pull/1876/commits/b368c1cf04d53a536b575cdc56b81a8fed6cecf6)

internal/module/pnp/utils/normalize_path.go
```diff
@@ -0,0 +1,64 @@
+package utils
+
+import (
+	"os"
+	"strings"
+)
+
+func NormalizePath(original string) string {
```
jakebailey: I would think we don't need this sort of thing; the `tspath` package already defines the way we typically normalize paths, which is already forward slashed and I think compatible with accessing zips.

internal/module/pnp/utils/portable_path.go
```diff
@@ -0,0 +1,47 @@
...
+	"regexp"
+	"runtime"
+)
+
+var (
+	reWindowsPath     = regexp.MustCompile(`^([a-zA-Z]:.*)$`)
+	reUNCWindowsPath  = regexp.MustCompile(`^[\/\\][\/\\](\.[\/\\])?(.*)$`)
+	rePortablePath    = regexp.MustCompile(`^\/([a-zA-Z]:.*)$`)
+	reUNCPortablePath = regexp.MustCompile(`^\/unc\/(\.dot\/)?(.*)$`)
+)
```
jakebailey: Using regexes is definitely not going to be a good idea; these are notoriously slow. I would hope we could avoid needing any of this.

internal/module/pnp/utils/portable_path.go
```diff
@@ -0,0 +1,47 @@
...
+		if m[1] != "" {
+			return "/unc/.dot/" + m[2]
+		}
+		return "/unc/" + m[2]
+	}
+	return s
+}
+
+func fromPortablePath(s string) string {
+	if runtime.GOOS != "windows" {
```
jakebailey: Depending on `runtime.GOOS` is a bad idea; all of our other path handling is platform independent.

internal/module/pnp/utils/portable_path.go
```diff
@@ -0,0 +1,47 @@
+package utils
```
jakebailey: Meta comment, but this package name is too simple. we'd probably name this `pnpmutils` or something.

internal/module/pnp/builtins.go
```diff
@@ -0,0 +1,62 @@
+package pnp
+
+import "slices"
+
+var NodeJSBuiltins = []string{
```
jakebailey: We already have this array defined somewhere else.

internal/module/pnp/lib.go
```diff
@@ -0,0 +1,491 @@
...
+
+func (e *MissingPeerDependencyError) Error() string { return e.Message }
+
+type FailedManifestHydrationError struct {
+	Message      string
+	ManifestPath string
+	Err          error
+}
+
+func (e *FailedManifestHydrationError) Error() string {
```
jakebailey: As a meta comment, we typically aren't using `error` at all in the codebase outside interacting with things that need it (like the LSP layer). I sort of expect that a more final version of this PR look more like `resolver` in the way it tracks things.

internal/module/pnp/manifest.go
```diff
@@ -0,0 +1,231 @@
...
+	"errors"
+	"regexp"
+
+	"github.com/go-json-experiment/json"
+	"github.com/go-json-experiment/json/jsontext"
+
+	"github.com/microsoft/typescript-go/internal/module/pnp/utils"
+)
+
+type Trie[T any] = utils.Trie[T]
```
jakebailey: This to me says this should just be defined in this package?

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+		zipFiles:      make(map[string]*zipFile),
+	}
+}
+
+func (fs *PNPFS) checkForZip(path string, kind EntryKind) (*zipFile, string) {
+	var zipPath string
+	var pathTail string
+
+	// Do a quick check for a ".zip" in the path at all
+	path = strings.ReplaceAll(path, "\\", "/")
```
jakebailey: `path` is guaranteed to not contain any backslashes.

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+	if i := strings.Index(path, ".zip/"); i != -1 {
```
jakebailey: Probably `strings.Cut` would suffice here?

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+	contents string
+	err      error
+	wasRead  bool
+}
+
+func From(baseFS vfs.FS) *PNPFS {
+	return &PNPFS{
+		inner: baseFS,
+
+		zipFilesMutex: sync.Mutex{},
```
jakebailey: There's no need to init this, the zero value is useful.

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// If there is one, then check whether it's a file on the file system or not
+	fs.zipFilesMutex.Lock()
+	archive := fs.zipFiles[zipPath]
+	if archive != nil {
+		fs.zipFilesMutex.Unlock()
+		archive.wait.Wait()
+	} else {
+		archive = &zipFile{}
+		archive.wait.Add(1)
```
jakebailey: I'm pretty sure this could be done instead with a `sync.Once` or channel-closed based pattern; using a waitgroup for this feels like overkill / unclear.

internal/vfs/zipvfs/zipvfs.go
```diff
@@ -0,0 +1,399 @@
...
+	if fs.inner.FileExists(path) {
+		return true
+	}
+
+	zip, pathTail := fs.checkForZip(path, FileEntry)
+	if zip == nil {
+		return false
+	}
+
+	_, ok := zip.files[strings.ToLower(pathTail)]
```
jakebailey, line 196: I am suspicious about `ToLower` here, compared to `tspath.Path`.

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+	return fs.inner.WriteFile(path, data, writeByteOrderMark)
+}
+
+func (fs *PNPFS) Remove(path string) error {
+	return fs.inner.Remove(path)
+}
+
+func (fs *PNPFS) Chtimes(path string, aTime time.Time, mTime time.Time) error {
+	return fs.inner.Chtimes(path, aTime, mTime)
+}
```
jakebailey: Probably these funcs should error or panic if ever tried on a zip path?

internal/vfs/pnpvfs/pnpvfs.go
```diff
@@ -0,0 +1,399 @@
...
+		return entries
+	}
+
+	dir, ok := zip.dirs[strings.ToLower(pathTail)]
+	if !ok {
+		return entries
+	}
+
+	files := make([]string, 0)
+	dirs := make([]string, 0)
```
jakebailey: `make([]string, 0)` is just an allocate-y version of `var files []string`; don't intentionally make an empty slice unless `nil` vs non-`nil` matters.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/microsoft/typescript-go
 
-go 1.25
+go 1.25.1
```
jakebailey: Please undo this.

internal/module/pnp/lib.go
```diff
@@ -0,0 +1,491 @@
...
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			break
+		}
+		dir = parent
+	}
+	return "", false
+}
+
+var rePNP = regexp.MustCompile(`(?s)(const[\ \r\n]+RAW_RUNTIME_STATE[\ \r\n]*=[\ \r\n]*|hydrateRuntimeState\(JSON\.parse\()'`)
```
jakebailey: This scares me a bit; because of the regex, but also because this reads JS files? Does the spec really require such a thing?

internal/module/pnp/lib.go
```diff
@@ -0,0 +1,491 @@
...
+		return "", nil, &FailedManifestHydrationError{
+			Message:      "Invalid specifier",
+			ManifestPath: spec,
+		}
+	}
+	return pkg, sub, nil
+}
+
+func FindClosestPNPManifestPath(start string) (string, bool) {
+	dir := filepath.Clean(start)
```
jakebailey: Nothing in our repo should ever make use of `filepath`; everything should use `tspath`.

internal/module/pnp/manifest.go
```diff
@@ -0,0 +1,231 @@
...
+			innerMap[k2] = info
+		}
+		res[k1] = innerMap
+	}
+	*m = res
+	return nil
+}
+
+func (r *RegexDef) compile() (*regexp.Regexp, error) {
+	return regexp.Compile(r.Source)
```
jakebailey: This will fail because `regex` is re2 syntax, not ECMAScript. You'd have to use `regexp2` in ECMAScript mode.

That being said, is this really ever used? Does the PnP spec really require regex parsing? (How do esbuild and so on do this without pulling on totally different regex implementations?)

gun-yu:
@jakebailey Thank you for the thoughtful review! I know feedback like this takes a lot of time, and I truly appreciate it. Reading your comments made me realize how unfamiliar I still am with the TypeScript Go codebase—I’m sorry I didn’t reference the existing code more. Consistent code style is especially important in open source, so I’m happy to adopt all of your suggestions. I’ll start by addressing your comments.

internal/module/pnp/manifest.go
gun-yu: Good point! esbuild uses the JS AST for parsing. Since I was primarily focused on porting pnp-rs, I initially brought the code over as-is, but as you mentioned in another comment, there are definitely some risky areas. I believe the TypeScript Go codebase also includes a JS parser, so I’ll try using that for parsing instead.

internal/module/pnp/utils/normalize_path.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I replaced normalizePath with tspath.normalizePath.

internal/module/pnp/utils/portable_path.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I removed this file as normalizePath has been removed.

internal/module/pnp/utils/portable_path.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I removed this file as normalizePath has been removed.

internal/module/pnp/utils/portable_path.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I removed this file as normalizePath has been removed.

internal/module/pnp/builtins.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I’m using NodeCoreModules instead.

internal/module/pnp/manifest.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
I removed the utils folder and moved its contents into the pnp package.
Does this direction look okay to you, or would it be better to place it under internal/collections instead?
I’ve implemented only the Trie features I needed for now, so I included it in pnp for the time being.

internal/module/pnp/lib.go
gun-yu: [305fd49](https://github.com/microsoft/typescript-go/pull/1876/commits/305fd49550f74402b0dd28cffa7adfb4d090ead6)
using tspath instead

internal/module/pnp/lib.go
gun-yu: Good point! esbuild uses the JS AST for parsing. Since I was primarily focused on porting pnp-rs, I initially brought the code over as-is, but as you mentioned in another comment, there are definitely some risky areas. I believe the TypeScript Go codebase also includes a JS parser, so I’ll try using that for parsing instead.

gun-yu: According to the official spec, in a Node.js environment, pnp.cjs patches Node’s module resolution, so it’s executed first to override the default behavior. In Go, however, this part needs to be implemented manually.

internal/module/pnp/lib.go
gun-yu: Great, you mean handling it using `trace.write` like in `resolver.go`, right?
I think it makes sense to follow that error handling approach.
However, it seems that these error messages generally follow some kind of convention. I’m not sure if it’s okay for me to add messages arbitrarily in a file like `diagnostics.generate`. Is there any guideline I should follow for this?

internal/module/pnp/manifest.go
gun-yu: Thanks for the feedback!
The original `ignorePatternData` looks like this:

```json
"ignorePatternData": "(^(?:\\\\.yarn\\\\/sdks(?:\\\\/(?!\\\\.{1,2}(?:\\\\/|$))(?:(?:(?!(?:^|\\\\/)\\\\.{1,2}(?:\\\\/|$)).)*?)|$))$)",
```

[Esbuild](https://github.com/evanw/esbuild/blob/main/internal/resolver/yarnpnp.go#L327) also compiles and uses the regex in the same way. Since this is part of the Yarn PnP spec, it seems unavoidable to use a regular expression here.

For now, I’ve optimized this commit to avoid the inefficiency of compiling the regex on every use. [f0ee55b](https://github.com/microsoft/typescript-go/pull/1876/commits/f0ee55baae946a3fcbace07443488162481dedca)

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [f0ee55b](https://github.com/microsoft/typescript-go/pull/1876/commits/f0ee55baae946a3fcbace07443488162481dedca)
remove replaceAll

internal/module/pnp/lib.go
gun-yu: Ultimately, parsing pnp.cjs is necessary to read the metadata, so this part seems unavoidable.
If we parse it using an AST like esbuild does, it might be a bit more reliable — would that be okay?

internal/module/pnp/lib.go
jakebailey: Yes, https://github.com/microsoft/typescript-go/blob/main/internal/diagnostics/extraDiagnosticMessages.json

You may want to look at the existing PnP PR for inspiration as well

internal/module/pnp/manifest.go
jakebailey: It could go there, yes.

internal/module/pnp/lib.go
jakebailey: That makes more sense to me. Though in general I really thought that PnP had been specified to just emit a plain JSON file...

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [ae272f1](https://github.com/microsoft/typescript-go/pull/1876/commits/ae272f16e2024b1221878883eafd07dbbef28182)

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [ae272f1](https://github.com/microsoft/typescript-go/pull/1876/commits/ae272f16e2024b1221878883eafd07dbbef28182)

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [ae272f1](https://github.com/microsoft/typescript-go/pull/1876/commits/ae272f16e2024b1221878883eafd07dbbef28182)
add sync.Once and remove waitgroup

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [ae272f1](https://github.com/microsoft/typescript-go/pull/1876/commits/ae272f16e2024b1221878883eafd07dbbef28182)
add assertion on writeFile, Remove, Chtimes

internal/vfs/pnpvfs/pnpvfs.go
gun-yu: [ae272f1](https://github.com/microsoft/typescript-go/pull/1876/commits/ae272f16e2024b1221878883eafd07dbbef28182)

internal/vfs/zipvfs/zipvfs.go
gun-yu: I'm not sure I fully understood. Which part do you find suspicious?
If it’s to be replaced, do they mean to use `tspath.NormalizePath` instead?

internal/module/pnp/lib.go
gun-yu: [67752dc](https://github.com/microsoft/typescript-go/pull/1876/commits/67752dcd841cfc9901ee38f8d09a9a3354af21c1)
I’ve added error messages following the tsgo format and removed the custom errors.

The PnP APIs ultimately return a single error. Would it be okay to handle that error by tracing it (or something similar) at the point where it’s used?

I was also considering attaching a trace logger or instantiating one for the PnP APIs, but I couldn’t find a suitable initialization point, so I’m not sure what would be a good design approach.

Still, if returning the error directly doesn’t align well with the overall code style, I think attaching it as a method of resolutionState might be a better fit.

internal/module/pnp/lib.go
gun-yu: [a9dbc68](https://github.com/microsoft/typescript-go/pull/1876/commits/a9dbc68bf4868159b92cd2e267b91eb621fca7ab)
I’ve changed the implementation from using regex to parsing with an AST.
The original pnp-rs implementation (the official one by the Yarn team) also uses regex, so I believe the regex-based approach is stable.
However, if switching to the AST-based approach is considered better, I’m happy to proceed with it — the code works fine as is.

go.mod
gun-yu: df40d62

gun-yu:
@jakebailey 
I believe I’ve incorporated all the review feedback.
Please let me know if there’s anything I might have overlooked.

internal/module/pnp/lib.go
gun-yu: Also, I’m a bit concerned that including the full error in the trace could make the messages too verbose. That’s because if it fails to find tens of thousands of files, the trace would become extremely long.

internal/module/pnp/lib.go
gun-yu: In my project, where pnp.cjs is around 200,000 lines, the regex implementation performed slightly better.

jakebailey: https://github.com/yarnpkg/pnp-rs was happy with a regex, then I don't mind behind happy with a regex. I don't know if this is even in the hot path. My main concern was just whether or not it was "stable".

internal/module/pnp/manifest.go
jakebailey: If esbuild gets away with it, then that's fine I suppose.

Out of curiosity, is this string _always_ the same?

internal/module/pnp/lib.go
jakebailey: The trace is for debugging, it's generally okay for it to be long, though there is a balance. How long are you talking?

internal/vfs/zipvfs/zipvfs.go
jakebailey: lowercasing for normalization is sometimes unsafe due to weird unicode rules, hence why ToPath has its own algorithm to do so.

If pnp-rs/pnp ifself get away with it, then it could be fine to do so.

internal/compiler/pnp.go
```diff
@@ -0,0 +1,54 @@
+package compiler
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
```
jakebailey: The compiler should not depend on these; any accesses it makes must go through a provided FS.

internal/module/pnp/manifest.go
```diff
@@ -0,0 +1,248 @@
...
+	PackageDependencies FallbackPool `json:"packageDependencies"`
+}
+
+type PackageDependency struct {
+	Reference string
+	Alias     [2]string
+	IsAlias   bool
+}
+
+func (p *PackageDependency) UnmarshalJSON(data []byte) error {
```
jakebailey: Not strictly required, but all of these UnmarshalJSON functions would perform better and probably be simpler to write using the v2 versions of this, as in `UnmarshalJSONFrom(dec *jsontext.Decoder) error`).

internal/module/resolver.go
```diff
@@ -1973,7 +2003,7 @@ func extensionIsOk(extensions extensions, extension string) bool {
 }
 
 func ResolveConfig(moduleName string, containingFile string, host ResolutionHost) *ResolvedModule {
-	resolver := NewResolver(host, &core.CompilerOptions{ModuleResolution: core.ModuleResolutionKindNodeNext}, "", "")
+	resolver := NewResolver(host, &core.CompilerOptions{ModuleResolution: core.ModuleResolutionKindNodeNext}, "", "", nil)
```
jakebailey, line 2007: This is an interesting one; `nil` here means that one can't load tsconfigs through PnP, which is probably fine, but I am curious how the old PnP patch did this (if at all).

internal/module/resolver_pnp.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (r *resolutionState) loadPNPResolutionPath(moduleName string) (string, error) {
+	resolution, err := pnp.ResolveToUnqualified(moduleName, r.containingDirectory, r.resolver.pnpResolutionConfig)
+	if err != nil {
+		return "", err
+	}
+
+	// trim trailing slash makes a bug in packageJsonInfoCache.Set
+	// like @emotion/react/ -> @emotion/react after packageJsonInfoCache.Set
+	// check why it's happening in packageJsonInfoCache and need to fix it
```
jakebailey: Yes, because the cache I think assumes normalized paths with trailing trimmed. Definitely need to ensure paths are minimal like that. I think.

internal/project/ata/ata.go
jakebailey: Unless you plan to do it in this PR, we this should have a comment in the code.

internal/compiler/host.go
```diff
@@ -49,12 +53,20 @@ func NewCompilerHost(
 	if trace == nil {
 		trace = func(msg string) {}
 	}
+
+	pnpResolutionConfig := TryGetPnpResolutionConfig(currentDirectory)
+
+	if pnpResolutionConfig != nil {
+		fs = zipvfs.From(fs)
+	}
```
jakebailey: I do wonder if this is the right layer for this. Surely the language server also needs to know this information? For example, we need to read random files to do source mapping in the LS layer, which would have to dig into zips.

But, we don't want cached info to be there indefinitely either.

I also don't know where this should go in relation to the caching layer...

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/microsoft/typescript-go
 
-go 1.25
+go 1.25.0
```
jakebailey: This is still the wrong version.


```suggestion
go 1.25
```

internal/compiler/pnp.go
jakebailey: (https://github.com/microsoft/typescript-go/pull/1911 will enforce this in the future)

internal/module/pnp/lib.go
gun-yu: I understand. pnp-rs uses regex while esbuild uses AST, and both are stable implementations.
As you mentioned, this isn’t a hot path and there shouldn’t be any significant difference in performance, so I think it’s fine.
I’ll keep the AST-based implementation as it is in the commit.

internal/vfs/zipvfs/zipvfs.go
gun-yu: Thanks for the great point! I agree that using ToPath would be a more stable and reliable approach.
There’s no need to go against what you mentioned, and your concerns are absolutely valid.

By the way, the zipvfs implementation was inspired by esbuild’s fszip.
Do you think it would be better to explicitly mention that reference in our license file as well?
(esbuild is under the MIT license.)

internal/module/pnp/manifest.go
gun-yu: According to the specification, I understand that this value is dynamic.
[yarn berry](https://github.com/yarnpkg/berry/blob/a592371321068c3d63b0932030ebbdede611610c/packages/docusaurus/docs/advanced/03-pnp/pnp-spec.mdx#L251)

internal/module/resolver.go
gun-yu: I actually wanted to get feedback on this part, but I forgot to mention it in the description — thanks for pointing it out!

I think adding PnP resolution would be the right approach, but for now, I set it to nil because I wasn’t sure how best to pass the pnpResolutionConfig from the host to NewResolver given the current structure.

(I’ll also mention this here since the concern is similar to the zipvfs layer you pointed out.)

I considered two possible approaches:

1. Inject both pnpResolutionConfig and zipvfs externally.

2. Handle them inside NewCompilerHost (which is how it’s currently implemented in this PR).

The first approach would make the PnP-related code spread across quite a few parts of the system, which I’d prefer to avoid — I’d like to keep PnP logic as isolated as possible.

The second approach avoids scattering PnP logic across tsgo, but it introduces some implicit behavior in NewCompilerHost and prevents taking advantage of cachedvfs.

From a maintainer’s perspective, do you have any thoughts on which direction would be preferable?

gun-yu:
>Unless you plan to do it in this PR, we this should have a comment in the code.

Once the direction for the NewCompilerHost design is decided, I think I can add everything.

However, I consider the ata file related to LSP behavior, and I haven’t really tested LSP properly yet, nor do I fully understand it, so I’m a bit concerned about that part.

I’m fine either working on it immediately or continuing in the next PR. I can proceed according to your guidance.

internal/module/pnp/lib.go
gun-yu: It was just a concern I hadn’t really thought through: if there are 10,000 files, we might end up printing 10,000 error logs.

If you’re fine with it, I don’t see a problem either, and I can go ahead and add it via trace.write.

Also, is the way I added the error messages okay with you? [67752dc](https://github.com/microsoft/typescript-go/pull/1876/commits/67752dcd841cfc9901ee38f8d09a9a3354af21c1)

go.mod
gun-yu: [bfa05c3](https://github.com/microsoft/typescript-go/pull/1876/commits/bfa05c307e0fb1b2262cc4a2b1bf3a6c01dadfc2)
sorry for my mistake

internal/module/pnp/manifest.go
gun-yu: [bfa05c3](https://github.com/microsoft/typescript-go/pull/1876/commits/bfa05c307e0fb1b2262cc4a2b1bf3a6c01dadfc2)

jakebailey: That's not exactly what I meant per se; you want to read from the input in a more streaming manner. I would look at other instances of the method, since by calling Value you are just doing what the old code did and double decoding.

gun-yu: [e697dce](https://github.com/microsoft/typescript-go/pull/1876/commits/e697dcee8beb826f317b153f5f7ea78f61548daa)
I did some work on it, but the code has gotten quite complicated, so I'm not sure if it's heading in the right direction. There might be some minor memory or performance gains, but I haven’t benchmarked it yet.

internal/compiler/pnp.go
gun-yu: [1041038](https://github.com/microsoft/typescript-go/pull/1876/commits/10410382498dc30d98b70c2b29673a50fe8aae93)
I replaced the implementation to use Yarn’s FindManifest API and switched from using the OS-based implementation to a VFS-based one.

internal/module/pnp/manifest.go
gun-yu: I did a simple timing test, and the new implementation was about twice as fast. Nice!

internal/compiler/host.go
gun-yu: [01b7670](https://github.com/microsoft/typescript-go/pull/1876/commits/01b76704a7280dea56675eaa85f95a0c986795aa)
I’ve modified the interface to inject pnpResolutionConfig as shown in this commit. What do you think about this approach?

internal/module/resolver.go
gun-yu: [01b7670](https://github.com/microsoft/typescript-go/pull/1876/commits/01b76704a7280dea56675eaa85f95a0c986795aa)
I’ve modified the interface to inject pnpResolutionConfig as shown in this commit. What do you think about this approach?

internal/project/ata/ata.go
gun-yu: [6b2f0c1](https://github.com/microsoft/typescript-go/pull/1876/commits/6b2f0c186e17a6feacc6c7860670d3418cfa6111)
I’ve added a comment noting that automatic type installation is not supported in the PnP environment.

internal/module/resolver_pnp.go
gun-yu: [6747e23](https://github.com/microsoft/typescript-go/pull/1876/commits/6747e2343a417e93ed6ca025249ad8f41ebf5a5d)
It’s a bit unfortunate that I have to explain this behavior in a comment, but I’ve written it this way for now.

gun-yu:
@jakebailey 
I noticed that there’s another PR with a similar goal: [#1966](https://github.com/microsoft/typescript-go/pull/1966).
To avoid overlapping work, could you please advise on the best way to move forward?

GGomez99:
> I noticed that there’s another PR with a similar goal: microsoft/typescript-go#1966. To avoid overlapping work, could you please advise on the best way to move forward?

Hey @gun-yu! We were aware of your PR, and we are totally open to collaborate on this feature, especially since we both implemented features that the other didn't focus on. We already sent you a couple of email on this 🙇

gun-yu:
> > I noticed that there’s another PR with a similar goal: microsoft/typescript-go#1966. To avoid overlapping work, could you please advise on the best way to move forward?
> 
> Hey @gun-yu! We were aware of your PR, and we are totally open to collaborate on this feature, especially since we both implemented features that the other didn't focus on. We already sent you a couple of email on this 🙇

Thank you for your response! I’d be very happy to collaborate if possible.

I also took a look at your PR — I saw that it already implements both the PnP VFS and metadata parsing.
I was particularly thinking about where the PnP VFS should fit in the project structure, and I found the approach in that PR quite inspiring.

I just noticed your email recently — I don’t check my inbox very often, so I apologize for the delayed reply!

gun-yu:
As part of the joint work with the Datadog team, I’ve merged some of my implementations into @GGomez99  branch. [6e2eaff](https://github.com/microsoft/typescript-go/pull/1966/commits/6e2eaff0392be6aa4abea0cfbc03d22a3c14d6ef)
I’m planning to close this PR, would that be alright, @jakebailey ?

jakebailey:
Of course, sure

PR #1884, opened 2025-10-15
Use `LocationLink` in go to definition

### LINKED ISSUE #1439: Support `LocationLink` in go-to-definition

gabritto:
This PR makes go to definition and go to type definition use `LocationLink`s, which has roughly the same shape of what we return in Strada.
It also adds support for location link's origin selection range in the baseline tests. In Strada, we had two `verify.baseline...` methods for go to def, one which included the origin selection range (response body's `textSpan`), and one which did not, so now we do the same here.

The origin selection range also shows up in baselines, although in a buggy way: it only shows up if there are no definition locations in the same file. I believe this is a bug in Strada's fourslash, but I've preserved the behavior so it's easier to compare the baselines. Another oddity here is that Strada's fourslash does not use the `contextSpan` property of definitions in the baselines, even though baselines for e.g. rename do use context spans (they show up as `<| ... |>` in baselines).

I'll enable go to def baseline diffs in another PR, and this PR also helps reduce noise in the diffs.

Fixes #1439


internal/lsp/server.go
```diff
@@ -921,3 +921,10 @@ func getCompletionClientCapabilities(params *lsproto.InitializeParams) *lsproto.
 	}
 	return params.Capabilities.TextDocument.Completion
 }
+
+func getDefinitionClientCapabilities(params *lsproto.InitializeParams) *lsproto.DefinitionClientCapabilities {
```
jakebailey: At some point I want to try and consolidate this with helpers on `ctx`, since everything seems to use a different mechanism to control what they send (no action needed now)

internal/lsp/server.go
```diff
@@ -763,11 +763,11 @@ func (s *Server) handleSignatureHelp(ctx context.Context, languageService *ls.La
 }
 
 func (s *Server) handleDefinition(ctx context.Context, ls *ls.LanguageService, params *lsproto.DefinitionParams) (lsproto.DefinitionResponse, error) {
-	return ls.ProvideDefinition(ctx, params.TextDocument.Uri, params.Position)
+	return ls.ProvideDefinition(ctx, params.TextDocument.Uri, params.Position, getDefinitionClientCapabilities(s.initializeParams))
 }
 
 func (s *Server) handleTypeDefinition(ctx context.Context, ls *ls.LanguageService, params *lsproto.TypeDefinitionParams) (lsproto.TypeDefinitionResponse, error) {
-	return ls.ProvideTypeDefinition(ctx, params.TextDocument.Uri, params.Position)
+	return ls.ProvideTypeDefinition(ctx, params.TextDocument.Uri, params.Position, getDefinitionClientCapabilities(s.initializeParams))
```
jakebailey: There's actually another set of capabilities for TypeDefinition that has to be used here. Annoying, sorry

jakebailey (changes requested):

internal/lsp/server.go
jakebailey: Given the caps only have `linkSupport` (and then `dynamicRegistration` which we will never use), you could probably get away with a bool for both and just extract `linkSupport` here.

jakebailey (approved):

PR #1886, opened 2025-10-15
Don’t look in JSExportAssignment and CommonJSExport for nodes

andrewbranch:
When looking for imports to resolve, we were hitting the cloned function inside the fake reparsed export, which says it has JSDoc attached, but isn’t actually the key into the JSDoc map because it’s the clone.


jakebailey:
I missed this while I was gone, but I find this special casing strange; why is it that such a generic algorithm needs an exception for these two nodes and not other JSDoc related stuff? Shouldn't we make the map have both keys or something rather than checking everywhere?

Issue #1657, opened 2025-09-01
tsgo preview: fails to show import suggestions for node modules and packages that are linked through PNPM workspaces


naman-luthra-rubrik:
## Steps to reproduce
Try importing any module from node_modules or pnpm linked workspaces

## Behavior with `typescript@5.8`
We get import suggestions for the modules

## Behavior with `tsgo`
We don't get any import suggestions

jakebailey:
Do you have a repo or example we could test?

iisaduan:
cc @donprasetiyo, @AlokTakshak from comments on #1084 

> This seems to only work for non-nodemodules files. Is this intentional?
> This is also true for packages that are linked through PNPM workspaces.

For the time being, yes. There's two related featuresthat could be causing what you're seeing
- `packageJsonAutoImportProvider`: At the moment, the LS can only see _files_ that are in the current project, so if you are trying to make a reference to a node modules package (or any other kind of package) that is not yet referenced in your current project, auto imports will not be able to see the new package. 
- cross project LS: We also don't have cross project support for any of the LS features yet.

Both are currently being worked on.

This means that autoimports _should_ work on already referenced files, so you've already referenced the package in your project, but you can't get new auto-imports for that package, please provide a repo so we can take a look!

christophe-g:
> Do you have a repo or example we could test?

The repo used in #836 should illustrate it: 

<img width="1339" height="485" alt="Image" src="https://github.com/user-attachments/assets/bc46c401-925e-490d-8161-5488f2d6e4b0" />

The expected behavior would be that an import from `@vaadin/button/vaadin-button.js` is proposed for `Button`. 

By contrast an auto-import is proposed for `SemVer` - semver being imported from nmp and not symlinked on locale. 

## Steps to reproduce

git clone [git@github.com](mailto:git@github.com):christophe-g/tsgo-test-linked-dep.git
git clone --depth 1 [git@github.com](mailto:git@github.com):vaadin/web-components.git
cd tsgo-test-linked-dep
pnpm i
open in VScode with `typescriptteam.native-preview` extension

## Dependencies

```json
  "dependencies": {
    "@vaadin/button": "file:../web-components/packages/button",
    "semver": "^7.7.1"
  }
```

juliomuhlbauer:
Using bun with linker = "isolated" does not work as well

In my case I have an error:
```
[Error - 12:09:09 PM] Request textDocument/completion failed.
[...]

Issue #1347, opened 2025-07-02
Destructuring re-exports using type from symlinked node-modules results in relative paths used in `import()` type


AlCalzone:
Feels like I'm hitting all the edge cases. This can be reproduced using the following test case:

```ts
// @Filename: /packages/b/package.json
{
  "name": "package-b",
  "type": "module",
  "exports": {
    ".": "./index.js"
  }
}

// @Filename: /packages/b/index.js
export {};

// @Filename: /packages/b/index.d.ts
export interface B {
	b: "b";
}

// @Filename: /packages/a/package.json
{
  "name": "package-a",
  "type": "module",
  "imports": {
    "#re_export": "./src/re_export.ts"
  },
  "exports": {
    ".": "./dist/index.js"
  }
}


// @Filename: /packages/a/tsconfig.json
{
  "compilerOptions": {
    "module": "nodenext",
    "outDir": "dist",
    "rootDir": "src",
    "declaration": true,
  },
  "include": ["src/**/*.ts"]
}

// @Filename: /packages/a/src/re_export.ts
import type { B } from "package-b";
declare function foo(): Promise<B>
export const re = { foo };

// @Filename: /packages/a/src/index.ts
import { re } from "#re_export";
const { foo } = re;
export { foo };

// @link: /packages/b -> /packages/a/node_modules/package-b
```

The relevant part are the type declarations generated for `package-a`:
```ts
//// [re_export.d.ts]
import type { B } from "package-b";
declare function foo(): Promise<B>;
export declare const re: {
    foo: typeof foo;
};
export {};
//// [index.d.ts]
declare const foo: () => Promise<import("../../b/index.js").B>;
export { foo };
```

I would have expected the following declaration:
```
declare const foo: () => Promise<import("package-b").B>;
export { foo };
```

jakebailey:
Thanks for reporting these, especially for coming up with actual compiler tests.

PR #1902, opened 2025-10-18
Fix #1034: Improve symlink resolution in module specifier generation

### LINKED ISSUE #1034: [pnpm] Error 2742 The inferred type of xxx cannot be named
With same code, just replacing tsc with tsgo gives me multiple TS 2742 errors
like these
```
error TS2742: The inferred type of 'router' cannot be named without a reference to '../../node_modules/@trpc/server/dist/unstable-core-do-not-import.js'. This is likely not portable. A type annotation is necessary.
```

Repro with instructions
https://github.com/vjau/tsgo-error

chase:
Extends the symlink support in GetEachFileNameOfModule to properly resolve module specifiers across symlinked packages and workspaces. 

Key changes:
- Move knownsymlinks from compiler to dedicated symlinks package
- Implement active resolution via ResolveModuleName to populate cache
- Add dependency resolution from package.json to detect symlinks early
- Improve ignored path handling (node_modules/., .git, .# emacs locks)
- Add comprehensive test coverage for symlink resolution
- Fix declaration emit to prefer original paths over symlink paths

Fixes https://github.com/microsoft/typescript-go/issues/1657 
Fixes https://github.com/microsoft/typescript-go/issues/1034
Fixes https://github.com/microsoft/typescript-go/issues/1347


chase:
My intention is to help fix #1034, which is the only thing left blocking my team from adopting tsgo.

I'm not very experienced with Go, but I wanted to expand upon @shinichy's work here: https://github.com/shinichy/typescript-go/tree/symlinks

neo773:
We have the same issue in our team that's stopping us from adopting `tsgo`

So I compiled your branch and ran a test, it doesn't seem to make any difference in our code base. Just wanted to let you know.

```
neo@neos-MacBook-Pro twenty-server % /Users/neo/Desktop/typescript-go/built/local/tsgo          
src/engine/api/mcp/services/tools/mcp-metadata-tools.service.ts:90:9 - error TS2742: The inferred type of 'send' cannot be named without a reference to '../../../../../../../../node_modules/axios/index.d.cts'. This is likely not portable. A type annotation is necessary.

90   async send(requestContext: RequestContext, data: Query) {
           ~~~~

src/engine/api/rest/metadata/rest-api-metadata.service.ts:24:9 - error TS2742: The inferred type of 'get' cannot be named without a reference to '../../../../../../../node_modules/axios/index.d.cts'. This is likely not portable. A type annotation is necessary.

24   async get(request: Request) {
           ~~~

src/engine/api/rest/metadata/rest-api-metadata.service.ts:38:9 - error TS2742: The inferred type of 'create' cannot be named without a reference to '../../../../../../../node_modules/axios/index.d.cts'. This is likely not portable. A type annotation is necessary.

38   async create(request: Request) {
           ~~~~~~

src/engine/api/rest/metadata/rest-api-metadata.service.ts:52:9 - error TS2742: The inferred type of 'update' cannot be named without a reference to '../../../../../../../node_modules/axios/index.d.cts'. This is likely not portable. A type annotation is necessary.

52   async update(request: Request) {
           ~~~~~~

src/engine/api/rest/metadata/rest-api-metadata.service.ts:66:9 - error TS2742: The inferred type of 'delete' cannot be named without a reference to '../../../../../../../node_modules/axios/index.d.cts'. This is likely not portable. A type annotation is necessary.

66   async delete(request: Request) {
           ~~~~~~

src/engine/metadata-modules/constants/search-vector-field.constants.ts:3:14 - error TS2742: The inferred type of 'SEARCH_VECTOR_FIELD' cannot be named without a reference to '../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

3 export const SEARCH_VECTOR_FIELD = {
               ~~~~~~~~~~~~~~~~~~~

src/engine/twenty-orm/relation-nested-queries/utils/formatConnectRecordNotFoundErrorMessage.util.ts:5:14 - error TS2742: The inferred type of 'formatConnectRecordNotFoundErrorMessage' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

5 export const formatConnectRecordNotFoundErrorMessage = (
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/prefill-views.ts:21:14 - error TS2742: The inferred type of 'prefillViews' cannot be named without a reference to '../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

21 export const prefillViews = async (
                ~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/companies-all.view.ts:12:14 - error TS2742: The inferred type of 'companiesAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

12 export const companiesAllView = (
                ~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/custom-all.view.ts:6:14 - error TS2742: The inferred type of 'customAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

6 export const customAllView = (
               ~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/dashboards-all.view.ts:11:14 - error TS2742: The inferred type of 'dashboardsAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

11 export const dashboardsAllView = (
                ~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/notes-all.view.ts:10:14 - error TS2742: The inferred type of 'notesAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

10 export const notesAllView = (
                ~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/opportunities-all.view.ts:8:14 - error TS2742: The inferred type of 'opportunitiesAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

8 export const opportunitiesAllView = (
               ~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/opportunity-by-stage.view.ts:8:14 - error TS2742: The inferred type of 'opportunitiesByStageView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

8 export const opportunitiesByStageView = (
               ~~~~~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/opportunity-table-by-stage.view.ts:8:14 - error TS2742: The inferred type of 'opportunitiesTableByStageView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

8 export const opportunitiesTableByStageView = (
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/people-all.view.ts:11:14 - error TS2742: The inferred type of 'peopleAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

11 export const peopleAllView = (
                ~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-all.view.ts:10:14 - error TS2742: The inferred type of 'tasksAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

10 export const tasksAllView = (
                ~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-assigned-to-me.ts:11:14 - error TS2742: The inferred type of 'tasksAssignedToMeView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

11 export const tasksAssignedToMeView = (
                ~~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-by-status.view.ts:10:14 - error TS2742: The inferred type of 'tasksByStatusView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

10 export const tasksByStatusView = (
                ~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/workflow-runs-all.view.ts:8:14 - error TS2742: The inferred type of 'workflowRunsAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

8 export const workflowRunsAllView = (
               ~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/workflow-versions-all.view.ts:11:14 - error TS2742: The inferred type of 'workflowVersionsAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

11 export const workflowVersionsAllView = (
                ~~~~~~~~~~~~~~~~~~~~~~~

src/engine/workspace-manager/standard-objects-prefill-data/views/workflows-all.view.ts:11:14 - error TS2742: The inferred type of 'workflowsAllView' cannot be named without a reference to '../../../../../../../node_modules/@lingui/core/dist/index.d.cts'. This is likely not portable. A type annotation is necessary.

11 export const workflowsAllView = (
                ~~~~~~~~~~~~~~~~

src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts:14:28 - error TS2503: Cannot find namespace 'DOMPurify'.

14   private readonly purify: DOMPurify.DOMPurify;
                              ~~~~~~~~~

src/utils/image.ts:27:18 - error TS2749: 'Axios' refers to a value, but is being used as a type here. Did you mean 'typeof Axios'?

27   axiosInstance: Axios,
                    ~~~~~

test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts:1:23 - error TS2307: Cannot find module '@faker-js/faker/.' or its corresponding type declarations.

1 import { faker } from '@faker-js/faker/.';
                        ~~~~~~~~~~~~~~~~~~~


Found 25 errors in 22 files.

Errors  Files
     1  src/engine/api/mcp/services/tools/mcp-metadata-tools.service.ts:90
     4  src/engine/api/rest/metadata/rest-api-metadata.service.ts:24
     1  src/engine/metadata-modules/constants/search-vector-field.constants.ts:3
     1  src/engine/twenty-orm/relation-nested-queries/utils/formatConnectRecordNotFoundErrorMessage.util.ts:5
     1  src/engine/workspace-manager/standard-objects-prefill-data/prefill-views.ts:21
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/companies-all.view.ts:12
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/custom-all.view.ts:6
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/dashboards-all.view.ts:11
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/notes-all.view.ts:10
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/opportunities-all.view.ts:8
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/opportunity-by-stage.view.ts:8
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/opportunity-table-by-stage.view.ts:8
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/people-all.view.ts:11
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-all.view.ts:10
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-assigned-to-me.ts:11
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/tasks-by-status.view.ts:10
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/workflow-runs-all.view.ts:8
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/workflow-versions-all.view.ts:11
     1  src/engine/workspace-manager/standard-objects-prefill-data/views/workflows-all.view.ts:11
     1  src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts:14
     1  src/utils/image.ts:27
     1  test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts:1

neo@neos-MacBook-Pro twenty-server % 
```

chase:
> We have the same issue in our team that's stopping us from adopting `tsgo`
> 
> So I compiled your branch and ran a test, it doesn't seem to make any difference in our code base. Just wanted to let you know.

Interesting. This is what I get in Linux after setting up your repo and replacing `tsc` in `node_modules/.bin` so that nx uses it instead:

```
src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts(14,28): error TS2503: Cannot find namespace 'DOMPurify'.
src/utils/image.ts(27,18): error TS2749: 'Axios' refers to a value, but is being used as a type here. Did you mean 'typeof Axios'?
test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts(1,23): error TS2307: Cannot find module '@faker-js/faker/.' or its corresponding type declarations.
```

Seems like this might need some fixes for macOS.

chase:
@neo773 should be fixed with the latest commit, the above errors seem like they might be actual errors or at least ones that can be fixed trivially:
<details><summary>Patch</summary>
<p>


``` diff 
diff --git a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts
index f81f151..335fe77 100644
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts
@@ -8,10 +8,12 @@ import * as planer from 'planer';
 
 import { safeDecodeURIComponent } from 'src/modules/messaging/message-import-manager/drivers/imap/utils/safe-decode-uri-component.util';
 
+type DOMPurifyInstance = ReturnType<typeof DOMPurify>;
+
 @Injectable()
 export class ImapMessageTextExtractorService {
   private readonly jsdomInstance: JSDOM;
-  private readonly purify: DOMPurify.DOMPurify;
+  private readonly purify: DOMPurifyInstance;
 
   constructor() {
     this.jsdomInstance = new JSDOM('');
diff --git a/packages/twenty-server/src/utils/image.ts b/packages/twenty-server/src/utils/image.ts
index 8e98729..ad78167 100644
--- a/packages/twenty-server/src/utils/image.ts
+++ b/packages/twenty-server/src/utils/image.ts
@@ -1,4 +1,4 @@
-import { type Axios } from 'axios';
+import { type AxiosInstance } from 'axios';
 
 const cropRegex = /([w|h])([0-9]+)/;
 
@@ -24,7 +24,7 @@ export const getCropSize = (value: ShortCropSize): CropSize | null => {
 
 export const getImageBufferFromUrl = async (
   url: string,
-  axiosInstance: Axios,
+  axiosInstance: AxiosInstance,
 ): Promise<Buffer> => {
   const response = await axiosInstance.get(url, {
     responseType: 'arraybuffer',
diff --git a/packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts b/packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts
index 0aa4bb7..2981c77 100644
--- a/packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts
+++ b/packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/failing-field-metadata-morph-relation-creation.integration-spec.ts
@@ -1,4 +1,4 @@
-import { faker } from '@faker-js/faker/.';
+import { faker } from '@faker-js/faker';
 import { createOneFieldMetadata } from 'test/integration/metadata/suites/field-metadata/utils/create-one-field-metadata.util';
 import { createOneObjectMetadata } from 'test/integration/metadata/suites/object-metadata/utils/create-one-object-metadata.util';
 import { deleteOneObjectMetadata } from 'test/integration/metadata/suites/object-metadata/utils/delete-one-object-metadata.util';
```

</p>
</details>

neo773:
@chase 
With the new commit it finally works, as for the remaining type errors they don't seem to affect `Strada` version, but it doesn't matters.

Thank you for your work on this.

tmm1:
Thanks for working on this!

Fixes https://github.com/microsoft/typescript-go/issues/1034#issuecomment-3041199644 and #1657 also

internal/ls/autoimports.go
```diff
@@ -164,7 +164,8 @@ func (e *exportInfoMap) add(
 	}
 
 	moduleName := stringutil.StripQuotes(moduleSymbol.Name)
-	id := e.exportInfoId + 1
+	id := e.exportInfoId
+	e.exportInfoId += 1
```
andrewbranch: Heh, I just found this too 😄

AlCalzone:
Uhh, this is great! It seems that this fixes my issue https://github.com/microsoft/typescript-go/issues/1347 too.

@chase could you include the compiler test case I posted in that issue?

internal/checker/nodebuilderimpl.go
```diff
@@ -568,6 +568,12 @@ func (b *nodeBuilderImpl) symbolToTypeNode(symbol *ast.Symbol, mask ast.SymbolFl
 			}
 
 			if attributes == nil {
+				// This avoids TS2742 errors for internal library types when skipLibCheck is enabled.
+				// TODO: Remove this workaround once syntacticNodeBuilder is implemented. The root cause is that
+				// tsgo deeply infers types while upstream reuses existing type nodes, avoiding internal library types.
+				if b.ch.compilerOptions.SkipLibCheck.IsTrue() && targetFile != nil && targetFile.IsDeclarationFile {
```
jakebailey: This workaround seems pretty arbitrary; I'm not sure it's good even as a temporary fix, but I haven't looked into the exact situation you're trying to work around here. Is there a test for this?

chase: If I have time this weekend, I'll try to write this up. What's the right way to add test cases for this and the one in #1347 that @AlCalzone mentioned?

jakebailey: You'd want to create a compiler test in `testdata/tests/cases/compiler`. Symlinks are supported (see `_submodules/TypeScript/tests/cases/compiler/moduleResolutionWithSymlinks_notInNodeModules.ts` for an example).

AlCalzone: The code in my issue should be able to be used 1:1 in such a compiler test. At least it worked when I tried my hand at this issue and failed.

iisaduan: You can either use `f.BaselineAutoImport()` (ex. `internal/fourslash/tests/autoImportCompletion_test.go`) or `f.VerifyApplyCodeActionFromCompletion()` (ex. `internal/fourslash/tests/gen/completionsUniqueSymbol_import_test.go`)

chase: @jakebailey Here's a quick summary of the bug, which occurs in a real code base.

There is a pnpm workspace with: `/workspace/x/y/z/src/components/Button/Button.tsx`

Library: `@base-ui-components/react` is located in the `.pnpm` store with `exports` field blocking internal paths, this is symlinked to `/workspace/node_modules`

The `tsconfig` has `skipLibCheck: true`, the library at `/workspace/x/y/z` has types built using `tsgo --build --emitDeclarationOnly`

This is what occurs in the compiler, without the workaround:

1. Declaration emitter processes `Button.tsx` with `export const Button = forwardRef<T, P>(...)`
2. `serializeTypeForDeclaration()` calls `typeToTypeNode()` to infer the return type, since we're missing `syntacticNodeBuilder`
3. `tsgo` analyzes the function body, sees `<Tooltip>` component used conditionally
4. Serializes `Tooltip`'s type, encounters internal `useAnchorPositioning` from `@base-ui-components/react`
5. `symbolToTypeNode()` tries to generate a module specifier via `getSpecifierForModuleSymbol()`
6. `tryGetModuleNameAsNodeModule()` gets blocked by the package's `exports` field and returns empty
7. This causes it to fall back to `getLocalModuleSpecifier()`, generating `../../../../../../node_modules/@base-ui-components/...`
8. It detects `/node_modules/` in specifier, reports TS2742 error

Upstream, tsc, uses `syntacticNodeBuilder.serializeTypeOfDeclaration()` to reuse `ForwardRefExoticComponent<...>` without diving into function body and never encounters internal types.

The intention of this workaround:

When `skipLibCheck: true` **AND** can't generate clean specifier for library declaration file, then return `nil` to elide reference instead of erroring.

This aligns with skipLibCheck semantics in that the user explicitly opts out of library type errors.

---

I've added a compiler test case that demonstrates that without the workaround the compiler incorrectly fails with:

```
error TS2742: The inferred type of 'createComponent' cannot be named without a reference to '../node_modules/.pnpm/@base-lib+react@1.0.0/node_modules/@base-lib/react/esm/Tooltip'
error TS2742: ... reference to '../node_modules/.pnpm/@base-lib+react@1.0.0/node_modules/@base-lib/react/esm/utils/internal'
```

jakebailey: I understand the bug, but `skipLibCheck` is an option that determines which files have the checker run over them. Note how it's only used in `SkipTypeChecking`. It is not supposed to be used for anything else.

There was a time before the syntactic node builder, and this hack was not required. That's what confuses me.

chase: I suppose maybe since I don't know the internals as well as you, I don't know where the correct place is to fix this, but nonetheless this behavior doesn't occur in tsc. I referenced how things were done in tsc and enough was missing or has changed that instead I addressed it where the problem terminated from my perspective.

From my perspective, as a user this works as expected: these files, which are library files, should not be type checked. They were being type checked and causing an error that is not present in upstream.

Maybe there is a better point where this should have been avoided, before module paths are even resolved? Can you point me to a better place to solve this discrepancy?

chase: If this is a major blocker to this PR, I can revert this workaround and hand it over to the experts to fix properly, now that my test case has been added.

andrewbranch: > these files, which are library files, should not be type checked. They were being type checked and causing an error that is not present in upstream

The error is issued because we are failing to _generate_ valid declaration files from your own `.tsx` file. That error is _never_ silenceable. `--skipLibCheck` is for silencing errors that occur in _existing_ declaration files. If there’s missing functionality that causes us not to be able to synthesize a name for a type in a declaration file, we _must_ issue the “we couldn’t name this type” error, or else folks will accidentally publish completely broken .d.ts files to npm.

chase: @andrewbranch To clarify, these are not errors in my own `.tsx` file, the types are valid and the code is incredibly simple. These are not errors that occur in `tsc`, so unless things are getting more strict in ways that don't match `tsc`, I don't think this is correct.

These are errors in an existing type declaration file for a third party library installed with pnpm, which as far as I'm aware, are exactly what `skipLibCheck` is meant to skip.

Specifically, the code where this errors incorrectly is very similar to this:
``` tsx

import { forwardRef } from 'react';
import { Tooltip } from '@base-ui-components/react/tooltip';

export const Button = forwardRef<HTMLButtonElement, {tooltip: string}>(
function Component({tooltip}: {tooltip: string}, ref) {
  return tooltip ? <Tooltip content={tooltip} ... /> : <button  ref={ref} ... />
}
);
```

https://github.com/mui/base-ui/blob/master/packages/react/src/tooltip/index.ts
-> https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/packages/react/src/utils/useAnchorPositioning.ts#L104
-> https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/packages/react/src/utils/useAnchorPositioning.ts#L684-L688
-> https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/packages/react/src/utils/useAnchorPositioning.ts#L669-L682

Types:
`package.json` which does not export `useAnchorPositioning`:  
https://unpkg.com/@base-ui-components/react@1.0.0-beta.4/package.json

chase: I think I need to rewrite the compiler case, since it doesn't accurately cover the case above.

jakebailey: Declaration diagnostics should never be calculated for a `d.ts` file (it's already a declaration), and even if so (which would be a bug), declaration files do not contain incomplete types that would require inferring anything. That's why I am personally so confused by the above explanation.

chase: Okay, suppose you ignore my explanation (which is likely established upon multiple layers of misunderstandings, since I'm new to this code base and Golang in general): why would `tsgo` check this type and emit this error when `tsc` does not?

This is `tsgo --build --emitDeclarationOnly`'s output:

```
error TS2742: The inferred type of 'Button' cannot be named without a reference to '../../../../../../node_modules/@base-ui-components/react/esm/utils/useAnchorPositioning'. This is likely not portable. A type annotation is necessary.

export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
```

 It seems to be traversing and checking things it should not, following a path it does not follow in `tsc`.
 
 Specifically, it is checking a type for `useAnchorPositioning` which is not exported by `package.json` for `@base-ui-components/react`, which is used in `Tooltip` from that library, which is used in a component library in our pnpm workspace.

jakebailey: What file is that error attributed to? And in what project are you running this?

jakebailey: From my understanding, that is in the file that contains the snippet from https://github.com/microsoft/typescript-go/pull/1902#discussion_r2461628121, in which case it's user code, and the symlink cache stuff is supposed to be fixing this problem because the problem is that for _that snippet_, it can't seem to name the type in mui.

andrewbranch: To make sure we’re understanding the problem correctly, do you have a real-world repro you can publish somewhere?

chase: This error is for the file similar to the snippet above, which is proprietary code. This is in a pnpm workspace, wherein this Button component is exported as part of an internal library for the workspace.

This code is valid in `tsc`. The code that fails is a type that's not used in the result type of `Tooltip`, the component used. It is used by TooltipPositioner: https://github.com/mui/base-ui/blob/e9f98ed126597668ace576d45904b31bad13b17c/packages/react/src/tooltip/positioner/TooltipPositioner.tsx#L5
Which is exported in the barrel file: https://github.com/mui/base-ui/blob/master/packages/react/src/tooltip/index.parts.ts#L4
But the package.json does not export the utility's types, and block access by: `"./esm": null`

So, where `tsgo` seems to respect the `exports` field for deeply inferred types, `tsc` doesn't seem to follow that far.

What I inferred was this area of code, dealing with module resolution styles (including respecting `exports`) had a bug which leads to an erroneous fallback that skips over the correctly resolved symlinks for the library: https://github.com/chase/typescript-go/blob/main/internal/checker/nodebuilderimpl.go#L529-L596

andrewbranch: What module/moduleResolution are you using in your project? tsgo always respects `exports` because every mode that doesn’t respect them is deprecated and not ported.

chase: > To make sure we’re understanding the problem correctly, do you have a real-world repro you can publish somewhere?

Sorry, somehow missed your replies! Didn't mean to talk past you.

Unfortunately, I'm not permitted to provide anything like that without permission. As an alternative, I'm trying to make a compiler case for it, but unfortunately they all pass even without the workaround 😂

> What module/moduleResolution are you using in your project? tsgo always respects `exports` because every mode that doesn’t respect them is deprecated and not ported.

We use `"moduleResolution": "Bundler"` and  `"module": "ESNext"`.

chase: I tried making a minimal reproduction as a separate repo, no error, so there's likely some complex interaction in the real codebase that causes it to misbehave.

Might revisit this later, for now the workaround has been reverted.

internal/compiler/program.go
```diff
@@ -1630,6 +1640,50 @@ func (p *Program) SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmi
 	return sourceFileMayBeEmitted(sourceFile, p, forceDtsEmit)
 }
 
+func (p *Program) GetSymlinkCache() *symlinks.KnownSymlinks {
+	// if p.Host().GetSymlinkCache() != nil {
+	// 	return p.Host().GetSymlinkCache()
+	// }
+	if p.knownSymlinks == nil {
```
andrewbranch: This condition looks to be impossible as-written, but I think lazy initialization is a good idea. However, you need to guard the field initialization with a `sync.Once` like the other lazy computed caches.

internal/compiler/program.go
```diff
@@ -1630,6 +1640,50 @@ func (p *Program) SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmi
...
+		p.knownSymlinks = symlinks.NewKnownSymlink(p.GetCurrentDirectory(), p.UseCaseSensitiveFileNames())
+		// In declaration-only builds, the symlink cache might not be populated yet
+		// because module resolution was skipped. Populate it now if we have resolutions.
```
andrewbranch: I don’t recall exactly how this happened in Strada, but I don’t think this comment applies in Corsa.

internal/modulespecifiers/specifiers.go
```diff
@@ -163,38 +164,117 @@ func getAllModulePaths(
...
+	// Mark this package as being processed to avoid redundant work
+	links.MarkPackageAsPopulated(packageJsonPath)
+
+	cwd := host.GetCurrentDirectory()
+	caseSensitive := host.UseCaseSensitiveFileNames()
+
+	// Helper to resolve dependencies without creating intermediate slices
+	resolveDeps := func(deps map[string]string) {
+		for depName := range deps {
+			resolved := host.ResolveModuleName(depName, packageJsonPath, options.OverrideImportMode)
```
andrewbranch: I’d like to get this out of here by frontloading these resolutions to program construction. That will allow us to remove `ResolveModuleName` from the host interfaces and should also remove all concurrent writes from the symlink cache. In moving it there, I also think the program could do a better job of determining which of these resolutions are necessary and avoid doing redundant ones. The original comment that said most of these resolutions will already be cached is not currently true in Corsa.

It is likely the case that we get 90% of the way there without this block of logic, as this was added to Strada long after the symlink cache existed with only the resolutions it made during program construction. Dropping it from this PR and leaving it as a to-do would be fine. The priority is to eliminate file system reads and cache mutations post-program-load.

internal/compiler/program.go
chase: This occurs regularly in pnpm workspaces when running `tsgo --build --emitDeclarationsOnly`, if I recall correctly.

internal/modulespecifiers/specifiers.go
chase: I'm worried that since the current test suite don't adequately cover handling pnpm workspaces, that removing this and relocating some of this willl cause a regression for some of the cases that were solved for others.

I might have time to work on this over the weekend, but if your or another wants to pick up this branch and get it up to standard, I don't want to to block getting a fix merged.

chase:
Unfortunately I won't be able to continue working on this for the foreseeable future due to some unforeseen life events. I'm not sure if this should be closed, but I'd be happy if someone else could pick this up or build from this while I'm away from keyboard.

neo773:
@andrewbranch Just wanted to know what the core team's stance/priority is on this one? It's a huge blocker for mono-repos.

andrewbranch:
I will try to pick this up next week, unless @iisaduan gets to it first.

internal/compiler/program.go
```diff
@@ -1651,10 +1654,46 @@ func (p *Program) GetSymlinkCache() *symlinks.KnownSymlinks {
 		if len(p.resolvedModules) > 0 || len(p.typeResolutionsInFile) > 0 {
 			p.knownSymlinks.SetSymlinksFromResolutions(p.ForEachResolvedModule, p.ForEachResolvedTypeReferenceDirective)
 		}
+		p.populateSymlinkCacheFromResolutions()
 	}
 	return p.knownSymlinks
 }
 
+func (p *Program) populateSymlinkCacheFromResolutions() {
+	p.Host().FS().WalkDir(
```
andrewbranch: I don’t think we should do a WalkDir here. Basically, I think we should find the package.json for each non-declaration input file, and then resolve the dependencies for each unique package.json. Finding the package.json could be moved to file loading to avoid a separate loop over all the files (and in fact, this already happens in order to determine the module kind under certain settings/situations; with this we could make it unconditional).

internal/compiler/program.go
```diff
@@ -1651,10 +1654,46 @@ func (p *Program) GetSymlinkCache() *symlinks.KnownSymlinks {
...
+			}
+			pkgJsonInfo := p.GetPackageJsonInfo(packageJsonPath)
+			if pkgJsonInfo == nil {
+				return nil
+			}
+			pkgJson := pkgJsonInfo.GetContents()
+			if pkgJson == nil {
+				return nil
+			}
+			p.knownSymlinks.PopulateFromResolutions(packageJsonPath, pkgJson, p.ResolveModuleName)
```
andrewbranch: There are two ways to save some work here:
- For a lot of these dependencies, we probably already did a resolution of that package as part of normal program construction, which means we already found the same symlink by iterating over the program's resolved modules. If fileLoader were to keep a set of package names it resolved, we could skip anything that's already been done.
- For resolutions we still need to make, we could save a few file system hits by exposing a resolver function that just gets the dependency’s package.json instead of its main entrypoint.

andrewbranch: I also think the naming of this (both the function we're in and the function being called here) should say something about dependencies

internal/compiler/program.go
```diff
@@ -240,6 +248,11 @@ func (p *Program) UpdateProgram(changedFilePath tspath.Path, newHost CompilerHos
 	result.filesByPath = maps.Clone(result.filesByPath)
 	result.filesByPath[newFile.Path()] = newFile
 	updateFileIncludeProcessor(result)
+	result.knownSymlinks = symlinks.NewKnownSymlink(result.GetCurrentDirectory(), result.UseCaseSensitiveFileNames())
+	if len(result.resolvedModules) > 0 || len(result.typeResolutionsInFile) > 0 {
+		result.knownSymlinks.SetSymlinksFromResolutions(result.ForEachResolvedModule, result.ForEachResolvedTypeReferenceDirective)
+	}
+	p.populateSymlinkCacheFromResolutions()
```
andrewbranch: I think we should just copy the symlink cache on program clone. I think it's intended to be immutable after initialization, so by copy, I mean the new program can share the same reference. (If the cache were mutable, we would have to add a clone method on it.) This function needs to be dirt cheap, and calling it means you're asserting that the _only_ relevant change that's happened in the Program's whole world is to the file you're passing in.

jakebailey:
This should be marked as fixing #1034, yes?

andrewbranch:
About to push an update; here’s my response to my own suggestions:

> Basically, I think we should find the package.json for each non-declaration input file, and then resolve the dependencies for each unique package.json. Finding the package.json could be moved to file loading to avoid a separate loop over all the files (and in fact, this already happens in order to determine the module kind under certain settings/situations; with this we could make it unconditional).

Done, and checked that this has no measurable performance impact by switching `microsoft/vscode` to `--module esnext` and comparing parse time between nightly and this PR’s build.

> For a lot of these dependencies, we probably already did a resolution of that package as part of normal program construction, which means we already found the same symlink by iterating over the program's resolved modules. If fileLoader were to keep a set of package names it resolved, we could skip anything that's already been done.

It’s not quite that simple, because the resolution of a package name can vary based on the location we’re starting from. For my original suggestion to work, we’d have to store not only the package name, but also the requesting location and the resolved location, and then do some calculation based on those later to figure out whether that resolution holds for the package.json location we’re interested in. That didn’t seem worth either the code complexity or allocations. However, we can do almost as well for free by checking to see whether the symlink cache already has an entry for a given dependency in the node_modules directory adjacent to the package.json. If so, we can skip the resolution and realpath call, because the resolution is guaranteed to resolve to that first node_modules directory if it exists.

> That will allow us to remove `ResolveModuleName` from the host interfaces and should also remove all concurrent writes from the symlink cache.

@iisaduan did this, so I started to remove the SyncMaps and SyncSets, but then I noticed that `KnownSymlinks` is still used concurrently by `projectReferenceDtsFakingHost`. However, I only noticed just now while writing this that it’s not sharing Program’s `KnownSymlinks`, so Program does indeed write to it fully synchronously. Maybe a future improvement could be to split those, so Program doesn’t have to pay the cost of using SyncMaps and doesn’t expose a mutable API to the module specifier generator. Not a big deal for now. The important thing for LSP purposes was that module specifier generation during diagnostics/auto-import doesn’t try to access the file system unexpectedly, which is fixed.

testdata/baselines/reference/submodule/compiler/declarationEmitReexportedSymlinkReference3.js.diff
```diff
@@ -6,6 +6,6 @@
 
 +//// [keys.d.ts]
 +import { MetadataAccessor } from "@raymondfeng/pkg2";
-+export declare const ADMIN: MetadataAccessor<boolean, import("../../pkg1/dist").IdType>;
++export declare const ADMIN: any;
```
jakebailey, line 9: I was going to ask why this regressed, but this file shouldn't be emitted anyhow

andrewbranch: Weird...

jakebailey:
I think I expected the PR to be fixing a bunch of other symlink related issues in the fourslash tests, so I'm not sure what else we are missing. Maybe it's the stuff @johnfav03 has commented out in #1793?

andrewbranch:
My auto-imports rewrite is not going to go through the same code path, so I’m not sure if it matters much right now. Can you give me an example of a test you’re thinking of?

jakebailey:
https://github.com/microsoft/typescript-go/blob/bd7c18dc8ed3c5ed960d72f1e329353f0a594bcc/internal/fourslash/tests/gen/autoImportCrossProject_symlinks_toSrc_test.go or similar, the tests I skipped when I added auto import fixes

andrewbranch:
@jakebailey all those also rely on the package.json auto import provider, so they never even try to pull on the symlink cache. Those are blocked on my auto-imports rewrite (which is going to have to deal with symlinks separately)

andrewbranch:
Thank you for getting this started, @shinichy and @chase!

jakebailey:
Hm, this ended up failing all of the tests.

andrewbranch:
Oops, I committed the removal of `t.Skip` on the test I was debugging

jakebailey:
Seems like three tests are being fixed:

```
 internal/fourslash/_scripts/failingTests.txt                        | 3 ---
 internal/fourslash/tests/gen/autoImportPnpm_test.go                 | 2 +-
 internal/fourslash/tests/gen/autoImportSymlinkCaseSensitive_test.go | 2 +-
 internal/fourslash/tests/gen/importNameCodeFix_pnpm1_test.go        | 2 +-
```

And linting needs to be fixed.

chase:
> Thank you for getting this started, @shinichy and @chase!

Thank you @andrewbranch and @iisaduan for bringing this over the finish line while I get my ducks back in a row.

nnnnoel:
Really appreciate you all guys!! It would be wonderful innovation for pnpm ecosystem 🫶

PR #1904, opened 2025-10-18
Consistently error on full circle of circular import aliases

### LINKED ISSUE #1851: Circular alias error locations differ based on number of checkers
Discovered in https://github.com/microsoft/typescript-go/pull/1797#issuecomment-3379139315 but silenced for now. @ahejlsberg mentioned that we may be able to reuse `pushTypeResolution` for this purpose.

ahejlsberg:
With this PR we consistently error on all nodes in the full circle of circular import alias declarations. This in particular means that our error reporting is unaffected by the number of checkers used to check a program.

Fixes #1851.

### DIFF
--- a/internal/testrunner/compiler_runner.go
+++ b/internal/testrunner/compiler_runner.go
@@ -337,11 +337,6 @@ func newCompilerTest(
 }
 
 var concurrentSkippedErrorBaselines = map[string]string{
-	"circular1.ts": "Circular error reported in an extra position.",
-	"circular3.ts": "Circular error reported in an extra position.",
-	"recursiveExportAssignmentAndFindAliasedType1.ts": "Circular error reported in an extra position.",
-	"recursiveExportAssignmentAndFindAliasedType2.ts": "Circular error reported in an extra position.",
-	"recursiveExportAssignmentAndFindAliasedType3.ts": "Circular error reported in an extra position.",
 	"typeOnlyMerge2.ts": "Type-only merging is not detected when files are checked on different checkers.",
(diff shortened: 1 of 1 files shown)


jakebailey:
Can some of these be unskipped, then? https://github.com/microsoft/typescript-go/blob/8c62e06ecea9786eec1132705ceee8e7da94e9b0/internal/testrunner/compiler_runner.go#L342

ahejlsberg:
> Can some of these be unskipped, then?

Yes, all but two of them!

internal/checker/checker.go
```diff
@@ -15652,29 +15633,25 @@ func (c *Checker) resolveAlias(symbol *ast.Symbol) *ast.Symbol {
...
-	} else if links.aliasTarget == c.resolvingSymbol {
-		links.aliasTarget = c.unknownSymbol
 	}
 	return links.aliasTarget
 }
 
 func (c *Checker) tryResolveAlias(symbol *ast.Symbol) *ast.Symbol {
 	links := c.aliasSymbolLinks.Get(symbol)
-	if links.aliasTarget != c.resolvingSymbol {
+	if links.aliasTarget != nil || c.findResolutionCycleStartIndex(symbol, TypeSystemPropertyNameAliasTarget) < 0 {
```
jakebailey, line 15654: Hopefully this extra `find` isn't too expensive, given how often we resolve aliases.

ahejlsberg: `tryResolveAlias` is only called when generating spelling correction suggestions, so not an issue at all. And anyway, every call to `pushTypeResolution` has a call to that `find` method with no issues.

PR #1911, opened 2025-10-20
Forbid platform specific package uses in agnostic files

jakebailey:
Mainly for external PRs which don't realize that our paths are usually incompatible with the stdlib packages (especially `filepath`) outside certain circumstances.

Same goes for `os`, `runtime.GOOS`.

### DIFF
--- a/cmd/tsgo/lsp.go
+++ b/cmd/tsgo/lsp.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
 	"os/signal"
 	"runtime"
 	"syscall"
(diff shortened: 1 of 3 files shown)


cmd/tsgo/lsp.go
```diff
@@ -49,9 +53,17 @@ func runLSP(args []string) int {
 		FS:                 fs,
 		DefaultLibraryPath: defaultLibraryPath,
 		TypingsLocation:    typingsLocation,
+		NpmInstall: func(cwd string, args []string) ([]byte, error) {
+			cmd := exec.Command("npm", args...)
+			cmd.Dir = cwd
```
jakebailey, line 58: A bit of a hack but I prefer this greatly to there being any chance that tests run `npm install`.

sheetalkamat (approved):

PR #1935, opened 2025-10-22
Create clickable links in quick info from @link JSDoc tags

### LINKED ISSUE #1895: Add support for JSDoc links in editors
## Feature request

Support rendering the JSDoc `{@link ...}` and `{@linkcode }` syntax. These links should be rendered in hovers and suggestion details

[Here's the VS Code side of the implementation](https://github.com/microsoft/vscode/blob/af6adc61cebb854d26c9b0e065df66f6f560080f/extensions/typescript-language-features/src/languageFeatures/util/textRendering.ts#L138) where we convert `SymbolDisplayPart` to a markdown string for use in the editor

We use these links heavily in our typings for extension authors and have found them to be very useful

ahejlsberg:
Fixes #1895.

### DIFF
--- a/internal/ls/hover.go
+++ b/internal/ls/hover.go
@@ -455,7 +455,7 @@ func (l *LanguageService) writeJSDocLink(b *strings.Builder, c *checker.Checker,
 		if linkText == "" {
 			linkText = getEntityNameString(name) + text[:prefixLen]
 		}
-		linkUri := fmt.Sprintf("%v#%v,%v-%v,%v", loc.Uri, loc.Range.Start.Line+1, loc.Range.Start.Character+1, loc.Range.End.Line+1, loc.Range.End.Character+1)
+		linkUri := fmt.Sprintf("%s#%d,%d-%d,%d", loc.Uri, loc.Range.Start.Line+1, loc.Range.Start.Character+1, loc.Range.End.Line+1, loc.Range.End.Character+1)
 		writeMarkdownLink(b, linkText, linkUri, quote)
 		return
 	}


internal/ls/hover.go
```diff
@@ -411,61 +412,80 @@ func writeCode(b *strings.Builder, lang string, code string) {
...
+	declarations := getDeclarationsFromLocation(c, name)
+	if len(declarations) != 0 {
+		declaration := declarations[0]
+		file := ast.GetSourceFileOfNode(declaration)
+		pos := l.converters.PositionToLineAndCharacter(file, core.TextPos(scanner.GetTokenPosOfNode(declaration, file, false /*includeJSDoc*/)))
+		linkText := strings.Trim(text, " ")
+		if linkText == "" {
+			linkText = getEntityNameString(name)
 		}
+		writeMarkdownLink(b, linkText, fmt.Sprintf("%v#%v,%v", string(FileNameToDocumentURI(file.FileName())), pos.Line+1, pos.Character+1), quote)
```
jakebailey: Can we do a range here instead of just a position? IIRC `#L,C-L-C` should work.

jakebailey: Basically, I mean reuse the logic we use for go-to-def to choose the range from a node and then return that.

ahejlsberg: Now implemented.

internal/ls/hover.go
```diff
@@ -411,61 +411,90 @@ func writeCode(b *strings.Builder, lang string, code string) {
...
+		declaration := declarations[0]
+		file := ast.GetSourceFileOfNode(declaration)
+		node := core.OrElse(ast.GetNameOfDeclaration(declaration), declaration)
+		loc := l.getMappedLocation(file.FileName(), createRangeFromNode(node, file))
+		prefixLen := core.IfElse(strings.HasPrefix(text, "()"), 2, 0)
+		linkText := trimCommentPrefix(text[prefixLen:])
+		if linkText == "" {
+			linkText = getEntityNameString(name) + text[:prefixLen]
+		}
+		linkUri := fmt.Sprintf("%v#%v,%v-%v,%v", loc.Uri, loc.Range.Start.Line+1, loc.Range.Start.Character+1, loc.Range.End.Line+1, loc.Range.End.Character+1)
```
jakebailey:
```suggestion
		linkUri := fmt.Sprintf("%s#%d,%d-%d,%d", loc.Uri, loc.Range.Start.Line+1, loc.Range.Start.Character+1, loc.Range.End.Line+1, loc.Range.End.Character+1)
```

Nit, but, `%v` means deferring to `String()` when defined, so best to not use it when we know what format we want.

ahejlsberg: Done.

PR #1937, opened 2025-10-22
Plumb through TokenFlagsSingleQuote; use for auto import quote detection

### LINKED ISSUE #1697: Respect quoteStyle when adding imports
For auto imports, we'd like the `quoteStyle` preferences in VS Code to be respected. In the VS code codebase we use single quotes for everything but auto imports always uses double quotes

andrewbranch:
Fixes #1697


internal/scanner/scanner.go
```diff
@@ -1455,6 +1455,9 @@ func (s *Scanner) scanIdentifierParts() string {
 
 func (s *Scanner) scanString(jsxAttributeString bool) string {
 	quote := s.char()
+	if quote == '\'' {
+		s.tokenFlags |= ast.TokenFlagsSingleQuote
```
jakebailey, line 1459: Oops...

andrewbranch: This wasn't a porting mistake; in Strada we didn’t have this token flag, and we (for some reason unknown to me) intentionally didn't set the single-quote boolean on StringLiteral from the parser. It only existed for synthesized nodes, which means we had to check the source text on parse nodes and the property on synthesized nodes. The way we're doing it now seems like a better design all around. 🤷

jakebailey: Ah, gotcha.

jakebailey:
Darn, conflicts with the other LS settings PR.

jakebailey (approved):

iisaduan (approved):

PR #1939, opened 2025-10-23
Implement selection ranges

### LINKED ISSUE #1896: Add support for selection range requests
Tracks adding [selection range support](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) for editors. This is the feature that powers `expand selection` in vscode

It's useful during editing and can be used by tooling to expand to enclosing scopes

jakebailey:
Fixes #1896

This is not a pure port; the old code relied heavily on `SyntaxList`. I tried doing it with the children helpers we have in the new code, but it always seemed really off.

Instead, I am taking a strategy similar to `astnav` and instead just building the stack of ranges from node visiting directly. This seems to do a very good job. Some differences remain.

### DIFF
--- a/internal/fourslash/_scripts/convertFourslash.mts
+++ b/internal/fourslash/_scripts/convertFourslash.mts
@@ -1427,7 +1427,7 @@ function parseBaselineSignatureHelp(args: ts.NodeArray<ts.Expression>): Cmd {
 function parseBaselineSmartSelection(args: ts.NodeArray<ts.Expression>): Cmd {
     if (args.length !== 0) {
         // All calls are currently empty!
-        throw new Error("Expected no arguments in verify.baselineSignatureHelp");
+        throw new Error("Expected no arguments in verify.baselineSmartSelection");
     }
     return {
         kind: "verifyBaselineSmartSelection",
@@ -1789,7 +1789,6 @@ function generateCmd(cmd: Cmd): string {
             return `f.VerifyRenameSucceeded(t, ${cmd.preferences})`;
         case "renameInfoFailed":
             return `f.VerifyRenameFailed(t, ${cmd.preferences})`;
-
         default:
             let neverCommand: never = cmd;
             throw new Error(`Unknown command kind: ${neverCommand as Cmd["kind"]}`);


testdata/baselines/reference/submodule/fourslash/smartSelection/smartSelection_behindCaret.baseline.diff
```diff
@@ -0,0 +1,8 @@
+--- old.smartSelection_behindCaret.baseline
++++ new.smartSelection_behindCaret.baseline
+@@= skipped -0, +0 lines =@@
+ // === Smart Selection ===
+ let/**/ x: string
+
+-let
```
jakebailey, line 7: The visitor does not touch this token, so we don't get it.

testdata/baselines/reference/submodule/fourslash/smartSelection/smartSelection_bindingPatterns.baseline.diff
```diff
@@ -0,0 +1,10 @@
+--- old.smartSelection_bindingPatterns.baseline
++++ new.smartSelection_bindingPatterns.baseline
+@@= skipped -20, +20 lines =@@
+ const { x, y: a, .../**/zs = {} } = {};
+
+                     zs             
+-                 ...zs
```
jakebailey, line 7: This also differs because we don't walk the token itself.

testdata/baselines/reference/submodule/fourslash/smartSelection/smartSelection_functionParams2.baseline.diff
```diff
@@ -0,0 +1,21 @@
...
+-
+-
+-           ↲
+-••a,
+-  b↲
+-
++  b↲
++
++  a,
++  b↲
```
jakebailey, line 18: This I think is some sort of difference in the baselining output that I'm not able to figure out so far...

testdata/baselines/reference/submodule/fourslash/smartSelection/smartSelection_behindCaret.baseline.diff
andrewbranch: Have you tried this in VS Code? They have word/token-based providers, and what the user experiences is the union of all the ranges provided, so this probably gets highlighted first anyway.

PR #1945, opened 2025-10-23
Fix leading source file comment emit bugs

### LINKED ISSUE #1526: Emit adds generated code before file header/copyright comment
## Steps to reproduce

Compile a ts file with a copyright comment:

```ts
/*---------------------------------------------------------------------------------------------
 *  Copyright
 *--------------------------------------------------------------------------------------------*/

export const a = 1;
```

The generated helpers and module code is added after the comment:

```js
"use strict";
/*---------------------------------------------------------------------------------------------
 *  Copyright
 *--------------------------------------------------------------------------------------------*/
exports.a = void 0;
exports.a = 1;
```

## Behavior with `tsgo`
The copyright comment after the added code:

```js
"use strict";
exports.a = void 0;
/*---------------------------------------------------------------------------------------------
 *  Copyright
 *--------------------------------------------------------------------------------------------*/
exports.a = 1;
```

jakebailey:
Fixes #1526

#1938 sort of found the right area but I found some misports that fix way more baselines still.

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -4400,6 +4400,8 @@ func (p *Printer) emitSourceFile(node *ast.SourceFile) {
 		if node.IsDeclarationFile {
 			p.emitTripleSlashDirectives(node)
 		}
+	} else {
+		state = p.emitDetachedCommentsBeforeStatementList(node.AsNode(), node.Statements.Loc)
 	}
 
 	// !!! Emit triple-slash directives


internal/printer/printer.go
```diff
@@ -5083,7 +5087,7 @@ func (p *Printer) emitDetachedCommentsBeforeStatementList(node *ast.Node, detach
 	containerPos := p.containerPos
 	containerEnd := p.containerEnd
 	declarationListContainerEnd := p.declarationListContainerEnd
-	skipLeadingComments := emitFlags&EFNoLeadingComments == 0 && !ast.PositionIsSynthesized(detachedRange.Pos())
+	skipLeadingComments := ast.PositionIsSynthesized(detachedRange.Pos()) || emitFlags&EFNoLeadingComments != 0
```
jakebailey, line 5092: This was a porting bug.

internal/printer/printer.go
```diff
@@ -4384,14 +4384,18 @@ func (p *Printer) emitSourceFile(node *ast.SourceFile) {
...
 	p.generateAllNames(node.Statements)
 
 	index := 0
+	var state *commentState
 	if node.ScriptKind != core.ScriptKindJSON {
 		p.emitShebangIfNeeded(node)
 		index = p.emitPrologueDirectives(node.Statements)
+		if !p.writer.IsAtStartOfLine() {
+			p.writeLine()
+		}
```
jakebailey, line 4397: This is required. I am not sure where Strada did this particular newline.

DanielRosenwasser (approved):

internal/printer/printer.go
weswigham: The first line of `emitSourceFile` is an unconditional `writeLine`, which is after `writeFile`'s `emitShebangIfNeeded` and `emitPrologueDirectivesIfNeeded` calls. Which puts this in exactly the right place.

internal/printer/printer.go
```diff
@@ -4384,18 +4384,24 @@ func (p *Printer) emitSourceFile(node *ast.SourceFile) {
...
+		state = p.emitDetachedCommentsBeforeStatementList(node.AsNode(), node.Statements.Loc)
 		p.emitHelpers(node.AsNode())
 		if node.IsDeclarationFile {
 			p.emitTripleSlashDirectives(node)
 		}
+	} else {
```
weswigham, line 4403: I don't think a JSON document will ever have detached leading comments (since it only ever has one statement and no prologues), so this *could* probably be a hardcoded initial empty comment emit state, but it's probably fine.

internal/printer/printer.go
weswigham: the ol' "let's invert this condition" and then fail to invert the usage/name.

internal/printer/printer.go
jakebailey: Yeah, no test change when I added this back (nil is fine), but since the original code did this unconditionally (both codebases) I thought I would add it back. I bet we just don't have any tests which reprint a JSON file without downleveling it...

PR #1953, opened 2025-10-25
Properly handle hovering on `this`

ahejlsberg:
Currently, hovering on `this` in an expression context displays `any`. With this PR we display the correct type.


jakebailey:
Huh, did we really not have an existing hover test for this? Surprising

ahejlsberg:
> Huh, did we really not have an existing hover test for this? Surprising

We undoubtedly have hover tests for `this` in class methods (which was working), but apparently not for methods in object literals.

PR #1958, opened 2025-10-26
fix: track untyped function-like signatures in js files

a-tarasyuk:


testdata/baselines/reference/submoduleAccepted/compiler/jsFileFunctionParametersAsOptional2.errors.txt.diff
```diff
@@ -0,0 +1,34 @@
+--- old.jsFileFunctionParametersAsOptional2.errors.txt
++++ new.jsFileFunctionParametersAsOptional2.errors.txt
+@@= skipped -0, +0 lines =@@
+-bar.ts(1,1): error TS2554: Expected 3 arguments, but got 0.
```
jakebailey: Seems like these errors are missing.

jakebailey: Seems like these errors are missing.

a-tarasyuk: I’m not quite sure — this seems like a special case where the parameters exist but have no explicit types.

```ts
/**
 * @param a
 * @param b
 * @param c
 */
function f(a, b, c) { }
```

ahejlsberg:
I have a more comprehensive fix in #1960.

a-tarasyuk:
@ahejlsberg ok

PR #1960, opened 2025-10-27
Several fixes to JS typing of functions and methods

### LINKED ISSUE #1521: Incorrect Argument Count Error for optional params.
## Summary

When using TypeScript Native Preview with a standard ESM setup and the `semver` package, calling `semver.gt` with two arguments in a JavaScript file results in a TypeScript error about missing arguments. This does not occur with the same code in other TypeScript environments, suggesting a compatibility issue specific to TypeScript Native Preview.

## Environment

- **TypeScript Native Preview**: `@typescript/native-preview@7.0.0-dev.20250804.1`
- **Node.js**: >=20.6.0
- **semver**: `^7.6.0`
- **jsconfig.json**: (using `"module": "esnext"`, `"target": "es2020"`)
- **Repro file**: `index.js`
- **Command**: `tsgo -p ./jsconfig.json && node ./index.js`

## Repro Steps
Clone this and run the main command: https://github.com/Bashamega/jsconfig-repro

## Actual Behavior
The following Error is produced:```bash
> jsconfig-repro@1.0.0 main
> tsgo -p ./jsconfig.json && node ./index.js

index.js:9:20 - error TS2554: Expected 3 arguments, but got 2.

9 console.log(semver.gt(version, '1.2.0')); // true
                     ~~

  node_modules/semver/functions/gt.js:2:19 - An argument for 'loose' was not provided.
    2 const gt = (a, b, loose) => compare(a, b, loose) > 0
                        ~~~~~

Found 1 error in index.js:9
```

## Expected Behavior
It should return `true`

Originally discovered in:
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2104

### LINKED ISSUE #1957: TS2554 [ERROR]: Expected 2 arguments, but got 1.
## Steps to reproduce

a.js
```js
export function c(a, b) {
}
```
b.ts
```ts
import { c } from "./a.js";

c(4);
```

no error

## Behavior with `tsgo`

TS2554 [ERROR]: Expected 2 arguments, but got 1.
c(4);

ahejlsberg:
In this PR:

* Properly compute `SignatureFlagsIsUntypedSignatureInJSFile` such that calls to untyped functions and methods declared in .js files allow arguments to be omitted.
* Limit `@type` JSDoc annotations to functions and methods (specifically, ignore `@type` annotations in constructors and accessors).
* Use full signature and parameter names from `@type` annotations (similar to Strada).
* Ensure `export` modifier is respected when `@type` annotations are present.

Fixes #1521.
Fixes #1957.

### DIFF
--- a/testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
+++ b/testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
@@ -6,21 +6,18 @@
  test.js(7,5): error TS2322: Type '(prop: any) => void' is not assignable to type '{ prop: string; }'.
 -test.js(10,12): error TS8030: The type of a function declaration must match the function's signature.
 -test.js(23,12): error TS8030: The type of a function declaration must match the function's signature.
--test.js(27,7): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
--  Target signature provides too few arguments. Expected 1 or more, but got 0.
--test.js(30,7): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
--  Target signature provides too few arguments. Expected 1 or more, but got 0.
--test.js(34,3): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
--  Target signature provides too few arguments. Expected 1 or more, but got 0.
--
--
--==== test.js (7 errors) ====
 +test.js(10,12): error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
 +test.js(23,12): error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
+ test.js(27,7): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+   Target signature provides too few arguments. Expected 1 or more, but got 0.
+ test.js(30,7): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+   Target signature provides too few arguments. Expected 1 or more, but got 0.
+-test.js(34,3): error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+-  Target signature provides too few arguments. Expected 1 or more, but got 0.
 +test.js(33,14): error TS8030: A JSDoc '@type' tag on a function must have a signature with the correct number of arguments.
-+
-+
-+==== test.js (5 errors) ====
+
+
+ ==== test.js (7 errors) ====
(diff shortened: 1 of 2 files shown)


testdata/baselines/reference/submoduleAccepted/conformance/checkJsdocTypeTag6.errors.txt.diff
```diff
@@ -43,7 +46,16 @@
...
+     const variableWithMoreParameters = function (more) {}; // error
+-          ~~~~~~~~~~~~~~~~~~~~~~~~~~
+-!!! error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+-!!! error TS2322:   Target signature provides too few arguments. Expected 1 or more, but got 0.
+     
+     /** @type {() => void} */
+     const arrowWithMoreParameters = (more) => {}; // error
+-          ~~~~~~~~~~~~~~~~~~~~~~~
+-!!! error TS2322: Type '(more: any) => void' is not assignable to type '() => void'.
+-!!! error TS2322:   Target signature provides too few arguments. Expected 1 or more, but got 0.
```
jakebailey: Curious loss of error; it's maybe fine but I would have expected this to still error given this is contextually typed by the declarations?

ahejlsberg: This looks to be an issue with reparsing. In Corsa, we currently consider the `@type` annotation to only apply to the variable declaration and not the arrow function. In Strada we consider the annotation to apply to both. I'll look at fixing this, provided we like the Strada approach.

jakebailey: I feel like this is something @sandersn changed but I can't remember anymore.

If it's a big change I don't think it's a big deal to ignore for this PR either.

ahejlsberg: And, I should add, in Corsa, because the `@type` annotation doesn't apply to the arrow function, the arrow function is considered completely untyped and therefore gets the `SignatureFlagsIsUntypedSignatureInJSFile` flag and the more permissive arity check.

ahejlsberg: And, I should further add, the arrow function is considered completely untyped because the contextual type doesn't have an applicable signature. That's a bit too aggressive, so with the latest commit, any contextual type for a function in a .js file causes us to not consider that function untyped.

PR #1962, opened 2025-10-27
Actually transform KindCommonJSExport in declaration emit

### LINKED ISSUE #1178: "invalid memory address or nil pointer dereference"
Hello,
I found an error from running tsgo. I am not exactly sure what file caused it but I believe it's reproducible. it's a large-ish open source project here.

## Repro

```
git clone git@github.com:GMOD/jbrowse-components
cd jbrowse-components
git clone tsgo_error_repro #branch to checkout with tsgo
yarn
yarn tsgo
```

## Environment

running on node 24.1.0, ubuntu 24

"@typescript/native-preview": "^7.0.0-dev.20250612.1",

## Log

```
❯❯❯ y tsgo
yarn run v1.22.22
$ /home/cdiesh/src/jbrowse-components/node_modules/.bin/tsgo
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6f32b2]

goroutine 10700 [running]:
github.com/microsoft/typescript-go/internal/ast.IsBlock(...)
        github.com/microsoft/typescript-go/internal/ast/ast.go:3078
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitConciseBody(0xc0fa3d6600?, 0xcea298?)
        github.com/microsoft/typescript-go/internal/printer/printer.go:2570 +0x12
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitArrowFunction(0xc0fa3d6600, 0xc0f94497c0)
        github.com/microsoft/typescript-go/internal/printer/printer.go:2593 +0x20c
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitExpression(0xc0fa3d6600, 0xc0f94497c0, 0x2)
        github.com/microsoft/typescript-go/internal/printer/printer.go:3030 +0x5a9
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitInitializer(0xc0fa3d6600, 0xc0f94497c0, 0x7fc, 0xc0fad32fc0)
        github.com/microsoft/typescript-go/internal/printer/printer.go:1480 +0xa6
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitCommonJSExport(0xc0fa3d6600, 0xc0fad32fc0)
        github.com/microsoft/typescript-go/internal/printer/printer.go:3760 +0x134
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitStatement(0x76b0b5774248?, 0xc0f4fab860?)
        github.com/microsoft/typescript-go/internal/printer/printer.go:3967 +0x46f
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitListItems(0xc0fa3d6600, 0xc00ef0, 0xc0fa3cc508, {0xc0fad50ae0, 0x2, 0x20?}, 0x1, 0x0, {0x0, 0xb08})
        github.com/microsoft/typescript-go/internal/printer/printer.go:4676 +0x4b9
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitListRange(0xc0fa3d6600, 0xc00ef0, 0xc0fa3cc508, 0xc0fad358a0, 0x1, 0xaca940?, 0xc00616c788?)
        github.com/microsoft/typescript-go/internal/printer/printer.go:4473 +0x395
github.com/microsoft/typescript-go/internal/printer.(*Printer).emitSourceFile(0xc0fa3d6600, 0xc0fa3cc508)
        github.com/microsoft/typescript-go/internal/printer/printer.go:4395 +0x208
github.com/microsoft/typescript-go/internal/printer.(*Printer).Write(0xc0fa3d6600, 0xc0fa3cc508, 0xc0f7076fc0?, {0xd04590, 0xc0f4081e30}, 0x1?)
        github.com/microsoft/typescript-go/internal/printer/printer.go:4925 +0x100c
github.com/microsoft/typescript-go/internal/compiler.(*emitter).printSourceFile(0xc0f4c7fc20, {0xc0132a8b80, 0x74}, {0x0, 0x0}, 0xc0fa3cc508, 0xc0fa3d6600, 0x0)
        github.com/microsoft/typescript-go/internal/compiler/emitter.go:158 +0x289
github.com/microsoft/typescript-go/internal/compiler.(*emitter).emitDeclarationFile(0xc0f4c7fc20, 0xc00616c788, {0xc0132a8b80, 0x74}, {0x0, 0x0})
        github.com/microsoft/typescript-go/internal/compiler/emitter.go:127 +0x5c5
github.com/microsoft/typescript-go/internal/compiler.(*emitter).emit(0xc0f4c7fc20)
        github.com/microsoft/typescript-go/internal/compiler/emitter.go:43 +0x5d
github.com/microsoft/typescript-go/internal/compiler.(*Program).Emit.func2()
        github.com/microsoft/typescript-go/internal/compiler/program.go:762 +0x149
github.com/microsoft/typescript-go/internal/core.(*parallelWorkGroup).Queue.func1()
        github.com/microsoft/typescript-go/internal/core/workgroup.go:39 +0x50
created by github.com/microsoft/typescript-go/internal/core.(*parallelWorkGroup).Queue in goroutine 1
        github.com/microsoft/typescript-go/internal/core/workgroup.go:37 +0x85

```

## Maybe related

Saw this issue, somewhat similar error https://github.com/microsoft/typescript-go/issues/898

weswigham:
Fixes #1178

The bug was less that something was broken, and more like nobody ever added handling for the newly reparsed node kind to declaration emit at all. I'm not 100% sure how these should be translated to declarations, but I think I have reasonable transforms in the code. This is enough implementation to fix the crashes, but, looking at the baselines, behavior in some cases stands to be improved (since this is mostly just variants of the export assignment handling logic), as it produces some pretty ick output. In particular, multiple declarations not merging is... OK (different from strada, ofc), but maybe could use some checks to see if they can combine/be elided to improve the output. Additionally, some namespacey patterns that are seemingly not well supported checking-ways anymore have seemingly equally poor declaration emit (eg, `var ns: typeof ns`). Additionally, `const`-ness and maybe-const-initializer-y-ness of the variables isn't correctly handled yet (instead the non-const type is always emitted), which requires integrating some of the `export var` logic into the export assignment logic I reused. IMO, those are all reasonable follow-up improvements since I'm still trying to focus on normal emit, and really just wanted to fix the crash here.


testdata/baselines/reference/submodule/compiler/modulePreserve4.js.diff
```diff
@@ -30,7 +30,7 @@
  //// [a.d.ts]
 -export const x: 0;
 +export declare const x = 0;
-+export var y = 0;
++export declare var y: number;
```
jakebailey, line 33: I saw this when trying to fix this crash too; definitely some weird widening interplay here

testdata/baselines/reference/submodule/conformance/assignmentToVoidZero1.js.diff
```diff
@@ -19,7 +19,6 @@
...
 -export const x: 1;
 -export const y: 2;
-+export var y = exports.x = void 0;
-+export var x = 1;
-+export var y = 2;
-+export {};
\ No newline at end of file
++export declare var y: undefined;
++export declare var x: number;
++export declare var y: undefined;
```
jakebailey, line 24: Weird, we're taking the initial assignment for the type of these?

jakebailey:
Agh bitten by leftover baselines (main shouldn't do this anymore, usually)

testdata/baselines/reference/submodule/compiler/modulePreserve4.js.diff
weswigham: Well, the old output was because the `KindCommonJSExport` declaration was just being emitted as-is. :S That's why we had things like `export var x = { x: "x" }` in the output which just...isn't actually something we should ever have in a declaration file? And is an error?

testdata/baselines/reference/submodule/conformance/assignmentToVoidZero1.js.diff
weswigham: `ensureType` uses the serialized type of the declaration which, generally speaking, is inferred from the first initializer. It's just whatever `getTypeOfSymbol` returns, without any special handling for `const`-y-ness, excluding some special cases it already handles for TS code.

weswigham:
> Probably worth a new issue about it generally.

JS declaration emit, in general, needs a polish/completeness pass - nobody's just gone through the js declaration baselines and fixed all the diffs yet, since we've still been working on JS handing at the reparser step, afaik. What works is mostly by chance and commonality with TS (which is perhaps unsurprisingly a lot of it!).

jakebailey (approved):

PR #1965, opened 2025-10-28
Fix various module emit differences

jakebailey:
Various things I noticed while looking at #1963


testdata/baselines/reference/submodule/compiler/aliasBug.js.diff
```diff
@@ -1,10 +1,9 @@
 --- old.aliasBug.js
 +++ new.aliasBug.js
-@@= skipped -36, +36 lines =@@
-         })(baz = bar.baz || (bar.baz = {}));
+@@= skipped -37, +37 lines =@@
      })(bar = foo.bar || (foo.bar = {}));
  })(foo || (foo = {}));
--var provide = foo;
+ var provide = foo;
 -var booz = foo.bar.baz;
```
jakebailey: Still not quite there, apparently...

testdata/baselines/reference/submodule/compiler/constEnums.js.diff
```diff
@@ -4,7 +4,7 @@
          })(C = B.C || (B.C = {}));
      })(B = A2.B || (A2.B = {}));
  })(A2 || (A2 = {}));
--var I2 = A2.B;
++var I1 = A1.B;
```
jakebailey: Regression here.

testdata/baselines/reference/submodule/compiler/importDeclWithClassModifiers.js.diff
```diff
@@ -1,11 +1,10 @@
...
- "use strict";
+@@= skipped -15, +15 lines =@@
  Object.defineProperty(exports, "__esModule", { value: true });
--exports.c = exports.b = exports.a = void 0;
--exports.a = x.c;
+ exports.c = exports.b = exports.a = void 0;
+ exports.a = x.c;
 -exports.b = x.c;
--exports.c = x.c;
++b = x.c;
```
jakebailey: Regression here.

testdata/baselines/reference/submodule/compiler/importDeclWithExportModifierAndExportAssignment.js.diff
```diff
@@ -4,7 +4,7 @@
 
  //// [importDeclWithExportModifierAndExportAssignment.js]
  "use strict";
--exports.a = void 0;
--exports.a = x.c;
--module.exports = x;
-+Object.defineProperty(exports, "__esModule", { value: true });
\ No newline at end of file
++Object.defineProperty(exports, "__esModule", { value: true });
```
jakebailey, line 7: Regression here.

testdata/baselines/reference/submodule/conformance/exportsAndImports3-es6.js.diff
```diff
@@ -1,24 +1,21 @@
...
-@@= skipped -19, +19 lines =@@
+@@= skipped -16, +16 lines =@@
  var M;
  (function (M) {
  })(M || (exports.M1 = exports.M = M = {}));
 -exports.a = M.x;
--exports.a1 = exports.a;
++exports.a1 = exports.a = M.x;
+ exports.a1 = exports.a;
  //// [t2.js]
```
jakebailey: Sort of a regression but it's just as bad as the extra export above on `v`.

testdata/baselines/reference/submodule/compiler/importDeclWithClassModifiers.js.diff
jakebailey: Though, this is a syntax error. Just, weird

testdata/baselines/reference/submodule/compiler/importDeclWithExportModifierAndExportAssignment.js.diff
jakebailey:
```==== importDeclWithExportModifierAndExportAssignment.ts (3 errors) ====
    module x {
        interface c {
        }
    }
    export import a = x.c;
                      ~
!!! error TS2708: Cannot use namespace 'x' as a value.
                        ~
!!! error TS2694: Namespace 'x' has no exported member 'c'.
    export = x;
    ~~~~~~~~~~~
!!! error TS2309: An export assignment cannot be used in a module with other exported elements.
```

Hard to feel bad about breaking this one though...

internal/binder/binder.go
```diff
@@ -794,8 +794,11 @@ func (b *Binder) bindModuleDeclaration(node *ast.Node) {
 		state := b.declareModuleSymbol(node)
 		if state != ast.ModuleInstanceStateNonInstantiated {
 			symbol := node.AsModuleDeclaration().Symbol
-			if symbol.Flags&(ast.SymbolFlagsFunction|ast.SymbolFlagsClass|ast.SymbolFlagsRegularEnum) != 0 || state != ast.ModuleInstanceStateConstEnumOnly {
-				// if module was already merged with some function, class or non-const enum, treat it as non-const-enum-only
+			if (symbol.Flags&(ast.SymbolFlagsFunction|ast.SymbolFlagsClass|ast.SymbolFlagsRegularEnum) == 0) &&
+				state == ast.ModuleInstanceStateConstEnumOnly &&
+				symbol.Flags&ast.SymbolFlagsConstEnumOnlyModule == 0 {
+				symbol.Flags |= ast.SymbolFlagsConstEnumOnlyModule
+			} else {
```
jakebailey: The old compiler used a `boolean | undefined` on every symbol for this. Corsa uses a flag but didn't make the behavior match the tristate. Doing this instead to avoid adding a Tristate to all symbols.

internal/transformers/moduletransforms/commonjsmodule.go
```diff
@@ -1080,7 +1080,14 @@ func (tx *CommonJSModuleTransformer) visitTopLevelVariableStatement(node *ast.Va
 					propertyAccess,
 					v.Name().Clone(tx.Factory()),
 				))
+			} else if ast.IsIdentifier(v.Name()) {
+				expression := tx.transformInitializedVariable(v)
```
jakebailey, line 1078: The old compiler had this func instead which did something special with `exports`, unlike `ConvertVariableDeclarationToAssignmentExpression` which is new. Not extremely familiar, though.

testdata/baselines/reference/submodule/compiler/importDeclWithExportModifierAndExportAssignment.js.diff
jakebailey: No, it was already broken before too, nevermind.

internal/binder/binder.go
jakebailey: Hm, I am not sure that is actually correct, thinking harder about it. I think this will fire when you just stick two const enums together in the same scope.

jakebailey: Drafting one so I can figure out what the deal is

jakebailey:
I think this is reviewable now.

PR #1977, opened 2025-10-29
Move change tracker, converters, utils to separate packages

andrewbranch:
I started down this path while working on auto-imports and pulled it out to a standalone PR as it's likely to cause conflicts.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -687,3 +687,13 @@ func DeduplicateSorted[T any](slice []T, isEqual func(a, b T) bool) []T {
 
 	return deduplicated
 }
+
+// CompareBooleans treats true as greater than false.
+func CompareBooleans(a, b bool) int {
+	if a && !b {
+		return -1
+	} else if !a && b {
+		return 1
+	}
+	return 0
+}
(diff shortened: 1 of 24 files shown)


jakebailey (approved):
Seems good, though I am not sure if the name `change` will be annoying or not given the genericness of that name.

jakebailey (approved):
`lsutils` seems like a weird home for `UserPreferences`; makes me think that this whole thing should be inverted somehow

andrewbranch:
I feel like UserPreferences : something :: DocumentURI : lsproto, so I think it makes sense for it to go in a lower-level, dependency-light package that can be depended upon by various bits of LS functionality. I agree `lsutils` isn’t an amazing name for that, but there was no technical need to make a separate `lsprefs` or `lstypes` package or something.

andrewbranch:
If _every_ LS feature moved into its own package (not unreasonable, but a big lift at the moment), that would free up `ls` to be the set of common types and utilities that all the features use, which is probably what you mean by inverting. This is a move in that direction, but because it's a partial move, the name `ls` is still the mega-dependency-heavy thing, so common types have to move out for now.

jakebailey:
Yeah. I wasn't intending to imply something actionable here, I definitely see where this is going and it's all good!

PR #1978, opened 2025-10-29
Export all types referenced through other exported APIs, enforce

jakebailey:
Getting this lint rule prototype out of my git stash.

Exported APIs containing unexported elements is generally annoying, because it means consuming code can't write down the type name but can do things with it (even still creating values of that type via generic helpers!). It also means people doing funky stuff like tsgolint have to work around APIs that aren't fully declared.

This PR adds a lint rule to find these cases and complain, then fixes everything.


jakebailey:
FYI @camc314

sheetalkamat (approved):

PR #1979, opened 2025-10-30
Consistent rules for mixing `@type`, `@param`, `@return`, `@template`

### LINKED ISSUE #1970: TS8040: A JSDoc '@type' tag may not occur with a '@param' or '@returns' tag.
## Steps to reproduce

```js
/**
 * @typedef {object} AliasInitConfig
 * @prop {`file://${string}`} AliasInitConfig.location The name or fully resolved location of the tsconfig.
 */
/**
 * @type {import('node:module').InitializeHook}
 * @param {AliasInitConfig} [config] Configuration object to customise Alias loader.
 */
function initialiseAlias(config) {
}
```

It works fine

## Behavior with `tsgo`

It (incorrectly) errors

---

Possibly fixed by https://github.com/microsoft/typescript-go/pull/1960 (description looks kinda related).

ahejlsberg:
This PR implements rules for mixing of `@type`, `@param`, `@return`, and `@template` JSDoc annotations and removes the restriction from #1468 that made it an error to mix `@type` with any of the other annotations. The specific rules implemented by the PR are:

* `@type` on a function is applicable when there are no preceding `@param`, `@return`, or `@template` annotations and no `@type` annotations on the function's parameters.
* `@param` for a parameter is applicable when there is no applicable `@type` annotation for the function and no `@type` annotation on the parameter.
* `@return` is applicable when there is no applicable `@type` annotation for the function.
* `@template` is applicable when there is no applicable `@type` annotation for the function.
* `@type` on a parameter is always applicable.

It isn't an error for an annotation to be inapplicable. Rather, the type specified in the annotation is simply ignored.

These rules aren't necessarily ideal (whatever that might be), but they are consistent and enforceable without excessively complicated logic.

Fixes #1970.

### DIFF
--- a/internal/diagnostics/diagnostics_generated.go
+++ b/internal/diagnostics/diagnostics_generated.go
@@ -3518,8 +3518,6 @@ var Decorators_may_not_appear_after_export_or_export_default_if_they_also_appear
 
 var A_JSDoc_template_tag_may_not_follow_a_typedef_callback_or_overload_tag = &Message{code: 8039, category: CategoryError, key: "A_JSDoc_template_tag_may_not_follow_a_typedef_callback_or_overload_tag_8039", text: "A JSDoc '@template' tag may not follow a '@typedef', '@callback', or '@overload' tag"}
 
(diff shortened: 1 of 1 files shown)


jakebailey:
Needs a `go generate ./internal/diagnostics` to resync the diags that were modified.

jakebailey (approved):

JakobJingleheimer:
Thanks!

PR #1981, opened 2025-10-30
Update golangci-lint, fix issues, modernize

jakebailey:
Updates the version, fixes or silences existing issues, then also enables the modernize fixes and fixes them except for the small number of cases I would like to ignore.

### DIFF
--- a/internal/vfs/vfstest/vfstest_test.go
+++ b/internal/vfs/vfstest/vfstest_test.go
@@ -285,6 +285,7 @@ func TestStress(t *testing.T) {
 			}
 		})
 	}
+	wg.Wait()
 }
 
 func TestParentDirFile(t *testing.T) {


internal/core/core.go
```diff
@@ -160,7 +160,7 @@ func Same[T any](s1 []T, s2 []T) bool {
 }
 
 func Some[T any](slice []T, f func(T) bool) bool {
```
jakebailey, line 162: This func arguably shouldn't exist but that's larger diff.

internal/tsoptions/tsconfigparsing.go
```diff
@@ -1288,7 +1288,7 @@ func parseJsonConfigFileContentWorker(
 			projectReferences = []*core.ProjectReference{}
 			for _, reference := range newReferencesOfRaw.sliceValue {
 				for _, ref := range parseProjectReference(reference) {
-					if reflect.TypeOf(ref.Path).Kind() != reflect.String {
+					if reflect.TypeFor[string]().Kind() != reflect.String {
```
jakebailey: This code is suspect. Always true.

PR #1985, opened 2025-10-30
Add --checkers to control number of checkers per Program

jakebailey:
Much simplified version of #1176.

`--checkers` controls how many checkers are created in a `Program` during a non-editor compile. The default is still 4. Parse/bind/emit are still massively-concurrent.

If you pass `--singleThreaded`, everything is still single threaded as before; `checkers` will be forced to `1` and all normally concurrent operations will instead proceed in a deterministic order.

Intended to be used in tandem with a not-yet-existing `--builders` flag for controlling `-b` concurrency (probably via a revived #1692; @sheetalkamat).

### DIFF
--- a/testdata/baselines/reference/tsbuild/commandLine/help.js
+++ b/testdata/baselines/reference/tsbuild/commandLine/help.js
@@ -127,7 +127,7 @@ Run in single threaded mode.
 Generate pprof CPU/memory profiles to the given directory.
 
 [94m--checkers[39m
-Set the number of checkers per program.
+Set the number of checkers per project.
 
 [94m--verbose, -v[39m
 Enable verbose logging.
--- a/testdata/baselines/reference/tsc/commandLine/help-all.js
+++ b/testdata/baselines/reference/tsc/commandLine/help-all.js
@@ -18,7 +18,7 @@ Show all compiler options.
 Build one or more projects and their dependencies, if out of date
 
 [94m--checkers[39m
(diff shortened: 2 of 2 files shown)


internal/compiler/program.go
```diff
@@ -247,7 +247,13 @@ func (p *Program) initCheckerPool() {
 	if p.opts.CreateCheckerPool != nil {
 		p.checkerPool = p.opts.CreateCheckerPool(p)
 	} else {
-		p.checkerPool = newCheckerPool(core.IfElse(p.SingleThreaded(), 1, 4), p)
+		checkers := 4
+		if p.SingleThreaded() {
+			checkers = 1
+		} else if p.Options().Checkers != nil {
+			checkers = max(*p.Options().Checkers, 1)
```
ahejlsberg: Do we want a sanity check that imposes an upper limit (say 1000 or something)?

jakebailey: I was thinking of topping it at 256, though I was hoping to have this unbounded for goofy testing too. I can cut it off. I feel like 256 is plenty high...

internal/diagnostics/extraDiagnosticMessages.json
```diff
@@ -11,6 +11,14 @@
         "category": "Message",
         "code": 100002
     },
+    "Set the number of checkers per program.": {
```
jakebailey: Perhaps this should actually say "project".

PR #1990, opened 2025-10-31
Implement semantic tokens

### LINKED ISSUE #1894: Add semantic highlighting support for editors
Mentioned in #1008 but creating an issue specifically about semantic highlighting

## Feature request
Add semantic highlighting support for editors

In VS Code, we've gotten pretty used to this feature so its absence is one of the first things people noticed when switching to tsgo

jakebailey:
Fixes #1894

semantic-tokens-2

I am a little scared of this one. VS Code seems to ask for doc highlights, semantic tokens, diags, all in quick succession. All three of those require checkers, which means we are likely creating 3 checkers to handle these. 4 if we merge inlay hints.


DanielRosenwasser:
> VS Code seems to ask for doc highlights, semantic tokens, diags, all in quick succession

Wasn't the same true for document highlights and semantic tokens?

Does it feel reasonable in a larger codebase?

DanielRosenwasser:
Ah, but we didn't make more checkers.

internal/lsp/server.go
```diff
@@ -971,8 +1000,15 @@ func shouldEnableWatch(params *lsproto.InitializeParams) bool {
 }
 
 func getCompletionClientCapabilities(params *lsproto.InitializeParams) *lsproto.CompletionClientCapabilities {
-	if params == nil || params.Capabilities == nil || params.Capabilities.TextDocument == nil {
+	if params.Capabilities == nil || params.Capabilities.TextDocument == nil {
```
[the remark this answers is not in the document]
jakebailey: Params is never nil, but I don't know why this got touched

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,580 @@
...
+		endPos := converters.PositionToLineAndCharacter(file, core.TextPos(tokenEnd))
+
+		// Length is the character difference when on the same line
+		var tokenLength uint32
+		if startPos.Line == endPos.Line {
+			tokenLength = endPos.Character - startPos.Character
+		} else {
+			// Multi-line tokens shouldn't happen for identifiers, but handle it
+			tokenLength = endPos.Character
+		}
```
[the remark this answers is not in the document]
jakebailey: Probably I should just panic, given we aren't expecting this ever.

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,580 @@
...
+		return tokenTypeEnum, true
+	}
+	if flags&ast.SymbolFlagsTypeAlias != 0 {
+		return tokenTypeType, true
+	}
+	if flags&ast.SymbolFlagsInterface != 0 {
+		if meaning&ast.SemanticMeaningType != 0 {
+			return tokenTypeInterface, true
+		}
+	}
```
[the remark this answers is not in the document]
jakebailey: This is a pure port:

```ts
else if (flags & SymbolFlags.Interface) {
    if (meaning & SemanticMeaning.Type) {
        return TokenType.interface;
    }
}
```

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -1439,6 +1461,71 @@ function parseBaselineSmartSelection(args: ts.NodeArray<ts.Expression>): Cmd {
...
+    if (!ts.isStringLiteralLike(formatArg)) {
+        console.error("semanticClassificationsAre first argument must be a string literal");
+        return undefined;
+    }
+
+    const format = formatArg.text;
+
+    // Only handle "2020" format for semantic tokens
+    if (format !== "2020") {
+        // Skip other formats like "original" - return sentinel to continue parsing
```
DanielRosenwasser: I don't remember what the distinction was, but it feels like you should just mark these as "unparsed" for later

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,580 @@
...
+	case ast.KindPropertySignature:
+		return tokenTypeProperty
+	case ast.KindInterfaceDeclaration:
+		return tokenTypeInterface
+	case ast.KindTypeAliasDeclaration:
+		return tokenTypeType
+	case ast.KindTypeParameter:
+		return tokenTypeTypeParameter
+	case ast.KindPropertyAssignment, ast.KindShorthandPropertyAssignment:
+		return tokenTypeProperty
```
DanielRosenwasser, line 369: Original code technically doesn't handle `PropertyAssignment`, but that's fine.

Any reason you did this as a switch over a map?

internal/fourslash/tests/gen/syntacticClassificationsConflictDiff3Markers1_test.go
```diff
@@ -0,0 +1,28 @@
...
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `class C {
+<<<<<<< HEAD
+    v = 1;
+||||||| merged common ancestors
+    v = 3;
+=======
+    v = 2;
+>>>>>>> Branch - a
+}`
```
gyanu1, line 25: looks like conflict

internal/fourslash/tests/gen/syntacticClassificationsConflictMarkers1_test.go
```diff
@@ -0,0 +1,26 @@
...
+func TestSyntacticClassificationsConflictMarkers1(t *testing.T) {
+	t.Parallel()
+
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `class C {
+<<<<<<< HEAD
+    v = 1;
+=======
+    v = 2;
+>>>>>>> Branch - a
```
gyanu1, line 22: same here

internal/fourslash/tests/gen/syntacticClassificationsConflictDiff3Markers2_test.go
```diff
@@ -0,0 +1,25 @@
...
+	t.Parallel()
+
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `<<<<<<< HEAD
+class C { }
+||||||| merged common ancestors
+class E { }
+=======
+class D { }
+>>>>>>> Branch - a`
```
gyanu1, line 23: here

internal/fourslash/tests/gen/syntacticClassificationsMergeConflictMarker1_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func TestSyntacticClassificationsMergeConflictMarker1(t *testing.T) {
+	t.Parallel()
+
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `<<<<<<< HEAD
+"AAAA"
+=======
+"BBBB"
+>>>>>>> Feature`
```
gyanu1, line 21: :)

internal/fourslash/tests/gen/syntacticClassificationsConflictMarkers1_test.go
jakebailey: Nope, note the test name; we have special parsing and tests explicitly for merge conflict markers in user code.

internal/lsp/server.go
```diff
@@ -690,6 +692,17 @@ func (s *Server) handleInitialize(ctx context.Context, params *lsproto.Initializ
 					},
 				},
 			},
+			SemanticTokensProvider: &lsproto.SemanticTokensOptionsOrRegistrationOptions{
+				Options: &lsproto.SemanticTokensOptions{
+					Legend: ls.SemanticTokensLegend(s.clientCapabilities.TextDocument.SemanticTokens),
```
jakebailey, line 1103: This is calculated but never saved anywhere for later use; I think this is an oops and I am recalculating it each time or something

jakebailey:
Drafting this for now, until we can improve the checker situation IMO

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,575 @@
...
+		return tokenTypeParameter
+	case ast.KindPropertyDeclaration:
+		return tokenTypeProperty
+	case ast.KindModuleDeclaration:
+		return tokenTypeNamespace
+	case ast.KindEnumDeclaration:
+		return tokenTypeEnum
+	case ast.KindEnumMember:
+		return tokenTypeEnumMember
+	case ast.KindClassDeclaration:
```
DanielRosenwasser: No support for `ClassExpression`?

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,575 @@
...
+						nodeFlags := ast.GetCombinedNodeFlags(decl)
+
+						if modifiers&ast.ModifierFlagsStatic != 0 {
+							tokenModifier |= tokenModifierStatic
+						}
+						if modifiers&ast.ModifierFlagsAsync != 0 {
+							tokenModifier |= tokenModifierAsync
+						}
+						if tokenType != tokenTypeClass && tokenType != tokenTypeInterface {
+							if (modifiers&ast.ModifierFlagsReadonly != 0) || (nodeFlags&ast.NodeFlagsConst != 0) || (symbol.Flags&ast.SymbolFlagsEnumMember != 0) {
```
DanielRosenwasser, line 254: Go just boils these down into a single check, right?

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,575 @@
...
+	visit = func(node *ast.Node) bool {
+		// Check for cancellation
+		if ctx.Err() != nil {
+			return false
+		}
+
+		if node == nil {
+			return false
+		}
+		if node.Flags&ast.NodeFlagsReparsed != 0 {
```
DanielRosenwasser, line 198: So we won't highlight in JSDoc then, right? Same with `module.export`s?

jakebailey: We won't walk into the fake nodes, but we will walk into the real ones, so the latter yes. To do the former requires walking into `JSDoc()`.

internal/fourslash/_scripts/convertFourslash.mts
jakebailey: I can't, the "unparsed" is all or nothing, and we have tests I am partially converting as most semantic token tests have both old and new (2020).

internal/ls/semantictokens.go
jakebailey: I was confused by your statement until I realized you meant "switch _instead_ of a map"; I was busy looking for a switch actually on a map somehow.

Switches are in general much faster than `map`

internal/ls/semantictokens.go
jakebailey: Strada didn't, actually

internal/ls/semantictokens.go
jakebailey: These are on three different flags, so I don't think so?

internal/ls/semantictokens.go
jakebailey: The trouble is that this results in:

```diff
 	f.VerifySemanticTokens(t, []fourslash.SemanticToken{
 		{Type: "class.declaration", Text: "x"},
-		{Type: "class", Text: "C"},
+		{Type: "class.declaration", Text: "C"},
 		{Type: "class.declaration", Text: "C"},
 		{Type: "class.declaration", Text: "D"},
-		{Type: "class", Text: "B"},
+		{Type: "class.declaration", Text: "B"},
 	})
 }
```

It's already a class, this just determines whether or not we say it's a declaration or not (which it isn't).

internal/ls/semantictokens.go
DanielRosenwasser: Oops, misread

internal/ls/semantictokens.go
jakebailey: Then again, this test is:

```ts
var x = class /*0*/C {}
class /*1*/C {}
class /*2*/D extends class /*3*/B{} { }`
```

So, maybe they are actually all "declarations" because they're named?

jakebailey: Oh duh, a class without a name has no token to colorize. So it's just down to whether or not a class expression is a "declaration".

DanielRosenwasser: I don't know how it's getting class but not class declaration. I'm also surprised the test doesn't actually have any references to its name.

```ts
var x = class C {
  equals(other: C) { return this == other; }
};
```

```ts

```

internal/ls/semantictokens.go
DanielRosenwasser: > We won't walk into the fake nodes, but we will walk into the real ones

I really hope this code is resilient against symbol lookups then.

internal/ls/semantictokens.go
```diff
@@ -0,0 +1,575 @@
...
+		prevInJSXElement := inJSXElement
+		if ast.IsJsxElement(node) || ast.IsJsxSelfClosingElement(node) {
+			inJSXElement = true
+		} else if ast.IsJsxExpression(node) {
+			inJSXElement = false
+		}
+
+		if ast.IsIdentifier(node) && node.Text() != "" && !inJSXElement && !isInImportClause(node) && !isInfinityOrNaNString(node.Text()) {
+			symbol := c.GetSymbolAtLocation(node)
+			if symbol != nil {
```
DanielRosenwasser, line 215: Might be worth inverting the nil check

internal/ls/semantictokens.go
DanielRosenwasser: Looks like it is with that big `if` check

DanielRosenwasser:
If you can add fuzzing support at https://github.com/microsoft/typescript-error-deltas that would be swell. I guess you'd have to wait for the next publish though?

internal/ls/semantictokens.go
jakebailey: I'll add a test that references it and see.

jakebailey: Done; does work as expected, the reference is class, but the declaration is declaration.

internal/ls/semantictokens.go
jakebailey: I could though Strada did not, actaully

jakebailey:
I'll look into the fuzzing, though this is sort of an irritating call due to the full/range/delta split

PR #1991, opened 2025-10-31
Multiproject requests like find all refs, rename and workspace symbols

sheetalkamat:


internal/lsp/lspservertests/testserver.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/microsoft/typescript-go/internal/ls/lsconv"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/project"
+	"github.com/microsoft/typescript-go/internal/testutil/fsbaselineutil"
+	"github.com/microsoft/typescript-go/internal/testutil/lsptestutil"
+	"github.com/microsoft/typescript-go/internal/testutil/projecttestutil"
+	"gotest.tools/v3/assert"
+)
+
+type testServer struct {
```
sheetalkamat: Added this to baseline the project state and watching stuff - also allows to not open all files and control which one to open instead of fourslash opening all files. (to control which project is open). 

i havent yet ported tests for watching and lifetime of project but i will do that separately

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -526,6 +526,14 @@ function generateCode() {
             writeLine(`\treturn s.TextDocument.Uri`);
             writeLine(`}`);
             writeLine("");
+
+            if (hasTextDocumentPosition(structure)) {
```
jakebailey, line 666: Was the new code intended to be within this block our outside?

jakebailey: Ah, I guess you combined two interfaces, okay.

internal/sourcemap/lineinfo.go
```diff
@@ -15,6 +15,9 @@ func CreateECMALineInfo(text string, lineStarts core.ECMALineStarts) *ECMALineIn
 }
 
 func (li *ECMALineInfo) LineCount() int {
+	if li == nil {
```
jakebailey: When would this happen? It seems like we wouldn't want this to happen...

internal/lsp/lspservertests/testserver.go
```diff
@@ -0,0 +1,254 @@
...
+		return text
+	}
+	return s.files[fileName].(string)
+}
+
+func (s *testServer) hoverToWriteProjectStatus(fileName string) {
+	// Do hover so we have snapshot to check things on!!
+	_, _, resultOk := lsptestutil.SendRequest(s.t, s.server, lsproto.TextDocumentHoverInfo, &lsproto.HoverParams{
+		TextDocument: lsproto.TextDocumentIdentifier{
+			Uri: lsproto.DocumentUri("file://" + fileName),
```
jakebailey: Maybe this is a pure code move from somewhere else, but this is technically wrong given escaping and such means we should be using the func that converts these.

jakebailey: (there seem to be a lot of these)

internal/lsp/server.go
```diff
@@ -480,16 +498,18 @@ var handlers = sync.OnceValue(func() handlerMap {
 	registerLanguageServiceDocumentRequestHandler(handlers, lsproto.TextDocumentDefinitionInfo, (*Server).handleDefinition)
 	registerLanguageServiceDocumentRequestHandler(handlers, lsproto.TextDocumentTypeDefinitionInfo, (*Server).handleTypeDefinition)
```
jakebailey, line 506: Do definition requests also need this? Since something can be augmented between projects?

internal/testutil/lsptestutil/lsptestutil.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	defaultTypeDefinitionCapabilities = &lsproto.TypeDefinitionClientCapabilities{
+		LinkSupport: ptrTrue,
+	}
+	defaultHoverCapabilities = &lsproto.HoverClientCapabilities{
+		ContentFormat: &[]lsproto.MarkupKind{lsproto.MarkupKindMarkdown, lsproto.MarkupKindPlainText},
+	}
+)
+
+type TestLspServer struct {
```
jakebailey: Nit, but, the Go convention is `TestLSPServer`. Not critical, we already did this elsewhere, so...

internal/testutil/lsptestutil/baseline.go
```diff
@@ -0,0 +1,589 @@
...
+	// [| some /*MARKER*/ text |]
+	// [| some text |]/*MARKER*/
+	// Stable sort should handle first two cases but with that marker will be before rangeEnd if locations match
+	// So we will defer writing marker in this case by checking and finding index of rangeEnd if same
+	var deferredMarkerIndex *int
+
+	for index, detail := range details {
+		if detail.span == nil && deferredMarkerIndex == nil {
+			// If this is marker position and its same as textEnd and/or contextEnd we want to write marker after those
+			for matchingEndPosIndex := index + 1; matchingEndPosIndex < len(details); matchingEndPosIndex++ {
```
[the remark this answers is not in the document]
jakebailey: This is true but not new.

jakebailey (commented):
Some comments; have not read the entire thing yet.

internal/lsp/server.go
sheetalkamat: no definition requests are never augmented .. infact we use definitions to find references in other projects

internal/sourcemap/lineinfo.go
sheetalkamat: this happens when file does not exist

sheetalkamat: Updated TryGetSourceMappingURL to handle it instead if thats better?

internal/lsp/lspservertests/testserver.go
andrewbranch: Finer control of open files would be useful in a lot of scenarios, but overall I find fourslash a lot easier to work with. Could we just add some options/functionality to fourslash instead of this?

internal/compiler/projectreferencefilemapper.go
```diff
@@ -107,31 +110,49 @@ func (mapper *projectReferenceFileMapper) getResolvedReferenceFor(path tspath.Pa
...
+) bool {
 	for index, path := range references {
 		if !seenRef.AddIfAbsent(path) {
 			continue
 		}
 		config, _ := mapper.configToProjectReference[path]
-		fn(path, config, parent, index)
-		mapper.forEachResolvedReferenceWorker(mapper.referencesInConfigFile[path], fn, config, seenRef)
+		if fn(path, config, parent, index) {
+			return true
```
andrewbranch: This is opposite of the Go `.Range` convention. I would suggest inverting and renaming everything to use the verb "range" instead of "forEach" so there's no guessing.

internal/project/configfileregistry.go
```diff
@@ -108,6 +109,73 @@ func (c *ConfigFileRegistry) clone() *ConfigFileRegistry {
...
+	FileName           string
+	RetainingProjects  iter.Seq[tspath.Path]
+	RetainingOpenFiles iter.Seq[tspath.Path]
+	RetainingConfigs   iter.Seq[tspath.Path]
+}
+
+// For testing
+func (c *ConfigFileRegistry) ForEachTestConfigEntry(cb func(tspath.Path, *TestConfigEntry)) {
+	if c != nil {
+		for path, entry := range c.configs {
```
andrewbranch, line 123: Don’t forget that this is a random order iteration—I didn’t see where you’re using this, but it could mess up baselining.

internal/lsp/lspservertests/testserver.go
sheetalkamat: Will do

internal/project/session.go
```diff
@@ -371,42 +375,158 @@ func (s *Session) Snapshot() (*Snapshot, func()) {
...
+	// if program doesnt contain this file any more ignore it
+	if !s.projectContainsFile(project, uri) {
+		return nil
+	}
+	return ls.NewLanguageService(project.GetProgram(), snapshot)
+}
+
+func (s *Session) projectContainsFile(project *Project, uri lsproto.DocumentUri) bool {
+	path := s.toPath(uri.FileName())
+	return project.containsFile(path)
```
andrewbranch: I was going to say project.containsFile should just be public, but this Session method is also private, so I'm not sure what this is for.

internal/project/session.go
```diff
@@ -371,42 +375,158 @@ func (s *Session) Snapshot() (*Snapshot, func()) {
...
+		ctx,
+		snapshotChangeRequest{requestedProjectTrees: &projectTreeRequest{requestedProjectTrees}},
+		func(snapshot *Snapshot, updatedSnapshot bool) UpdateReason {
+			return core.IfElse(!updatedSnapshot, UpdateReasonRequestedLoadProjectTree, UpdateReasonUnknown)
+		},
+	)
+	return snapshot
+}
+
+func (s *Session) ForEachProjectLocationLoadingProjectTree(
```
andrewbranch: This feels like an extremely specific API not well suited to Session.

internal/project/session.go
```diff
@@ -371,42 +375,158 @@ func (s *Session) Snapshot() (*Snapshot, func()) {
 	}
 }
 
-func (s *Session) GetLanguageService(ctx context.Context, uri lsproto.DocumentUri) (*ls.LanguageService, error) {
+func (s *Session) getSnapshot(
+	ctx context.Context,
+	changeRequest snapshotChangeRequest,
+	updateIf func(snapshot *Snapshot, updatedSnapshot bool) UpdateReason,
```
andrewbranch: I like the idea of an abstraction around getting a snapshot with conditional updates, but this feels off as-is. Returning `UpdateReasonUnknown` to mean “don’t update” is confusing, and ideally the callers of this function shouldn’t care if the snapshot was just updated with file changes. I'm pretty sure the update line I originally had should have been like this:

```diff
- if project == nil && !updateSnapshot || project != nil && project.dirty
+ if !updatesSnapshot && (project == nil || project.dirty)
```

so that should at least allow removing the `updatedSnapshot` parameter.

internal/project/snapshot.go
```diff
@@ -128,13 +136,33 @@ type APISnapshotRequest struct {
 	UpdateProjects *collections.Set[tspath.Path]
 }
 
+type projectTreeRequest struct {
+	// If null, all project trees need to be loaded, otherwise only those that are referenced
+	referencedProjects map[tspath.Path]struct{}
+}
+
+type snapshotChangeRequest struct {
```
andrewbranch: Naming nit: these are resources that must be loaded in preparation for a language service operation. The other fields of SnapshotChange are mostly data that has changed. Maybe something like

```go
type ResourceRequest struct {
  Documents   []lsproto.DocumentUri
  Projects    []tspath.Path
  ProjectTree *ProjectTreeRequest
}
```

?

internal/project/snapshot.go
```diff
@@ -128,13 +136,33 @@ type APISnapshotRequest struct {
...
+	// requestedURIs are URIs that were requested by the client.
+	// The new snapshot should ensure projects for these URIs have loaded programs.
+	requestedURIs []lsproto.DocumentUri
+	// Update requested projects.
+	// this is used when we want to get LS and from all the projects the file can be part of
+	requestedProjectUpdates []tspath.Path
+	// Ensure default project for these URIs
+	// These are documents that might not be open, but we need default project for them
+	// eg for Find all references when the definition is in a mapped location that is not yet open
+	ensureDefaultProjectForURIs []lsproto.DocumentUri
```
andrewbranch: Does this actually need to be separate from requestedURIs? It sounds like the preparation would be the same, except that for requestedURIs, we assume the default project already exists.

internal/project/configfileregistry.go
sheetalkamat: the diff table sorts it before printing

internal/project/snapshot.go
sheetalkamat: this is where we "create a default project if it doesnt exist".. The other uses (getting LS for request ) already have project and throw if project isnt present. Trying to create project if it does not exist seems like something was intentional so created separate one.

andrewbranch: I would have said we could remove the panic and combine them, but we have seen that panic come up in issue reports, so I guess it might be a good idea to keep them separate, if only for diagnostic purposes 🤔

andrewbranch: Although, we can easily determine whether the project should already exist by whether the document URI requested is open.

```go
if b.fs.isOpenFile(requestedDocument) {
  // panic if project doesn't exist
} else {
  // create project if it doesn't exist
}
```

sheetalkamat:
converting tests to fourslash uncovered some more ordering issues with results. Looking into that as well as handling the feedback

jakebailey:
I'll give it another pass shortly, though I was waiting until it was mostly passing CI.

internal/fourslash/_scripts/updateFailing.mts
```diff
@@ -13,15 +13,15 @@ function main() {
     const go = which.sync("go");
     let testOutput: string;
     try {
-        testOutput = cp.execFileSync(go, ["test", "./internal/fourslash/tests/gen"], { encoding: "utf-8" });
+        testOutput = cp.execFileSync(go, ["test", "-v", "./internal/fourslash/tests/gen"], { encoding: "utf-8" });
```
jakebailey, line 16: Does this end up printing a lot?

sheetalkamat:
> All my concerns are addressed, hopefully someone else can review as well 👍

I am still looking into deduplicating results as there is still something going on there.

internal/fourslash/_scripts/updateFailing.mts
sheetalkamat: This is when it panics we can diagnose the issue rather than having to manually run those steps and tests again to figure out which test had panic that was not expected. (eg i got one when i was not associating reqID for panics when operating on projects inside those go routines.)

sheetalkamat:
> > All my concerns are addressed, hopefully someone else can review as well 👍
> 
> I am still looking into deduplicating results as there is still something going on there.


Have resolved this now

internal/fourslash/baselineutil.go
```diff
@@ -885,3 +902,64 @@ func (t *textWithContext) getIndex(i any) *int {
 func codeFence(lang string, code string) string {
 	return "```" + lang + "\n" + code + "\n```"
 }
+
+func symbolInformationToData(symbol *lsproto.SymbolInformation) string {
+	var symbolKindToString string
+	switch symbol.Kind {
+	case lsproto.SymbolKindFile:
```
jakebailey, line 909: Not for this PR of course, but this could be very easily generated in the LSP type generator; I should arguably do that for any of these enums that aren't just strings already.

internal/fourslash/fourslash.go
```diff
@@ -432,6 +457,39 @@ func (f *FourslashTest) readMsg(t *testing.T) *lsproto.Message {
...
+func sendRequest[Params, Resp any](t *testing.T, f *FourslashTest, info lsproto.RequestInfo[Params, Resp], params Params) Resp {
+	t.Helper()
+	prefix := f.getCurrentPositionPrefix()
+	f.baselineState(t)
+	f.baselineRequestOrNotification(t, info.Method, params)
+	resMsg, result, resultOk := sendRequestWorker(t, f, info, params)
+	f.baselineState(t)
+	if resMsg == nil {
+		t.Fatalf(prefix+"Nil response received for %s request", info.Method)
+	}
```
jakebailey, line 469: Thanks for cleaning this all up, I hated the duplication

internal/ls/symbols.go
```diff
@@ -240,17 +240,17 @@ type DeclarationInfo struct {
...
 			if !sourceFile.IsDeclarationFile {
-				sourceFiles.Add(sourceFile)
+				sourceFiles[sourceFile.Path()] = sourceFile
 			}
 		}
 	}
 	// Create DeclarationInfos for all declarations in the source files.
 	var infos []DeclarationInfo
-	for sourceFile := range sourceFiles.Keys() {
+	for _, sourceFile := range sourceFiles {
```
jakebailey, line 253: Does this end up causing the infos to be randomly ordered? Or do we sort this somewhere else?

jakebailey:
I just tested this out on TypeScript. This is unbelievably fast.

Cold, only `debug.ts` open, FAR on `Debug` itself does a FAR in 414ms. Repeating it again, hot, 180ms. Wowza.

jakebailey:
Compare to tsserver... which does it cold in more than 4 seconds, and hot in something like 2 seconds.

10x!

internal/ls/symbols.go
sheetalkamat: They were never ordered. (collections.Set is map[key]struct{} The declarations are ordered using some sort of declarationInfo comparison and used. https://github.com/microsoft/typescript-go/pull/1991/files/98fe20eee9ad3d03e247b8f2d053688cb53ab93a#diff-5ddd4836d96fb50708a55022fee6f2e9de817b1cc54b846fd126f7e1a486d4aaL268

internal/fourslash/baselineutil.go
sheetalkamat: Yeah .. lets do that as separate PR?

internal/ls/symbols.go
jakebailey: Ah, duh!

internal/fourslash/baselineutil.go
jakebailey: https://github.com/microsoft/typescript-go/pull/2148 😄

PR #1999, opened 2025-11-01
Only export `@typedef` type aliases in modules

### LINKED ISSUE #1988: Reparsed `@typedef` turns into `export type` even in script files
This is an extension of https://github.com/microsoft/typescript-go/issues/1685

Given this test case:

```ts
// @module: nodenext
// @target: esnext
// @outDir: ./out
// @allowJs: true
// @checkJs: true
// @noUnusedLocals: true
// @noUnusedParameters: true
// @declaration: true
// @moduleDetection: legacy

// @filename: test.js
/** @type {MyType} */
const obj = { a: 42, b: "hello" };
console.log(obj);

// @filename: types.js
/** @typedef {{ a: number, b: string }} MyType */
```

These files are not modules, they are scripts (per `moduleDetction: legacy`, but `auto` also works). The `typedef` is _not_ an export, it's declaring a global type the other file can reference.

```ts
//// [test.d.ts]
/** @type {MyType} */
declare const obj: MyType;
//// [types.d.ts]
type MyType = {
    a: number;
    b: string;
};
```

But, in Corsa, the reparser turns `typedef` into `export type`, leading to a different output:

```ts
//// [test.d.ts]
/** @type {MyType} */
declare const obj: MyType;
//// [types.d.ts]
export type MyType = {
    a: number;
    b: string;
};
/** @typedef {{ a: number, b: string }} MyType */
```

And an error:

```
test.js(1,12): error TS2304: Cannot find name 'MyType'.


==== test.js (1 errors) ====
    /** @type {MyType} */
               ~~~~~~
!!! error TS2304: Cannot find name 'MyType'.
    const obj = { a: 42, b: "hello" };
    console.log(obj);
    
==== types.js (0 errors) ====
    /** @typedef {{ a: number, b: string }} MyType */
```

This is all a problem, because the reparser does this transformation _before_ we determine if a file is a module or not, and because it changes the declaration emit.

Strada avoids this by not doing anything at all with JSDoc comments, until special cases in the checker, declaraion emit, etc.

This looks like a situation similar to the one that required us to introduce `ExportAssignment` vs `JSExportAssignment`.

But, I am looking, and we _already have_ `KindJSTypeAliasDeclaration`. So this may actually just come down to us doing module detection wrong in the face of these? Or something? Some other bug?

ahejlsberg:
In this PR:

* Only export `@typedef` type aliases when they occur in modules (i.e. don't export in script files).
* Mark reparsed export assignment nodes as a TypeScript construct such that they're omitted from emitted .js.
* Fix LSP crash by adding missing case for `ast.KindJSTypeAliasDeclaration`.

Most of the baseline changes result from not marking script files as modules and no longer including invalid `export =` in emitted .js files.

Fixes #1988.


testdata/baselines/reference/fourslash/findAllReferences/findAllRefsTypedef_importType.baseline.jsonc
```diff
@@ -12,10 +12,6 @@
 // /** @typedef {number} /*FIND ALL REFS*/[|Foo|] */
 // const dummy = 0;
 
-// === /b.js ===
-// /** @type {import('./a').[|Foo|]} */
-// const x = 0;
-
```
jakebailey, line 18: This seems like a regression, maybe? `a.js` theoretically is a module because it has `module.exports` in it, I think, but now we are losing that export and therefore the reference.

testdata/baselines/reference/submodule/conformance/commonJSAliasedExport.js.diff
```diff
@@ -10,13 +10,12 @@
  function funky(declaration) {
      return false;
  }
-+export = donkey;
  module.exports = donkey;
 +export var funky = funky;
```
jakebailey, line 14: Not your bug but definitely an interesting one.

testdata/baselines/reference/fourslash/findAllReferences/findAllRefsTypedef_importType.baseline.jsonc
ahejlsberg: Yeah, that's an oversight. We need to handle the implied export of `@typedef` in find-all-references.

testdata/baselines/reference/submodule/conformance/commonJSAliasedExport.js.diff
ahejlsberg: This is actually a bug fix I made part of this PR. We now mark reparsed export assignment nodes as a TypeScript construct such that they're properly omitted from the emitted .js.

jakebailey: Yes, right, I was just commenting on the remaining diff in the file.

internal/parser/reparser.go
jakebailey:
```suggestion
```

testdata/baselines/reference/submodule/conformance/commonJSAliasedExport.js.diff
ahejlsberg: Got it. Definitely funky!

internal/parser/reparser.go
ahejlsberg: Done!

internal/transformers/declarations/transform.go
```diff
@@ -1582,6 +1582,9 @@ func (tx *DeclarationTransformer) ensureModifierFlags(node *ast.Node) ast.Modifi
 		mask ^= ast.ModifierFlagsAmbient
 		additions = ast.ModifierFlagsNone
 	}
+	if ast.IsJSTypeAliasDeclaration(node) && parentIsFile && ast.IsExternalOrCommonJSModule(node.Parent.AsSourceFile()) {
+		additions |= ast.ModifierFlagsExport
+	}
 	return maskModifierFlags(tx.host, node, mask, additions)
 }
```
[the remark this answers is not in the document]
jakebailey: Is this worth addressing if we are actually copy and pasting this code?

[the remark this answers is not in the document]
ahejlsberg: Yup, done.

PR #2007, opened 2025-11-03
Fix unused identifier diags, LSP tag diags

### LINKED ISSUE #2001: Unused declaration behavior differs from tsserver
Given this code:

```ts
function unused() {}
```

In tsserver without `noUnusedLocals`, this is grayed out:

Image

With `noUnusedLocals`, it's grayed out and also gets a warning squiggle:

Image

But, with tsgo's LSP implementation, these differ.

Without `noUnusedLocals`, there are no diagnostics at all, no graying:

Image

```
[Trace - 10:37:58 PM] Received response 'textDocument/diagnostic - (14)' in 4ms.
Result: {
    "kind": "full",
    "items": []
}
```

With `noUnusedLocals`, it gets an error-level squiggle:

Image

```
[Trace - 10:37:16 PM] Received response 'textDocument/diagnostic - (13)' in 8ms.
Result: {
    "kind": "full",
    "items": [
        {
            "range": {
                "start": {
                    "line": 0,
                    "character": 9
                },
                "end": {
                    "line": 0,
                    "character": 15
                }
            },
            "severity": 1,
            "code": 6133,
            "source": "ts",
            "message": "'unused' is declared but its value is never read."
        }
    ]
}
```

We are clearly missing diagnostics and the correct info on diagnostics when we have them.

jakebailey:
Fixes #2001 (mostly)

Issues:

- The checker's ported code was for the _non-suggestion_ side of things, meaning it entirely skipped these diags when the `noUnusedLocals`/`noUnusedParameters` was off, so the grayed out suggestion errors were never even calculated.
- `reportUnused` didn't actually use `suggestionDiagnostics` at all, nor have the bit that sets the category to `Suggestion` on the fly; I re-ported that code.
- The LSP code for checking client capabilities for diagnostics was "correct", except apparently the VS Code LSP client does not send the full set of capabilities for _both_ push and pull diagnostics (specifically, `tag support`). Pull the info from `publishDiagnostics` as a backup, which is a hack but works around the problem until we can get v10 of the LSP client.
  ```
  "publishDiagnostics": {
      "relatedInformation": true,
      "versionSupport": false,
      "tagSupport": {
          "valueSet": [
              1,
              2
          ]
      },
      "codeDescriptionSupport": true,
      "dataSupport": true
  },
  "diagnostic": {
      "dynamicRegistration": true,
      "relatedDocumentSupport": false
  }
  ```

The warning vs error thing is actually a VS Code setting, something we will have to do once those are passed in and handled.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -2130,7 +2130,7 @@ func (c *Checker) checkSourceFile(ctx context.Context, sourceFile *ast.SourceFil
 		}
 		if ctx.Err() == nil {
 			// This relies on the results of other lazy diagnostics, so must be computed after them
-			if !sourceFile.IsDeclarationFile {
+			if !sourceFile.IsDeclarationFile && (c.compilerOptions.NoUnusedLocals.IsTrue() || c.compilerOptions.NoUnusedParameters.IsTrue()) {
 				c.checkUnusedIdentifiers(links.identifierCheckNodes)
 			}
 			if !sourceFile.IsDeclarationFile {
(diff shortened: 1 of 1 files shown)


internal/ls/diagnostics.go
```diff
@@ -73,13 +73,14 @@ func (l *LanguageService) toLSPDiagnostic(clientOptions *lsproto.DiagnosticClien
 		}
 	}
 
+	// We do not check client capabilities for tags; the LSP spec says clients must handle unknown tags.
```
jakebailey: I regressed this (kinda) in #1980.

internal/checker/checker.go
```diff
@@ -2130,7 +2130,7 @@ func (c *Checker) checkSourceFile(ctx context.Context, sourceFile *ast.SourceFil
 		}
 		if ctx.Err() == nil {
 			// This relies on the results of other lazy diagnostics, so must be computed after them
-			if !sourceFile.IsDeclarationFile && (c.compilerOptions.NoUnusedLocals.IsTrue() || c.compilerOptions.NoUnusedParameters.IsTrue()) {
+			if !sourceFile.IsDeclarationFile {
```
jakebailey: The old code did this conditionally by having extra code in `GetSuggestionDiagnostics`, which would technically save CPU/memory for batch compiles where we never expose suggestion diags.

I can attempt to do this, though there's a bit more goofiness there given that asking for diagnostics is currently side-effect free.

internal/ls/diagnostics.go
jakebailey, line 64: Would you believe that the client is not sending this for pull diags either? Seems like I have to fix another thing here :(

jakebailey: Addressed in the latest commit. Upgrading the client to v10 fixies this but is a harder change I need to send in another PR.

sheetalkamat:
are there any tests. i dont see any change in tests

jakebailey:
Unfortunately not, the current fourslash runner does not do diagnostics yet and that's a bigger ask I can tackle separately. I have tested this in the editor, however.

internal/checker/checker.go
jakebailey: Now fixed.

jakebailey:
Okay, this is now a pure port and should work fine.

sheetalkamat (approved):

PR #2013, opened 2025-11-04
Detect Windows junctions with GetFileAttributesEx

jakebailey:
Faster than doing a realpath + comparison.

```
goos: windows
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/vfs/osvfs
                                    │   out.txt   │
                                    │   sec/op    │
Realpath/target-20                    76.62µ ± 1%
Realpath/link-20                      87.68µ ± 2%
IsSymlinkOrJunction/regular_file-20   28.60µ ± 2%
IsSymlinkOrJunction/junction-20       22.62µ ± 2%
IsSymlinkOrJunction/nonexistent-20    11.45µ ± 0%
geomean                               34.63µ

                                    │  out.txt   │
                                    │    B/op    │
Realpath/target-20                    944.0 ± 0%
Realpath/link-20                      944.0 ± 0%
IsSymlinkOrJunction/regular_file-20   192.0 ± 0%
IsSymlinkOrJunction/junction-20       176.0 ± 0%
IsSymlinkOrJunction/nonexistent-20    192.0 ± 0%
geomean                               356.8

                                    │  out.txt   │
                                    │ allocs/op  │
Realpath/target-20                    8.000 ± 0%
Realpath/link-20                      8.000 ± 0%
IsSymlinkOrJunction/regular_file-20   1.000 ± 0%
IsSymlinkOrJunction/junction-20       1.000 ± 0%
IsSymlinkOrJunction/nonexistent-20    1.000 ± 0%
geomean                               2.297
```

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -20,6 +20,7 @@ runs:
       id: install-go
       shell: pwsh
       run: |
+        # https://github.com/microsoft/go-infra/blob/main/goinstallscript/powershell/go-install.ps1
         ${{ github.action_path }}/go-install.ps1 -Version ${{ inputs.go-version }} -GitHubActionsPath
 
         $goVersionOutput = go version
--- a/internal/compiler/checkerpool.go
+++ b/internal/compiler/checkerpool.go
@@ -12,10 +12,11 @@ import (
 )
 
 type CheckerPool interface {
+	Count() int
 	GetChecker(ctx context.Context) (*checker.Checker, func())
 	GetCheckerForFile(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func())
 	GetCheckerForFileExclusive(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func())
-	GetAllCheckers(ctx context.Context) ([]*checker.Checker, func())
+	ForEachCheckerParallel(ctx context.Context, cb func(idx int, c *checker.Checker))
 	Files(checker *checker.Checker) iter.Seq[*ast.SourceFile]
 }
 
@@ -40,6 +41,10 @@ func newCheckerPool(checkerCount int, program *Program) *checkerPool {
 	}
 
 	return pool
+}
+
+func (p *checkerPool) Count() int {
+	return p.checkerCount
 }
 
 func (p *checkerPool) GetCheckerForFile(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func()) {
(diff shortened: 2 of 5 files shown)


internal/vfs/osvfs/reparsepoint_windows.go
```diff
@@ -0,0 +1,27 @@
+package osvfs
```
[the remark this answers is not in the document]
jakebailey: Incorrect; `_windows` in the filename suffices, and `realpath_windows.go` does not use `//go:build`, so this is a lie

DanielRosenwasser:
Any specific differences you've seen in perf?

jakebailey:
Haven't tested yet, just cleaning up some TODOs with my name on it. On any project I have it probably won't be very noticable...

internal/vfs/internal/internal.go
```diff
@@ -93,12 +93,12 @@ func (vfs *Common) GetAccessibleEntries(path string) (result vfs.Entries) {
...
-			// Could be a Windows junction. Try Realpath.
-			// TODO(jakebailey): use syscall.Win32FileAttributeData instead
+		if entryType&fs.ModeIrregular != 0 && vfs.IsSymlinkOrJunction != nil {
+			// Could be a Windows junction or other reparse point.
+			// Check using the OS-specific helper.
 			fullPath := path + "/" + entry.Name()
-			if realpath := vfs.Realpath(fullPath); fullPath != realpath {
-				if stat := vfs.Stat(realpath); stat != nil {
+			if vfs.IsSymlinkOrJunction(fullPath) {
+				if stat := vfs.Stat(fullPath); stat != nil {
```
jakebailey, line 101: If something is a junction, maybe we don't even need to stat it and just say that it's a dir? (I don't believe junctions/reparse points can be files, in which case I could rename funcs too)

internal/vfs/osvfs/reparsepoint_windows.go
```diff
@@ -0,0 +1,27 @@
...
+// isSymlinkOrJunction checks if the given path is a symlink or junction point
+// on Windows by checking the FILE_ATTRIBUTE_REPARSE_POINT attribute.
+func isSymlinkOrJunction(path string) bool {
+	pathUTF16, err := syscall.UTF16PtrFromString(path)
+	if err != nil {
+		return false
+	}
+
+	var data syscall.Win32FileAttributeData
+	err = syscall.GetFileAttributesEx(
```
RyanCavanaugh, line 19: We might want to prepend `\\?\` to automatically handle long filepaths? https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileattributesexa

jakebailey: Ah, crap, converting those is a whole thing that is hard and not available to us outside stdlib. I'd need to do something like I did in realpath

```go
func realpath(path string) (string, error) {
	var h windows.Handle
	if len(path) < 248 {
		var err error
		h, err = openMetadata(path)
		if err != nil {
			return "", err
		}
		defer windows.CloseHandle(h) //nolint:errcheck
	} else {
		// For long paths, defer to os.Open to run the path through fixLongPath.
		f, err := os.Open(path)
		if err != nil {
			return "", err
		}
		defer f.Close()

		// Works on directories too since https://go.dev/cl/405275.
		h = windows.Handle(f.Fd())
	}
```

Guess I'll wait on things until I figure that out.

DanielRosenwasser (approved):

internal/vfs/osvfs/reparsepoint_windows.go
RyanCavanaugh: I wouldn't block this on it

jakebailey: It's pretty easy to hit this on the use cases where it matters, e.g. pnpm ones, so I should definitely check it in afraid

internal/vfs/internal/internal.go
jakebailey: `FILE_ATTRIBUTE_REPARSE_POINT` will be set on even symlinks, so I have to keep statting. Maybe this means the func should be renamed to `isReparsePoint`.

jakebailey:
To actually test this, I switched us over to the Microsoft fork of Go, which disables the SEH bit that auto-handles longpaths. I can split that out into its own PR for clarity if desired.

jakebailey:
Okay, this PR is now solely the junctions change.

RyanCavanaugh (approved):

PR #2016, opened 2025-11-04
fix(2015): abstract property created, overshadowing override

### LINKED ISSUE #2015: abstract property created, overshadowing override
## Steps to reproduce

```
abstract class Base {
	abstract value: number
}

class MyClass extends Base {
	get value() {
		return 123
	}
}

const obj = new MyClass()
console.log(obj.value)
```

The abstract property "value" is not created in "Base":

```
"use strict";
class Base {
}
class MyClass extends Base {
    get value() {
        return 123;
    }
}
const obj = new MyClass();
console.log(obj.value);
//# sourceMappingURL=transpileme.js.map
```

## Behavior with `tsgo`

The abstract property "value" **is created** in "Base":

```
"use strict";
class Base {
    value;
}
class MyClass extends Base {
    get value() {
        return 123;
    }
}
const obj = new MyClass();
console.log(obj.value);
//# sourceMappingURL=transpileme.js.map
```

When running the two transpiled files, I see that the "tsc version" emits "123" but the "tsgo version" emits "undefined".

Possibly related: https://github.com/microsoft/TypeScript/issues/40632

a-tarasyuk:
Fixes #2015

--- 

This PR adds the missing [ModifierFlagsAbstract](https://github.com/microsoft/TypeScript/blob/6f4fb0145845db22791188c4d38e3a1a0edfd449/src/compiler/transformers/ts.ts#L1298) to eliminate property declarations

```ts
const isAmbient = node.flags & NodeFlags.Ambient || hasSyntacticModifier(node, ModifierFlags.Abstract);

```


birgersp:
🤯 You guys are fast 🚀

PR #1636, opened 2025-08-26
lsp: enable custom configs to be passed into tsgo LSP

johnnycheng0210:
## Context
Address the feat request in this [ticket](https://github.com/microsoft/typescript-go/issues/945).

Issue #945, opened 2025-05-26
[feat][lsp] add support for custom config file name(s)


mjames-c:
The `tsgo` LSP is currently hardcoded to look for `tsconfig.json` and `jsconfig.json` files for figuring out project configuration for a given source file (similar to TSv5). My request is to make the config file names the LSP searches for configurable (e.g. a vscode option like `typescript.native-preview.configNames: string[]`)

This would allow my company (Canva) to experiment and incrementally rollout different project configurations to our 600+ frontend engineers.

johnnycheng0210:
Have a proposed feature change [here](https://github.com/microsoft/typescript-go/pull/1636) @jakebailey not sure what the flow is to request reviewers.

The [contributing](https://github.com/microsoft/typescript-go?tab=contributing-ov-file) doc isn't extremely clear with instructions 😅

jakebailey:
This was done in #2020

PR #2020, opened 2025-11-05
Support for custom configurations

johnnycheng0210:
Follow up more up to date rebase of this https://github.com/microsoft/typescript-go/pull/1636.

Address the feat request in this [ticket](https://github.com/microsoft/typescript-go/issues/945), by allowing custom configs to be passed to the tsgo LSP.

## Testing

- Load `.ts` file
- `findAllReferences` -> returns correct amount of references for this singular project
- update `"typescript.native-preview.customConfigFileName": "tsconfig.all.json"` to `solution` style project config -> returns all references in the entire repo (correct behavior)
- `tsconfig.all.json` is essentially `"include": ["./**/*"]` at root of our project.

### DIFF
--- a/internal/ls/lsutil/userpreferences.go
+++ b/internal/ls/lsutil/userpreferences.go
@@ -735,7 +735,13 @@ func (p *UserPreferences) Set(name string, value any) bool {
 	case "implementationscodelensshowonallclassmethods":
 		p.CodeLens.ImplementationsCodeLensShowOnAllClassMethods = parseBoolWithDefault(value, false)
 	case "customconfigfilename":
-		p.CustomConfigFileName = strings.TrimSpace(tsoptions.ParseString(value))
+		name := strings.TrimSpace(tsoptions.ParseString(value))
+		// Validate that the custom config file name is a plain base file name
+		// (no path separators or ".." segments) to prevent path traversal.
+		if name != "" && (strings.ContainsAny(name, "/\\") || name == ".." || name == ".") {
+			name = ""
+		}
+		p.CustomConfigFileName = name
 	default:
 		if p.FormatCodeSettings == nil {
 			p.FormatCodeSettings = GetDefaultFormatCodeSettings()
--- a/internal/project/customconfigfilename_test.go
+++ b/internal/project/customconfigfilename_test.go
@@ -152,4 +152,74 @@ func TestCustomConfigFileName(t *testing.T) {
 			"expected RefreshDiagnostics to be called after customConfigFileName change, got %d calls (baseline %d)",
 			refreshCount, baselineRefreshCount)
 	})
+
+	t.Run("rejects path traversal in customConfigFileName", func(t *testing.T) {
+		t.Parallel()
+		for _, invalidName := range []string{
+			"/etc/passwd",
+			"../tsconfig.json",
+			"configs/tsconfig.all.json",
+			"..\\tsconfig.json",
+			"sub\\dir\\tsconfig.json",
+			"..",
+			".",
+		} {
+			prefs := lsutil.NewDefaultUserPreferences()
+			prefs.Set("customConfigFileName", invalidName)
+			assert.Equal(t, prefs.CustomConfigFileName, "",
+				"expected customConfigFileName to be cleared for invalid value %q", invalidName)
+		}
+	})
+
+	t.Run("accepts plain base file names in customConfigFileName", func(t *testing.T) {
+		t.Parallel()
+		for _, validName := range []string{
+			"tsconfig.all.json",
+			"tsconfig.editor.json",
+			"jsconfig.custom.json",
+		} {
+			prefs := lsutil.NewDefaultUserPreferences()
+			prefs.Set("customConfigFileName", validName)
+			assert.Equal(t, prefs.CustomConfigFileName, validName,
+				"expected customConfigFileName to be %q", validName)
+		}
+	})
+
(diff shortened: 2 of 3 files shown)


_extension/src/client.ts
```diff
@@ -90,10 +90,11 @@ export class Client {
         this.exe = exe;
         this.outputChannel.appendLine(`Resolved to ${this.exe.path}`);
 
-        // Get pprofDir
+		// Get pprofDir
```
DanielRosenwasser: Switch this and every other TS line back to spaces

_extension/src/client.ts
```diff
@@ -108,12 +109,17 @@ export class Client {
...
-            this.clientOptions,
-        );
+		this.client = new LanguageClient(
+			"typescript.native-preview",
+			"typescript.native-preview-lsp",
+			serverOptions,
+			{
+				...this.clientOptions,
+				initializationOptions: {
+					customConfigFileName,
```
DanielRosenwasser: Rather than be part of initialization, we think this sort of thing would belong more on user preferences ([see here](https://github.com/microsoft/typescript-go/pull/1729).

Whether we just say "you need to restart the server" or we track this sort of thing on the server in some way is a bigger question.

DanielRosenwasser (dismissed):
We haven't discussed this more broadly, but custom config specification has been a long-standing request with a lot of interest. The scope of implementing could grow a bit from what's here if we need something beyond a naive restart, but maybe this is good to start. I'll let others weigh in.

jakebailey:
I'm not sure what happened with your PRs, but can you keep the one you intend us to look at open and keep pushing to that one?

johnnycheng0210:
ah sorry - i was rebasing PR's - i will close the previous PR and use this one instead

_extension/src/client.ts
johnnycheng0210: hey @DanielRosenwasser i've just made the changes to have it be included in user preferences and have tested it works - could you take another look please

jakebailey:
One thing this might be missing is some handling of on-change events; `(*Session).Configure` handles things; maybe `pendingConfigChanges` is sufficient to catch this?

_extension/package.json
```diff
@@ -31,6 +31,12 @@
             {
                 "title": "TypeScript Native Preview",
                 "properties": {
+                    "typescript.native-preview.customConfigFileName": {
+                        "type": "string",
+                        "default": "",
+                        "description": "Custom config file name to use, before defaulting to tsconfig.json/jsconfig.json.",
```
jakebailey: An example value would be helpful here, like `tsconfig.editor.json`, just to show that we're looking for a _basename_, not a full path.

jakebailey:
I tried this out and it seems to work. What I did to test it was to create `oops.ts` with:

```ts
declare const x: string | undefined;

const y: string = x;
```

Which fails normally as the default inferred project has strict=true.

Then I added `tsconfig.oops.json`:

```json
{
    "compilerOptions": {
        "strict": false
    },
    "files": [
        "oops.ts"
    ]
}
```

Then configured `tsconfig.oops.json` in the config. The diags refreshed and the error went away, so it did pick up the config.

Undoing the config didn't do anything, however, and it took a restart to get the diags to differ (even after modifying).

I'm not sure if that's good or bad, but it does seems suspicious.

johnnycheng0210:
@jakebailey i did observe similar behavior with `findAllReferences` for example, when going from single project config -> singular solution project config -> single project config. The results after singular solution project -> single project config, didn't change. I thought it was to do with findAllReferences implementation but it appears its wider than that.

any clue why this might be happening? seems odd that one update works, but the subsequent one seems to remember the old ones config.

johnnycheng0210:
I do wonder if a more naive approach, of requiring a window reload between custom config file name changes in user preferences is preferred. less likely to run into these edge cases, and still feels reasonably ergonomic?

jakebailey:
I'm definitely confused, because when I write a test for this, the test seemingly does the right thing.

johnnycheng0210:
hey @jakebailey any clue on this one? i've taken another look and am pretty confused too, it feels like its unrelated to the new config file name being updated, but moreso on the state of the language server when config files refresh? 🙏

jakebailey:
I honestly have no idea ☹️

johnnycheng0210:
> I do wonder if a more naive approach, of requiring a window reload between custom config file name changes in user preferences is preferred. less likely to run into these edge cases, and still feels reasonably ergonomic?

@jakebailey what do you think about this approach? when using this approach, we don't run into the above described edge case, albeit does feel less ergonomic.

jakebailey:
I don't want to do that; we do not have anything that requires a client restart or a full window reload. I think we can figure out what's going on here. (I just have not personally had the chance to. Maybe @andrewbranch will know off the top of his head.)

andrewbranch:
I'm not sure off the top of my head, I'd have to debug and step through.

sspencer-canva:
This has been on hold for 3 weeks. It's a high-priority piece for us, and a number of engineers across multiple spaces are super keen to get it working here at Canva. Is there anything we can do to help? 🙏

jakebailey:
This PR needs a merge from main at least (the snapshot code has changed a bit), but I still have not been able to explain the behavior I saw in testing. I'm trying to write some sort of test to show it.

jakebailey:
I think I have a test and a fix. Will push it to the PR so we can get this moving.

PR #2023, opened 2025-11-05
Fix crash in find-all-refs on `exports.xxx` in .js file

ahejlsberg:
This PR fixes a crash that occurs when doing find-all-refs in a .js file:

```js
exports.a = 1; // Find-all-refs on the exports identifier
```

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -1172,7 +1172,7 @@ func (l *LanguageService) getReferencedSymbolsForModule(ctx context.Context, pro
 			references: references,
 		}}
 	}
-	return nil
+	return []*SymbolAndEntries{}
 }
 
 func getReferenceAtPosition(sourceFile *ast.SourceFile, position int, program *compiler.Program) *refInfo {


internal/ls/findallreferences.go
```diff
@@ -1172,7 +1172,7 @@ func (l *LanguageService) getReferencedSymbolsForModule(ctx context.Context, pro
 			references: references,
 		}}
 	}
-	return nil
+	return []*SymbolAndEntries{}
```
jakebailey, line 1175: What calling code depends on the difference? 😞

ahejlsberg: Two places in `findallreferences.go`. They specifically check for `nil` and branch accordingly. Ugly, but basically the same behavior with had in Strada (i.e. `undefined` vs possibly empty array).

ahejlsberg: Actually, only one place depends on seeing `nil`:

https://github.com/microsoft/typescript-go/blob/3d247ef578b7291790fe059fc06a1ae9eaad9187/internal/ls/findallreferences.go#L732

jakebailey (approved):

PR #2024, opened 2025-11-05
Fix losing options from command line in watch mode

### LINKED ISSUE #1645: --noEmit does not work properly in watchMode
`tsgo --noEmit` doesn't emit files.
`tsgo --noEmit -w` emit files.

haoqixu:
`watcher` loses options from command line after reloading options from the config file.

Fixes #1645.


jakebailey (approved):
This seems right at a glance; @sheetalkamat does this seem right to you?

PR #2026, opened 2025-11-05
Implement regexp literal syntax checking

jakebailey:
This PR implements regexp literal syntax checking. The main difference is that it is no longer a scanning step, as that would make it a parser error, which would mean we would have to go back to having separate ASTs depending on the language version being used (something we worked so hard to stop doing).

Instead, the regex checks are grammar checks that happen during checking. Additionally, the code doesn't live in the scanner package at all, as it is entirely self-contained (much to the original code's credit).

This is a really complicated piece of code largely due to the sheer scale of these checks, but also the annoying edge cases due to the old compiler's reliance of TS being in JS already and therefore getting identical string behavior when decoding. This required just a load of fiddling to get working.

The bulk of this is copilot ported, with a bunch of cleanups and fixes. The result is that every regexp related test matches, except for `regularExpressionCharacterClassRangeOrder.errors.txt.diff` and `regularExpressionWithNonBMPFlags.errors.txt.diff`, which I believe are _improving_ because of better handling of positioning in weird characters, e.g.:

```diff
     // See https://en.wikipedia.org/wiki/Mathematical_Alphanumeric_Symbols
     const regexes: RegExp[] = [
     	/[𝘈-𝘡][𝘡-𝘈]/,
-    	   ~~~
+    	  ~~~
 !!! error TS1517: Range out of order in character class.
-    	          ~~~
+    	       ~~~
 !!! error TS1517: Range out of order in character class.
     	/[𝘈-𝘡][𝘡-𝘈]/u,
-    	         ~~~~~
+    	       ~~~
 !!! error TS1517: Range out of order in character class.
     	/[𝘈-𝘡][𝘡-𝘈]/v,
-    	         ~~~~~
+    	       ~~~
 !!! error TS1517: Range out of order in character class.
     
     	/[\u{1D608}-\u{1D621}][\u{1D621}-\u{1D608}]/,
```

But probably aren't really, it's just that the offsets are UTF-8 now and these are weird chars. tsserver in the editor shows similar ranges so this was probably just a bug in the old baseline squiggler.

### DIFF
--- a/internal/regexpchecker/regexpchecker.go
+++ b/internal/regexpchecker/regexpchecker.go
@@ -717,9 +717,18 @@ func (v *regExpValidator) scanEscapeSequence(atomEscape bool) string {
 		}
 		return string(ch)
 
-	case 'b', 't', 'n', 'v', 'f', 'r':
-		// Standard escape sequences
-		return string(ch)
+	case 'b':
+		return "\b"
+	case 't':
+		return "\t"
+	case 'n':
+		return "\n"
+	case 'v':
+		return "\v"
+	case 'f':
+		return "\f"
+	case 'r':
+		return "\r"
 
 	case 'x':
 		// Hex escape '\xDD'
@@ -1126,8 +1135,9 @@ func (v *regExpValidator) scanClassSetExpression() {
 				}
 				secondStart := v.pos
 				secondOperand := v.scanClassSetOperand()
-				// Don't report TS1518 for the second operand of a range
-				// expressionMayContainStrings tracking is still needed
+				if isCharacterComplement && v.mayContainStrings {
+					v.error(diagnostics.Anything_that_would_possibly_match_more_than_a_single_character_is_invalid_inside_a_negated_character_class, secondStart, v.pos-secondStart)
+				}
 				expressionMayContainStrings = expressionMayContainStrings || v.mayContainStrings
 				if secondOperand == "" {
 					v.error(diagnostics.A_character_class_range_must_not_be_bounded_by_another_character_class, secondStart, v.pos-secondStart)


jakebailey:
Based on the little differences I've noticed, the testing of this feature is not entirely complete...

jakebailey:
@graphemecluster I'm not sure if you are around and willing to look at this, but since you are looking at regex stuff again maybe you're willing to look at this again.

Not 100% confident in it, and I don't like the duplication due to the fact that we need some weird scanning stuff to deal with things the new scanner does not much care about...

internal/regexpchecker/tables.go
```diff
@@ -0,0 +1,560 @@
+package regexpchecker
+
+import "github.com/microsoft/typescript-go/internal/collections"
+
+// Table 67: Binary Unicode property aliases and their canonical property names
+// https://tc39.es/ecma262/#table-binary-unicode-properties
+var binaryUnicodeProperties = collections.NewSetFromItems(
```
DanielRosenwasser, line 7: It's annoying that we can't just use [`unicode.Categories`](https://pkg.go.dev/unicode#pkg-variables) for this because of differences like `Bidi_Control` *and* `Bidi_C` existing.

internal/regexpchecker/utf16.go
```diff
@@ -0,0 +1,134 @@
...
+	if n := utf16.RuneLen(ch); n > 0 {
+		return n
+	}
+	return 1 // fallback for invalid runes
+}
+
+// utf16Length returns the UTF-16 length of a string, matching JavaScript's string.length.
+// This counts UTF-16 code units, where surrogate pairs count as 2 units.
+// Handles both UTF-8 encoded strings and special 2-byte UTF-16BE surrogate encodings.
+func utf16Length(s string) int {
```
DanielRosenwasser: We must have this elsewhere, no?

DanielRosenwasser: And if not, maybe we should be using this everywhere else.

graphemecluster:
Let me see if I can get some time this weekend :D

internal/regexpchecker/regexpchecker.go
```diff
@@ -0,0 +1,1351 @@
...
+				// Check if both are "complete" characters in JavaScript's UTF-16 representation.
+				// A character is complete if its UTF-16 length matches the expected size.
+				// In JavaScript, string.length returns the UTF-16 code unit count.
+				minUTF16Length := utf16Length(atom)
+				maxUTF16Length := utf16Length(rangeEnd)
+
+				minIsComplete := minUTF16Length == minExpectedSize
+				maxIsComplete := maxUTF16Length == maxExpectedSize
+
+				if minIsComplete && maxIsComplete && minCodePoint > maxCodePoint {
```
graphemecluster, line 984: The original intention is merely to determine whether an atom has only a single character (codepoint) for a valid order comparison. I believe `utf8.RuneCountInString(atom) == 1` is already sufficient (and is actually an equivalent regardless of Unicode mode, since it is impossible to get a non-BMP atom in non-Unicode mode), but I might be wrong. (In the original TypeScript code, I could have written `[...atom].length === 1` – comparing by `charSize` is purely for avoiding array allocation.)

graphemecluster: And this is always equivalent to `r, n := utf8.DecodeRuneInString(s); len(s) == n`, so you can just do what you did in `scanClassSetExpression`.

internal/regexpchecker/regexpchecker.go
```diff
@@ -0,0 +1,1351 @@
...
+	}
+}
+
+// scanSourceCharacter scans and returns a single "character" from the source.
+// In Unicode mode (u or v flags), returns complete Unicode code points.
+// In non-Unicode mode, mimics JavaScript's UTF-16 behavior where literal characters
+// >= U+10000 are treated as surrogate pairs and consumed across two sequential calls.
+func (v *regExpValidator) scanSourceCharacter() string {
+	// Check if we have a pending low surrogate from the previous call
+	if v.surrogateState != nil {
```
graphemecluster, line 900: Just for safety, I would suggest checking the expected position too (by adding a property to the state)

internal/checker/grammarchecks.go
```diff
@@ -54,43 +55,30 @@ func (c *Checker) grammarErrorOnNodeSkippedOnNoEmit(node *ast.Node, message *dia
...
+		var lastError *ast.Diagnostic
+
+		nodeStart := scanner.GetTokenPosOfNode(node.AsNode(), sourceFile, false)
+		onError := func(message *diagnostics.Message, start int, length int, args ...any) {
+			start += nodeStart
+
+			if message.Category() == diagnostics.CategoryMessage && lastError != nil &&
+				start == lastError.Pos() && length == lastError.Len() {
+				err := ast.NewDiagnostic(nil, core.NewTextRange(start, start+length), message, args...)
+				lastError.AddRelatedInfo(err)
```
graphemecluster, line 70: This was kind of workaround since I couldn't expand scanner effectively in Strada. But now that `regExpValidator` is a standalone `struct`, I would suggest adding a dedicated method for related info (spelling suggestions).

internal/regexpchecker/utf16.go
```diff
@@ -0,0 +1,134 @@
...
+	utf16Length int
+}
+
+// makeRegExpChar creates a regExpChar from a code point
+func makeRegExpChar(codePoint rune) regExpChar {
+	return regExpChar{
+		codePoint:   codePoint,
+		utf16Length: charSize(codePoint),
+	}
+}
```
graphemecluster, line 134: I don't think they are referenced anywhere, do you intend to use them at a later time?

internal/regexpchecker/regexpchecker.go
```diff
@@ -0,0 +1,1351 @@
...
+				v.mayContainStrings = true
+			}
+		} else if !isValidUnicodePropertyName(propertyNameOrValue) {
+			v.error(diagnostics.Unknown_Unicode_property_name_or_value, propertyNameOrValueStart, v.pos-propertyNameOrValueStart)
+			// Provide spelling suggestion from general category values, binary properties, and binary properties of strings
+			candidates := make([]string, 0, generalCategoryValues.Len()+binaryUnicodeProperties.Len()+binaryUnicodePropertiesOfStrings.Len())
+			candidates = slices.AppendSeq(candidates, maps.Keys(generalCategoryValues.M))
+			candidates = slices.AppendSeq(candidates, maps.Keys(binaryUnicodeProperties.M))
+			candidates = slices.AppendSeq(candidates, maps.Keys(binaryUnicodePropertiesOfStrings.M))
+			suggestion := core.GetSpellingSuggestion(propertyNameOrValue, candidates, core.Identity)
```
graphemecluster, line 603: Did you consider making `GetSpellingSuggestion` accept an iterator to avoid allocation? (I am not particularly familiar with Go; is the succession of hidden `for` loops unavoidable after all?)

graphemecluster (commented):
I don't think I should overwhelm this page with nitpicking reviews, so I've decided to put the remaining suggested changes in a separate PR #2107 for your convenience.

jakebailey:
@graphemecluster Thanks for looking, and for the PR!

I have a bad feeling my approach is bad in general in that I have managed to duplicate functionality of the scanner itself, such that this code needs to move again...

I need some time to think harder about this (and your other PR on the main repo; microsoft/TypeScript#61042)...

graphemecluster:
`CharacterEscape` and `EscapeSequence` are separate productions in the spec after all, it was just that the they are "similar" enough that I combined them together in my old implementation. Only `HexEscapeSequence` and `LegacyOctalEscapeSequence` are real overlaps within the two productions. So I am not sure if it is really worthwhile abstracting the method out.
But well, I must not ignore the fact that they do account for a large portion of `scanEscapeSequence`…

graphemecluster:
Ah, you mean the identifier name part. Well obviously `RegExpIdentifierName` and IdentifierName` are distinct productions, and they do have quite some subtle differences.
As for the utils dealing with surrogate pair, I would definitely suggest moving them to core or perhaps a standalone text manipulation folder.

graphemecluster:
Just to be sure, did you close this in favor of #3061?

jakebailey:
Yes, correct.

PR #2028, opened 2025-11-06
Split "use strict" into separate transformer, fix bugs with prologues

jakebailey:
This fixes a massive number of diffs relating to prologues, including missing `"use strict"` or duplicating prologues/comments in general.

Before, `"use strict"` was added in the CJS transform. But this meant we were not adding it to non-CJS files that also needed it, e.g. scripts.

This PR splits the functionality out into a fine-grained transform which handles _just_ adding `"use strict"`, drops the code from CJS, then fixes related bugs in transforms/emit that were causing things to break with this fix in place.

After this PR, there are no missing `"use strict"` in any diffs, except for files which are missing entire files or are in general not being detected as ESM correctly. (only 9 diffs last I checked)


jakebailey:
@sheetalkamat I don't really know how to handle this one; the test seems to be flaky. It seems to me like we are not correctly noting that strict/alwaysStrict are emit affecting (or even bind affecting!)

Issue #1701, opened 2025-09-10
surrogate pair and lone surrogate support in stringLiteral


hardfist:
## Steps to reproduce
for following code tsgo and typescript generate differents token text 
```
"🦀\ud7ff\ud800\ud801\uD83E\uDD80"
```
It seems tsgo using go string to store codePoint(from JS string),
https://github.com/microsoft/typescript-go/blob/0216862d44c9b14717b7400818cf300f99ec5d1f/internal/ast/ast.go#L5813
but JS string is not strict UTF16 string which may contain lone surrogate while go string will convert lone surrogate to U+FFFD which is a lossy conversion and lose the origin info 

## Behavior with `typescript@5.8`
```
🦀\ud7ff\ud800\ud801\uD83E\uDD80
```
https://ts-ast-viewer.com/#code/ESPg3AG7A6CuAmDsAzB0YA4AM6UYIzQCKoDMAogYesEA
## Behavior with `tsgo`
```
🦀퟿����
```
https://rslint.rs/playground/?tab=ast&code=%22%F0%9F%A6%80%5Cud7ff%5Cud800%5Cud801%5CuD83E%5CuDD80%22

jakebailey:
Go strings don't do any conversion themselves (it's not like Rust strings; Go strings are just bytes); we scan directly from JS source into the string, and then back out into JS source.

When I make this test:

```ts
export const x = "🦀\ud7ff\ud800\ud801\uD83E\uDD80";
```

We output:

```ts
//// [tests/cases/compiler/tsgo1701.ts] ////

//// [tsgo1701.ts]
export const x = "🦀\ud7ff\ud800\ud801\uD83E\uDD80";


//// [tsgo1701.js]
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.x = void 0;
exports.x = "🦀\ud7ff\ud800\ud801\uD83E\uDD80";
```

Which correctly roundtrips the string.

Are you sure this is not an issue with this alternative playground or an external project?

hardfist:
That's strange, I actually debug using [encoder_test](https://github.com/microsoft/typescript-go/blob/0216862d44c9b14717b7400818cf300f99ec5d1f/internal/api/encoder/encoder_test.go#L20)
It seems that tsgo will try to convert unicode sequence into rune and then convert rune into string when do lexer 
[...]

PR #2032, opened 2025-11-07
Fix scanning of valid surrogate pairs

haoqixu:
Some characters can be represented as a pair of UTF16 code units in TS/JS, for example, `'😃'` == `'\ud83d\ude03'`. `tsgo` does not handle these strings correctly and fails to produce valid binary `ast` via api's encoder.

This PR might be related to #1701, but it only focuses on fixing the valid surrogate pairs.


internal/scanner/scanner.go
```diff
@@ -1629,6 +1629,13 @@ func (s *Scanner) scanEscapeSequence(flags EscapeSequenceScanningFlags) string {
 		codePoint := s.scanUnicodeEscape(flags&EscapeSequenceScanningFlagsReportInvalidEscapeErrors != 0)
 		if codePoint < 0 {
 			return s.text[start:s.pos]
+		} else if codePointIsHighSurrogate(codePoint) && s.char() == '\\' && s.charAt(1) == 'u' {
+			savedPos := s.pos
+			nextCodePoint := s.scanUnicodeEscape(flags&EscapeSequenceScanningFlagsReportInvalidEscapeErrors != 0)
+			if codePointIsLowSurrogate(nextCodePoint) {
```
DanielRosenwasser, line 1635: I think I need to see what happens when a user has provided a single high surrogate, a low surrogate, or mismatched surrogates.

Maybe we need a compiler test for this in `tests/cases/compiler` or `tests/cases/conformance` with the following:

```ts
// @declaration: true

// low-high surrogate pair - the "correct" case
export const highLow = "\ud83d\ude03" as const;

// high surrogate
export const high = "\ud83d" as const;

// low surrogate
export const low = "\ude03" as const;

// two high surrogates
export const highHigh = "\ud83d\ud83d" as const;

// two low surrogates
export const lowLow = "\ude03\ude03" as const;

// swapped expected order of surrogates
export const lowHigh = "\ude03\ud83d" as const;
```

I think you are currently doing the right thing by only consuming when you have a correct pair, but I'm guessing that we're missing coverage here, so might as well add the test now.

internal/scanner/utilities.go
```diff
@@ -8,6 +8,25 @@ import (
...
+func codePointIsHighSurrogate(r rune) bool {
+	return surr1 <= r && r < surr2
+}
+
+func codePointIsLowSurrogate(r rune) bool {
+	return surr2 <= r && r < surr3
+}
+
+func surrogatePairToCodepoint(r1, r2 rune) rune {
+	return (r1-surr1)<<10 | (r2 - surr2) + surrSelf
```
DanielRosenwasser: I don't know if we're better off with


```suggestion
	return utf16.DecodeRune(r1, r2)
```

but this is probably fine.

internal/scanner/scanner.go
jakebailey: Also, I think I have done some very similar stuff in #2026, which (unfortunately) duplicates some of this scanning code in another package and fixes the bugs.

haoqixu: > Maybe we need a compiler test for this in `tests/cases/compiler` or `tests/cases/conformance` with the following:
> 
I have added these testcases to `tests/cases/compiler`.

haoqixu: > Also, I think I have done some very similar stuff in #2026, which (unfortunately) duplicates some of this scanning code in another package and fixes the bugs.

Got it. I noticed that `internal/regexpchecker/utf16.go` in #2026 handles some similar unicode escapes processing, but I've verified that it doesn't actually fix the issue here. The `internal/scanner` stores the interpreted string for string literal while preserving the original text for regex literal in `tokenValue`, which is used by `internal/api/encoder` for encoding `ast` into binary foramt. We still need to update the scanner's code to properly decode valid surrogate pairs for strings.

As for invalid escapes, it involves encoding them into some kind of non-strict utf8 without data loss, transmitting them through the binary format and restoring them in the TypeScript API. I think it would be better to track that in a separate issue (perhaps in #1701) :thinking:

haoqixu: > I noticed that `internal/regexpchecker/utf16.go` in #2026 handles some similar unicode escapes processing

> We still need to update the scanner's code to properly decode valid surrogate pairs for strings.

Perhaps we could consider refactoring by extracting the code from `internal/regexpchecker/utf16.go` into a reusable package. This way, it could be repurposed for use within `internal/scanner` as well

jakebailey: I suspect all of the code in that PR needs to move back into the scanner package anyway

DanielRosenwasser: > As for invalid escapes, it involves encoding them into some kind of non-strict utf8 without data loss

I guess we should look into https://wtf-8.codeberg.page. Maybe I've missed something, but I think that just says that we should preserve unmatched surrogates as-is.

testdata/tests/cases/compiler/unicodeSurrogatesInStringLiterals.ts
DanielRosenwasser, line 1: I'd rename this `unicodeSurrogatesInStringLiterals.ts`.

We will eventually need to look at Unicode escapes in identifiers too.

internal/scanner/scanner.go
jakebailey: I think this PR is right. But, we have a bigger problem elsewhere long-term when it comes to dealing with these.

haoqixu:
The CI failure does not seem related to this PR, I've submitted another PR (#2056) to fix the panic. :thinking: Considering that this failure only occurs in data race mode, I am not sure whether it implies other problems.

jakebailey:
Yeah, it's a flake. I have rerun it.

PR #2033, opened 2025-11-07
Port tsc --init

### LINKED ISSUE #1602: Initializing tsconfig in VSCode workspace results in error exit code
## Steps to reproduce

1. Create workspace with two folders
2. Initialize each folder with `npm init`
3. Add latest (currently `7.0.0-dev.20250818.1`) `@typescript/native-preview`
4. Run `npx tsgo --init`

## Repository example

https://github.com/XIshArkIX/typescript-go-issue-example

Generated `tsconfig.json`

## Behavior with `tsgo`

Failed to generate. Exit code 5.

haoqixu:
This PR ports `tsc --init` from `Strada`.

Fixes #1602

### DIFF
--- a/internal/execute/tsc/init.go
+++ b/internal/execute/tsc/init.go
@@ -15,7 +15,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
-func WriteConfigFile(sys System, reportDiagnostic DiagnosticReporter, options *core.CompilerOptions) {
+func WriteConfigFile(sys System, reportDiagnostic DiagnosticReporter, options *collections.OrderedMap[string, any]) {
 	getCurrentDirectory := sys.GetCurrentDirectory()
 	file := tspath.NormalizePath(tspath.CombinePaths(getCurrentDirectory, "tsconfig.json"))
 	if sys.FS().FileExists(file) {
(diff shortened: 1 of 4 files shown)


internal/execute/tsctests/tsc_test.go
```diff
@@ -36,6 +36,28 @@ func TestTscCommandline(t *testing.T) {
 			subScenario:     "when build not first argument",
 			commandLineArgs: []string{"--verbose", "--build"},
 		},
+		{
+			subScenario:     "init",
```
jakebailey: Can you do all of the tests listed in `_submodules/TypeScript/src/testRunner/unittests/config/initializeTSConfig.ts`?

haoqixu: Sure. No problem. It's done.

internal/execute/tsc/init.go
```diff
@@ -0,0 +1,238 @@
...
+			continue
+		}
+
+		// Get the field name, considering 'json' tag if present
+		fieldName := field.Name
+		if jsonTag := field.Tag.Get("json"); jsonTag != "" {
+			fieldName, _, _ = strings.Cut(jsonTag, ",")
+		}
+
+		if fieldName != "" && fieldName != "init" && fieldName != "help" && fieldName != "watch" {
```
jakebailey: I think there are other options like `--all` that also have to be dropped. Do we have some preexisting way to identify these rather than hardcoding them?

internal/execute/tsc/init.go
```diff
@@ -0,0 +1,238 @@
...
+		} else {
+			return formatSingleValue(value, option.EnumMap())
+		}
+	}
+
+	// commentedNever': Never comment this out
+	// commentedAlways': Always comment this out, even if it's on commandline
+	// commentedOptional': Comment out unless it's on commandline
+	const (
+		commentedNever    = 0
```
jakebailey: You can probably declare a local type and use `iota` here.

internal/execute/tsc/init.go
jakebailey: Ah, no, this is actually straight ported. Fun. That's probably a bug!

testdata/baselines/reference/tsc/commandLine/Initialized-TSConfig-with-advanced-options.js
```diff
@@ -0,0 +1,61 @@
...
+    // Recommended Options
+    "strict": true,
+    "jsx": "react-jsx",
+    "verbatimModuleSyntax": true,
+    "isolatedModules": true,
+    "noUncheckedSideEffectImports": true,
+    "moduleDetection": "force",
+    "skipLibCheck": true,
+
+    "declarationDir": "/home/src/workspaces/project/lib",
```
jakebailey: This path is printed incorrectly; it shouldn't include an absolute path.

jakebailey: I think this might come down to the fact that our compiler options are fully resolved now, whereas the old code was not? In which case, you may need to resolve any path-like options relative to the working directory. But since you're just reflecting over options instead of going via options declarations, I am not sure how simple that is.

Perhaps this means we should change the method to _not_ be like Strada and instead use the options declarations to find values... Which would probably also let us not special case the things I mentioned above.

haoqixu: I've updated the code to generate `tsconfig` from `commandLine.Raw` instead of `commandLine.CompilerOptions()`.

jakebailey: That works!

PR #2044, opened 2025-11-09
Ignore reparsed nodes when determining external module indicator

jakebailey:
These fake nodes should not contribute to the module-ness of a file.


testdata/baselines/reference/submodule/conformance/untypedModuleImport_allowJs.errors.txt
```diff
@@ -0,0 +1,12 @@
+/a.ts(2,5): error TS2339: Property 'bar' does not exist on type 'typeof import("/node_modules/foo/index")'.
```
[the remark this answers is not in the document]
jakebailey: Nope, see the diff file that goes with this

DanielRosenwasser:
Can you give more context on why this was happening? I don't get what was being injected that was module-like but wouldn't imply that the output should be a module.

testdata/baselines/reference/submodule/compiler/jsFileAlternativeUseOfOverloadTag.js.diff
andrewbranch, line 1: I can't immediately figure out what's going on with this diff, but it's the only one that didn't shrink

jakebailey: The diff makes it look worse than it is; this is actually a preexisting bug @sandersn also just found in https://github.com/microsoft/typescript-go/issues/2047

jakebailey:
> Can you give more context on why this was happening? I don't get what was being injected that was module-like but wouldn't imply that the output should be a module.

Yes, for example `@callback` can get turned into `export type Foo = () => void`, and this code would find that node and go "oh, there's an export" and use that to determine if it's a module. Or `module.exports = {}` turned into `export = {}` and that gets caught too.

Basically, the thinking is that the old tree in Strada would never have any of these new reparsed nodes, so to match algorithms, we need to ignore them too.

DanielRosenwasser:
I'm sure this is an okay fix, but I think the more appropriate thing would have been for the `@callback` to transform differently based on if the containing file was originally a module. Having `export`ed nodes in a global file is very suspicious for any later analysis.

jakebailey:
That isn't possible; reparsing happens on the fly, whereas module-ness is determined after the fact.

I agree that export keywords should not appear; I suspect you're looking for a fix more like #1999 which I think may have just missed `@callback`.

jakebailey:
Filed #2048

ahejlsberg:
I'm not sure I agree with this PR. Reparsed nodes are intended to _replace_ JSDoc nodes (from a semantic point of view) such that downstream logic doesn't have to examine the JSDoc. But if we ignore them, then we push the responsibility downstream--exactly what we were trying to avoid. Once I get my fix for #2048 up, reparsed `JSTypeAliasDeclaration` nodes resulting from `@callback` wont have an `export` modifier, at which point this PR doesn't serve any purpose.

ahejlsberg:
I recommend we revert this PR.

jakebailey:
I can do that if you have the follow up prepared.

jakebailey:
#2062

PR #2049, opened 2025-11-10
Update vscode-languageclient to v10

jakebailey:
This has two negative effects:

- All of our current logging appears as "error" output because it went over stderr.
  - "Fixed" by introducing leveled logging.
- All LSP logging is hidden by default, which is perhaps good for end users but sort of annoying for debugging.
  - related: https://github.com/microsoft/vscode/issues/223536
  - You can at least select the "default" log level for a channel, which will persist. That's probably good enough for regular debugging.

But, fixes client bugs we've experienced.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -130,8 +130,9 @@ var (
 )
 
 type Server struct {
-	r      Reader
-	w      Writer
+	r Reader
+	w Writer
+
 	stderr io.Writer
 
 	logger                  *logger


_extension/package.json
```diff
@@ -108,7 +108,7 @@
         "bundle": "esbuild src/extension.ts --bundle --external:vscode --platform=node --format=cjs --outfile=dist/extension.bundle.js --minify"
     },
     "dependencies": {
-        "vscode-languageclient": "^9.0.1"
+        "vscode-languageclient": "^10.0.0-next.17"
```
DanielRosenwasser: Looks like we might be better off moving to `next.18` at this point - @dbaeumer mentioned at https://github.com/microsoft/vscode/issues/280515#issuecomment-3610975738 that there are some pull diagnostics based fixes/improvements worth trying out.

jakebailey:
Alternatively, I can implement my own `LogOutputChannel` (it's just an interface) which logs to a non-log output, but I don't think that's going to work so well with the log level / event emitter :\

Probably best to acquiesce

PR #2056, opened 2025-11-11
Fix panic in syncmap on loading nil

haoqixu:
`SyncMap` uses a type assertion to return values, which is not a safe operation for a `nil`. This PR handles `nil` specifically to avoid panic.

see https://github.com/microsoft/typescript-go/actions/runs/19256942354/job/55053261658?pr=2032#step:10:48

### DIFF
--- a/internal/collections/syncmap_test.go
+++ b/internal/collections/syncmap_test.go
@@ -8,6 +8,8 @@ import (
 )
 
 func TestSyncMapWithNil(t *testing.T) {
+	t.Parallel()
+
 	var m collections.SyncMap[string, any]
 
 	got1, ok := m.Load("foo")


internal/collections/syncmap.go
```diff
@@ -13,7 +13,7 @@ type SyncMap[K comparable, V any] struct {
 
 func (s *SyncMap[K, V]) Load(key K) (value V, ok bool) {
 	val, ok := s.m.Load(key)
-	if !ok {
+	if !ok || val == nil {
```
jakebailey, line 16: I don't understand when this would happen. If the map holds `V`, why would we ever have stored a `nil` except when `V` is `any` and therefore the conversion should succeed?

haoqixu: The panic occurs only if `V` is an interface, and we store a `nil` into the map and then load it. For example, `cachedvfs` holds a `statCache` whose type is `collections.SyncMap[string, vfs.FileInfo]`. If the implementation of its underlying `fs` returns a `nil` literal in its `Stat`, that `nil` will cause a panic when we load it from the map. :thinking:

haoqixu: > If the implementation of its underlying `fs` returns a `nil` literal in its `Stat`, that `nil` will cause a panic when we load it from the map.

It looks like this: https://go.dev/play/p/AmDx9AdDj6h

jakebailey: Darn, ok.

PR #2062, opened 2025-11-11
Revert PR 2044

jakebailey:
I guess to replace my big hammer with a smaller hammer.

#2044

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -617,6 +617,8 @@ func (n *Node) StatementList() *NodeList {
 		return n.AsBlock().Statements
 	case KindModuleBlock:
 		return n.AsModuleBlock().Statements
+	case KindCaseClause, KindDefaultClause:
+		return n.AsCaseOrDefaultClause().Statements
 	}
 	panic("Unhandled case in Node.StatementList: " + n.Kind.String())
 }
@@ -1038,6 +1040,10 @@ func (n *Node) Statement() *Statement {
 		return n.AsForStatement().Statement
 	case KindForInStatement, KindForOfStatement:
 		return n.AsForInOrOfStatement().Statement
+	case KindWithStatement:
+		return n.AsWithStatement().Statement
+	case KindLabeledStatement:
+		return n.AsLabeledStatement().Statement
 	}
 	panic("Unhandled case in Node.Statement: " + n.Kind.String())
 }
@@ -1068,8 +1074,11 @@ func (n *Node) ElementList() *NodeList {
 		return n.AsNamedExports().Elements
 	case KindObjectBindingPattern, KindArrayBindingPattern:
 		return n.AsBindingPattern().Elements
+	case KindArrayLiteralExpression:
+		return n.AsArrayLiteralExpression().Elements
+	case KindTupleType:
+		return n.AsTupleTypeNode().Elements
 	}
-
 	panic("Unhandled case in Node.ElementList: " + n.Kind.String())
 }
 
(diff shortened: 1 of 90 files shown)


ahejlsberg (approved):

jakebailey:
Going to wait on the other PR since this exposes more problems that would have to be fixed... I am not super content with reverting it

jakebailey:
Updated; there are still many regressions by removing this.

ahejlsberg:
What is the pattern to the regressions you're seeing? I'm still surprised we'd consider excluding reparsed nodes from module determination, just seems wrong to do so.

jakebailey:
I don't see why we would include them; the old algorithm did _not_ walk into JSDoc nodes, and any other new stuff we stick in the tree to handle expandos or CJS exports and whatnot were not there either, right?

If the PR is filtered to diffs, you can see what's regressing: https://github.com/microsoft/typescript-go/pull/2062/files?file-filters%5B%5D=.go&file-filters%5B%5D=.diff

I can annotate the PR with a few examples.

jakebailey:
To be clear, I agree that ideally we wouldn't do this, but for the most part the problem is that there are other bugs that this is sort of helping with. I don't think we need to keep it forever.

testdata/baselines/reference/submodule/conformance/assignmentToVoidZero1.js.diff
jakebailey, line 16: This is a regression but you'll note that the emit is already wrong; we are emitting actual `export var x` when we shouldn't. These nodes existing in the tree means we are both detecting it as a module and then printing them back.

testdata/baselines/reference/submodule/compiler/modulePreserve4.errors.txt.diff
```diff
@@ -30,19 +31,21 @@
...
      import a1 = require("./a"); // { x: 0 }
      const a2 = require("./a"); // Error in TS
                 ~~~~~~~
-@@= skipped -42, +40 lines =@@
+@@= skipped -37, +35 lines =@@
+     
+     import g1 from "./g"; // { default: 0 }
+     g1.default;
++       ~~~~~~~
++!!! error TS2339: Property 'default' does not exist on type '0'.
```
jakebailey, line 48: Strange error appearing here. `g.js` is `exports.default = 0` which breaks for some reason.

testdata/baselines/reference/submodule/conformance/jsDeclarationsClassExtendsVisibility.js.diff
```diff
@@ -1,6 +1,15 @@
 --- old.jsDeclarationsClassExtendsVisibility.js
 +++ new.jsDeclarationsClassExtendsVisibility.js
-@@= skipped -25, +25 lines =@@
+@@= skipped -17, +17 lines =@@
+ }
+ module.exports = Bar;
+ //// [cls.js]
++"use strict";
++Object.defineProperty(exports, "__esModule", { value: true });
```
jakebailey, line 8: This file is becoming strict when it shouldn't be; strict is off, alwaysStrict is off, so the determination is only whether or not there's an external module indicator, which Strada must not have had.

testdata/baselines/reference/submodule/conformance/moduleExportAliasImported.types.diff
```diff
@@ -31,5 +31,5 @@
  === importer.js ===
  import('./bug28014')
 ->import('./bug28014') : Promise<{ version: 1; default: { (): void; version: 1; }; }>
-+>import('./bug28014') : Promise<{ default: () => void; }>
++>import('./bug28014') : Promise<() => void>
  >'./bug28014' : "./bug28014"
```
jakebailey, line 35: The type here changes.

testdata/baselines/reference/submodule/conformance/untypedModuleImport_allowJs.errors.txt.diff
```diff
@@ -0,0 +1,16 @@
...
+-==== /a.ts (1 errors) ====
+-    import foo from "foo";
+-    foo.bar();
+-        ~~~
+-!!! error TS2339: Property 'bar' does not exist on type 'typeof import("/node_modules/foo/index")'.
+-    
+-==== /node_modules/foo/index.js (0 errors) ====
+-    exports.default = { bar() { return 0; } }
+-    
++<no content>
```
jakebailey, line 16: This error disappears.

ahejlsberg:
I think a lot of these issues are caused by the re-parser adding a synthetic `export` modifier to `KindCommonJSExport` nodes. I don't know why that code is there, but it is causing the `isAnExternalModuleIndicatorNode` function to return true when it shouldn't. I think a first step should be to remove that `export` modifier.

ahejlsberg:
Ok, it appears the only reason for that synthetic `export` modifier is so that the binder considers `KindCommonJSExport` nodes to declare exported symbols. When I remove the `export` modifier and change the code in `declareModuleMember` in the binder to always consider `KindCommonJSExport` nodes to be exports, there are no baseline changes. I will put up a PR to that effect, I think it is a better solution than the blunt instrument of ignoring all reparsed nodes.

Also, I agree there are still issues that need to be fixed in this JS module related code.

PR #2063, opened 2025-11-11
Don't add `export` modifier to `JSTypeAliasDeclaration` from `@callback`

### LINKED ISSUE #2048: `@callback` in JS files should not get export modifier, similar to `@typedef` fix
#1999 applied to `@typedef`, but other syntax like `@callback` and maybe even `@overload` also need to be careful to not confuse things; see https://github.com/microsoft/typescript-go/pull/2044#issuecomment-3513672020.

I am not sure if we have tests specifically for this case; the places I'm seeing a diff with `@callback` appear to be in files that are already modules through other means?

ahejlsberg:
Fixes #2048.

### DIFF
--- a/testdata/baselines/reference/compiler/jsDocCallbackExport1.js
+++ b/testdata/baselines/reference/compiler/jsDocCallbackExport1.js
@@ -1,0 +1,21 @@
+//// [tests/cases/compiler/jsDocCallbackExport1.ts] ////
+
+//// [x.js]
+/**
+ * @callback Foo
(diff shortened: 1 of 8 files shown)


jakebailey (approved):
This seems correct, but doesn't seem to fix everything that my fix caught for the external module indicator, so I guess there's more work on that end (or my change was correct, which I think it is in theory...)

Checked by applying https://github.com/microsoft/typescript-go/pull/2062 on top of this PR and looking at the new diffs

PR #2067, opened 2025-11-12
Move unreachable checks to checker, allowing more AST reuse

jakebailey:
They say it's not normal to have a "white whale", but this is one of mine...

---

Currently, we define these "source affecting compiler options":

```go
type SourceFileAffectingCompilerOptions struct {
	AllowUnreachableCode     Tristate
	AllowUnusedLabels        Tristate
	BindInStrictMode         bool
	ShouldPreserveConstEnums bool
}

// ...
options.sourceFileAffectingCompilerOptions = SourceFileAffectingCompilerOptions{
	AllowUnreachableCode:     options.AllowUnreachableCode,
	AllowUnusedLabels:        options.AllowUnusedLabels,
	BindInStrictMode:         options.AlwaysStrict.IsTrue() || options.Strict.IsTrue(),
	ShouldPreserveConstEnums: options.ShouldPreserveConstEnums(),
}
// ...
```

This is a part of the input/cache key for ASTs. A difference in any of these options between projects means we cannot reuse the ASTs and therefore reparse files, which can have a memory impact for the language server, build mode, even our test suite. (Declaration files are exempt from this as none of these options matter for them; since #1158.)

We're already planning on making all files strict (https://github.com/microsoft/TypeScript/issues/62213) meaning `BindInStrictMode` will go away, so that leaves three other options.

It turns out that all three of them have to do with unreachable code detection, including `ShouldPreserveConstEnums`.

After some futzing, I was able to move all of the checks the binder previously did into the checker.

The binder still does its usual `unreachableFlow` stuff, but also sets `NodeFlagUnreachable` on statements/declarations to be reported by the checker.

These diagnostics are still presented as suggestion diagnostics when needed, and appear to work for me, including in JS files with/without `// @ts-check`.

---

If we go through with the strict mode change, there will be _no more_ source file affecting compiler options, leaving just this as the input to parse/bind (which are a pure function in Corsa):

```go
type SourceFileParseOptions struct {
	FileName                       string
	Path                           tspath.Path
	ExternalModuleIndicatorOptions ExternalModuleIndicatorOptions
	JSDocParsingMode               JSDocParsingMode
}
```

Plus the text itself, and the `ScriptKind`.

`JSDocParsingMode` is stable, and `ExternalModuleIndicatorOptions` is effectively derived from the path, so this would be as minimal as it can get!

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -2366,7 +2366,7 @@ func (c *Checker) checkSourceElementUnreachable(node *ast.Node) bool {
 		}
 	}
 
-	start = scanner.GetRangeOfTokenAtPosition(sourceFile, start).Pos()
+	start = scanner.SkipTrivia(sourceFile.Text(), start)
 
 	diagnostic := ast.NewDiagnostic(sourceFile, core.NewTextRange(start, end), diagnostics.Unreachable_code_detected)
 	c.addErrorOrSuggestion(c.compilerOptions.AllowUnreachableCode == core.TSFalse, diagnostic)


testdata/baselines/reference/submodule/conformance/neverReturningFunctions1.errors.txt.diff
```diff
@@ -0,0 +1,32 @@
...
+         if (typeof x.a === "string") {
+             fail();
+             x;    // Unreachable
+-            ~~
+-!!! error TS7027: Unreachable code detected.
++            ~~~~~~~~~~~~~~~~~~~~
+             x.a;  // Unreachable
+-            ~~~~
++    ~~~~~~~~~~~~
+ !!! error TS7027: Unreachable code detected.
```
jakebailey, line 30: This is the one diff. For some reason, we used to issue two errors here instead of one. I would like to call this an improvement.

internal/checker/checker.go
```diff
@@ -2306,6 +2332,114 @@ func (c *Checker) checkSourceElementWorker(node *ast.Node) {
...
+	scanAndReportExecutableStatements := func(statements []*ast.Node, index int) {
+		var first, last *ast.Node
+		for _, s := range statements[index:] {
+			if isExecutableStatement(s) {
+				if first == nil {
+					first = s
+				}
+				last = s
+			} else {
+				break
```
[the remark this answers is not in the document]
jakebailey: This is true, but one of the weird things I had to change to make the errors match; I believe the checker still goes down the statements anyway because this is happening in checkSourceElement.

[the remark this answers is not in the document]
jakebailey: (I believe given the way the binder used to walk the code and report unreachable, that it _also_ stopped early simply because it reached code that had not yet been bound...)

internal/ast/ast.go
```diff
@@ -3564,8 +3564,9 @@ func (node *DebuggerStatement) Clone(f NodeFactoryCoercible) *Node {
 
 type LabeledStatement struct {
 	StatementBase
-	Label     *IdentifierNode // IdentifierNode
-	Statement *Statement      // Statement
+	Label        *IdentifierNode // IdentifierNode
+	Statement    *Statement      // Statement
+	IsReferenced bool            // Set by binder to indicate if the label is used by a break or continue statement
```
ahejlsberg: I would use a `NodeFlags` for this.

internal/binder/binder.go
```diff
@@ -1657,94 +1660,6 @@ func (b *Binder) bindEachStatementFunctionsFirst(statements *ast.NodeList) {
 	}
 }
 
-func (b *Binder) checkUnreachable(node *ast.Node) bool {
```
ahejlsberg, line 1660: I'm thinking you could keep all of this logic in the binder, but set a `NodeFlags` on each unreachable node instead of reporting errors. Then the checker could just do the error reporting when requested based on the flags. Anything to keep logic out of the checker!

ahejlsberg: This might in fact be a useful flag to have for other tools that operate on the AST.

jakebailey: I don't see how that is possible; whether or not a specific piece of code is marked as unreachable directly depends on the compiler options thanks  to ShouldPreserveConstEnums. I'd have to unconditionally mark it as an unused node and then add conditions somewhere else...

ahejlsberg: Right, I would just mark any enum declaration as unreachable and then suppress the diagnostic depending on `ShouldPreserveConstEnums`.

ahejlsberg: Or just not report this on enum declarations. Seems kinda odd to begin with.

jakebailey: I'll try setting it unconditionally.

I'm not so sure that it'll be less code, and the flag will only ever be able to indicate what the binder says is unreachable; the checker already has its own extra unreachable detection via type analysis that is much more expensive.

It seemed to me simpller to just absolve the binder of this duty.

ahejlsberg: I'll leave it to your discretion. I was just hoping to keep some of the code out of the checker.

jakebailey: I think I have an alternative that uses a flag-based approach, but the downside is that it reports unreachable for every statement unless I duplicate the code that does the range-combining. Maybe it's not such a big deal but it could cause a load of diagnostic churn.

jakebailey: I have a _much_ simpler version of this PR, but it _only_ works if we also fix https://github.com/microsoft/TypeScript/issues/55562 / do kinda sorta https://github.com/microsoft/TypeScript/pull/55588.

jakebailey: I've pushed a revamped version. This one incorporates the tests from the above linked PR and effectively fixes that issue.

internal/ast/ast.go
```diff
@@ -631,6 +631,15 @@ func (n *Node) Statements() []*Node {
 	return nil
 }
 
+func (n *Node) CanHaveStatements() bool {
```
jakebailey, line 634: Alternatively, `Statements()` could return nil. I am unsure if that's a great idea.

internal/checker/checker.go
```diff
@@ -2306,6 +2310,60 @@ func (c *Checker) checkSourceElementWorker(node *ast.Node) {
...
+	if !c.isSourceElementUnreachable(node) {
+		return false
+	}
+
+	c.reportedUnreachableNodes.Add(node)
+
+	isError := c.compilerOptions.AllowUnreachableCode == core.TSFalse
+	sourceFile := ast.GetSourceFileOfNode(node)
+
+	start := scanner.GetRangeOfTokenAtPosition(sourceFile, node.Pos()).Pos()
```
DanielRosenwasser:
```suggestion
	start := scanner.SkipTrivia(sourceFile, node.Pos())
```

internal/checker/checker.go
```diff
@@ -2306,6 +2310,60 @@ func (c *Checker) checkSourceElementWorker(node *ast.Node) {
...
+	if node.Flags&ast.NodeFlagsUnreachable != 0 {
+		switch node.Kind {
+		case ast.KindEnumDeclaration:
+			return !ast.IsEnumConst(node) || c.compilerOptions.ShouldPreserveConstEnums()
+		case ast.KindModuleDeclaration:
+			return ast.IsInstantiatedModule(node, c.compilerOptions.ShouldPreserveConstEnums())
+		default:
+			return true
+		}
+	} else if ast.KindFirstStatement <= node.Kind && node.Kind <= ast.KindLastStatement {
```
DanielRosenwasser: Can you leave a comment either here or at `NodeFlagsUnreachable` about why only one of these checks isn't sufficient?

jakebailey:
In porting this back to Strada, I've found other bugs in this that are being silenced by the main test for unreachable code being all unreachable itself. Will figure that one out.

jakebailey:
Okay, seems to be working as expected now. I forgot about the `var x;` carve-out.

I also made https://github.com/microsoft/TypeScript/pull/62751 which helped me find that bug. Will be interested in those results.

internal/checker/checker.go
```diff
@@ -2177,13 +2182,13 @@ func (c *Checker) checkSourceElementWorker(node *ast.Node) {
 			}
 		}
 	}
-	kind := node.Kind
-	if kind >= ast.KindFirstStatement && kind <= ast.KindLastStatement {
-		flowNode := node.FlowNodeData().FlowNode
-		if flowNode != nil && !c.isReachableFlowNode(flowNode) {
-			c.errorOrSuggestion(c.compilerOptions.AllowUnreachableCode == core.TSFalse, node, diagnostics.Unreachable_code_detected)
+
+	if !c.withinUnreachableCode {
```
ahejlsberg: If feels like this could benefit from keeping the is-potentially-executable check on the `kind` here to quickly skip over nodes that are never going to be unreachable. Right now the always-false code path involves function calls and a map lookup.

internal/checker/checker.go
```diff
@@ -2306,6 +2311,65 @@ func (c *Checker) checkSourceElementWorker(node *ast.Node) {
...
+}
+
+func (c *Checker) isSourceElementUnreachable(node *ast.Node) bool {
+	if node.Flags&ast.NodeFlagsUnreachable != 0 {
+		switch node.Kind {
+		case ast.KindEnumDeclaration:
+			return !ast.IsEnumConst(node) || c.compilerOptions.ShouldPreserveConstEnums()
+		case ast.KindModuleDeclaration:
+			return ast.IsInstantiatedModule(node, c.compilerOptions.ShouldPreserveConstEnums())
+		case ast.KindVariableStatement:
```
ahejlsberg: Couldn't you just check each `KindVariableDeclaration` here and rely on the outer logic to iterate over variable declaration lists?

testdata/baselines/reference/submodule/compiler/reachabilityChecks1.errors.txt
```diff
@@ -1,58 +1,77 @@
...
 reachabilityChecks1.ts(60,5): error TS7027: Unreachable code detected.
 reachabilityChecks1.ts(69,5): error TS7027: Unreachable code detected.
 
 
-==== reachabilityChecks1.ts (7 errors) ====
+==== reachabilityChecks1.ts (5 errors) ====
     while (true);
     var x = 1;
     ~~~~~~~~~~
-!!! error TS7027: Unreachable code detected.
```
ahejlsberg, line 14: So are the changes here because the new logic does a better job of merging adjacent unreachable code blocks? Looks like a nice improvement.

jakebailey: In this test, kinda; this test made everything below the first line unreachable, so this changing is actually a fix for https://github.com/microsoft/TypeScript/issues/55562.

jakebailey: I'd look at the errors in reachabilityChecks11.ts for the non-broken example.

internal/checker/checker.go
jakebailey: Yeah, I'll play around with it. I think there's an extra simplification I can make here.

jakebailey: Done; that's now checked first before we do anything else.

internal/checker/checker.go
jakebailey: No, since reachability only works on statements, and we don't want to piecemeal report only half of say `var x, y = 1` as unreachable/reachable or something in that carveout.

jakebailey: That being said, the latest iteration of this PR drops this special branch anyway!

jakebailey:
I'm still investigating these breaks: https://github.com/microsoft/TypeScript/pull/62751#issuecomment-3525796162

ahejlsberg (approved):
One thing to keep in mind is that if we ever do region based semantic diagnostics (like we do in Strada), the logic in this PR may get the merged error spans wrong. We'd have to add some sort of backtracking do determine where a merged span starts.

jakebailey:
Agh, yeah. I have this as a PR ported to Strada so I'm going to have to see if that's majorly broken ☹️

jakebailey:
Alternatively, I could just issue errors at every unreachable statement, but that's probably too chatty...

jakebailey:
> One thing to keep in mind is that if we ever do region based semantic diagnostics (like we do in Strada), the logic in this PR may get the merged error spans wrong. We'd have to add some sort of backtracking do determine where a merged span starts.

I've added that code, commented out for now until we do region diags.

jakebailey:
Not in this PR is the modification to the options declarations to not declare these as source affecting, but that code is not even used at the moment so I'm going to skip it for now.

PR #2068, opened 2025-11-12
Accurately recognize fourslash test as submodule

gabritto:
Noticed in #2040 that we were treating all fourslash tests which we currently diff as submodule tests. This PR makes it so we only treat tests in the `/gen` and `/manual` subfolders as submodule tests, since that's where the ported tests should go.


internal/fourslash/baselineutil.go
```diff
@@ -229,6 +234,10 @@ func getBaselineOptions(command string) baseline.Options {
 	}
 }
 
+func isSubmoduleTest(testPath string) bool {
+	return strings.Contains(testPath, "fourslash/tests/gen") || strings.Contains(testPath, "fourslash/tests/manual")
```
jakebailey, line 238: I was worried about this, but runtime.Callers always returns fourward slash paths, so it should work.

jakebailey (approved):

internal/fourslash/baselineutil.go
gabritto: Hadn't really thought about that, but I tested on Windows. 🙂

PR #2079, opened 2025-11-13
Port non-baseline diagnostics tests

gabritto:
This PR ports diagnostics fourslash tests. I think about 60% of tests that rely on `verify.getXDiagnostics` functions also rely on `verify.codeFix` though, which isn't ported, so the tests in this PR are the remaining 40%.

Note that in Strada, we had `verify.getSyntacticDiagnostics`, `verify.getSemanticDiagnostics`, and `verify.getSuggestionDiagnostics`, but in LSP there is no such distinction. I considered adding some information to the diagnostics we produce so that we could have those three kinds of testing functions, but that seemed not worth the effort, as I think long term we should think of diagnostics in the terms set by LSP, and there aren't that many diagnostics tests to start with.
So instead I've added `VerifyDiagnostics`, `VerifySuggestionDiagnostics`, and `VerifyNonSuggestionDiagnostics`. `VerifyDiagnostics` is for verifying all diagnostics for a file, and what I think we should use for new tests in Corsa, while `VerifyNonSuggestionDiagnostics` is like `verify.getSyntacticDiagnostics` + `verify.getSemanticDiagnostics`, and is implemented by filtering out the diagnostics sent by server that look like suggestion diagnostics, based on their severity and tags. This distinction was enough to get all but one ported test passing (modulo ones that fail for non-diagnostics reasons), and I think it's useful because it's common for a test to have unused identifier errors that don't really matter for the test, and those end up as suggestions.

I'll add baseline diagnostics in my next PR.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -2524,19 +2524,24 @@ func (f *FourslashTest) verifyDiagnostics(t *testing.T, expected []*lsproto.Diag
 	}
 	actualDiagnostics = core.Filter(actualDiagnostics, filterDiagnostics)
 	emptyRange := lsproto.Range{}
-	for _, diag := range expected {
+	expectedWithRanges := make([]*lsproto.Diagnostic, len(expected))
+	for i, diag := range expected {
 		if diag.Range == emptyRange {
 			rangesInFile := f.getRangesInFile(f.activeFilename)
 			if len(rangesInFile) == 0 {
 				t.Fatalf("No ranges found in file %s to assign to diagnostic with empty range", f.activeFilename)
 			}
-			diag.Range = rangesInFile[0].LSRange
+			diagWithRange := *diag
+			diagWithRange.Range = rangesInFile[0].LSRange
+			expectedWithRanges[i] = &diagWithRange
+		} else {
+			expectedWithRanges[i] = diag
 		}
 	}
-	if len(actualDiagnostics) == 0 {
-		actualDiagnostics = nil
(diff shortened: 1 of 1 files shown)


internal/fourslash/fourslash.go
```diff
@@ -2474,6 +2490,65 @@ func (f *FourslashTest) VerifyBaselineInlayHints(
...
+		}
+	}
+	if len(actualDiagnostics) == 0 && len(expectedWithRanges) == 0 {
+		return
+	}
+	assertDeepEqual(t, actualDiagnostics, expectedWithRanges, "Diagnostics do not match expected", diagnosticsIgnoreOpts)
+}
+
+func isSuggestionDiagnostic(diag *lsproto.Diagnostic) bool {
+	return diag.Tags != nil && len(*diag.Tags) > 0 ||
```
jakebailey, line 2548: Suggestions don't necessarily have to have tags; e.g. if we were to do the "convert to ESM" fix, that would just be a hint, no tags. The tags just apply a visual change like strikethrough or greying out.

gabritto: This is an heuristic, as I think we don't have a consistent way of signaling a diagnostic is a suggestion, right? e.g. the unreachable code diagnostic has a tag, but its severity is `Error`.

jakebailey: It is so long as allowUnreachableCode is true; if it's undefined then it's `Hint`. But if that's the difference you're checking then yeah we can leave it.

jakebailey (approved):

PR #2089, opened 2025-11-14
Use SkipTrivia instead of GetRangeOfTokenAtPosition where possible

jakebailey:
@DanielRosenwasser noticed I made this same mistake. `GetRangeOfTokenAtPosition` implicitly skips trivia, but then also scans a token. This is wasteful for callers that could just skip trivia.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -3968,7 +3968,7 @@ func (c *Checker) checkWithStatement(node *ast.Node) {
 	c.checkExpression(node.Expression())
 	sourceFile := ast.GetSourceFileOfNode(node)
 	if !c.hasParseDiagnostics(sourceFile) {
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -3968,7 +3968,7 @@ func (c *Checker) checkWithStatement(node *ast.Node) {
 	c.checkExpression(node.Expression())
 	sourceFile := ast.GetSourceFileOfNode(node)
 	if !c.hasParseDiagnostics(sourceFile) {
-		start := scanner.GetRangeOfTokenAtPosition(sourceFile, node.Pos()).Pos()
+		start := scanner.SkipTrivia(sourceFile.Text(), node.Pos())
```
DanielRosenwasser, line 3971: We're going to be so much faster when erroring on `with` statements now. 😎

PR #2095, opened 2025-11-14
Fully resolve LSP client caps to non-pointers, pass by context

jakebailey:
This PR modifies the LSP generator to also generate a version of `ClientCapabilities` which is fully "resolved", that is, free of pointers and safe to access all the way down to the deepest field, along with a function that does this. On `initialized`, we resolve the caps, fix up the weird diagnostic cap bug (for now), then save that result for later.

This allows us to stop checking nil 1000 times over, eliminating a load of random helpers, if statements, etc, and preventing us from making the same mistake in the future.

This is however a lossy process; we won't know if something was `nil` versus `""`, or something. But I do not believe his matters for anything in the caps, so I think the ergonomics is worth it.

In addition, I've replaced the one-off passing of options with a context value; this is passed to LSP handler methods, replacing our very random use of capabilities (or, derived-from-capability values).

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -14,6 +14,7 @@ import (
 	"github.com/go-json-experiment/json"
 	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/jsonutil"
 	"github.com/microsoft/typescript-go/internal/ls"
 	"github.com/microsoft/typescript-go/internal/ls/lsconv"
 	"github.com/microsoft/typescript-go/internal/ls/lsutil"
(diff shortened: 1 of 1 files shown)


internal/lsp/server.go
```diff
@@ -206,10 +208,7 @@ func (s *Server) UnwatchFiles(ctx context.Context, id project.WatcherID) error {
 
 // RefreshDiagnostics implements project.Client.
 func (s *Server) RefreshDiagnostics(ctx context.Context) error {
-	if s.initializeParams.Capabilities == nil ||
-		s.initializeParams.Capabilities.Workspace == nil ||
-		s.initializeParams.Capabilities.Workspace.Diagnostics == nil ||
-		!ptrIsTrue(s.initializeParams.Capabilities.Workspace.Diagnostics.RefreshSupport) {
+	if !s.clientCapabilities.Workspace.Diagnostics.RefreshSupport {
```
jakebailey, line 211: No ctx here, since this is not an LSP method, but rather a callback given to Session.

gabritto (approved):

jakebailey:
@sheetalkamat I'm sorry if this causes you to have to solve conflicts on #1991 again ☹️

Issue #2081, opened 2025-11-13
Missing error when reference directive file path doesn't exist


RyanCavanaugh:
## Steps to reproduce

```ts
/// <reference path="notfound.d.ts" />
```

## Behavior with `typescript@5.8`
error:
```
D:\Throwaway\corsarepro>tsc
a.ts:1:22 - error TS6053: File 'D:/Throwaway/corsarepro/notfound.d.ts' not found.

1 /// <reference path="notfound.d.ts" />
                       ~~~~~~~~~~~~~
Found 1 error in a.ts:1
```

## Behavior with `tsgo`

No error issued

TranNhatHan:
Is it considered as the new TS design or a bug? I would like to fix this.

jakebailey:
Bug, there should be an error; some code is just not ported.

PR #2096, opened 2025-11-14
fix TS6053 missing file diagnostics to match tsc (#2081)

TranNhatHan:
I tried to fix the bug, but I am unable to resolve the parentFile and textLoc for the Diagnostic (I did note by commenting). However, it now raises an error when reference directive file path doesn't exist.


internal/compiler/program_test.go
```diff
@@ -312,3 +312,46 @@ func BenchmarkNewProgram(b *testing.B) {
 		}
 	})
 }
+
+func TestMissingReferenceFile(t *testing.T) {
```
jakebailey: No need for this; if you fixed the bug, the `declarationEmitIndexTypeNotFound` test should have been updated when you ran the tests.

TranNhatHan: i have just fixed the bug, should i open a new pull request or just commit on this request?

TranNhatHan:
Failed the tests, trying to resolve

internal/compiler/program_test.go
jakebailey: Just push to this one, please

TranNhatHan:
runner.go:131: Test when global file is added, the signatures are updated has unexpected diff Edit [3]:: Add new file and update main file

**Changes observed in incremental build:**
```diff
            -src/main.ts:3:22 - error TS6053: File '/home/src/workspaces/project/src/fileNotFound.ts' not found.
            +src/anotherFileWithSameReferenes.ts:2:22 - error TS6053: File '/home/src/workspaces/project/src/fileNotFound.ts' not found.

            -3 /// <reference path="./fileNotFound.ts"/>
            +2 /// <reference path="./fileNotFound.ts"/>
                                    ~~~~~~~~~~~~~~~~~
            -Found 1 error in src/main.ts:3
            +Found 1 error in src/anotherFileWithSameReferenes.ts:2

**Comment by TranNhatHan** [comment]
passed all the tests locally

File: internal/compiler/program.go [hunk 2]
```diff
@@ -1222,6 +1223,47 @@ func (p *Program) getDiagnosticsHelper(ctx context.Context, sourceFile *ast.Sour
 	return SortAndDeduplicateDiagnostics(result)
 }
 
+func (p *Program) addProgramDiagnostic(diagnostic *ast.Diagnostic) {
+	p.programDiagnostics = append(p.programDiagnostics, diagnostic)
```
jakebailey: Can you just inline this? The other users of `p.programDiagnostics` do.

internal/compiler/program.go
```diff
@@ -1553,11 +1595,23 @@ func (p *Program) GetIncludeReasons() map[tspath.Path][]*FileIncludeReason {
 
 // Testing only
 func (p *Program) IsMissingPath(path tspath.Path) bool {
-	return slices.ContainsFunc(p.missingFiles, func(missingPath string) bool {
+	return slices.ContainsFunc(p.missingFilePaths(), func(missingPath string) bool {
 		return p.toPath(missingPath) == path
```
jakebailey: Couldn't this just be over missingPath still and then check `m.path` instead of us allocating a temp slice?

internal/compiler/program.go
```diff
@@ -1222,6 +1223,47 @@ func (p *Program) getDiagnosticsHelper(ctx context.Context, sourceFile *ast.Sour
...
+
+		parent := p.filesByPath[data.file]
+		if parent == nil {
+			continue
+		}
+
+		var ref *ast.FileReference
+
+		for _, r := range parent.ReferencedFiles {
+			if r.FileName == m.path {
```
jakebailey: This seems like a `find` call we have a helper for already.

internal/compiler/program_test.go
jakebailey: This is still an issue; the PR is not fixing the test I mentioned above

TranNhatHan: In the last commit, I only made changes to loaderfile.go and program.go. I had previously added a new test (TestMissingReferenceFile), but it has been removed following the your guidance. Please let me know if I have misunderstood your point.

jakebailey: If the fix in this PR worked, then the test case Ryan was showing, `declarationEmitIndexTypeNotFound`, would fail and ask you to update the baselines, and likely other tests. The fact that this PR contains code but no test baseline changes means the fix is not working.

Are you testing this some other way?

TranNhatHan: Oh, got it! so if the bug is fixed, the tests are expected to fail and I need to update the baseline, right?

Previously, I ran a test locally for debugging and compiled an erroneous .ts file to check whether the error matched what tsc would produce. After that, I ran go test, which failed. I then code again to pass the test but forgot to manually verify it again, so I just realized that my last commit didn’t actually solve the issue.

Now I have fixed the bug again and I’ve updated the baseline, but it resulted in 64 changes, and I’m not sure how to verify whether the updated baseline is correct.

jakebailey: It seems like an improvement with some regressions; if you view https://github.com/microsoft/typescript-go/pull/2096/files?file-filters%5B%5D=.go&file-filters%5B%5D=.diff (filtered to diffs), the goal would be to eliminate those by improving the fix.

TranNhatHan:
@jakebailey Please take a look. I'm still unsure whether the bug is actually fixed. When I compile with a not-found reference file, it behaves like tsc and the tests fail. After I accept the new baseline the tests pass, but no files are generated, and compiling with a not-found reference file still produces the same error as tsc.

TranNhatHan:
sorry it is not fixed yet

TranNhatHan:
I’ve come to realize this task is a bit beyond my current scope. I really appreciate all your help and feedback. I’ll close the pull request.

TranNhatHan:
I noticed that if a reference file is not found, only that error appears while other errors don’t, and some errors, like a file without an extension, are now reported as ‘file not found.’ Could this be because I called it too early? If this was fixed elsewhere, I’d appreciate it if you could let me know when. I’ve closed the PR but would like to understand how it was resolved.

PR #2097, opened 2025-11-14
Port baseline diagnostics tests

gabritto:
Follow up to #2079.

In Strada fourslash, we obtained compiler diagnostics directly and reused the compiler test harness functions for producing a baseline file for diagnostics. In Corsa fourslash, we get LSP diagnostics instead of `ast.Diagnostic`, and so to reuse the existing diagnostic baseline functions I had to refactor them to accept an interface with common methods for both the LSP and internal diagnostic types.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -2569,7 +2569,7 @@ func isSuggestionDiagnostic(diag *lsproto.Diagnostic) bool {
 }
 
 func (f *FourslashTest) VerifyBaselineNonSuggestionDiagnostics(t *testing.T) {
-	var diagnostics []*Diagnostic
+	var diagnostics []*fourslashDiagnostic
 	var files []*harnessutil.TestFile
 	for fileName, scriptInfo := range f.scriptInfos {
 		if tspath.HasJSONFileExtension(fileName) {
@@ -2580,7 +2580,7 @@ func (f *FourslashTest) VerifyBaselineNonSuggestionDiagnostics(t *testing.T) {
 			f.getDiagnostics(t, fileName),
 			func(d *lsproto.Diagnostic) bool { return !isSuggestionDiagnostic(d) },
 		)
-		diagnostics = append(diagnostics, core.Map(lspDiagnostics, func(d *lsproto.Diagnostic) *Diagnostic {
+		diagnostics = append(diagnostics, core.Map(lspDiagnostics, func(d *lsproto.Diagnostic) *fourslashDiagnostic {
 			return f.toDiagnostic(scriptInfo, d)
 		})...)
 	}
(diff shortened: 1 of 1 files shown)


internal/diagnosticwriter/diagnosticwriter.go
```diff
@@ -370,15 +455,15 @@ func prettyPathForFileError(file *ast.SourceFile, fileErrors []*ast.Diagnostic,
...
 	for _, diagnostic := range diagnostics {
 		WriteFormatDiagnostic(output, diagnostic, formatOpts)
 	}
 }
 
-func WriteFormatDiagnostic(output io.Writer, diagnostic *ast.Diagnostic, formatOpts *FormattingOptions) {
+func WriteFormatDiagnostic(output io.Writer, diagnostic Diagnostic, formatOpts *FormattingOptions) {
 	if diagnostic.File() != nil {
-		line, character := scanner.GetECMALineAndCharacterOfPosition(diagnostic.File(), diagnostic.Loc().Pos())
+		line, character := scanner.GetECMALineAndCharacterOfPosition(diagnostic.File(), diagnostic.Pos())
```
jakebailey, line 466: We haven't really vetted these calls yet, but I am worried that this is going to be wrong given this func accepts AST diags (UTF-8 offsets), LSP diags (UTF-16 by default?), and then `GetECMALineAndCharacterOfPosition` returns a rune count (UTF-32???) up to `Pos` for some reason.

We can probably kick the can down the road if it's not _crashing_ yet, but, yeah

internal/fourslash/fourslash.go
```diff
@@ -2556,11 +2543,217 @@ func (f *FourslashTest) verifyDiagnostics(t *testing.T, expected []*lsproto.Diag
...
+		})...)
+	}
+	slices.SortFunc(files, func(a, b *harnessutil.TestFile) int {
+		return strings.Compare(a.UnitName, b.UnitName)
+	})
+	result := tsbaseline.GetErrorBaseline(t, files, diagnostics, compareDiagnostics, false /*pretty*/)
+	f.addResultToBaseline(t, "Syntax and Semantic Diagnostics", result)
+}
+
+type Diagnostic struct {
```
jakebailey: Are these types ever used outside this package?

testdata/baselines/reference/fourslash/syntaxandSemanticDiagnostics/getJavaScriptSyntacticDiagnostics13.baseline
```diff
@@ -0,0 +1,8 @@
+// === Syntax and Semantic Diagnostics ===
+/a.js(1,8): error TS8010: Type annotations can only be used in TypeScript files.
```
jakebailey, line 2: Are these files diffable?

internal/diagnosticwriter/diagnosticwriter.go
```diff
@@ -16,6 +16,87 @@ import (
...
+	return result
+}
+
+func ToDiagnostics[T Diagnostic](diags []T) []Diagnostic {
+	result := make([]Diagnostic, len(diags))
+	for i, d := range diags {
+		result[i] = d
+	}
+	return result
+}
```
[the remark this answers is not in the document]
jakebailey: copilot doesn't know about variance, apparently

testdata/baselines/reference/fourslash/syntaxandSemanticDiagnostics/getJavaScriptSyntacticDiagnostics13.baseline
gabritto: Forgot to mention in my description. It's only 20 something tests, and would be annoying enough to diff because Strada distinguishes between syntactic and semantic diagnostics. I just visually diffed them.

internal/fourslash/fourslash.go
gabritto: No, I just keep forgetting the uppercase matters in Go. 🤦‍♀️

internal/diagnosticwriter/diagnosticwriter.go
gabritto: We are using UTF-8 encoding for LSP in tests when we set the server up, and the location on LSP diagnostics gets converted anyways, so at least I think this function would be equally wrong for both kinds of diagnostics right now, unless I'm missing something. But yes, this is an existing suspicious usage of the function.

jakebailey: Ah, if we are using UTF-8 in fourslash, then that that is good for this code, at least.

PR #2098, opened 2025-11-15
Fix hover on `module.exports`

ahejlsberg:
This PR fixes our odd behavior when hovering on `module.exports` in a JS file. Previously we'd display `{ "exports=": { ... } }`, now we display `module "xxx"` when hovering on `module` and `exports: { ... }` when hovering on `exports`. In addition, find-all-references on `module` finds all locations where the module is imported.

The PR also fixes the following panic:

```ts
module.exports = {}
exports.foo = 42 // Click on `exports` causes panic
```

### DIFF
--- a/internal/fourslash/tests/gen/findAllRefsModuleDotExports_test.go
+++ b/internal/fourslash/tests/gen/findAllRefsModuleDotExports_test.go
@@ -9,7 +9,7 @@ import (
 
 func TestFindAllRefsModuleDotExports(t *testing.T) {
 	t.Parallel()
-	t.Skip()
+
 	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
 	const content = `// @allowJs: true
 // @Filename: /a.js
--- a/testdata/baselines/reference/fourslash/findAllReferences/findAllRefsModuleDotExports.baseline.jsonc
+++ b/testdata/baselines/reference/fourslash/findAllReferences/findAllRefsModuleDotExports.baseline.jsonc
@@ -1,0 +1,21 @@
+// === findAllReferences ===
+// === /a.js ===
+// /*FIND ALL REFS*/const [|b|] = require("./b");
+
+
+
+// === findAllReferences ===
+// === /a.js ===
+// const b = require("/*FIND ALL REFS*/[|./b|]");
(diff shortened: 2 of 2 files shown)


ahejlsberg:
The many baseline changes are all because `module.exports` now resolves to the symbol for the module source file instead of a synthetic `module.exports` symbol with no associated declaration.

testdata/baselines/reference/submodule/compiler/checkJsTypeDefNoUnusedLocalMarked.symbols.diff
```diff
@@ -7,5 +7,5 @@
 ->module.exports : Symbol(module.exports, Decl(something.js, 0, 0))
 ->module : Symbol(export=, Decl(something.js, 0, 0))
 +>module.exports : Symbol(export=, Decl(something.js, 0, 0))
-+>module : Symbol(module.exports)
++>module : Symbol("something", Decl(something.js, 0, 0))
```
jakebailey, line 10: Kind of funky that before, `module.exports` was `module.exports` and `module` was `export=`, but now `module.exports` is `export=` and `module` is module name...

ahejlsberg: I think the new behavior is defensible. The old one--not so much.

jakebailey:
Seems like you fixed a test and just need to run `npm run updatefailing`.

ahejlsberg:
> Seems like you fixed a test and just need to run npm run updatefailing.

Tried, but didn't seem to do anything.

ahejlsberg:
Hmm, I'm not understanding what to do here! No changes are generated when I run `npm run updatefailing`.

jakebailey:
I think this is a bad interaction of the script and baselines (since generating new baselines also fails the test, and the CI check is smarter)

We can fix this manually for now; remove the test from the failing list, then also remove the t.Skip() call from the generated test itself, then if you run the test you can accept the baselines. Then you can rerun updatefailing again.

I'll look into fixing this properly later.

jakebailey (approved):

PR #2100, opened 2025-11-15
fix(2047): Incomplete declaration emit of callback tag with no return tag

### LINKED ISSUE #2047: Incomplete declaration emit of callback tag with no return tag
## Steps to reproduce
From

test: submodule/conformance/jsDeclarationsFunctionPrototypeStatic.ts
after #826 is merged

```js
/** @callback {Done}
 * @param {number} failures - the number of failures
 */
```

```ts
export type Done = (failures: number) => any;
```

## Behavior with `tsgo`

```ts
export type Done = (failures: number) ;
```

a-tarasyuk:
Fixes #2047


jakebailey (approved):
I feel like this is acceptable for `@callback` without any return. `void` is probably bad. Perhaps this needs an error?

PR #2111, opened 2025-11-17
Display inherited JSDoc documentation in quick info

### LINKED ISSUE #2035: Classes and interfaces lose hover JSDoc comments from parent class/interface when re-declaring properties
### Extension Version

v0.20251107.1 

### VS Code Version

1.105.1

Also occurs when using TSGo with the Zed editor (extension version `0.0.3` and editor version `0.211.6`), so it's likely not a VS Code impl issue.

### Operating system Version

Ubuntu x86-64 (WSL, though I highly doubt this matters)

### Steps to reproduce

Copy the following MRE into a new `.ts` file in the editor.
```ts
interface foo {
  /** base jsdoc */
  bar(k: string): number;
  /** other jsdoc */
  other: 24;
}

interface bar extends foo {
  bar(k: string | symbol): number | 99;
}

declare class thing {
  /** doc comment */
  method(s: string): void;
}

declare class potato extends thing {
  method(s: "1234"): void;
}

declare const f: foo;
declare const b: bar;
declare const t: thing;
declare const p: potato;

// b and p lack comments on the properties they overrode (`bar` and `method` respectively)
f.bar;
f.other;
b.bar;
b.other;
t.method;
p.method;
```

Hover over the bottom 6 symbols and notice how `b` and `p`'s overridden methods lack hover comments.

### Issue

The interfaces and classes should inherit the parents' doc comments if they themself lack an accompanying JSDoc comment, as happens with normal `tsc`.

NB: This also occurs with classes implementing interfaces and interfaces extending classes.

### Images
Playground = tsc, editor = tsgo

Image

Image

Image

Image

### Remarks
Our codebase uses this feature a very large amount to avoid re-documenting interface properties on implementing classes, so a fix would be greatly appreciated.

ahejlsberg:
In this PR:

* Display inherited JSDoc documentation in quick info.
* Display JSDoc for actual signature when hovering over an overload or implementation signature.
* Consistently display constructor signature when hovering over `new SomeClass(...)` expression.

Fixes #2035.


internal/fourslash/tests/manual/quickInfoForOverloadOnConst1_test.go
```diff
@@ -31,7 +31,7 @@ c.x1(1, (x/*10*/x) => { return 1; } );`
 	f.VerifyQuickInfoAt(t, "1", "(method) I.x1(a: number, callback: (x: \"hi\") => number): any", "")
 	f.VerifyQuickInfoAt(t, "2", "(method) C.x1(a: number, callback: (x: \"hi\") => number): any", "")
 	f.VerifyQuickInfoAt(t, "3", "(parameter) callback: (x: \"hi\") => number", "")
-	f.VerifyQuickInfoAt(t, "4", "(method) C.x1(a: number, callback: (x: \"hi\") => number): any", "")
+	f.VerifyQuickInfoAt(t, "4", "(method) C.x1(a: number, callback: (x: string) => number): void", "")
```
jakebailey, line 34: You can't modify a generated test like this since it'll be overweitten; you have to do `npm run makemanual TestQuickInfoForOverloadOnConst1` to exempt the test from being a generated one then update the version moved to the "manual" dir where we put generated tests we manually modify.

jakebailey: Though I'm not sure why this is changing...

ahejlsberg: It's changing because we now display the actual signature and documentation for the declaration being hovered on. The Strada language service displays the signature of the first overload when hovering over the implementation which doesn't make much sense.

PR #2116, opened 2025-11-17
Update dependencies

jakebailey:
Largely to fix audit issues, but also because it's good.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,12 @@ jobs:
       TSGO_HEREBY_NOEMBED: ${{ (matrix.config.noembed && 'true') || 'false' }}
 
     steps:
+      - uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+        with:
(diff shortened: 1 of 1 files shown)


jakebailey:
I do not understand at all why `generate` consistently fails in CI ☹️

andrewbranch (approved):

.github/workflows/ci.yml
```diff
@@ -233,6 +233,12 @@ jobs:
   generate:
     runs-on: ubuntu-latest
     steps:
+      - uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
```
DanielRosenwasser, line 242: Why do we need this now?

jakebailey: On PRs that update dependencies, the caches that `setup-go` pulls from main are basically unused and therefore CI doubles up in `GOCACHE` so any job that runs all the tests needs a bit more space.

jakebailey:
ugh linting also needs the space freeing

PR #2118, opened 2025-11-18
Provide Program diagnostics as push diags in tsconfig.json

jakebailey:
This uses LSP push diags to send `ProgramDiagnostics` when present, removing them when a project is closed. The actual implementation of it seems fine, but the testing is pretty annoying.

There are unit tests for this, but fourslash cannot handle client-initiated anything except the two hardcoded init ones. So, I'm just forcibly disabling those via an init option set for fourslash. Nasty but making fourslash more like a real LSP client is going to take a lot of work.

Also, we have program diags that are not associated with files at all. I've just hacked it so that these diags get reported at a zero position.

### DIFF
--- a/internal/testutil/projecttestutil/projecttestutil.go
+++ b/internal/testutil/projecttestutil/projecttestutil.go
@@ -209,12 +209,13 @@ func SetupWithOptionsAndTypingsInstaller(files map[string]any, options *project.
 	// Use provided options or create default ones
 	if options == nil {
 		options = &project.SessionOptions{
-			CurrentDirectory:   "/",
-			DefaultLibraryPath: bundled.LibPath(),
-			TypingsLocation:    TestTypingsLocation,
-			PositionEncoding:   lsproto.PositionEncodingKindUTF8,
-			WatchEnabled:       true,
-			LoggingEnabled:     true,
+			CurrentDirectory:       "/",
+			DefaultLibraryPath:     bundled.LibPath(),
+			TypingsLocation:        TestTypingsLocation,
+			PositionEncoding:       lsproto.PositionEncodingKindUTF8,
+			WatchEnabled:           true,
+			LoggingEnabled:         true,
+			PushDiagnosticsEnabled: true,
 		}
 	}
 


internal/lsp/server.go
```diff
@@ -726,43 +743,36 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
...
+		NpmExecutor:            s,
+		ParseCache:             s.parseCache,
+		DisablePushDiagnostics: disablePushDiagnostics,
 	})
 
-	if s.initializeParams != nil && s.initializeParams.InitializationOptions != nil && *s.initializeParams.InitializationOptions != nil {
-		// handle userPreferences from initializationOptions
-		userPreferences := s.session.NewUserPreferences()
-		userPreferences.Parse(*s.initializeParams.InitializationOptions)
-		s.session.InitializeWithConfig(userPreferences)
```
jakebailey, line 740: this code was unused

internal/project/snapshot_test.go
```diff
@@ -11,6 +11,25 @@ import (
 	"gotest.tools/v3/assert"
 )
 
+// noopClient is a minimal Client implementation for snapshot tests
+type noopClient struct{}
```
jakebailey: `ClientMock` creates an import cycle that's annoying to resolve, so I just made a noop here for the lower-level tests.

andrewbranch:
Can a single-file edit that results in a fast program clone affect program diagnostics? If not, we could gate the collection on `project.ProgramUpdateKind > ProgramUpdateKindCloned` to avoid doing it on every keystroke.

jakebailey:
No, I don't believe it can, since we check if a file's important bits change before doing the easy fast clone for that reason. I can do that.

andrewbranch (approved):
🎉

jakebailey:
This is still failing the API tests, I'm trying to figure out why ☹️

jakebailey:
It's the client nil check; just dropped that since there's a flag that enables the push diags... the API is a little weird given it's sort of a language server sort of not.

PR #2123, opened 2025-11-18
Enable localization

jakebailey:
This PR enables localization, both for CLI with `--locale`, but also for in the language server via the client-provided locale init option.

To make this happen, we needed to stop eagerly formatting diagnostic messages. The reason this worked in the old codebase was because we used a global locale, so any time we made a diag, it happened to be produce after the global locale was set.

Doing that again sounds like an utter nightmare to test, so instead I make diagnostics remember their "message key" and stringified args. Then, when we finally go to produce the diagnostics, we then use the locale available to us to finalize the messsage.

Interestingly, this also makes `tsbuildinfo` locale-invariant, which is nice.

I also had to sort of redo the JSX element message hack; we were eagerly resolving a message to remove its placeholders, turning that into _another_ `*diagnostic.Message`, and then passing that into relations. That doesn't work so well, so it's been replaced with something that seems to produce equivalent errors.

Localization files are currently just embedded into the binary. I would like to use a system similar to what we do with `.d.ts` files, but there's a notable difference between them. The lib files need to be in the filesystem, so we use an FS wrapper to provide them, or use a real disk path. But for diagnostics, we really don't want to have to lug an FS around to where diags are localized (at least, that seems bad), so reusing the existing infra is not so great.

The localization files are `.json.gz` files; generating code like we do the diagnostic objects themselves, or even embedding `.json` files in a similar format to the old codebase would increase the binary size by some 5MB. No good!

What is here I think is the least bad balance until I can refactor the bundled-file system to better handle localizations.

### DIFF
--- a/internal/execute/tsc/emit.go
+++ b/internal/execute/tsc/emit.go
@@ -22,7 +22,7 @@ func GetTraceWithWriterFromSys(w io.Writer, locale locale.Locale, testing Comman
 			fmt.Fprintln(w, msg.Localize(locale, args...))
 		}
 	} else {
-		return testing.GetTrace(w)
+		return testing.GetTrace(w, locale)
 	}
 }
 
--- a/internal/execute/tsc/compile.go
+++ b/internal/execute/tsc/compile.go
@@ -9,6 +9,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/compiler"
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 	"github.com/microsoft/typescript-go/internal/execute/incremental"
+	"github.com/microsoft/typescript-go/internal/locale"
 	"github.com/microsoft/typescript-go/internal/tspath"
 	"github.com/microsoft/typescript-go/internal/vfs"
 )
@@ -57,7 +58,7 @@ type CommandLineTesting interface {
 	OnBuildStatusReportEnd(w io.Writer)
 	OnWatchStatusReportStart()
 	OnWatchStatusReportEnd()
-	GetTrace(w io.Writer) func(msg *diagnostics.Message, args ...any)
+	GetTrace(w io.Writer, locale locale.Locale) func(msg *diagnostics.Message, args ...any)
 	OnProgram(program *incremental.Program)
 }
 
--- a/internal/execute/tsctests/sys.go
+++ b/internal/execute/tsctests/sys.go
@@ -288,13 +288,13 @@ func (s *TestSys) OnWatchStatusReportEnd() {
 	fmt.Fprintln(s.Writer(), watchStatusReportEnd)
 }
 
-func (s *TestSys) GetTrace(w io.Writer) func(msg *diagnostics.Message, args ...any) {
+func (s *TestSys) GetTrace(w io.Writer, locale locale.Locale) func(msg *diagnostics.Message, args ...any) {
 	return func(msg *diagnostics.Message, args ...any) {
 		fmt.Fprintln(w, traceStart)
 		defer fmt.Fprintln(w, traceEnd)
 		// With tsc -b building projects in parallel we cannot serialize the package.json lookup trace
 		// so trace as if it wasnt cached
-		str := msg.Localize(locale.Default, args...)
+		str := msg.Localize(locale, args...)
 		s.tracer.TraceWithWriter(w, str, w == s.Writer())
 	}
 }


internal/ast/diagnostic.go
```diff
@@ -185,13 +203,15 @@ func EqualDiagnosticsNoRelatedInfo(d1, d2 *Diagnostic) bool {
 	return getDiagnosticPath(d1) == getDiagnosticPath(d2) &&
 		d1.Loc() == d2.Loc() &&
 		d1.Code() == d2.Code() &&
-		d1.Message() == d2.Message() &&
+		d1.MessageKey() == d2.MessageKey() &&
```
jakebailey: Within the same exe, I don't even think it's possible for the key to differ (the generator asserts that codes are unique), so probably this and other checks can go away.

internal/compiler/program.go
```diff
@@ -1565,10 +1565,10 @@ func (p *Program) ExplainFiles(w io.Writer) {
 	for _, file := range p.GetSourceFiles() {
 		fmt.Fprintln(w, toRelativeFileName(file.FileName()))
 		for _, reason := range p.includeProcessor.fileIncludeReasons[file.Path()] {
-			fmt.Fprintln(w, "  ", reason.toDiagnostic(p, true).Message())
+			fmt.Fprintln(w, "  ", reason.toDiagnostic(p, true).Localize(p.Host().Locale()))
```
jakebailey: This might be the only reason we even need the locale here (or in CompilerHost); I should recheck this as it could be that this should live outside the program.

jakebailey: It isn't; resolve/file loading tracing prints stuff, and undoing that gets very complicated.

jakebailey: I can try, however; the gotcha here is not this code (this is easy to refactor), but rather just resolution tracing, which I can defer like diags but might be a little invasive...

internal/execute/build/compilerHost.go
```diff
@@ -27,6 +28,10 @@ func (h *compilerHost) GetCurrentDirectory() string {
 	return h.host.GetCurrentDirectory()
 }
 
+func (h *compilerHost) Locale() locale.Locale {
```
sheetalkamat: Wondering if instead of plumbing it through the users of program can convert and write with correct locale? 

Eg: Conver the compilerHost.Trace to take diagnpostic

and writers of diagnostic actually `GetTraceWithWriterFromSys` is the one that writes the correct locale string?
Same for diagnostic printer ?

jakebailey: Yes, I have about 90% of that change already as I was trying to: https://github.com/microsoft/typescript-go/pull/2123#discussion_r2543978606

internal/compiler/program.go
jakebailey: Done, much cleaner, I think?

internal/execute/build/compilerHost.go
jakebailey: Done, much cleaner, I think?

jakebailey: The main problem is that there's not a convenient function that just "logs a string" anymore

internal/tsoptions/parsedcommandline.go
```diff
@@ -379,3 +383,10 @@ func (p *ParsedCommandLine) ReloadFileNamesOfParsedCommandLine(fs vfs.FS) *Parse
 	}
 	return &parsedCommandLine
 }
+
+func (p *ParsedCommandLine) Locale() locale.Locale {
+	p.localeOnce.Do(func() {
+		p.locale = locale.Parse(p.CompilerOptions().Locale)
```
sheetalkamat: Do we need to store errors from parsing and report those like strada did?

jakebailey: We could but honestly, I don't know if it's worth it. The error message given back by Go's parser is unlikely to itself be dependable, though I guess we could error with something more generic.

If anything, though, I would rather do that in a follow-up, unless you know the best place off the top of your head to do that.

sheetalkamat: I am looking at these errors
Locale_must_be_of_the_form_language_or_language_territory_For_example_0_or_1 - shouldnt this be reported. but then this could be options parsing error ?

// Confirm if I am wrong in assuming that below ones are unlikely since the files will be bundled?
Unable_to_open_file_0
Corrupted_locale_file_0

jakebailey: I can add that first one in; I see where it's reasonable to do so now.

The latter ones cannot happen, at least not yet.

jakebailey: Updated with the error message, with tests.

internal/execute/tsc/emit.go
```diff
@@ -10,14 +10,16 @@ import (
...
-func GetTraceWithWriterFromSys(w io.Writer, testing CommandLineTesting) func(msg string) {
+func GetTraceWithWriterFromSys(w io.Writer, locale locale.Locale, testing CommandLineTesting) func(msg *diagnostics.Message, args ...any) {
 	if testing == nil {
-		return func(msg string) {
-			fmt.Fprintln(w, msg)
+		return func(msg *diagnostics.Message, args ...any) {
+			fmt.Fprintln(w, msg.Localize(locale, args...))
 		}
 	} else {
 		return testing.GetTrace(w)
```
sheetalkamat: Pass the locale here ? to testing just like writer so it uses correct locale ?

internal/execute/tsctests/tsc_test.go
```diff
@@ -205,6 +205,14 @@ func TestTscCommandline(t *testing.T) {
 			},
 			commandLineArgs: []string{"-p", "."},
 		},
+		{
+			subScenario:     "locale",
+			commandLineArgs: []string{"--locale", "en", "--version"},
```
sheetalkamat: Test with something other than en would be good too

sheetalkamat (approved):

PR #2130, opened 2025-11-19
Port tests for go to implementation and diff definitions tests

gabritto:
In this PR:
- Port of baseline go to implementation fourslash tests.
- Baseline diffing for go to definition, go to type definition, and go to implementation.
- Include lib files in baselines (+ fix to embed's `walkDir`)
- Re-add support for printing context spans (i.e. `LocationLink`'s target range) in baselines.

Note: I also added a method to fourslash tests in Corsa to mark them as originally server tests, because due to a bug in the fourslash server test harness, we don't include context spans in server tests, so I added the same special behavior (which we can remove later) to make diffing easier.

### DIFF
--- a/internal/bundled/embed.go
+++ b/internal/bundled/embed.go
@@ -131,7 +131,7 @@ func (vfs *wrappedFS) walkDir(rest string, walkFn vfs.WalkDirFunc) error {
 			return err
 		}
 		if entry.IsDir() {
-			if err := vfs.walkDir(entry.Name(), walkFn); err != nil {
+			if err := vfs.walkDir(strings.TrimPrefix(name, "/"), walkFn); err != nil {
 				return err
 			}
 		}


testdata/baselines/reference/submodule/fourslash/goToType/goToTypeDefinition4.baseline.jsonc.diff
```diff
@@ -0,0 +1,8 @@
+--- old.goToTypeDefinition4.baseline.jsonc
++++ new.goToTypeDefinition4.baseline.jsonc
+@@= skipped -1, +1 lines =@@
+ // === /bar.ts ===
+ // import { T } from "./foo";
+ // let x: /*GOTO TYPE*/T;
+-
+-
```
jakebailey: Wonder if this can be worked around

gabritto: I'll try to fix it, I just happened to not have looked at this particular diff before.

jakebailey: I just so happened to randomly click on one, I guess 😄

internal/bundled/embed.go
```diff
@@ -131,7 +131,7 @@ func (vfs *wrappedFS) walkDir(rest string, walkFn vfs.WalkDirFunc) error {
 			return err
 		}
 		if entry.IsDir() {
-			if err := vfs.walkDir(name, walkFn); err != nil {
```
jakebailey, line 134: Why is this needed? Isn't `entry.name()` the basename of the path, whereas walkDir is expecting a full path?

gabritto: If I remember correctly, starting from `rest` = `""`, we were never actually visiting the lib files, because when we got to this line, `name` = `"/libs"`, and then we hit the `default: return nil` case of the recursive call here.

gabritto: Basically we were calling `vfs.walkDir("/libs", ...)`, which doesn't work, and we need to call `vfs.walkDir("libs", ...)` instead. There might be a less weird way to do that though.

jakebailey: Well, the trouble would be if we had nested dirs, like `/foo/bar/libs`, we just happen to not have those. Maybe this just needs a prefix trim?

PR #2142, opened 2025-11-21
Place DO NOT EDIT comment on top of generated fourslash tests

jakebailey:
These files are not meant to be edited by hand. Placing this special string at the top tells all Go-knowing tooling that these are generated and should be ignored, hides their diffs in GitHub diffs, tells `gopls` to issue a warning when you touch these files, and probably helps Copilot know it shouldn't touch these files in general.


gabritto:
Did the conversion script fail to format?

jakebailey:
I think it did, because these are generated files and so are not formatted, I guess.

I don't think this is worth the churn. Probably we should just update the copilot instructions.

PR #2144, opened 2025-11-21
Display all symbol meanings in quick info

ahejlsberg:
With this PR we display all meanings of a symbol in Quick Info. An `(alias)` prefix is displayed for non-local meanings that arise from `import` and `export` aliases.

image

### DIFF
--- a/internal/ls/hover.go
+++ b/internal/ls/hover.go
@@ -9,6 +9,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/astnav"
 	"github.com/microsoft/typescript-go/internal/checker"
+	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
 )
@@ -135,6 +136,7 @@ func formatQuickInfo(quickInfo string) string {
 
 func getQuickInfoAndDeclarationAtLocation(c *checker.Checker, symbol *ast.Symbol, node *ast.Node) (string, *ast.Node) {
 	var b strings.Builder
+	var visitedAliases collections.Set[*ast.Symbol]
 	container := getContainerNode(node)
 	if node.Kind == ast.KindThisKeyword && ast.IsInExpressionContext(node) {
 		return c.TypeToStringEx(c.GetTypeAtLocation(node), container, typeFormatFlags), nil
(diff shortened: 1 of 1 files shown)


jakebailey:
This fixes a massive number of fourslash tests that just need to be regenerated; unfortunately you'll probably need to do the goofy manual fix for `getOccurrencesIfElseBroken` if that isn't fixed by a plain `npm run updatefailing`.

I haven't had time to go rewrite these scripts to fix this yet (sorry...)

jakebailey:
Hm, it maybe it doesn't fix a bunch of tests. It seems like it's actually breaking something to do with the test fix detection. I am now looking into this.

internal/ls/hover.go
```diff
@@ -131,149 +134,163 @@ func formatQuickInfo(quickInfo string) string {
...
-			b.WriteString(" = ")
-			b.WriteString(c.TypeToStringEx(c.GetDeclaredTypeOfSymbol(symbol), container, typeFormatFlags|checker.TypeFormatFlagsInTypeAlias))
+		return declaration
+	}
+	var writeSymbol func(*ast.Symbol, bool) *ast.Node
+	writeSymbol = func(symbol *ast.Symbol, isAlias bool) *ast.Node {
+		var declaration *ast.Node
+		// Recursively write all meanings of alias
+		if symbol.Flags&ast.SymbolFlagsAlias != 0 {
+			if aliasedSymbol := c.GetAliasedSymbol(symbol); aliasedSymbol != c.GetUnknownSymbol() {
```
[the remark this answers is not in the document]
jakebailey: GetAliasedSymbol can never return nil

internal/ls/hover.go
```diff
@@ -131,149 +134,163 @@ func formatQuickInfo(quickInfo string) string {
...
-	default:
-		b.WriteString(c.TypeToStringEx(c.GetTypeOfSymbol(symbol), container, typeFormatFlags))
-	}
-	return b.String(), declaration
+		// Write the value meaning, if any
+		declaration = core.OrElse(declaration, writeSymbolMeaning(symbol, ast.SymbolFlagsValue|ast.SymbolFlagsSignature, isAlias))
+		// Write the type meaning, if any
+		declaration = core.OrElse(declaration, writeSymbolMeaning(symbol, ast.SymbolFlagsType&^ast.SymbolFlagsValue, isAlias))
+		// Write the namespace meaning, if any
+		declaration = core.OrElse(declaration, writeSymbolMeaning(symbol, ast.SymbolFlagsNamespace&^ast.SymbolFlagsValue, isAlias))
```
jakebailey, line 290: Is unconditionally calling writeSymbolMeaning intentional? I think it is, but assumed this was a bug when I read it 😄

jakebailey:
Ah, the fourslash code is breaking because `TestAliasMergingWithNamespace` when unskipped causes a stack overflow. Probably should fix that lest a user hit it.

internal/ls/hover.go
ahejlsberg: It's intentional. It only writes something if the symbol has the given meaning.

jakebailey (approved):

PR #2145, opened 2025-11-21
Implement support for Code Lenses

### LINKED ISSUE #1897: Add support for references code lenses in VS Code extension
When switching to tsgo, the references code lenses are a noticeable missing feature

I believe this feature can be implemented entirely on the extension side. Today we use `navtree` and `references` requests to implement them. We could do the same with tsgo or they could be implemented natively using the LSP

DanielRosenwasser:
This PR implements support for `textDocument/codeLens` and `codeLens/resolve` on the server side, alongside a custom client command called `typescript.codeLens.showLocations`.

`CodeLens` objects are produced by `textDocument/codeLens` as indicators that an editor client can get more information "later on". When `codeLens/resolve` is called for a given `CodeLens` object, the `Title`, `Command`, and command `Arguments` are populated appropriately so that a user can see some high-level information and potentially click on the UI decoration to trigger a command.

We produce two types of `CodeLens` objects: "references" code lens, and "implementations" code lens. These act as markers/indicators so that when an editor like VS Code scrolls into the view of a code lens, it will call on the server to populate each lens with a title and list of locations to preview upon clicking. This amounts to a find-all-references or go-to-implementation call when scrolling into the view of a given code lens.

The VS Code extension wires this up with a minimal client-side command which just converts the LSP objects into VS Code ones before redirecting the values to the editor's "peek" pane.

Much of the logic is roughly ported over from [VS Code's Code Lens implementations](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features/src/languageFeatures/codeLens), though it is not one-to-one. On an implementation level, instead of two separate providers/walks like the client did, this implementation does one walk and may produce up to two `CodeLens` objects per implementation, each which encodes how the `CodeLens` should be resolved later. Since we have full syntactic and semantic information, we can make better judgments on what kinds of syntax should be decorated with a `CodeLens` if we ever want to modify the behavior.

That said, we have some differences - some that I think are desirable, and some that are just bugs.

side-by-side comparison of client-side codelens versus server-side in tsgo

One thing I implemented here was support for `includeDeclaration` in `textDocument/references` (aside: I find this behavior iffy for callers who "forget" to pass in `includeDeclaration`). I used this as a shortcut to avoid filtering out declarations in find-all-references; however, as you can see above, this gets a different "References" count because it drops overrides/implementations. I actually think that this might not be the right move, and that maybe we should only discount references containing the source position.

Current TODOs:

* [x] Add testing capabilities and some basic tests
* [x] Sweep up TODOs (there's one left for type aliases, which weren't previously handled anyway)
* [x] Fix up overloads
* [x] Decide on how references filtering works for declarations (this is just a `!!!` for now)
* [x] Fix up how `Data` is handled given Jake's PR at https://github.com/microsoft/typescript-go/pull/2141
* [x] Do some sort of client/server handshake to decide on the client-side command name for `codeLens/resolve`.

Fixes #1897

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -1469,7 +1469,7 @@ func getReferencedSymbolsForSymbol(originalSymbol *ast.Symbol, node *ast.Node, s
 		// When renaming at an export specifier, rename the export and not the thing being exported.
 		// state.getReferencesAtExportSpecifier(exportSpecifier.Name(), symbol, exportSpecifier.AsExportSpecifier(), state.createSearch(node, originalSymbol, comingFromUnknown /*comingFrom*/, "", nil), true /*addReferencesHere*/, true /*alwaysGetReferences*/)
 	} else if node != nil && node.Kind == ast.KindDefaultKeyword && symbol.Name == ast.InternalSymbolNameDefault && symbol.Parent != nil {
-		state.addReference(node, symbol, entryKindNone)
+		state.addReference(node, symbol, entryKindNode)
 		state.searchForImportsOfExport(node, symbol, &ExportInfo{exportingModuleSymbol: symbol.Parent, exportKind: ExportKindDefault})
 	} else {
 		search := state.createSearch(node, symbol, ImpExpKindUnknown /*comingFrom*/, "", state.populateSearchSymbolSet(symbol, node, options.use == referenceUseRename, options.useAliasesForRename, options.implementations))
@@ -1621,33 +1621,12 @@ func getReferenceEntriesForShorthandPropertyAssignment(node *ast.Node, checker *
 	}
 }
 
-func climbPastPropertyAccess(node *ast.Node) *ast.Node {
-	if ast.IsRightSideOfPropertyAccess(node) {
-		return node.Parent
-	}
-	return node
-}
-
-func isNewExpressionTarget(node *ast.Node) bool {
-	if node.Parent == nil {
-		return false
-	}
-	return node.Parent.Kind == ast.KindNewExpression && node.Parent.Expression() == node
-}
-
-func isCallExpressionTarget(node *ast.Node) bool {
-	if node.Parent == nil {
-		return false
-	}
-	return node.Parent.Kind == ast.KindCallExpression && node.Parent.Expression() == node
-}
-
 func isMethodOrAccessor(node *ast.Node) bool {
 	return node.Kind == ast.KindMethodDeclaration || node.Kind == ast.KindGetAccessor || node.Kind == ast.KindSetAccessor
 }
 
 func tryGetClassByExtendingIdentifier(node *ast.Node) *ast.ClassLikeDeclaration {
-	return ast.TryGetClassExtendingExpressionWithTypeArguments(climbPastPropertyAccess(node).Parent)
+	return ast.TryGetClassExtendingExpressionWithTypeArguments(ast.ClimbPastPropertyAccess(node).Parent)
 }
 
 func getClassConstructorSymbol(classSymbol *ast.Symbol) *ast.Symbol {
@@ -1680,7 +1659,7 @@ func findOwnConstructorReferences(classSymbol *ast.Symbol, sourceFile *ast.Sourc
 				body := decl.Body()
 				if body != nil {
 					forEachDescendantOfKind(body, ast.KindThisKeyword, func(thisKeyword *ast.Node) {
-						if isNewExpressionTarget(thisKeyword) {
+						if ast.IsNewExpressionTarget(thisKeyword, false, false) {
 							addNode(thisKeyword)
 						}
 					})
@@ -1701,7 +1680,7 @@ func findSuperConstructorAccesses(classDeclaration *ast.ClassLikeDeclaration, ad
 			body := decl.Body()
 			if body != nil {
 				forEachDescendantOfKind(body, ast.KindSuperKeyword, func(node *ast.Node) {
-					if isCallExpressionTarget(node) {
+					if ast.IsCallExpressionTarget(node, false, false) {
 						addNode(node)
 					}
 				})
@@ -1902,8 +1881,8 @@ func (state *refState) getReferencesAtLocation(sourceFile *ast.SourceFile, posit
 }
 
 func (state *refState) addConstructorReferences(referenceLocation *ast.Node, symbol *ast.Symbol, search *refSearch, addReferencesHere bool) {
-	if isNewExpressionTarget(referenceLocation) && addReferencesHere {
-		state.addReference(referenceLocation, symbol, entryKindNone)
+	if ast.IsNewExpressionTarget(referenceLocation, false, false) && addReferencesHere {
+		state.addReference(referenceLocation, symbol, entryKindNode)
 	}
 
 	pusher := func() func(*ast.Node, entryKind) {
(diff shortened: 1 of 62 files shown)


_extension/src/client.ts
```diff
@@ -119,10 +122,32 @@ export class Client {
         await this.client.start();
         vscode.commands.executeCommand("setContext", "typescript.native-preview.serverRunning", true);
         this.onStartedCallbacks.forEach(callback => callback());
-        return new vscode.Disposable(() => {
-            if (this.client) {
-                this.client.stop();
+
+        const codeLensLocationsCommand = vscode.commands.registerCommand("typescript.codeLens.showLocations", (...args: unknown[]) => {
```
jakebailey: Is this needed? We could just run `editor.action.showReferences` directly if the editor is VS Code.

internal/ls/codelens.go
```diff
@@ -0,0 +1,224 @@
...
+func GetCodeLensData(item *lsproto.CodeLens) (*CodeLensData, error) {
+	bytes, err := json.Marshal(item.Data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal completion item data: %w", err)
+	}
+	var itemData CodeLensData
+	if err := json.Unmarshal(bytes, &itemData); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal completion item data: %w", err)
+	}
+	return &itemData, nil
```
jakebailey: You should now be able to update `convert.mts` with your info.

_extension/src/client.ts
DanielRosenwasser: We cannot because that command concretely expects a `vscode.Uri`, a `vscode.Position`, and `vscode.Location`s. If you try to invoke it with plain LSP objects, it will error.

jakebailey: Reallyyyyyyyy 🫠

jakebailey: In that case, we should probably an an init option that is set only in our extension to say "this command will be supported", then conditionally enable code lens support in the handshake...

jakebailey:
Thankfully this is opt-in, otherwise I'd be even more worried about merging this given this _also_ uses the checker.

_extension/package.json
```diff
@@ -91,6 +91,11 @@
                 "title": "Report Issue",
                 "enablement": "typescript.native-preview.serverRunning",
                 "category": "TypeScript Native Preview"
+            },
+            {
+                "title": "Show References of CodeLens",
+                "command": "typescript.codeLens.showLocations",
```
jakebailey: Is this duplicated? Probably we should use one scoped to this extension given we don't want someone else handling this, right?

internal/lsp/server.go
```diff
@@ -971,6 +976,20 @@ func (s *Server) handleInlayHint(
 	return languageService.ProvideInlayHint(ctx, params)
 }
 
+func (s *Server) handleCodeLens(ctx context.Context, ls *ls.LanguageService, params *lsproto.CodeLensParams) (lsproto.CodeLensResponse, error) {
+	return ls.ProvideCodeLenses(ctx, params.TextDocument.Uri)
+}
+
+func (s *Server) handleCodeLensResolve(ctx context.Context, codeLens *lsproto.CodeLens, reqMsg *lsproto.RequestMessage) (*lsproto.CodeLens, error) {
```
jakebailey, line 1304: This is likely to conflict with Sheetal's PR, but probably in a good way given the codelens needs to do cross-project too.

_extension/package.json
DanielRosenwasser: This command doesn't exist in the original extension - the VS Code side would do it all client-side and trigger `editor.action.showReferences`.

jakebailey: Ah, sure.

DanielRosenwasser: The command is now namespaced to the extension.

_extension/src/client.ts
DanielRosenwasser: Now we do - the server will use `codeLensShowLocationsCommandName` if provided. Code lenses will still be presented either way, but they functionally won't do anything except display the count.

internal/lsp/server.go
```diff
@@ -692,6 +694,9 @@ func (s *Server) handleInitialize(ctx context.Context, params *lsproto.Initializ
 			InlayHintProvider: &lsproto.BooleanOrInlayHintOptionsOrInlayHintRegistrationOptions{
 				Boolean: ptrTo(true),
 			},
+			CodeLensProvider: &lsproto.CodeLensOptions{
+				ResolveProvider: ptrTo(true),
```
jakebailey, line 955: This should be conditional on `CodeLensShowLocationsCommandName`, otherwise all clients will get lenses they can't resolve, right?

DanielRosenwasser: No, you can resolve without a command. So you can still display the count even if your editor doesn't have support for executing this command.

DanielRosenwasser: This already happens when you have `0 references` displayed.

internal/lsp/server.go
DanielRosenwasser: I'm noticing that that PR (https://github.com/microsoft/typescript-go/pull/1991) doesn't touch go-to-implementations, which in theory is also multi-project. I think that that should be a follow-up to this so the two have a common way to walk references across related projects.

sheetalkamat: I can do that part as a separate PR but strada didn’t do it so I didn’t port it in my earlier pr

internal/ls/codelens.go
```diff
@@ -0,0 +1,218 @@
...
+			Uri:  fileUri,
+		},
+	}
+}
+
+func isValidImplementationsCodeLensNode(node *ast.Node, userPrefs *lsutil.UserPreferences) bool {
+	switch node.Kind {
+	// Always show on interfaces
+	case ast.KindInterfaceDeclaration:
+		// TODO: ast.KindTypeAliasDeclaration?
```
DanielRosenwasser, line 179: Probably won't address this here. VS Code just didn't show them on type aliases.

internal/ls/findallreferences.go
```diff
@@ -599,11 +599,24 @@ func (l *LanguageService) ProvideSymbolsAndEntries(ctx context.Context, uri lspr
...
 
 func (l *LanguageService) ProvideImplementations(ctx context.Context, params *lsproto.ImplementationParams) (lsproto.ImplementationResponse, error) {
+	return l.provideImplementationsEx(ctx, params, provideImplementationsOpts{})
+}
+
+type provideImplementationsOpts struct {
+	// Force the result to be Location objects.
+	requireLocationsResult bool
+	// Omit node(s) containing the original position.
+	dropOriginNodes bool
```
DanielRosenwasser, line 616: Part of me wonders if I should just do the filtering after the fact. That's what the client used to do, and I'm not sure if `IncludeDeclarations` was implemented correctly here.

internal/ls/findallreferences.go
```diff
@@ -599,11 +599,24 @@ func (l *LanguageService) ProvideSymbolsAndEntries(ctx context.Context, uri lspr
...
 	return lsproto.LocationsOrNull{Locations: &locations}, nil
 }
 
 func (l *LanguageService) ProvideImplementations(ctx context.Context, params *lsproto.ImplementationParams) (lsproto.ImplementationResponse, error) {
+	return l.provideImplementationsEx(ctx, params, provideImplementationsOpts{})
+}
+
+type provideImplementationsOpts struct {
+	// Force the result to be Location objects.
+	requireLocationsResult bool
```
DanielRosenwasser, line 614: Instead of any options bag, I could just set up a new context for this one.

internal/ls/codelens.go
```diff
@@ -0,0 +1,230 @@
...
+	if sourceFile == nil ||
+		l.converters.PositionToLineAndCharacter(sourceFile, core.TextPos(sourceFile.End())).Line < codeLens.Range.Start.Line {
+		// This can happen if a codeLens/resolve request comes in after a program change.
+		// While it's true that handlers should latch onto a specific snapshot
+		// while processing requests, we just set `Data.Uri` based on
+		// some older snapshot's contents. The content could have been modified,
+		// or the file itself could have been removed from the session entirely.
+		// Note this won't bail out on every change, but will prevent crashing
+		// based on non-existent files and line maps from shortened files.
+		return codeLens, lsproto.ErrContentModified
```
jakebailey: Needs to be `lsproto.ErrorCodeContentModified` now.

jakebailey (approved):

PR #2146, opened 2025-11-21
Port workspace symbols tests

gabritto:
Relevant test changes:
- Preference `excludeLibrarySymbols` is now true by default, for reasons which include the fact it's true by default in vscode, and also that you almost never want to see library symbols in the tests either.

Relevant symbol changes:
- Implemented `excludeLibrarySymbols`.
- Made the file exclusion logic similar to Strada.
- Added missing JS cases in `computeDeclarationMap()`: we were skipping `CommonJSExport` and `BinaryExpression` cases (i.e. JS expandos).
- Added missing cases to `getSymbolFromNodeKind`.

I also had to manually fix some tests to address changes in symbol `Kind` (e.g. `exports.y = function () ...` now has kind variable, had kind function in JS), and to test turning off `excludeLibrarySymbols`.

Another difference in Strada vs Corsa here is that in Corsa, we are including imported symbols even if they come from a library or node_modules and `excludeLibrarySymbols` is true. See the modified test https://github.com/microsoft/typescript-go/pull/2146/files#diff-8f459c0ecfef2ad46a1a166f4a39e0bb2e3e71152181f52597cf4a8befc02b2aR43. It would be possible to filter those symbols out in some cases in Corsa while still not using the checker, but for now, I'm leaving this new behavior in.


internal/fourslash/tests/manual/navto_excludeLib2_test.go
```diff
@@ -0,0 +1,52 @@
...
+	const content = `// @filename: /index.ts
+import { [|someName as weirdName|] } from "bar";
+// @filename: /tsconfig.json
+{}
+// @filename: /node_modules/bar/index.d.ts
+export const someName: number;
+// @filename: /node_modules/bar/package.json
+{}`
+	f := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	f.VerifyWorkspaceSymbol(t, []*fourslash.VerifyWorkspaceSymbolCase{
```
sheetalkamat, line 26: Especially for exact - can we baselineWorkspaceSymbols instead?

gabritto: I was planning on doing that in a future PR alongside the document symbol tests, so that new tests added to Corsa would be baseline tests, but the old ones would be roughly the same, would that be ok?

sheetalkamat (approved):

PR #2151, opened 2025-11-23
fix error missing bug for referencing files are not existing

TranNhatHan:
fix (#2081)


TranNhatHan:
Is there any issue that causes the test to fail?

jakebailey:
Superseded by #3206

PR #2168, opened 2025-11-29
Revise type-only alias logic to not depend on full type checks

### LINKED ISSUE #2114: Stamp out remaining ordering differences requiring multi-checker error
Apply this patch to disable the code that merges errors for multiple checkers in a single file, and unskips the tests that fail with it.

Then, run `hereby test --concurrentTestPrograms`

```diff
diff --git a/internal/compiler/program.go b/internal/compiler/program.go
index 2dbe96c47..d0f364076 100644
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -1034,7 +1034,7 @@ func (p *Program) getSemanticDiagnosticsForFileNotFilter(ctx context.Context, so
 		if sourceFile == nil || checker == fileChecker {
 			checkerDiags[idx] = checker.GetDiagnostics(ctx, sourceFile)
 		} else {
-			checkerDiags[idx] = checker.GetDiagnosticsWithoutCheck(sourceFile)
+			// checkerDiags[idx] = checker.GetDiagnosticsWithoutCheck(sourceFile)
 		}
 	})
 	if ctx.Err() != nil {
diff --git a/internal/testrunner/compiler_runner.go b/internal/testrunner/compiler_runner.go
index a79e1874a..8e2c60bb4 100644
--- a/internal/testrunner/compiler_runner.go
+++ b/internal/testrunner/compiler_runner.go
@@ -351,8 +351,8 @@ func newCompilerTest(
 }
 
 var concurrentSkippedErrorBaselines = map[string]string{
-	"typeOnlyMerge2.ts": "Type-only merging is not detected when files are checked on different checkers.",
-	"typeOnlyMerge3.ts": "Type-only merging is not detected when files are checked on different checkers.",
+	// "typeOnlyMerge2.ts": "Type-only merging is not detected when files are checked on different checkers.",
+	// "typeOnlyMerge3.ts": "Type-only merging is not detected when files are checked on different checkers.",
 }
 
 func (c *compilerTest) verifyDiagnostics(t *testing.T, suiteName string, isSubmodule bool) {
```

You'll get these four tests failing:

```
=== Failed
=== FAIL: internal/testrunner TestSubmodule/objectTypeHidingMembersOfExtendedObject.ts/error (0.00s)
    baseline.go:206: the baseline file /home/jabaile/work/TypeScript-go/testdata/baselines/reference/submodule/conformance/objectTypeHidingMembersOfExtendedObject.errors.txt has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
    baseline.go:201: new baseline created at /home/jabaile/work/TypeScript-go/testdata/baselines/local/submodule/conformance/objectTypeHidingMembersOfExtendedObject.errors.txt.diff.

=== FAIL: internal/testrunner TestSubmodule/duplicateNumericIndexers.ts/error (0.00s)
    baseline.go:206: the baseline file /home/jabaile/work/TypeScript-go/testdata/baselines/reference/submodule/conformance/duplicateNumericIndexers.errors.txt has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
    baseline.go:201: new baseline created at /home/jabaile/work/TypeScript-go/testdata/baselines/local/submodule/conformance/duplicateNumericIndexers.errors.txt.diff.

=== FAIL: internal/testrunner TestSubmodule/objectTypeHidingMembersOfExtendedObject.ts (0.62s)

=== FAIL: internal/testrunner TestSubmodule/duplicateNumericIndexers.ts (0.64s)

=== FAIL: internal/testrunner TestSubmodule/typeOnlyMerge3.ts/error (0.00s)
    baseline.go:206: the baseline file /home/jabaile/work/TypeScript-go/testdata/baselines/reference/submodule/conformance/typeOnlyMerge3.errors.txt has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
    baseline.go:201: new baseline created at /home/jabaile/work/TypeScript-go/testdata/baselines/local/submodule/conformance/typeOnlyMerge3.errors.txt.diff.

=== FAIL: internal/testrunner TestSubmodule/typeOnlyMerge3.ts (0.00s)

=== FAIL: internal/testrunner TestSubmodule/typeOnlyMerge2.ts/error (0.00s)
    baseline.go:206: the baseline file /home/jabaile/work/TypeScript-go/testdata/baselines/reference/submodule/conformance/typeOnlyMerge2.errors.txt has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
    baseline.go:201: new baseline created at /home/jabaile/work/TypeScript-go/testdata/baselines/local/submodule/conformance/typeOnlyMerge2.errors.txt.diff.

=== FAIL: internal/testrunner TestSubmodule/typeOnlyMerge2.ts (0.01s)

=== FAIL: internal/testrunner TestSubmodule (1.76s)
```

If we were to fix the checker so it is able to issue errors _everywhere_ and consistently like in #1904, we can probably eliminate the need to merge diagnostics between checkers for the same file, and then drop certain locking we have to do (and maybe eventually make Program not own checkers at all). This also would continue to harden us in the editor.

ahejlsberg:
This PR extensively revises our type-only import/export alias checking to not depend on full type checks. With the PR, it is no longer necessary to fully type-check all files in a program to obtain a stable set of diagnostics for a given source file--rather, it is sufficient just to check the particular file.

The new checking scheme works as follows: Resolution of each alias records any type-only node that affects _that_ alias (i.e. non-transitively). Checks for whether an alias should be considered type-only then walk up the alias chain (possibly resolving aliases along the way) until it finds a type-only affecting node, if one exists. Propagation of alias information is not attempted as it would require storing more information just to save a typically short and inexpensive walk.

This removes the last known difference between single-threaded and concurrent type checking.

Fixes #2114.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -14647,11 +14647,11 @@ func (c *Checker) markSymbolOfAliasDeclarationIfTypeOnly(aliasDeclaration *ast.N
 	// If the declaration itself is type-only, mark it and return. No need to check what it resolves to.
 	sourceSymbol := c.getSymbolOfDeclaration(aliasDeclaration)
 	links := c.aliasSymbolLinks.Get(sourceSymbol)
-	if ast.IsTypeOnlyImportOrExportDeclaration(aliasDeclaration) {
+	if links.typeOnlyDeclaration == nil && ast.IsTypeOnlyImportOrExportDeclaration(aliasDeclaration) {
 		links.typeOnlyDeclaration = aliasDeclaration
 		return true
 	}
-	if exportStarDeclaration != nil {
+	if links.typeOnlyDeclaration == nil && exportStarDeclaration != nil {
 		links.typeOnlyDeclaration = exportStarDeclaration
 		return true
 	}
@@ -15799,11 +15799,12 @@ func (c *Checker) resolveAlias(symbol *ast.Symbol) *ast.Symbol {
 	return links.aliasTarget
 }
 
-func (c *Checker) resolveIndirectionAlias(symbol *ast.Symbol, target *ast.Symbol) *ast.Symbol {
+func (c *Checker) resolveIndirectionAlias(source *ast.Symbol, target *ast.Symbol) *ast.Symbol {
 	result := c.resolveAlias(target)
-	targetLinks := c.aliasSymbolLinks.Get(target)
-	if targetLinks.typeOnlyDeclaration != nil {
-		c.aliasSymbolLinks.Get(symbol).typeOnlyDeclaration = targetLinks.typeOnlyDeclaration
+	if targetLinks := c.aliasSymbolLinks.Get(target); targetLinks.typeOnlyDeclaration != nil {
+		if sourceLinks := c.aliasSymbolLinks.Get(source); sourceLinks.typeOnlyDeclaration == nil {
+			sourceLinks.typeOnlyDeclaration = targetLinks.typeOnlyDeclaration
+		}
 	}
 	return result
 }


testdata/baselines/reference/submodule/conformance/typeOnlyMerge3.errors.txt
```diff
@@ -19,21 +16,12 @@ c.ts(4,1): error TS2349: This expression is not callable.
...
-!!! related TS1377 a.ts:2:15: 'A' was exported here.
     A.displayName;
-    ~
-!!! error TS1362: 'A' cannot be used as a value because it was exported using 'export type'.
-!!! related TS1377 a.ts:2:15: 'A' was exported here.
     A();
     ~
-!!! error TS1362: 'A' cannot be used as a value because it was exported using 'export type'.
-!!! related TS1377 a.ts:2:15: 'A' was exported here.
-    ~
```
ahejlsberg, line 36: These (previously inconsistently reported) errors are now no longer reported.

testdata/baselines/reference/submodule/conformance/importEquals3.errors.txt
```diff
@@ -21,16 +20,13 @@ c.ts(3,13): error TS1380: An import alias cannot reference a declaration that wa
...
+==== c.ts (1 errors) ====
     import * as b from './b';
     import A = b.a.A; // Error
                ~~~~~
 !!! error TS1380: An import alias cannot reference a declaration that was imported using 'import type'.
 !!! related TS1376 b.ts:1:18: 'a' was imported here.
     import AA = b.A; // Error
-                ~~~
-!!! error TS1380: An import alias cannot reference a declaration that was imported using 'import type'.
-!!! related TS1376 b.ts:1:18: 'a' was imported here.
```
ahejlsberg, line 33: An error was already reported on the declaration of `A` so this follow-on error isn't necessary.

testdata/baselines/reference/submodule/conformance/umd-augmentation-3.types
```diff
@@ -48,7 +48,7 @@ var t = p.x;
 
 === node_modules/math2d/index.d.ts ===
 export as namespace Math2d;
->Math2d : typeof import("node_modules/math2d/index.d.ts")
+>Math2d : typeof import("node_modules/math2d/index")
```
jakebailey, line 51: Odd change; wonder where this is coming from...

jakebailey (approved):
I tested this against 1, 2, 3, 5, 7, 13, 31, 61, 127, and 251 checkers, and it all seems to work.

I don't know enough to know if the Copilot comment is correct, however.

We can go and clean up the whole checker pool thing later now that we are sure we can drop multi-file merging.

internal/checker/checker.go
```diff
@@ -8157,7 +8138,7 @@ func (c *Checker) checkImportCallExpression(node *ast.Node) *Type {
 	// resolveExternalModuleName will return undefined if the moduleReferenceExpression is not a string literal
 	moduleSymbol := c.resolveExternalModuleName(node, specifier, false /*ignoreErrors*/)
 	if moduleSymbol != nil {
-		esModuleSymbol := c.resolveESModuleSymbol(moduleSymbol, specifier, true /*dontResolveAlias*/, false /*suppressInteropError*/)
+		esModuleSymbol := c.resolveExternalModuleSymbol(moduleSymbol, true /*dontResolveAlias*/)
```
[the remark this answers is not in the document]
ahejlsberg: This PR doesn't affect the actual module resolution process, so this is an unrelated suggestion. I don't think we want to change anything here since `resolveESModuleSymbol` has to do with ESM modules referencing CommonJS modules with an `export=` that merges with a something else, something that we don't necessarily want to support with `import()` calls. Either way, it would be a separate PR.

testdata/baselines/reference/submodule/conformance/umd-augmentation-3.types
ahejlsberg: Don't know, suspect it comes from something in emit.

PR #2171, opened 2025-12-01
Port call hierarchy, fourslash tests

jakebailey:
This is all very well tested; fourslash tests use baseline diffs, and the only diffs are ones related to the fact that tsserver differed from LSP, and so the code here is once again sort of a hybrid of the old code plus the post-processing steps VS Code does on top of tsserver, with the difference that I made script files not report themselves as `variable`, which VS Code happened to do because of its fallback condition (now, `file`).


internal/fourslash/fourslash.go
```diff
@@ -268,7 +269,23 @@ func getBaseFileNameFromTest(t *testing.T) string {
...
+		name = "callHierarchyFunctionAmbiguity.1"
+	case "callHierarchyFunctionAmbiguity2":
+		name = "callHierarchyFunctionAmbiguity.2"
+	case "callHierarchyFunctionAmbiguity3":
+		name = "callHierarchyFunctionAmbiguity.3"
+	case "callHierarchyFunctionAmbiguity4":
+		name = "callHierarchyFunctionAmbiguity.4"
+	case "callHierarchyFunctionAmbiguity5":
+		name = "callHierarchyFunctionAmbiguity.5"
+	}
```
jakebailey, line 286: really annoying but these tests had dots in their names ☹️

internal/ls/callhierarchy.go
```diff
@@ -0,0 +1,1032 @@
...
+			}
+		}
+		if ast.IsModuleBlock(parent.Parent.Parent.Parent) {
+			modParent := parent.Parent.Parent.Parent.Parent
+			if ast.IsModuleDeclaration(modParent) {
+				if name := modParent.Name(); name != nil && ast.IsIdentifier(name) {
+					return name.Text()
+				}
+			}
+		}
```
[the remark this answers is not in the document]
jakebailey: This is a pure port of this frankly hilarious code:

```ts
    if (isAssignedExpression(node)) {
        if (isPropertyDeclaration(node.parent) && isClassLike(node.parent.parent)) {
            return isClassExpression(node.parent.parent) ? getAssignedName(node.parent.parent)?.getText() : node.parent.parent.name?.getText();
        }
        if (isModuleBlock(node.parent.parent.parent.parent) && isIdentifier(node.parent.parent.parent.parent.parent.name)) {
            return node.parent.parent.parent.parent.parent.name.getText();
        }
        return;
    }
```

jakebailey:
Yeesh, either our tests are bad or I did something very wrong (well, I'm wrong either way); this does not seem to work in the editor past a single resolve.

jakebailey:
I'm apparently very silly and chose a test func that has no callers... 🤦 

Choosing a better func like `assert`...

image

testdata/baselines/reference/submodule/fourslash/callHierarchy/callHierarchyAccessor.callHierarchy.txt.diff
```diff
@@ -0,0 +1,10 @@
+--- old.callHierarchyAccessor.callHierarchy.txt
++++ new.callHierarchyAccessor.callHierarchy.txt
+@@= skipped -0, +0 lines =@@
+ // === Call Hierarchy ===
+ ╭ name: bar
+-├ kind: getter
```
gabritto: Couldn't we also fix this kind of diff (and same for `script` -> `file` below) in the baseline process so we get rid of these diff files?

jakebailey: Oh, great idea.

jakebailey: Done!

PR #2174, opened 2025-12-02
Server-to-client refresh requests for user preference updates

DanielRosenwasser:
Adds refresh support for

- inlay hints
- code lenses

Also

- Adds a level of nesting to `UserPreferences`, which is helpful for comparing entire objects and mirroring some of the hierarchy we have in our actual user preferences.
- Makes `hereby` run `go generate -v` so we can see progress.
- Fixes a few typos with inlay hints preference parsing.


DanielRosenwasser:
Looks like, at least for VS Code, a refresh request doesn't actually trigger a new inlay hints request.

https://github.com/microsoft/vscode/issues/280812

jakebailey (approved):

PR #2197, opened 2025-12-03
Make Program diagnostic API clearer

jakebailey:
Now that we do not need to ask multiple checkers for their errors anymore, we can restructure the funcs to make a bit more sense.

I've tried to simplify things. Notably:

- Were asking for binding way too often. We only need to do it before creating a checker (we need this during startup for globals/ambients), and then for `GetBindDiagnostics` which is is "special" and is used solely for timing `tsc`.
- We sometimes combine bind diags with semantic diags, so we could have the same diag objects when deduping. I've added fast paths to the diag comparison methods.
- `ForEachCheckerParallel` is now limited to non-`Project` `Program`s, used for stats/testing.

### DIFF
--- a/internal/compiler/program.go
+++ b/internal/compiler/program.go
@@ -368,8 +368,10 @@ func (p *Program) GetTypeChecker(ctx context.Context) (*checker.Checker, func())
 	return p.checkerPool.GetChecker(ctx)
 }
 
-func (p *Program) ForEachCheckerParallel(ctx context.Context, cb func(idx int, c *checker.Checker)) {
-	p.checkerPool.ForEachCheckerParallel(ctx, cb)
+func (p *Program) ForEachCheckerParallel(cb func(idx int, c *checker.Checker)) {
+	if pool, ok := p.checkerPool.(*checkerPool); ok {
+		pool.ForEachCheckerParallel(cb)
+	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/compiler/program.go
```diff
@@ -1139,28 +1131,12 @@ func (p *Program) getSuggestionDiagnosticsForFile(ctx context.Context, sourceFil
...
-		defer done()
-	}
+	fileChecker, done := p.checkerPool.GetCheckerForFile(ctx, sourceFile)
+	defer done()
 
+	// Getting a checker will force a bind, so this will be populated.
 	diags := slices.Clip(sourceFile.BindSuggestionDiagnostics)
-
-	checkerDiags := make([][]*ast.Diagnostic, p.checkerPool.Count())
-	p.checkerPool.ForEachCheckerParallel(ctx, func(idx int, checker *checker.Checker) {
```
andrewbranch, line 1152: Can this method be deleted now?

jakebailey: Almost; it's used for getting stats and for testing. I might be able to finagle those away.

andrewbranch (approved):

internal/compiler/program.go
jakebailey: I removed it as much as possible without giving up on things we need.

PR #2207, opened 2025-12-04
Port document symbol tests + fixes

gabritto:
In Strada, the document symbol tests (i.e. `navigationTree`) were not baseline tests. I initially ported those tests as they were in Strada, but some 40 tests failed, and mostly due to intentional differences, and so I just went through every failing test, fixed the most relevant differences, and re-ported the tests as baseline tests. Long term I think baseline tests are what we want for document symbol, and I think keeping diffs was going to be too noisy anyways.

I had to add support in document symbol for expandos and some other missing things like name truncation.

Noticeable differences:
- Strada used to mark some function and variable declarations as classes in JS whenever they had an expando property, and that meant we added a "constructor" document symbol child to those symbols. I'm not doing this in Corsa.
- In Corsa we don't look deeply into a node to determine its kind. So if e.g. you have a property assignment with a function, its kind is property, not method.
- In Corsa we don't include symbols for nodes with missing names, i.e. cases where you'd have a parse error anyways (e.g. `import *{} from 'foo'` won't include an unnamed namespace import).
- Corsa still doesn't add symbols for default export assignments. I think we added this in Strada in order to support code lenses, I am not sure if we still need to add those now.

Still missing:
- Support for expandos via `Object.defineProperty(...)` is missing because we still haven't added support for this construct yet in Corsa. See `navigationBarFunctionPrototype_test.go`. If we add support for it by reparsing it into something else, we'll just need to update the `newDocumentSymbol` to allow that reparsed node, but that should be a quick change.

If the baseline format is ok, I can add a similar baselining function for workspace symbols too, since it's likely we'll want to use that for new tests going forward.

### DIFF
--- a/internal/ast/diagnostic.go
+++ b/internal/ast/diagnostic.go
@@ -195,17 +195,11 @@ func getDiagnosticPath(d *Diagnostic) string {
 }
 
 func EqualDiagnostics(d1, d2 *Diagnostic) bool {
-	if d1 == d2 {
-		return true
-	}
 	return EqualDiagnosticsNoRelatedInfo(d1, d2) &&
 		slices.EqualFunc(d1.RelatedInformation(), d2.RelatedInformation(), EqualDiagnostics)
 }
 
 func EqualDiagnosticsNoRelatedInfo(d1, d2 *Diagnostic) bool {
-	if d1 == d2 {
-		return true
-	}
 	return getDiagnosticPath(d1) == getDiagnosticPath(d2) &&
 		d1.Loc() == d2.Loc() &&
 		d1.Code() == d2.Code() &&
@@ -214,9 +208,6 @@ func EqualDiagnosticsNoRelatedInfo(d1, d2 *Diagnostic) bool {
 }
 
 func equalMessageChain(c1, c2 *Diagnostic) bool {
-	if c1 == c2 {
-		return true
-	}
 	return c1.Code() == c2.Code() &&
 		slices.Equal(c1.MessageArgs(), c2.MessageArgs()) &&
 		slices.EqualFunc(c1.MessageChain(), c2.MessageChain(), equalMessageChain)
@@ -267,9 +258,6 @@ func compareRelatedInfo(r1, r2 []*Diagnostic) int {
 }
 
 func CompareDiagnostics(d1, d2 *Diagnostic) int {
-	if d1 == d2 {
-		return 0
-	}
 	c := strings.Compare(getDiagnosticPath(d1), getDiagnosticPath(d2))
 	if c != 0 {
 		return c
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -15800,7 +15800,7 @@ func (c *Checker) resolveAlias(symbol *ast.Symbol) *ast.Symbol {
 }
 
 func (c *Checker) resolveIndirectionAlias(source *ast.Symbol, target *ast.Symbol) *ast.Symbol {
-	result := c.resolveAlias(target)
+	result := c.getMergedSymbol(c.resolveAlias(target))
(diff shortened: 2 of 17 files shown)


internal/fourslash/_scripts/manualTests.txt
```diff
@@ -2,6 +2,14 @@ completionListInClosedFunction05
 completionsAtIncompleteObjectLiteralProperty
 completionsSelfDeclaring1
 completionsWithDeprecatedTag4
+navigationBarFunctionPrototype
```
gabritto, line 5: All of those tests had .js files and needed `// @allowJs: true` to be added. Strada didn't need this because document symbol did not need to obtain the file from a program, it just obtained the file from a syntax cache (?). In Corsa, we try to get the requested file from the program, so these tests crashed. One thing I've been wanting to do is to use the same inferred project default settings for fourslash, but this would break a few import code fix tests, so I didn't do it in this PR.

internal/fourslash/fourslash.go
```diff
@@ -3665,3 +3674,33 @@ func verifyIncludesSymbols(
 		assertDeepEqual(t, actualSym, sym, fmt.Sprintf("%s: Symbol '%s' at location '%v' mismatch", prefix, sym.Name, sym.Location))
 	}
 }
+
+var marshalSymbolKind = func(v lsproto.SymbolKind) ([]byte, error) {
```
jakebailey:
```suggestion
func marshalSymbolKind(v lsproto.SymbolKind) ([]byte, error) {
```

Seems like this would work too

testdata/baselines/reference/fourslash/documentSymbols/navbar01.baseline.md
```diff
@@ -0,0 +1,602 @@
...
+		"kind": "Interface",
+		"range": {
+			"start": {
+				"line": 1,
+				"character": 0
+			},
+			"end": {
+				"line": 7,
+				"character": 1
+			}
```
jakebailey: In some way, I wonder if we should also override printing of ranges in the baseline into something more compact.

gabritto: Do you have ideas as to what would be more compact? I guess I could either make it print the ranges single line, or print the text of the range instead, though that would mean stringifying the text to produce valid json.

jakebailey: I could imagine `[[1, 2], [3, 4]]` or `"1,2-3,4"`, but I don't really know.

I don't think we've been this clever with other tests, though things like sig help baselines previously used `postition` and not line/char. I don't mind leaving this as-is, but it is a lot of baselines, so I was speculating.

DanielRosenwasser: Crazy idea but what about printing inline like

```
/*Method getDist L2C4-L2C22*/«‹getDist›(): number»
```

DanielRosenwasser: Or you can just do something like

```
(Interface) IPoint - selectionRange L1C10-L1C16, range L1C0-L7C1
└── (Method) getDist - selectionRange L2C4-L2C11, range L2C4-L2C22
```

gabritto: > Crazy idea but what about printing inline like
> 
> ```
> /*Method getDist L2C4-L2C22*/«‹getDist›(): number»
> ```

I think I'm gonna do what we do for workspace symbols and other baselines where we place markers in the file to indicate positions.

> Or you can just do something like
> 
> ```
> (Interface) IPoint - selectionRange L1C10-L1C16, range L1C0-L7C1
> └── (Method) getDist - selectionRange L2C4-L2C11, range L2C4-L2C22
> ```

And then do something like this for indicating the hierarchy maybe?

PR #2235, opened 2025-12-05
Implementations, CodeLens, IncomingCalls across projects

### LINKED ISSUE #2177: Differences/issues with Code Lens references
### Extension Version

0.20251202.1

### VS Code Version

1.107.0-insider

### Operating system Version

Windows_NT x64 10.0.26220

### Steps to reproduce

1. Open https://github.com/gitkraken/vscode-gitlens/blob/cdf7cd3fc2f9246c10081b0b038b68328981d413/src/webviews/apps/home/home.ts#L61
2. And see that there are 70 references using TS, but 31 using TSGO

### Issue

Related to #2145
Seeing differences in the references between TS and TSGO

(Loving that this feature is back tho!)

sheetalkamat:
Fixes #2177

### DIFF
--- a/internal/ls/crossproject.go
+++ b/internal/ls/crossproject.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"iter"
+	"runtime/debug"
 	"sync"
 
 	"github.com/microsoft/typescript-go/internal/collections"
@@ -39,7 +40,6 @@ type CrossProjectOrchestrator interface {
 	GetLanguageServiceForProjectWithFile(ctx context.Context, project Project, uri lsproto.DocumentUri) *LanguageService
 	GetProjectsForFile(ctx context.Context, uri lsproto.DocumentUri) ([]Project, error)
 	GetProjectsLoadingProjectTree(ctx context.Context, requestedProjectTrees *collections.Set[tspath.Path]) iter.Seq[Project]
-	RecoverWith(r any) string
 }
 
 func handleCrossProject[Req lsproto.HasTextDocumentPosition, Resp any](
@@ -86,7 +86,8 @@ func handleCrossProject[Req lsproto.HasTextDocumentPosition, Resp any](
 			}
 			defer func() {
 				if r := recover(); r != nil {
-					panicOccured := orchestrator.RecoverWith(r)
+					stack := debug.Stack()
+					panicOccured := fmt.Sprintf("panic handling request: %v\n%s", r, string(stack))
 					panicMu.Lock()
 					panicsOccured = append(panicsOccured, panicOccured)
 					panicMu.Unlock()
@@ -206,7 +207,7 @@ func handleCrossProject[Req lsproto.HasTextDocumentPosition, Resp any](
 		wg.RunAndWait()
 		// No need to use mu here since we are not in parallel at this point
 		if panicsOccured != nil {
-			panic(fmt.Sprintf("Panics occured during cross-project handling: %v", panicsOccured))
+			panic(fmt.Sprintf("Panics occurred during cross-project handling: %v", panicsOccured))
 		}
 		if ctx.Err() != nil {
 			return resp, ctx.Err()
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -652,15 +652,6 @@ func registerMultiProjectReferenceRequestHandler[Req lsproto.HasTextDocumentPosi
 	}
 }
 
-func (s *Server) getLanguageServiceAndCrossProjectOrchestrator(ctx context.Context, uri lsproto.DocumentUri, req *lsproto.RequestMessage) (*ls.LanguageService, ls.CrossProjectOrchestrator, error) {
-	defaultProject, defaultLs, allProjects, err := s.session.GetLanguageServiceAndProjectsForFile(ctx, uri)
-	var orchestrator ls.CrossProjectOrchestrator
-	if err == nil {
-		orchestrator = &crossProjectOrchestrator{s, req, defaultProject, allProjects}
-	}
-	return defaultLs, orchestrator, err
-}
-
 type crossProjectOrchestrator struct {
 	server         *Server
 	req            *lsproto.RequestMessage
@@ -696,13 +687,13 @@ func (c *crossProjectOrchestrator) GetProjectsLoadingProjectTree(ctx context.Con
 	}
 }
 
-func (c *crossProjectOrchestrator) RecoverWith(r any) string {
-	stack := debug.Stack()
-	c.server.logger.Error("panic handling request", c.req.Method, r, string(stack))
-	if c.req.ID == nil {
-		c.server.logger.Error("unhandled panic in notification", c.req.Method, r)
+func (s *Server) getLanguageServiceAndCrossProjectOrchestrator(ctx context.Context, uri lsproto.DocumentUri, req *lsproto.RequestMessage) (*ls.LanguageService, ls.CrossProjectOrchestrator, error) {
+	defaultProject, defaultLs, allProjects, err := s.session.GetLanguageServiceAndProjectsForFile(ctx, uri)
+	var orchestrator ls.CrossProjectOrchestrator
+	if err == nil {
+		orchestrator = &crossProjectOrchestrator{s, req, defaultProject, allProjects}
 	}
-	return fmt.Sprintf("panic handling request %s: %v", c.req.Method, r)
+	return defaultLs, orchestrator, err
 }
 
 func (s *Server) recover(req *lsproto.RequestMessage) {


internal/lsp/server.go
```diff
@@ -1346,13 +1393,110 @@ func (s *Server) handleCodeLens(ctx context.Context, ls *ls.LanguageService, par
 }
 
 func (s *Server) handleCodeLensResolve(ctx context.Context, codeLens *lsproto.CodeLens, reqMsg *lsproto.RequestMessage) (*lsproto.CodeLens, error) {
-	ls, err := s.session.GetLanguageService(ctx, codeLens.Data.Uri)
-	if err != nil {
-		return nil, err
+	textDocument := lsproto.TextDocumentIdentifier{Uri: codeLens.Data.Uri}
```
jakebailey: I feel like we're in general putting too much code in this file. It seems like now basically all of the code lens code is here.

Does this multi-project stuff need to live here? Can it be done safely outside of this package?

sheetalkamat: it can go in project package i guess

sheetalkamat: ls cannot have this code because its better to just run the "references" or "implementations" request than having to wire that part again in multi project handler

sheetalkamat: Btw i havent done this for call hierarchy but we need to do something similar to implementaions for call hierarchy as well

sheetalkamat: Also remembered why i had not put it in session. We have `s.recover` that needs to be called in each go routine that gets spined for projects and that part of recovery only exists on lsp server and not on session?

internal/lsp/server.go
```diff
@@ -655,224 +655,242 @@ type response[Resp any] struct {
...
 		if err != nil {
 			return err
 		}
-		defer s.recover(req)
+		s.sendResult(req.ID, resp)
+		return nil
+	}
+}
+
+func multiProjectRequestHandling[Req lsproto.HasTextDocumentPosition, Resp any](
```
jakebailey: Nit, but I would call this `handleMultiProjectRequest`.

internal/lsp/server.go
```diff
@@ -655,224 +655,242 @@ type response[Resp any] struct {
...
-				results.Range(func(key tspath.Path, response *response[Resp]) bool {
-					if response.complete {
-						requestedProjectTrees[key] = struct{}{}
-					}
-					return true
-				})
+		wg = core.NewWorkGroup(false)
+		hasMoreWork := false
+		if defaultDefinition != nil {
+			requestedProjectTrees := make(map[tspath.Path]struct{})
```
jakebailey:
```suggestion
			var requestedProjectTrees collections.Set[tspath.Path]
```

internal/lsp/server.go
```diff
@@ -655,224 +655,242 @@ type response[Resp any] struct {
 func registerMultiProjectReferenceRequestHandler[Req lsproto.HasTextDocumentPosition, Resp any](
 	handlers handlerMap,
 	info lsproto.RequestInfo[Req, Resp],
-	fn func(*Server, context.Context, *ls.LanguageService, Req, *ast.Node, []*ls.SymbolAndEntries) (Resp, error),
-	combineResults func(iter.Seq[Resp]) Resp,
+	fn func(*Server, context.Context, *ls.LanguageService, Req, ls.SymbolAndEntriesData, ls.SymbolEntryTransformOptions) (Resp, error),
+	combineResults func(iter.Seq[*Resp]) Resp,
```
jakebailey: I think this needs to stay as `Resp`; we don't want responses to be `nil`, as the protocol (for this) doesn't want "missing" responses, only something or explicitly `null`, which is captured by the union type.

If you change this to `Resp`, and then fix `generate.mts` to generate `GetLocations` with a value receiver, then this all works out:

```diff
diff --git a/internal/lsp/lsproto/_generate/generate.mts b/internal/lsp/lsproto/_generate/generate.mts
index c1083847f..6e0412f79 100644
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -1584,7 +1584,7 @@ function generateCode() {
 
         // Generate GetLocations method
         if (hasLocations) {
-            writeLine(`func (o *${name}) GetLocations() *[]Location {`);
+            writeLine(`func (o ${name}) GetLocations() *[]Location {`);
             writeLine(`\treturn o.Locations`);
             writeLine(`}`);
             writeLine("");
diff --git a/internal/lsp/lsproto/lsp_generated.go b/internal/lsp/lsproto/lsp_generated.go
index 54c7a31f7..90d97b981 100644
--- a/internal/lsp/lsproto/lsp_generated.go
+++ b/internal/lsp/lsproto/lsp_generated.go
@@ -27101,7 +27101,7 @@ func (o *LocationOrLocationsOrDefinitionLinksOrNull) UnmarshalJSONFrom(dec *json
 	return fmt.Errorf("invalid LocationOrLocationsOrDefinitionLinksOrNull: %s", data)
 }
 
-func (o *LocationOrLocationsOrDefinitionLinksOrNull) GetLocations() *[]Location {
+func (o LocationOrLocationsOrDefinitionLinksOrNull) GetLocations() *[]Location {
 	return o.Locations
 }
 
@@ -27195,7 +27195,7 @@ func (o *LocationOrLocationsOrDeclarationLinksOrNull) UnmarshalJSONFrom(dec *jso
 	return fmt.Errorf("invalid LocationOrLocationsOrDeclarationLinksOrNull: %s", data)
 }
 
-func (o *LocationOrLocationsOrDeclarationLinksOrNull) GetLocations() *[]Location {
+func (o LocationOrLocationsOrDeclarationLinksOrNull) GetLocations() *[]Location {
 	return o.Locations
 }
 
@@ -27951,7 +27951,7 @@ func (o *LocationsOrNull) UnmarshalJSONFrom(dec *jsontext.Decoder) error {
 	return fmt.Errorf("invalid LocationsOrNull: %s", data)
 }
 
-func (o *LocationsOrNull) GetLocations() *[]Location {
+func (o LocationsOrNull) GetLocations() *[]Location {
 	return o.Locations
 }
 
diff --git a/internal/lsp/server.go b/internal/lsp/server.go
index 47d996249..ac9d7af20 100644
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -656,7 +656,7 @@ func registerMultiProjectReferenceRequestHandler[Req lsproto.HasTextDocumentPosi
 	handlers handlerMap,
 	info lsproto.RequestInfo[Req, Resp],
 	fn func(*Server, context.Context, *ls.LanguageService, Req, ls.SymbolAndEntriesData, ls.SymbolEntryTransformOptions) (Resp, error),
-	combineResults func(iter.Seq[*Resp]) Resp,
+	combineResults func(iter.Seq[Resp]) Resp,
 ) {
 	handlers[info.Method] = func(s *Server, ctx context.Context, req *lsproto.RequestMessage) error {
 		resp, err := handleMultiProjectRequest(s, ctx, info, fn, combineResults, req, ls.SymbolEntryTransformOptions{})
@@ -673,7 +673,7 @@ func handleMultiProjectRequest[Req lsproto.HasTextDocumentPosition, Resp any](
 	ctx context.Context,
 	info lsproto.RequestInfo[Req, Resp],
 	fn func(*Server, context.Context, *ls.LanguageService, Req, ls.SymbolAndEntriesData, ls.SymbolEntryTransformOptions) (Resp, error),
-	combineResults func(iter.Seq[*Resp]) Resp,
+	combineResults func(iter.Seq[Resp]) Resp,
 	req *lsproto.RequestMessage,
 	options ls.SymbolEntryTransformOptions,
 ) (Resp, error) {
@@ -782,11 +782,11 @@ func handleMultiProjectRequest[Req lsproto.HasTextDocumentPosition, Resp any](
 		}
 	}
 
-	getResultsIterator := func() iter.Seq[*Resp] {
-		return func(yield func(*Resp) bool) {
+	getResultsIterator := func() iter.Seq[Resp] {
+		return func(yield func(Resp) bool) {
 			var seenProjects collections.SyncSet[tspath.Path]
 			if response, loaded := results.Load(defaultProject.Id()); loaded && response.complete {
-				if !yield(&response.result) {
+				if !yield(response.result) {
 					return
 				}
 			}
@@ -794,7 +794,7 @@ func handleMultiProjectRequest[Req lsproto.HasTextDocumentPosition, Resp any](
 			for _, project := range allProjects {
 				if seenProjects.AddIfAbsent(project.Id()) {
 					if response, loaded := results.Load(project.Id()); loaded && response.complete {
-						if !yield(&response.result) {
+						if !yield(response.result) {
 							return
 						}
 					}
@@ -803,14 +803,14 @@ func handleMultiProjectRequest[Req lsproto.HasTextDocumentPosition, Resp any](
 			// Prefer the searches from locations for default definition
 			results.Range(func(key tspath.Path, response *response[Resp]) bool {
 				if !response.forOriginalLocation && seenProjects.AddIfAbsent(key) && response.complete {
-					return yield(&response.result)
+					return yield(response.result)
 				}
 				return true
 			})
 			// Then the searches from original locations
 			results.Range(func(key tspath.Path, response *response[Resp]) bool {
 				if response.forOriginalLocation && seenProjects.AddIfAbsent(key) && response.complete {
-					return yield(&response.result)
+					return yield(response.result)
 				}
 				return true
 			})
@@ -886,7 +886,7 @@ func handleMultiProjectRequest[Req lsproto.HasTextDocumentPosition, Resp any](
 	} else {
 		// Single result, return that directly
 		for value := range getResultsIterator() {
-			resp = *value
+			resp = value
 			break
 		}
 	}
@@ -1236,7 +1236,7 @@ func combineResponseLocations[T lsproto.HasLocations](results iter.Seq[T]) *[]ls
 	return &combined
 }
 
-func combineReferences(results iter.Seq[*lsproto.ReferencesResponse]) lsproto.ReferencesResponse {
+func combineReferences(results iter.Seq[lsproto.ReferencesResponse]) lsproto.ReferencesResponse {
 	return lsproto.LocationsOrNull{Locations: combineResponseLocations(results)}
 }
 
@@ -1245,7 +1245,7 @@ func (s *Server) handleImplementations(ctx context.Context, ls *ls.LanguageServi
 	return ls.ProvideImplementationsFromSymbolAndEntries(ctx, params, data, options)
 }
 
-func combineImplementations(results iter.Seq[*lsproto.ImplementationResponse]) lsproto.ImplementationResponse {
+func combineImplementations(results iter.Seq[lsproto.ImplementationResponse]) lsproto.ImplementationResponse {
 	var combined []*lsproto.LocationLink
 	var seenLocations collections.Set[lsproto.Location]
 	for resp := range results {
@@ -1329,7 +1329,7 @@ func (s *Server) handleRename(ctx context.Context, ls *ls.LanguageService, param
 	return ls.ProvideRenameFromSymbolAndEntries(ctx, params, data, options)
 }
 
-func combineRenameResponse(results iter.Seq[*lsproto.RenameResponse]) lsproto.RenameResponse {
+func combineRenameResponse(results iter.Seq[lsproto.RenameResponse]) lsproto.RenameResponse {
 	combined := make(map[lsproto.DocumentUri][]*lsproto.TextEdit)
 	seenChanges := make(map[lsproto.DocumentUri]*collections.Set[lsproto.Range])
 	// !!! this is not used any more so we will skip this part of deduplication and combining
```

internal/lsp/server.go
DanielRosenwasser: Do you think it makes sense to have an abstraction called `ls.CrossProjectBroadcaster` (or `ls.CrossProjectAggregator` or whatever) which has a `RequestAcrossProjects` method? 

Basically the idea is that on the server side, it would repackage the server/session in a custom type, and `RequestAcrossProjects` could just call `handleMultiProjectRequest`.

sheetalkamat: I will look into this as a separate PR request so we can review the refactoring separately.

sheetalkamat: ok i think,  I have made progress and will push the change to this PR itself once ready

internal/ls/crossproject.go
```diff
@@ -0,0 +1,347 @@
...
+	forOriginalLocation bool
+}
+
+type response[Resp any] struct {
+	complete            bool
+	result              Resp
+	forOriginalLocation bool
+}
+
+type CrossProjectOrchestrator interface {
```
sheetalkamat, line 37: naming suggestions welcome

sheetalkamat:
Failure is because of not correctly calling recover to assign request for panic. will take a look .

internal/ls/crossproject.go
```diff
@@ -0,0 +1,366 @@
...
+		if _, loaded := results.LoadOrStore(item.project.Id(), &response); loaded {
+			return
+		}
+		wg.Queue(func() {
+			if ctx.Err() != nil {
+				return
+			}
+			defer func() {
+				if r := recover(); r != nil {
+					orchestrator.RecoverWith(r)
```
jakebailey: If multiple items panic, we'll be sending the error response multiple times, so we will have to do this slightly differently to avoid that.

internal/ls/crossproject.go
```diff
@@ -198,10 +204,13 @@ func handleCrossProject[Req lsproto.HasTextDocumentPosition, Resp any](
 	for {
 		// Process existing known projects first
 		wg.RunAndWait()
+		// No need to use mu here since we are not in parallel at this point
+		if panicsOccured != nil {
+			panic(fmt.Sprintf("Panics occured during cross-project handling: %v", panicsOccured))
```
jakebailey: If these are handled here anyway, you might as well ditch `RecoverWith` and just inline the code; it will be equivalent.

sheetalkamat: ok let me do that

PR #2251, opened 2025-12-06
  Fix: Add mutex protection for projects map to prevent race conditions

Catsayer-Chan:
## Summary
  This PR adds proper mutex protection to the `projects` map in the API struct to prevent potential race conditions.

  ## Problem
  The `projects` map in `internal/api/api.go` was the only map in the API struct without mutex protection, while `files`, `symbols`, and `types` maps
  were already protected. This inconsistency could lead to data races if the API handles concurrent requests.

  ## Solution
  - Added `sync.RWMutex` to protect the `projects` map
  - Applied read locks (`RLock`) for all read operations (4 locations)
  - Applied write locks (`Lock`) for write/delete operations (2 locations)
  - Used `RWMutex` instead of `Mutex` for better read performance

  ## Testing
  - ✅ Code compiles without errors
  - ✅ Passes race detector: `go test -race ./internal/api/...`
  - ✅ No functional changes, backward compatible

  ## Modified Files
  - `internal/api/api.go`

  ## Impact
  - Improves thread safety and code consistency
  - Prepares codebase for concurrent request handling
  - Negligible performance impact in single-threaded scenarios

  Author: Catsayer

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -32,13 +32,14 @@ type API struct {
 	logger  logging.Logger
 	session *project.Session
 
-	projects  map[Handle[project.Project]]tspath.Path
-	filesMu   sync.Mutex
-	files     handleMap[ast.SourceFile]
-	symbolsMu sync.Mutex
-	symbols   handleMap[ast.Symbol]
-	typesMu   sync.Mutex
-	types     handleMap[checker.Type]
+	projectsMu sync.RWMutex
+	projects   map[Handle[project.Project]]tspath.Path
+	filesMu    sync.Mutex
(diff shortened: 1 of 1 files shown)


Catsayer-Chan:

jakebailey:
Where are you using this API such that it matters for you? It's not guaranteed to work and I'm not sure we are expecting people to use it truly concurrently on the client side yet?

jakebailey:
The API code has changed a lot since this PR was sent. Is this still needed?

jakebailey:
This is really behind and I haven't gotten a reply. I'm going to close this for housekeeping reasons.

PR #2272, opened 2025-12-08
Revamp user preferences serialization/deserialization

jakebailey:
In working on improving the fourslash testing, I wanted to make user preferences serializable, such that we can directly go from JSON to prefs and back, and therefore also test out the "actual" JSON encoding.

Making that work sort of snowballed into a total refactor, which eliminates the error-prone switch case for decoding, replacing it with struct tags and reflect.

Prefs now look like this:

```go
type CodeLensUserPreferences struct {
	ReferencesCodeLensEnabled                     bool `raw:"referencesCodeLensEnabled" config:"referencesCodeLens.enabled"`
	ImplementationsCodeLensEnabled                bool `raw:"implementationsCodeLensEnabled" config:"implementationsCodeLens.enabled"`
	ReferencesCodeLensShowOnAllFunctions          bool `raw:"referencesCodeLensShowOnAllFunctions" config:"referencesCodeLens.showOnAllFunctions"`
	ImplementationsCodeLensShowOnInterfaceMethods bool `raw:"implementationsCodeLensShowOnInterfaceMethods" config:"implementationsCodeLens.showOnInterfaceMethods"`
	ImplementationsCodeLensShowOnAllClassMethods  bool `raw:"implementationsCodeLensShowOnAllClassMethods" config:"implementationsCodeLens.showOnAllClassMethods"`
}
```

A struct tag gives the name for the JSON propery (used by Strada), and also the name over LSP. https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/package.json.

All prefs can still be set by the `unstable` object (which happens first, based on the raw name).

Structs can be nested, e.g.:

```go
type UserPreferences struct {
	// ...
	ModuleSpecifier ModuleSpecifierUserPreferences
	// ...
}

type ModuleSpecifierUserPreferences struct {
	ImportModuleSpecifierPreference modulespecifiers.ImportModuleSpecifierPreference `raw:"importModuleSpecifierPreference" config:"preferences.importModuleSpecifier"` // !!!
	// Determines whether we import `foo/index.ts` as "foo", "foo/index", or "foo/index.js"
	ImportModuleSpecifierEnding       modulespecifiers.ImportModuleSpecifierEndingPreference `raw:"importModuleSpecifierEnding" config:"preferences.importModuleSpecifierEnding"`             // !!!
	AutoImportSpecifierExcludeRegexes []string                                               `raw:"autoImportSpecifierExcludeRegexes" config:"preferences.autoImportSpecifierExcludeRegexes"` // !!!
}
```

And, for prefs that are actually defined elsewhere (like the above), conversion works so long as the structure is the same, making this less error prone:

```go
func (p *UserPreferences) ModuleSpecifierPreferences() modulespecifiers.UserPreferences {
	return modulespecifiers.UserPreferences(p.ModuleSpecifier)
}
```

`UserPreferences` are also now considered to be immutable. They should never be mutated for any reason. `Copy()` can be used if you need to make a deep copy.

User prefs are also now value types; pointers to them are not passed around. (Slightly untrue given they contain slices, but, oh well).

### DIFF
--- a/internal/ls/lsutil/userpreferences.go
+++ b/internal/ls/lsutil/userpreferences.go
@@ -769,18 +769,22 @@ func (p UserPreferences) IsModuleSpecifierExcluded(moduleSpecifier string) bool 
 func ParseUserPreferences(items map[string]any) UserPreferences {
 	prefs := NewDefaultUserPreferences()
 	// Apply editor settings first (tabSize, indentSize, etc.) as raw-name defaults,
-	// then overlay js/ts settings which take precedence.
+	// then overlay language-specific settings with increasing precedence:
+	// editor < javascript < typescript < js/ts
 	if editorItem, ok := items["editor"]; ok && editorItem != nil {
 		if editorSettings, ok := editorItem.(map[string]any); ok {
 			prefs = prefs.withConfig(map[string]any{"unstable": editorSettings})
 		}
 	}
-	if jsTsItem, ok := items["js/ts"]; ok && jsTsItem != nil {
-		switch jsTsSettings := jsTsItem.(type) {
-		case map[string]any:
-			prefs = prefs.withConfig(jsTsSettings)
-		case UserPreferences:
-			prefs = prefs.WithOverrides(jsTsSettings)
+	// Apply javascript, then typescript, then js/ts (highest precedence).
+	for _, section := range []string{"javascript", "typescript", "js/ts"} {
+		if item, ok := items[section]; ok && item != nil {
+			switch settings := item.(type) {
+			case map[string]any:
+				prefs = prefs.withConfig(settings)
+			case UserPreferences:
+				prefs = prefs.WithOverrides(settings)
+			}
 		}
 	}
 	return prefs
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -224,11 +224,27 @@ func NewFourslash(t *testing.T, capabilities *lsproto.ClientCapabilities, conten
 func (f *FourslashTest) handleServerRequest(_ context.Context, req *lsproto.RequestMessage) *lsproto.ResponseMessage {
 	switch req.Method {
 	case lsproto.MethodWorkspaceConfiguration:
-		// Return current user preferences
+		// Return current user preferences for each requested section.
+		// The server requests multiple sections (js/ts, typescript, javascript, editor);
+		// we return user preferences for "js/ts" and nil for others.
+		params, ok := req.Params.(*lsproto.ConfigurationParams)
+		if !ok || params == nil || params.Items == nil {
+			return &lsproto.ResponseMessage{
+				ID:      req.ID,
+				JSONRPC: req.JSONRPC,
+				Result:  []any{f.userPreferences},
+			}
+		}
+		results := make([]any, len(params.Items))
+		for i, item := range params.Items {
+			if item.Section != nil && *item.Section == "js/ts" {
+				results[i] = f.userPreferences
+			}
+		}
 		return &lsproto.ResponseMessage{
 			ID:      req.ID,
 			JSONRPC: req.JSONRPC,
-			Result:  []any{f.userPreferences},
+			Result:  results,
 		}
 
 	case lsproto.MethodClientRegisterCapability:
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -338,6 +338,12 @@ func (s *Server) RequestConfiguration(ctx context.Context) (lsutil.UserPreferenc
 				Section: new("js/ts"),
 			},
 			{
+				Section: new("typescript"),
+			},
+			{
+				Section: new("javascript"),
+			},
+			{
 				Section: new("editor"),
 			},
 		},
@@ -351,12 +357,18 @@ func (s *Server) RequestConfiguration(ctx context.Context) (lsutil.UserPreferenc
 		case 0:
 			configMap["js/ts"] = config
 		case 1:
+			configMap["typescript"] = config
+		case 2:
+			configMap["javascript"] = config
+		case 3:
 			configMap["editor"] = config
 		}
 	}
 	s.logger.Logf(
-		"received options from workspace/configuration request:\njs/ts: %+v\n\neditor: %+v\n",
+		"received options from workspace/configuration request:\njs/ts: %+v\n\ntypescript: %+v\n\njavascript: %+v\n\neditor: %+v\n",
 		configMap["js/ts"],
+		configMap["typescript"],
+		configMap["javascript"],
 		configMap["editor"],
 	)
 	return lsutil.ParseUserPreferences(configMap), nil
@@ -1139,7 +1151,7 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 				RegisterOptions: &lsproto.RegisterOptions{
 					WorkspaceDidChangeConfiguration: &lsproto.DidChangeConfigurationRegistrationOptions{
 						Section: &lsproto.StringOrStrings{
-							Strings: &[]string{"js/ts", "editor"},
+							Strings: &[]string{"js/ts", "typescript", "javascript", "editor"},
 						},
 					},
 				},


internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,511 @@ const (
...
+		case OrganizeImportsTypeOrderLast:
+			return "last"
+		case OrganizeImportsTypeOrderInline:
+			return "inline"
+		case OrganizeImportsTypeOrderFirst:
+			return "first"
+		default:
+			return "auto"
+		}
+	},
```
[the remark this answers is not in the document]
jakebailey: No, these are strings, so don't need special handling.

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,511 @@ const (
...
-				p.set("organizeimportsnumericcollation", numeric)
-			}
-			if accent, ok := prefsMap["accentCollation"]; ok {
-				p.set("organizeimportsaccentcollation", accent)
-			}
-			if caseFirst, ok := prefsMap["caseFirst"]; ok && !p.OrganizeImportsIgnoreCase.IsTrue() {
-				p.set("organizeimportscasefirst", caseFirst)
+}
+
+func (p *UserPreferences) MarshalJSONTo(enc *jsontext.Encoder) error {
```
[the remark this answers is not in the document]
jakebailey: If it's nil, things will fall over. That's just how unmarshalling works.

internal/ls/lsutil/userpreferences_test.go
```diff
@@ -0,0 +1,87 @@
...
+		assert.NilError(t, err2)
+		assert.DeepEqual(t, original, parsed)
+	})
+
+	t.Run("parseWorker", func(t *testing.T) {
+		t.Parallel()
+		var config map[string]any
+		err2 := json.Unmarshal(jsonBytes, &config)
+		assert.NilError(t, err2)
+		parsed := &UserPreferences{}
```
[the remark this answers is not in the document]
jakebailey: This is intentional for testing.

internal/ls/lsutil/userpreferences.go
```diff
@@ -1,33 +1,43 @@
...
+	ProvideRefactorNotApplicableReason: true,
+	IncludeCompletionsWithSnippetText:  core.TSTrue,
+	DisplayPartsForJSDoc:               true,
+	DisableLineTextInReferences:        true,
+	ReportStyleChecksAsWarnings:        true,
 
-		ExcludeLibrarySymbolsInNavTo: true,
-	}
+	ExcludeLibrarySymbolsInNavTo: true,
 }
```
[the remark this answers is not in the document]
jakebailey: So, yes, but I think this is a bit much.

internal/ls/lsutil/userpreferences.go
```diff
@@ -38,87 +48,85 @@ type UserPreferences struct {
...
-	ImportModuleSpecifierPreference modulespecifiers.ImportModuleSpecifierPreference // !!!
-	// Determines whether we import `foo/index.ts` as "foo", "foo/index", or "foo/index.js"
-	ImportModuleSpecifierEnding       modulespecifiers.ImportModuleSpecifierEndingPreference // !!!
-	IncludePackageJsonAutoImports     IncludePackageJsonAutoImports                          // !!!
-	AutoImportSpecifierExcludeRegexes []string                                               // !!!
-	AutoImportFileExcludePatterns     []string                                               // !!!
-	PreferTypeOnlyAutoImports         bool                                                   // !!!
+	ModuleSpecifier ModuleSpecifierUserPreferences
+
+	IncludePackageJsonAutoImports IncludePackageJsonAutoImports `pref:"preferences.includePackageJsonAutoImports"` // !!!
```
DanielRosenwasser: Do you think it would be better to just omit these until the code is used?

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,505 @@ const (
...
+var fieldInfoCache = sync.OnceValue(func() []fieldInfo {
+	var infos []fieldInfo
+	collectFieldInfos(reflect.TypeFor[UserPreferences](), nil, &infos)
+	return infos
+})
+
+// lowerNameIndex maps lowercase field names to fieldInfo index for O(1) lookup
+var lowerNameIndex = sync.OnceValue(func() map[string]int {
+	infos := fieldInfoCache()
+	index := make(map[string]int, len(infos)*2)
```
DanielRosenwasser: `*2` is a little overkill, it looks like there's only one alias right now.

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,505 @@ const (
...
+})
 
-func (p *UserPreferences) ModuleSpecifierPreferences() modulespecifiers.UserPreferences {
-	return modulespecifiers.UserPreferences{
-		ImportModuleSpecifierPreference:   p.ImportModuleSpecifierPreference,
-		ImportModuleSpecifierEnding:       p.ImportModuleSpecifierEnding,
-		AutoImportSpecifierExcludeRegexes: p.AutoImportSpecifierExcludeRegexes,
-	}
-}
+func collectFieldInfos(t reflect.Type, indexPath []int, infos *[]fieldInfo) {
```
DanielRosenwasser: I might have just made `collectFieldInfos` just return `[]fieldInfo` instead but this is probably fine.

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,505 @@ const (
...
-		case "objectLiteralMethodSnippets":
-			if v, ok := value.(map[string]any); ok {
-				p.set("includeCompletionsWithObjectLiteralMethodSnippets", parseEnabledBool(v))
-			}
-		case "classMemberSnippets":
-			if v, ok := value.(map[string]any); ok {
-				p.set("includeCompletionsWithClassMemberSnippets", parseEnabledBool(v))
+		case "preferences", "suggest", "inlayHints", "referencesCodeLens",
+			"implementationsCodeLens", "workspaceSymbols", "format", "tsserver", "tsc", "experimental":
+			continue
```
DanielRosenwasser: What's an example of a path that doesn't need to be skipped?

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +249,511 @@ const (
...
+		return serializer(field.Interface())
+	}
+
+	switch field.Kind() {
+	case reflect.Bool:
+		return field.Bool()
+	case reflect.Int:
+		return int(field.Int())
+	case reflect.String:
+		return field.String()
```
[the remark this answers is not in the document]
DanielRosenwasser: Do we need some sort of `omitzero`?

internal/ls/lsutil/userpreferences_test.go
```diff
@@ -0,0 +1,87 @@
...
+			if field.Type().Elem().Kind() == reflect.String {
+				field.Set(reflect.ValueOf([]string{"test"}))
+			}
+		case reflect.Struct:
+			fillNonZeroValues(field, fieldPath)
+		}
+	}
+}
+
+func getValidStringValue(t reflect.Type, path string) string {
```
DanielRosenwasser: `path` doesn't seem to be used - maybe you needed it before? If you don't, you can get rid of `fieldPath` too above.

internal/ls/lsutil/userpreferences.go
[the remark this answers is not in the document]
jakebailey: Eh, not sure. The roundtripping is not perfect because these types were switched away from things like `*string` for convenience.

internal/ls/lsutil/userpreferences.go
jakebailey: I don't even think we need the alias functionality at all. It's only used for one thing because VS Code called the option something different than TS, but I would be utterly shocked if anyone had been using `providePrefixAndSuffixTextForRename`, especially since `unstable` options get overridden by other VS Code defaults anyhow

internal/ls/lsutil/userpreferences.go
jakebailey: IIRC this is upcoming, so it would just be churn to remove I think...

internal/ls/lsutil/userpreferences.go
jakebailey: Yeah, I think this is actually dead code.

jakebailey: Ah, so this was here to roundtrip unstable settings, but it isn't doing that right.

I'm going to rethink unstable settings. Probably just make them explicit.

jakebailey: Okay, so the sort of problem is that we used to allow user prefs from like, anywhere, so you could have a object with say, `{ "quotePreference": ... }` and that would be read in, which is a valid `UserPreferences` object. But, now over LSP, we are expecting things over `typescript`. So it's maybe a question of what non-vscode people do.

jakebailey: And, VS Code calls it `quoteStyle`, while preferences called it `quotePreference`, so that's uh, great

jakebailey: I split things apart so that we can both have the raw JS name we used to have, but also the LSP-based preference name.

internal/ls/lsutil/userpreferences.go
```diff
@@ -308,414 +247,503 @@ const (
...
-func (p *UserPreferences) Copy() *UserPreferences {
-	if p == nil {
-		return nil
-	}
-	prefCopy := *p
-	prefCopy.AutoImportSpecifierExcludeRegexes = slices.Clone(p.AutoImportSpecifierExcludeRegexes)
-	prefCopy.AutoImportFileExcludePatterns = slices.Clone(p.AutoImportFileExcludePatterns)
-	return &prefCopy
+type fieldInfo struct {
+	rawName      string // raw name for unstable section lookup (e.g., "quotePreference")
```
DanielRosenwasser, line 450: So is the idea that to get to this, a user can only write the explicit path or `typescript.unstable.quotePreference`?

jakebailey: Yeah. Either, you are using `unstable` and you were working by spread directly into `UserPreferences` in Strada, or you were using the "real" name VS Code had, which takes precedence via the spread.

Arguably, this means we should entirely delete any unstable thing with a real name, though, because VS Code actually entirely overwrote these (by nature of `config.get` having defaults): https://github.com/microsoft/vscode/blob/f9e3a710d84879d41431e7c17d8a8fe8612dfb26/extensions/typescript-language-features/src/languageFeatures/fileConfigurationManager.ts#L186

So maybe I went overkill and we shouldn't allow both names.

jakebailey: Basically, it's hard to balance "LSP VS Code style settings" with "the old tsserver UserPreferences", defaults, spread, and then maybe even other editors...

So, I don't know what is exactly right.

DanielRosenwasser: I think in this case, needing to support all these `unstable` settings years later kind of defeats the purpose - but we'll probably need to support new `unstable` properties again anyway, right? So I'm not against keeping support around for it as a concept, and it is virtually no work for us to keep the current ones going if we want 6.0/7.0 to be smoother.

DanielRosenwasser: (So your call but I think this is fine)

jakebailey: Probably I'll just keep this.

jakebailey:
I think I'm going to wait for the other two user pref related PRs to finish and just fix this PR instead of forcing that on everyone.

internal/ls/lsutil/userpreferences.go
```diff
@@ -315,448 +250,506 @@ const (
...
-	return defaultV
+// --- Helper methods ---
+
+// WithOverrides returns a copy of p with non-zero fields from overrides applied on top.
+// This is safe because all preference fields use types where zero = "not set":
+// Tristate (TSUnknown=0), int (0), string (""), slice (nil).
+func (p UserPreferences) WithOverrides(overrides UserPreferences) UserPreferences {
+	mergeNonZeroFields(reflect.ValueOf(&p).Elem(), reflect.ValueOf(&overrides).Elem())
+	return p
 }
```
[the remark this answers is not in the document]
jakebailey: Yes, though I don't think it matters; I don't think anyone should be mutating the slices themselves.

internal/ls/lsutil/userpreferences.go
```diff
@@ -315,448 +250,506 @@ const (
...
+	if jsTsItem, ok := items["js/ts"]; ok && jsTsItem != nil {
+		switch jsTsSettings := jsTsItem.(type) {
+		case map[string]any:
+			prefs = prefs.withConfig(jsTsSettings)
+		case UserPreferences:
+			prefs = prefs.WithOverrides(jsTsSettings)
+		}
 	}
-	return false
+	return prefs
```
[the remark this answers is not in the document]
jakebailey: I did this on purpose, but could bring it back, I suppose.

internal/ls/lsutil/userpreferences.go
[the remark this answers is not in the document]
iisaduan: Yes, once parsed, there shouldn't be any mutating. The one potential case is if you are parsing into an existing (non default) userpreferences object-- in that case you'd need to parse into a copy

jakebailey:
Any opinions on whether or not the `typescript`/`javascript` fallback is preserved? Not hard to stick back in.

jakebailey:
I'm going to restore it, it wasn't really the main point of this change but something I did in the middle of the refactor.

PR #2283, opened 2025-12-08
Fix perfromance when there's lots of error

GabrielCastro:
DiagnosticsCollection#Add() uses InsertSort, this causes a ton of memcopys on every error. This is O(n) for every error.

Instead I moved the sort to lazily happen just before each GetDiagnostics call this is O(nlogn), but gets happen way less often than Inserts.

This happens on codebases with lots of usages of react-hook-form due to https://github.com/microsoft/typescript-go/issues/988 . While this isn't a fix for that it allows builds to fail instead of hanging for hours

### DIFF
--- a/internal/ast/diagnostic.go
+++ b/internal/ast/diagnostic.go
@@ -5,6 +5,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 )
@@ -123,7 +124,7 @@ func NewCompilerDiagnostic(message *diagnostics.Message, args ...any) *Diagnosti
 
 type DiagnosticsCollection struct {
 	fileDiagnostics          map[string][]*Diagnostic
-	fileDiagnosticsSorted    map[string]bool
+	fileDiagnosticsSorted    collections.Set[string]
 	nonFileDiagnostics       []*Diagnostic
 	nonFileDiagnosticsSorted bool
 }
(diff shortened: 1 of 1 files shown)


jakebailey:
This looks fine, but you should run `hereby format`.

internal/ast/diagnostic.go
```diff
@@ -132,18 +134,23 @@ func (c *DiagnosticsCollection) Add(diagnostic *Diagnostic) {
 		if c.fileDiagnostics == nil {
 			c.fileDiagnostics = make(map[string][]*Diagnostic)
 		}
-		c.fileDiagnostics[fileName] = core.InsertSorted(c.fileDiagnostics[fileName], diagnostic, CompareDiagnostics)
+		if c.fileDiagnosticsSorted == nil {
```
sheetalkamat: Do we really need sorting here as diagnostics are anyways sorted and deduplicated at end points?

jakebailey: It's sorted so `lookup` can find diags later; it's a weird system but yeah

internal/ast/diagnostic.go
```diff
@@ -140,8 +122,10 @@ func NewCompilerDiagnostic(message *diagnostics.Message, args ...any) *Diagnosti
 }
 
 type DiagnosticsCollection struct {
-	fileDiagnostics    map[string][]*Diagnostic
-	nonFileDiagnostics []*Diagnostic
+	fileDiagnostics          map[string][]*Diagnostic
+	fileDiagnosticsSorted    map[string]bool
```
jakebailey: You can just use `core.Set[string]` here, I think

jakebailey: Oops, yes, collections 😅

jakebailey:
This turns out to be unsafe; it causes a data race when concurrently accessing processor diagnostics. We don't see the problem in main, though, only on a WIP branch I have for bringing back #2197.

jakebailey:
And, since it sorts the list on the fly, if something previously grabbed a list of diags (say, the global diags), and then they changed later, the list would potentially be concurrently modified.

PR #2288, opened 2025-12-09
Use xxhash for composite checker keys instead of strings

jakebailey:
Often when I see people's profiles, the top allocator is `WriteByte`, all from the checker building runtime map keys.

We stringify data to use as map keys for union types, intersection types, etc. In JS, this is cheap thanks to the "rope" optimization where the strings are never actually produced unless required (which does not include `Map` lookups!).

But in Go, this doesn't work so well because we do actually have to back strings with real fully-written memory, and it's not easy to do a lookup otherwise. There's no magic string type that can expand itself when read. A Go stdlib map that allows custom hashing/equality is seemingly not coming soon.

An alternative is to just hash the data and use that hash as a key. A while ago I had tried use `sha256` to do this (like gopls does), but did not see much success. And, introducing `crypto` into the binary is a can of worms I don't want to open. So, I gave up on that idea.

We have since then added xxhash to our dependencies for incremental mode, the LSP, etc.

A 128-bit xxhash could be argued to have enough collision resistance; UUIDs are of course also 128-bit. What are the chances of a UUID-level collision happening within a single compile? 😅 

This PR tries that out, swapping our key builder out for one that hashes the data as it comes in instead.

The effect seems pretty good. VS Code before:

```
Memory used:    3565146K
Memory allocs:  34630016
Config time:      0.130s
Parse time:       0.520s
Bind time:        0.097s
Check time:       6.276s
Emit time:        1.394s
Total time:       8.425s
```

And after:

```
Memory used:    3533519K
Memory allocs:  25587150
Config time:      0.132s
Parse time:       0.549s
Bind time:        0.085s
Check time:       6.136s
Emit time:        1.378s
Total time:       8.289s
```

The time in general seems unchanged, but with 35% fewer allocations.

Downside is of course that the map keys become unreadable and irreversible. Right now, they're only "unreadable" thanks to us making numbers into shorter strings. We could have a mode which preserves the data behind a build tag. Not sure about that.

Thankfully, these keys are never actually exposed through any API, so should be opaque enough.

### DIFF
--- a/internal/checker/flow.go
+++ b/internal/checker/flow.go
@@ -12,6 +12,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 	"github.com/microsoft/typescript-go/internal/evaluator"
 	"github.com/microsoft/typescript-go/internal/scanner"
+	"github.com/zeebo/xxh3"
 )
 
 type FlowType struct {
@@ -1291,7 +1292,7 @@ func (c *Checker) getTypeAtFlowLoopLabel(f *FlowState, flow *ast.FlowNode) FlowT
 	if f.refKey.IsZero() {
 		f.refKey = c.getFlowReferenceKey(f)
 	}
-	if f.refKey.IsZero() {
+	if f.refKey == nonDottedNameCacheKey {
 		// No cache key is generated when binding patterns are in unnarrowable situations
 		return FlowType{t: f.declaredType}
 	}
@@ -1607,16 +1608,18 @@ func (c *Checker) isMatchingReference(source *ast.Node, target *ast.Node) bool {
 	return false
 }
 
+var nonDottedNameCacheKey = CacheHashKey(xxh3.HashString128("?"))
+
 // Return the flow cache key for a "dotted name" (i.e. a sequence of identifiers
 // separated by dots). The key consists of the id of the symbol referenced by the
 // leftmost identifier followed by zero or more property names separated by dots.
-// The result is a zero hash if the reference isn't a dotted name.
+// The result is nonDottedNameCacheKey if the reference isn't a dotted name.
 func (c *Checker) getFlowReferenceKey(f *FlowState) CacheHashKey {
 	var b keyBuilder
 	if c.writeFlowCacheKey(&b, f.reference, f.declaredType, f.initialType, f.flowContainer) {
 		return b.hash()
 	}
-	return CacheHashKey{} // Reference isn't a dotted name
+	return nonDottedNameCacheKey // Reference isn't a dotted name
 }
 
 func (c *Checker) writeFlowCacheKey(b *keyBuilder, node *ast.Node, declaredType *Type, initialType *Type, flowContainer *ast.Node) bool {


internal/checker/checker.go
```diff
@@ -161,23 +162,23 @@ type UnionOfUnionKey struct {
 	id1 TypeId
 	id2 TypeId
 	r   UnionReduction
-	a   string
+	a   xxh3.Uint128
```
ahejlsberg: Maybe we should declare a friendlier local name for this?

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
-	b.WriteByte(',')
-	b.WriteType(indexType)
-	b.WriteByte(',')
-	b.WriteUint64(uint64(accessFlags))
-	b.WriteAlias(alias)
-	return b.String()
+func getIndexedAccessKey(objectType *Type, indexType *Type, accessFlags AccessFlags, alias *TypeAlias) xxh3.Uint128 {
+	var b keyBuilder
+	b.writeType(objectType)
+	b.writeByte(',')
```
ahejlsberg: No need for this separator since `writeType` now writes a fixed number of bytes.

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
+	b.writeType(indexType)
+	b.writeByte(',')
```
ahejlsberg: No need for this either.

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
-	b.WriteTypes(types)
-	b.WriteByte('|')
+func getTemplateTypeKey(texts []string, types []*Type) xxh3.Uint128 {
+	var b keyBuilder
+	b.writeTypes(types)
+	b.writeByte('|')
 	for i, s := range texts {
 		if i != 0 {
-			b.WriteByte(',')
+			b.writeByte(',')
```
ahejlsberg: No need for this.

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
 
-func getRelationKey(source *Type, target *Type, intersectionState IntersectionState, isIdentity bool, ignoreConstraints bool) string {
+func getRelationKey(source *Type, target *Type, intersectionState IntersectionState, isIdentity bool, ignoreConstraints bool) (xxh3.Uint128, bool) {
 	if isIdentity && source.id > target.id {
 		source, target = target, source
 	}
-	var b KeyBuilder
+	var b keyBuilder
 	var constrained bool
 	if isTypeReferenceWithGenericArguments(source) && isTypeReferenceWithGenericArguments(target) {
```
ahejlsberg, line 17111: I think we need something to change here. The old code relied on the fact that `writeGenericTypeReferences` would write strings that never start with a character that could occur in a `writeType` string. That's no longer the case because `writeType` now writes four bytes that could be anything. I think we need to write an explicit distinguishing prefix byte to indicate which branch we took.

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
-		constrained = b.WriteGenericTypeReferences(source, target, ignoreConstraints)
+		constrained = b.writeGenericTypeReferences(source, target, ignoreConstraints)
 	} else {
-		b.WriteType(source)
-		b.WriteByte(',')
-		b.WriteType(target)
+		b.writeType(source)
+		b.writeByte(',')
```
ahejlsberg: No need for this.

internal/checker/checker.go
```diff
@@ -16951,193 +16956,186 @@ func (b *KeyBuilder) WriteGenericTypeReferences(source *Type, target *Type, igno
...
+		b.writeType(target)
 	}
 	if intersectionState != IntersectionStateNone {
```
ahejlsberg: Just unconditionally write the 32-bit intersection state.

internal/checker/flow.go
```diff
@@ -1288,10 +1290,10 @@ func (c *Checker) getUnionOrEvolvingArrayType(f *FlowState, types []*Type, subty
 }
 
 func (c *Checker) getTypeAtFlowLoopLabel(f *FlowState, flow *ast.FlowNode) FlowType {
-	if f.refKey == "" {
-		f.refKey = c.getFlowReferenceKey(f)
+	if !f.refKeyValid {
+		f.refKey, f.refKeyValid = c.getFlowReferenceKey(f)
```
ahejlsberg: Can we just use the zero value here to indicate an invalid key? I mean, that's no more likely to collide than a regular valid key.

ahejlsberg:
In general, in places where we now write a fixed number of bytes (e.g. four bytes for a type ID), we don't need separators between multiple consecutive IDs.

internal/checker/flow.go
jakebailey: Yeah, I was flip flopping between them.

internal/checker/checker.go
jakebailey: Added a byte before each.

internal/checker/checker.go
```diff
@@ -161,23 +162,23 @@ type UnionOfUnionKey struct {
...
-	SignatureKeyOuter          string = ">"
-	SignatureKeyImplementation string = "+"
+var (
+	SignatureKeyErased         = CacheHashKey(xxh3.HashString128("-"))
+	SignatureKeyCanonical      = CacheHashKey(xxh3.HashString128("*"))
+	SignatureKeyBase           = CacheHashKey(xxh3.HashString128("#"))
+	SignatureKeyInner          = CacheHashKey(xxh3.HashString128("<"))
+	SignatureKeyOuter          = CacheHashKey(xxh3.HashString128(">"))
+	SignatureKeyImplementation = CacheHashKey(xxh3.HashString128("+"))
 )
```
jakebailey, line 182: These would be better as functions returning the structs, to actually be constants. It's unfortunate to not have const structs.

internal/checker/checker.go
```diff
@@ -16864,85 +16865,87 @@ func (c *Checker) isThislessInterface(symbol *ast.Symbol) bool {
...
+		byte(v >> 56),
+	})
 }
 
-func (b *KeyBuilder) WriteUint64(value uint64) {
-	for value != 0 {
-		b.WriteByte(base64chars[value&0x3F])
-		value >>= 6
-	}
+type CacheHashKey xxh3.Uint128
```
jakebailey, line 16892: Bikeshed, but could also be `CompositeMapKey`

internal/checker/flow.go
```diff
@@ -1288,10 +1288,10 @@ func (c *Checker) getUnionOrEvolvingArrayType(f *FlowState, types []*Type, subty
 }
 
 func (c *Checker) getTypeAtFlowLoopLabel(f *FlowState, flow *ast.FlowNode) FlowType {
-	if f.refKey == "" {
+	if f.refKey.IsZero() {
 		f.refKey = c.getFlowReferenceKey(f)
 	}
-	if f.refKey == "?" {
+	if f.refKey.IsZero() {
```
jakebailey: I messed up here; I need a sentinel value.

jakebailey: Fixed now, wish I had caught this before asking for a rereview 😄

ahejlsberg (approved):

PR #2302, opened 2025-12-09
API over LSP implementation

aleksei-berezkin:
Hi! This is our attempt to implement a TS-Go API over the LSP protocol using custom messages. We are already actively using it to query types from the `tsgo` server, so we think it could be of interest to you. Here are the key points of this implementation:

* It uses LSP as the transport layer rather than a custom protocol. We find this much easier to maintain. In terms of performance, the transport has never been a bottleneck for us.
* `Type`s and `Symbol`s are loaded lazily. For example, for an `ObjectType`, properties are loaded only when accessed, and their types are also requested on demand. This helps minimize traffic and maintain good performance.
* While this code is fully functional, we understand that the implementation is not complete. Only a subset of fields and types is supported, but we believe it can be developed further into a production-ready solution.

Thank you.


aleksei-berezkin:
@microsoft-github-policy-service agree company="JetBrains"

aleksei-berezkin:
@microsoft-github-policy-service agree company="JetBrains"

nnnnoel:
Hello @aleksei-berezkin , I'm WebStorm user that really hope to support typescript-go 100% compatibility in WebStorm.
Would [this PR](https://github.com/microsoft/typescript-go/pull/2831) be helpful to your works?

aleksei-berezkin:
Hi @nnnnoel, thank you for your comment. You can already use TypeScript-Go in WebStorm: just declare a `@typescript/native-preview` in your dependencies, and make `npm install`, and WebStorm will use it as an LSP server.

The only thing which is not implemented yet is the new "Service-Powered Type Engine" (available in Settings | Languages and Frameworks | TypeScript) because this requires the integration with TypeScript compiler via plugin API. We're following closely the progress of the API implementation; right now it looks very promising, yet some crucial methods are missing. Once they are ready, and we observe some stability, we'll reimplement it for TypeScript-Go. Until then, you can use our fork (available via the same settings), where we implemented the needed API subset for Service-Powered Type Engine to work properly.

aleksei-berezkin:
We're closing this PR. It served as an example implementation, and its main purpose was to demonstrate the way we would like to use the TS-Go API.

For the moment, the TS-Go already has a working API, and, while it's still not officially announced, it seems to be quite robust, and meets most or our our needs. So instead of investing further into our version of the API implmenetaion in [our fork](https://github.com/JetBrains/typescript-go), we will focus on supporting the upstream TS-Go API. Hope to get it working in WebStorm 2026.2 with TS 7.1.

Thanks everyone who took a moment and checked this PR.

PR #2304, opened 2025-12-09
Fix getReferencedSymbolsForSymbol, implement getReferencesAtExportSpecifier

### LINKED ISSUE #2204: [bug] TSGO fails to follow re-exports when finding references
### Extension Version

 0.20251203.1

### VS Code Version

1.106.3 (Universal)

### Operating system Version

MacOS

### Steps to reproduce

Attempt to use find refs against `symbolA` in https://github.com/mjames-c/tsgo-playground

### Issue

TSGO can follow re-exports of symbols when a user uses find refs. Re-exports are pretty common in the Canva code base so having this would be pretty huge for us!

jakebailey:
Fixes #2204

- `getReferencesAtExportSpecifier` had not been ported.
- `getReferencedSymbolsForSymbol` had a condition backwards.

Finding the latter involved debugging a panic; in this PR is also a fix to fourslash to actually print out an error message instead of "nil response".

### DIFF
--- a/internal/fourslash/_scripts/failingTests.txt
+++ b/internal/fourslash/_scripts/failingTests.txt
@@ -539,7 +539,6 @@ TestQuickinfoWrongComment
 TestRecursiveInternalModuleImport
 TestReferencesInEmptyFile
 TestRegexDetection
-TestRenameForAliasingExport02
 TestRenameFromNodeModulesDep1
 TestRenameFromNodeModulesDep2
 TestRenameFromNodeModulesDep3
--- a/internal/fourslash/tests/gen/renameForAliasingExport02_test.go
+++ b/internal/fourslash/tests/gen/renameForAliasingExport02_test.go
@@ -9,7 +9,7 @@ import (
 
 func TestRenameForAliasingExport02(t *testing.T) {
 	t.Parallel()
-	t.Skip()
+
 	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
 	const content = `// @Filename: foo.ts
 let x = 1;


internal/ls/findallreferences.go
```diff
@@ -1528,14 +1528,12 @@ func getReferencedSymbolsForSymbol(originalSymbol *ast.Symbol, node *ast.Node, s
 	state := newState(sourceFiles, sourceFilesSet, node, checker /*, cancellationToken*/, searchMeaning, options)
 
 	var exportSpecifier *ast.Node
-	if !isForRenameWithPrefixAndSuffixText(options) || len(symbol.Declarations) == 0 {
+	if isForRenameWithPrefixAndSuffixText(options) && len(symbol.Declarations) != 0 {
```
gabritto, line 1531: Not specific to this PR, but maybe we want to rename this function since the actual user preference is `useAliasesForRenames`?

jakebailey: It was called `isForRenameWithPrefixAndSuffixText` in Strada, we are just seemingly inconsistent all around...

gabritto: I know, that's what I meant. Strada used `providePrefixAndSuffixTextForRename` as the preference name and related functions, but it was called `useAliasesForRenames` in vscode, and now in Corsa we confusingly have both terms.

jakebailey: Yeah. Probably better fixed in another PR. #2272 at least lets us choose whichever name we want.

PR #2307, opened 2025-12-09
Add --builders to limit number of projects can build concurrently

sheetalkamat:
Per https://github.com/microsoft/typescript-go/issues/1622#issuecomment-3633202774


internal/execute/build/orchestrator.go
```diff
@@ -396,5 +398,9 @@ func NewOrchestrator(opts Options) *Orchestrator {
 	} else {
 		orchestrator.errorSummaryReporter = tsc.CreateReportErrorSummary(opts.Sys, opts.Command.Locale(), opts.Command.CompilerOptions)
 	}
+	// If we want to build more than one project at a time, create a semaphore to limit concurrency
+	if builders := opts.Command.BuildOptions.Builders; builders != nil {
+		orchestrator.buildSemaphore = make(chan struct{}, *builders)
```
jakebailey, line 403: Do we need to `max(1, *builders)` like we do for the checker count?

jakebailey: Ah, I see that you added something to config parsing for this.

internal/execute/build/buildtask.go
```diff
@@ -200,6 +200,10 @@ func (t *BuildTask) updateDownstream(orchestrator *Orchestrator, path tspath.Pat
 }
 
 func (t *BuildTask) compileAndEmit(orchestrator *Orchestrator, path tspath.Path) {
+	if orchestrator.buildSemaphore != nil {
+		orchestrator.buildSemaphore <- struct{}{}        // acquire slot
+		defer func() { <-orchestrator.buildSemaphore }() // release slot
```
jakebailey, line 205: Just to double check, we don't actually load anything related to the project until after this point, right?

jakebailey (approved):
I tested, and this seems to work. On the old compiler, `--builders 1` performs the same as letting it be unconstrained. I need to find a better test case.

internal/execute/build/buildtask.go
sheetalkamat: correct. we actually create program and do emit after this.
this does not affect uptodate ness checks

PR #2309, opened 2025-12-09
Fix contextual type for array completions

gabritto:
Fixes some problems caused by #2258:

1. If we're after a comma, we may also be inside an argument list and should go through the code for that when the parent is not an array literal.
2. Unlike the comment Copilot added, we can't just pass -1 as the spread indices, it'll result in a wrong contextual type if there are spread elements present. I think -1 length is ok to use, since it'll just opt us out of counting from the end of the array, which I think we should avoid since the array literal may be incomplete.

### DIFF
--- a/internal/ls/completions.go
+++ b/internal/ls/completions.go
@@ -1369,7 +1369,10 @@ func (l *LanguageService) getCompletionData(
 		if importAttributes.AsImportAttributes().Attributes != nil {
 			elements = importAttributes.AsImportAttributes().Attributes.Nodes
 		}
-		existing := collections.NewSetFromItems(core.Map(elements, (*ast.Node).Text)...)
+		attributeNames := core.Map(elements, func(el *ast.Node) string {
+			return el.AsImportAttribute().Name().Text()
+		})
+		existing := collections.NewSetFromItems(attributeNames...)
 		uniques := core.Filter(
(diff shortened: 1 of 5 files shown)


internal/ls/completions.go
```diff
@@ -3023,22 +3003,30 @@ func getContextualType(previousToken *ast.Node, position int, file *ast.SourceFi
...
-		if argInfo != nil {
-			return typeChecker.GetContextualTypeForArgumentAtIndex(argInfo.invocation, argInfo.argumentIndex)
-		} else if isEqualityOperatorKind(previousToken.Kind) && ast.IsBinaryExpression(parent) && isEqualityOperatorKind(parent.AsBinaryExpression().OperatorToken.Kind) {
-			// completion at `x ===/**/`
-			return typeChecker.GetTypeAtLocation(parent.AsBinaryExpression().Left)
-		} else {
-			contextualType := typeChecker.GetContextualType(previousToken, checker.ContextFlagsCompletions)
-			if contextualType != nil {
-				return contextualType
+	case ast.KindCommaToken:
```
DanielRosenwasser, line 3006: I might still co-locate the `[` and `]` cases, but this is fine.

I was thinking about whether there are cases where we should also guard on `]`, but I couldn't imagine where you'd want or have a contextual type immediately after a `]`.

PR #2314, opened 2025-12-10
Reapply Program diagnostic func update, CheckerPool cleanup

jakebailey:
This reapplies #2197 after it was reverted in #2250.

This includes a few changes over #2197.

- Actually check concurrently. Oops.
  - Caveat; each file gets its own goroutine. I measure no difference, and checkers are only locked on for the duration of their checks. Not needing to loop over the checkers is IMO a good simplification and means we have the same flow for both the CLI and for the LS.
- Make `DiagnosticsCollection` concurrency safe, and return copies of diags. See: https://github.com/microsoft/typescript-go/pull/2283#issuecomment-3635213002
- Remove the code that orchestrates multi-file checking in checkers themselves. The callers do this; nothing was actually asking the checker to check all files. Now there is no exported `CheckSourceFile` method.
  - This technically also means we don't need to store the file list in the checker anymore. That cleanup is not in this PR. It's sort of inconsequential given we have to give those to the checker anyway for other reasons.
- Move `SkipTypeChecking` to a `Program` method, now that the callers are simplified. We were doubly checking this for every file. Oops.
  - `IsSourceFromProjectReference` goes away on `checker.Program`.
- Remove `Count` and `Files` from `CheckerPool` (as I did `ForEachCheckerParallel` before). These are only used in the CLI compilation, and they have been deleted out of the LSP checker pool.

There are further refactors I haven't done that should be possible in the future:

- `CheckerPool` does not really need `GetTypeChecker` anymore. The only callers of this are \~mistakenly not asking for the checker for a given file. Thankfully, these are LS users where the request-scoped checker system avoids problems. It's not a hard to fix this but it is somewhat noisy. Might stick in this PR but the ported code is a bit wonky (my fault in many of these cases...).
- In this cleanup, I've discovered that `maxCheckers` on `project.CheckerPool` does nothing. Not sure what to make of that but it seems scary.
- And of course, the change that makes us able to use differing checkers for diags and LS features, now that there are way fewer code paths to implement.

### DIFF
--- a/internal/ast/diagnostic.go
+++ b/internal/ast/diagnostic.go
@@ -174,9 +174,9 @@ func (c *DiagnosticsCollection) Lookup(diagnostic *Diagnostic) *Diagnostic {
 
 	var diagnostics []*Diagnostic
 	if diagnostic.File() != nil {
-		diagnostics = c.getDiagnosticsForFile(diagnostic.File().FileName())
+		diagnostics = c.getDiagnosticsForFileLocked(diagnostic.File().FileName())
 	} else {
-		diagnostics = c.getGlobalDiagnostics()
+		diagnostics = c.getGlobalDiagnosticsLocked()
 	}
 	if i, ok := slices.BinarySearchFunc(diagnostics, diagnostic, CompareDiagnostics); ok {
 		return diagnostics[i]
@@ -188,10 +188,10 @@ func (c *DiagnosticsCollection) GetGlobalDiagnostics() []*Diagnostic {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	return c.getGlobalDiagnostics()
+	return c.getGlobalDiagnosticsLocked()
 }
 
-func (c *DiagnosticsCollection) getGlobalDiagnostics() []*Diagnostic {
+func (c *DiagnosticsCollection) getGlobalDiagnosticsLocked() []*Diagnostic {
 	if !c.nonFileDiagnosticsSorted {
 		slices.SortStableFunc(c.nonFileDiagnostics, CompareDiagnostics)
 		c.nonFileDiagnosticsSorted = true
@@ -203,10 +203,10 @@ func (c *DiagnosticsCollection) GetDiagnosticsForFile(fileName string) []*Diagno
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	return c.getDiagnosticsForFile(fileName)
+	return c.getDiagnosticsForFileLocked(fileName)
 }
 
-func (c *DiagnosticsCollection) getDiagnosticsForFile(fileName string) []*Diagnostic {
+func (c *DiagnosticsCollection) getDiagnosticsForFileLocked(fileName string) []*Diagnostic {
 	if !c.fileDiagnosticsSorted.Has(fileName) {
 		slices.SortStableFunc(c.fileDiagnostics[fileName], CompareDiagnostics)
 		c.fileDiagnosticsSorted.Add(fileName)


internal/ast/diagnostic.go
```diff
@@ -141,13 +141,20 @@ func NewCompilerDiagnostic(message *diagnostics.Message, args ...any) *Diagnosti
 }
 
 type DiagnosticsCollection struct {
+	mu                       sync.Mutex
+	count                    int
 	fileDiagnostics          map[string][]*Diagnostic
 	fileDiagnosticsSorted    collections.Set[string]
```
DanielRosenwasser, line 147: While we're here, I wonder if you're open to renaming this `filesWithSortedDiagnostics` or something

internal/checker/checker_test.go
```diff
@@ -61,25 +61,6 @@ foo.bar;`
 	}
 }
 
-func TestCheckSrcCompiler(t *testing.T) {
```
DanielRosenwasser, line 64: Wow that's funny - so every full test run was checking `TypeScript/src/compiler`?

internal/compiler/program.go
```diff
@@ -985,41 +1031,29 @@ func (p *Program) GetGlobalDiagnostics(ctx context.Context) []*ast.Diagnostic {
 		return nil
 	}
 
-	globalDiagnostics := make([][]*ast.Diagnostic, p.checkerPool.Count())
-	p.checkerPool.ForEachCheckerParallel(ctx, func(idx int, checker *checker.Checker) {
+	pool := p.checkerPool.(*checkerPool)
```
DanielRosenwasser, line 1034: Why the interface and struct implementation if we're always going to cast to a concrete instance in these methods?

internal/project/checkerpool.go
```diff
@@ -80,7 +79,7 @@ func (p *CheckerPool) GetCheckerForFile(ctx context.Context, file *ast.SourceFil
 // GetCheckerForFileExclusive is the same as GetCheckerForFile but also locks a mutex associated with the checker.
 // Call `done` to free the lock.
 func (p *CheckerPool) GetCheckerForFileExclusive(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func()) {
-	panic("unimplemented") // implement if used by LS
+	return p.GetCheckerForFile(ctx, file)
```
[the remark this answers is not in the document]
DanielRosenwasser: Wait I don't understand, why was this marked resolved. One method just calls the other, right?

internal/checker/checker_test.go
jakebailey: Yeah but it was quick.

internal/compiler/program.go
jakebailey: Because this is also provided by the language server, which doesn't provide these methods.

internal/project/checkerpool.go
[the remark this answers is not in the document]
jakebailey: Exclusiveness is handled by the pool here already, in that a checked is associated with a request.

internal/compiler/program.go
jakebailey: Though, you made me realize that I changed this such that it fails here, which means that we haven't been getting global diags in the editor, which is sort of a problem but not new...

jakebailey: To be clearer, the LS never actually calls this function. I think that's in a bug, but fixing that is turning into something more annoying so I'll probably follow that up

internal/ast/diagnostic.go
jakebailey: I don't mind the existing name, personally...

jakebailey:
I'm going to merge this as-is; I have a followup about options diags and another for the GetTypeChecker thing but they are not critical.

PR #2322, opened 2025-12-10
Remove GetOptionsDiagnostics, expose global diags in LSP

jakebailey:
"Options diagnostics" were just config file diagnostics + global diagnostics.

This was odd, because we then took that in places and combined it with... config file diags and global diags. And then not deduping it, sometimes.

This PR just removes `GetOptionsDiagnostics` altogether, instead leaving behind just `GetConfigFileParsingDiagnostics` and `GetGlobalDiagnostics`.

~The language server still does not ever call `GetGlobalDiagnostics`. I am not sure where such a thing would go, since global diagnostics are generated as a side effect of doing checking (e.g., when we figure out that `AsyncIterable` was never declared or something).~

The language server now keeps track of global diagnostics on the fly as it creates and destroys checkers. The diags are eventually bubbled out as diags on the tsconfig.

There's a baseline change, but I think it's correct and positive.

### DIFF
--- a/internal/project/project_test.go
+++ b/internal/project/project_test.go
@@ -326,7 +326,8 @@ func TestPushDiagnostics(t *testing.T) {
 		// Request semantic diagnostics to trigger checking, which triggers the global type resolvers.
 		ls, err := session.GetLanguageService(projecttestutil.WithRequestID(context.Background()), lsproto.DocumentUri("file:///src/index.ts"))
 		assert.NilError(t, err)
-		ls.ProvideDiagnostics(projecttestutil.WithRequestID(context.Background()), lsproto.DocumentUri("file:///src/index.ts"))
+		_, err = ls.ProvideDiagnostics(projecttestutil.WithRequestID(context.Background()), lsproto.DocumentUri("file:///src/index.ts"))
+		assert.NilError(t, err)
 		// Enqueue global diagnostics publishing (normally done by the LSP server after each request).
 		session.EnqueuePublishGlobalDiagnostics()
 		session.WaitForBackgroundTasks()


internal/compiler/program.go
```diff
@@ -1031,7 +1031,10 @@ func (p *Program) GetGlobalDiagnostics(ctx context.Context) []*ast.Diagnostic {
 		return nil
 	}
 
-	pool := p.checkerPool.(*checkerPool)
+	pool, ok := p.checkerPool.(*checkerPool)
+	if !ok {
+		return nil // !!! Global diagnostics in the editor?
```
jakebailey: Whenever we get around to revamping how the checkers are allocated, I think the fix is going to be to (in the editor) collect global diags from diagnostic-producing checkers, and then re-collect them whenever we check a new file, then expose those diags there.

internal/compiler/program.go
```diff
@@ -1045,17 +1048,6 @@ func (p *Program) GetDeclarationDiagnostics(ctx context.Context, sourceFile *ast
 	return p.collectDiagnostics(ctx, sourceFile, true /*concurrent*/, p.getDeclarationDiagnosticsForFile)
 }
 
-func (p *Program) GetOptionsDiagnostics(ctx context.Context) []*ast.Diagnostic {
-	return SortAndDeduplicateDiagnostics(core.Concatenate(p.GetGlobalDiagnostics(ctx), p.getOptionsDiagnosticsOfConfigFile()))
-}
-
-func (p *Program) getOptionsDiagnosticsOfConfigFile() []*ast.Diagnostic {
-	if p.Options() == nil || p.Options().ConfigFilePath == "" {
```
jakebailey, line 1191: This used to be here because we could fail to get file parsing diags. But then we made `ParsedCommandLine` mandatory, so this call always succeeds, so there's no reason to special case anything.

internal/compiler/program.go
```diff
@@ -1494,13 +1485,9 @@ func GetDiagnosticsOfAnyProgram(
 	allDiagnostics = append(allDiagnostics, program.GetProgramDiagnostics()...)
 
 	if len(allDiagnostics) == configFileParsingDiagnosticsLength {
-		// Options diagnostics include global diagnostics (even though we collect them separately),
-		// and global diagnostics create checkers, which then bind all of the files. Do this binding
-		// early so we can track the time.
+		// Do binding early so we can track the time.
 		getBindDiagnostics(ctx, file)
 
-		allDiagnostics = append(allDiagnostics, program.GetOptionsDiagnostics(ctx)...)
```
jakebailey, line 1648: Note above how we get config file diags, then get options diags, and then get global diags. This means we could have the same diags repeated twice!

(This func does not sort and dedupe, for some reason)

internal/execute/incremental/program.go
```diff
@@ -365,7 +359,6 @@ func (p *Program) ensureHasErrorsForState(ctx context.Context, program *compiler
 		len(program.GetConfigFileParsingDiagnostics()) > 0 ||
 		len(program.GetSyntacticDiagnostics(ctx, nil)) > 0 ||
 		len(program.GetProgramDiagnostics()) > 0 ||
-		len(program.GetOptionsDiagnostics(ctx)) > 0 ||
```
jakebailey, line 368: This list already contains `GetConfigFileParsingDiagnostics` and `GetGlobalDiagnostics`, so this is a noop.

internal/compiler/program.go
```diff
@@ -1494,18 +1485,16 @@ func GetDiagnosticsOfAnyProgram(
...
-
 		if program.Options().ListFilesOnly.IsFalseOrUnknown() {
 			allDiagnostics = append(allDiagnostics, program.GetGlobalDiagnostics(ctx)...)
 
 			if len(allDiagnostics) == configFileParsingDiagnosticsLength {
 				allDiagnostics = append(allDiagnostics, getSemanticDiagnostics(ctx, file)...)
+				// Ask for the global diagnostics again (they were empty above); we may have found new during checking, e.g. missing globals.
+				allDiagnostics = append(allDiagnostics, program.GetGlobalDiagnostics(ctx)...)
```
sheetalkamat, line 1632: Strada use to add them to semantic Diagnostics - this ensured that all callers of this API didnt have to worry about it?

jakebailey: We can do that, if and only if the caller passes in a `nil` file, theoretically.

jakebailey: Pushed a change which I think works to delete `GetGlobalDiagnostics` too.

sheetalkamat: well - strada did the exact opposite. It added to "per file semanticDiagnostics" because thats how editor gets diagnostics and it shouldnt miss them?

```typescript
function getDiagnosticsWorker(sourceFile: SourceFile, nodesToCheck: Node[] | undefined): Diagnostic[] {
        if (sourceFile) {
            ensurePendingDiagnosticWorkComplete();
            // Some global diagnostics are deferred until they are needed and
            // may not be reported in the first call to getGlobalDiagnostics.
            // We should catch these changes and report them.
            const previousGlobalDiagnostics = diagnostics.getGlobalDiagnostics();
            const previousGlobalDiagnosticsSize = previousGlobalDiagnostics.length;

            checkSourceFileWithEagerDiagnostics(sourceFile, nodesToCheck);
            const semanticDiagnostics = diagnostics.getDiagnostics(sourceFile.fileName);
            if (nodesToCheck) {
                // No need to get global diagnostics.
                return semanticDiagnostics;
            }
            const currentGlobalDiagnostics = diagnostics.getGlobalDiagnostics();
            if (currentGlobalDiagnostics !== previousGlobalDiagnostics) {
                // If the arrays are not the same reference, new diagnostics were added.
                const deferredGlobalDiagnostics = relativeComplement(previousGlobalDiagnostics, currentGlobalDiagnostics, compareDiagnostics);
                return concatenate(deferredGlobalDiagnostics, semanticDiagnostics);
            }
            else if (previousGlobalDiagnosticsSize === 0 && currentGlobalDiagnostics.length > 0) {
                // If the arrays are the same reference, but the length has changed, a single
                // new diagnostic was added as DiagnosticCollection attempts to reuse the
                // same array.
                return concatenate(currentGlobalDiagnostics, semanticDiagnostics);
            }

            return semanticDiagnostics;
        }

        // Global diagnostics are always added when a file is not provided to
        // getDiagnostics
        forEach(host.getSourceFiles(), file => checkSourceFileWithEagerDiagnostics(file));
        return diagnostics.getDiagnostics();
    }
```

sheetalkamat: Not suire if we have enough coverage for diagnostics in fourslash but all this logic was for editor to not miss reporting errors as for command line you can always get all semantic diagnostics and then global diagnsotics?

jakebailey: Hm. With concurrent checking, that might be a bit challenging. But I suppose we could somehow save those off during `getBindAndCheckDiagnosticsForFile`. The problem is that it'll be nondeterministic depending on the order in which files are checked, I think.

jakebailey: > Not suire if we have enough coverage for diagnostics in fourslash but all this logic was for editor to not miss reporting errors as for command line you can always get all semantic diagnostics and then global diagnsotics?

The current LS can miss these global diagnostics, yes. Though, that isn't new...

jakebailey: The last commit I did was bad, though, because it locks off ever asking for global diags in the LS where we never ask for all files, so I'll undo that

sheetalkamat: > > Not suire if we have enough coverage for diagnostics in fourslash but all this logic was for editor to not miss reporting errors as for command line you can always get all semantic diagnostics and then global diagnsotics?
> 
> The current LS can miss these global diagnostics, yes. Though, that isn't new...

It is new right, given we use to add the new diagnostics to semantic diagnostics and send it back to client ?


> getBindAndCheckDiagnosticsForFile

It shouldnt be part of this as these are the ones that are cached in buildInfo and need to be consistent

jakebailey: I'm confused enough with how to make this work that I wish we just didn't have global diags at all, they just really do depend on whether or not a file was checked and in which order, which does not feel reasonable to manage?

sheetalkamat: Havent checked how diagnostics are handled in corsa but with push diagnostics we can control the order but not sure how to handle that as part of pull diagnostics esp since global diagnostics are checker diagnostics without any file info. (may be we just need to make sure that checker cannot generate global diagnostics) ?

internal/compiler/program.go
sheetalkamat: Yes before this the duplicae config parsing diagnostics is why we didnt report semantic errors which we are doing now (that is when there are no syntax errors in source files and program options are ok.

andrewbranch:
> global diagnostics are generated as a side effect of doing checking (e.g., when we figure out that AsyncIterable was never declared or something).

Isn't it the case that some diagnostics are generated during checker initialization due to symbol table merging? Or are those only ever duplicate declaration diagnostics, which always have a location?

It seems like maybe a distinction we should have instead of whatever mess we inherited from Strada is diagnostics that got generated during initialization, because those (whether they have a location or not) will always reliably be there regardless of what files have been checked in what order.

jakebailey:
Yes, there are diagnostics that are generated without a location, so can't be exposed any other way.

andrewbranch:
I know that some diagnostics are generated without a location; I'm asking whether "has a location" is orthogonal to "is dependent on what files have been checked in what order" (I think it is, at least in part)

jakebailey:
Ah. I think we still can report global diagnostics on things like "Iterable is missing", which only happens when you finally ask for it somewhere. Would need to recheck.

andrewbranch:
I think this change is fine as-is; I was just suggesting that it could be useful to have one bucket of diagnostics specifically from checker initialization because that would be known to be populated the exact same way no matter how many files you have or haven't checked. If that _happened_ to include any global (location-less) diagnostics, we could reliably report them in the LSP without jumping through extra hoops. But I recognize that it wouldn't contain _all_ the global diagnostics (and maybe in practice it wouldn't include any). The broader question of global diagnostics in the LSP will have to be figured out later—as you said, it's not new in this PR. I think Sheetal meant that it's new in Corsa, which is true I guess. I would approve this if you're ready to undraft it.

internal/project/checkerpool.go
```diff
@@ -166,13 +171,49 @@ func (p *CheckerPool) createRelease(requestId string, index int, checker *checke
...
+	globals := c.GetGlobalDiagnostics()
+	if len(globals) == p.globalDiagCheckerCount[index] {
+		return
+	}
+	p.globalDiagCheckerCount[index] = len(globals)
+	before := len(p.globalDiagAccumulated)
+	p.globalDiagAccumulated = compiler.SortAndDeduplicateDiagnostics(append(p.globalDiagAccumulated, globals...))
+	if len(p.globalDiagAccumulated) != before {
+		p.globalDiagChanged = true
+	}
```
jakebailey, line 196: Okay, this is my new fresh idea. The language server keeps track of global diagnostics, merging them into a big list that's kept even if checkers go away. Those get exposed as diags on the tsconfig, and it's all checked periodically.

internal/compiler/checkerpool.go
```diff
@@ -14,6 +14,7 @@ type CheckerPool interface {
 	GetChecker(ctx context.Context) (*checker.Checker, func())
 	GetCheckerForFile(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func())
 	GetCheckerForFileExclusive(ctx context.Context, file *ast.SourceFile) (*checker.Checker, func())
+	GetGlobalDiagnostics() []*ast.Diagnostic
```
jakebailey, line 17: There's technically not a great reason to bother putting this in here. In a future PR, I'm going to have to redo the way the pool works again anyway, but I can undo this specific bit and shove it back into Program as a special case.

jakebailey:
Probably should just take #3285

PR #2329, opened 2025-12-10
Properly format `@example` in quick info

### LINKED ISSUE #2313: Bare TypeScript code in `@example` is interpreted as Markdown/HTML
## Steps to reproduce

<div lang=ja>

```tsx
/**
 * @example
 * <Ruby text="ほっけ">𩸽</Ruby>
 * <Ruby text="ビャンビャン">𰻞𰻞</Ruby>麺
 */
export const Ruby = ({ children, text }: { children: string; text: string }) => (
  <ruby>
    {children}
    <rp>（</rp>
    <rt>{text}</rt>
    <rp>）</rp>
  </ruby>
);
```

</div>

1. The above file is `.tsx` file
2. Set `typescript.experimental.useTsgo` to `true` in `settings.json`
3. See the description of `Ruby` by holding your mouse pointer on "`Ruby`"
4. See how `@example` is rendered

Note: this occurs even in non-JSX `.ts` files (e.g. tokens are not highlighted or newlines are removed), but is not as critical as in `.tsx` files.

Interpreted as TS code:

Image

insiders.vscode.dev

## Behavior with `tsgo`

Interpreted as Markdown/HTML:

Image

Workaround: wrap it with ` ```tsx `–` ```` `

Image

This also works in Strada:

Image

To fix (Strada probably adopts this method):

1. Find a code block marker (` ``` `)
2. If exists, treat the content as Markdown
3. If missing, treat the content as TypeScript code

ahejlsberg:
Fixes #2313.

### DIFF
--- a/internal/ls/hover.go
+++ b/internal/ls/hover.go
@@ -113,7 +113,7 @@ func (l *LanguageService) getDocumentationFromDeclaration(c *checker.Checker, de
 					if tag.Kind == ast.KindJSDocTag && tag.TagName().Text() == "example" {
 						b.WriteString("\n")
 						commentText := strings.TrimRight(getCommentText(comments), " \t\r\n")
-						if len(commentText) >= 6 && strings.HasPrefix(commentText, "```") && strings.HasSuffix(commentText, "```") {
+						if len(commentText) > 6 && strings.HasPrefix(commentText, "```") && strings.HasSuffix(commentText, "```") && strings.Contains(commentText, "\n") {
 							b.WriteString(commentText)
 							b.WriteString("\n")
 						} else {


testdata/baselines/reference/fourslash/quickInfo/quickInfoForJSDocCodefence.baseline
```diff
@@ -36,9 +36,12 @@
 // | ```
 // | 
 // | 
-// | *@example* — ``
+// | *@example*
+// | ```tsx
```
jakebailey, line 40: I don't know if this one should work; the old baseline was:


```
=== /tests/cases/fourslash/quickInfoForJSDocCodefence.ts ===
// /**
//  * @example
//  * ```
//  * 1 + 2
//  * ```
//  */
// function foo() {
//          ^^^
// | function foo(): string
// | @example ```
// | 1 + 2
// | ```
//     return '2';
// }
// /**
//  * @example
//  * ``
//  * 1 + 2
//  * `
//  */
// function boo() {
//          ^^^
// | function boo(): string
// | @example ``
// | 1 + 2
// | `
//     return '2';
// }
```

ahejlsberg: I think the new baselines look right. Unless the `@example` text starts and ends with exactly three backticks, we want to put a code fence around it.

jakebailey: I guess I don't know what the original intent of this weird case was, so it's probably fine.

Technically, one can use an arbitrary number of opening backticks to be able to stick even more backticks into a block, but it's probably okay to ignore that for now

internal/ls/hover.go
```diff
@@ -109,14 +110,19 @@ func (l *LanguageService) getDocumentationFromDeclaration(c *checker.Checker, de
 						}
 					}
 					comments := tag.Comments()
-					if len(comments) != 0 {
-						if commentHasPrefix(comments, "```") {
+					if tag.Kind == ast.KindJSDocTag && tag.TagName().Text() == "example" {
+						b.WriteString("\n")
+						commentText := strings.TrimRight(getCommentText(comments), " \t\r\n")
+						if len(commentText) >= 6 && strings.HasPrefix(commentText, "```") && strings.HasSuffix(commentText, "```") {
```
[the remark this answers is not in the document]
ahejlsberg: Reasonable suggestion. Will fix.

testdata/baselines/reference/fourslash/quickInfo/quickInfoForJSDocCodefence.baseline
ahejlsberg: We really just want to recognize code that's already fenced and avoid adding another fence and the current check is fine for that. The `writeCode` helper function we use to fence an un-fenced block does indeed ensure we use a sufficient number of backticks.

PR #2331, opened 2025-12-10
Organize Imports

johnfav03:
Implementation of organize imports


internal/ls/change/trackerimpl.go
```diff
@@ -126,9 +126,13 @@ func (t *Tracker) getNonformattedText(node *ast.Node, sourceFile *ast.SourceFile
...
 	eofToken := t.Factory.NewToken(ast.KindEndOfFile)
 	if ast.IsStatement(node) {
+		text := ""
+		if node.Pos() >= 0 {
+			text = sourceFile.Text()
+		}
 		nodeIn = t.Factory.NewSourceFile(
 			ast.SourceFileParseOptions{FileName: sourceFile.FileName(), Path: sourceFile.Path()},
-			"",
+			text,
```
jakebailey, line 138: It's sort of spooky that this is needed; my assumption here was that the node is _not_ the one from the original source file, so the text won't line up, and therefore we want empty so that we know when we are mistakenly accessing it. Is that not the case here?

internal/ls/organizeimports/organizeimports.go
```diff
@@ -2,11 +2,25 @@ package organizeimports
...
 import (
 	"cmp"
+	"context"
 	"math"
+	"slices"
+	"strings"
+	"unicode"
+
+	"golang.org/x/text/collate"
+	"golang.org/x/text/language"
```
jakebailey, line 7: I'd remove this empty line, for consistency with other files.

internal/ls/organizeimports/organizeimports.go
```diff
@@ -2,11 +2,25 @@ package organizeimports
 
 import (
 	"cmp"
+	"context"
 	"math"
+	"slices"
+	"strings"
+	"unicode"
+
+	"golang.org/x/text/collate"
```
jakebailey: It's unfortunate that this is required, given this package is basically unmaintained and outdated. Hopefully we can get away with it.

internal/ls/codeactions.go
```diff
@@ -47,12 +50,41 @@ var codeFixProviders = []*CodeFixProvider{
...
 	program, file := l.getProgramAndFile(params.TextDocument.Uri)
 
 	var actions []lsproto.CommandOrCodeAction
 
+	// Handle source actions (like organize imports)
+	if params.Context != nil && params.Context.Only != nil {
+		for _, kind := range *params.Context.Only {
+			switch kind {
+			case lsproto.CodeActionKindSourceOrganizeImports:
+				organizeAction := l.createOrganizeImportsAction(ctx, program, file, params.TextDocument.Uri, organizeimports.OrganizeImportsModeAll)
```
jakebailey: We're repeating code here; could `organizeimports` just use `lsproto`'s constants and then the switch goes away?

(The new code actions could be declared there too, probably)

internal/ls/codeactions.go
```diff
@@ -96,6 +128,62 @@ func (l *LanguageService) ProvideCodeActions(ctx context.Context, params *lsprot
...
+
+	// Determine the code action kind based on the mode
+	var kind lsproto.CodeActionKind
+	switch mode {
+	case organizeimports.OrganizeImportsModeSortAndCombine:
+		kind = CodeActionKindSourceOrganizeImportsSortAndCombine
+	case organizeimports.OrganizeImportsModeRemoveUnused:
+		kind = CodeActionKindSourceOrganizeImportsRemoveUnused
+	default:
+		kind = lsproto.CodeActionKindSourceOrganizeImports
```
jakebailey: Yeah, this is the kind of thing I think we could avoid.

internal/ls/organizeimports/organizeimports.go
johnfav03: I think this is enforced by goimports, it automatically adds the empty line on file save

jakebailey: We don't use `goimports`, though. Maybe your editor is misconfigured? I'd check that you have `settings.template.json` copied to `settings.json`.

internal/ls/organizeimports/organizeimports.go
```diff
@@ -52,12 +45,11 @@ func OrganizeImports(
 	changeTracker *change.Tracker,
 	program *compiler.Program,
 	preferences *lsutil.UserPreferences,
-	mode OrganizeImportsMode,
+	kind lsproto.CodeActionKind,
 ) {
-	shouldSort := mode == OrganizeImportsModeSortAndCombine || mode == OrganizeImportsModeAll
+	shouldSort := kind == "source.organizeImports.sortAndCombine" || kind == lsproto.CodeActionKindSourceOrganizeImports
```
jakebailey: Can you just define these constants in the `lsproto` package instead of inlining them?

internal/ls/change/trackerimpl.go
iisaduan: I agree, very spooky. What goes wrong if you dont do this change?

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -3041,6 +3239,10 @@ function generateQuickInfoCommand({ kind, marker, text, docs }: VerifyQuickInfoC
     }
 }
 
+function generateOrganizeImports({ expectedContent, mode, preferences }: VerifyOrganizeImportsCmd): string {
+    return `f.VerifyOrganizeImports(t, ${getGoMultiLineStringLiteral(expectedContent)}, ${mode}, ${preferences})`;
```
iisaduan: Because the expected content can be rather long, the tests become a lot easier to read with newlines between the args

```suggestion
    return `f.VerifyOrganizeImports(t,
        ${getGoMultiLineStringLiteral(expectedContent)},
        ${mode},
        ${preferences}
    )`;
```

internal/ls/change/trackerimpl.go
johnfav03: nodeIn is passed to Write(), and further down the call stack functions like SkipTrivia() use p.currentSourceFile which tries to index the empty text using the old positions, which results in out of bounds errors on many tests. I'm investigating an alternate solution that leaves the empty string unmodified

jakebailey: But are the nodes we pass in actually from the real tree? If they're synthetic, the positions won't map to the text, I don't think. (Is this something that the original code did?)

johnfav03: Yes, this was something the original code did - I just discovered it after using the tracker in my code

internal/ls/organizeimports/organizeimports.go
iisaduan: I think the packages need to be rearranged. `OrganizeImports` shares code with FAR and the FAR stuff can't be used if FAR is in `ls`, leading to duplicated code here, and I don't think either (both) need to be their own package. 

My proposal: 
- Move all the comparers (anything that existed before this PR/is needed by ls-dependent packages) to `lsutil` (so that `changeTracker` and `autoimports` can continue to use them)
- Move everything else to `ls`

Moving the main function to `ls` also allows us to make its format match the other LS functions (to be a method on `ls`)

internal/ls/codeactions.go
```diff
@@ -96,6 +107,51 @@ func (l *LanguageService) ProvideCodeActions(ctx context.Context, params *lsprot
...
+	organizeimports.OrganizeImports(
+		ctx,
+		file,
+		changeTracker,
+		program,
+		l.UserPreferences(),
+		kind,
+	)
+
+	changes := changeTracker.GetChanges()
```
iisaduan:
```suggestion
	changes := l.OrganizeImports(
		ctx,
		file,
		program,
		kind,
	)
```
Change tracker should be created and `.GetChanges()` should be called within `OrganizeImports()`. Seems weird  that we could potentially pass a not-new changetracker into the function

internal/ls/organizeimports.go
```diff
@@ -20,10 +34,1031 @@ var (
...
+//  2. Coalescing imports from the same module
+//  3. Sorting imports
+func OrganizeImports(
+	ctx context.Context,
+	sourceFile *ast.SourceFile,
+	changeTracker *change.Tracker,
+	program *compiler.Program,
+	preferences *lsutil.UserPreferences,
+	kind lsproto.CodeActionKind,
+) {
```
iisaduan: Along with the other change,

```suggestion
func (l *LanguageService) OrganizeImports(
	ctx context.Context,
	sourceFile *ast.SourceFile,
	preferences *lsutil.UserPreferences,
	kind lsproto.CodeActionKind,
)  map[string][]*lsproto.TextEdit {
```

internal/ls/codeactions.go
```diff
@@ -96,6 +107,51 @@ func (l *LanguageService) ProvideCodeActions(ctx context.Context, params *lsprot
 	return lsproto.CommandOrCodeActionArrayOrNull{CommandOrCodeActionArray: &actions}, nil
 }
 
+// createOrganizeImportsAction creates the organize imports code action
+func (l *LanguageService) createOrganizeImportsAction(
+	ctx context.Context,
+	program *compiler.Program,
+	file *ast.SourceFile,
+	uri lsproto.DocumentUri,
```
iisaduan:
```suggestion
```

unused arg

internal/ls/change/trackerimpl.go
iisaduan: Organize imports uses the nodes that existed in the old tree since it never has to generate new nodes. This allows us to re-emit nodes while keeping comments (and other benefits)

I took a closer look, and I think the change is generally correct. While this specifically isn't necessary for organize imports to work correctly (we could do something else to skip name generation), it's probably the easiest way to fix it and should hold up for future codefixes that reuse nodes

iisaduan: Suggestion to 
- change the if statement to something like `!ast.IsSynthesized()` or `node.Pos != -1` to make it more clear that it's testing if the node is a real node or not
- add a comment with some explanation (perhaps routing back to this PR). I think there's a chance that some of the code in the latter half of the function might need to be updated because of this change, but we wouldn't know until we hit a specific case (a mix of synthesized and old nodes in `nodeIn`)

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,743 @@
...
+	}
+	return comparer(name1, name2)
+}
+
+func compareImportKind(s1 *ast.Statement, s2 *ast.Statement) int {
+	return cmp.Compare(getImportKindOrder(s1), getImportKindOrder(s2))
+}
+
+// getImportKindOrder returns the sort order for different import kinds:
+// 1. Side-effect imports
```
andrewbranch, line 248: Not important, but seems like these could be constants

internal/ls/change/trackerimpl.go
andrewbranch: Yeah, I think this is right. At least in Strada, the emitter is very smart about detecting per-node whether it can take a slice of the source file text or if it needs to be printed fresh.

jakebailey:
Has this been tested in the editor itself yet?

johnfav03:
> Has this been tested in the editor itself yet?

Yes - I think my change to `CodeActionProvider` got lost during one of the bug fixes, but I just committed it again and I've tested locally in the editor using the vscode codebase

andrewbranch (approved):

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+					usedImports = append(usedImports, importDecl)
+				}
+			}
+		}
+	}
+
+	return usedImports
+}
+
+func isDeclarationUsed(
```
iisaduan: This and `isSymbolReferencedInFile` could be methods on checker in `checker/services.go`. This way you can just have a function `checker.IsDeclarationUsed(sourcefile, identifier)` and not need to get/pass around so many extra args in these functions and `removeUnusedImports`

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	for _, stmt := range statements {
+		if stmt.Kind == ast.KindImportDeclaration {
+			result = append(result, stmt)
+		}
+	}
+	return result
+}
+
+// GroupByNewlineContiguous groups declarations by blank lines between them.
+func GroupByNewlineContiguous(sourceFile *ast.SourceFile, decls []*ast.Statement) [][]*ast.Statement {
```
iisaduan: move this and `isNewGroup` to `ls/organizeimports.go`

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	caseSensitiveOrganizeImportsComparer   = []func(a, b string) int{getOrganizeImportsOrdinalStringComparer(false)}
+	organizeImportsComparers               = []func(a, b string) int{
+		caseInsensitiveOrganizeImportsComparer[0],
+		caseSensitiveOrganizeImportsComparer[0],
+	}
+)
+
+// FilterImportDeclarations filters out non-import declarations from a list of statements.
+func FilterImportDeclarations(statements []*ast.Statement) []*ast.Statement {
+	var result []*ast.Statement
```
iisaduan: could this be done with `core.Filter`?

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// RangeIsOnSingleLine returns true if the given text range is on a single line.
+func RangeIsOnSingleLine(r core.TextRange, sourceFile *ast.SourceFile) bool {
```
iisaduan: This function already exists in `printer` (but is unexported). Can you export and use that one?

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	if r.Pos() < 0 || r.End() < 0 {
+		return true
+	}
+	startLine, _ := scanner.GetECMALineAndCharacterOfPosition(sourceFile, r.Pos())
+	endLine, _ := scanner.GetECMALineAndCharacterOfPosition(sourceFile, r.End())
+	return startLine == endLine
+}
+
+// GetImportAttributesKey returns a key for grouping imports by their attributes.
+func GetImportAttributesKey(attributes *ast.ImportAttributesNode) string {
```
iisaduan: move to `ls/organizeimports.go`

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+		importEquals := declaration.AsImportEqualsDeclaration()
+		if importEquals.ModuleReference.Kind == ast.KindExternalModuleReference {
+			return importEquals.ModuleReference.Expression()
+		}
+		return nil
+	case ast.KindImportDeclaration:
+		return declaration.ModuleSpecifier()
+	case ast.KindVariableStatement:
+		variableStatement := declaration.AsVariableStatement()
+		declarations := variableStatement.DeclarationList.AsVariableDeclarationList().Declarations.Nodes
```
iisaduan: An fyi--
```suggestion
		declarations := declaration.AsVariableStatement().DeclarationList.AsVariableDeclarationList().Declarations.Nodes
```
you can chain these, especially if you're not using the intermediate variables for anything else (checking for validity between "casts")

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+		if len(declarations) > 0 {
+			decl := declarations[0]
+			initializer := decl.Initializer()
```
iisaduan: here too 
```suggestion
			initializer :=  declarations[0].Initializer()
```

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+// GetExternalModuleName returns the module name from a module specifier expression.
+func GetExternalModuleName(specifier *ast.Expression) string {
+	if specifier != nil && ast.IsStringLiteralLike(specifier.AsNode()) {
+		return specifier.Text()
+	}
+	return ""
+}
+
+// CompareModuleSpecifiersWorker compares two module specifiers using the given comparer.
+func CompareModuleSpecifiersWorker(m1 *ast.Expression, m2 *ast.Expression, comparer func(a, b string) int) int {
```
iisaduan:
```suggestion
func CompareModuleSpecifiers(m1 *ast.Expression, m2 *ast.Expression, comparer func(a, b string) int) int {
```
we don't have a non-"worker" compareModuleSpecifiers

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	default: // OrganizeImportsTypeOrderLast
+		if cmp := core.CompareBooleans(s1.IsTypeOnly(), s2.IsTypeOnly()); cmp != 0 {
+			return cmp
+		}
+		return comparer(s1Name, s2Name)
+	}
+}
+
+// CompareExportSpecifiers compares two export specifiers considering type order.
+func CompareExportSpecifiers(s1 *ast.Node, s2 *ast.Node, comparer func(a, b string) int, typeOrder OrganizeImportsTypeOrder) int {
```
iisaduan:
```suggestion
```
Function is redundant (see comment in `ls/organizeimports.go`)

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	comparer organizeImportsComparerSettings,
+	sourceFile *ast.SourceFile,
+	changeTracker *change.Tracker,
+) {
+	if len(oldExportDecls) == 0 {
+		return
+	}
+
+	specifierComparerFunc := func(s1, s2 *ast.Node) int {
+		return lsutil.CompareExportSpecifiers(s1, s2, comparer.namedImportComparer, comparer.typeOrder)
```
iisaduan: can be replaced with `GetNamedImportSpecifierComparer` like on line 156, or even `compareImportOrExportSpecifiers` if you're certain `comparer.namedImportComparer` is non-nil, as you have here (I need to think more about if it will be or not, I'm not sure)

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	i := 0
+	for j := range len(arr) - 1 {
+		if comparer(arr[j], arr[j+1]) > 0 {
+			i++
+		}
+	}
+	return i
+}
+
+// GetNamedImportSpecifierComparerWithDetection returns a specifier comparer based on detecting the existing sort order.
```
iisaduan:
```suggestion
// GetNamedImportSpecifierComparerWithDetection returns a specifier comparer based on the existing sort order within a single import statement
```

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+		isSorted:            bestDiff[OrganizeImportsTypeOrderLast] == 0,
+	}
+}
+
+type caseSensitivityDetectionResult struct {
+	comparer func(a, b string) int
+	isSorted bool
+}
+
+// DetectModuleSpecifierCaseBySort detects the case sensitivity of module specifiers based on existing sort order.
```
iisaduan:
```suggestion
// DetectModuleSpecifierCaseBySort detects the order of import statements by their module specifiers based on import statements throughout the module/file
```

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	return organizeImportsComparers
+}
+
+type namedImportSortResult struct {
+	namedImportComparer func(a, b string) int
+	typeOrder           OrganizeImportsTypeOrder
+	isSorted            bool
+}
+
+// DetectNamedImportOrganizationBySort detects the organization of named imports by analyzing the existing sort order.
```
iisaduan:
```suggestion
// detects the order of named imports throughout the file by considering the named imports in each statement as a group
```
Not really your fault but the existing copilot comments are bothering me. The comments are rather meaningless since the function names are already exteremely verbose 

_Wait I didn't know this function was about organization of imports and it did "analyzing" :000000000 🙃_

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+}
+
+// GetImportDeclarationInsertIndex returns the index at which to insert a new import declaration.
+func GetImportDeclarationInsertIndex(sortedImports []*ast.Statement, newImport *ast.Statement, comparer func(a, b *ast.Statement) int) int {
+	return core.FirstResult(core.BinarySearchUniqueFunc(sortedImports, func(mid int, value *ast.Statement) int {
+		return comparer(value, newImport)
+	}))
+}
+
+// GetOrganizeImportsStringComparerWithDetection returns a comparer based on detection of the existing import order.
```
iisaduan:
```suggestion
// GetOrganizeImportsStringComparerWithDetection returns a string comparer based on detecting the order of import statements by module specifiers
```

jakebailey:
I tested this on `packages/retag/src/index.ts` in https://github.com/microsoft/DefinitelyTyped-tools.

Organizing started with:

```ts
#!/usr/bin/env node

import assert = require("assert");
import yargs from "yargs";
import process = require("process");

import { TypeScriptVersion } from "@definitelytyped/typescript-versions";
import {
  Logger,
  NpmPublishClient,
  consoleLogger,
  logUncaughtErrors,
  loggerWithErrors,
  LoggerWithErrors,
  cacheDir,
  nAtATime,
  retry,
} from "@definitelytyped/utils";
import { AnyPackage, TypingsData, AllPackages, getDefinitelyTyped } from "@definitelytyped/definitions-parser";
import * as pacote from "pacote";
import * as semver from "semver";

if (require.main === module) {
```

But then produced:

```ts
#!/usr/bin/env node

import assert = require("assert");
#!/usr/bin/env node
import yargs from "yargs";
import process = require("process");

#!/usr/bin/env node
import { AllPackages, AnyPackage, TypingsData, getDefinitelyTyped } from "@definitelytyped/definitions-parser";
#!/usr/bin/env node
import { TypeScriptVersion } from "@definitelytyped/typescript-versions";
#!/usr/bin/env node
import {
    Logger,
    LoggerWithErrors,
    NpmPublishClient,
    cacheDir,
    consoleLogger,
    logUncaughtErrors,
    loggerWithErrors,
    nAtATime,
    retry,
} from "@definitelytyped/utils";
#!/usr/bin/env node
import * as pacote from "pacote";
#!/usr/bin/env node
import * as semver from "semver";
```

jakebailey:
On `packages/dts-gen/src/definitely-typed.ts`, I organized and got:

```ts
import { existsSync, mkdirSync, writeFileSync } from "fs";
import { STATUS_CODES } from "http";
import { get } from "https";
import { homedir } from "os";
import { join as joinPaths } from "path";
import { parse as parseUrl } from "url";
import { getDTName } from "./names";
import parseGitConfig = require("parse-git-config");
```

Maybe this was the case before, but it's interesting that the require got pushed to the bottom.

In `packages/definitions-parser/src/git.ts`, I organized this:

```ts
import {
  Logger,
  assertDefined,
  cacheDir,
  consoleLogger,
  execAndThrowErrors,
  joinPaths,
  symmetricDifference,
} from "@definitelytyped/utils";
import * as pacote from "pacote";
import * as semver from "semver";
import { inspect } from "util";
import { PreparePackagesResult, getAffectedPackages } from "./get-affected-packages";
import { parseVersionFromDirectoryName } from "./lib/definition-parser";
import { AllPackages, NotNeededPackage, PackageId, formatTypingVersion, getDependencyFromFile } from "./packages";
import { readFile } from "fs/promises";
```

And got:

```ts
import {
    Logger,
    assertDefined,
    cacheDir,
    consoleLogger,
    execAndThrowErrors,
    joinPaths,
    symmetricDifference,
} from "@definitelytyped/utils";
import { readFile } from "fs/promises";
import * as pacote from "pacote";
import * as semver from "semver";
import { inspect } from "util";
import { PreparePackagesResult, getAffectedPackages } from "./get-affected-packages";
import { parseVersionFromDirectoryName } from "./lib/definition-parser";
import { AllPackages, NotNeededPackage, PackageId, formatTypingVersion, getDependencyFromFile } from "./packages";
```

So, it lost my indent level somewhere (this repo is 2 space).

This repo might be a good one to spot test. I think I was just lucky I happened to have a shebang file open on my first try!

iisaduan:
Some of the indenting logic is not yet ported. Not sure if it affects organize imports since it's theoretically all old nodes, and that's a formatting option that should've been picked up on, but it could be why it loses the smaller indent

johnfav03:
I've updated the code to account for the shebang issue, and I've also updated `textwriter.go` to use `ls.formatOptions.indentSize` instead of a hardcoded string of 4 spaces. However, there's still an issue where the LSP doesn't sync indentSize settings from the editor, which I mentioned to @iisaduan. As for the require being sorted below the others, this behavior was in Typescript as well, with import equals declarations being sorted in their own category.

```typescript
// 1. Side-effect imports
// 2. Type-only imports
// 3. Namespace imports
// 4. Default imports
// 5. Named imports
// 6. ImportEqualsDeclarations
// 7. Require variable statements
function getImportKindOrder(s1: AnyImportOrRequireStatement) {
    switch (s1.kind) {
        case SyntaxKind.ImportDeclaration:
            if (!s1.importClause) return 0;
            if (s1.importClause.isTypeOnly) return 1;
            if (s1.importClause.namedBindings?.kind === SyntaxKind.NamespaceImport) return 2;
            if (s1.importClause.name) return 3;
            return 4;
        case SyntaxKind.ImportEqualsDeclaration:
            return 5;
        case SyntaxKind.VariableStatement:
            return 6;
    }
}
```

jakebailey:
I think there's a bug in the way we load indent size out of the settings; it's some discrepency i noticed when working on my user prefs refactor. VS Code's setting has a different name than we do.

andrewbranch (approved):

internal/printer/printer.go
```diff
@@ -4429,7 +4432,7 @@ func (p *Printer) emitDirective(kind string, refs []*ast.FileReference) {
 
 func (p *Printer) emitList(emit func(p *Printer, node *ast.Node), parentNode *ast.Node, children *ast.NodeList, format ListFormat) {
 	if p.shouldEmitOnMultipleLines(parentNode) {
-		format |= LFPreferNewLine
+		format |= LFPreferNewLine | LFIndented
```
iisaduan, line 4435: Where does this change come from?

internal/printer/changetrackerwriter.go
```diff
@@ -19,9 +19,12 @@ type triviaPositionKey interface { // *astNode | *ast.NodeList
 	End() int
 }
 
-func NewChangeTrackerWriter(newline string) *ChangeTrackerWriter {
+func NewChangeTrackerWriter(newline string, indentSize int) *ChangeTrackerWriter {
+	if indentSize <= 0 {
```
iisaduan: I think `-1` is better for "use the default" since technically `indentsize = 0` is valid since it just means no indenting.

We should also comment a `TODO` here to actually pass in indent options in the cases where you've currently used `0`. I don't think it's a blocker for this PR, but there is indenting related formatting that isn't ported yet

internal/printer/textwriter.go
```diff
@@ -104,22 +105,25 @@ func (w *textWriter) updateLineCountAndPosFor(s string) {
 	w.lineStart = false
 }
 
-func getIndentString(indent int) string {
-	switch indent {
-	case 0:
+const defaultIndentSize = 4
+
+// GetDefaultIndentSize returns the default indent size (4 spaces) used when no specific indent size is configured.
+func GetDefaultIndentSize() int {
```
iisaduan, line 111: Can you update the relevant places in `formatCodeSettings` to also use this default?

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+
+// OrganizeImports organizes imports by:
+//  1. Removing unused imports
+//  2. Coalescing imports from the same module
+//  3. Sorting imports
+func (l *LanguageService) OrganizeImports(
+	ctx context.Context,
+	sourceFile *ast.SourceFile,
+	program *compiler.Program,
+	preferences *lsutil.UserPreferences,
```
iisaduan: preferences should be used with `l.UserPreferences()`
```suggestion
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	comparersToTest, typeOrdersToTest := lsutil.GetDetectionLists(preferences)
+	defaultComparer := comparersToTest[0]
+
+	moduleSpecifierComparer := defaultComparer
+	namedImportComparer := defaultComparer
+	typeOrder := lsutil.OrganizeImportsTypeOrderAuto
+	if preferences != nil {
+		typeOrder = preferences.OrganizeImportsTypeOrder
+	}
+
```
iisaduan: preferences when gotten from `ls.UserPreferences()` should never be nil
```suggestion
	typeOrder := preferences.OrganizeImportsTypeOrder
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	if preferences == nil || preferences.OrganizeImportsIgnoreCase.IsUnknown() {
```
iisaduan:
```suggestion
	if preferences.OrganizeImportsIgnoreCase.IsUnknown() {
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+		result, _ := lsutil.DetectModuleSpecifierCaseBySort(topLevelImportGroupDecls, comparersToTest)
+		moduleSpecifierComparer = result
+	}
+
+	if typeOrder == lsutil.OrganizeImportsTypeOrderAuto || (preferences != nil && preferences.OrganizeImportsIgnoreCase.IsUnknown()) {
```
iisaduan:
```suggestion
	if typeOrder == lsutil.OrganizeImportsTypeOrderAuto || preferences.OrganizeImportsIgnoreCase.IsUnknown() {
```

internal/ls/lsutil/organizeimports.go
```diff
@@ -0,0 +1,754 @@
...
+	typeOrder           OrganizeImportsTypeOrder
+	isSorted            bool
+}
+
+// DetectNamedImportOrganizationBySort detects the organization of named imports by analyzing the existing sort order.
+func DetectNamedImportOrganizationBySort(
+	originalGroups []*ast.Statement,
+	comparersToTest []func(a, b string) int,
+	typesToTest []OrganizeImportsTypeOrder,
+) (comparer func(a, b string) int, typeOrder OrganizeImportsTypeOrder, isSorted bool, found bool) {
```
iisaduan:
```suggestion
) (comparer func(a, b string) int, typeOrder OrganizeImportsTypeOrder, found bool) {
```

Is this arg ever used?

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+		namedImportComparer2, typeOrder2, _, found := lsutil.DetectNamedImportOrganizationBySort(topLevelImportDecls, comparersToTest, typeOrdersToTest)
+		if found {
+			if namedImportComparer == nil || (preferences != nil && preferences.OrganizeImportsIgnoreCase.IsUnknown()) {
```
iisaduan:
```suggestion
			if namedImportComparer == nil || preferences.OrganizeImportsIgnoreCase.IsUnknown() {
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	shouldSort := kind == lsproto.CodeActionKindSourceOrganizeImportsModeSortAndCombine || kind == lsproto.CodeActionKindSourceOrganizeImports
+	shouldCombine := shouldSort
+	shouldRemove := kind == lsproto.CodeActionKindSourceOrganizeImportsModeRemoveUnused || kind == lsproto.CodeActionKindSourceOrganizeImports
+	topLevelImportDecls := lsutil.FilterImportDeclarations(sourceFile.Statements.Nodes)
+	topLevelImportGroupDecls := lsutil.GroupByNewlineContiguous(sourceFile, topLevelImportDecls)
+
+	comparersToTest, typeOrdersToTest := lsutil.GetDetectionLists(preferences)
+	defaultComparer := comparersToTest[0]
+
+	moduleSpecifierComparer := defaultComparer
```
iisaduan: Is the nil handling done somewhere else? I think currently, these defaults differ from strada
```suggestion
	var moduleSpecifierComparer []func(a, b string) int
	var namedImportComparer []func(a, b string) int
    if !preferences.OrganizeImportsIgnoreCase.IsUnknown() {
	    moduleSpecifierComparer = defaultComparer
	    namedImportComparer = defaultComparer
    }
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+		newImportDecls = processedImports
+	}
+
+	if shouldSort && !shouldCombine {
+		slices.SortFunc(newImportDecls, func(a, b *ast.Statement) int {
+			return lsutil.CompareImportsOrRequireStatements(a, b, comparer.moduleSpecifierComparer)
+		})
+	}
+
+	if len(oldImportDecls) > 0 {
```
iisaduan: this outer `if` shouldn't be needed since we already checked it at the beginning of the function
```suggestion
```

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	result := make([][]*ast.Statement, 0, len(order))
+	for _, key := range order {
+		result = append(result, groups[key])
+	}
+	return result
+}
+
+func removeUnusedImports(oldImports []*ast.Statement, sourceFile *ast.SourceFile, program *compiler.Program, ctx context.Context) []*ast.Statement {
+	typeChecker, done := program.GetTypeCheckerForFile(ctx, sourceFile)
+	defer done()
```
iisaduan:
```suggestion
func removeUnusedImports(oldImports []*ast.Statement, sourceFile *ast.SourceFile, typeChecker *checker) []*ast.Statement {
```
Since `organizeImports` is only called on statements within the same file, its better to share the checker between each call to `removeUnusedImports` instead of making a new one every time

internal/ls/organizeimports.go
```diff
@@ -0,0 +1,910 @@
...
+	changeTracker *change.Tracker,
+	ctx context.Context,
+) {
+	if len(oldImportDecls) == 0 {
+		return
+	}
+
+	// Header comment preservation is handled via LeadingTriviaOptionExclude in the change tracker below
+
+	processedImports := oldImportDecls
```
iisaduan: this just copies the pointer, so it's possible for this code to mutuate `oldImportDecls` before it's referenced again at the end of the function

iisaduan (changes requested):
part 1 of my review-- most are proofreading things, but I started finding some bugs. I am doing some more testing, but am posting the first part so you can get a start on these in the meantime

internal/printer/printer.go
johnfav03: I was running into a bug where the first element of each list was printing on a newline but was unindented, whereas the rest of the list was properly indented.

internal/lsp/lsproto/lsp.go
```diff
@@ -231,3 +231,8 @@ func PreferredMarkupKind(formats []MarkupKind) MarkupKind {
 	}
 	return MarkupKindPlainText
 }
+
+const (
+	CodeActionKindSourceOrganizeImportsModeRemoveUnused   CodeActionKind = "source.organizeImports.removeUnused"
+	CodeActionKindSourceOrganizeImportsModeSortAndCombine CodeActionKind = "source.organizeImports.sortAndCombine"
+)
```
iisaduan, line 239: ^delete if we dont plan to implemnt this user side?

internal/lsp/server.go
```diff
@@ -918,6 +918,9 @@ func (s *Server) handleInitialize(ctx context.Context, params *lsproto.Initializ
 				CodeActionOptions: &lsproto.CodeActionOptions{
 					CodeActionKinds: &[]lsproto.CodeActionKind{
 						lsproto.CodeActionKindQuickFix,
+						lsproto.CodeActionKindSourceOrganizeImports,
+						lsproto.CodeActionKindSourceOrganizeImportsModeRemoveUnused,
+						lsproto.CodeActionKindSourceOrganizeImportsModeSortAndCombine,
```
iisaduan:
```suggestion
						lsproto.CodeActionKindSourceOrganizeImports,
```
^delete if we don't plan to implement this user side?

iisaduan (commented):
~Do we want to continue supporting these? `SortAndCombine` and `RemoveUnused` aren't (currently) triggerable in the ts-go extension.
We also had a separate (not yet ported) codefix to "remove unused imports" which would cover the `RemoveUnused` case (the more popular sub command).
@andrewbranch @mjbvz~

Discussed offline--We will continue to have the `SortAndCombine` and `RemoveUnused` actions. The change requested is to register them so they are user-invokable

internal/fourslash/tests/organizeImportsShebang_test.go
iisaduan, line 1: Move this up a level to `fourslash/tests` since `manual` are the tests that existed in the submodule and are manually changed (as opposed to just generated)

internal/ls/codeactions.go
```diff
@@ -105,13 +114,52 @@ func (l *LanguageService) ProvideCodeActions(ctx context.Context, params *lsprot
...
+	case lsproto.CodeActionKindSourceOrganizeImportsModeSortAndCombine:
+		return "Sort and Combine Imports"
+	default:
+		return "Organize Imports"
+	}
+}
+
+// getOrganizeImportsActionsForKind returns the organize imports code action kinds that should be
+// returned for the given requested kind and trigger.
+// When triggered automatically (e.g., on save), only the base organizeImports action is returned.
```
iisaduan: the user should be able to configure this each within the editor, so this should allow "automatic" triggers of the other options as well.

I think all of these should always be available in `ProvideCodeActions` and leave it to the editor to request/user to pick the correct one?

strada:
Image

iisaduan (changes requested):

iisaduan (approved):
:D

bradleyayers:
@johnfav03 thanks for this work, I've found an issue with comments on exports being duplicated, but I'm not sure where I should report it so I'll mention here. 

Here are a few example test cases that show the issue:

**Input:**

```
// a
export { a } from "a";
```

**Output (bad):**
```
// a
// a
export { a } from "a";
```

---

**Input:**

```
// a
// a
export { a } from "a";
```

**Output (bad):**
```
// a
// a
// a
// a
export { a } from "a";
```

---

**Input:**

```
// a

export { a } from "a";
```

**Output (bad):**
```
// a

// a
export { a } from "a";
```

johnfav03:
@bradleyayers Thank you for letting me know! I've put up a PR to fix this: https://github.com/microsoft/typescript-go/pull/3162

Issue #2326, opened 2025-12-10
JSX emit does not preserve string delimiters


dragomirtitian:
## Steps to reproduce

```ts
let v = <div id='value' />
```

## Behavior with `typescript@5.9`

Emitted JS uses the same delimiter for the string 'value' as in original source code:

```js
let v = React.createElement("div", { id: 'value' });
```

## Behavior with `tsgo`

Emitted JS always uses `"` as the delimiter:

```js
let v = React.createElement("div", { id: "value" });
```

jakebailey:
Does this actually matter for something you're doing, or did you just notice it in a comparison?

(I think we have a bit less fidelity for this in some cases but likely should still be able to do this given we need to have this preservation work for the editor or something.)

dragomirtitian:
Does not really matter. Was on the fence about reporting it, but I thought it might be good to document it. Feel free to close it. 

The only small issue is it makes it more difficult for us to spot any significant JS differences since we do text comparison on output.

PR #2333, opened 2025-12-11
Add token flags parameter to token creation functions

gabritto:
Also propagates token flags in the places where Strada did.
This could have helped prevent #2326, probably.

### DIFF
--- a/internal/transformers/moduletransforms/utilities.go
+++ b/internal/transformers/moduletransforms/utilities.go
@@ -57,7 +57,7 @@ func getExternalModuleNameLiteral(factory *printer.NodeFactory, importNode *ast.
 		if name == nil {
 			name = tryRenameExternalModule(factory, moduleName, sourceFile)
 		}
-		if name == nil {
+		if name == nil { // !!! propagate token flags (will produce new diffs)
 			name = factory.NewStringLiteral(moduleName.Text(), ast.TokenFlagsNone)
 		}
 		return name
--- a/internal/transformers/tstransforms/runtimesyntax.go
+++ b/internal/transformers/tstransforms/runtimesyntax.go
@@ -190,7 +190,7 @@ func (tx *RuntimeSyntaxTransformer) getExpressionForPropertyName(member *ast.Enu
 		return tx.Visitor().VisitNode(n.Expression)
 	case ast.KindIdentifier:
 		return tx.Factory().NewStringLiteralFromNode(name)
-	case ast.KindStringLiteral:
+	case ast.KindStringLiteral: // !!! propagate token flags (will produce new diffs)
 		return tx.Factory().NewStringLiteral(name.Text(), ast.TokenFlagsNone)
 	case ast.KindNumericLiteral:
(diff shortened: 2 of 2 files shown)


testdata/baselines/reference/submodule/conformance/assignmentCompatWithObjectMembersStringNumericNames.symbols.diff
```diff
@@ -1,49 +1,20 @@
...
-->'1' : Symbol(S2['1'], Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 9, 18))
-+>'1' : Symbol(S2["1"], Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 9, 18))
- >bar : Symbol(S2.bar, Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 9, 31))
+@@= skipped -17, +17 lines =@@
 
      interface T2 { '1.0': string; baz?: string }
  >T2 : Symbol(T2, Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 9, 46))
 ->'1.0' : Symbol(T2['1.0'], Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 10, 18))
-+>'1.0' : Symbol(T2["10"], Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 10, 18))
++>'1.0' : Symbol(T2['10'], Decl(assignmentCompatWithObjectMembersStringNumericNames.ts, 10, 18))
```
jakebailey: Nothing you did in this PR, but I bet this is a very simple oops somewhere

gabritto: Looking

internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -191,9 +191,9 @@ func (tx *RuntimeSyntaxTransformer) getExpressionForPropertyName(member *ast.Enu
 	case ast.KindIdentifier:
 		return tx.Factory().NewStringLiteralFromNode(name)
 	case ast.KindStringLiteral:
-		return tx.Factory().NewStringLiteral(name.Text())
```
gabritto, line 194: We could preserve flags here, but we didn't in Strada (I think because `cloneNode` doesn't copy the `singleQuote` property, but haven't looked deeply), so if we propagate flags here it ends up producing a couple diffs.

internal/transformers/moduletransforms/utilities.go
```diff
@@ -59,7 +58,7 @@ func getExternalModuleNameLiteral(factory *printer.NodeFactory, importNode *ast.
 			name = tryRenameExternalModule(factory, moduleName, sourceFile)
 		}
 		if name == nil {
-			name = factory.NewStringLiteral(moduleName.Text())
```
gabritto, line 62: Same as other comment above, we could preserve flags here but since Strada didn't, we end up with 190 new diffs if we try and preserve them now.

jakebailey (approved):
Worth leaving comments about those cases?

PR #2338, opened 2025-12-11
Fix unquote

gabritto:
Fixes misported regex. Noticed in https://github.com/microsoft/typescript-go/pull/2333#discussion_r2608791712.


jakebailey (approved):
Why is this even a regex? 😄

PR #2343, opened 2025-12-11
--experimentalDecorators and --emitDecoratorMetadata

### LINKED ISSUE #914: Not transpiling decorators
It seems to emit decorators as-is in both experimental and regular mode.

weswigham:
This PR ports emit for `--experimentalDecorators` and `--emitDecoratorMetadata`, since many users of the former also enable the later (which isn't even supported with standardized decorators in strada, apparently).

This looks pretty solid from the baselines at this point, I've only noticed two things I still want to follow up on, or at least get someone else's opinion on:
* The single sourcemap test'd diffs are still all outta line - there's just way more mappings. Something is probably missing a `EFNoSourceMaps` that should have it. Or it could just be class fields not being transformed yet. I'm unsure. (See [sourceMapValidationDecorators.sourcemap.txt](https://github.com/microsoft/typescript-go/compare/main...weswigham:decos?expand=1#diff-35872429e3befd00a2f5f5f7160f3a76e5650591059bab94b0fe1a9d8a8df19b)) Somewhere there's a nicer tool for viewing sourcemap diffs than our textual baselines, I just need to find it again...
* There's a `target: es5` test of a constructor reference to a decorated class within a static of that class - binding behavior here intentionally changes after es6, and the test is effectively targeting `es6` now (thanks to the new minimum target), so it's probably fine, but I'm not wholly convinced.

Additionally, I haven't included the conditional lookup of the `BigInt` global in emit for targets < es2020 at present. I could add it back, I'm just not completely sure why it was needed in the first place, and with `Symbol` definitely no longer being maybe-conditional with min target es6, it was the only use of the conditional global type reference emit logic.

Most of the remaining diffs in the touched baselines that weren't outright deleted are just other unported emit features remaining in the diffs - class fields, mostly, and a little `await`.

Fixes #914

### DIFF
--- a/internal/printer/factory.go
+++ b/internal/printer/factory.go
@@ -554,7 +554,7 @@ func (f *NodeFactory) NewMetadataHelper(metadataKey string, metadataValue *ast.N
 		nil, /*questionDotToken*/
 		nil, /*typeArguments*/
 		f.NewNodeList([]*ast.Node{
-			f.NewStringLiteral(metadataKey),
+			f.NewStringLiteral(metadataKey, ast.TokenFlagsNone),
 			metadataValue,
 		}),
 		ast.NodeFlagsNone,
@@ -567,7 +567,7 @@ func (f *NodeFactory) NewParamHelper(expression *ast.Node, parameterOffset int, 
 		f.NewUnscopedHelperName("__param"),
 		nil, /*questionDotToken*/
 		nil, /*typeArguments*/
-		f.NewNodeList([]*ast.Expression{f.NewNumericLiteral(strconv.Itoa(parameterOffset)), expression}),
+		f.NewNodeList([]*ast.Expression{f.NewNumericLiteral(strconv.Itoa(parameterOffset), ast.TokenFlagsNone), expression}),
 		ast.NodeFlagsNone,
 	)
 	helper.Loc = location
--- a/internal/transformers/tstransforms/legacydecorators.go
+++ b/internal/transformers/tstransforms/legacydecorators.go
@@ -1029,7 +1029,7 @@ func (tx *LegacyDecoratorsTransformer) getExpressionForPropertyName(member *ast.
 		}
 		return name.AsComputedPropertyName().Expression
 	} else if ast.IsIdentifier(name) {
-		return tx.Factory().NewStringLiteral(name.Text())
+		return tx.Factory().NewStringLiteral(name.Text(), ast.TokenFlagsNone)
 	} else {
 		return tx.Factory().DeepCloneNode(name)
 	}
--- a/internal/transformers/tstransforms/typeserializer.go
+++ b/internal/transformers/tstransforms/typeserializer.go
@@ -451,7 +451,7 @@ func (s *metadataSerializer) serializeEntityNameAsExpressionFallback(node *ast.E
  */
 func (s *metadataSerializer) createCheckedValue(left *ast.Node, right *ast.Node) *ast.Node {
 	return s.f.NewLogicalANDExpression(
-		s.f.NewStrictInequalityExpression(s.f.NewTypeOfExpression(left), s.f.NewStringLiteral("undefined")),
+		s.f.NewStrictInequalityExpression(s.f.NewTypeOfExpression(left), s.f.NewStringLiteral("undefined", ast.TokenFlagsNone)),
 		right,
 	)
 }


internal/printer/emitcontext.go
```diff
@@ -446,6 +446,14 @@ func (c *EmitContext) SetOriginal(node *ast.Node, original *ast.Node) {
 	c.SetOriginalEx(node, original, false)
 }
 
+func (c *EmitContext) UnsetOriginal(node *ast.Node) {
+	if c.original == nil {
+		c.original = make(map[*ast.Node]*ast.Node)
+	}
+
+	c.original[node] = nil
```
jakebailey: `delete(c.original, node)` would work regardless of nilness, too

testdata/baselines/reference/submodule/conformance/esDecorators-classDeclaration-parameterDecorators(target=es2015).js.diff
```diff
@@ -0,0 +1,49 @@
...
++    x) { }
+ }
+ (class C {
+-    constructor(x) { }
+-    method(x) { }
+-    set x(x) { }
+-    static method(x) { }
+-    static set x(x) { }
++    constructor(
++    @dec
```
jakebailey, line 35: This is a new diff; what's going on here?

testdata/baselines/reference/submodule/conformance/usingDeclarationsWithLegacyClassDecorators.11(module=esnext,target=es5).js.diff
```diff
@@ -4,17 +4,14 @@
...
--C = __decorate([
--    dec
--], C);
--export { C };
- var after;
-+@dec
-+class C {
-+}
-+export { C };
++var after;
```
jakebailey, line 7: There are a few tests like this where `after` is misplaced; I don't know how much it matters, though, given it's a var?

(Oh, hm, maybe this is just "it's es5 so who cares")

jakebailey (commented):
I read every diff and it all looks great! Awesome!

internal/printer/emitcontext.go
jakebailey: (Unless explicitly set `nil` is required)

jakebailey (approved):
Everything I mentioned above are small nits, I think, so I would be happy with this as is!

jakebailey:
The only other thing is to remove `!!! emitDecoratorMetadata` from `program.go` where we would have emitted an error on this option.

tmm1:
Thank you for tackling this 🙏

tmm1:
> * Somewhere there's a nicer tool for viewing sourcemap diffs than our textual baselines, I just need to find it again...

Perhaps you're thinking of https://evanw.github.io/source-map-visualization

testdata/baselines/reference/submodule/conformance/esDecorators-classDeclaration-parameterDecorators(target=es2015).js.diff
weswigham: Decorators without `--experimentalDecorators` don't have a transform (yet), so aren't being transformed in the output. The TS transform formerly was eliding all parameter modifier/decorators until I fixed that here, which is why it's now preserved in the output.

testdata/baselines/reference/submodule/conformance/usingDeclarationsWithLegacyClassDecorators.11(module=esnext,target=es5).js.diff
weswigham: Functionally identical for sure, I looked at this one and eventually just went with "eh, probably doesn't matter". I actually don't get the old behavior - hoisted variables usually get injected at the top of the scope, so I'm not sure why the old behavior had it near the bottom near the actual assignment.

weswigham:
> The only other thing is to remove !!! emitDecoratorMetadata from program.go where we would have emitted an error on this option.

AFAIK that's already gone, replaced with the error requiring `emitDecoratorMetadata` be specified alongside `experimentalDecorators`.

jycouet:
Great news 🎉

mlarabi:
I am so happy to see this PR. I now know what to experiment this night. Thank you @weswigham and everyone involved for the work done.

PR #2357, opened 2025-12-12
List crashing tests

### LINKED ISSUE #2350: Differentiate between crashing vs failing tests
Currently the test runner allows failing requests - but in actuality, these are recovered crashes. We should find a way to track these differently so that we can burn these issues down.

gabritto:
Fixes #2350.

Not sure what the best way to surface this information is, but here's the list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -276,22 +276,7 @@ jobs:
           npx hereby test || true
 
       - run: git add .
-
-      - name: Check for changes
-        id: check-changes
-        run: |
-          if git diff --staged --exit-code --quiet; then
-            git diff --staged --exit-code --stat
-            git diff --staged > generate_changes.patch
-            exit 1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -276,7 +276,22 @@ jobs:
...
+      
+      - name: Check for changes
+        id: check-changes
+        run: |
+          if ! git diff --staged --exit-code --quiet; then
+            git diff --staged --exit-code --stat
+            git diff --staged > generate_changes.patch
+            exit 1
+          fi
+
```
DanielRosenwasser:
```suggestion

      - name: Check for changes
        id: check-changes
        run: |
          if ! git diff --staged --exit-code --quiet; then
            git diff --staged --exit-code --stat
            git diff --staged > generate_changes.patch
            exit 1
          fi

```

PR #2358, opened 2025-12-12
fix(2356): omit numeric separators when targeting es2020

### LINKED ISSUE #2356: Omit numeric separators when targeting es2020
Currently, regardless of target it appears that

```ts
let x = 1_2_3 + 123;
```

will be emitted as-is.

Numeric separators were introduced in ES2021, so we will need to modify either our es2021 transformer or the printer to respect `--target es2020`.

a-tarasyuk:
Fixes #2356

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -206,9 +206,7 @@ func (p *Printer) getLiteralTextOfNode(node *ast.LiteralLikeNode, sourceFile *as
 			}
 		}
 	}
-
 	// !!! Printer option to control whether to terminate unterminated literals
-	// !!! If necessary, printer option to control whether to preserve numeric separators
 	if p.emitContext.EmitFlags(node)&EFNoAsciiEscaping != 0 {
 		flags |= getLiteralTextFlagsNeverAsciiEscape
 	}


jakebailey (approved):
Took me a bit to figure out what the Strada code did but, yeah:

```ts
        const flags = (neverAsciiEscape ? GetLiteralTextFlags.NeverAsciiEscape : 0)
            | (jsxAttributeEscape ? GetLiteralTextFlags.JsxAttributeEscape : 0)
            | (printerOptions.terminateUnterminatedLiterals ? GetLiteralTextFlags.TerminateUnterminatedLiterals : 0)
            | (printerOptions.target && printerOptions.target >= ScriptTarget.ES2021 ? GetLiteralTextFlags.AllowNumericSeparator : 0);

        return getLiteralText(node, sourceFile, flags);
```

Unbelievably, this is the only reason we pass the target into the printer.

(I do bet we have other things that are missing, though.)

jakebailey (changes requested):
Er, actually, no, `getLiteralTextOfNode` is where this should live, which is where I pulled the above code from. There is a TODO there for this (and related code).

internal/printer/printer.go
jakebailey, line 211: This comment can go away

a-tarasyuk: @jakebailey removed

internal/printer/printer.go
```diff
@@ -206,9 +206,6 @@ func (p *Printer) getLiteralTextOfNode(node *ast.LiteralLikeNode, sourceFile *as
 			}
 		}
 	}
-
-	// !!! Printer option to control whether to terminate unterminated literals
```
jakebailey, line 210: You removed this, but is this actually done?

a-tarasyuk: @jakebailey Yep, I've reverted this comment

PR #2359, opened 2025-12-12
Permit URLs in `@see` tags and make links for name references

### LINKED ISSUE #2306: The `@see` JSDoc tag cannot create links from URLs
## Steps to reproduce

1. Open VSCode with the [TypeScript Native Preview](https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.native-preview) extension.
2. Paste in this code example
   ```ts
   /**
    * @see https://example.com
    */
   const test = 0;
   ```
3. Hover over the `test` variable
4. The link does not work and is rendered incorrectly

Perhaps it's some new parsing of `@see` tags for JSDoc with the Language server, but it's very different from TypeScript 5.9. This seems to be related to how [`@see <namepath>`](https://jsdoc.app/tags-see) is parsed.

```ts
const foo = () => { /* .. */ };

/**
 * @see foo
 * @see foo description
 * @see {@link foo}
 * @see {@link foo} description
 */
const bar = () => { /* .. */ };
```

| `typescript@5.9` | `tsgo` |
| --- | --- |
| Image | Image |

Markdown generated is `_@see_ &mdash [https://example.com](https://example.com)`

Image

## Behavior with `tsgo`

Markdown generated is similar to ``_@see_ `https` &mdash; ://example.com``

Image

## Extra notes

It's not a regression compared to TS 5.9, but it's a bit weird that with this code example:

```ts
const foo = () => { /* .. */ };

/**
 * @see foo
 * @see foo description
 */
const bar = () => { /* .. */ };
```

The following is true:

- The `foo` name can use "Go to definition" for `@see foo [...]`
- The markdown generated for both TSGO and TS 5.9 is not a clickable link
- Must be converted to `{@link foo}` for `foo` to be navigable for both Markdown and "Go to definition"

Given the description on https://jsdoc.app/tags-see (below), I would expect `@see <namepath>` to work the same as `@see {@link <namepath>}`.

> You can provide either **a symbol's namepath** or free-form text. If you provide a namepath, JSDoc's default template **automatically converts the namepath to a link**.

ahejlsberg:
Fixes #2306.

### DIFF
--- a/internal/parser/jsdoc.go
+++ b/internal/parser/jsdoc.go
@@ -787,7 +787,7 @@ func (p *Parser) parseTypeTag(previousTags []*ast.Node, start int, tagName *ast.
 }
 
 func (p *Parser) parseSeeTag(start int, tagName *ast.IdentifierNode, indent int, indentText string) *ast.Node {
-	hasNameReference := p.isIdentifier() && !p.scanner.TextAtTokenIs("http://") && !p.scanner.TextAtTokenIs("https://") ||
+	hasNameReference := p.isIdentifier() && !strings.HasPrefix(p.sourceText[p.scanner.TokenEnd():], "://") ||
 		p.token == ast.KindOpenBraceToken && p.lookAhead((*Parser).nextTokenIsIdentifierOrKeyword)
 	var nameExpression *ast.Node
 	if hasNameReference {
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -277,10 +277,6 @@ func (s *Scanner) TokenText() string {
 	return s.text[s.tokenStart:s.pos]
 }
 
-func (s *Scanner) TextAtTokenIs(str string) bool {
-	return s.text[s.tokenStart:min(s.tokenStart+len(str), len(s.text))] == str
-}
-
 func (s *Scanner) TokenValue() string {
 	return s.tokenValue
 }


internal/parser/jsdoc.go
```diff
@@ -794,11 +787,10 @@ func (p *Parser) parseTypeTag(previousTags []*ast.Node, start int, tagName *ast.
 }
 
 func (p *Parser) parseSeeTag(start int, tagName *ast.IdentifierNode, indent int, indentText string) *ast.Node {
-	isMarkdownOrJSDocLink := p.token == ast.KindOpenBracketToken || p.lookAhead(func(p *Parser) bool {
-		return p.nextTokenJSDoc() == ast.KindAtToken && tokenIsIdentifierOrKeyword(p.nextTokenJSDoc()) && isJSDocLinkTag(p.scanner.TokenValue())
-	})
+	hasNameReference := p.isIdentifier() && !p.scanner.TextAtTokenIs("http://") && !p.scanner.TextAtTokenIs("https://") ||
+		p.token == ast.KindOpenBraceToken && p.lookAhead((*Parser).nextTokenIsIdentifierOrKeyword)
```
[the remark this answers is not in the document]
ahejlsberg: Huh? Those parentheses wouldn't make a difference.

internal/parser/jsdoc.go
DanielRosenwasser: I just wouldn't gate this on http(s), and I wouldn't add a new method to `Scanner`

```suggestion
	var hasNameReference bool
	if p.isIdentifier() {
		// Specifically *not* a link (e.g. `@see x.y`)
		pos := p.scanner.TokenEnd()
		hasNameReference = !strings.HasPrefix(p.sourceText[pos:], "://")
	} else if p.token == ast.KindOpenBraceToken {
		// @see {@link ...}
		hasNameReference = p.lookAhead((*Parser).nextTokenIsIdentifierOrKeyword)
	}
```

ahejlsberg: Yup, will do.

PR #2361, opened 2025-12-12
Add package deduplication, --deduplicatePackages

### LINKED ISSUE #1080: Deduplicate doubly-installed packages
We have some package (in our case `@sentry/browser`) installed two places in our tree (in our case, two different npm workspaces which npm in its infinite wisdom did not figure out how to deduplicate, although I think this can come about from dependencies-of-dependencies as well). This package exports a class with a protected method. Assigning an instance of one to the other works in TS, but fails in Go, because I guess it didn't deduplicate the package like TS does. (There are also other cases where deduplicating affects behavior, mostly in very complex types, and probably many more where it affects performance.)

Let me know if more details are helpful to repro, it's a bit more ceremony to construct than most repros and it sounded like @jakebailey already knew enough to be aware of the problem, but I can set something up if it's more of an edge case and you need one.

jakebailey:
Fixes #1080

The old compiler did "package deduplication" through source file redirection. This allowed two duplicate packages in different parts of `node_modules` (but with the same name / package.json version) to be "compatible" by way of them being the same ASTs even though they are at different paths (and so at runtime, probably are _not_ compatible).

The old compiler did this by essentially proxying the original `SourceFile` via `Object.create(sourceFile)` (the original file is the prototype 😨), then overriding `symbol`, `fileName`, etc, on that new instance.

The old compiler also chose the "winning"/canonical packages as they were walked, saving work and producing a deterministic result. The new compiler's loading is fully concurrent, so that is a challenge.

We can't really do the same thing as we did before. It is technically feasible to move data off `SourceFile` and embed a pointer to a data sidecar, which can be shared safely, but the problem is that we want the `symbol` to also be shared, and that's not movable due to the way nodes are laid out.

Instead, this PR lets loading proceed as "normal", no deduplication. Then, during `collectFiles`, it deduplicates packages. Deduplication is done by quite literally replacing duplicates with the canonical files' ASTs. This means looking them up later will give the "right" source file, with the "right" nodes and symbols, as before.

This _seems_ to be sufficient; it's basically just what we did before, but less optimal, and without creating a completely new SourceFile node with different. It just means that if you ever query `filesByPath`, you might get a file with a different path than you expect. But, only in deps.

I've also made a flag that controls this behavior, `--deduplicatePackages`, defaulting to `true`.

---

I tried also reintroducing the concept of "redirected files" by moving data off of `SourceFile` into an embedded pointer, but that hit some walls where things didn't work. I'm not sure if it's worth doing if this method seems to work fine?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,6 +286,12 @@ jobs:
   smoke:
     runs-on: ubuntu-latest
     steps:
+      - uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+        with:
+          large-packages: false
+          docker-images: false
+          swap-storage: false
+          dotnet: false
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           submodules: true


testdata/baselines/reference/submodule/fourslash/goToDefinition/duplicatePackageServices.baseline.jsonc.diff
```diff
@@ -0,0 +1,11 @@
+--- old.duplicatePackageServices.baseline.jsonc
++++ new.duplicatePackageServices.baseline.jsonc
+@@= skipped -10, +10 lines =@@
+
+
+ // === goToDefinition ===
+-// === /node_modules/a/node_modules/x/index.d.ts ===
++// === /node_modules/b/node_modules/x/index.d.ts ===
```
jakebailey: This diff exists because go-to-def is not aware of the deduping, instead opting to go to the "actual" file rather than the deduped winner.

testdata/baselines/reference/submodule/compiler/declarationEmitForGlobalishSpecifierSymlink.types.diff
```diff
@@ -0,0 +1,9 @@
+--- old.declarationEmitForGlobalishSpecifierSymlink.types
++++ new.declarationEmitForGlobalishSpecifierSymlink.types
+@@= skipped -39, +39 lines =@@
+
+ === /p2/index.d.ts ===
+ export const a: import("typescript-fsa").A;
+->a : import("/p2/node_modules/typescript-fsa/index").A
++>a : import("/p1/node_modules/typescript-fsa/index").A
```
jakebailey, line 8: This is technically fixed by #1604 in that that PR would instead say `import("typescript-fsa")` instead.

internal/compiler/filesparser.go
```diff
@@ -331,6 +342,36 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
...
+						packageIdToCanonicalPath[pkgId] = resolvedPath
+						sourceFileToPackageName[resolvedPath] = packageName
+						deduplicatedPathMap[resolvedPath] = resolvedPath
+					}
+				}
+			}
+
 			if subTasks := task.subTasks; len(subTasks) > 0 {
 				collectFiles(subTasks, seen)
 			}
```
jakebailey: The fact that we aren't skipping collecting deps out of deduped files I think is a problem, since we can accidentally load files we don't mean into the global scope from their deps; going to make a test and see if this is fixable

jakebailey: Now fixed (with a test that shows it's working in the diff)

internal/compiler/filesparser.go
```diff
@@ -331,8 +342,49 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
 			for _, trace := range task.resolutionsTrace {
 				loader.opts.Host.Trace(trace.Message, trace.Args...)
 			}
-			if subTasks := task.subTasks; len(subTasks) > 0 {
-				collectFiles(subTasks, seen)
+
+			if packageIdToCanonicalPath != nil {
+				for _, resolution := range task.resolutionsInFile {
```
jakebailey: I don't particularly like that this loop is like this, versus having the package ID before collect and then deduping with it, but there are some weird ordering constraints here

andrewbranch (approved):

internal/compiler/filesparser.go
jakebailey: I stared at this for a while and figured out how to actually make it behave the way I wanted without this loop

jakebailey: Basically, this loop was working around the fact that packages get their IDs from incoming edges, so there was a race. Now it's way less weird

internal/compiler/program.go
```diff
@@ -241,6 +242,14 @@ func (p *Program) UpdateProgram(changedFilePath tspath.Path, newHost CompilerHos
 	if !canReplaceFileInProgram(oldFile, newFile) {
 		return NewProgram(newOpts), false
 	}
+	// If this file is part of a package redirect group (same package installed in multiple
+	// node_modules locations), we need to rebuild the program because the redirect targets
+	// might need recalculation. A file is in a redirect group if it's either a canonical
+	// file that others redirect to, or if it redirects to another file.
+	if _, ok := p.deduplicatedPathMap[changedFilePath]; ok {
```
jakebailey: On second reading, I think this is wrong; this mapping is always populated even if something was not deduped.

I will have to redo this to instead build a list of files that are actually a part of deduping, I think. Or, not bother. I have not written a test for this case, so...

internal/compiler/filesparser.go
```diff
@@ -283,6 +292,14 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
 	var sourceFilesFoundSearchingNodeModules collections.Set[tspath.Path]
 	libFilesMap := make(map[tspath.Path]*LibFile, libFileCount)
 
+	var redirectTargetsMap map[tspath.Path][]string
+	var deduplicatedPaths collections.Set[tspath.Path]
+	var packageIdToCanonicalPath map[module.PackageId]tspath.Path
+	if !loader.opts.Config.CompilerOptions().DeduplicatePackages.IsFalse() {
```
RyanCavanaugh, line 298:
```suggestion
	if loader.opts.Config.CompilerOptions().DeduplicatePackages.IsTrueOrUnknown() {
```
(is this equivalent?)

RyanCavanaugh (approved):

internal/compiler/filesparser.go
jakebailey: It is, I waffled between the two and just chose the not false mode. Can followup PR this if you want (the PR automerged, oops) but we are generally inconsistent (the OrUnknown forms are "new")

RyanCavanaugh: Not worth a followup IMO

internal/compiler/filesparser.go
```diff
@@ -220,6 +224,11 @@ func (w *filesParser) start(loader *fileLoader, tasks []*parseTask, depth int) {
 				}
 			}
 
+			// Propagate packageId to data if we have one and data doesn't yet
```
sheetalkamat, line 227: In strada we didnt get the package ids to evaluate on already found file - but i think this is better behavior.. (that is if file was root file and then added through import (with package id - it wasnt added to "redirect" map)  But just noting it for change in behavior.

jakebailey: Yeah, this was something I realized but didn't note, and I feel like it's a better behavior too.

PR #2363, opened 2025-12-12
Cache name table on SourceFile

jakebailey:
Same as Strada:

```ts
export function getNameTable(sourceFile: SourceFile): Map<__String, number> {
    if (!sourceFile.nameTable) {
        initializeNameTable(sourceFile);
    }

    return sourceFile.nameTable!; // TODO: GH#18217
}

function initializeNameTable(sourceFile: SourceFile): void {
    const nameTable = sourceFile.nameTable = new Map();
    sourceFile.forEachChild(function walk(node) {
```

### DIFF
--- a/internal/ls/utilities.go
+++ b/internal/ls/utilities.go
@@ -483,7 +483,6 @@ func isObjectBindingElementWithoutPropertyName(bindingElement *ast.Node) bool {
 		bindingElement.PropertyName() == nil
 }
 
-
 func isRightSideOfPropertyAccess(node *ast.Node) bool {
 	return node.Parent.Kind == ast.KindPropertyAccessExpression && node.Parent.Name() == node
 }


internal/ast/ast.go
```diff
@@ -10944,6 +10946,61 @@ func (node *SourceFile) ECMALineMap() []core.TextPos {
 	return lineMap
 }
 
+// GetNameTable returns a map of all names in the file to their positions.
+// If the name appears more than once, the value is -1.
+func (file *SourceFile) GetNameTable() map[string]int {
+	file.nameTableOnce.Do(func() {
+		nameTable := make(map[string]int, file.IdentifierCount)
+
+		isTagName := func(node *Node) bool {
```
gabritto: We have that function in `ls/utilities.go`, not super important but we could move it and reuse it here.

gabritto: Unless this is somehow an optimization?

jakebailey: This is that helper; it's been deleted out of the other package as this code was the only use. 

But, I see one potential future use, so I could split it back out again.

`isArgumentOfElementAccessExpression` is duplicated, though, and nobody noticed.

jakebailey: Ah, no, my bad for not deleting that latter one too. It really is only for this code.

gabritto: Yeah, I missed that `getNameTable()` used to be in `ls` and got confused.

gabritto (approved):

PR #2370, opened 2025-12-12
Format options + fourslash testing + fix formatting bugs

### LINKED ISSUE #2450: Review ported code from formatter
Original code:

```
const parent = findPrecedingToken(tokenInfo.end, sourceFile, enclosingNode)?.parent || previousParent!;
```

https://github.com/microsoft/TypeScript/blob/9d364978d3e8e4de01bb2bec8083b58c1f710232/src/services/formatting/formatting.ts#L567

New code:

```go
parent := astnav.FindPrecedingToken(w.sourceFile, tokenInfo.Loc.End())
if parent == nil {
	parent = w.previousParent
}
```

https://github.com/microsoft/typescript-go/blob/05278024e75c0410505d8c4ee4e669b624bb1a6a/internal/format/span.go#L304-L307

Looks like we at least need something like

```diff 
- parent := astnav.FindPrecedingToken(w.sourceFile, tokenInfo.Loc.End())
+ var parent *ast.Node
+ prevToken := astnav.FindPrecedingToken(w.sourceFile, tokenInfo.Loc.End())
+ if prevToken != nil {
+ 	parent = prevToken.Parent
+ }
  if parent == nil {
  	parent = w.previousParent
  }
```

iisaduan:
Adds formatting options and links up the server's `FormatCodeOptions` with the formatting functions. Also adds in basic structure for `js`, `javascript`, `ts`, and `typescript` scoped settings, and redirects `LanguageService.UserPreferences()` to return the preferences for the active file.

Moves `formatCodeSettings` to `lsutil`

Formatting is now being tested in fourslash, although not all the commands are implemented due to non-formatting tests crashing on insertion. I've added a fourslashTest property to skip the check for those specific crashes, so that the rest of the test can proceed as normal. The flag(s) can easily be commented out if investigating these crashes.

Also fixes a formatting bug that was adding extra spaces after newlines.

____

Fixes #2450

### DIFF
--- a/internal/ls/lsutil/configuration.go
+++ b/internal/ls/lsutil/configuration.go
@@ -81,7 +81,7 @@ func ParseNewUserConfig(items []any) *UserConfig {
 			newConfig := &UserConfig{}
 			switch i {
 			case 0:
-				// if provided, parse and set "ts/js" as base config
+				// if provided, parse and set "js/ts" as base config
 				defaultPref = NewUserConfig(defaultPref.ts.ParseWorker(config))
 				c = defaultPref.Copy()
 				continue
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -281,7 +281,7 @@ func (s *Server) RequestConfiguration(ctx context.Context) (*lsutil.UserConfig, 
 	configs, err := sendClientRequest(ctx, s, lsproto.WorkspaceConfigurationInfo, &lsproto.ConfigurationParams{
 		Items: []*lsproto.ConfigurationItem{
 			{
-				Section: ptrTo("ts/js"),
+				Section: ptrTo("js/ts"),
 			},
 			{
 				Section: ptrTo("typescript"),
@@ -970,8 +970,7 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 				RegisterOptions: &lsproto.RegisterOptions{
 					DidChangeConfiguration: &lsproto.DidChangeConfigurationRegistrationOptions{
 						Section: &lsproto.StringOrStrings{
-							// !!! Both the 'javascript' and 'js/ts' scopes need to be watched for settings as well.
-							Strings: &[]string{"typescript"},
+							Strings: &[]string{"js/ts", "typescript", "javascript"},
 						},
 					},
 				},
@@ -1010,7 +1009,7 @@ func (s *Server) handleDidChangeWorkspaceConfiguration(ctx context.Context, para
 		s.session.Configure(lsutil.ParseNewUserConfig(settings))
 	} else if settings, ok := params.Settings.(map[string]any); ok {
 		// fourslash case
-		s.session.Configure(lsutil.ParseNewUserConfig([]any{settings["ts/js"], settings["typescript"], settings["javascript"]}))
+		s.session.Configure(lsutil.ParseNewUserConfig([]any{settings["js/ts"], settings["typescript"], settings["javascript"]}))
 	}
 	return nil
 }


internal/fourslash/fourslash.go
```diff
@@ -606,21 +631,29 @@ func (f *FourslashTest) writeMsg(t *testing.T, msg *lsproto.Message) {
 }
 
 func sendRequest[Params, Resp any](t *testing.T, f *FourslashTest, info lsproto.RequestInfo[Params, Resp], params Params) Resp {
-	t.Helper()
+	// t.Helper()
```
jakebailey: Why remove this?

internal/fourslash/fourslash.go
```diff
@@ -606,21 +631,29 @@ func (f *FourslashTest) writeMsg(t *testing.T, msg *lsproto.Message) {
...
-	}
-	resp := resMsg.AsResponse()
-	if resp.Error != nil {
-		t.Fatalf(prefix+"%s request returned error: %s", info.Method, resp.Error.String())
-	}
-	if !resultOk {
-		t.Fatalf(prefix+"Unexpected %s response type: %T", info.Method, resp.Result)
+	switch info.Method {
+	case lsproto.MethodTextDocumentOnTypeFormatting:
+		if !f.reportFormatOnTypeCrash {
```
jakebailey, line 746: This seems like a leftover hack, maybe?

internal/fourslash/fourslash.go
```diff
@@ -642,6 +675,13 @@ func (f *FourslashTest) updateState(method lsproto.Method, params any) {
 	}
 }
 
+func (f *FourslashTest) SetFormatOption(t *testing.T, optionName string, value any) {
+	t.Helper()
+	newPreferences := f.userPreferences.Copy()
+	newPreferences.Set(optionName, value)
```
jakebailey: I don't particularly setting these as strings; can we generate these differently so we can stay static? My options refactor will remove any sort of string-based setting

internal/fourslash/fourslash.go
```diff
@@ -1348,13 +1429,14 @@ func (f *FourslashTest) VerifyImportFixAtPosition(t *testing.T, expectedTexts []
...
+	currentCaretPosition := f.currentCaretPosition
 	params := &lsproto.CodeActionParams{
 		TextDocument: lsproto.TextDocumentIdentifier{
 			Uri: lsconv.FileNameToDocumentURI(f.activeFilename),
 		},
 		Range: lsproto.Range{
-			Start: f.currentCaretPosition,
-			End:   f.currentCaretPosition,
+			End:   currentCaretPosition,
+			Start: currentCaretPosition,
```
jakebailey, line 1635: Seems like a random change

internal/fourslash/fourslash.go
```diff
@@ -2630,29 +2748,41 @@ func (f *FourslashTest) Replace(t *testing.T, start int, length int, text string
 
 // Inserts the text currently at the caret position character by character, as if the user typed it.
 func (f *FourslashTest) typeText(t *testing.T, text string) {
+	// temprorary -- this disables tests failing if format crashes; this unblocks unrelated tests such as codefixes
```
jakebailey, line 2948: Hm, I see; how much does this matter?

internal/ls/lsutil/formatcodeoptions.go
```diff
@@ -0,0 +1,218 @@
...
+	InsertSpaceBeforeFunctionParenthesis                        core.Tristate
+	PlaceOpenBraceOnNewLineForFunctions                         core.Tristate
+	PlaceOpenBraceOnNewLineForControlBlocks                     core.Tristate
+	InsertSpaceBeforeTypeAnnotation                             core.Tristate
+	IndentMultiLineObjectLiteralBeginningOnBlankLine            core.Tristate
+	Semicolons                                                  SemicolonPreference
+	IndentSwitchCase                                            core.Tristate
+}
+
+func FromLSFormatOptions(f *FormatCodeSettings, opt *lsproto.FormattingOptions) *FormatCodeSettings {
```
jakebailey, line 90: Is there any way you can add tests for some of the expected decoding, so I know what is expected? I had forgotten that the LSP has formatting options itself

internal/ls/lsutil/userpreferences.go
```diff
@@ -408,17 +400,22 @@ func (p *UserPreferences) parseWorker(config map[string]any) {
...
 		case "experimental":
 			// !!!
 		default:
-			p.set(name, values)
+			p.Set(name, values)
+			formatSettingsParsed = p.FormatCodeSettings.Set(name, values) || formatSettingsParsed
 		}
 	}
+	if !formatSettingsParsed {
+		p.FormatCodeSettings = nil
```
jakebailey: Do we have to reset this, or could we just set it to the default?

Maybe this is moot if I'm going to change this otherwise, though.

internal/lsp/server.go
```diff
@@ -269,30 +268,32 @@ func (s *Server) RefreshCodeLens(ctx context.Context) error {
...
-		return s.session.NewUserPreferences(), nil
+		return project.NewConfig(s.session.NewUserPreferences()), nil
 	}
 	configs, err := sendClientRequest(ctx, s, lsproto.WorkspaceConfigurationInfo, &lsproto.ConfigurationParams{
 		Items: []*lsproto.ConfigurationItem{
 			{
 				Section: ptrTo("typescript"),
 			},
+			{
+				Section: ptrTo("ts"),
```
jakebailey: The actual section is `ts/js`, but we should just ignore those for now since we don't pull inferred project compiler options out of it yet.

internal/lsp/server.go
```diff
@@ -924,17 +925,14 @@ func (s *Server) handleExit(ctx context.Context, params any) error {
...
 func (s *Server) handleDidChangeWorkspaceConfiguration(ctx context.Context, params *lsproto.DidChangeConfigurationParams) error {
-	settings, ok := params.Settings.(map[string]any)
-	if !ok {
+	// !!! only implemented because needed for fourslash
+	if params.Settings == nil {
 		return nil
+	} else if settings, ok := params.Settings.([]any); ok {
+		s.session.Configure(project.ParseConfiguration(settings))
+	} else if settings, ok := params.Settings.(map[string]any); ok {
+		s.session.Configure(project.ParseConfiguration([]any{settings["typescript"], settings["ts"], settings["javascript"], settings["js"]}))
```
jakebailey: If we want to respond to this, we're going to have to also register these in `handleInitialized`

internal/project/configuration.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (c *Config) Copy() *Config {
+	return &Config{
+		Ts: c.Ts.CopyOrDefault(),
+		js: c.js.CopyOrDefault(),
+	}
+}
+
+// any non-nil field in b is copied into a
+func (a *Config) CopyInto(b *Config) *Config {
```
jakebailey: Do we need mutability for any of the prefs like this? I had thought the change I made before to make it all immutable worked pretty well

internal/project/session.go
```diff
@@ -190,40 +193,52 @@ func (s *Session) GetCurrentDirectory() string {
...
+	return s.workspaceConfig.Copy()
+}
+
+// !!! ts/js preferences
+// Gets copy of current UserPreferences
 func (s *Session) UserPreferences() *lsutil.UserPreferences {
 	s.configRWMu.Lock()
 	defer s.configRWMu.Unlock()
-	return s.userPreferences.Copy()
+	return s.workspaceConfig.Ts.Copy()
```
jakebailey: What is the plan for returning the "right" one?

If we aren't handling that yet, can we just skip doing the JS stuff for the time being?

internal/scanner/scanner.go
```diff
@@ -2458,7 +2458,7 @@ func GetECMAEndLinePosition(sourceFile *ast.SourceFile, line int) int {
 	for {
 		ch, size := utf8.DecodeRuneInString(sourceFile.Text()[pos:])
 		if size == 0 || stringutil.IsLineBreak(ch) {
-			return pos
+			return pos - 1
```
jakebailey, line 2461: I feel suspicious of this; some compiler baselines have changed

internal/fourslash/fourslash.go
iisaduan: oops, I commented it out while debugging because it was easier to stack trace without it marked as helper

internal/scanner/scanner.go
iisaduan: I think the error message was using the wrong function. I will fix that

internal/ls/lsutil/userpreferences.go
iisaduan: Yeah, we should set it to the default if nil

internal/fourslash/fourslash.go
iisaduan: there's currently ~10 passing tests that this unblocks, and some more (~10?) in `failingtests.txt` that fail for other reasons besides the formatting crashes. There's more unparsed tests (they have commands like `format.Range` for example) that I didn't include due to trying to get this pr out faster, that also crash. These numbers are a lot lower after the formatting bugs that I've fixed+included in this PR. I'm very pro-keeping this until either all of autoimports and completions tests are passing or formatting bugs are more ironed out. It'll be a lot less annoying for anyone working on completions or autoimports to work on their features without the formatter doing unrelated things.

iisaduan: more info: 
The current disabling only disables reporting crashes when calling format on keystroke, which in strada fourslash was called after every character passed into `edit.insert`. However, completions and autoimport/codefix tests don't usually use `edit.insert` to simulate keystrokes-- they use `edit.insert` to simulate the file result after accepting a completion or codefixaction, which in the real world, returns already formatted text, and doesn't format on keystroke. We also don't really lose any formatting test coverage because if the formatter crashes, the check on formatted text will be wrong

internal/scanner/scanner.go
iisaduan: yup, the compiler error function was ported wrong, probably because this function was ported wrong

internal/project/configuration.go
iisaduan: This is mostly to be careful in the case that if we don't parse some scope, we don't overwrite the other scope(s) with our defaults

internal/lsp/server.go
iisaduan: This is another one of those cases that was for fourslash, I'll add a comment. I had to update it to work with Daniel's changes

internal/fourslash/fourslash.go
iisaduan: I fixed the function so that it updates the caret position correctly. This makes it a bit more clear where the original caret position comes from

internal/project/session.go
iisaduan: Added JS handling

internal/project/configuration.go
```diff
@@ -0,0 +1,60 @@
+package project
+
+import (
+	"github.com/microsoft/typescript-go/internal/ls/lsutil"
+)
+
+type Config struct {
```
jakebailey: I think `Config` is a bit too generic of a name, and it does seem like this should live in `lsutil` since it's basically just a way to encode two user preferences structs and that isn't project specific.

internal/lsp/server.go
jakebailey: I would still delete `ts` and `js` here, I don't think they are real?

internal/ls/autoimport/registry.go
```diff
@@ -210,10 +210,11 @@ type Registry struct {
 	specifierCache map[tspath.Path]*collections.SyncMap[tspath.Path, string]
 }
 
-func NewRegistry(toPath func(fileName string) tspath.Path) *Registry {
+func NewRegistry(toPath func(fileName string) tspath.Path, preferences *lsutil.UserPreferences) *Registry {
 	return &Registry{
-		toPath:      toPath,
-		directories: make(map[tspath.Path]*directory),
+		toPath:          toPath,
+		userPreferences: preferences,
```
jakebailey, line 216: Is this ever used?

jakebailey: I guess it must be, because I don't see the field introduced in this PR

iisaduan: it's from #2390 and eventually end up being passed to locations in `snapshot` which are supposed to be non-nil in this PR

internal/project/configuration.go
iisaduan: `ServerConfig`?

jakebailey: `CombinedUserPreferences`? `AllUserPreferences`? I didn't come up with a good name so I didn't suggest one before, which is not helpful 😄

My hope is that in a future change, this struct is immutable too (probably in a rebase of my PR), just to make this easier to work with

iisaduan: I had originally imagined that we would factor out the stuff configures server side preferences (stuff that would never need to be passed into `LanguageService`). There's also `tsserverOptions` in strada, I figured we'd have similar options here (or are they all just on `Server` already?

jakebailey: Maybe I'm totally missing something, but the only reference to `tsserverOptions` is in a comment in Corsa...

iisaduan: I think I'm referring to a bunch of options that sometimes got tacked onto `userPreferences`, and I decided to name it `tsserverOptions` here incase we needed to parse any server-related options. I don't think that's necessary anymore, as it looks like extra options have been moved to other places

iisaduan: So now this struct only has `userPreferences`, so rename to `UserPreferenceConfig`? I like keeping "config" in the name because of the LSP config command

jakebailey: At that point, I guess `UserConfiguration` is probably fine, given it encompasses all user configuration?

internal/ls/lsutil/configuration.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (c *UserConfig) Copy() *UserConfig {
+	return &UserConfig{
+		ts: c.ts.CopyOrDefault(),
+		js: c.js.CopyOrDefault(),
+	}
+}
+
+// any non-nil field in b is copied into a
+func (a *UserConfig) CopyInto(b *UserConfig) *UserConfig {
```
jakebailey: Shouldn't this be `CopyFrom`, if it copies from `b` to `a`?

internal/ls/lsutil/configuration.go
```diff
@@ -0,0 +1,95 @@
...
+	return newUserConfig
+}
+
+func (c *UserConfig) Ts() *UserPreferences {
+	return c.ts
+}
+
+func (c *UserConfig) Js() *UserPreferences {
+	return c.js
+}
```
jakebailey: Nit, but I would do `TS()` and `JS()`.

internal/project/snapshot.go
```diff
@@ -360,14 +356,9 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
...
-	config := s.config
+	config := s.allUserPreferences
 	if change.newConfig != nil {
-		if change.newConfig.tsUserPreferences != nil {
-			config.tsUserPreferences = change.newConfig.tsUserPreferences.CopyOrDefault()
-		}
-		if change.newConfig.formatOptions != nil {
-			config.formatOptions = change.newConfig.formatOptions
-		}
+		config = config.CopyInto(change.newConfig)
```
jakebailey: `CopyInto` isn't mutating `config`, right? So maybe it's acutally `Merge`? I think I am just very confused by the name when it's not mutating something, especially given this example

internal/lsp/server.go
```diff
@@ -272,30 +272,29 @@ func (s *Server) RefreshCodeLens(ctx context.Context) error {
...
 		Items: []*lsproto.ConfigurationItem{
+			{
+				Section: ptrTo("ts/js"),
+			},
 			{
 				Section: ptrTo("typescript"),
 			},
+			{
+				Section: ptrTo("javascript"),
+			},
```
jakebailey, line 291: I think we need to change fourslash to provide these, in the right order, right? I didn't see that in the other file.

iisaduan: Fourslash sends us a map with the languages as the key. We order them here in `didChangeConfiguration` https://github.com/microsoft/typescript-go/pull/2370/changes#diff-0077cc8440363b2137f8e3e405549e59083242b4281d50b08702a6b7153492a8R1011-R1013

jakebailey: Ah, right, I forgot we still did that in testing.

internal/lsp/server.go
jakebailey: This needs to be updated now (in the same order).

Issue #2383, opened 2025-12-15
LSP: Path suggestion missing (import/export)


besserwisser:
## Steps to reproduce

Given you are for example in a file `src/my-app/index.ts` and you have a folder next to the `index.ts` called `lib` and in this lib folder a file called `test.ts`

`import { test } from './lib/t`
`export * from './lib/t`

When typing one of the above

## Behavior with `typescript@5.9`

Would suggest to complete to `./lib/test` 

## Behavior with `tsgo`

Does not suggest file path completion.

dalisoft:
Maybe related to https://github.com/microsoft/typescript-go/issues/2189

besserwisser:
Thanks. I saw it, but I am not sure if it is related. Because auto import might be something different than file path completion in insert mode. At least as far as I understand from a users perspective :D

andrewbranch:
Module path completions are a known to-do.

besserwisser:
> Module path completions are a known to-do.

Ahh okay, thank you. Is there a list of to-dos? I tried to check before creating the issue, but was sure where. 

Thanks

andrewbranch:
Yes and no; we have a list of failing/differing tests, and I traced these to a function that was empty except for a code comment.

Issue #2336, opened 2025-12-11
Crash for signature help in arrow function parameters


DanielRosenwasser:
```ts
const comparers: Array<(a: any, b: any) => boolean> = [];

comparers.push((a,/**/ b) => true);
```

Somehow this brings down the whole server without sending back a proper stack trace.

gabritto:
The underlying issue happens because we try to print the signature label for type symbol `(a: any, b: any) => boolean`, which is an anonymous type, so we print `\xfetype` as its name, but `\xfe` is not valid UTF-8, so JSON marshalling fails and the server stops.

jakebailey:
We should probably make it so that the node builder detects internal types and auto-escapes them like the old compiler?

gabritto:
> We should probably make it so that the node builder detects internal types and auto-escapes them like the old compiler?

Where was the old compiler escaping this? Because I couldn't find it, and when I tested this same signature help scenario in Strada, we returned "__type(...)" as label.

jakebailey:
Well, yes, by nature of symbol names and so on like this having those `__` prefix (the underscore escaping). But, those names were sometimes leaked, which is a no-go in the Go code due to the UTF-8 trick (which I have not been able to figure out a better alternative to...)

PR #2390, opened 2025-12-16
New auto-import infrastructure

### LINKED ISSUE #1677: panic handling request textDocument/completion (autoimports)
## Stack trace

```
panic handling request textDocument/completion Could not find symbol 'InternalIdentifier' by key 'InternalIdentifier' in module "c:/Projects/a/b/node_modules/@types/lodash/index" goroutine 1441067 [running]:
runtime/debug.Stack()
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0xc0000f4dc0, 0xc119702d80)
	github.com/microsoft/typescript-go/internal/lsp/server.go:532 +0x58
panic({0x7ff72291a7e0?, 0xc07381ee40?})
	runtime/panic.go:783 +0x132
github.com/microsoft/typescript-go/internal/ls.(*exportInfoMap).rehydrateCachedInfo(0xc136951090, 0xc0ebc4e008, {0x1, {0xc0392e5d71, 0xa}, {0xc0392e5d71, 0xa}, {0x0, 0x0}, {0xc052f70361, ...}, ...})
	github.com/microsoft/typescript-go/internal/ls/autoimports.go:302 +0x494
github.com/microsoft/typescript-go/internal/ls.(*exportInfoMap).search.func1(...)
	github.com/microsoft/typescript-go/internal/ls/autoimports.go:231
github.com/microsoft/typescript-go/internal/core.Map[...](...)
	github.com/microsoft/typescript-go/internal/core/core.go:57
github.com/microsoft/typescript-go/internal/ls.(*exportInfoMap).search(0xc136951090, 0xc0ebc4e008, {0xc108e5aa80?, 0xc0606f8230?}, 0x0, 0xc1369511f8, 0xc136951338)
	github.com/microsoft/typescript-go/internal/ls/autoimports.go:230 +0x3d4
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).searchExportInfosForCompletions(0xc008579ec0, {0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc0ebc4e008, 0xc0f1e1d8c8, 0xc0991e7950, 0x0, 0x0, 0x1, {0xc1c426d168, ...}, ...)
	github.com/microsoft/typescript-go/internal/ls/autoimportsexportinfo.go:172 +0x3da
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionData.func9()
	github.com/microsoft/typescript-go/internal/ls/completions.go:1271 +0x26c
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionData.func17()
	github.com/microsoft/typescript-go/internal/ls/completions.go:1686 +0x5f6
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionData.func18(...)
	github.com/microsoft/typescript-go/internal/ls/completions.go:1713
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionData(0xc008579ec0, {0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc0ebc4e008, 0xc0f1e1d8c8, 0x2d71, 0xc0991e7950)
	github.com/microsoft/typescript-go/internal/ls/completions.go:1741 +0x1ee2
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionsAtPosition(0xc008579ec0, {0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc0f1e1d8c8, 0x2d71, 0x0, 0xc0991e7950, 0xc000282f90)
	github.com/microsoft/typescript-go/internal/ls/completions.go:399 +0x2dd
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideCompletion(0xc008579ec0, {0x7ff722ca5cb0, 0xc0bafbddb0}, {0xc104aefc80?, 0x1?}, {0xe4febe70?, 0xc0?}, 0xc082166140, 0xc000282f90, 0xc0991e7950)
	github.com/microsoft/typescript-go/internal/ls/completions.go:50 +0xe7
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleCompletion(0xc0003501c0?, {0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc008579ec0, 0xc119702c60)
	github.com/microsoft/typescript-go/internal/lsp/server.go:754 +0xff
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].15({0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc119702d80)
	github.com/microsoft/typescript-go/internal/lsp/server.go:518 +0x130
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0xc0000f4dc0, {0x7ff722ca5cb0, 0xc0bafbddb0}, 0xc119702d80)
	github.com/microsoft/typescript-go/internal/lsp/server.go:424 +0x10d
	github.com/microsoft/typescript-go/internal/lsp/server.go:329 +0x3a
	github.com/microsoft/typescript-go/internal/lsp/server.go:349 +0x8aa

[Error - 9:52:05 AM] Request textDocument/completion failed.
  Message: InternalError: panic handling request textDocument/completion: Could not find symbol 'InternalIdentifier' by key 'InternalIdentifier' in module "c:/Projects/a/b/node_modules/@types/lodash/index"
  Code: -32603 
```

## Steps to reproduce

Unsure, happened while editing. I've changed InternalIdentifier - it was a internal identifier that is definitely not defined in lodash, is not in the source code of the fileI was editing. In fact the file I was editing didn't use lodash either.

I have the latest extension - 20250903.1

### LINKED ISSUE #2143: Panic handling request completionItem/resolve Some exportInfo should match the specified exportMapKey
## Stack trace

```
[ERROR][2025-11-21 09:50:50] /usr/share/nvim/runtime/lua/vim/lsp/log.lua:151	"rpc"	"tsgo"	"stderr"	"panic handling request completionItem/resolve Some exportInfo should match the specified exportMapKey goroutine 38736 [running]:\nruntime/debug.Stack()\n\truntime/debug/stack.go:26 +0x5e\ngithub.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0xc000238008, 0xc01ded01e0)\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:572 +0x58\npanic({0xb46f40?, 0xee83f0?})\n\truntime/panic.go:783 +0x132\ngithub.com/microsoft/typescript-go/internal/ls.(*LanguageService).getImportCompletionAction(0xc01da9e420, {0xef1228, 0xc01da9e3c0}, 0xc029a2b908, 0xc029086b88?, 0xc00b787bd0?, 0xc02a61fb08, 0xb22, {{0xc005e06714, 0x4}, ...}, ...)\n\tgithub.com/microsoft/typescript-go/internal/ls/autoimports.go:367 +0x30c\ngithub.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionItemActions(0xc01da9e420, {0xef1228, 0xc01da9e3c0}, 0xc029a2b908, 0xc02a61fb08, 0xb23, 0xc00091aa00, 0xc00091b1c0)\n\tgithub.com/microsoft/typescript-go/internal/ls/completions.go:5328 +0x49a\ngithub.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionItemDetails(0xc01da9e420, {0xef1228, 0xc01da9e3c0}, 0xc00009c180?, 0xb23, 0xc02a61fb08, 0xc0612e21e0, 0xc00091aa00)\n\tgithub.com/microsoft/typescript-go/internal/ls/completions.go:5072 +0x5ff\ngithub.com/microsoft/typescript-go/internal/ls.(*LanguageService).ResolveCompletionItem(0xc01da9e420, {0xef1228, 0xc01da9e3c0}, 0xc0612e21e0, 0xc00091aa00)\n\tgithub.com/microsoft/typescript-go/internal/ls/completions.go:4996 +0x85\ngithub.com/microsoft/typescript-go/internal/lsp.(*Server).handleCompletionItemResolve(0xc000238008, {0xef1228, 0xc01da9e3c0}, 0xc0612e21e0, 0xc01ded01e0)\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:896 +0x125\ngithub.com/microsoft/typescript-go/internal/lsp.init.func1.registerRequestHandler[...].30({0xef1228, 0xc01da9e3c0}, 0xc01ded01e0)\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:534 +0xb6\ngithub.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0xc000238008, {0xef1260?, 0xc037a952c0?}, 0xc01ded01e0)\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:451 +0x153\ngithub.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func1()\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:354 +0x3a\ncreated by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 34\n\tgithub.com/microsoft/typescript-go/internal/lsp/server.go:374 +0x9ad\n\n"
[ERROR][2025-11-21 09:50:50] /usr/share/nvim/runtime/lua/vim/lsp/log.lua:151	"LSP[tsgo]"	"on_error"	{ code = "NO_RESULT_CALLBACK_FOUND", err = { id = 129, jsonrpc = "2.0", result = vim.NIL } }

```

## Steps to reproduce

1.Copy this code 

```tsx
      [].reduce((acc, curr) => {
        acc[curr.id] ??= {
          id: curr.id,
          // type data
        };
        return acc;
      }, {} as Record<number, any>)
```
2. type `data`

https://github.com/user-attachments/assets/a8181932-3da6-486e-bfc3-32f61f5114c7

### LINKED ISSUE #2189: LSP doesn't offer auto-importing packages
### Extension Version

0.20251203.1

### VS Code Version

1.107.0-insider

### Operating system Version

MacOS 26.1

### Steps to reproduce

1. Create an empty TypeScript project
2. Install `typescript-result`
3. Create index.ts
4. Type `Result` and nothing happens

### Issue

I noticed somewhere that auto-imports are mostly done, but in my work project, it doesn't auto-import most of the packages that I have installed. It usually auto-imports built-in dependencies such as `node:crypto` or global types, but not installed deps.

Seems not to be an issue with pnpm, catalog dependencies, etc. It happens with npm as well.

I'm really sorry if this is a current limitation and has been mentioned somewhere, and I didn't notice.

### LINKED ISSUE #2301: Crash for value usage of type-only import
```ts
// @allowImportingTsExtensions: true
// @module: preserve

// @filename: abc.ts
class Abcdef {}

// @filename: usage.ts
import { type Abcdef } from "./abc.ts";

new A/**/
```

Request completions and resolve the completion for `Abcdef`.

```
[error] panic handling requestcompletionItem/resolvesymbolOriginInfo: unknown data type for moduleSymbol(): *ls.symbolOriginInfoTypeOnlyAliasgoroutine 1038 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x64
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0x4000120008, 0x4002318f00)
	github.com/microsoft/typescript-go/internal/lsp/server.go:701 +0x40
panic({0x7ff6d7381c80?, 0x40033a1bf0?})
	runtime/panic.go:783 +0x120
github.com/microsoft/typescript-go/internal/ls.(*symbolOriginInfo).moduleSymbol(...)
	github.com/microsoft/typescript-go/internal/ls/completions.go:238
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionItemActions(0x4002b513b0, {0x7ff6d7868b88, 0x4002b51350}, 0x4001326608, 0x4001857208, 0x35, 0x4002338ac0, 0x400258d1c0)
	github.com/microsoft/typescript-go/internal/ls/completions.go:5339 +0x458
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).getCompletionItemDetails(0x4002b513b0, {0x7ff6d7868b88, 0x4002b51350}, 0x1?, 0x35, 0x4001857208, 0x40027d1d60, 0x4002338ac0)
	github.com/microsoft/typescript-go/internal/ls/completions.go:5089 +0x48c
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ResolveCompletionItem(0x4002b513b0, {0x7ff6d7868b88, 0x4002b51350}, 0x40027d1d60, 0x4002338ac0)
	github.com/microsoft/typescript-go/internal/ls/completions.go:5025 +0x68
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleCompletionItemResolve(0x4000120008, {0x7ff6d7868b88, 0x4002b51350}, 0x40027d1d60, 0x4002318f00)
	github.com/microsoft/typescript-go/internal/lsp/server.go:1026 +0xd8
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerRequestHandler[...].35({0x7ff6d7868b88, 0x4002b51350}, 0x4002318f00)
	github.com/microsoft/typescript-go/internal/lsp/server.go:593 +0xc0
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x4000120008, {0x7ff6d7868bc0?, 0x4002529c20?}, 0x4002318f00)
	github.com/microsoft/typescript-go/internal/lsp/server.go:501 +0x128
	github.com/microsoft/typescript-go/internal/lsp/server.go:404 +0x34
	github.com/microsoft/typescript-go/internal/lsp/server.go:424 +0x7a0
```

andrewbranch:
This PR introduces a completely redesigned auto-import collection infrastructure. As we ported Strada's auto-import code to Corsa, two things became clear:

1. The cache infrastructure would need to be redesigned to work with the snapshot-based LSP.
2. We were leaving significant parallelization opportunities on the table.

We wanted to address these issues before (or while) adding package.json-based dependency discovery, the main feature missing from the initial port of Strada's auto-imports code.

While the data collection backend has been completely redesigned, the fix generation code is mostly unchanged, although the files have been moved from `internal/ls` into `internal/ls/autoimport`.

I’m closing existing all bugs/crashes filed against this repo’s auto-imports implementation with this PR. Please comment if you have a bug that still repros the exact same way after this PR.

- Closes https://github.com/Microsoft/TypeScript/issues/60704
- Closes https://github.com/Microsoft/TypeScript/issues/62829
- Closes #2301
- Closes #1677 
- Closes #2189
- Closes #2143 

## Strada's collection and cache infrastructure

The earliest version of auto-imports in TypeScript had no caching or separate collection phase. During every global completions request, we used the type checker to gather every export from every source file in the program. This was actually fairly fast, in part because we deferred computing the actual import module specifier until a completion item was highlighted. Eventually, we wanted to address two common complaints: (1) users wanted to see the import module specifier for every item in the completion list without having to highlight it first, and (2) we would often suggest imports from transitive dependencies, bloating the list with entries that should really never be imported. Computing module specifiers for every item in the list slowed things down enough that we implemented a caching layer to compensate. Since the cache included module specifiers and filtering based on what dependencies were accessible to the file requesting completions, much of the cache was specific to the requesting file, and needed to be rebuilt whenever the user started editing a different file. 

Later, we added a very commonly requested feature: auto-imports from package.json dependencies. Since the data source for auto-imports was source files already in the program, a common complaint was that immediately after `npm install`ing a dependency, auto-imports from that package wouldn't show up. Users would have to manually write an import from the new package once, then auto-imports would start appearing, even in other files. To address this without completely overhauling all the existing infrastructure, which was designed to use a type checker to extract exports from source files in a program, we added a separate program and type checker “in the background” (scare quotes because we couldn’t actually do anything in a separate process; we just added it onto project loading) containing entrypoints from package.json dependencies that were absent from the main program. Then, the language service could basically just run its existing auto-import collection code twice: once for the main program, and once for the package.json dependencies program. Since this had the potential to add significant amounts of work both to project loading and completions requests, we added a conservative bail-out if more than ten entrypoints were going to be loaded into the background program. This bail-out strategy was a big and imprecise hammer, since those ten entrypoints could trigger an arbitrarily large or small amount of work through transitive imports that would be eventually be filtered out. 

In short, auto-imports was due for a rewrite even before typescript-go.

## New infrastructure

The primary innovations of the new system are:

### Granular buckets

The new infrastructure stores exports in separate buckets that are independent of each other and independent of the requesting file. The `autoimport.Registry` contains one bucket for each project owning the editor's open files, plus one bucket for each `node_modules` directory visible to the open files. Project buckets contain exports from the project's local, non-`node_modules` files. This means that when switching between open files, any buckets that are already populated can be reused. When a project updates, all `node_modules` buckets are unaffected; when an `npm install` happens, only the `node_modules` directories that were touched need to be updated.

### Highly parallelized collection

Previously, collection of exports for package.json dependencies worked by creating a program containing all the entrypoints we care about, then using the type checker to extract exports from every source file in that program. The disadvantage of creating a program for this purpose is that we would resolve and parse transitive dependencies that have absolutely no impact on the exports we actually care about. In the new infrastructure, we instead create a mock program (`autoimport.aliasResolver`) for each dependency package, seeded with just the entrypoint files for that package. We then walk the exports of those entrypoints, and only use a checker when one of those exports is a re-export from another module. The result is that we only process exactly what we need, we limit our reliance on type checkers, and when we do need a type checker, the data backing it is much smaller, so we can create many checkers very cheaply.

### Pre-computed module specifiers for `node_modules`

Since the old infrastructure originally started from a program's source files, module specifier generation always went through the full process of “importing file name + destination file name => module specifier.” That process could be complicated for node_modules files, since the destination file could be the realpath of a symlinked package, or could have been a transitive import that’s not actually reachable from the importing file, or could be blocked by the package’s package.json `"exports"`. Without knowing how that file got into the program, there were a dozen edge cases that needed to be considered to reverse engineer the correct module specifier, or even to determine whether a valid one exists at all. With the new directory-based buckets and extraction only from dependency entrypoints, we actually know the module specifier for every entrypoint as we find it, so we just store it alongside the export information.

## Limitations / observable differences compared to Strada

* Since we use `sourceFile.Symbol.Exports` instead of `checker.GetExportsAndPropertiesOfModule`, properties of `export =` assignments are not collected. Many of these properties were filtered out in the old logic because they were highly suspicious anyway (e.g. class instance methods, properties of arrays). There is a special syntactic carve-out for JavaScript `module.exports = { foo, bar, ... }` patterns, which are still collected. This limitation did not cause any new test failures.
* For node_modules packages, since we only initialize the type checker with entrypoints from that package, re-exports of globals declared in other packages or non-entrypoint files of that package will not resolve, and so may lack proper categorization and deduplication with other exports. (This limiation does not apply to re-exports from ambient modules declared elsewhere, which has special handling.) Exporting a global is rare; I would be surprised if this limitation were ever noticed.
* Since we only use a type checker when we encounter a re-export, some categorization features that eventually determine the icon for a completion entry may differ. In particular, `export declare const foo: SomeType` would appear as a function if `SomeType` declared call signatures, but now appears as a variable.
* Previously, an unresolvable module augmentation would not contribute anything to auto-imports. Now, they do, even though importing them will cause an error. This seems possibly desirable, since the user may fix the underlying module resolution issue.
* For now, untyped dependency packages don't show up. Previously, we would parse JavaScript file entrypoints as a last resort. There are currently bugs in JavaScript reparsing that cause crashes, performance problems, and phantom exports when loading bundled JavaScript files, so I've disabled JS loading until those are addressed.
* Previously, completions would usually avoid offering multiple ways of importing the same symbol based on several ranking criteria, and then falling back to an arbitrary symbol-based order. The same ranking criteria is still applied, but instead of falling back to a totally arbitrary order, same-ranked entries generate multiple completion items.

## Performance

It's extremely hard to compare this to `main` to Strada, because each has its own limitations that results in very different behavior, not just different timing characteristics:

* Strada: volatile cache, package.json dependencies parsed on project load, more info cached on completions request
* Corsa main: no cache, no package.json dependencies, everything done during completions
* Corsa new: stable cache, package.json dependencies parsed on first edit, local module specifiers cached on completions request

That said, here are some numbers on two example projects.

### Signal-Desktop

|            | Project load  | First edit | First completion | Subsequent completion | Number of entries |
|------------|---------------|------------|------------------|-----------------------|-------------------|
| Strada     | 301 ms        |            | 348 ms           | 101 ms                | 13,688            |
| Corsa main |               |            | 387 ms           | 352 ms                | 14,576            |
| Corsa PR   | 0.25 ms       | 129 ms     | 99 ms            | 73 ms                 | 13,589            |

### VS Code

|            | Project load | First edit | First completion | Subsequent completion | Number of entries |
|------------|--------------|------------|------------------|-----------------------|-------------------|
| Strada     |              |            | 318 ms           | 95 ms                 | 15,864            |
| Corsa main |              |            | 434 ms           | 398 ms                | 15,851            |
| Corsa PR   | .09 ms       | 100 ms     | 125 ms           | 86 ms                 | 16,310            |

## Follow-up work

* Investigate including untyped packages after aforementioned JS reparsing bugs are fixed. Even after the fix, we may want some extra heuristics for avoiding minified or bundled files.
* There’s currently no specific integration with automatic type acquisition. I think Strada had some special logic for this that’s not ported.
* Import path string completions are not yet implemented (#2383), and the node_modules registry buckets implemented here will be very helpful for simplifying the complex logic that exists in Strada.
* `completionItem/resolve` is surprisingly slow (not just for auto-imports), and may be missing info like JSDoc that was previously(?) there for auto-imports


jakebailey:
> Removing devDependencies from the search should improve parsing/loading time a good bit!

IIRC this isn't desirable because then you won't get auto imports in things like build scripts or tests.

(I have not read the PR quite yet.)

andrewbranch:
I think it’s desirable, because you edit build scripts a lot less than you edit your app code! This is how it always worked in Strada. Almost every single one of our users has `typescript` as a dependency, and polluting all of their projects with our hundreds of API exports doesn’t make any sense.

jakebailey:
Ah, sure, I guess I'm not sure how I've been getting fairly good auto imports in say, Herebyfile

internal/core/core.go
```diff
@@ -621,15 +655,20 @@ func DiffMaps[K comparable, V comparable](m1 map[K]V, m2 map[K]V, onAdded func(K
 	DiffMapsFunc(m1, m2, comparableValuesEqual, onAdded, onRemoved, onChanged)
 }
 
-func DiffMapsFunc[K comparable, V any](m1 map[K]V, m2 map[K]V, equalValues func(V, V) bool, onAdded func(K, V), onRemoved func(K, V), onChanged func(K, V, V)) {
-	for k, v2 := range m2 {
-		if _, ok := m1[k]; !ok {
-			onAdded(k, v2)
+func DiffMapsFunc[K comparable, V1 any, V2 any](m1 map[K]V1, m2 map[K]V2, equalValues func(V1, V2) bool, onAdded func(K, V2), onRemoved func(K, V1), onChanged func(K, V1, V2)) {
+	if onAdded != nil {
```
jakebailey, line 659: I've been meaning to do this 👍

internal/ls/autoimport/aliasresolver.go
```diff
@@ -0,0 +1,219 @@
...
+func (r *aliasResolver) IsSourceFromProjectReference(path tspath.Path) bool {
+	panic("unimplemented")
+}
+
+// SourceFileMayBeEmitted implements checker.Program.
+func (r *aliasResolver) SourceFileMayBeEmitted(sourceFile *ast.SourceFile, forceDtsEmit bool) bool {
+	panic("unimplemented")
+}
+
+var _ checker.Program = (*aliasResolver)(nil)
```
jakebailey, line 227: it continues to be nice that we can just "do this"

internal/ls/autoimport/registry.go
```diff
@@ -0,0 +1,1030 @@
...
+}
+
+func (b BucketState) hasDirtyFileBesides(file tspath.Path) bool {
+	return b.multipleFilesDirty || b.dirtyFile != "" && b.dirtyFile != file
+}
+
+type RegistryBucket struct {
+	state BucketState
+
+	Paths map[tspath.Path]struct{}
```
jakebailey: `collections.Set[tspath.Path]`?

internal/ls/autoimport/registry.go
```diff
@@ -0,0 +1,1030 @@
...
+				checker, done := getChecker()
+				defer done()
+				extractor := b.newExportExtractor("", "", checker)
+				fileExports := extractor.extractFromFile(file)
+				mu.Lock()
+				exports[file.Path()] = fileExports
+				mu.Unlock()
+				stats := extractor.Stats()
+				atomic.AddInt32(&combinedStats.exports, stats.exports)
+				atomic.AddInt32(&combinedStats.usedChecker, stats.usedChecker)
```
jakebailey: These can be `atomic.Int32` on the fields themselves (unless you're relying on a struct copy being """legal""").

internal/ls/autoimport/registry.go
```diff
@@ -0,0 +1,1030 @@
...
+			if len(packageEntrypoints.Entrypoints) == 0 {
+				return
+			}
+
+			entrypointsMu.Lock()
+			entrypoints = append(entrypoints, packageEntrypoints)
+			entrypointsMu.Unlock()
+
+			aliasResolver := createAliasResolver(packageName, packageEntrypoints.Entrypoints)
+			checker, _ := checker.NewChecker(aliasResolver)
```
jakebailey: How often does this happen? Creating checkers is quite expensive, technically, so I assume not often

internal/ls/autoimport/util.go
```diff
@@ -0,0 +1,174 @@
...
+		checker, done := program.GetTypeChecker(ctx)
+		for name := range unresolvedPackageNames.Keys() {
+			if symbol := checker.TryFindAmbientModule(name); symbol != nil {
+				declaringFile := ast.GetSourceFileOfModule(symbol)
+				if packageName := modulespecifiers.GetPackageNameFromDirectory(declaringFile.FileName()); packageName != "" {
+					resolvedPackageNames.Add(packageName)
+				}
+			}
+		}
+		done()
```
jakebailey: deferring this would still be fine, given the func returns.

internal/ls/autoimport/util.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return resolvedPackageNames
+}
+
+func createCheckerPool(program checker.Program) (getChecker func() (*checker.Checker, func()), closePool func(), getCreatedCount func() int32) {
+	maxSize := int32(runtime.GOMAXPROCS(0))
```
jakebailey, line 166: Can we limit this? Or do these actually do so little as to not matter? On my machines that's something like 16 or 20 checkers, which would normally scare me

internal/ls/autoimport/registry.go
andrewbranch: The only thing expensive about creating checkers is global symbol merges, but part of the point of feeding it with the alias resolver is seeding the checker with only the package’s entrypoint files, so very very few globals if any (typically just ambient modules declared by the entrypoints). This is very fast. On my machine, creating fewer checkers than processors was measurably slower.

internal/ls/autoimport/util.go
andrewbranch: Again, limiting this was measurably slower on my machine on projects I tried (including VS Code, which is a single large program)

cmd/tsgo/lsp.go
```diff
@@ -64,6 +64,7 @@ func runLSP(args []string) int {
 	defer stop()
 
 	if err := s.Run(ctx); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
andrewbranch, line 67: I haven’t looked at #2336 yet, but this could be what’s missing to show a stack trace.

internal/checker/checker.go
```diff
@@ -14430,6 +14430,9 @@ func (c *Checker) getEmitSyntaxForModuleSpecifierExpression(usage *ast.Node) cor
 }
 
 func (c *Checker) errorNoModuleMemberSymbol(moduleSymbol *ast.Symbol, targetSymbol *ast.Symbol, node *ast.Node, name *ast.Node) {
+	if c.compilerOptions.NoCheck.IsTrue() {
```
andrewbranch, line 14433: Auto-imports uses checkers for minimal alias resolution, and these code paths go way down a rabbit hole hitting the file system and using the node builder to generate errors. Added some targeted short circuits.

internal/fourslash/_scripts/failingTests.txt
andrewbranch, line 1: Most of the added failures are differing in a way that I think is unobjectionable or even preferable. I mostly focused on not regressing, so I haven’t looked at all of the tests that were previously failing and still are. I think some follow-up to categorize/address all the remaining ones would be good.

internal/module/resolver.go
```diff
@@ -2022,3 +2017,209 @@ func GetAutomaticTypeDirectiveNames(options *core.CompilerOptions, host Resoluti
...
+					for _, prevCondition := range prevConditions {
+						if excludeConditions == nil {
+							excludeConditions = &collections.Set[string]{}
+						}
+						excludeConditions.Add(prevCondition)
+					}
+				}
+
+				prevConditions = append(prevConditions, condition)
+				loadEntrypointsFromTargetExports(subpath, newIncludeConditions, excludeConditions, export)
```
andrewbranch, line 2184: I think this is neat: in order to build a node_modules cache that works for any file in any project, regardless of `customConditions` or the file’s resolution mode, we walk through every conditional export of a package, and alongside the file the entrypoint corresponds to, we record the conditions you would _need to have_ to get there (as we descend into a conditions object) and the conditions you would _need not to have_ (as iterate past a previous condition in the same object). Then, when actually computing completions, we can iterate through these candidate entrypoints and compare the include/exclude sets against the reality for the project and file being queried (ls/autoimport/specifiers.go). Example:

```jsonc
{
  "exports": {
    ".": {
      "node": {
        "custom": "./custom.d.ts",
        "import": "./esm.d.mts",
        "default": "./cjs.d.cts"
      }
    }
  }
}
```

- To offer auto-imports from `custom.d.ts`, the importing file must have `node` and `custom`
- To offer auto-imports from `esm.d.mts`, the importing file must have `node` and `import` and must _not_ have `custom`
- To offer auto-imports from `cjs.d.cts`, the importing file must have `node` and must _not_ have `custom` or `import`.

This is one of the strategy changes from Strada that lets us build a cache that can be used by more than one file.

testdata/baselines/reference/fourslash/autoImports/autoImportErrorMixedExportKinds.baseline.md
```diff
@@ -0,0 +1,12 @@
+// === Auto Imports === 
+```ts
+// @FileName: /b.ts
+foo/**/
+
+``````ts
```
jakebailey, line 6: The baselines seem to need a newline between blocks

andrewbranch: I noticed this earlier, but I didn’t touch the baseline implementation in this PR

jakebailey: Oh, sorry, I thought it was new.

andrewbranch:
> Ah, sure, I guess I'm not sure how I've been getting fairly good auto imports in say, Herebyfile

So, the functionality where you get imports from packages you've already imported even if they're not in your package.json dependencies is also preserved from Strada, so you'll get auto-imports for any of these things that are already present:

image

Strada also seems to give you imports from the ATA cache, which at least for me, in Herebyfile, includes some pretty random stuff (like React) that I think shouldn't be there. I think it may be including everything in the global typings cache rather than just what Herebyfile is actually pulling on? This was the case for both Strada and Corsa, so I actually just filtered those out in Corsa for now.

Including devDependencies also sounds like something that would be reasonable to expose as an option, especially considering you could combine that with targeted exclusions.

jakebailey:
Tried this out, it's very fast!

In `DefinitelyTyped-tools`'s `packages/eslint-plugin/test/eslint.test.ts`, I commented out the `globSync` import; when I then tried to auto import it again, `globSync` was not presented to me unless I had `glob` imported. I guess this is the devDep filter at work? The same happens to `jest-file-snapshot`. But, `main` also has this problem.

Commenting out `normalizeSlashes` nets me some extra import options that look undesirable:

image

`main` only has the first one.

In `packages/eslint-plugin/src/rules/no-import-of-dev-dependencies.ts`, I commented out `import { TSESTree } from "@typescript-eslint/utils";`; the auto import suggested both `@typescript-eslint/types` and `@typescript-eslint/utils`, which I think could be correct. Interestingly, `main`, only suggests `/types`, which is worse.

andrewbranch:
I’ll check out the example in the screenshot. The good one is probably from discovering the symlink in node_modules, while the bad ones probably originated from the realpath file in the program. (I’m not sure why the one with src is duplicated though, that’s kinda weird.) Interestingly, I think the solution is disable the newly lit up symlink cache in module specifier generation for program-sourced files 🙃 

> Interestingly, main, only suggests /types, which is worse.

I forgot to mention this behavior difference I intentionally made because it’s pretty obscure and in the weeds. In completions, we typically try to deduplicate re-exports of the same symbol and just pick the best one, mostly based on module specifier. When two module specifiers tie in ranking, like `@typescript-eslint/utils` and `@typescript-eslint/types`, we would fall back to program order, which IIRC ended up preferring the least deep file, or in other words the most-re-exported symbol. Since we no longer have program order as an input to node_modules files, I noticed some arbitrary swaps of these kinds of ties in test failures in the new system, so I changed the rules slightly:

- re-exports from different package names are never deduped, always treated as separate
- for re-exports from the same package (or no package, as local project files), we return multiple completions in the case where there's a tie for the winner, rather than falling back to depth or another totally arbitrary measure

andrewbranch:
Hm, ok, yeah, symlinked monorepo packages are going to take more thought than I previously thought. I guess the good news is that the current failure mode is that you might get too many suggestions. The bad news is that the ones that come from node_modules might get stale as you edit your local packages, since we won't figure out from the watch event on the realpath that something stored in the node_modules bucket changed. Getting my thoughts out in writing since I may not have time to finish this out before I’m gone for the holidays:

The current way we include node_modules packages that would normally be filtered out by your package.json dependencies, but are directly imported by your program files, is when we iterate over program files during project bucket building, if the file is in node_modules, we only include it if the node_modules bucket corresponding to the node_modules directory the file is in is _not_ going to include it (we've already computed the set of dependency names for any node_modules buckets that are being built concurrently). If the node_modules bucket doesn't have that package in its dependencies, we just process the file as part of the project bucket. This has pros and cons:

* Pro: the dependency that should arguably be filtered out is only offered in the project that's already importing it; it doesn't "leak" into any other projects you may have open.
* Con: only the files that are already in the program become available for auto-import. If the package has a bunch of different entrypoints, e.g. like an icon library with a file per icon, importing "cool-icons/react/chevron-left" isn't going to make "cool-icons/react/chevron-right" appear, which is pretty annoying.
* Con: since we didn't discover the file by entrypoint searching its package.json, it doesn't get the new fast pre-computed module specifier. It goes through the old module specifier generation code. Not a very big deal, but it makes me slightly nervous that the code paths are pretty different.

I think it might be a better idea to invert the current logic, and instead of precomputing the node_modules bucket dependencies so we can add extra node_modules files to project buckets, we should precompute the node_modules packages referenced by active programs and add those to the dependency list for those node_modules buckets. I think that will make everything else more consistent, and node_modules files will truly always be in node_modules buckets. The dependency computation is slightly more complicated, but there's not this weird exception to the rule of where exports are indexed.

Either way, when we go through the program files to find resolved node_modules packages, we need to check to see if non-node_modules file names are actually the realpath of a node_modules package so we can know to let the node_modules buckets handle them. Additionally, when we handle a file change event for one of those files, we need to mark the node_modules bucket dirty, not the project bucket dirty.

alexandr2110pro:
Hey there. Reading the code and discussions here is a pleasure.

While you're working on this, maybe you can also check [this little repo](https://github.com/alexandr2110pro/zed-playground-eslint) too.

It is a standard Nx ts-project-reference-based monorepo (this is their current default mode. Previously it was based on `compilerOptions.paths`) with nextjs and a bunch of example libs.

I've created it to explore how Zed's eslint and TS lsp integrations work, and it has troubles with autocompletions (and autoimport actions) of members exported from local packages. By looking at the vtsls logs, it doesn't return the completions, so it is probably not the sole zed's fault.
In my experience, vscode struggles with this too, although it seems to work a bit better. 

A little insight - if you have monorepo package A with files fooA.ts and barA.ts and B with files fooB.ts and barB.ts 
then, for all files in package A will not receive any suggestions for members of package B... Until you manually add import from B into any of files of A: eg import fooB into fooA => then both fooB.ts and barB.ts are present in completions in both fooA.ts and barA.ts.

If this problem has roots in the tsserver implementation, you could fix it the new implementation and drastically improve DX for large Nx monorepos.

jansedlon:
This looks solid. I'm wondering why is the difference in subsequent completion so small between TS and Go implementation. Do you think the gap widens with project size?

rubnogueira:
Following this. I was using TypeScript 0.20251205.1 and I bumped to the latest master version (I saw all your commits), and unfortunately a lot of auto-imports and related features in our pnpm mono/polyrepo stopped working, and I had to revert it. 

Looking at this PR, I believe you fixed a lot of my issues, but unfortunately I've been very busy these past days, but I'm really excited to test it and give some feedback.

andrewbranch:
> I'm wondering why is the difference in subsequent completion so small between TS and Go implementation. Do you think the gap widens with project size?

I’m not sure. On one hand, I expect Go to be a lot faster than TS, but looking at the absolute numbers, I just kept being impressed how fast the TS-based completions actually are (when it's not doing work that could have been cached). I think there are still some opportunities to optimize non-auto-import completions in the Go code, but for the auto-import path, I've removed all the hot spots in the profiles until I can't see how any part of it could get faster.

internal/collections/set.go
```diff
@@ -68,6 +111,33 @@ func (s *Set[T]) Equals(other *Set[T]) bool {
 	return maps.Equal(s.M, other.M)
 }
 
+func (s *Set[T]) IsSubsetOf(other *Set[T]) bool {
+	if s == nil {
+		return true
+	}
+	if other == nil {
+		return false
```
gabritto: It probably doesn't matter, but if `s` is empty, shouldn't this return `true` in this case?

gabritto: Or maybe `s.M` is never empty? But if it can be, I have the same question for `Intersects` below: what happens if `s.M` is empty?

internal/fourslash/_scripts/failingTests.txt
gabritto: The tests that differ in a good way should be updated then and not show up here, ideally.

internal/fourslash/tests/gen/completionListWithLabel_test.go
gabritto, line 11: Why is this being skipped?

testdata/baselines/reference/fourslash/autoImports/autoImportCrossProjectNodeModules.baseline.md
```diff
@@ -0,0 +1,10 @@
+// === Auto Imports === 
+```ts
+// @FileName: /project-b/index.ts
+pkg_/**/
+``````ts
```
gabritto, line 5: I think auto-imports baseline is missing a newline here? Should be \```\n``` ts

internal/fourslash/fourslash.go
```diff
@@ -928,6 +928,7 @@ type MarkerInput = any
 // !!! user preferences param
 // !!! completion context param
 func (f *FourslashTest) VerifyCompletions(t *testing.T, markerInput MarkerInput, expected *CompletionsExpectedList) VerifyCompletionsResult {
+	t.Helper()
```
gabritto, line 931: Why does this need to be marked helper?

internal/project/session.go
```diff
@@ -514,6 +530,22 @@ func (s *Session) GetSnapshotLoadingProjectTree(
 	return snapshot
 }
 
+// GetLanguageServiceWithAutoImports clones the current snapshot with a request to
+// prepare auto-imports for the given URI, then returns a LanguageService for the
+// default project of that URI. It should only be called after GetLanguageService.
+// !!! take snapshot that GetLanguageService initially returned
```
gabritto, line 536: Will we actually need to take the original LS snapshot? At least from the current usage of `GetLanguageServiceWithAutoImports`, the old snapshot is only used to determine whether we need autoimports, so I don't see why we'd need to worry about using a possibly newer snapshot...

testdata/baselines/reference/fourslash/autoImports/autoImportCrossProjectNodeModules.baseline.md
andrewbranch: Yes, but it was always like this, so I don't want to change every baseline in this PR.

internal/fourslash/fourslash.go
andrewbranch: When an assertion in this function fails, it makes the test runner report the line/column of the VerifyCompletions call that failed instead of the line of the assertion inside fourslash.go. If there are multiple VerifyCompletions calls in a single test, you can't tell which one failed without this.

internal/project/session.go
andrewbranch: I forgot to revisit this TODO. The completions request was made against a certain document version, and when we come back to here, we are continuing that request. If the document changed in the meantime, the document position might not make sense anymore. I think in theory this is important, but in practice, there are only a couple ms during which things could have changed, and if a change did sneak into that window, it's probably a single character insertion, which wouldn't change things much. It would add a little complexity to handle the theoretical edge cases correctly, so I'm on the fence about whether to address it now. I'll try to take a look and see how bad it is.

gabritto: Ah, I see. Yeah, I remember I had to do some handling of this in #1871, so it may be worth fixing.

internal/fourslash/tests/gen/completionListWithLabel_test.go
andrewbranch: It's not being skipped; it was moved to manual because of an arbitrary sorting change.

internal/collections/set.go
andrewbranch: Good catch on this one. I think `Intersects` is correct. You’re allowed to range over a nil map, so if `s` and `other` are non-nil but `s.M` is nil or empty, you'll fall out the bottom `return false`. I think the definition of `Intersects` should be that the intersection is non-empty, so `s.Intersects(other)` when both are empty is still false.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -11,34 +11,21 @@ TestAutoImportCompletionExportListAugmentation1
...
-TestAutoImportCrossProject_symlinks_toSrc
 TestAutoImportFileExcludePatterns2
 TestAutoImportFileExcludePatterns3
 TestAutoImportJsDocImport1
 TestAutoImportModuleNone1
 TestAutoImportNodeModuleSymlinkRenamed
 TestAutoImportNodeNextJSRequire
 TestAutoImportPackageJsonImportsCaseSensitivity
-TestAutoImportProvider_exportMap1
+TestAutoImportPackageRootPath
```
andrewbranch, line 21: This one is failing because of the note I made about not searching node_modules for untyped JS. We might improve that later so I'll leave it marked failing here.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -11,34 +11,21 @@ TestAutoImportCompletionExportListAugmentation1
...
+TestAutoImportPathsNodeModules
```
andrewbranch, line 22: This failure represents a new edge case limitation that would incur a performance penalty to fix, but can be worked around with user preferences. The test asserts that auto-imports respects tsconfig.json `paths` as an override for a node_modules package’s module specifier. Now that node_modules specifiers are computed outside the context of a program and compiler options, we short circuit long before the code that comes up with the paths-based specifier. I don't want to support this and make everyone who uses `paths` pay the cost of us double checking every node_modules specifier against their `paths`, but a user could make it work by excluding the node_modules specifier in the `autoImportSpecifierExcludeRegexes` user preference, which would make us fall back to trying the older slower specifier generation that takes `paths` into account.

I opted to leave "wontfix" issues like this as failing rather than move them to manual and reverse the assertion.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -11,34 +11,21 @@ TestAutoImportCompletionExportListAugmentation1
...
 TestAutoImportProvider_exportMap9
 TestAutoImportProvider_globalTypingsCache
-TestAutoImportProvider_namespaceSameNameAsIntrinsic
-TestAutoImportProvider_pnpm
-TestAutoImportProvider_wildcardExports1
-TestAutoImportProvider_wildcardExports2
-TestAutoImportProvider_wildcardExports3
-TestAutoImportProvider4
 TestAutoImportProvider9
+TestAutoImportRelativePathToMonorepoPackage
```
andrewbranch, line 28: This is a wontfix. The test contains a file that imports from a monorepo package via relative path instead of via node_modules package specifier, and asserts that an import fix of a different file from the same monorepo package also gets written as a relative path, overriding default specifier generation ranking by detecting an unusual pattern that already exists in the file.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -132,6 +119,7 @@ TestCompletionListInNamedFunctionExpression
 TestCompletionListInNamedFunctionExpression1
 TestCompletionListInNamedFunctionExpressionWithShadowing
 TestCompletionListInScope
+TestCompletionListInScope_doesNotIncludeAugmentations
```
andrewbranch, line 122: I think I noted this difference in the description. We now offer auto-imports from module augmentations that don't resolve to anything. I think this has pretty neutral desirability. It will result in a checker error if you accept the import, but maybe it would be fixed after an `npm install`. It doesn't seem very important. I guess I could move this one to manual and reverse the assertion.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -177,19 +165,19 @@ TestCompletionsImport_default_anonymous
 TestCompletionsImport_details_withMisspelledName
 TestCompletionsImport_exportEquals_global
 TestCompletionsImport_filteredByInvalidPackageJson_direct
+TestCompletionsImport_filteredByPackageJson_ambient
```
andrewbranch: The behavior of this one improved. I'll move it.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -177,19 +165,19 @@ TestCompletionsImport_default_anonymous
...
 TestCompletionsImport_filteredByPackageJson_direct
 TestCompletionsImport_filteredByPackageJson_nested
 TestCompletionsImport_filteredByPackageJson_peerDependencies
 TestCompletionsImport_filteredByPackageJson_typesImplicit
 TestCompletionsImport_filteredByPackageJson_typesOnly
-TestCompletionsImport_jsxOpeningTagImportDefault
-TestCompletionsImport_mergedReExport
+TestCompletionsImport_importType
```
andrewbranch: I don't think I meant to skip this; it was unintentionally broken. Fixing.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -177,19 +165,19 @@ TestCompletionsImport_default_anonymous
...
 TestCompletionsImport_named_didNotExistBefore
+TestCompletionsImport_named_namespaceImportExists
```
andrewbranch: Same, for the same root cause as `_importType`. Fixing.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -344,6 +332,8 @@ TestImportNameCodeFix_trailingComma
 TestImportNameCodeFix_uriStyleNodeCoreModules2
 TestImportNameCodeFix_uriStyleNodeCoreModules3
 TestImportNameCodeFix_withJson
+TestImportNameCodeFixDefaultExport4
```
andrewbranch, line 332: Wontfix. Asserts that an `export default a` shows up as a fix for the identifier `foo` if the exporting file is named foo.ts. We will still use filename as a fallback, but not if the export has a local name. We used to jump through a lot of hoops to find every possible name. Now we stop collecting lower priority names once we have something.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -344,6 +332,8 @@ TestImportNameCodeFix_trailingComma
...
+TestImportNameCodeFixDefaultExport7
```
andrewbranch: Same. The original test made a similar assertion to the one mentioned above, but this one was only intended to repro a Strada crash. Actually, I'll move this to manual and assert no fixes to show it doesn't crash.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -362,6 +352,7 @@ TestImportNameCodeFixNewImportFileQuoteStyle2
 TestImportNameCodeFixNewImportFileQuoteStyleMixed0
 TestImportNameCodeFixNewImportFileQuoteStyleMixed1
 TestImportNameCodeFixNewImportTypeRoots1
+TestImportNameCodeFixOptionalImport0
```
andrewbranch: This one now offers an extra valid fix. I'll update and move to manual.

internal/ls/autoimport/fix.go
```diff
@@ -0,0 +1,1176 @@
...
+// Sorting with this comparator will place the best fix first.
+// Fixes of equal desirability will be considered equal.
+func (v *View) CompareFixesForRanking(a, b *Fix) int {
+	if res := compareFixKinds(a.Kind, b.Kind); res != 0 {
+		return res
+	}
+	return v.compareModuleSpecifiersForRanking(a, b)
+}
+
+func compareFixKinds(a, b lsproto.AutoImportFixKind) int {
```
jakebailey, line 896: This could just be `cmp.Compare` at the one call site, though I suppose there are other comparisons here anyhow

jansedlon:
@andrewbranch Sorry to bother you, I see that it has been released. I was eager to try that out but I don't see a difference in the autoimport speed. I didn't want to ping you here in case it hasn't been release but i can see in the vscode logs
```
Building autoimport registry
```

Which has been introduced in your PR so I guess I have it.

I just tried to import a single identifier in my (private) project. I tested both TS version and Go version.

This is what i found in the logs with the Go version

```
2026-01-11 18:47:14.403 [info] handled method 'textDocument/didChange' in 5.042µs
2026-01-11 18:47:14.403 [info] Processed 1 file changes in 57.792µs
2026-01-11 18:47:14.440 [info] handled method 'textDocument/didChange' in 1.916µs
2026-01-11 18:47:14.462 [info] handled method 'textDocument/documentSymbol' in 31.161292ms
2026-01-11 18:47:14.463 [info] ======== Cloning snapshot 63 ========
[18:47:14.403] Reason: RequestedLanguageService (pending file changes) -  Documents: [file:///Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/server/websockets-processor.ts]
[18:47:14.403] DidChangeFiles
	[18:47:14.403] Checking for changes affecting config files
		[18:47:14.403] Summarizing file changes
		[18:47:14.403] Checking if any changed files are config files
	[18:47:14.403] Marking project /Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/tsconfig.json as dirty due to changes in /users/jansedlon/developer/flixy.worktrees/webinars/apps/web/server/websockets-processor.ts
[18:47:14.403] DidRequestFile
	[18:47:14.403] Acquiring config for project
	[18:47:14.403] CompilerHost
	[18:47:14.458] Program update for /Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/tsconfig.json completed in 55.773417ms
[18:47:14.458] UpdateAutoImports
	[18:47:14.458] Building autoimport registry
		[18:47:14.459] Updated buckets and directories in 22.917µs
		[18:47:14.459] Built autoimport registry in 32.083µs
[18:47:14.459] Finished cloning snapshot 63 into snapshot 64 in 55.985666ms

2026-01-11 18:47:14.463 [info] 
======== Cache Statistics ========
2026-01-11 18:47:14.463 [info] Open file count:        2
2026-01-11 18:47:14.464 [info] Cached disk files:  13766
2026-01-11 18:47:14.464 [info] Project count:          1
2026-01-11 18:47:14.465 [info] Config count:           3
2026-01-11 18:47:14.465 [info] Parse cache size:            10941
2026-01-11 18:47:14.465 [info] Program count:                   1
2026-01-11 18:47:14.465 [info] Extended config cache size:      2
2026-01-11 18:47:14.465 [info] Auto Imports:
2026-01-11 18:47:14.465 [info] 	Project buckets:
2026-01-11 18:47:14.465 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars/apps/web/tsconfig.json (dirty):
2026-01-11 18:47:14.466 [info] 			Files: 1011
2026-01-11 18:47:14.466 [info] 			Exports: 4211
2026-01-11 18:47:14.466 [info] 	node_modules buckets:
2026-01-11 18:47:14.466 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars:
2026-01-11 18:47:14.466 [info] 			Files: 0
2026-01-11 18:47:14.466 [info] 			Exports: 6069
2026-01-11 18:47:14.467 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars/apps/web:
2026-01-11 18:47:14.467 [info] 			Files: 0
2026-01-11 18:47:14.467 [info] 			Exports: 46223
2026-01-11 18:47:14.467 [info] 
2026-01-11 18:47:14.467 [info] Updated watches in 6.208µs
2026-01-11 18:47:14.467 [info] Processed 1 file changes in 20.458µs
2026-01-11 18:47:14.472 [info] handled method 'textDocument/inlayHint' in 43.601959ms
2026-01-11 18:47:14.475 [info] handled method 'textDocument/diagnostic' in 71.877875ms
2026-01-11 18:47:14.510 [info] ======== Cloning snapshot 64 ========
[18:47:14.465] Reason: RequestedLanguageService (pending file changes) -  Documents: [file:///Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/server/websockets-processor.ts]
[18:47:14.465] DidChangeFiles
	[18:47:14.465] Checking for changes affecting config files
		[18:47:14.465] Summarizing file changes
		[18:47:14.465] Checking if any changed files are config files
	[18:47:14.465] Marking project /Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/tsconfig.json as dirty due to changes in /users/jansedlon/developer/flixy.worktrees/webinars/apps/web/server/websockets-processor.ts
[18:47:14.465] DidRequestFile
	[18:47:14.465] Acquiring config for project
	[18:47:14.465] CompilerHost
	[18:47:14.505] Program update for /Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/tsconfig.json completed in 40.153083ms
[18:47:14.505] UpdateAutoImports
	[18:47:14.505] Building autoimport registry
		[18:47:14.505] Updated buckets and directories in 17.25µs
		[18:47:14.505] Built autoimport registry in 24.625µs
[18:47:14.505] Finished cloning snapshot 64 into snapshot 65 in 40.29075ms

2026-01-11 18:47:14.511 [info] 
======== Cache Statistics ========
2026-01-11 18:47:14.511 [info] Open file count:        2
2026-01-11 18:47:14.511 [info] Cached disk files:  13766
2026-01-11 18:47:14.511 [info] Project count:          1
2026-01-11 18:47:14.511 [info] Config count:           3
2026-01-11 18:47:14.511 [info] Parse cache size:            10941
2026-01-11 18:47:14.512 [info] Program count:                   1
2026-01-11 18:47:14.512 [info] Extended config cache size:      2
2026-01-11 18:47:14.512 [info] Auto Imports:
2026-01-11 18:47:14.512 [info] 	Project buckets:
2026-01-11 18:47:14.512 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars/apps/web/tsconfig.json (dirty):
2026-01-11 18:47:14.512 [info] 			Files: 1011
2026-01-11 18:47:14.512 [info] 			Exports: 4211
2026-01-11 18:47:14.512 [info] 	node_modules buckets:
2026-01-11 18:47:14.512 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars:
2026-01-11 18:47:14.512 [info] 			Files: 0
2026-01-11 18:47:14.512 [info] 			Exports: 6069
2026-01-11 18:47:14.512 [info] 		/users/jansedlon/developer/flixy.worktrees/webinars/apps/web:
2026-01-11 18:47:14.512 [info] 			Files: 0
2026-01-11 18:47:14.512 [info] 			Exports: 46223
2026-01-11 18:47:14.512 [info] 
2026-01-11 18:47:14.512 [info] Updated watches in 6.667µs
2026-01-11 18:47:14.520 [info] handled method 'textDocument/inlayHint' in 54.423542ms
2026-01-11 18:47:14.567 [info] handled method 'textDocument/didChange' in 1.542µs
2026-01-11 18:47:14.589 [info] handled method 'textDocument/didChange' in 1.583µs
2026-01-11 18:47:14.615 [info] Processed 2 file changes in 55.792µs
2026-01-11 18:47:14.641 [info] handled method 'textDocument/didChange' in 2.167µs
2026-01-11 18:47:14.663 [info] Processed 1 file changes in 43.416µs
// ...
//  Repeating multiple times ...

2026-01-11 18:47:20.122 [info] handled method 'textDocument/completion' in 5.667879291s
```

So it takes about 5.6s to complete completion.

Now the TS version

```
Info 4687 [19:24:50.177] collectAutoImports: resolved 628 module specifiers, plus 1848 ambient and 1739 from cache
Info 4688 [19:24:50.177] collectAutoImports: response is complete
Info 4689 [19:24:50.177] collectAutoImports: 619.9200000000001
Info 4690 [19:24:50.177] getCompletionData: Semantic work: 1146.5165420000012
Info 4691 [19:24:50.198] getCompletionsAtPosition: getCompletionEntriesFromSymbols: 21.33741700000064
Perf 4692 [19:24:50.199] 19::completionInfo: elapsed time (in milliseconds) 1300.9486
Info 4693 [19:24:50.202] response:
    {"seq":0,"type":"response","command":"completionInfo","request_seq":19,"success":true,"performanceData":{...}
Info 4694 [19:24:50.213] request:
    {
      "seq": 20,
      "type": "request",
      "command": "updateOpen",
      "arguments": {
        "changedFiles": [
          {
            "fileName": "/Users/jansedlon/Developer/flixy.worktrees/webinars/apps/web/server/websockets-processor.ts",
            "textChanges": [
              {
                "newText": "n",
                "start": {
                  "line": 21,
                  "offset": 2
                },
                "end": {
                  "line": 21,
                  "offset": 2
                }
              }
            ]
          }
        ],
        "closedFiles": [],
        "openFiles": []
      }
    }
    ```

**Comment by rubnogueira** [comment]
@jansedlon I had some issues with auto-import on the previous versions, and I was having the same issues as you yesterday when I first upgraded.

But as of today, it's working properly for me and way faster. I don't have a conclusion as to why that happened, but I will update you if I find something relevant.

@andrewbranch Just out of curiosity, when I press ctrl+cmd (even 2/3 consecutive times), it's taking around 3 to 4 seconds for the dropdown to show up. I wonder if it's expected behavior (with this new cache infrastructure) or if I have something happening on my side.

Autocomplete and auto-import are super fast now, but it's just this initial slowdown.

```
2026-01-12 11:04:25.069 [info] Updated watches in 77.125µs
2026-01-12 11:04:26.323 [info] handled method 'textDocument/hover' in 46.041µs
2026-01-12 11:04:28.195 [info] handled method 'textDocument/completion' in 3.514961541s
2026-01-12 11:04:31.240 [info] handled method 'completionItem/resolve' in 3.033017083s
2026-01-12 11:04:32.307 [info] handled method 'textDocument/completion' in 2.97733275s
2026-01-12 11:04:35.967 [info] handled method 'completionItem/resolve' in 3.641896959s
2026-01-12 11:04:39.861 [info] handled method 'textDocument/hover' in 2.024584ms
2026-01-12 11:04:40.417 [info] handled method 'textDocument/hover' in 51.25µs
2026-01-12 11:04:44.047 [info] handled method 'textDocument/completion' in 3.76113425s
2026-01-12 11:04:46.990 [info] handled method 'completionItem/resolve' in 2.924262208s
```

0.20260111.1

edit: it varies a lot depending on the file. `handled method 'textDocument/completion' in 9.921624417s` but once done, autocomplete is super fast, but if I try to open again, it takes the same time.

jansedlon:
@rubnogueira Interestering, i completely restarted vscode, updated from 0.20260111.1 to 0.20260112 and i'm now encountering https://github.com/microsoft/typescript-go/issues/2467 so something has definitely changed. Will report when it's resolved.

andrewbranch:
@rubnogueira @jansedlon in tonight's nightly, #2467 will be fixed and #2474 will allow you to run "Developer: TypeScript Native Preview: Start CPU Profile" from the VS Code command palette. I would love to see a profile of some of these multi-second completion times. You can email me at [firstname].[lastname]@microsoft.com. Thanks for trying this out!

jansedlon:
@andrewbranch  Thank you for fixing this 🙏 I have contacted you on email

PR #2396, opened 2025-12-16
Use non-diagnostic reporting symbol lookups for API uses of getResolvedSymbol 

### LINKED ISSUE #2380: Explosion in emit time on VS Code
With https://github.com/microsoft/typescript-go/pull/2378, we can compile VS Code again without `--noEmit`, but we now have a dramatic jump in build time. It's hard for me to post all the details, but here's `--extendedDiagnostics` with `--noCheck`:

```
Files:              5386
Lines:           1741915
Identifiers:     2870355
Symbols:         4285627
Types:            962343
Instantiations:  1598817
Memory used:    2784695K
Memory allocs:  30215701
Config time:      0.158s
Parse time:       0.760s
Bind time:        0.182s
Check time:       0.003s
Emit time:       74.479s <- ⚠️
Total time:      75.624s
```

and here's the pprof graph view:

Image

weswigham:
…to avoid endlessly repeating diagnostic work we never need in checkers we're only using for emit.

Fixes #2380.

Brings `vscode` `noCheck` `incremental: false` emit time down to 3.8s on my machine, from 41s.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -13450,6 +13450,22 @@ func (c *Checker) getResolvedSymbolOrNil(node *ast.Node) *ast.Symbol {
 	return c.symbolNodeLinks.Get(node).resolvedSymbol
 }
 
+func (c *Checker) getResolvedSymbolNoDiagnostics(node *ast.Node) *ast.Symbol {
+	links := c.symbolNodeLinks.Get(node)
+	if links.resolvedSymbol != nil {
+		return links.resolvedSymbol
+	}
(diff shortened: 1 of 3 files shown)


weswigham:
I still feel like 3.8s is too long for native-code based emit (even with how much code vscode is), so there's probably still optimizations to be made (there's probably more than a few things that ported with subtly poor perf); but I'm probably gonna focus on the remaining emit features before I get back to checking in on it so long as it's acceptable with just this obvious fix for now.

PR #2409, opened 2025-12-18
Enhance error handling in readLoop

youngxhui:
Add recover at goroutine call sites to prevent panics inside goroutines from affecting the main process; also add assertion checks for type conversions to avoid panics caused by failed assertions


jakebailey:
Why add this? When does `readLoop` fail? When would a cancel request's params not be the expected kind?

PR #2429, opened 2025-12-27
fix(2426): goToDefinition for pattern ambient module no longer goes to implementation 

### LINKED ISSUE #2426: goToDefinition for pattern ambient module no longer goes to implementation
After switching to `tsgo`, command-clicking a `.css` import no longer opens the `.css` file. It just opens the pattern ambient module declaration.

## Steps to reproduce

I'm pretty sure this test suite diff is already reproducing this issue:

https://github.com/microsoft/typescript-go/blob/d9178cc1fef3cedc3c1a48a652e63dd83310ea20/testdata/baselines/reference/submodule/fourslash/goToDefinition/goToDefinitionCSSPatternAmbientModule.baseline.jsonc.diff#L1-L10

Go-to-Definition = Go-to-Implementation 👍 

## Behavior with `tsgo`

Go-to-Definition = Go-to-Type-Definition 👎

a-tarasyuk:
Fixes #2426


internal/ls/definition.go
```diff
@@ -121,13 +129,34 @@ type fileRange struct {
...
+		targetRange := lsproto.Range{
+			Start: lsproto.Position{
+				Line:      0,
+				Character: 0,
+			},
+			End: lsproto.Position{
+				Line:      0,
+				Character: 0,
+			},
+		}
```
DanielRosenwasser, line 151: To minimize some of the diffs, can you do something like this?

```
l.converters.ToLSPRange(reference.file, reference.file.Loc)
```

(not sure if `file` might be `nil` though)

a-tarasyuk: @DanielRosenwasser thanks for the feedback. The file can be `nil`., even if it exists, this case is handled using a range with zero positions

https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/src/services/goToDefinition.ts#L123-L127

https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/src/services/goToDefinition.ts#L764-L774

if the file is used for the range, the selection will look like this

```
-// [||] export class Foo {}
+// [| export class Foo {}|]
```

Several differences in this PR are due to handling whitespace at the start of the file, for example

https://github.com/microsoft/typescript-go/pull/2429/files#diff-f36f4a55771065e64894e0d32a5c744439364e2a7804d6267af778e0d4ac09ddR3

https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/tests/cases/fourslash/goToDefinitionExternalModuleName.ts#L8

https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/goToDefinitionExternalModuleName.baseline.jsonc#L3

Before
```
// [||]export class Foo {}
```

After
```
//[||] export class Foo {}
```

DanielRosenwasser: > Several differences in this PR are due to handling whitespace at the start of the file, for example

In that case, `createLspRangeFromNode` or `createLspRangeFromRange` should take care of those differences.

a-tarasyuk: These diffs are related to missing [chompLeadingSpace](https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/src/harness/fourslashImpl.ts#L4977) from the test parser. I’m not sure whether this behavior should be preserved.

a-tarasyuk: There are a couple of failing formatting tests, but they’re all in `gen` folder, which shouldn’t be modified directly. We probably need to regenerate the tests for these cases. For example, Strada behaviour for some of them,

this test has 4 spaces after `///`
https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/tests/cases/fourslash/semicolonFormattingInsideAStringLiteral.ts#L3

but the behavior is validated with 3 spaces
https://github.com/microsoft/TypeScript/blob/f3770c9b89ba7969b163c5cea7805d5df23b2086/tests/cases/fourslash/semicolonFormattingInsideAStringLiteral.ts#L9

Anyway, the remaining question is whether we should proceed with or without the `chompLeadingSpace` helper.

jakebailey: If you want to fix the generated tests, you should be able to do something like `npm run makemanual semicolonFormattingInsideAStringLiteral` and then fix the test.

a-tarasyuk: @jakebailey, thanks for pointing that out. I've updated the failing tests to align with Strada.

internal/fourslash/_scripts/manualTests.txt
jakebailey, line 1: Did the script sort this or something?

jakebailey:
PR is failing because the failing tests need to be updated

internal/fourslash/tests/manual/semicolonFormattingInsideAStringLiteral_test.go
```diff
@@ -11,7 +11,7 @@ func TestSemicolonFormattingInsideAStringLiteral(t *testing.T) {
 	fourslash.SkipIfFailing(t)
```
jakebailey: `SkipIfFailing` should be removed from the manual tests. The script unfortunately does not do this

internal/fourslash/_scripts/manualTests.txt
a-tarasyuk: I ran the script as-is, without any manual modifications.

internal/ls/definition.go
```diff
@@ -121,13 +129,34 @@ type fileRange struct {
...
+		locations = append(locations, &lsproto.LocationLink{
+			OriginSelectionRange: originSelectionRange,
+			TargetUri:            lsconv.FileNameToDocumentURI(reference.fileName),
```
gabritto, line 154: Could this ever need to go through the source mapper?

PR #2432, opened 2025-12-30
Include import specifiers as references in code lens

### LINKED ISSUE #2367: References code lens does not include imports as references
For a project:

```ts
// abc.ts
export function abc() { }
```

```ts
// other.ts
import { abc } from "./abc";
```

The references code lens currently shows `0` references

I can understand why it's doing this but it also feels a bit unexpected. TS 6.0 includes the imports as references

darwin808:
## Summary
- Fixes issue where import specifiers were not counted as references in code lens
- For `export function foo()` and `import { foo } from "./a"`, the import `foo` is now properly counted as a reference
- Aligns behavior with TS 6.0

## Changes
- Added import specifiers to `singleReferences` in `getSearchesFromDirectImports` (`importTracker.go`)
- Export specifiers are excluded since they're already handled by `getReferencesAtExportSpecifier`
- Added deduplication in `shouldAddSingleReference` to prevent duplicate entries when traversing import/export chains
- Updated duplicate detection logic in `getReferencesAtLocation` to handle simple import specifiers (not just `import { foo as bar }` patterns)

## Test plan
- [x] All existing code lens tests pass with updated baselines
- [x] All findAllReferences tests pass with updated baselines
- [x] Verified `foo` now shows 4 references (was 3) in `codeLensFunctionsAndConstants01` test

Fixes #2367

### DIFF
--- a/testdata/baselines/reference/fourslash/findAllReferences/findAllRefsOfConstructor_multipleFiles.baseline.jsonc
+++ b/testdata/baselines/reference/fourslash/findAllReferences/findAllRefsOfConstructor_multipleFiles.baseline.jsonc
@@ -6,7 +6,7 @@
 // export { [|B|] as [|B1|] } from "./f";
 
 // === /b.ts ===
-// import B, { B1 } from "./a";
+// import B, { [|B1|] } from "./a";
 // const d = new [|B|]("b");
 // const d1 = new [|B1|]("b1");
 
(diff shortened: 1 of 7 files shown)


darwin808:

testdata/baselines/reference/fourslash/findAllReferences/findAllRefsOfConstructor_multipleFiles.baseline.jsonc
```diff
@@ -6,7 +6,7 @@
 // export { [|B|] as [|B1|] } from "./f";
 
 // === /b.ts ===
-// import B, { B1 } from "./a";
+// import B, { [|B1|] } from "./a";
```
jakebailey, line 9: Seems wrong, given `_submodules/TypeScript/tests/baselines/reference/findAllRefsOfConstructor_multipleFiles.baseline.jsonc` does not include it

PR #2435, opened 2025-12-30
Support identifier location in inlay hints

gabritto:
In Strada, identifiers had symbols, and when we synthesized an identifier in a type node, we'd store a symbol in it, which we would later use in inlay hints to obtain a location for an identifier in the inlay hint label parts.
This PR adapts that for Corsa, but instead of attaching a symbol to an identifier, we now have to store this information in a separate map that is provided to the node builder implementation.

All inlay hint existing diffs are now good changes as far as I can tell. The newly accepted diffs are due to a few cases where Strada couldn't provide a symbol for an identifier but Corsa can.

I also had to add code to avoid converting bundled file names to URIs in the tests, as that would cause a crash, and I left a `!!!` for possibly avoiding type node reuse in inlay hints once that's implemented, because I think otherwise we can't easily keep track of the mapping of identifiers to symbols.


internal/ls/inlay_hints.go
```diff
@@ -310,20 +310,24 @@ func (s *inlayHintState) getParameterDeclarationTypeHints(symbol *ast.Symbol) *l
...
+		InlayHintLabelParts: ptrTo(s.getInlayHintLabelParts(typeNode, idToSymbol)),
 	}
 }
 
 func (s *inlayHintState) typePredicateToInlayHintParts(typePredicate *checker.TypePredicate) lsproto.StringOrInlayHintLabelParts {
 	flags := nodebuilder.FlagsIgnoreErrors | nodebuilder.FlagsAllowUniqueESSymbolType |
 		nodebuilder.FlagsUseAliasDefinedOutsideCurrentScope
-	typeNode := s.checker.TypePredicateToTypePredicateNode(typePredicate, nil /*enclosingDeclaration*/, flags)
+	idToSymbol := make(map[*ast.IdentifierNode]*ast.Symbol)
+	// !!! Avoid type node reuse so we collect identifier symbols.
```
weswigham, line 326: q: If we reused a node, wouldn't a `node.Symbol()` suffice for that node, inside `getInlayHintLabelParts`, rather than only looking for the (potentially not) synthesized node's symbol?

gabritto: I think we don't store symbols on identifiers anymore in Corsa, so `node.Symbol()` for an AST identifier would always return `nil`.

gabritto: We could try to obtain the symbol for an AST identifier via checker methods here though. I might consider that solution once we get to it, I don't know which option would be faster.

Issue #2399, opened 2025-12-16
Implement JSX tag-closing completions


navya9singh:
## Steps to reproduce
Create a file.jsx and add the following lines:
`MyComponent = () => <div`

## Behavior with `typescript@5.9`
On adding `>` a closing html tag `</div>` gets auto inserted
## Behavior with `tsgo`
 `</div>` does not get inserted

Issue #2369, opened 2025-12-12
Code Lens provider registration error after switching tsdk location


andrewbranch:
### Extension Version

0.20251212.1

### VS Code Version

1.108.0-insider (Universal)

### Operating system Version

macOS

### Steps to reproduce

1. Open a TS file in VS Code
2. In your user preferences JSON, add or remove the `typescript.native-preview.tsdk` option pointing to your local tsgo build to make VS Code switch between the local build and its bundled build
3. In the TS file, run the Restart Language Server command

### Issue

An error "command 'typescript.native-preview.codeLens.showLocations' already exists" appears, and then you’re left with two instances of the client running!

<img width="646" height="83" alt="Image" src="https://github.com/user-attachments/assets/abbc749f-439a-4fed-87d8-5234b1beb5d3" />

DanielRosenwasser:
I can fix this in this case, but the lifecycle around start/restart is a little weird because it doesn't dispose of the previous instance at all.

DanielRosenwasser:
And it's also a bit odd because `LanguageClient`s can't synchronously dispose/stop either...

jakebailey:
The command doesn't use the client at all. It should just be registered outside `Client`, unconditionally on extension activation.

DanielRosenwasser:
Sure, that one does and that's easy. I'm working on other stuff that the `Client` needs to manage though.

jakebailey:
All in all, I would personally say we should just require there only be one Client at a time, then use globals or something

rubenferreira97:
@andrewbranch Was this wrongly closed? I am using `7.0.0-dev.20260122.3` and suspect this issue might still be persisting.

When I set up `"typescript.native-preview.tsdk": "./node_modules/@typescript/native-preview"`, I get:

<img width="638" height="302" alt="Image" src="https://github.com/user-attachments/assets/0df595cd-fc58-4e39-bd29-7e6cfd3b06f7" />

When I use `Format Document With...` we can also see that multiple `Typescript (Native Preview)` are present:

<img width="928" height="194" alt="Image" src="https://github.com/user-attachments/assets/842e8e1f-e4b9-44d3-921e-6a5166d8c7eb" />

PR #2451, opened 2026-01-07
Port the `jsxClosingTag` command

DanielRosenwasser:
Fixes https://github.com/microsoft/typescript-go/issues/2369
Fixes https://github.com/microsoft/typescript-go/issues/2399

### DIFF
--- a/internal/ls/jsxclosingtag.go
+++ b/internal/ls/jsxclosingtag.go
@@ -9,13 +9,13 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
-func (l *LanguageService) ProvideClosingTagCompletion(ctx context.Context, params *lsproto.TextDocumentPositionParams) (*lsproto.CustomClosingTagCompletion, error) {
+func (l *LanguageService) ProvideClosingTagCompletion(ctx context.Context, params *lsproto.TextDocumentPositionParams) (lsproto.CustomClosingTagCompletionResponse, error) {
 	_, sourceFile := l.getProgramAndFile(params.TextDocument.Uri)
 	position := l.converters.LineAndCharacterToPosition(sourceFile, params.Position)
 
 	token := astnav.FindPrecedingToken(sourceFile, int(position))
 	if token == nil {
-		return nil, nil
+		return lsproto.CustomClosingTagCompletionResponse{}, nil
 	}
 
 	var element *ast.Node
@@ -27,11 +27,11 @@ func (l *LanguageService) ProvideClosingTagCompletion(ctx context.Context, param
 
 	if element != nil && isUnclosedTag(element.AsJsxElement()) {
 		tagNameNode := element.AsJsxElement().OpeningElement.TagName()
-		result := &lsproto.CustomClosingTagCompletion{
+		result := lsproto.CustomClosingTagCompletion{
 			// Slight divergence from Strada - we don't use the verbatim text from the opening tag.
-			NewText: strPtrTo("</" + ast.EntityNameToString(tagNameNode, scanner.GetTextOfNode) + ">"),
+			NewText: "</" + ast.EntityNameToString(tagNameNode, scanner.GetTextOfNode) + ">",
 		}
-		return result, nil
+		return lsproto.CustomClosingTagCompletionResponse{CustomClosingTagCompletion: &result}, nil
 	}
 
 	var fragment *ast.Node
@@ -42,13 +42,13 @@ func (l *LanguageService) ProvideClosingTagCompletion(ctx context.Context, param
 	}
 
 	if fragment != nil && isUnclosedFragment(fragment.AsJsxFragment()) {
-		result := &lsproto.CustomClosingTagCompletion{
-			NewText: strPtrTo("</>"),
+		result := lsproto.CustomClosingTagCompletion{
+			NewText: "</>",
 		}
-		return result, nil
+		return lsproto.CustomClosingTagCompletionResponse{CustomClosingTagCompletion: &result}, nil
 	}
 
-	return nil, nil
+	return lsproto.CustomClosingTagCompletionResponse{}, nil
 }
 
 func isUnclosedTag(node *ast.JsxElement) bool {
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -152,13 +152,12 @@ const customStructures: Structure[] = [
         ],
     },
     {
+        // Longer-term, we may just want to use TextEdit.
         name: "CustomClosingTagCompletion",
         properties: [
-            // TODO: TextEdit?
             {
                 name: "newText",
                 type: { kind: "base", name: "string" },
-                optional: true,
                 documentation: "The text to insert at the closing tag position.",
             },
         ],
@@ -222,7 +221,13 @@ const customRequests: Request[] = [
         method: "custom/textDocument/closingTagCompletion",
         typeName: "CustomClosingTagCompletionRequest",
         params: { kind: "reference", name: "TextDocumentPositionParams" },
-        result: { kind: "reference", name: "CustomClosingTagCompletion" },
+        result: {
+            kind: "or",
+            items: [
+                { kind: "reference", name: "CustomClosingTagCompletion" },
+                { kind: "base", name: "null" },
+            ],
+        },
         messageDirection: "clientToServer",
         documentation: "Request to get the closing tag completion at a given position.",
     },
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -2829,11 +2829,12 @@ func (f *FourslashTest) VerifyJsxClosingTag(t *testing.T, markersToNewText map[s
 			},
 			Position: f.currentCaretPosition,
 		}
-		result := sendRequest(t, f, lsproto.CustomTextDocumentClosingTagCompletionInfo, params)
+
+		requestResult := sendRequest(t, f, lsproto.CustomTextDocumentClosingTagCompletionInfo, params)
 
 		var actualText *string
-		if result != nil {
-			actualText = result.NewText
+		if closingTag := requestResult.CustomClosingTagCompletion; closingTag != nil {
+			actualText = &closingTag.NewText
 		}
 		assertDeepEqual(t, actualText, expectedText, f.getCurrentPositionPrefix()+"JSX closing tag text mismatch")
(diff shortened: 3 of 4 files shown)


internal/lsp/lsproto/customproto.go
```diff
@@ -0,0 +1,10 @@
+package lsproto
+
+type JSXClosingTagParams = TextDocumentPositionParams
```
jakebailey: I am shocked this works, because the request parsing relies on being able to know exactly which JSON objects are valid. I think this is only working becuase we are not roundtripping fourslash over JSON.

You should override the LSP generator to add in your custom stuff and the rest will fall out

internal/fourslash/tests/manual/autoCloseTag_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+// @Filename: /8.tsx
+const x = <div>
+    <div>/*8*/</div>
+</div>;
+
+// @Filename: /9.tsx
+const x = <p>
+    <div>
+        <div>/*9*/
```
DanielRosenwasser, line 57: This one isn't working - the call to FindPrecedingToken is coming up with a `>` (`KindGreaterThanToken`) with a parent of `JsxElement`. The code is only meant to operate on

* `>` with a `JsxOpeningElement` parent
* `JsxText` with a `JsxElement` parent

I haven't debugged in Strada, but I'm guessing there's something a little bit different about the way FindPrecedingToken works on JSX (and maybe specifically JSX text since that often needed to be specially-handled).

Maybe @andrewbranch has a clue?

internal/lsp/lsproto/customproto.go
DanielRosenwasser, line 1: This is a temporary file, I assume @jakebailey will kill me if I do this.

internal/ls/jsxclosingtag.go
```diff
@@ -0,0 +1,80 @@
...
+	if token.Kind == ast.KindGreaterThanToken && ast.IsJsxOpeningElement(token.Parent) {
+		element = token.Parent.Parent
+	} else if ast.IsJsxText(token) && ast.IsJsxElement(token.Parent) {
+		element = token.Parent
+	}
+
+	if element != nil && isUnclosedTag(element.AsJsxElement()) {
+		result := lsproto.JSXClosingTagResponse{
+			// TODO: Slight divergence from Strada - we don't use the verbatim text from the opening tag.
+			NewText: strPtrTo("</" + element.AsJsxElement().OpeningElement.TagName().Text() + ">"),
```
DanielRosenwasser: I could switch this back - but as it turns out, we have absolutely no tests on anything other than identifier tag names...

I actually don't think we should use the old behavior either, and `.Text()` works pretty nicely. I'll likely add a test here.

internal/fourslash/tests/manual/autoCloseTag_test.go
DanielRosenwasser: Also, we should probably have a test with `<div>>` or something. Funny enough, Strada completes it.

internal/lsp/lsproto/customproto.go
DanielRosenwasser: > I am shocked this works, because the request parsing relies on being able to know exactly which JSON objects are valid. 

I think it works because `JSXClosingTagParams` is just an alias anyway, right? So it's just piggy-backing on existing logic for `TextDocumentPositionParams`

jakebailey: No, you don't have the method in `unmarshalParams`, so it would not actually work in practice. It only works because the current transport used in the fourslash tester sends values as-is which retains their types.

jakebailey: Well, okay, I guess it's really just because you tricked it into using a totally different unused method with a potentially fine request, though the response is not as expected

internal/fourslash/tests/manual/autoCloseTag_test.go
DanielRosenwasser: Good news, it's a parser porting bug. https://github.com/microsoft/typescript-go/pull/2452

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -195,6 +208,18 @@ const customEnumerations: Enumeration[] = [
     },
 ];
 
+// Custom requests to add to the model (TypeScript-specific LSP extensions)
+const customRequests: Request[] = [
+    {
+        method: "textDocument/closingTagCompletion",
```
jakebailey: we'll want to name this something TS specific, like `typescript/closingTagCompletion`, since there's no guarantee that this name is not used or something

internal/ls/jsxclosingtag.go
DanielRosenwasser: Well of course

```
Error requesting closing tag completion: Error: InternalError: panic handling request custom/TypeScript/textDocument/closingTagCompletion: Unhandled case in Node.Text: *ast.PropertyAccessExpression
```

Probably not a bad idea to figure out why `JsxNamespacedName` is handled the way it is in `(*Node).Text()`

```go
return n.AsJsxNamespacedName().Namespace.Text() + ":" + n.AsJsxNamespacedName().name.Text()
```

jakebailey: Yeah, I was about to say, `Text()` won't work on tag names, we have helpers to deal with that we've had use in recent fixes for that kind of crash

DanielRosenwasser:
I am strongly considering just redoing this as a server -> client request.

____

Edit: I don't want to have to figure out the testing infrastructure for that at the moment. I think a custom command is acceptable for now.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -203,6 +216,18 @@ const customEnumerations: Enumeration[] = [
     },
 ];
 
+// Custom requests to add to the model (TypeScript-specific LSP extensions)
+const customRequests: Request[] = [
+    {
+        method: "custom/TypeScript/textDocument/closingTagCompletion",
```
jakebailey: Can we do `custom/`?

internal/lsp/server.go
```diff
@@ -503,7 +503,7 @@ func (s *Server) handleRequestOrNotification(ctx context.Context, req *lsproto.R
 		s.logger.Info("handled method '", req.Method, "' in ", time.Since(start))
 		return err
 	}
-	s.logger.Warn("unknown method", req.Method)
+	s.logger.Warn("unknown method ", req.Method)
```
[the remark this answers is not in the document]
DanielRosenwasser: This was intentional, but the suggestion isn't bad.

_extension/src/languageFeatures/util/dependentRegistration.ts
```diff
@@ -0,0 +1,83 @@
...
+    ) {
+        for (const condition of conditions) {
+            condition.onDidChange(() => this.update());
+        }
+        this.update();
+    }
+
+    public dispose() {
+        this.state?.registration?.dispose();
+        this.state = undefined;
```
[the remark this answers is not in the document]
DanielRosenwasser: @mjbvz original code in VS Code might have the same issue.

_extension/src/languageFeatures/util/dependentRegistration.ts
```diff
@@ -0,0 +1,83 @@
...
+        this._value = this.getValue();
+
+        onUpdate(() => {
+            const newValue = this.getValue();
+            if (newValue !== this._value) {
+                this._value = newValue;
+                this._onDidChange.fire();
+            }
+        });
+    }
```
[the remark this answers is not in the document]
DanielRosenwasser: onUpdate returns void.

_extension/src/languageFeatures/util/dependentRegistration.ts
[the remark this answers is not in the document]
DanielRosenwasser: Same with `onDidChangeSubscription`

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -151,6 +151,19 @@ const customStructures: Structure[] = [
             },
         ],
     },
+    {
+        name: "CustomClosingTagCompletion",
+        properties: [
+            // TODO: TextEdit?
```
DanielRosenwasser: I think in an official API this would be a `TextEdit`. If it was a server -> client event, it would possibly be an `AnnotatedTextEdit` to make it easier to test.

But I'm actually kind of fine with it as-is.

_extension/src/languageFeatures/tagClosing.ts
```diff
@@ -0,0 +1,184 @@
...
+
+            const params: TextDocumentPositionParams = {
+                textDocument: { uri: document.uri.toString() },
+                position: { line: position.line, character: position.character },
+            };
+            this.cancel = new vscode.CancellationTokenSource();
+
+            let response;
+            try {
+                response = await this.client.sendRequest<{ newText: string; } | null>(
```
DanielRosenwasser, line 108: I think the type is wrong?

```suggestion
                response = await this.client.sendRequest<{ newText?: string; }>(
```

Should we switch to `{ newText: string } | null`?

_extension/src/languageFeatures/tagClosing.ts
```diff
@@ -0,0 +1,184 @@
...
+                    : position;
+
+                activeEditor.insertSnippet(snippet, insertionPositions);
+            }
+        }, CLOSING_TAG_COMPLETION_DELAY);
+    }
+}
+
+function requireActiveDocumentSetting(languageConfigSection: "typescript" | "javascript", selector: vscode.DocumentSelector) {
+    return new Condition(
```
jakebailey, line 151: I know this is copy pasted but it sure does feel complicated to have all of these dynamic registrations, disposables, rather than just checking a boolean in the on change handler?

_extension/src/languageFeatures/tagClosing.ts
jakebailey: You did:

```
            {
                name: "newText",
                type: { kind: "base", name: "string" },
                optional: true,
                documentation: "The text to insert at the closing tag position.",
            },
```

So your suggestion would be correct, yeah.

DanielRosenwasser: I went the other way instead - the LS now responds with the expected type written here. It's more consistent with how other responses look, the naming is technically more clear (without the alias), and now there's less worrying about a pointer to a pointer.

Issue #1983, opened 2025-10-30
textDocument/diagnostic failed


lukpsaxo:
I seem to get this all the time:

<img width="493" height="108" alt="Image" src="https://github.com/user-attachments/assets/8d7f4991-8228-483b-845c-645cafdcda64" />

last time I checked the logs I couldn't find anything around the time I checked, but this time I see this panic (and this panic is in my logs several times a day)

Is the below panic expected or the cause of the above?

Logs:

```
======== Cloning snapshot 339 ========
[14:56:30.043] Reason: RequestedLanguageService (project dirty) - [file:///c%3A/Projects/x/y/z]
[14:56:30.043] DidRequestFile
	[14:56:30.043] Acquiring config for project
	[14:56:30.043] CompilerHost

panic: parse cache entry not found [recovered, repanicked]

goroutine 4680420 [running]:
github.com/microsoft/typescript-go/internal/project.(*Snapshot).Clone.func1()
	github.com/microsoft/typescript-go/internal/project/snapshot.go:174 +0x7d
panic({0x7ff6411298a0?, 0x7ff6414bd560?})
	runtime/panic.go:783 +0x132
github.com/microsoft/typescript-go/internal/project.(*ParseCache).Ref(0xc0daf614a0?, 0xc0132fd310?)
	github.com/microsoft/typescript-go/internal/project/parsecache.go:69 +0x194
github.com/microsoft/typescript-go/internal/project.(*Project).CreateProgram(0xc125859c00)
	github.com/microsoft/typescript-go/internal/project/project.go:293 +0x30e
github.com/microsoft/typescript-go/internal/project.(*projectCollectionBuilder).updateProgram.func1.2(0xc125859c00)
	github.com/microsoft/typescript-go/internal/project/projectcollectionbuilder.go:797 +0x1fd
github.com/microsoft/typescript-go/internal/project/dirty.(*SyncMapEntry[...]).changeLocked(0x7ff6414db520, 0xc11107ef40)
	github.com/microsoft/typescript-go/internal/project/dirty/syncmap.go:124 +0x18b
github.com/microsoft/typescript-go/internal/project/dirty.(*lockedEntry[...]).Change(...)
	github.com/microsoft/typescript-go/internal/project/dirty/syncmap.go:27
github.com/microsoft/typescript-go/internal/project.(*projectCollectionBuilder).updateProgram.func1({0x7ff6414c7f00, 0xc10ed3b5f0})
[...]

PR #2456, opened 2026-01-08
Fix LS bugs triggered by file moving (and probably other scenarios)

andrewbranch:
Fixes two bugs:

1. The more complicated fix and probably less commonly hit one involves moving a source file via the VS Code UI, which triggers a file open in the new location before a file create, which we weren't handling correctly. Test comments explain exactly what was happening in that sequence.
2. The second was a very simple fix to a condition that I messed up during a refactor long ago which has the potential to mess up default project lookups on any requests immediately following a configured project update that changes file inclusion. This incorrect file/project association will eventually correct itself after another file open, which is probably why these bugs have been hard to pin down a repro for in reports. I discovered this accidentally by moving a tsconfig.json file in the VS Code UI, but there are probably simpler and more common ways to trigger it. Hoping this fixes a lot of the "No project found for URI" messages.

### DIFF
--- a/internal/project/projectlifetime_test.go
+++ b/internal/project/projectlifetime_test.go
@@ -280,6 +280,7 @@ func TestProjectLifetime(t *testing.T) {
 
 		// Should now have one configured project only (file is now under src/)
 		_, err = session.GetLanguageService(context.Background(), srcIndexUri)
+		assert.NilError(t, err)
 		snapshot, release = session.Snapshot()
 		defer release()
 		assert.Equal(t, len(snapshot.ProjectCollection.Projects()), 1)


internal/project/projectcollectionbuilder.go
```diff
@@ -89,7 +89,7 @@ func (b *ProjectCollectionBuilder) Finalize(logger *logging.LogTree) (*ProjectCo
 		newProjectCollection.configuredProjects = configuredProjects
 	}
 
-	if !changed && !maps.Equal(b.fileDefaultProjects, b.base.fileDefaultProjects) {
+	if !maps.Equal(b.fileDefaultProjects, b.base.fileDefaultProjects) {
```
andrewbranch, line 92: This is the fix for the second bug. Everything else is handling the first.

internal/project/projectcollectionbuilder.go
```diff
@@ -349,6 +359,9 @@ func (b *ProjectCollectionBuilder) DidRequestFile(uri lsproto.DocumentUri, logge
 		if result := b.findDefaultProject(fileName, path); result != nil {
 			hasChanges = b.updateProgram(result, logger) || hasChanges
 			if result.Value() != nil {
+				if hasChanges {
+					b.cleanupInferredProject(logger)
```
andrewbranch, line 363: Updating (or possibly removing) the inferred project slightly more eagerly here was not strictly necessary for the bug fix, but I think it's correct/desirable. The file that got moved into a configured project via a watch change was staying in the inferred project as well, keeping it alive for longer than needed. It would have gotten cleaned up on the next file open.

internal/project/projectlifetime_test.go
```diff
@@ -217,4 +217,130 @@ func TestProjectLifetime(t *testing.T) {
...
+
+		// Simulate file move: create src/index.ts on disk
+		err := utils.FS().WriteFile("/home/projects/TS/p1/src/index.ts", files["/home/projects/TS/p1/index.ts"].(string), false)
+		assert.NilError(t, err)
+		err = utils.FS().Remove("/home/projects/TS/p1/index.ts")
+		assert.NilError(t, err)
+
+		// Simulate file move sequence as it would happen in an editor:
+		// 1. didOpen src/index.ts (new location)
+		// Open comes in before file create event, so the config file is not marked as needing a file name reload,
```
jakebailey, line 255: Totally weird that `Open` comes in before `create`, but I suppose that can make sense given they are different systems

jakebailey (approved):

internal/project/projectlifetime_test.go
andrewbranch: Yeah, that violated my original assumptions, but it can happen in another way too, where you have an editor tab open representing a recently deleted file, and then you restore it on disk while still open.

andrewbranch:
Confirmed this fixes my comment https://github.com/microsoft/typescript-go/issues/1983#issuecomment-3550064071 🎉

internal/project/configfileregistrybuilder.go
```diff
@@ -446,13 +446,25 @@ func (c *configFileRegistryBuilder) DidChangeFiles(summary FileChangeSummary, lo
...
+			var createdOpenFile bool
 			entry.ChangeIf(
 				func(config *configFileEntry) bool {
-					if config.commandLine == nil || config.rootFilesWatch == nil || config.pendingReload != PendingReloadNone {
+					if config.pendingReload != PendingReloadNone {
 						return false
 					}
 					logger.Logf("Checking if any of %d created files match root files for config %s", len(createdFiles), entry.Key())
 					for _, fileName := range createdFiles {
+						if _, ok := config.retainingOpenFiles[c.fs.toPath(fileName)]; ok {
```
sheetalkamat, line 457: Interesting that we are doing it this way. 
In strada we did this other way round - where on file open (if we didnt have infio about it / it didnt exist before) we would make sure that wild card watchers for the config are re-evaluated. As otherwise you would get incorrect project for the open file and wrong diagnostics. Dont we need that. or is the flicker so fast that it doesn't matter?
https://github.com/microsoft/TypeScript/blob/main/src/server/editorServices.ts#L4868

andrewbranch: I thought about this—and this conversation is about to be fragmented between here and https://github.com/microsoft/typescript-go/pull/2515#discussion_r2695686775 —and I'm really not sure what the right approach is. The problem is that VS Code has the ability to have files that are both open and deleted (they display with a strikethrough in the tab title). At first, I thought that if something is open, we should treat it as if it is definitely not deleted. But agent mode has a bad habit of keeping everything it ever made open in the background, even if it's been deleted. We have so far not plumbed overlays through the FS that gets read for config wildcards, so the status quo is that files have to actually exist to belong to a tsconfig. I ended up deciding that that's fine, but it's certainly possible to adopt a different model.

PR #2469, opened 2026-01-11
Update JsxOpeningElement type

ArnaudBarre:
Given it can only be the child of a JsxElement, I think it's a mistake to mark it as an Expression


andrewbranch:
It's `Expression` in Strada, but it ends up not mattering. I agree it doesn’t make sense to consider it an expression.

PR #2474, opened 2026-01-12
Add pprof commands to VS Code extension

andrewbranch:


internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -203,6 +203,74 @@ const customEnumerations: Enumeration[] = [
     },
 ];
 
+// Custom requests to add to the model (tsgo-specific)
+const customRequests: Request[] = [
+    {
+        method: "$/runGC",
```
jakebailey: I don't think `$/` is intended to be used like this; the spec says that requests made with `$/` prefixes are supposed to be rejected by a server: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests (and notifications are allowed to be ignored)

If we just need a custom prefix, we probably just want to align this PR and #2399?

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -203,6 +203,74 @@ const customEnumerations: Enumeration[] = [
     },
 ];
 
+// Custom requests to add to the model (tsgo-specific)
+const customRequests: Request[] = [
+    {
+        method: "custom/TypeScript/runGC",
```
jakebailey: Can we do `custom/`?

PR #2475, opened 2026-01-12
fix: use `enclosingDeclaration` for `resolveName` in `trackComputedName`

fubhy:
This changes the `resolveName` call in `trackComputedName` to use `enclosingDeclaration` as the location parameter, matching TypeScript's implementation from [checker.ts (line 8049)](https://github.com/microsoft/TypeScript/blob/9e8eaa1746b0d09c3cd29048126ef9cf24f29c03/src/compiler/checker.ts#L8409)


jakebailey:
The test in this PR passes even if I revert the checker change to `main`.

fubhy:
> The test in this PR passes even if I revert the checker change to `main`.

I did a lot of back and forth on this one. I can definitely confirm this fix works on the Effect example I linked whereas `main` errors on it. Let me fix the test though.

jakebailey:
Note that the original code says:

```ts
        function isAccessible(symbolFromSymbolTable: Symbol, resolvedAliasSymbol?: Symbol, ignoreQualification?: boolean) {
            return (symbol === (resolvedAliasSymbol || symbolFromSymbolTable) || getMergedSymbol(symbol) === getMergedSymbol(resolvedAliasSymbol || symbolFromSymbolTable)) &&
                // if the symbolFromSymbolTable is not external module (it could be if it was determined as ambient external module and would be in globals table)
                // and if symbolFromSymbolTable or alias resolution matches the symbol,
                // check the symbol can be qualified, it is only then this symbol is accessible
                !some(symbolFromSymbolTable.declarations, hasNonGlobalAugmentationExternalModuleSymbol) &&
                (ignoreQualification || canQualifySymbol(getMergedSymbol(symbolFromSymbolTable), meaning));
        }
```

So I would be very suspicious of any change that makes it different from that. I was hoping the test would let me see if making the func _closer_ to the original code was the real fix.

fubhy:
@jakebailey Looks like `getExportSymbolOfValueSymbolIfExported` is the key here.

jakebailey:
Yes, but that isn't mentioned in the original func that we are porting, no?

fubhy:
> Yes, but that isn't mentioned in the original func that we are porting, no?

EDIT: Ok, found the place where the code diverges: 

TypeScript: https://github.com/microsoft/TypeScript/blob/9e8eaa1746b0d09c3cd29048126ef9cf24f29c03/src/compiler/checker.ts#L8409

Go: https://github.com/microsoft/typescript-go/blob/931f1c4221bd5374c99f08b89dfa99e971c9f489/internal/checker/nodebuilderimpl.go#L2091

fubhy:
The corresponding TS code is this:

```typescript
        function trackComputedName(accessExpression: EntityNameOrEntityNameExpression, enclosingDeclaration: Node | undefined, context: NodeBuilderContext) {
            if (!context.tracker.canTrackSymbol) return;
            // get symbol of the first identifier of the entityName
            const firstIdentifier = getFirstIdentifier(accessExpression);
            const name = resolveName(enclosingDeclaration, firstIdentifier.escapedText, SymbolFlags.Value | SymbolFlags.ExportValue, /*nameNotFoundMessage*/ undefined, /*isUse*/ true);
            if (name) {
                context.tracker.trackSymbol(name, enclosingDeclaration, SymbolFlags.Value);
            }
            else {
                // Name does not resolve at target location, track symbol at dest location (should be inaccessible)
                const fallback = resolveName(firstIdentifier, firstIdentifier.escapedText, SymbolFlags.Value | SymbolFlags.ExportValue, /*nameNotFoundMessage*/ undefined, /*isUse*/ true);
                if (fallback) {
                    context.tracker.trackSymbol(fallback, enclosingDeclaration, SymbolFlags.Value);
                }
            }
        }
```

jakebailey:
Good catch; can you update the PR title/description? The test does work to show the issue, thanks!

fubhy:
> Good catch; can you update the PR title/description? The test does work to show the issue, thanks!

Sure. Will do. The updated test baseline files seem borked though.

jakebailey:
Yeah, something is regressing here, unfortunately.

jakebailey:
Ah, the func is missing the fallback case, no?

fubhy:
> Ah, the func is missing the fallback case, no?

Of course. Good catch ;-)

fubhy:
This should be it then. Thanks for the coaching!

PR #2479, opened 2026-01-12
Exhaustive case snippet completions

gabritto:
In this PR:
- Exhaustive case snippet completions
- Basic import adder logic needed for snippets

We're sending the additional text edits for exhaustive case completion items right away instead of only upon a resolve request. Seemed unnecessary to do extra work to support lazily resolving additional text edits in this case.

`exhaustiveCaseCompletions6_test.go` is currently being skipped because it needs user preferences to pass, I'll add that in my next PR and fix it.
Some now ported tests that are unrelated to exhaustive case snippet completions are failing. I have added them to the list of failing tests and will investigate later.

### DIFF
--- a/internal/ls/autoimport/import_adder.go
+++ b/internal/ls/autoimport/import_adder.go
@@ -53,9 +53,7 @@ func newImportsKey(moduleSpecifier string, topLevelTypeOnly bool) string {
 type importAdder struct {
 	// Context
 	ctx           context.Context
-	program       *compiler.Program
 	checker       *checker.Checker
-	file          *ast.SourceFile
 	view          *View
 	formatOptions *format.FormatCodeSettings
 	converters    *lsconv.Converters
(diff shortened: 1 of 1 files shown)


internal/fourslash/_scripts/makeManual.mts
andrewbranch, line 1: TIL this script exists, I've been doing it by hand

internal/ls/autoimport/import_adder.go
```diff
@@ -0,0 +1,487 @@
...
+type importAdder struct {
+	// Context
+	ctx           context.Context
+	program       *compiler.Program
+	checker       *checker.Checker
+	file          *ast.SourceFile
+	view          *View
+	formatOptions *format.FormatCodeSettings
+	converters    *lsconv.Converters
+	preferences   *lsutil.UserPreferences
```
andrewbranch: Nit: `View` stores most of these things already; prefer to access them from there

gabritto:
> Nice! I'm really pleased it seems like the new auto-imports infrastructure didn't make this too hard. I didn't look too closely at the completions implementation, but I assume that's was a pretty straightforward port. I don't have any concerns; did you have any questions or want me to look closer at anything in particular before I approve?

No, the completions code is a 1:1 port except we now send the edits eagerly.

andrewbranch (approved):

PR #2485, opened 2026-01-12
Fix race condition in ProjectCollection.ConfigFileRegistry

fubhy:
This attempts to tackle a race condition I ran into in CI multiple times whilst working on the last fix.


fubhy:
There seem to be more, unrelated race conditions e.g. https://github.com/microsoft/typescript-go/actions/runs/20938617746/job/60167089124 that the previous CI run ran into. That one appears to be a problem with the scope of the mutex in `flushChanges`.

fubhy:
> There seem to be more, unrelated race conditions e.g. https://github.com/microsoft/typescript-go/actions/runs/20938617746/job/60167089124 that the previous CI run ran into. That one appears to be a problem with the scope of the mutex in `flushChanges`.

Attempting to fix that here: https://github.com/microsoft/typescript-go/pull/2495

PR #2495, opened 2026-01-13
Fix race condition between flush and snapshot update

fubhy:
This should fix a race condition that was causing intermittent CI failures in `TestCompletionListCladule` (and likely other tests) when running with `-race`. (e.g. see https://github.com/microsoft/typescript-go/actions/runs/20938617746/job/60167089124)

The problem is that `getSnapshot` would flush pending changes and then call `UpdateSnapshot`, but another goroutine could slip in, see an empty pending changes queue, and return with a stale snapshot.

The fix adds a mutex (`snapshotUpdateMu`) that serializes the flush+update sequence. I considerered recycling the existing snapshot mutexes for this but couldn't see a way to do that ergonomically and idiomatically.

EDIT: It's difficult to reproduce but the CI failure in the run linked above basically shows the race condition in action https://github.com/microsoft/TypeScript/blob/d7cd405bb23eacda2833f3e7fe1be6d1b377f882/tests/cases/fourslash/completionListCladule.ts where the cursor seemingly ends up elsewhere (not at `.` after `c3) and then shows all global completions instead of the expected `getStuff` completion.

So basically: When completions are calculated using stale file content (without the `.`):
- The LSP position (line/character after `.`) maps to a different offset in the old file
- This position may fall at end of identifier `f` or elsewhere
- Triggers global scope completions instead of member access

### DIFF
--- a/internal/project/session.go
+++ b/internal/project/session.go
@@ -406,20 +406,20 @@ func (s *Session) getSnapshot(
 ) *Snapshot {
 	var snapshot *Snapshot
 	s.snapshotUpdateMu.Lock()
+	defer s.snapshotUpdateMu.Unlock()
+
 	fileChanges, overlays, ataChanges, newConfig := s.flushChanges(ctx)
 	updateSnapshot := !fileChanges.IsEmpty() || len(ataChanges) > 0 || newConfig != nil
 	if updateSnapshot {
 		// If there are pending file changes, we need to update the snapshot.
 		// Sending the requested URI ensures that the project for this URI is loaded.
-		snapshot = s.UpdateSnapshot(ctx, overlays, SnapshotChange{
+		return s.UpdateSnapshot(ctx, overlays, SnapshotChange{
 			reason:          UpdateReasonRequestedLanguageServicePendingChanges,
 			fileChanges:     fileChanges,
 			ataChanges:      ataChanges,
 			newConfig:       newConfig,
 			ResourceRequest: request,
 		})
-		s.snapshotUpdateMu.Unlock()
-		return snapshot
 	}
 	// If there are no pending file changes, we can try to use the current snapshot.
 	s.snapshotMu.RLock()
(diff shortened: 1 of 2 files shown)


internal/project/session.go
```diff
@@ -404,20 +405,22 @@ func (s *Session) getSnapshot(
...
 		// Sending the requested URI ensures that the project for this URI is loaded.
-		return s.UpdateSnapshot(ctx, overlays, SnapshotChange{
+		snapshot = s.UpdateSnapshot(ctx, overlays, SnapshotChange{
 			reason:          UpdateReasonRequestedLanguageServicePendingChanges,
 			fileChanges:     fileChanges,
 			ataChanges:      ataChanges,
 			newConfig:       newConfig,
 			ResourceRequest: request,
 		})
+		s.snapshotUpdateMu.Unlock()
```
DanielRosenwasser: This should probably follow the `Lock` via an immediate `defer s.snapshotUpdateMu.Unlock()`, right?

internal/project/session.go
```diff
@@ -457,6 +460,7 @@ func (s *Session) getSnapshot(
 			ResourceRequest: request,
 		})
 	}
+	s.snapshotUpdateMu.Unlock()
```
DanielRosenwasser: and doing that would allow you to delete this line too.

internal/project/session.go
```diff
@@ -100,8 +100,9 @@ type Session struct {
 	snapshotID atomic.Uint64
 
 	// snapshot is the current immutable state of all projects.
-	snapshot   *Snapshot
-	snapshotMu sync.RWMutex
+	snapshot         *Snapshot
+	snapshotMu       sync.RWMutex
+	snapshotUpdateMu sync.Mutex
```
DanielRosenwasser, line 105: I kind of wonder if it would be better to just stick with `snapshotMu` , make `UpdateSnapshot` not lock, and switch all uses of `snapshotUpdateMu` to `Lock()` (i.e. in write/exclusive mode).

Not 100% sure though.

fubhy: Wouldn't that result in us needlessly blocking all snapshot readers during the entire flush+update op?

andrewbranch (approved):
Great catch, thank you!

jakebailey (approved):

PR #2499, opened 2026-01-13
Fix potential deadlock in case of lsp shutdown

fubhy:
The LSP server could deadlock during shutdown when background tasks or handlers attempted to send messages (logs, diagnostics, responses) to the `outgoingQueue` after the `writeLoop` had exited. With no consumer draining the queue, these blocking channel sends would previously hang forever.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -931,7 +931,7 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 	}
 
 	s.session = project.NewSession(&project.SessionInit{
-		BackgroundCtx: ctx,
+		BackgroundCtx: s.backgroundCtx,
 		Options: &project.SessionOptions{
 			CurrentDirectory:       cwd,
 			DefaultLibraryPath:     s.defaultLibraryPath,


jakebailey:
Thanks for looking into this. We really shouldn't have code that actually uses `Background()` if it's long running or interacts like this; there should be a path to plumb a proper context, I would think.

fubhy:
> There should be a path to plumb a proper context, I would think.

Along these lines ... ?

internal/lsp/server.go
```diff
@@ -130,8 +130,9 @@ var (
 )
 
 type Server struct {
-	r Reader
-	w Writer
+	r   Reader
+	w   Writer
+	ctx context.Context
```
jakebailey: Context saved on a struct is in general a no-no as the context is supposed to be a request scoped value passed down the call chain, but could be "okay" to embed in some cases. Makes it feel kind of gross to have, though

internal/lsp/server.go
```diff
@@ -246,7 +247,7 @@ func (s *Server) RefreshDiagnostics(ctx context.Context) error {
 // PublishDiagnostics implements project.Client.
 func (s *Server) PublishDiagnostics(ctx context.Context, params *lsproto.PublishDiagnosticsParams) error {
 	notification := lsproto.TextDocumentPublishDiagnosticsInfo.NewNotificationMessage(params)
-	s.outgoingQueue <- notification.Message()
+	s.send(notification.Message())
```
jakebailey: This is odd, because there's a context right here, but we ignore it. Surely if the entire server is cancelled, the context here must be cancelled, as it should be derived from the main context.

internal/project/background/queue.go
```diff
@@ -25,7 +25,16 @@ func (q *Queue) Enqueue(ctx context.Context, fn func(context.Context)) {
 	}
 	q.mu.RUnlock()
 
+	// Don't start new tasks if context is already cancelled
+	if ctx != nil && ctx.Err() != nil {
```
jakebailey: Contexts should never be nil, so checking nil should not be required.

internal/lsp/server.go
fubhy: Would it be okay in case of the logger?

jakebailey: Yeah, unfortunately I don't see a way around that one.

fubhy: What about storing the context in a closure there? Not sure if that's less awkward. I'll come back with a mockup.

internal/lsp/server.go
```diff
@@ -301,7 +299,21 @@ func (s *Server) RequestConfiguration(ctx context.Context) (*lsutil.UserPreferen
...
+	// After Run(): trySend is nil again, logs go to stderr.
+	s.logger.setSender(func(msg *lsproto.Message) bool {
+		select {
+		case s.outgoingQueue <- msg:
+			return true
+		case <-ctx.Done():
+			return false
+		}
+	})
+	defer s.logger.setSender(nil)
```
fubhy: What about this @jakebailey ?

jakebailey: This seems weird, honestly, I would prefer the context as a field over mutability

internal/lsp/server.go
jakebailey: Can you rename this and the other such struct cases `backgroundCtx`? I want the name to be clear that it's for background tasks

internal/project/session.go
```diff
@@ -54,6 +54,7 @@ type SessionOptions struct {
 }
 
 type SessionInit struct {
+	Ctx         context.Context
```
jakebailey: This one should be renamed too

internal/lsp/server.go
```diff
@@ -493,8 +495,18 @@ func (s *Server) sendError(id *lsproto.ID, err error) {
...
+	s.send(ctx, resp.Message())
+}
+
+// send writes a message to the outgoing queue, respecting context cancellation.
+func (s *Server) send(ctx context.Context, msg *lsproto.Message) {
+	select {
+	case s.outgoingQueue <- msg:
+		// sent
+	case <-ctx.Done():
+		// shutting down, drop message
```
jakebailey: I have a bad feeling about this dropping.

Since the server can make requests back to the client, we could in the future be in the middle of a request that is dropped, then hang waiting for a response. I think `send` needs to return `ctx.Err()` so we can bubble that back up and make sure we bail out.

fubhy: Yeah makes sense.

fubhy:
The latest CI error is again caused by this: https://github.com/microsoft/typescript-go/pull/2485

jakebailey:
I don't know how you're hitting these races in CI so often when I don't even remember seeing them anywhere until now 😅

internal/fourslash/fourslash.go
```diff
@@ -206,11 +206,10 @@ func NewFourslash(t *testing.T, capabilities *lsproto.ClientCapabilities, conten
 	fsFromMap := vfstest.FromMap(testfs, true /*useCaseSensitiveFileNames*/)
 	fs := bundled.WrapFS(fsFromMap)
 
-	var err strings.Builder
 	server := lsp.NewServer(&lsp.ServerOptions{
 		In:  inputReader,
 		Out: outputWriter,
-		Err: &err,
+		Err: t.Output(),
```
jakebailey: This will cause every test to have a load of logging; we aren't ready to do this quite yet. I would just change this to a discard, or make a wrapper writer that's safe for concurrent use.

fubhy:
> I don't know how you're hitting these races in CI so often when I don't even remember seeing them anywhere until now 😅

Hah I guess that's just because I'm working on this in so many incremental steps because I'm so inexperienced with this codebase (and Go, frankly) that I trigger more CI runs than you maybe? ;-)

internal/lsp/server.go
```diff
@@ -493,8 +499,18 @@ func (s *Server) sendError(id *lsproto.ID, err error) {
 	})
 }
 
-func (s *Server) sendResponse(resp *lsproto.ResponseMessage) {
-	s.outgoingQueue <- resp.Message()
+func (s *Server) sendResponse(ctx context.Context, resp *lsproto.ResponseMessage) {
+	_ = s.send(ctx, resp.Message())
```
jakebailey: Why the discard? It seems like many of the senders can actually bubble out, e.g. readLoop sends some messages and could return.

fubhy: I let it bubble up further. Not sure what to do with the remaining discards though.

internal/lsp/server.go
```diff
@@ -406,11 +413,11 @@ func (s *Server) dispatchLoop(ctx context.Context) error {
 			handle := func() {
```
jakebailey, line 413: `handle` can return `error` here, which would allow propagation for the blocking methods.

internal/lsp/server.go
```diff
@@ -508,7 +530,7 @@ func (s *Server) handleRequestOrNotification(ctx context.Context, req *lsproto.R
 	}
 	s.logger.Warn("unknown method '", req.Method, "'")
 	if req.ID != nil {
-		s.sendError(req.ID, lsproto.ErrorCodeInvalidRequest)
+		_ = s.sendError(ctx, req.ID, lsproto.ErrorCodeInvalidRequest)
```
jakebailey: This could return the error, no?

internal/lsp/server.go
```diff
@@ -750,12 +768,12 @@ func (s *Server) getLanguageServiceAndCrossProjectOrchestrator(ctx context.Conte
...
 }
 
-func (s *Server) recover(req *lsproto.RequestMessage) {
+func (s *Server) recover(ctx context.Context, req *lsproto.RequestMessage) {
 	if r := recover(); r != nil {
 		stack := debug.Stack()
 		s.logger.Errorf("panic handling request %s: %v\n%s", req.Method, r, string(stack))
 		if req.ID != nil {
-			s.sendError(req.ID, fmt.Errorf("%w: panic handling request %s: %v", lsproto.ErrorCodeInternalError, req.Method, r))
+			_ = s.sendError(ctx, req.ID, fmt.Errorf("%w: panic handling request %s: %v", lsproto.ErrorCodeInternalError, req.Method, r))
```
jakebailey, line 776: This one seems unavoidable, but if we are panicking, we will have backed out of the code already, so it should be okay.

internal/lsp/server.go
jakebailey: Well, it's probably not strictly required, since we immediately check the context above. So, nevermind.

internal/lsp/server.go
```diff
@@ -913,6 +931,7 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 	}
 
 	s.session = project.NewSession(&project.SessionInit{
+		BackgroundCtx: ctx,
```
jakebailey: Missed this; this should be `s.backgroundCtx`, because otherwise this is the context tied to the incoming request.

fubhy: Thanks for the patience

jakebailey: "missed" meaning I missed it in review 😅

jakebailey (approved):
I feel like this is good, now.

PR #2502, opened 2026-01-14
Fix module specifier cache misses in symlinked package directories

andrewbranch:
Fixes slow completion times seen in https://github.com/microsoft/typescript-go/pull/2390#issuecomment-3735316463. In my repro, this goes from 8.3 seconds to 120ms 😄 

Big thank you to @rubnogueira and @jansedlon for the CPU profiles and investigation 🥳

### DIFF
--- a/internal/module/resolver.go
+++ b/internal/module/resolver.go
@@ -2026,19 +2026,28 @@ type ResolvedEntrypoints struct {
 type Ending int
 
 const (
+	// EndingFixed indicates that the module specifier cannot be changed without changing its resolution.
 	EndingFixed Ending = iota
+	// EndingExtensionChangeable indicates that the module specifier's extension portion was inferred from a
+	// file on disk, so an interchangeable one could be used instead (e.g. replacing .d.ts with .js).
 	EndingExtensionChangeable
+	// EndingChangeable indicates that the module specifier's file name and extension portion were inferred
+	// from a file on disk without being matched as part of an 'exports' pattern, so can be changed according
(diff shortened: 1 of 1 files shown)


andrewbranch:
I need to look closer at the project reference specific code. @jansedlon’s draft pointed out that we want those to take the slow path, but I want to see if that’s really necessary.

jakebailey (approved):

rubnogueira:
@andrewbranch happy to provide additional context if needed, thank you!

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -11,6 +11,8 @@ TestAutoImportCompletionExportListAugmentation1
 TestAutoImportCompletionExportListAugmentation2
 TestAutoImportCompletionExportListAugmentation3
 TestAutoImportCompletionExportListAugmentation4
+TestAutoImportCrossProject_symlinks_stripSrc
+TestAutoImportCrossProject_symlinks_toDist
```
andrewbranch, line 15: These two fall into a category I mentioned in #2390 where tsconfig.json `paths` were being used to override the `node_modules` package specifier, which I've decided not to support until/unless there's a well justified demand for it. It doesn't seem like a great idea. The tests came from an era before package.json `exports` were common; there are now better ways to solve the problem they were proving they could solve.

rubnogueira:
@andrewbranch just to let you know that I'm testing these changes with the 0.20260115.1 extension, and it's so, so much faster! Instant autocomplete, way more reliable, and it doesn't seem to "lag" the IDE as before. Thank you so much! :)

PR #2513, opened 2026-01-15
Support `Object.defineProperty` for expando properties and CommonJS exports in JS files

### LINKED ISSUE #2527: tsgo incorrectly emits both ESM and CJS exports when compiling `.js` files containing `exports.xxx` statements
## Steps to reproduce

Minimal reproducing example: https://github.com/psm14/tsgo-exports-bug-reproduction

## Source sample

```js
/**
 * Greets a user by name.
 * @param {string} name - The name to greet
 * @returns {string} The greeting message
 */
exports.greet = function greet(name) {
  return `${DEFAULT_PREFIX}, ${name}!`;
};
```

```js
"use strict";
/**
 * Greets a user by name.
 * @param {string} name - The name to greet
 * @returns {string} The greeting message
 */
exports.greet = function greet(name) {
    return `${DEFAULT_PREFIX}, ${name}!`;
};
```

## Behavior with `tsgo`

```js
"use strict";
/**
 * Greets a user by name.
 * @param {string} name - The name to greet
 * @returns {string} The greeting message
 */
export var greet = function greet(name: string): string {
    return `${DEFAULT_PREFIX}, ${name}!`;
};
/**
 * Greets a user by name.
 * @param {string} name - The name to greet
 * @returns {string} The greeting message
 */
exports.greet = function greet(name) {
    return `${DEFAULT_PREFIX}, ${name}!`;
};
```

ahejlsberg:
In this PR:

* In JS files, permit expando properties to be defined with `Object.defineProperty`.
* In JS files, permit CommonJS exports to be defined with `Object.defineProperty` applied to `exports` or `module.exports`.
* Fix circularity errors and unwanted implicit-any errors by deferring resolution of expando object property types.
* Fix error on `typeof module.exports` when used to define types in JSDoc comments in JS files.
* Cache creation of CommonJS module symbols in `getModuleSymbol`.
* Don't emit CommonJSExport nodes in JS files (those are synthetic reparsed nodes intended just for checking).
* Refactor declaration emit to support CommonJS exports defined with `Object.defineProperty`.
* Port https://github.com/microsoft/TypeScript/pull/60683 to fix stack overflow in test.
* Support chained assignments in reparsing.
* Revised handling of CommonJS assignment declarations with `undefined` initializers.

The revised handling of CommonJS assignment declarations now ignores `undefined` initializers when they occur as the initial assignment declaration of an export with multiple declarations. This provides for the common pattern of first initializing CommonJS exports to `undefined` and then subsequently assigning the actual exported value.

Fixes #2527.


testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameTemplateLiteralsDefinePropertyJs.baseline.jsonc
```diff
@@ -27,14 +27,18 @@
 
 // === findRenameLocations ===
 // === /a.js ===
-// --- (line: 5) skipped ---
-//     [`prop`]: 1
+// let obj = {};
+// 
+// Object.defineProperty(obj, `prop`, { value: 0 });
```
jakebailey: Seems like Strada's rename would also update the defineProperty. Not sure how possible that is now.

testdata/baselines/reference/submodule/conformance/jsDeclarationsExportDefinePropertyEmit.js.diff
```diff
@@ -76,22 +148,41 @@
...
++    export { _exported_6 as "h" };
++    declare const _exported_7: () => void;
++    export { _exported_7 as "i" };
++    declare const _exported_8: () => void;
++    export { _exported_8 as "ii" };
++    declare const _exported_9: () => void;
++    export { _exported_9 as "jj" };
++    declare const _exported_10: () => void;
++    export { _exported_10 as "j" };
++
```
jakebailey, line 188: The exports here seem to be causing `module` to not exist, or something? Definitely odd.

testdata/baselines/reference/submodule/compiler/jsExpandoObjectDefineProperty.errors.txt.diff
```diff
@@ -0,0 +1,13 @@
...
++++ new.jsExpandoObjectDefineProperty.errors.txt
+@@= skipped -0, +0 lines =@@
+-<no content>
++index.js(3,17): error TS2339: Property 'inspectedWindow' does not exist on type '{}'.
++
++
++==== index.js (1 errors) ====
++    var chrome = {}
++    Object.defineProperty(chrome, 'devtools', { value: {}, enumerable: true })
++    chrome.devtools.inspectedWindow = {}
```
jakebailey, line 11: I suppose this is consistentish, but it is weird that if I write:

```js
var chrome = {}
chrome.devtools = {}
chrome.devtools.inspectedWindow = {}
```

You get an error on the second line too, so expandos are sort of a way around a check.

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameTemplateLiteralsDefinePropertyJs.baseline.jsonc
ahejlsberg: I'll take a look.

testdata/baselines/reference/submodule/conformance/jsDeclarationsExportDefinePropertyEmit.js.diff
ahejlsberg: It's a declaration emit issue. The old declaration emitter wouldn't attempt to reference `module.exports` (it's only a CommonJS thing and can't be referenced in a `.d.ts` file), but instead would inline the actual type.

testdata/baselines/reference/submodule/compiler/jsExpandoObjectDefineProperty.errors.txt.diff
ahejlsberg: We could probably fix this.

ahejlsberg: This works in a JS file (but not in a TS file):

```ts
var chrome = {}
chrome.devtools = {}
chrome.devtools.inspectedWindow = {}
```

So you could argue that using `Object.defineProperty` for the middle one should work too. Which is the case in Strada.

ahejlsberg: Honestly, having expando objects that are rooted in `Object.defineProperty` seems like a totally esoteric feature. I'm going to hold off on implementing that.

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameTemplateLiteralsDefinePropertyJs.baseline.jsonc
ahejlsberg: Now fixed, except initiating a rename in a string literal doesn't work, but that's a general issue for Corsa that we should fix in a separate PR.

jakebailey:
Are you planning on putting more into this PR or is it ready?

ahejlsberg:
> Are you planning on putting more into this PR or is it ready?

The PR is ready now. The tests reveal issues in declaration emit, namely:

* Use of `typeof module.exports.xxx` (which isn't permitted in .ts files),
* Emitting multiple `export declare var foo: Foo` for multiple assignments to `exports.foo`, and
* Emitting `export declare var foo: Foo` instead of `export const foo: Foo` for CommonJS exports.

We can address these in separate PRs.

PR #2515, opened 2026-01-15
[LSP] Process file deletions against tsconfig `includes`

andrewbranch:
We weren't triggering tsconfig file name reloads in response to watcher deletion events that match `includes` wildcards, so we were holding onto source files that should have been deleted for too long.

### DIFF
--- a/internal/project/configfileregistrybuilder.go
+++ b/internal/project/configfileregistrybuilder.go
@@ -363,7 +363,8 @@ func (c *configFileRegistryBuilder) DidChangeFiles(summary FileChangeSummary, lo
 	logger.Log("Summarizing file changes")
 	hasExcessiveChanges := summary.HasExcessiveWatchEvents() && summary.IncludesWatchChangeOutsideNodeModules
 	createdFiles := make(map[tspath.Path]string, summary.Created.Len())
-	createdOrDeletedFiles := make(map[tspath.Path]struct{}, summary.Created.Len()+summary.Deleted.Len())
+	deletedFiles := make(map[tspath.Path]string, summary.Deleted.Len())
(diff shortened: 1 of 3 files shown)


internal/project/configfileregistrybuilder.go
```diff
@@ -423,24 +431,54 @@ func (c *configFileRegistryBuilder) DidChangeFiles(summary FileChangeSummary, lo
...
+		if hasExcessiveChanges {
+			return c.invalidateCache(logger)
 		}
+		c.configs.Range(func(entry *dirty.SyncMapEntry[tspath.Path, *configFileEntry]) bool {
+			entry.ChangeIf(
+				func(config *configFileEntry) bool {
+					if config.pendingReload != PendingReloadNone || config.commandLine == nil {
+						return false
+					}
+					if _, ok := config.commandLine.FileNamesByPath()[path]; ok {
```
sheetalkamat, line 462: do you need to handle case of getting notification for "open file"?

andrewbranch: Do you mean getting a file open for something that doesn't actually exist on disk?

sheetalkamat: file is already open and the delete event comes ?
In strada we sometimes use to not get some of the events (and switching between branches could send delete/create and if you missed open - the open file would fall off the project) so we would pretend all open files are present till they are closed.

andrewbranch: I'm going to open a new PR experimenting with this, along with what you pointed out in https://github.com/microsoft/typescript-go/pull/2456. But I think the changes here are still needed and improve things on their own.

andrewbranch:
Incorporating this into a larger PR

PR #2520, opened 2026-01-15
Fix missing content change during consecutive didClose/didOpen

### LINKED ISSUE #2511: Stale diagnostics on file reload
When Neovim detects that a file has changed on disk (e.g., on `:checktime`), it reloads it and notifies the LSP server by `textDocument/didClose` followed by `textDocument/didOpen`. When this happens, `tsgo` diagnostics are not updated until the file is modified in the editor again.

jakebailey:
This fixes #2511, but the tests are flaky, I think because the events are not always combined.

Would love some feedback about this ☹️

### DIFF
--- a/internal/project/overlayfs_test.go
+++ b/internal/project/overlayfs_test.go
@@ -227,8 +227,8 @@ func TestProcessChanges(t *testing.T) {
 			},
 		})
 
-		// Should be marked as opened
-		assert.Equal(t, result.Opened, testURI1)
+		// Should not be marked as opened since it was already open
+		assert.Assert(t, result.Opened == "", "close then open should not mark as opened")
 		// Should also be marked as changed since it was closed and reopened
 		assert.Assert(t, result.Changed.Has(testURI1), "close then open should mark as changed")
 		// Should have the new content


frangio:
I can confirm that this fixes #2511 for me.

internal/project/overlayfs.go
```diff
@@ -300,6 +304,9 @@ func (fs *overlayFS) processChanges(changes []FileChange) (FileChangeSummary, ma
 				panic("can only process one file open event at a time")
 			}
 			result.Opened = uri
+			if events.reopened {
+				result.Changed.Add(uri)
```
andrewbranch, line 307: We might need to see if the contents of the didOpen actually resulted in a change...

testdata/baselines/reference/fourslash/state/declarationMapsRenameWithDisableSourceOfProjectReferenceRedirect.baseline
```diff
@@ -426,7 +435,7 @@ Projects::
     /myproject/decls/FnS.d.ts  
     /myproject/main/main.ts    
   [/myproject/tsconfig.json] *deleted*
-  [/random/tsconfig.json] 
+  [/random/tsconfig.json] *modified*
```
andrewbranch: ...I'm guessing that would resolve all these changed baselines, but I'm not really sure what would have caused this.

internal/project/overlayfs.go
jakebailey: I think that worked!

jakebailey: Though, I am a little unsure about the logistics of multiple changes within the same FileChange, so maybe the continue is wrong?

jakebailey: Pushed something that I think is better

andrewbranch:
Seems good in principle but I really want to know why all those server baselines changed 🤔

jakebailey:
Yeah 🫤

andrewbranch:
I think I fixed it

internal/project/overlayfs.go
```diff
@@ -300,6 +304,9 @@ func (fs *overlayFS) processChanges(changes []FileChange) (FileChangeSummary, ma
 				panic("can only process one file open event at a time")
 			}
 			result.Opened = uri
+			if events.reopened && o != nil && o.Content() != events.openChange.Content {
```
andrewbranch: Honestly we don't even need the `reopened` flag to do this... it's a little weird that we're saying this was opened at all... your last commit "fixed" that but still broke baselines, so maybe I don't fully understand what's going on yet after all

andrewbranch: Ok, now I think it makes more sense and still passes tests

sheetalkamat (approved):

PR #2521, opened 2026-01-15
Port rootDir default to configFile directory and some module resoluton fixes

sheetalkamat:
Port of https://github.com/microsoft/TypeScript/pull/62418


andrewbranch:
Should we get https://github.com/microsoft/TypeScript/pull/62418 in the porting branch of the submodule first so the diffs are clean?

jakebailey:
Sheetal asked me about this, and I think actually we should start targeting TS 6.0 in the porting branch, which I wanted to start doing next week, it's just annoying due to some patches that had been applied.

In the meantime I think a few diffs is probably okay?

PR #2524, opened 2026-01-16
feat: detect deprecated via JSDoc and flag deprecated property access

AyomiCoder:
This PR implements the deprecated property access check in `internal/checker/checker.go`, resolving a TODO comment. The implementation translates the logic from the TypeScript compiler to Go.

Additionally, it updates `IsDeprecatedDeclaration` to fallback to checking `getJSDocDeprecatedTag`, ensuring robust detection even if `NodeFlagsDeprecated` is not set.

- **internal/checker/checker.go**:
  - Uncommented and implemented the deprecation check block in `checkPropertyAccessExpressionOrQualifiedName`.
  - Added logic to resolve the correct `deprecatedNode`.
  - Updated `IsDeprecatedDeclaration` to check JSDoc tags.

- **Compilation:** `go build ./cmd/tsgo` passes.
- **Manual Testing:** Tested with a `@deprecated` property.
- **Observation:** The check logic is correct, but the currently implemented Parser does not yet attach JSDoc comments to AST nodes, so `getJSDocDeprecatedTag` returns `nil`. This feature will automatically become active once JSDoc parsing support is added upstream.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -13567,7 +13567,10 @@ func (c *Checker) addErrorOrSuggestion(isError bool, diagnostic *ast.Diagnostic)
 }
 
 func (c *Checker) IsDeprecatedDeclaration(declaration *ast.Node) bool {
-	return c.getCombinedNodeFlagsCached(declaration)&ast.NodeFlagsDeprecated != 0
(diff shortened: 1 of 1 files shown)


AyomiCoder:
I accept

AyomiCoder:

PR #2530, opened 2026-01-16
Disable Go CI caching

jakebailey:
We spend a lot of time making caches and restoring caches. I tried pretty hard to make this work, but I'm not sure it helps all that much given we keep hitting storage limits and so on...

I'm curious what happens if we just don't bother anymore.


jakebailey:
From my comparison, the caching is doing nothing for us. I don't think it's worth having all of this stuff around anymore.

PR #2532, opened 2026-01-16
Incremental fixes

### LINKED ISSUE #1465: Intermittent crash with --incremental
## Stack trace

https://gist.github.com/tmm1/03a124c5e5c19dede914bf6566755a2c

## Steps to reproduce

1. run `tsgo --noEmit --extendedDiagnostics -p src --incremental` on large project at high frequency in CI

sheetalkamat:
I am going to try and create repro for this but i think this fixes the issue where if there is reference cycle with the const enum usage , it runs into stack overflow.

Fixes #1465

### DIFF
--- a/internal/execute/incremental/affectedfileshandler.go
+++ b/internal/execute/incremental/affectedfileshandler.go
@@ -200,7 +200,8 @@ func (h *affectedFilesHandler) handleDtsMayChangeOfAffectedFile(dtsMayChange dts
 		return
 	}
 
-	// At this point affectedFile is actually changed file with change in signature.
+	// At this point affectedFile is actually one of the changed files
+	// that has some change in its .d.ts signature.
 
 	// Since isolated modules dont change js files, files affected by change in signature is itself
 	// But we need to cleanup semantic diagnostics and queue dts emit for affected files
@@ -278,7 +279,8 @@ func (h *affectedFilesHandler) handleDtsMayChangeOfFileAndReferences(dtsMayChang
 	}
 	h.handleDtsMayChangeOf(dtsMayChange, filePath, invalidateJsFiles)
 
-	// Remove the diagnostics of files that imports this file and references to this (direct or indirect)
+	// Remove the diagnostics of files that import this file and
+	// any files that are referenced by it (directly or indirectly)
 	for referencingFilePath := range h.program.snapshot.referencedMap.getReferencedBy(filePath) {
 		if h.handleDtsMayChangeOfFileAndReferences(dtsMayChange, referencingFilePath, invalidateJsFiles) {
 			return true


internal/execute/incremental/affectedfileshandler.go
```diff
@@ -200,6 +200,8 @@ func (h *affectedFilesHandler) handleDtsMayChangeOfAffectedFile(dtsMayChange dts
 		return
 	}
 
+	// At this point affectedFile is actually changed file with change in signature.
```
DanielRosenwasser:
```suggestion
	// At this point affectedFile is actually one of the changed files
	// that has some change in its .d.ts signature.
```

internal/execute/incremental/affectedfileshandler.go
```diff
@@ -35,7 +35,7 @@ type affectedFilesHandler struct {
 	dtsMayChange                           []dtsMayChange
 	filesToRemoveDiagnostics               collections.SyncSet[tspath.Path]
 	cleanedDiagnosticsOfLibFiles           sync.Once
-	seenFileAndExportsOfFile               collections.SyncMap[tspath.Path, bool]
+	seenFileAndReferences                  collections.SyncMap[tspath.Path, bool]
```
DanielRosenwasser, line 38: Can we make this a `collections.SyncSet` instead?

internal/execute/incremental/affectedfileshandler.go
```diff
@@ -250,30 +252,35 @@ func (h *affectedFilesHandler) handleDtsMayChangeOfAffectedFile(dtsMayChange dts
...
 				return
 			}
 		}
 	}
 }
 
-func (h *affectedFilesHandler) handleDtsMayChangeOfFileAndExportsOfFile(dtsMayChange dtsMayChange, filePath tspath.Path, invalidateJsFiles bool) bool {
-	if existing, loaded := h.seenFileAndExportsOfFile.LoadOrStore(filePath, invalidateJsFiles); loaded && (existing || !invalidateJsFiles) {
+func (h *affectedFilesHandler) handleDtsMayChangeOfFileAndReferences(dtsMayChange dtsMayChange, filePath tspath.Path, invalidateJsFiles bool) bool {
+	if existing, loaded := h.seenFileAndReferences.LoadOrStore(filePath, invalidateJsFiles); loaded && (existing || !invalidateJsFiles) {
```
DanielRosenwasser, line 271: Since this is really a set, you don't need to handle `existing` and `loaded` differently.

internal/execute/incremental/affectedfileshandler.go
```diff
@@ -250,30 +252,35 @@ func (h *affectedFilesHandler) handleDtsMayChangeOfAffectedFile(dtsMayChange dts
...
+		h.seenFileAndReferences.Store(filePath, true)
 	}
+
 	if h.handleDtsMayChangeOfGlobalScope(dtsMayChange, filePath, invalidateJsFiles) {
 		return true
 	}
 	h.handleDtsMayChangeOf(dtsMayChange, filePath, invalidateJsFiles)
 
-	// Remove the diagnostics of files that import this file and handle all its exports too
+	// Remove the diagnostics of files that imports this file and references to this (direct or indirect)
```
DanielRosenwasser:
```suggestion
	// Remove the diagnostics of files that import this file and
	// any files that are referenced by it (directly or indirectly)
```

?

internal/execute/incremental/affectedfileshandler.go
sheetalkamat: This has value bool - true if js files were invalidated. otherwise false. So this is not a set but map.

internal/execute/incremental/affectedfileshandler.go
sheetalkamat: This isnt a set. The value is "true if we invalidated js files, otherwise false.

chriskrycho:
I can confirm that this fixed a stack overflow in our code (in this same code path)! **Edit:** I should say: one user who was consistently hitting it has seen this build resolve it, and I’m also looking for further confirmation from other users in our app who’ve hit it, but signs are promising.

PR #2539, opened 2026-01-18
Add support for importing JSON modules "as const"

james-pre:
This PR implements importing JSON modules with const context, which has the same effect as an `as const` assertion.

This was discussed in https://github.com/microsoft/TypeScript/issues/32063. The issue suggests using a `const` import attribute, however this PR uses a compiler option which is much simpler and should be more maintainable.

This feature is opt-in, so no existing code will be broken. It can be enabled by setting `compilerOptions.importJsonAsConst` to `true` in tsconfig.

I've already tested a patch of the compiler written in Typescript on some of my own code and it works great. I also added some new tests for completeness.


james-pre:

RyanCavanaugh:
Closing for housekeeping reasons (feature PRs, which should be for approved issues, would only happen post-7.0, which will be back in the `TypeScript` repo)

PR #2549, opened 2026-01-20
Include overlays in all file systems used in snapshot building

### LINKED ISSUE #1983: textDocument/diagnostic failed

### LINKED ISSUE #2506: Stale module resolution error after linking pnpm workspace dependency due to watcher configuration/handling
### Extension Version

main

### VS Code Version

1.109.0-insider (Universal)

### Operating system Version

macOS Tahoe 26.2

### Steps to reproduce

1. Create a pnpm workspace with two packages, but do not `pnpm install` yet.
2. Make one package depend on the other and write an import that will eventually resolve through a node_modules symlink.
3. `pnpm install`

### Issue

`pnpm install` creating the node_modules link should make the error go away, but the file watcher doesn't trigger. VS Code sees the directory symlink creation, but doesn't send it to us because it doesn't match our watch glob pattern which ends with `**/*.{js,jsx,mjs,cjs,ts,tsx,mts,cts,json}` (unless the directory name happens to have one of those extensions). Dropping the extension alone is not sufficient, since we check the created files from the `didChangeWatchedFiles` notification against `failedLookupLocations` keys, which are full files.

This is somewhat related to our other known directory watch issue, which is missing events when deleting a directory, since VS Code only sends a deletion event for the directory, not for the files inside.

andrewbranch:
@sheetalkamat pointed out that the fix in https://github.com/microsoft/typescript-go/pull/2456 was a little backwards, and that was because we never used the overlay-augmented file system in determining root files for tsconfigs; we were only using the on-disk file system. We were also a little inconsistent about converting opens/closes to creates/deletes; we wouldn't process a deletion for an open file (good), but we didn't simulate a creation event for an open that we had never seen before (bad), and we also weren't simulating a deletion for the closure of an open file that no longer exists on disk (bad).

This fixes #1983 (at least one very common reproduction of it; I'm optimistic). What was happening was:

1. Do a `git switch` or `mv` a file or do something that batches a deletion of a file currently open in your editor with the creation of a new file that matches the `include` of the same tsconfig.json as the deleted open file.
2. We would ignore the deletion event since the file remains open. However, the creation event would cause the file names for the impacted tsconfig.json to be reevaluated.
3. The file system used to reevaluate file names for the tsconfig.json was a cache in front of the disk, and the cache did still contain the deleted but open file since we ignored the deletion event, _but_ that file system delegated its `GetAccessibleEntries` (read directory) to the underlying OS file system, so the deleted file would get dropped from the tsconfig's root files.

To make matters even more confusing, _program construction_ always used the overlay file system, so if the deleted but open file was _imported_ by another program file, program construction _would_ find that file and include it, and everything would work—this is why the error was a little tricky to reproduce. You needed an open file that was a root of a tsconfig and not included in the program any other way.

This PR makes the following changes:

1. Overlays are included in the file system used everywhere throughout snapshot building
2. That file system properly implements `GetAccessibleEntries` so overlays appear in directory reads whether they actually exist on disk or not.
3. After Session makes a `FileChangeSummary` of opens/closes/creates/deletes/changes, that summary is compared against the previous snapshot's file system to convert opens/closes to creates/deletes.
4. I pulled #2515 into this PR; we were missing some deletions

The directory structure that enables (2) will also directly enable us to fix #2506 and the directory deletion file watcher issue, but I left those fixes out of this PR to keep the scope smaller.

### DIFF
--- a/testdata/baselines/reference/fourslash/state/callHierarchyAcrossProject.baseline
+++ b/testdata/baselines/reference/fourslash/state/callHierarchyAcrossProject.baseline
@@ -579,29 +579,28 @@ Open Files::
   }
 }
 
+Projects::
+  [/projects/container/compositeExec/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts            
+    /projects/container/compositeExec/index.ts  
+  [/projects/container/exec/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts   
+    /projects/container/exec/index.ts  
+  [/projects/container/lib/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts  
+    /projects/container/lib/bar.ts    
+    /projects/container/lib/baz.ts    
+  [/projects/container/tsconfig.json] *deleted*
+  [/projects/temp/tsconfig.json] 
+    /projects/temp/temp.ts  
 Open Files::
   [/projects/temp/temp.ts] *new*
     /projects/temp/tsconfig.json  (default) 
 Config::
-  [/projects/container/compositeExec/tsconfig.json] 
-    RetainingProjects:
-      /projects/container/compositeExec/tsconfig.json  
-      /projects/container/tsconfig.json                
-  [/projects/container/exec/tsconfig.json] 
-    RetainingProjects:
-      /projects/container/exec/tsconfig.json  
-      /projects/container/tsconfig.json       
-  [/projects/container/lib/tsconfig.json] *modified*
-    RetainingProjects:
-      /projects/container/compositeExec/tsconfig.json  
-      /projects/container/exec/tsconfig.json           
-      /projects/container/lib/tsconfig.json            
-      /projects/container/tsconfig.json                
-    RetainingOpenFiles: *modified*
-      /projects/container/lib/index.ts  *deleted*
-  [/projects/container/tsconfig.json] 
-    RetainingProjects:
-      /projects/container/tsconfig.json  
+  [/projects/container/compositeExec/tsconfig.json] *deleted*
+  [/projects/container/exec/tsconfig.json] *deleted*
+  [/projects/container/lib/tsconfig.json] *deleted*
+  [/projects/container/tsconfig.json] *deleted*
   [/projects/temp/tsconfig.json] 
     RetainingProjects:
       /projects/temp/tsconfig.json  
--- a/internal/project/filechange.go
+++ b/internal/project/filechange.go
@@ -34,9 +34,11 @@ type FileChange struct {
 
 type FileChangeSummary struct {
 	// Only one file can be opened at a time per request
-	Opened  lsproto.DocumentUri
-	Closed  collections.Set[lsproto.DocumentUri]
-	Changed collections.Set[lsproto.DocumentUri]
+	Opened lsproto.DocumentUri
+	// Reopened is set if a close and open occurred for the same file in a single batch of changes.
+	Reopened lsproto.DocumentUri
+	Closed   collections.Set[lsproto.DocumentUri]
+	Changed  collections.Set[lsproto.DocumentUri]
 	// Only set when file watching is enabled
 	Created collections.Set[lsproto.DocumentUri]
 	// Only set when file watching is enabled
@@ -48,7 +50,7 @@ type FileChangeSummary struct {
 }
 
 func (f FileChangeSummary) IsEmpty() bool {
-	return f.Opened == "" && f.Closed.Len() == 0 && f.Changed.Len() == 0 && f.Created.Len() == 0 && f.Deleted.Len() == 0
+	return f.Opened == "" && f.Reopened == "" && f.Closed.Len() == 0 && f.Changed.Len() == 0 && f.Created.Len() == 0 && f.Deleted.Len() == 0
 }
 
 func (f FileChangeSummary) HasExcessiveWatchEvents() bool {
--- a/internal/project/overlayfs.go
+++ b/internal/project/overlayfs.go
@@ -298,13 +298,15 @@ func (fs *overlayFS) processChanges(changes []FileChange) (FileChangeSummary, ma
 		o := newOverlays[path]
 
 		if events.openChange != nil {
-			if result.Opened != "" {
+			if result.Opened != "" || result.Reopened != "" {
 				panic("can only process one file open event at a time")
 			}
 			if o != nil && o.Content() != events.openChange.Content {
 				result.Changed.Add(uri)
 			} else if o == nil {
 				result.Opened = uri
+			} else {
+				result.Reopened = uri
 			}
 			newOverlays[path] = newOverlay(
 				uri.FileName(),
--- a/internal/project/projectcollectionbuilder.go
+++ b/internal/project/projectcollectionbuilder.go
@@ -244,10 +244,10 @@ func (b *ProjectCollectionBuilder) DidChangeFiles(summary FileChangeSummary, log
 	})
 
 	// Handle opened file
-	if summary.Opened != "" {
-		fileName := summary.Opened.FileName()
-		path := b.toPath(fileName)
+	if summary.Opened != "" || summary.Reopened != "" {
 		var toRemoveProjects collections.Set[tspath.Path]
+		fileName := core.FirstNonZero(summary.Opened, summary.Reopened).FileName()
+		path := b.toPath(fileName)
 		openFileResult := b.ensureConfiguredProjectAndAncestorsForFile(fileName, path, logger)
 		b.configuredProjects.Range(func(entry *dirty.SyncMapEntry[tspath.Path, *Project]) bool {
 			toRemoveProjects.Add(entry.Key())
--- a/internal/project/snapshot.go
+++ b/internal/project/snapshot.go
@@ -339,7 +339,7 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
 
 	// Clean cached disk files not touched by any open project. It's not important that we do this on
 	// file open specifically, but we don't need to do it on every snapshot clone.
-	if len(change.fileChanges.Opened) != 0 {
+	if change.fileChanges.Opened != "" || change.fileChanges.Reopened != "" {
 		// The set of seen files can change only if a program was constructed (not cloned) during this snapshot.
 		if len(projectsWithNewProgramStructure) > 0 {
 			cleanFilesStart := time.Now()
--- a/testdata/baselines/reference/fourslash/state/codeLensAcrossProjects.baseline
+++ b/testdata/baselines/reference/fourslash/state/codeLensAcrossProjects.baseline
@@ -736,29 +736,27 @@ Open Files::
   }
 }
 
+Projects::
+  [/projects/container/compositeExec/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts            
+    /projects/container/compositeExec/index.ts  
+  [/projects/container/exec/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts   
+    /projects/container/exec/index.ts  
+  [/projects/container/lib/tsconfig.json] *deleted*
+    /projects/container/lib/index.ts  
+    /projects/container/lib/bar.ts    
+  [/projects/container/tsconfig.json] *deleted*
+  [/projects/temp/tsconfig.json] 
+    /projects/temp/temp.ts  
 Open Files::
   [/projects/temp/temp.ts] *new*
     /projects/temp/tsconfig.json  (default) 
 Config::
-  [/projects/container/compositeExec/tsconfig.json] 
-    RetainingProjects:
-      /projects/container/compositeExec/tsconfig.json  
-      /projects/container/tsconfig.json                
-  [/projects/container/exec/tsconfig.json] 
-    RetainingProjects:
-      /projects/container/exec/tsconfig.json  
-      /projects/container/tsconfig.json       
-  [/projects/container/lib/tsconfig.json] *modified*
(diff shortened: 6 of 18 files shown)


testdata/baselines/reference/fourslash/state/callHierarchyAcrossProject.baseline
```diff
@@ -591,7 +605,7 @@ Projects::
     /projects/container/lib/bar.ts    
     /projects/container/lib/baz.ts    
   [/projects/container/tsconfig.json] *deleted*
-  [/projects/temp/tsconfig.json] 
+  [/projects/temp/tsconfig.json] *modified*
```
andrewbranch: Looking into why these baselines changed

testdata/baselines/reference/fourslash/state/callHierarchyAcrossProject.baseline
```diff
@@ -579,28 +579,29 @@ Open Files::
...
-    /projects/temp/temp.ts  
 Open Files::
   [/projects/temp/temp.ts] *new*
     /projects/temp/tsconfig.json  (default) 
 Config::
-  [/projects/container/compositeExec/tsconfig.json] *deleted*
-  [/projects/container/exec/tsconfig.json] *deleted*
-  [/projects/container/lib/tsconfig.json] *deleted*
-  [/projects/container/tsconfig.json] *deleted*
+  [/projects/container/compositeExec/tsconfig.json]
```
sheetalkamat: This seems like it should have deleted unused projects but its not?

andrewbranch: Yeah, looking into it

andrewbranch: I see what’s happening. Not sure why it wasn’t happening before, but #2520 means that this:

```ts
	f.CloseFileOfMarker(t, "call")
	f.CloseFileOfMarker(t, "temp")
	f.GoToMarker(t, "temp")
```

is no longer a file open, since the file for `temp` got closed and reopened in the same update without any changes. And we only clean up projects on file opens.

sheetalkamat: i think we shoukld be doing this on file reopen as well ?

testdata/baselines/reference/fourslash/state/findAllRefsTwoProjectsOpenAndOneProjectReferences.baseline
```diff
@@ -298,7 +298,7 @@ Projects::
     /myproject/indirectDisabledChildLoad1/src/file1.ts  
   [/myproject/indirectDisabledChildLoad2/tsconfig.json] *new*
     /myproject/indirectDisabledChildLoad2/src/file1.ts  
-  [/myproject/main/tsconfig.json] 
+  [/myproject/main/tsconfig.json] *modified*
```
andrewbranch, line 301: I believe this is caused by treating the previous `/myproject/core/src/file1.ts` open as a create (since that was the first time we ever tried to read that file), which we recognize could have affected `/myproject/main/tsconfig.json` because we checked `/myproject/core/tsconfig.json` (via references) while searching for the default project for `/myproject/main/src/file1.ts`. This ultimately resulted in a no-op program update, and we probably could short circuit that somewhere if we really try.

sheetalkamat: Why would we treat open as create ? we would have already known presence of the file when we created main/tsconfig.json project right. (because we have command line for core as well as part of creating program for it?)

andrewbranch: So, we did load file names for `/myproject/core/tsconfig.json` before this, but we didn’t create a program for it. We never read `/myproject/core/src/file1.ts`; we only saw it via a `GetAccessibleEntries` call, so we didn’t create a file in the cache for it. We could do that, but we’d have to add more tracking to clear it out of the cache when it’s no longer relevant. Right now we only track references to files that are directly read or checked for existence.

sheetalkamat: Isn’t accessible entry = check for existence true ?

andrewbranch: No, it's asking what files/directories exist in a directory. The cache entry gets created literally on `FileExists()` and `ReadFile()`.

andrewbranch: Also FWIW, I think it would be wasteful to create a cache entry for every result of `GetAccessibleEntries`, as a lot of those don't end up in the program. The tradeoff of considering opens as creates, with the premise that we might miss creates, is that we're going to fire some false positive creates. I think the problem, insofar as there is one, is that this create caused a seemingly unrelated update that ended up being a no-op.

PR #2556, opened 2026-01-21
[Experiment] avoid OrderedMap in widening context properties

a-tarasyuk:


DanielRosenwasser:
I'm guessing you also tried `NewOrderedMapWithSizeHint` and there's too much indirection for that to be compiled away.

PR #2577, opened 2026-01-24
Add --release flag to emulate real released binaries, rename build tag to noassert, test release in CI

jakebailey:
I want to be able to produce "real" release binaries, with stripping and trimming.

Shift stuff around to make that possible. Since `--release` is sort of overlapping with the build tag `release`, I renamed that to `noassert`, matching `noembed`.

I've also added in a way to do a "full release build", building all platforms, making packages and vsixes, etc, so we can verify that's working too.


rubiesonthesky:
nodebug seems logical flag, but without context, it’s easy to read it as “node bug” 🐛 probably no harm, just funny coincidence.

jakebailey:
Yes, I thought that was funny.

Can't think of a better name, though. I could probably just leave it as `release` but we haven't had any reasons to flip code around in release mode (usually a bad idea, I am questioning us doing it still).

jakebailey:
Thinking more, we should have a `--nodebug` like `--noembed`, except that aliases another flag too. So, it's probably a bad build tag name and release was fine. Maybe...

jakebailey:
Also TODO, going to add a task which actually exercises the release procedure, for CI testing.

internal/debug/assert.go
```diff
@@ -1,4 +1,4 @@
-//go:build !nodebug
+//go:build !noasert
```
jakebailey: Typo 🤦

DanielRosenwasser:
I think it's getting hard to keep track of where things are. So with this PR there's now a `--release` flag with hereby, but where can we keep track of all the build flags in our codebase, and is there a way to toggle them ad-hoc?

jakebailey:
We have `noembed` and `noassert` build tags. They are inverted so that `go test` uses embedded files and have debug asserts.

When you do `hereby build`, you get a binary with both set, so no embedded files and no asserts, like `tsc`.

When you build with `hereby build --debug` (what the debug attach task does), you still get `noembed`, but you don't get `noassert`, so asserts will fire, and the binary is deoptimized for attaching.

When you build with `hereby build --release`, you get `noembed` and `noassert` like `hereby build`, but the build also does `-trimpath '-ldflags=-s -w'`, which strips paths and debug info from the binary.

There are other flags, but they are only used for testing (controlling embedding, concurrent test programs, race mode, coverage) or to pass metadata in for the release build.

PR #2587, opened 2026-01-26
Module specifier completions

### LINKED ISSUE #2383: LSP: Path suggestion missing (import/export)

gabritto:
Fixes #2383

I opted to port the code as-is for now because it wasn't very clear to me how to modify the existing code to use the autoimports registry for non-relative specifier completions. In the future we might want to do this for performance or correctness reasons. It would likely require untangling the existing code, which ends up calling the same function for  `getCompletionEntriesForDirectoryFragment` and reading from the file system for both relative and non-relative imports. We'd probably also have to simplify how we collect module specifier candidates to make better use of the information in the registry, e.g. if a user typed `"packageFoo/bar/baz/f"`, right now we end up reading the directory `packageFoo/bar/baz` and computing completions for that path, but we could just use all of the module specifier information collected for `packageFoo` and rely on the client to do the filtering for us for entries that match `"packageFoo/bar/baz/f"` (or do the filtering ourselves after we collect the entries).

Note:
We bypass the snapshot's immutable view of files to be able to see what files and directories exist. This means there could be inconsistencies between the module specifier completions and the state of the current snapshot, but the worst that could happen here is extra or missing module specifiers in completions, so that seems fine.

Future work:
- This PR currently uses the existing `ReadDirectory` as ported from Strada that uses regexes. We'll need https://github.com/microsoft/typescript-go/pull/2407 to get rid of regex use. The usage of `ReadDirectory` by module specifier completions probably doesn't need full fidelity with the original Strada code, the only nice to haves I can think of are the filtering of `*.min.js` files and maybe of paths starting with `.`.

### DIFF
--- a/internal/ls/completions.go
+++ b/internal/ls/completions.go
@@ -5093,7 +5093,9 @@ func (l *LanguageService) getImportStatementCompletionInfo(contextToken *ast.Nod
 		result.replacementSpan = l.getSingleLineReplacementSpanForImportCompletionNode(candidate)
 		result.couldBeTypeOnlyImportSpecifier = couldBeTypeOnlyImportSpecifier(candidate, contextToken)
 		if ast.IsImportDeclaration(candidate) {
-			result.isTopLevelTypeOnly = candidate.ImportClause().IsTypeOnly()
+			if importClause := candidate.ImportClause(); importClause != nil {
+				result.isTopLevelTypeOnly = importClause.IsTypeOnly()
+			}
 		} else if candidate.Kind == ast.KindImportEqualsDeclaration {
 			result.isTopLevelTypeOnly = candidate.IsTypeOnly()
 		}
--- a/internal/fourslash/tests/completionImportKeywordNoCrash_test.go
+++ b/internal/fourslash/tests/completionImportKeywordNoCrash_test.go
@@ -1,0 +1,55 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+// Test that requesting completions after a keyword in an import statement
+// doesn't crash with a nil pointer dereference.
+// Reproduces issue: import super/*cursor*/
+// Before the fix, this would panic with: runtime error: invalid memory address or nil pointer dereference
+func TestCompletionImportKeywordNoCrash(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+
+	emptyCommitChars := []string{}
+
+	// Test with "super" keyword
+	{
+		const content = `import super/*1*/`
+		f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+		f.VerifyCompletions(t, "1", &fourslash.CompletionsExpectedList{
+			IsIncomplete: false,
+			ItemDefaults: &fourslash.CompletionsExpectedItemDefaults{
+				CommitCharacters: &emptyCommitChars,
+			},
+			Items: &fourslash.CompletionsExpectedItems{
+				Includes: []fourslash.CompletionsExpectedItem{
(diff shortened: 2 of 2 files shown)


internal/fourslash/_scripts/manualTests.txt
```diff
@@ -1,66 +1,83 @@
...
-completionsWithStringReplacementMode1
-jsdocParameterNameCompletion
-stringLiteralCompletionsInPositionTypedUsingRest
-tsxCompletion12
-importNameCodeFix_uriStyleNodeCoreModules1
-importNameCodeFixDefaultExport7
-importNameCodeFixOptionalImport0
-formattingFatArrowFunctions
\ No newline at end of file
+autoCloseFragment
```
gabritto: This file was in a bad state so I had to dedupe + sort it again.

internal/astnav/tokens.go
```diff
@@ -451,6 +451,9 @@ func FindPrecedingTokenEx(sourceFile *ast.SourceFile, position int, startNode *a
 }
 
 func isValidPrecedingNode(node *ast.Node, sourceFile *ast.SourceFile) bool {
+	if node.Kind == ast.KindEndOfFile {
```
gabritto, line 454: This matches a special case we already had in Strada and fixes a crash in one of the tests.

internal/ls/completions.go
```diff
@@ -4382,18 +4381,6 @@ func (l *LanguageService) createLSPCompletionItem(
 	if kindModifiers.Has(lsutil.ScriptElementKindModifierDeprecated) {
 		tags = &[]lsproto.CompletionItemTag{lsproto.CompletionItemTagDeprecated}
 	}
-	if kind == lsproto.CompletionItemKindFile {
```
gabritto, line 4385: Moved this logic to `string_completions.go`.

internal/module/resolver.go
```diff
@@ -656,12 +656,13 @@ func (r *resolutionState) loadModuleFromTargetExportOrImport(extensions extensio
 				if isPattern {
 					combinedLookup = strings.ReplaceAll(targetString, "*", subpath)
 				}
+				scopeContainingDirectory := tspath.EnsureTrailingDirectorySeparator(scope.PackageDirectory)
```
gabritto, line 659: This fixes a crash when `scopeContainingDirectory` is `/`.

internal/testrunner/test_case_parser.go
```diff
@@ -181,8 +182,9 @@ func ParseTestFilesAndSymlinksWithOptions[T any](
 			// New metadata statement after having collected some code to go with the previous metadata
 			if currentFileName != "" {
 				// Store result file - always save for regular tests, but skip empty implicit first file for fourslash
-				shouldSaveFile := currentFileContent.Len() != 0 || !options.AllowImplicitFirstFile
+				shouldSaveFile := !options.AllowImplicitFirstFile || currentFileContent.Len() != 0 || hasSeenFile
```
gabritto, line 185: This allows us to parse empty files in tests, e.g.:
```ts
// @Filename: foo.ts
// @Filename: bar.ts
```

internal/fourslash/tests/gen/completionForStringLiteralNonrelativeImport14_test.go
andrewbranch, line 1: Why did this (and two others) get deleted without a replacement in manualTests?

gabritto: They test `baseUrl`, so I figured it didn't make sense to keep it because IIRC we have other tests that have just `paths` in the tsconfig.json.

PR #2596, opened 2026-01-27
Support FORCE_COLOR

### LINKED ISSUE #2540: Should handle `FORCE_COLOR` environment variable
Current implementation
https://github.com/microsoft/typescript-go/blob/9799455da17eea568671ff4bc1d8bab299f84a59/internal/execute/tsc/diagnostics.go#L46-L48
tries to determine whether the terminal is a TTY. This prevents us from using `exec` to get colored help text, which makes the output very ugly when creating wrappers for tsgo (such as vue-tsgo).

jakebailey:
Fixes #2540

picocolors, chalk, Node, etc, all support this.

Strada doesn't, but it seems to matter for people trying to use `tsgo`, so, I think it's worth doing. And probably sticking in Strada if we want.

### DIFF
--- a/internal/execute/tsc/diagnostics.go
+++ b/internal/execute/tsc/diagnostics.go
@@ -44,7 +44,13 @@ func CreateDiagnosticReporter(sys System, w io.Writer, locale locale.Locale, opt
 }
 
 func defaultIsPretty(sys System) bool {
-	return sys.WriteOutputIsTTY() && sys.GetEnvironmentVariable("NO_COLOR") == ""
+	if sys.GetEnvironmentVariable("NO_COLOR") != "" {
+		return false
+	}
+	if sys.GetEnvironmentVariable("FORCE_COLOR") != "" {
+		return true
+	}
+	return sys.WriteOutputIsTTY()
 }
 
 func shouldBePretty(sys System, options *core.CompilerOptions) bool {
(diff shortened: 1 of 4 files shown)


internal/execute/tsc/diagnostics.go
```diff
@@ -44,7 +44,13 @@ func CreateDiagnosticReporter(sys System, w io.Writer, locale locale.Locale, opt
 }
 
 func defaultIsPretty(sys System) bool {
-	return sys.WriteOutputIsTTY() && sys.GetEnvironmentVariable("NO_COLOR") == ""
+	if sys.GetEnvironmentVariable("NO_COLOR") != "" {
+		return false
+	}
+	if sys.GetEnvironmentVariable("FORCE_COLOR") != "" {
```
DanielRosenwasser, line 50: It's weird that we just look for any content (e.g. someone could write `FALSE`), but I don't know how other tools deal with this.

DanielRosenwasser (approved):
If all tools deal with non-empty strings in this way, seems good.

internal/execute/tsc/diagnostics.go
jakebailey: They do what this does, more or less.

- https://github.com/alexeyraspopov/picocolors/blob/0e7c4af2de299dd7bc5916f2bddd151fa2f66740/picocolors.js#L3-L4 (truthy check)
- https://github.com/nodejs/node/blob/150d154bf1ba6cc28ca6276b94117b5a2813e6c8/lib/internal/tty.js#L126 (leveled, but truthy means a nonzero level)
- https://github.com/chalk/chalk/blob/aa06bb5ac3f14df9fda8cfb54274dfc165ddfdef/source/vendor/supports-color/index.js#L35 (same as node)

PR #2597, opened 2026-01-27
internal/transformers/tstransformers: fix unqualified references across merged namespace

dtscd:
Within a single file, if multiple namespace blocks of the same namespace are present (for instance if the files have been manually joined together) symbols found in the current namespace, but previous blocks, must be qualified.

container.Contains(location) incorrectly excluded references in other declaration blocks of the same namespace or enum.

It is up to the caller to ensure files are joined together in the correct order.

### DIFF
--- a/internal/transformers/tstransforms/runtimesyntax.go
+++ b/internal/transformers/tstransforms/runtimesyntax.go
@@ -1,7 +1,5 @@
 package tstransforms
 
-// !!! Unqualified enum member references across merged enum declarations are not currently supported (e.g `enum E {A}; enum E {B=A}`)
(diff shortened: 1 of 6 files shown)


dtscd:
@microsoft-github-policy-service agree company="Solid Core Data Inc"

jakebailey:
Is this not just going away from [microsoft/TypeScript#54500](https://github.com/microsoft/TypeScript/issues/54500#:~:text=Cross%2Dnamespace%20value%20qualification)?

jakebailey:
I think this is actually already fixed now

PR #2610, opened 2026-01-29
Send path completion item detail eagerly

### LINKED ISSUE #2606: Module specifier completion icons are lazily loaded/generic until item selection
When triggering IntelliSense for module imports (#2587), the icons in the suggestion widget do not immediately reflect the correct file type (e.g., TypeScript file). Instead, they display a generic text/keyword icon.

https://github.com/user-attachments/assets/fdfbf101-bc4b-40c3-895d-ec19b6b5da10
## Behavior with `tsgo`
https://github.com/user-attachments/assets/f91544bd-2994-467e-8b28-cd5388c589e7

gabritto:
Fixes #2606.
For completion items of file kind, vscode seems to rely on a file extension present in the `detail` field of a completion item. If we send the detail only upon completion item resolve, then the icon only shows once you select the item, so we need to send those details eagerly.

We weren't always computing the extension for a given file completion item, so I've also added code to do that. However, two of the code paths look at files on disk to figure out the extension to use for the item detail, and one of them doesn't (when we find a path mapping without a wildcard, e.g. `"foo": "./bar.js"`, we use extension `.js` from the pattern, even if it turns out there's a `bar.d.ts` on disk, because we don't have an actual path from disk to look at).
So right now which extension we pick is inconsistent in that one case. If this matters, we could add a file system lookup there to figure out what extension to use, or we could consistently always try and pick the extension from the pattern (i.e. from `"./bar.js"`), though I think a pattern could have no extension suffix.

See `pathCompletionsPackageJsonExportsWildcard3_test.go` vs `pathCompletionsPackageJsonExportsBundlerNoNodeCondition_test.go` for examples.

### DIFF
--- a/internal/fourslash/tests/manual/completionsPaths_pathMapping_nonTrailingWildcard1_test.go
+++ b/internal/fourslash/tests/manual/completionsPaths_pathMapping_nonTrailingWildcard1_test.go
@@ -43,12 +43,12 @@ import {} from "foo/_dir//*3*/";
 				&lsproto.CompletionItem{
 					Label:  "_a/suffix",
 					Kind:   PtrTo(lsproto.CompletionItemKindFile),
-					Detail: PtrTo("_a/suffix"),
+					Detail: PtrTo("_a/suffix.ts"),
 				},
 				&lsproto.CompletionItem{
 					Label:  "_b/suffix",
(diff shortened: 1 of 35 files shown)


jakebailey:
Popped out of the merge queue because of `declarationMapsRenameWithProjectReferencesEdit`, which I think is flaky?

PR #2613, opened 2026-01-29
Implement telemetry reporting in extension

DanielRosenwasser:
This change adds telemetry reporting infrastructure to help capture usage and error information. It follows [VS Code's guidelines for extension authors](https://code.visualstudio.com/api/extension-guides/telemetry) in order to respect user preferences (through using `@vscode/extension-telemetry` which takes care of all of that).

While most events are triggered via the client, the server now pushes events on recovered request failures. At a broad level, server telemetry lines up with the information our reporters take: an event name, properties, and measurements, along with the purpose (used to respect user preferences).

In the future we'll likely want to keep track of measurements on specific language server commands to keep an eye on performance issues, along with certain other server-recovered crashes.

### DIFF
--- a/_extension/src/telemetryReporting.ts
+++ b/_extension/src/telemetryReporting.ts
@@ -10,10 +10,10 @@ import type { TelemetryReporter as VSCodeTelemetryReporter } from "@vscode/exten
 // The "untyped" variants are provided for when properties/measurements are not known
 // from the editor client - for example, when forwarding telemetry events from the language server.
 export interface TelemetryReporter {
-    sendTelemetryEvent(eventName: "cmd.enableNativePreview"): void;
-    sendTelemetryEvent(eventName: "cmd.disableNativePreview"): void;
-    sendTelemetryEvent(eventName: "cmd.restartLanguageServer"): void;
-    sendTelemetryEvent(eventName: "cmd.reportIssue"): void;
+    sendTelemetryEvent(eventName: "command.enableNativePreview"): void;
+    sendTelemetryEvent(eventName: "command.disableNativePreview"): void;
+    sendTelemetryEvent(eventName: "command.restartLanguageServer"): void;
+    sendTelemetryEvent(eventName: "command.reportIssue"): void;
     sendTelemetryEvent(eventName: "languageServer.start", data: LSServerStart): void;
 
     sendTelemetryErrorEvent(eventName: "languageServer.connectionError", data: LSConnectionError): void;
--- a/_extension/src/commands.ts
+++ b/_extension/src/commands.ts
@@ -13,13 +13,13 @@ import { restartExtHostOnChangeIfNeeded } from "./util";
 export function registerEnablementCommands(context: vscode.ExtensionContext, telemetryReporter: tr.TelemetryReporter): void {
     context.subscriptions.push(vscode.commands.registerCommand("typescript.native-preview.enable", () => {
         // Fire and forget, because this will restart the extension host and cause an error if we await
-        telemetryReporter.sendTelemetryEvent("cmd.enableNativePreview");
+        telemetryReporter.sendTelemetryEvent("command.enableNativePreview");
         updateUseTsgoSetting(true);
     }));
 
     context.subscriptions.push(vscode.commands.registerCommand("typescript.native-preview.disable", () => {
         // Fire and forget, because this will restart the extension host and cause an error if we await
-        telemetryReporter.sendTelemetryEvent("cmd.disableNativePreview");
+        telemetryReporter.sendTelemetryEvent("command.disableNativePreview");
         updateUseTsgoSetting(false);
     }));
 }
@@ -34,7 +34,7 @@ export function registerLanguageCommands(
     const disposables: vscode.Disposable[] = [];
 
     disposables.push(vscode.commands.registerCommand("typescript.native-preview.restart", () => {
-        telemetryReporter.sendTelemetryEvent("cmd.restartLanguageServer");
+        telemetryReporter.sendTelemetryEvent("command.restartLanguageServer");
         return client.restart(context);
     }));
 
@@ -52,7 +52,7 @@ export function registerLanguageCommands(
     disposables.push(vscode.commands.registerCommand("typescript.native-preview.showMenu", showCommands));
 
     disposables.push(vscode.commands.registerCommand("typescript.native-preview.reportIssue", () => {
-        telemetryReporter.sendTelemetryEvent("cmd.reportIssue");
+        telemetryReporter.sendTelemetryEvent("command.reportIssue");
         vscode.commands.executeCommand("workbench.action.openIssueReporter", {
             extensionId: "TypeScriptTeam.native-preview",
         });


_extension/package.json
```diff
@@ -160,11 +160,12 @@
...
+        "@vscode/extension-telemetry": "^1.2.0",
         "vscode-languageclient": "^10.0.0-next.19"
     },
     "devDependencies": {
         "@types/vscode": "^1.106.1",
         "@vscode/vsce": "^3.7.1",
         "esbuild": "^0.27.1"
     }
-}
+}
```
jakebailey: please sir my newline

_extension/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es2020",
+        "target": "es2024",
```
jakebailey: why the change here?

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -21875,6 +21875,172 @@ func (s *CustomClosingTagCompletion) UnmarshalJSONFrom(dec *jsontext.Decoder) er
 	return nil
 }
 
+// TelemetryEvent contains information about a telemetry event.
+type TelemetryEvent struct {
```
jakebailey: Are we missing the code that overrides `telemetry/event`'s property to this type? I'm not seeing its `Info` var updated, which means it won't roundtrip.

internal/lsp/stack_sanitizer_test.go
```diff
@@ -0,0 +1,56 @@
+package lsp
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/testutil/baseline"
+)
+
+func TestSanitizedStackTraceCompletionsRequest(t *testing.T) {
```
jakebailey: Can you also add tests for the prod build, given that is what everyone will be using?

_extension/package.json
DanielRosenwasser: dprint does this

jakebailey: _removes_ a newline? certainly not

_extension/tsconfig.json
DanielRosenwasser: I previously wanted it for `replaceAll`, but I don't think I use that anymore. But why do we need a target this far back anyway? I guess maybe an eventual vscode.dev?

I can change it back if you want.

```suggestion
        "target": "es2020",
```

internal/lsp/lsproto/lsp_generated.go
DanielRosenwasser: Do we have any examples where an Info's type argument is overridden?

_extension/tsconfig.json
jakebailey: I think we need ES2023, given VS Code is Node 22.

_extension/package.json
DanielRosenwasser: Nope, I was wrong. This was in my `settings.json`.

```
    "[json]": {
        "editor.defaultFormatter": "vscode.json-language-features"
    }
```

internal/lsp/lsproto/lsp_generated.go
jakebailey: Yeah, `InitializationOptions` used to be `any`, but is now `InitializationOptions`.

jakebailey: this bit:

```ts
            // Replace initializationOptions type with custom InitializationOptions
            if (prop.name === "initializationOptions" && prop.type.kind === "reference" && prop.type.name === "LSPAny") {
                prop.type = { kind: "reference", name: "InitializationOptions" };
            }
```

internal/lsp/server.go
```diff
@@ -776,6 +777,16 @@ func (s *Server) recover(ctx context.Context, req *lsproto.RequestMessage) {
...
+			telMsg := lsproto.TelemetryEventInfo.NewNotificationMessage(&lsproto.TelemetryEvent{
+				EventName:        "languageServer.errorResponse",
+				TelemetryPurpose: lsproto.TelemetryPurposeError,
+				Properties: &lsproto.RequestFailureTelemetryProperties{
+					ErrorCode:     lsproto.ErrorCodeInternalError.String(),
+					RequestMethod: string(req.Method),
+					Stack:         sanitizeStackTrace(string(stack)),
+				},
+			})
+			s.send(telMsg.Message())
```
jakebailey: We should be able to use a type safe function for this, I think.

DanielRosenwasser: Do you just mean

```
func (s *Server) sendNotification(notification *lsproto.RequestMessage) error {
	return s.send(notification.Message())
}
```

?

_extension/src/client.ts
```diff
@@ -150,7 +166,29 @@ export class Client {
...
+                case "general":
+                    this.telemetryReporter.sendTelemetryEventUntyped(d.eventName, d.properties, d.measurements);
+                    break;
+                case "error":
+                    this.telemetryReporter.sendTelemetryErrorEventUntyped(d.eventName, d.properties, d.measurements);
+                    break;
+                default:
+                    const _: never = d.telemetryPurpose;
+                    break;
+            }
```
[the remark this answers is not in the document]
DanielRosenwasser: It would also be a decent idea to log unexpected values from the server instead of just returning.

internal/lsp/server.go
jakebailey: I mean that we should add a top level func based on `sendClientRequest` that forces a specific argument type, so we don't feel tempted to use `s.send` directly. But, `PublishDiagnostics` doesn't do this yet, so it's probably okay

_extension/src/client.ts
```diff
@@ -150,7 +166,32 @@ export class Client {
...
+                    this.telemetryReporter.sendTelemetryEventUntyped(d.eventName, d.properties, d.measurements);
+                    break;
+                case "error":
+                    this.telemetryReporter.sendTelemetryErrorEventUntyped(d.eventName, d.properties, d.measurements);
+                    break;
+                default:
+                    const _: never = d.telemetryPurpose;
+                    this.telemetryReporter.sendTelemetryErrorEvent("languageServer.unexpectedTelemetryPurpose", {
+                        telemetryPurpose: String(d.telemetryPurpose),
+                    });
```
jakebailey, line 188: This could be random data; do we actually need to collect this?

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -163,6 +163,53 @@ const customStructures: Structure[] = [
...
+            {
+                name: "properties",
+                type: { kind: "reference", name: "LSPAny" },
+                documentation: "The properties associated with the event.",
+            },
+            {
+                name: "measurements",
+                type: { kind: "reference", name: "LSPAny" },
+                documentation: "The measurements associated with the event.",
+            },
```
jakebailey: Rather than using any here, can we make the entire TelemetryEvent a union of events, and then define specific types for properties/measurements? That way we don't mess it up?

Then you can use a literal for the event name and purpose and not have to write those down somewhere explicitly.

internal/lsp/stack_sanitizer.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		i := 0
+		// Skip whitespace
+		for range line {
+			if line[i] != ' ' && line[i] != '\t' {
+				break
+			}
+			i++
+		}
```
jakebailey, line 32: I feel like this could just be a trim, but, it's ok

DanielRosenwasser: It's mostly to preserve formatting so the stack trace continues to be readable.

internal/lsp/lsproto/_generate/generate.mts
DanielRosenwasser: It feels a bit overkill at this point - we're just lobbing properties over and there's only one event right now. But I can do it.

_extension/src/client.ts
DanielRosenwasser: Yeah, I wasn't sure - but it indicates some sort of mismatch between client/server and it'd be nice to find out about it somehow without going crazy to keep client/server in sync.

internal/lsp/server.go
DanielRosenwasser: I added a `sendNotification` - I felt like `sendClientRequest` would have been a little weird here because it has no corresponding response. We can adjust later.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -163,6 +163,59 @@ const customStructures: Structure[] = [
...
+                name: "telemetryPurpose",
+                type: { kind: "reference", name: "TelemetryPurpose" },
+                documentation: "Indicates whether the reason for generating the event (e.g. general usage telemetry or errors).",
+            },
+            {
+                // Currently, only RequestFailureTelemetryProperties is sent.
+                // Update to a union if needed in the future.
+                name: "properties",
+                optional: true,
+                type: { kind: "reference", name: "RequestFailureTelemetryProperties" },
```
jakebailey, line 182: This isn't the event I was trying to imply; I was trying to say to make `TelemetryEvent` itself a union via `or`, and then fully type each event with its properties/measurements.

internal/lsp/server.go
```diff
@@ -508,6 +509,10 @@ func (s *Server) sendError(id *lsproto.ID, err error) error {
 	})
 }
 
+func (s *Server) sendNotification(notification *lsproto.RequestMessage) error {
+	return s.send(notification.Message())
+}
```
jakebailey:
```suggestion
func sendNotification[Params any](s *Server, info lsproto.NotificationInfo[Params], params Param) error {
	return s.send(info.NewNotificationMessage(params).Message())
}
```

_extension/src/client.ts
jakebailey: I'd probably err on the side of not doing anything; whatever we do will have to be forward compatible anyway...

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -213,6 +255,14 @@ const customEnumerations: Enumeration[] = [
             { name: "NotAllowed", value: 4, documentation: "Import cannot be marked type-only." },
         ],
     },
+    {
+        name: "TelemetryPurpose",
```
jakebailey: This is unused now, I believe.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -385,6 +435,18 @@ function patchAndPreprocessModel() {
         }
     }
 
+    for (const notification of model.notifications) {
+        if (notification.typeName === "TelemetryEventNotification") {
+            notification.params = {
```
jakebailey, line 446: We should give this a friendly name and then ref it here

DanielRosenwasser: Where would you recommend I do that? Add it to `model.typeAliases`?

jakebailey: Yeah, an alias would work to define it, I think.

_extension/src/telemetryReporting.ts
```diff
@@ -0,0 +1,77 @@
...
+// This helps ensure that the telemetry events used in the codebase are
+// properly typed, and that properties/measurements are never forgotten.
+//
+// The difference between `sendTelemetryEvent` and `sendTelemetryErrorEvent` is that
+// these methods respect user preferences around reporting (i.e. `telemetry.telemetryLevel`).
+//
+// The "untyped" variants are provided for when properties/measurements are not known
+// from the editor client - for example, when forwarding telemetry events from the language server.
+export interface TelemetryReporter {
+    sendTelemetryEvent(eventName: "cmd.enableNativePreview"): void;
```
jakebailey: I feel like these could be "command." And not abbreviated but it's probably ok

internal/lsp/server.go
```diff
@@ -508,6 +509,10 @@ func (s *Server) sendError(id *lsproto.ID, err error) error {
 	})
 }
 
+func sendNotification[Params any](s *Server, info lsproto.NotificationInfo[Params], params Params) error {
```
jakebailey, line 511: This could now be used for the other notification send but not important

jakebailey (approved):

jakebailey (approved):

Issue #2611, opened 2026-01-29
typeRoots behavior difference from tsc v9 (vitest/globals)


sammarks-stukent:
We are currently attempting to adopt the preview on a large monorepo with 600k LOC of TypeScript, and are running into tons of errors any time vitest globals are referenced.

I ran that build with `--traceResolution` and found that it wasn't attempting to import the `vitest/globals` export unless I imported it explicitly inside the code somewhere. It's currently configured using `types` inside `tsconfig.json`.

Luckily, I managed to get a repro of the issue in a small repo (at least on my machine).

According to the docs it _seems_ like I have `typeRoots` configured properly, but I wouldn't put it past myself to be missing one small detail that causes this to happen 😜

## Steps to reproduce

1. Clone https://github.com/sammarks-stukent/tsgo-typeroots
2. Install with `pnpm install` (I have tried installing dependencies with both `pnpm` and `npm` to rule out symlink issues, and the behavior appears to be the same across both).
3. Run `pnpm tsgo` and get error saying `vi` is not defined.
4. Run `pnpm tsc` and it builds successfully.
5. Remove `typeRoots` from `tsconfig.json`
6. Run `pnpm tsgo` and it builds successfully.

Tested with preview v7.0.0-dev.20260129.1, and with typescript v5.9.3.

I created a GitHub Actions workflow to run the repro steps in a different environment than my machine, and that appears to show the same issue. [Here is the most recent run illustrating the problem.](https://github.com/sammarks-stukent/tsgo-typeroots/actions/runs/21491063227/job/61913403376)

## Behavior with `typescript@5.9`

Builds successfully.

## Behavior with `tsgo`

```
src/index.ts:2:3 - error TS2304: Cannot find name 'vi'.

2   vi.doMock('something')
    ~~


Found 1 error in src/index.ts:2
```

sammarks-stukent:
Might be related to https://github.com/microsoft/typescript-go/issues/2147

PR #2614, opened 2026-01-29
fix: use typeRoot instead of candidate when resolving types references from config

sammarks-stukent:
Resolves https://github.com/microsoft/typescript-go/issues/2611.

**AI Disclaimer:** I used OpenCode to generate this PR, as I do not have much experience with go. The solution it came up with seemed to resolve the issue, and seemed simple enough that I figured I would go ahead and put up a PR for it just in case. If this PR is completely off base, feel free to summarily disregard it; you will not hurt my feelings :)

The solution could be _so simple_ that it ends up breaking other things, but it appears the full test suite still passes so that gives me some more confidence.

I also had the agent go back and create a test to protect against regressions for this in the future. I have verified locally that the test case fails whenever undoing the fix.

After running the full suite with the main TS repo cloned, it looked like a few tests inside there started failing. I gave it to OpenCode and had it fix, and it looks like it discovered updating the baselines for those tests is what we're looking for:

> A reduction/removal of .diff file baselines is ideal because it indicates the port has converged in behavior with the stable TypeScript codebase.

So hopefully the removal / update of those other tests makes sense!

Below is the original explanation the agent output for the fix:

---

## Problem

When `typeRoots` is specified in `tsconfig.json`, type reference directives (like `vitest/globals`) fail to resolve, even though they work correctly with the standard TypeScript compiler (tsc).

**Example failing configuration:**
```json
{
  "compilerOptions": {
    "types": ["vitest/globals"],
    "typeRoots": ["./node_modules"]
  }
}
```

**Error:**
```
src/index.ts:2:3 - error TS2304: Cannot find name 'vi'.
```

## Root Cause

In `internal/module/resolver.go`, the `resolveTypeReferenceDirective` function incorrectly checks if the **candidate** directory exists instead of checking if the **typeRoot** directory exists.

**Bug (line 340):**
```go
directoryExists := r.resolver.host.FS().DirectoryExists(candidate)
```

When resolving `vitest/globals` with typeRoot `./node_modules`:
- `candidate` = `./node_modules/vitest/globals`
- The candidate path doesn't exist as a directory (it resolves to a `.d.ts` file)
- `directoryExists` becomes `false`
- This causes `onlyRecordFailures=true` to be passed to the load functions
- Resolution fails because it only records failures instead of actually trying to resolve

**TypeScript reference implementation** (`src/compiler/moduleNameResolver.ts`):
```typescript
const directoryExists = directoryProbablyExists(typeRoot, host); // Checks TYPE ROOT, not candidate
```

## Fix

Change line 340 in `internal/module/resolver.go` from:
```go
directoryExists := r.resolver.host.FS().DirectoryExists(candidate)
```

To:
```go
directoryExists := r.resolver.host.FS().DirectoryExists(typeRoot)
```

This matches the TypeScript behavior where we check if the typeRoot directory exists, then attempt to resolve the candidate within it.

## Verification

1. Build passes: `npx hereby build`
2. All tests pass: `npx hereby test`
3. Lint passes: `npx hereby lint`
4. Format passes: `npx hereby format`
5. Manual verification with test project:
   - `vitest/globals` now resolves successfully
   - Trace shows: `Type reference directive 'vitest/globals' was successfully resolved to '.../vitest/globals.d.ts'`

## Files Changed

- `internal/module/resolver.go` - Single line fix (line 340)


sammarks-stukent:
@microsoft-github-policy-service agree company="Stukent"

andrewbranch (commented):
Thanks! Looks like this is correct. Since existing test baselines were modified in the right direction (diffs deleted), we don’t need new test coverage for this; would you mind deleting the new test?

sammarks-stukent:
@andrewbranch for sure! Extra tests have been removed 👍

sheetalkamat (approved):

PR #2617, opened 2026-01-30
ExplainFiles to handle duplicated files 

sheetalkamat:


internal/compiler/filesparser.go
```diff
@@ -401,7 +406,7 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
 			if task.libFile != nil {
 				libFiles = append(libFiles, file)
 				libFilesMap[path] = task.libFile
-			} else if existingCanonicalPath == "" {
+			} else {
 				files = append(files, file)
```
jakebailey, line 410: If we don't skip the files here, we will attempt to run the checker on these files, no?

sheetalkamat: The duplicated file is added explicitly and we continue above where we do that .. that ensures that we aren’t copying preprocessing diagnostics and stuff for this task ( only include reason is needed )

jakebailey: I'm not sure that answers my question; later on when we ask for diags of all files, we loop over this files list and ask for each ones diags. But I don't think we have anything that prevents us from doing that multiple times, and having the same file in the list twice will associate a file with multiple checkers

sheetalkamat: What I mean is that duplicate files aren’t added to files list ..

internal/compiler/filesparser.go
```diff
@@ -349,22 +349,31 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
...
+						fileName: task.normalizedFilePath,
+						path:     task.path,
+						target:   packageIdFile.Path(),
+					}
+					filesByPath[task.path] = packageIdFile
+					if data.lowestDepth > 0 {
+						sourceFilesFoundSearchingNodeModules.Add(task.path)
+					}
+					continue
+				} else if file != nil {
```
sheetalkamat, line 370: This is where we skip rest of the stuff for duplicate file , hence skipping adding file again to files list

jakebailey: Ah, I missed the continue.

PR #2620, opened 2026-01-30
Bring back async API, allow API connection to LSP process

### LINKED ISSUE #2630: Multiple Language Client instances persist after TSDK switch
### Extension Version

7.0.0-dev.20260201.1

### VS Code Version

1.108.2

### Operating system Version

Windows

### Steps to reproduce

When I set up `typescript.native-preview.tsdk": "./node_modules/@typescript/native-preview`, I get:

Image

When I use Format Document With... we can also see that multiple Typescript (Native Preview) are present:

Image

### Issue

The fix implemented in PR #2451 does not appear to have fully resolved the issues reported in #2369. I believe it was closed prematurely, as the previous fix only addressed #2399.

andrewbranch:
- Brings back the async, JSON-RPC-based API client that existed in intermediate commits of #711 but was ultimately deleted.
- Adds a proper JSON-RPC API server for said async client (rather than abusing the LSP server as the prototype in #711 did).
- The new API server stack is less tightly coupled between method handlers and transport/protocol, so now a single implementation of API handlers and session management (`api.Session`) can be used over either STDIO or UDS / named pipe, and communicate over either JSON-RPC or the custom sync protocol we built on top of MessagePack.
- The LSP now supports a custom command that initializes an (async) API server listening on a UDS / named pipe. In this mode, the data backing the API server is the Snapshot owned by the LSP session. Our VS Code extension exposes a command that calls this API initialization method and returns the API connection, which will allow third party extensions and LSP servers to query TypeScript’s LSP state.

The API surface area is still **early prototype quality and very incomplete**; this PR doesn’t add any methods. But the infrastructure provides the beginnings of a path toward replacing TS Server plugins, and will hopefully make it less cumbersome to maintain both a sync and async version of the API. (It’s still pretty annoying in the TS client code, thanks to async function coloring in JavaScript. Possibly future codegen can help us out a little there.)

Also incidentally fixes #2630, as I needed some additional features out of the extension as I started to play with this, which were difficult to add without fixing the lifetime issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,11 +85,15 @@
         },
         "node_modules/@azu/format-text": {
             "version": "1.0.2",
+            "resolved": "https://registry.npmjs.org/@azu/format-text/-/format-text-1.0.2.tgz",
+            "integrity": "sha512-Swi4N7Edy1Eqq82GxgEECXSSLyn6GOb5htRFPzBDdUkECGXtlf12ynO5oJSpWKPwCaUssOu7NfhDcCWpIC6Ywg==",
             "dev": true,
             "license": "BSD-3-Clause"
         },
         "node_modules/@azu/style-format": {
             "version": "1.0.1",
+            "resolved": "https://registry.npmjs.org/@azu/style-format/-/style-format-1.0.1.tgz",
+            "integrity": "sha512-AHcTojlNBdD/3/KxIKlg8sxIWHfOtQszLvOpagLTO+bjC3u7SAszu1lf//u7JJC50aUSH+BVWDD/KvaA6Gfn5g==",
             "dev": true,
             "license": "WTFPL",
             "dependencies": {
@@ -98,6 +102,8 @@
         },
         "node_modules/@azure/abort-controller": {
             "version": "2.1.2",
+            "resolved": "https://registry.npmjs.org/@azure/abort-controller/-/abort-controller-2.1.2.tgz",
+            "integrity": "sha512-nBrLsEWm4J2u5LpAPjxADTlq3trDgVZZXHNKabeXZtpq3d3AbN/KGO82R87rdDz5/lYB024rtEf10/q0urNgsA==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -109,6 +115,8 @@
         },
         "node_modules/@azure/core-auth": {
             "version": "1.10.1",
+            "resolved": "https://registry.npmjs.org/@azure/core-auth/-/core-auth-1.10.1.tgz",
+            "integrity": "sha512-ykRMW8PjVAn+RS6ww5cmK9U2CyH9p4Q88YJwvUslfuMmN98w/2rdGRLPqJYObapBCdzBVeDgYWdJnFPFb7qzpg==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -122,6 +130,8 @@
         },
         "node_modules/@azure/core-client": {
             "version": "1.10.1",
+            "resolved": "https://registry.npmjs.org/@azure/core-client/-/core-client-1.10.1.tgz",
+            "integrity": "sha512-Nh5PhEOeY6PrnxNPsEHRr9eimxLwgLlpmguQaHKBinFYA/RU9+kOYVOQqOrTsCL+KSxrLLl1gD8Dk5BFW/7l/w==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -139,6 +149,8 @@
         },
         "node_modules/@azure/core-rest-pipeline": {
             "version": "1.22.2",
+            "resolved": "https://registry.npmjs.org/@azure/core-rest-pipeline/-/core-rest-pipeline-1.22.2.tgz",
+            "integrity": "sha512-MzHym+wOi8CLUlKCQu12de0nwcq9k9Kuv43j4Wa++CsCpJwps2eeBQwD2Bu8snkxTtDKDx4GwjuR9E8yC8LNrg==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -156,6 +168,8 @@
         },
         "node_modules/@azure/core-tracing": {
             "version": "1.3.1",
+            "resolved": "https://registry.npmjs.org/@azure/core-tracing/-/core-tracing-1.3.1.tgz",
+            "integrity": "sha512-9MWKevR7Hz8kNzzPLfX4EAtGM2b8mr50HPDBvio96bURP/9C+HjdH3sBlLSNNrvRAr5/k/svoH457gB5IKpmwQ==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -167,6 +181,8 @@
         },
         "node_modules/@azure/core-util": {
             "version": "1.13.1",
+            "resolved": "https://registry.npmjs.org/@azure/core-util/-/core-util-1.13.1.tgz",
+            "integrity": "sha512-XPArKLzsvl0Hf0CaGyKHUyVgF7oDnhKoP85Xv6M4StF/1AhfORhZudHtOyf2s+FcbuQ9dPRAjB8J2KvRRMUK2A==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -180,6 +196,8 @@
         },
         "node_modules/@azure/identity": {
             "version": "4.13.0",
+            "resolved": "https://registry.npmjs.org/@azure/identity/-/identity-4.13.0.tgz",
+            "integrity": "sha512-uWC0fssc+hs1TGGVkkghiaFkkS7NkTxfnCH+Hdg+yTehTpMcehpok4PgUKKdyCH+9ldu6FhiHRv84Ntqj1vVcw==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -201,6 +219,8 @@
         },
         "node_modules/@azure/logger": {
             "version": "1.3.0",
+            "resolved": "https://registry.npmjs.org/@azure/logger/-/logger-1.3.0.tgz",
+            "integrity": "sha512-fCqPIfOcLE+CGqGPd66c8bZpwAji98tZ4JI9i/mlTNTlsIWslCfpg48s/ypyLxZTump5sypjrKn2/kY7q8oAbA==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -213,6 +233,8 @@
         },
         "node_modules/@azure/msal-browser": {
             "version": "4.28.1",
+            "resolved": "https://registry.npmjs.org/@azure/msal-browser/-/msal-browser-4.28.1.tgz",
+            "integrity": "sha512-al2u2fTchbClq3L4C1NlqLm+vwKfhYCPtZN2LR/9xJVaQ4Mnrwf5vANvuyPSJHcGvw50UBmhuVmYUAhTEetTpA==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -224,6 +246,8 @@
         },
         "node_modules/@azure/msal-common": {
             "version": "15.14.1",
+            "resolved": "https://registry.npmjs.org/@azure/msal-common/-/msal-common-15.14.1.tgz",
+            "integrity": "sha512-IkzF7Pywt6QKTS0kwdCv/XV8x8JXknZDvSjj/IccooxnP373T5jaadO3FnOrbWo3S0UqkfIDyZNTaQ/oAgRdXw==",
             "dev": true,
             "license": "MIT",
             "engines": {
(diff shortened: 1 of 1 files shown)


jakebailey:
> It’s still pretty annoying in the TS client code, thanks to async function coloring in JavaScript. Possibly future codegen can help us out a little there.

If we wanted to be very clever, we could do what babel does and use [`gensync`](https://www.npmjs.com/package/gensync) (which I am biased towards having written their types).

internal/api/callbackfs.go
```diff
@@ -0,0 +1,198 @@
...
+	CbRealpath             = "realpath"
+)
+
+// NewCallbackFS creates a new CallbackFS wrapping the given base filesystem.
+// The callbacks slice specifies which filesystem operations should be delegated
+// to the client (e.g., "readFile", "fileExists").
+func NewCallbackFS(base vfs.FS, callbacks []string) *CallbackFS {
+	enabled := make(map[string]bool, len(callbacks))
+	for _, cb := range callbacks {
+		enabled[cb] = true
```
gabritto, line 59: Should we validate that a known callback was passed here?

gabritto (approved):
Code looks ok to me, though I can't confidently review the protocol/connection changes.

A high level question I have: why is the user of the API responsible for releasing objects? How would that work?

andrewbranch:
> why is the user of the API responsible for releasing objects?

If you want to be able to do a sequence like

```ts
const symbol = await project.getSymbolAtPosition(file, pos);
const type = await project.getTypeOfSymbol(symbol);
const signatures = await type.getSignatures();
```

where each of those is a remote API call, the API needs to store its representation of `symbol` and `type` in a map, because there's no master lookup table for symbols/types in programs, checkers, source files, or anywhere. If you want to ask a follow-up question about a handle to a symbol you just queried, the server would either have to re-fetch the symbol the same way you fetched it the first time (how would it encode and remember that? The logic for retrieval of these things can be non-trivial.) or you’d have to store it in a side table by an id that gets passed back and forth between client and server. So if the server is going to store stuff _in case the client queries it again_, the client has to communicate when it no longer needs those things.

> How would that work?

```ts
using symbol = await project.getSymbolAtPosition(file, pos);
using type = await project.getTypeOfSymbol(symbol);
using signatures = await type.getSignatures();
```

will automatically clean those resources up once they go out of lexical scope.

gabritto:
> > why is the user of the API responsible for releasing objects?
> 
> If you want to be able to do a sequence like
> 
> ```ts
> const symbol = await project.getSymbolAtPosition(file, pos);
> const type = await project.getTypeOfSymbol(symbol);
> const signatures = await type.getSignatures();
> ```
> 
> where each of those is a remote API call, the API needs to store its representation of `symbol` and `type` in a map, because there's no master lookup table for symbols/types in programs, checkers, source files, or anywhere. If you want to ask a follow-up question about a handle to a symbol you just queried, the server would either have to re-fetch the symbol the same way you fetched it the first time (how would it encode and remember that? The logic for retrieval of these things can be non-trivial.) or you’d have to store it in a side table by an id that gets passed back and forth between client and server. So if the server is going to store stuff _in case the client queries it again_, the client has to communicate when it no longer needs those things.
> 
> > How would that work?
> 
> ```ts
> using symbol = await project.getSymbolAtPosition(file, pos);
> using type = await project.getTypeOfSymbol(symbol);
> using signatures = await type.getSignatures();
> ```
> 
> will automatically clean those resources up once they go out of lexical scope.

Ok, that makes sense. I think what I was wondering about is that as soon as there's a change in the project, it's possible none of the objects the API returned are usable anymore, and then I don't know what happens.

andrewbranch:
Sort of true—this isn’t included yet, but I think we’re going to leave the client in control of when it wants to adopt the latest snapshot from the LSP session. So it’s not like changes in the LSP are going to rug-pull the API client’s ability to ask about a type mid-query; instead, the client can hang onto a snapshot as long as it wants and then when it’s ready, trigger the API server to the latest state. At that point, some of the retained objects might be stale and others might not be—source files, along with binder-produced symbols, will be reused. I’m not sure if there’s a great way to purge objects that are really gone—thought about using `weak` but decided to keep it simple and keep the client in control for now. (These were decisions I made in #711, not here.) It’s not super different from in Strada, where you could get a type, make a new checker, and do something invalid with the old type and the new checker. The only difference here is you have to use some different syntax to clean up automatically, but you can easily hold onto stuff you shouldn’t in either system.

_extension/src/client.ts
```diff
@@ -195,21 +198,15 @@ export class Client {
...
         if (this.isDisposed) {
             return;
         }
         this.isDisposed = true;
 
-        this.client?.dispose();
+        await this.client?.dispose();
         while (this.disposables.length > 0) {
             const d = this.disposables.pop()!;
             d.dispose();
```
RyanCavanaugh: Is this missing an `await` too?

andrewbranch: I don't _think_ these are actually async, but it's too annoying to try to keep track. I'll push a change that awaits all of them. Also, I'll clean up this mutation to `this.disposables`, because we (as of this PR at least, not sure about before) always throw away the client right after disposing it.

It looks like @DanielRosenwasser was very careful to avoid the possibility of anything getting disposed twice, but I’m not sure if that actually matters. Is there something I'm missing, Daniel?

andrewbranch:
What? What version of tinybench did it install? 3.1.1 doesn't even have that many lines. How could it have installed anything else?

Edit: Jake changed it in main 😄

internal/api/transport_unix.go
```diff
@@ -0,0 +1,22 @@
+//go:build !windows
+
+package api
+
+import (
+	"net"
```
jakebailey, line 6: I checked, and building this binary on Linux for Linux causes us to produce a dynamically linked binary:

```
file ./built/local/tsgo        
./built/local/tsgo: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=a060a83df11b7a44a2bb262877c3325b828826fa, with debug_info, not stripped
```

We're going to have to set `CGO_ENABLED=0` somewhere in `Herebyfile.mjs` to prevent this from happening for the releases, otherwise we'll produce binaries that don't work across Linux systems or on Alpine.

jakebailey: Here's a suggested patch:

```patch
diff --git a/Herebyfile.mjs b/Herebyfile.mjs
index 0489a5b70..741f83963 100644
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -104,6 +104,10 @@ const goBuildFlags = [
     ...(options.debug ? ["-gcflags=all=-N -l"] : []),
 ];
 
+const goBuildEnv = {
+    CGO_ENABLED: "0",
+};
+
 /**
  * @template T
  * @param {() => T} fn
@@ -192,7 +196,8 @@ export const lib = task({
 function buildTsgo(opts) {
     opts ||= {};
-    return $({ cancelSignal: opts.abortSignal, env: opts.env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${options.debug ? goBuildTags("noembed") : goBuildTags("noembed", "release")} -o ${out} ./cmd/tsgo`;
+    const env = { ...goBuildEnv, ...opts.env };
+    return $({ cancelSignal: opts.abortSignal, env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${options.debug ? goBuildTags("noembed") : goBuildTags("noembed", "release")} -o ${out} ./cmd/tsgo`;
 }
 
 export const tsgoBuild = task({
@@ -297,6 +302,7 @@ function goTestFlags(taskName) {
 }
 
 const goTestEnv = {
+    ...goBuildEnv,
     ...(options.concurrentTestPrograms ? { TS_TEST_PROGRAM_SINGLE_THREADED: "false" } : {}),
     // Go test caching takes a long time on Windows.
     // https://github.com/golang/go/issues/72992
```

_extension/package.json
```diff
@@ -141,6 +141,12 @@
                 "title": "TypeScript Native Preview: Stop CPU Profile",
                 "enablement": "typescript.native-preview.cpuProfileRunning",
                 "category": "Developer"
+            },
+            {
+                "command": "typescript.native-preview.initializeAPIConnection.ui",
+                "title": "Initialize API Connection",
+                "enablement": "typescript.native-preview.serverRunning",
```
jakebailey, line 148: I don't think we actually need to declare this; having it here exposes it in the UI which I don't think will work as expected, and commands not in package.json are still runnable IIRC.

_packages/api/src/async/api.ts
```diff
@@ -0,0 +1,265 @@
+/// <reference path="../node.ts" preserve="true" />
```
jakebailey, line 1: I don't see node.ts in this PR

andrewbranch: It's not in this PR but it's there 😄

_packages/api/src/path.ts
```diff
@@ -134,3 +134,186 @@ function pathComponents(path: string, rootLength: number) {
 function lastOrUndefined<T>(array: T[]): T | undefined {
     return array.length ? array[array.length - 1] : undefined;
 }
+
+const bundledScheme = "bundled:///";
```
jakebailey, line 138: Does this actually happen? IIRC we only do bundling for testing, and the release binary doesn't have it.

_extension/package.json
andrewbranch: I put it in the UI explicitly for developer/debugging convenience. It does work as expected; I’ve been using it.

_packages/api/src/async/api.ts
jakebailey: I typo'd my `find` command and missed it, sigh

_packages/api/src/path.ts
andrewbranch: Probably not; this was a Copilot 1:1 port.

_extension/package.json
jakebailey: Ah, I assumed wrongly that it required args.

cmd/tsgo/api.go
```diff
@@ -15,22 +17,35 @@ import (
...
 		Cwd:                *cwd,
 		DefaultLibraryPath: defaultLibraryPath,
+		Callbacks:          callbacksList,
+		Async:              *async,
 	})
 
-	if err := s.Run(); err != nil && !errors.Is(err, io.EOF) {
-		fmt.Println(err)
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
```
jakebailey, line 45: Not in this PR but we should totally lift this out to `func main()` eventually, since everything should begin like this

_extension/package.json
andrewbranch: The pipe path is optional; one is generated for you if you omit it. Also note this command is actually `.ui` which I use to show an info message with a "Copy to clipboard" button. We can remove it at some point before stable launch or maybe put it behind a preference.

internal/api/conn_async.go
```diff
@@ -0,0 +1,182 @@
...
+		defer c.pendingMu.Unlock()
+		if ch, ok := c.pending[*id]; ok {
+			close(ch)
+			delete(c.pending, *id)
+		}
+	}()
+
+	// Send the request
+	c.writeMu.Lock()
+	err := c.protocol.WriteRequest(id, method, params)
```
jakebailey, line 159: This can't panic, right?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/microsoft/typescript-go
 go 1.25
 
 require (
+	github.com/Microsoft/go-winio v0.6.2
```
jakebailey, line 6: After merging, we'll need to update NOTICE (I'll do it, no worries)

package-lock.json
```diff
@@ -284,8 +258,6 @@
         },
         "node_modules/@babel/helper-validator-identifier": {
             "version": "7.28.5",
-            "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.28.5.tgz",
-            "integrity": "sha512-qSs4ifwzKJSV39ucNjsvc6WVHs6b7S03sOh2OcHF9UHfVPqWWALUsNUVzhSBiItjRZoLHx7nIarVjqKVusUZ1Q==",
```
jakebailey, line 288: I think sometihng went wrong in the lockfile; we're now missing a bunch of resolved/integrety strings.

internal/api/conn_async.go
andrewbranch: There's no panic call in our own code in the implementations of WriteRequest. Why do you ask?

jakebailey: Mainly because the lock isn't deferred, force of habit; the func looked like it could be complicated enough to break in a mysterious way that hangs things.

jakebailey (approved):

PR #2623, opened 2026-01-31
Fix inlay hints panic: guard against stale line map in LineAndCharacterToPosition

### LINKED ISSUE #2077: Inlayhint panic
## Stack trace

```
panic handling request textDocument/inlayHint bad line number. Line: 109, lineMap length: 109 goroutine 2804779 [running]:
runtime/debug.Stack()
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0xc0000ecdc0, 0xc1512c22d0)
	github.com/microsoft/typescript-go/internal/lsp/server.go:572 +0x58
panic({0x7ff658b6dae0?, 0xc0050479c0?})
	runtime/panic.go:783 +0x132
github.com/microsoft/typescript-go/internal/ls/lsconv.(*Converters).LineAndCharacterToPosition(0xc1798b0000, {0x7ff658f0be70, 0xc025c31b08}, {0x58a40240?, 0x7ff6?})
	github.com/microsoft/typescript-go/internal/ls/lsconv/converters.go:143 +0x2d9
github.com/microsoft/typescript-go/internal/ls/lsconv.(*Converters).FromLSPRange(0xc1798b0000, {0x7ff658f0be70, 0xc025c31b08}, {{0x25c31b08?, 0xc0?}, {0x0?, 0x0?}})
	github.com/microsoft/typescript-go/internal/ls/lsconv/converters.go:43 +0x4c
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideInlayHint(0xc181be80c0, {0x7ff658f0d860, 0xc11a99e050}, 0xc1512c22a0)
	github.com/microsoft/typescript-go/internal/ls/inlay_hints.go:38 +0x165
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleInlayHint(0xc000322b40?, {0x7ff658f0d860?, 0xc11a99e050?}, 0xc05ff13b60?, 0x60?)
	github.com/microsoft/typescript-go/internal/lsp/server.go:998 +0x25
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].27({0x7ff658f0d860, 0xc11a99e050}, 0xc1512c22d0)
	github.com/microsoft/typescript-go/internal/lsp/server.go:558 +0x11a
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0xc0000ecdc0, {0x7ff658f0d860, 0xc11a99e050}, 0xc1512c22d0)
	github.com/microsoft/typescript-go/internal/lsp/server.go:451 +0x10d
	github.com/microsoft/typescript-go/internal/lsp/server.go:356 +0x3a
	github.com/microsoft/typescript-go/internal/lsp/server.go:376 +0x9ad

[Error - 2:08:44 PM] Request textDocument/inlayHint failed.
  Message: InternalError: panic handling request textDocument/inlayHint: bad line number. Line: 109, lineMap length: 109

```

## Steps to reproduce

Sorry not sure - feel free to close if its not useful.

abelcha:
Fixes #2077

**Problem**
Inlay-hint requests (and potentially other position-based features) can panic the server with

```
runtime error: slice bounds out of range [3529:3526]
```

in `LineAndCharacterToPosition` → `converters.go:164` when the line map cache is stale after file edits (especially truncations or large deletions). The line start offset ends up > current text length, and `script.Text()[start:]` panics.

**Root cause**
`LineStarts[line]` can reference a byte offset that no longer exists in the current document version. This is particularly easy to trigger during rapid edits + inlay-hint refreshes.

**Fix**
Add a minimal guard before any slicing:

```go
textLen := core.TextPos(len(script.Text()))
if start >= textLen {
    return textLen  // safe fallback to EOF instead of panic
}
```

This keeps the happy path completely unchanged and prevents the server crash in the common stale-position case. Downstream consumers of the position (inlay hints, etc.) will simply receive EOF and skip rendering the bad hint — far better than crashing.

we could fix this by recomputing line maps on every request instead of caching, but that seems like a big performance hit for an edge case. Is my understanding right that the tradeoff isn't worth it here,

**Test**
Added `TestLineAndCharacterToPosition_StalePositionData` that explicitly creates a stale line map (LineStarts with offset > text length) and confirms:
- No panic
- Returns `textLen` as expected

Also added `TestReproduceIssue2077` which reproduces the exact crash scenario from issue #2077:
- Before fix: panic with `slice bounds out of range [3529:3526]`
- After fix: returns EOF position gracefully

**Notes**
- I chose to return `textLen` (EOF) rather than erroring out because many callers do not handle errors gracefully and the position is used only for rendering hints.
- No logging added (to keep it zero-overhead), but could be considered in debug mode if desired.

Thanks for the great work on typescript-go — happy to adjust anything that doesn't align with project conventions.


jakebailey:
I don't think we should do this. If this happens, the caller is at fault and needs to be fixed. It can only be a symptom of a worse problem above.

abelcha:
> I don't think we should do this. If this happens, the caller is at fault and needs to be fixed. It can only be a symptom of a worse problem above.

We could fix this by recomputing line maps on every request instead of caching, but that seems like a big performance hit for an edge case. Is my understanding right that the tradeoff isn't worth it here, and a guard makes more sense?

DanielRosenwasser:
There are basically 2 cases we probably need to handle

1. `/resolve`-like requests, where the content length changes since the last request and something like `testDocument/codelens/resolve` asks for a non-existent position (we regressed on this and we have no test for it unfortunately)
2. Incorrect logic that operates over nodes, but uses source text from the wrong source file.

The first is pretty uncommon, but easy to fix. The latter happens in subtle ways unfortunately.

Line maps will always need to be calculated when we care after a content change. The problem is using the right line map, or dismissing out-of-date requests.

DanielRosenwasser:
There's also the potential that there's an issue with our snapshot implementation - but the point of operating over snapshots is to avoid this class of problems where a request is being served but the rug (the current state of data) is pulled out from underneath it.

DanielRosenwasser:
I've opened up https://github.com/microsoft/typescript-go/issues/2682 to track discussion to understand the latter, since we do have examples of *something* going wrong that can't be explained by those two usual causes.

DanielRosenwasser:
We appreciate the PR, but ultimately we don't feel like this is the right approach - but also, we think most cases in https://github.com/microsoft/typescript-go/issues/2682 have been resolved.

PR #2625, opened 2026-01-31
Enhance logging for handled methods with request ID

### LINKED ISSUE #2624: *lsproto.ID.String() hangs if nil on android
Running the LSP server on android hangs when a notification gets handled and tries to print the req.ID, which is `nil`

https://github.com/microsoft/typescript-go/blob/main/internal/lsp/server.go#L525C1-L539C2
```go
func (s *Server) handleRequestOrNotification(ctx context.Context, req *lsproto.RequestMessage) error {
	ctx = lsproto.WithClientCapabilities(ctx, &s.clientCapabilities)

	if handler := handlers()[req.Method]; handler != nil {
		start := time.Now()
		err := handler(s, ctx, req)
==>		s.logger.Info("handled method '", req.Method, "' (", req.ID, ") in ", time.Since(start))
		return err
	}
	s.logger.Warn("unknown method '", req.Method, "'")
	if req.ID != nil {
		return s.sendError(req.ID, lsproto.ErrorCodeInvalidRequest)
	}
	return nil
}
```

cplepage:
Closes #2624

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -528,7 +528,11 @@ func (s *Server) handleRequestOrNotification(ctx context.Context, req *lsproto.R
 	if handler := handlers()[req.Method]; handler != nil {
 		start := time.Now()
 		err := handler(s, ctx, req)
-		s.logger.Info("handled method '", req.Method, "' (", req.ID, ") in ", time.Since(start))
+		idStr := ""
+		if req.ID != nil {
+			idStr = " (" + req.ID.String() + ")"
+		}
+		s.logger.Info("handled method '", req.Method, "'", idStr, " in ", time.Since(start))
 		return err
 	}
 	s.logger.Warn("unknown method '", req.Method, "'")


internal/lsp/lsproto/jsonrpc.go
```diff
@@ -43,7 +43,9 @@ func NewIDString(str string) *ID {
 }
 
 func (id *ID) String() string {
-	if id.str != "" {
+	if id == nil {
+		return "<nil>"
+	} else if id.str != "" {
```
jakebailey: This should never be the case; a nil ID means it's a notification, so we shouldn't print the ID at all. I would fix the calling code to do the nil check and not attempt to print the ID (which I think will produce a clearer log message)

cplepage: right!

But the function is `handleRequestOrNotification` which then logs 
```go
		s.logger.Info("handled method '", req.Method, "' (", req.ID, ") in ", time.Since(start))
```

we could switch on Request or Notification and log a different message ?

jakebailey: You should be able to just check if req.ID is nil or not. nil means notification.

I'd just do maybe:

```go
idStr := ""
if req.ID != nil {
	idStr = " (" + req.ID.String() + ")"
}
s.logger.Info("handled method '", req.Method, "'", idStr, " in ", time.Since(start))
```

PR #2628, opened 2026-01-31
Fix dedupe/redirect nondeterminism

jakebailey:
Noticed this in #1604 (and in other CI jobs with dedupes). Without this, redirected / deduplicated files randomly swap ordering because the inputs are randomly ordered.

The comparison change is the real fix, but I have also removed the stable sorts since that is misleading; `allFileNames` is a map!

### DIFF
--- a/internal/modulespecifiers/specifiers.go
+++ b/internal/modulespecifiers/specifiers.go
@@ -208,6 +208,11 @@ func getAllModulePathsWorker(
 		allFileNames[p.FileName] = p
 	}
 
+	useCaseSensitiveFileNames := info.UseCaseSensitiveFileNames
+	comparePaths := func(a, b ModulePath) int {
+		return comparePathsByRedirect(a, b, useCaseSensitiveFileNames)
+	}
+
 	// Sort by paths closest to importing file Name directory
 	sortedPaths := make([]ModulePath, 0, len(paths))
 	for directory := info.SourceDirectory; len(allFileNames) != 0; {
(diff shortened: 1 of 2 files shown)


internal/modulespecifiers/util.go
```diff
@@ -28,7 +28,10 @@ var (
 
 func comparePathsByRedirectAndNumberOfDirectorySeparators(a ModulePath, b ModulePath) int {
 	if a.IsRedirect == b.IsRedirect {
-		return strings.Count(a.FileName, "/") - strings.Count(b.FileName, "/")
+		if c := strings.Count(a.FileName, "/") - strings.Count(b.FileName, "/"); c != 0 {
+			return c
+		}
+		return strings.Compare(a.FileName, b.FileName)
```
jakebailey: I might be wrong, but I think we have a func that does something like this already...

jakebailey: Could just use `tspath.ComparePaths`

internal/modulespecifiers/specifiers.go
```diff
@@ -208,6 +208,11 @@ func getAllModulePathsWorker(
 		allFileNames[p.FileName] = p
 	}
 
+	useCaseSensitiveFileNames := info.UseCaseSensitiveFileNames
+	comparePaths := func(a, b ModulePath) int {
```
DanielRosenwasser, line 212: This doesn't actually allocate, right? I assume it'll keep everything on the stack.

DanielRosenwasser (approved):

internal/modulespecifiers/specifiers.go
jakebailey: Yeah, or at worst it'll be a single slot on the stack for the closure object

rb1974805-web:
I will distory you that's it

PR #2633, opened 2026-02-02
Leave extension as-is for TS extension comparison.

### LINKED ISSUE #2631: Code lenses and inlay hints only respected from `javascript.` settings since `0.20260127.1`
Ensure the following settings in VS Code

```json5
    "editor.inlayHints.enabled": "on",
    "typescript.inlayHints.parameterNames.enabled": "all",
    "typescript.inlayHints.parameterTypes.enabled": true,
    "typescript.inlayHints.variableTypes.enabled": true,
    "typescript.inlayHints.functionLikeReturnTypes.enabled": true,
    "editor.codeLens": true,
    "typescript.referencesCodeLens.enabled": true,
    "typescript.referencesCodeLens.showOnAllFunctions": true,
    "typescript.implementationsCodeLens.enabled": true,
    "typescript.implementationsCodeLens.showOnInterfaceMethods": true,
    "typescript.implementationsCodeLens.showOnAllClassMethods": true,
```

If you install `0.20260126.1`, and you type in the following code:

```ts
let x = foo(a => a);

function foo(f: (x: string) => string) {
    return f;
}
```

You'll see the following:

Image

However, if you install anything newer, it's all gone:

Image

Now, switch all these settings to `javascript.`:

```json5
    "editor.inlayHints.enabled": "on",
    "javascript.inlayHints.parameterNames.enabled": "all",
    "javascript.inlayHints.parameterTypes.enabled": true,
    "javascript.inlayHints.variableTypes.enabled": true,
    "javascript.inlayHints.functionLikeReturnTypes.enabled": true,
    "editor.codeLens": true,
    "javascript.referencesCodeLens.enabled": true,
    "javascript.referencesCodeLens.showOnAllFunctions": true,
    "javascript.implementationsCodeLens.enabled": true,
    "javascript.implementationsCodeLens.showOnInterfaceMethods": true,
    "javascript.implementationsCodeLens.showOnAllClassMethods": true,
```

Code lenses and inlay hints will work for *both* TypeScript and JavaScript files.

DanielRosenwasser:
Fixes #2631.

Not sure how to get this tested apart from a server test (do we have anything like that?) or an end-to-end VS Code extension test.

### DIFF
--- a/_packages/api/package.json
+++ b/_packages/api/package.json
@@ -3,6 +3,9 @@
     "name": "@typescript/api",
     "version": "1.0.0",
     "type": "module",
+    "engines": {
+        "node": ">=20.19"
+    },
     "imports": {
         "#symbolFlags": {
             "@typescript/source": {
@@ -42,7 +45,7 @@
         "test": "node --test --experimental-strip-types --no-warnings --conditions @typescript/source ./test/**/*.test.ts"
     },
     "devDependencies": {
-        "tinybench": "^3.1.1"
+        "tinybench": "^6.0.0"
     },
     "dependencies": {
         "@typescript/ast": "1.0.0",
--- a/_packages/ast/package.json
+++ b/_packages/ast/package.json
@@ -4,6 +4,9 @@
     "version": "1.0.0",
     "description": "TypeScript AST definitions",
     "type": "module",
+    "engines": {
+        "node": ">=20.19"
+    },
     "imports": {
         "#syntaxKind": {
             "@typescript/source": {
--- a/internal/compiler/fileInclude.go
+++ b/internal/compiler/fileInclude.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 
 	"github.com/microsoft/typescript-go/internal/ast"
-	"github.com/microsoft/typescript-go/internal/core"
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 	"github.com/microsoft/typescript-go/internal/module"
 	"github.com/microsoft/typescript-go/internal/scanner"
@@ -23,8 +22,6 @@ const (
 	fileIncludeKindLibReferenceDirective
 
 	fileIncludeKindRootFile
-	fileIncludeKindSourceFromProjectReference
-	fileIncludeKindOutputFromProjectReference
 	fileIncludeKindLibFile
 	fileIncludeKindAutomaticTypeDirectiveFile
 )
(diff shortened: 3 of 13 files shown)


gabritto:
Server tests are supposed to be fourslash tests, but the fourslash client might not behave like vscode. In particular, I think the issue might be that fourslash passes preferences directly as a map, unlike vscode. (The other fourslash difference I can remember is that we also send LSP messages directly, bypassing JSON marshalling etc.)

DanielRosenwasser:
Without coming up with a completely different way of configuring fourslash tests right now, I just wrote a unit test for `GetPreferences`. If we think a lot of this is eventually changing, then I can drop it.

internal/project/session_test.go
```diff
@@ -904,6 +904,7 @@ func TestSession(t *testing.T) {
 		assert.Equal(t, len(codeLensRefreshCalls), 1, "expected one RefreshCodeLens call after code lens preference change")
 		assert.Equal(t, len(inlayHintsRefreshCalls), 1, "expected one RefreshInlayHints call after inlay hints preference change")
 	})
+
```
jakebailey, line 907: rando change?

DanielRosenwasser: It was bothering me that everything else in the file had a line in between. I was originally going to write a test in this file.

internal/ls/lsutil/configuration.go
```diff
@@ -64,7 +62,7 @@ func (c *UserConfig) JS() *UserPreferences {
 }
 
 func (c *UserConfig) GetPreferences(activeFile string) *UserPreferences {
-	fileEnding := strings.TrimPrefix(tspath.GetAnyExtensionFromPath(activeFile, nil, true), ".")
+	fileEnding := tspath.GetAnyExtensionFromPath(activeFile, nil, true)
 	if activeFile == "" || tspath.ExtensionIsTs(fileEnding) {
```
jakebailey:
```suggestion
	if activeFile == "" || tspath.ExtensionIsTs(tspath.GetAnyExtensionFromPath(activeFile, nil, true)) {
```


this would make more sense to me, but, it's not harmful to get an extension from an empty string and ignore it

jakebailey (approved):

internal/ls/lsutil/configuration.go
DanielRosenwasser: Aside, it's weird that we consider an empty file name to be TS, but an extensionless file name to be JS.

PR #2627, opened 2026-01-31
Multiple improvements to Quick Info

ahejlsberg:
Changes in this PR:

* When an overload signature has no JSDoc, Quick Info now displays JSDoc from the first overload.
* Quick Info uses `getMeaningOfLocation` to determine what symbol meanings to display (value, type, or namespace, or all).
* Fixes a translation bug in `getMeaningOfLocation`.
* Hovering over `this` would sometimes display no Quick Info or a wrong type.

Fixes #2599.

PR #2641, opened 2026-02-02
Properly track first Quick Info declaration + fix issue in fourslash

ahejlsberg:
This is a follow-up PR to #2627 that fixes the issue reported [here](https://github.com/microsoft/typescript-go/pull/2627#issuecomment-3833435879).

The PR also fixes an issue in fourslash that caused baseline verification of Quick Info to not work in multi-file tests.

### DIFF
--- a/testdata/baselines/reference/fourslash/quickInfo/jsdocLink2.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/jsdocLink2.baseline
@@ -1,0 +1,56 @@
+// === QuickInfo ===
+=== /script.ts ===
+// /**
+//  * {@link C}
+//  * @wat Makes a {@link C}. A default one.
+//  * {@link C()}
+//  * {@link C|postfix text}
+//  * {@link unformatted postfix text}
+//  * @see {@link C} its great
+//  */
+// function CC() {
+//          ^^
+// | ----------------------------------------------------------------------
+// | ```tsx
+// | function CC(): void
+// | ```
+// | [C](file:///jsdocLink2.ts#1,7-1,8)
+// | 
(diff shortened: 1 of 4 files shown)


internal/ls/hover.go
```diff
@@ -238,19 +243,21 @@ func getQuickInfoAndDeclarationAtLocation(c *checker.Checker, symbol *ast.Symbol
 			flags = symbol.Flags & ast.SymbolFlagsType
 		case ast.SemanticMeaningNamespace:
 			flags = symbol.Flags & ast.SymbolFlagsNamespace
+		default:
+			flags = symbol.Flags & (ast.SymbolFlagsValue | ast.SymbolFlagsSignature | ast.SymbolFlagsType | ast.SymbolFlagsNamespace)
 		}
 		if flags == 0 {
+			if aliasLevel != 0 || b.Len() != 0 {
```
gabritto, line 250: I get checking the alias level, but why do we need to check if `b.Len() != 0` here?

jakebailey:
You fixed four generated tests, so you should be able to do `npm run updatefailing`

internal/ls/hover.go
ahejlsberg: The check is here such that we reach the subsequent logic only when `aliasLevel` is zero and nothing has been output yet. In that case we want to do a final sanity check where we output any meaning of the symbol, even if `getMeaningFromLocation` doesn't indicate it. There are cases where `getMeaningFromLocation` returns "wrong" results because of errors in the source file, for example during completion (and completion actually invokes Quick Info).

jakebailey (approved):

PR #2645, opened 2026-02-02
Add first-class unused baseline tracking

jakebailey:
Rather than having CI check this, we can make `hereby test` do this.

The general gist is that we set up a `TestMain` for every package which uses the baseline infrastructure, which allows for code to be run before and after all tests in a package. Then, if we are running a plain `hereby test`, `Herebyfile.mjs` informs the test binary of a place to store a list of baselines it saw, and every time one is written, it's listed there. Then the hereby task collects all of those outputs, scans the reference baseline dir, and complains about unknown files. It also creates the `.delete` files used by `baseline-accept` for easy acceptance.

This flow ends up looking like this:

image

If a test package starts using the baseline infra without setting up `TestMain`, the baseline writing will fail the test, which can then be effectively just copied to the package from another working one.

### DIFF
--- a/internal/testutil/baseline/testmain.go
+++ b/internal/testutil/baseline/testmain.go
@@ -1,6 +1,7 @@
 package baseline
 
 import (
+	"bufio"
 	"fmt"
 	"hash/fnv"
 	"os"
@@ -79,17 +80,24 @@ func writeRecordedBaselines(trackingPath string) {
 		return
 	}
 
+	if err := doWriteRecordedBaselines(trackingPath); err != nil {
+		fmt.Fprintf(os.Stderr, "baseline: failed to write tracking file %s: %v\n", trackingPath, err)
+		os.Exit(1)
+	}
+}
+
+func doWriteRecordedBaselines(trackingPath string) error {
 	f, err := os.Create(trackingPath)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "baseline: failed to create tracking file %s: %v\n", trackingPath, err)
-		os.Exit(1)
+		return err
 	}
 	defer f.Close()
 
+	w := bufio.NewWriter(f)
 	for baseline := range recordedBaselines.Keys() {
-		if _, err := fmt.Fprintln(f, baseline); err != nil {
-			fmt.Fprintf(os.Stderr, "baseline: failed to write to tracking file %s: %v\n", trackingPath, err)
-			os.Exit(1)
+		if _, err := fmt.Fprintln(w, baseline); err != nil {
+			return err
 		}
 	}
+	return w.Flush()
 }


internal/testutil/baseline/baseline.go
```diff
@@ -73,6 +79,10 @@ func Run(t *testing.T, fileName string, actual string, opts Options) {
 	{
 		localPath := filepath.Join(localRoot, unusedOutRoot, origSubfolder, diffFileName)
 		referencePath := filepath.Join(referenceRoot, unusedOutRoot, origSubfolder, diffFileName)
+
+		// Record this potential diff baseline for tracking unused baselines
+		recordBaseline(t, filepath.Join(unusedOutRoot, origSubfolder, diffFileName))
+
 		writeComparison(t, NoContent, localPath, referencePath, false)
```
jakebailey, line 86: Not sure how it works for NoContent; need to double check.

jakebailey: Ah, no, if NoContent, a `.delete` file will get created in `local`, which means the tests fail, which means we don't trigger an unused check anyway.

jakebailey: Rather, this is comparing a diff file, which should be empty, so I just misunderstood

Herebyfile.mjs
```diff
@@ -302,11 +303,66 @@ const goTestEnv = {
...
+async function collectUsedBaselines(trackingDir) {
+    /** @type {Set<string>} */
+    const usedBaselines = new Set();
+    if (!fs.existsSync(trackingDir)) {
+        return usedBaselines;
+    }
+
+    const trackingFiles = await fs.promises.readdir(trackingDir);
+    for (const file of trackingFiles) {
+        if (!file.endsWith(".txt")) continue;
```
DanielRosenwasser: Don't we have non-`txt` baselines?

internal/tsoptions/testmain_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/testutil/baseline"
+)
+
+func TestMain(m *testing.M) {
+	defer baseline.Track()()
+	m.Run()
```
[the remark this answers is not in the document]
jakebailey: Nope: https://pkg.go.dev/testing#M.Run:~:text=If%20TestMain%20returns%2C%20the%20test%20wrapper%20will%20pass%20the%20result%20of%20m.Run%20to%20os.Exit%20itself.

Herebyfile.mjs
jakebailey: These are the files written by the baseline package that contain baseline names. This filter isn't even required; i'll remove it

DanielRosenwasser: Okay, sorry, these are not baselines. These are lists of baseline files (usually in some temp directory), and you are okay with even reading stale ones.

jakebailey: There aren't any stale ones, as a new temp dir is created every time the tests run (and deleted after).

PR #2647, opened 2026-02-03
fix(2632): allow missing names in jsdoc param tags

### LINKED ISSUE #2632: `@param` JSDoc requires parameter names, did not in 6.0 and prior
```js
// @ts-check

/**
 * @param {(x: string) => string}
 */
function foo(f) {
    return f(123);
}
```

**Current**: No error on `f(123)`, but an "implicit `any`" error under `noImplicitAny`.

**Expected**: Assignability error for `number` to `string` on `f(123)`, no other errors.

a-tarasyuk:
Fixes #2632

### DIFF
--- a/_packages/native-preview/bin/tsgo.js
+++ b/_packages/native-preview/bin/tsgo.js
@@ -5,6 +5,16 @@ import { execFileSync } from "node:child_process";
 
 const exe = getExePath();
 
+if (typeof process.execve === "function") {
+    // > v22.15.0
+    try {
+        process.execve(exe, [exe, ...process.argv.slice(2)]);
+    }
+    catch {
+        // not available on windows, ignore the error and fallback
+    }
+}
+
 try {
     execFileSync(exe, process.argv.slice(2), { stdio: "inherit" });
 }
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -1684,7 +1684,7 @@ function generateCode() {
         }
 
         const paramType = request.params ? resolveType(request.params) : undefined;
-        const paramGoType = paramType ? (paramType.needsPointer ? `*${paramType.name}` : paramType.name) : "any";
+        const paramGoType = paramType ? (paramType.needsPointer ? `*${paramType.name}` : paramType.name) : "NoParams";
 
         writeLine(`// Type mapping info for \`${request.method}\``);
         if (responseTypeName) {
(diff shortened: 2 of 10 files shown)


testdata/baselines/reference/submodule/conformance/jsDeclarationsRestArgsWithThisTypeInJSDocFunction(target=es2015).errors.txt.diff
```diff
@@ -2,19 +2,20 @@
...
-+bug38550.js(5,10): error TS7006: Parameter 'functionDeclaration' implicitly has an 'any' type.
 +
 +
 +==== bug38550.js (2 errors) ====
 +    export class Clazz {
 +      /**
 +       * @param {function(this:Object, ...*):*} functionDeclaration
++                 ~~~~~~~~
++!!! error TS2552: Cannot find name 'function'. Did you mean 'Function'?
++!!! related TS2728 lib.es5.d.ts:--:--: 'Function' is declared here.
```
jakebailey, line 15: Do you know why this error is now appearing?

ahejlsberg: It's old Closure syntax that we decided not to support any longer.

jakebailey: I suppose yes, though I'm more wondering why this is just now appearing in this PR. I can only assume weird follow-on behavior.

a-tarasyuk: @jakebailey This is unsupported _Closure_ syntax. TypeScript handles it by parsing to [`parseJSDocFunctionType`](https://github.com/microsoft/TypeScript/blob/2a3bed2b4265fa1173c88771a21ce044e6480f75/src/compiler/parser.ts#L4615-L4616) when it sees `function` in JSDoc. In tsgo, parsing currently stops after the `function` keyword

For this example:

```ts
@param {function(this:Object, ...):} functionDeclaration
  ^             ^ stops at the `(` token
```

As `(` is not an identifier token, the parsed name becomes an empty identifier (`""`), and everything after the `(` is ignored.


Previously, tsgo lacked logic to resolve parameters by index. The current change assumes parameter index 0, treats the type as a function, and then tries to resolve/check it. 

@jakebailey @ahejlsberg Should the parser simply recover here, ignore the unsupported syntax, and attempt parsing at the parameter name?

jakebailey: I think this is fine. We have this same error anew in other diffs.

PR #2659, opened 2026-02-03
Fix generate CI task flakes

jakebailey:
We unfortunately have tests which produce baselines _after_ they fail for other reasons, so one test run is not enough.

https://github.com/microsoft/typescript-go/actions/runs/21642390137/job/62385368566


gabritto:
> We unfortunately have tests which produce baselines _after_ they fail for other reasons, so one test run is not enough.
> 
> https://github.com/microsoft/typescript-go/actions/runs/21642390137/job/62385368566

Why are those tests like that?

jakebailey:
They're "server" style tests which log out info from the test itself

PR #2689, opened 2026-02-05
Start targeting TS 6.0 behavior

jakebailey:
It's time to start targeting 6.0!

This bumps the submodule to the latest main.

I've preemptively partially ported a few things to make this less diff-y, but we will need to still go one at a time and port things over.

Notably:

- I have confined the fourslash tests to lib=ES5, now that the tests explicitly request that.
- `mappedTypeAsClauseRecursiveNoCrash1.ts` is skipped as it crashes the compiler, pending a port.
- `for-of29.ts` is skipped; it's now flaky, flipping between differing errors. (!!!)
- I (mostly) ported the `strict`, `alwaysStrict`, `target`, and `module` changes, enough to not end up with a billion diffs.

### DIFF
--- a/internal/core/scripttarget_stringer_generated.go
+++ b/internal/core/scripttarget_stringer_generated.go
@@ -24,6 +24,7 @@ func _() {
 	_ = x[ScriptTargetESNext-99]
 	_ = x[ScriptTargetJSON-100]
 	_ = x[ScriptTargetLatest-99]
+	_ = x[ScriptTargetLatestStandard-11]
 }
 
 const (


internal/testrunner/compiler_runner.go
```diff
@@ -210,6 +216,11 @@ func (r *CompilerBaselineRunner) runSingleConfigTest(t *testing.T, testName stri
 		t.Skipf("Skipping test %s with outFile set", testName)
 	}
 
+	switch compilerTest.options.Target {
+	case core.ScriptTargetES3, core.ScriptTargetES5:
```
jakebailey, line 220: I'm skipping ES3/ES5 here preemptively; will need to skip other things like `allowStrict=false`, `downlevelIteration`, etc etc. In later PRs.

internal/fourslash/tests/util/util.go
```diff
@@ -33,13273 +33,1138 @@ var CompletionUndefinedVarItem = &lsproto.CompletionItem{
...
+		Label:    "Lowercase",
+		Kind:     PtrTo(lsproto.CompletionItemKindClass),
 		SortText: PtrTo(string(ls.SortTextGlobalsOrKeywords)),
 	},
 	&lsproto.CompletionItem{
-		Label:    "GamepadEvent",
-		Kind:     PtrTo(lsproto.CompletionItemKindVariable),
-		SortText: PtrTo(string(ls.SortTextGlobalsOrKeywords)),
-	},
-	&lsproto.CompletionItem{
```
jakebailey, line 1010: Note the giant amount of stuff deleted here; Strada fourslash was doing ES5, but Corsa was not, leading to these. Since the target bump in Strada was going to break every test in ways that sucked, I instead annotated most everything with `@lib: es5` in the target bump changes for 6.0 in Strada. This change brings these hardcoded globals lists back to what they are in Strada, which is ES5.

PR #2704, opened 2026-02-06
Allow invalid UTF-8

### LINKED ISSUE #1531: tsgo panics with invalid UTF-8 when marshalling build info
## Stack trace

```
panic: Failed to marshal build info: json: cannot marshal from Go incremental.BuildInfoDiagnosticsOfFile within "/semanticDiagnosticsPerFile/0": invalid UTF-8 within "/1/1/message" after offset 319

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/incremental.(*Program).emitBuildInfo(0xc002a0c180, {0xdb5728, 0x14ad4e0}, {0x0?, 0xd8?, 0x0?})
        github.com/microsoft/typescript-go/internal/incremental/program.go:264 +0x588
github.com/microsoft/typescript-go/internal/incremental.(*emitFilesHandler).emitAllAffectedFiles(0xc000b35590, {0x0?, 0x1e?, 0x0?})
        github.com/microsoft/typescript-go/internal/incremental/emitfileshandler.go:132 +0x696
github.com/microsoft/typescript-go/internal/incremental.emitFiles({0xdb5728, 0x14ad4e0}, 0xc002a0c180, {0x0?, 0x0?, 0x0?}, 0x0)
        github.com/microsoft/typescript-go/internal/incremental/emitfileshandler.go:273 +0x1a5
github.com/microsoft/typescript-go/internal/incremental.(*Program).Emit(0xc002a0c180, {0xdb5728, 0x14ad4e0}, {0x0?, 0x0?, 0x0?})
        github.com/microsoft/typescript-go/internal/incremental/program.go:193 +0x307
github.com/microsoft/typescript-go/internal/execute.emitFilesAndReportErrors({0xdb8450, 0xc000144000}, {0xdbb000, 0xc002a0c180}, 0xc0001285e0)
        github.com/microsoft/typescript-go/internal/execute/tsc.go:384 +0x1cf
github.com/microsoft/typescript-go/internal/execute.emitAndReportStatistics({0xdb8450, 0xc000144000}, {0xdbb000?, 0xc002a0c180?}, 0xc000002000, 0xc0001560f0, 0x0?, 0x6325c, 0x299e9ef, 0x7ffd6e, ...)
        github.com/microsoft/typescript-go/internal/execute/tsc.go:308 +0x6d
github.com/microsoft/typescript-go/internal/execute.performIncrementalCompilation({0xdb8450, 0xc000144000}, 0xc0001560f0, 0xc0001285e0, 0xc00011e240, 0x6325c, 0x0)
        github.com/microsoft/typescript-go/internal/execute/tsc.go:250 +0x40c
github.com/microsoft/typescript-go/internal/execute.tscCompilation({0xdb8450, 0xc000144000}, 0xc000156000, 0x0)
        github.com/microsoft/typescript-go/internal/execute/tsc.go:194 +0xb53
github.com/microsoft/typescript-go/internal/execute.CommandLine({0xdb8450, 0xc000144000}, {0xc0000200a0, 0x0, 0x0}, 0x0)
        github.com/microsoft/typescript-go/internal/execute/tsc.go:63 +0x17e
main.runMain()
        github.com/microsoft/typescript-go/cmd/tsgo/main.go:23 +0x109
main.main()
        github.com/microsoft/typescript-go/cmd/tsgo/main.go:10 +0x13
```

## Steps to reproduce

1. Create a .ts file with this content:

```
const createFileListFromFiles = (files: File[]): FileList => {
  const fileList: FileList = {
    length: files.length,
    item: (index: number): File | null => files[index] || null,
    [Symbol.iterator]: function* (): IterableIterator<File> {
      for (const file of files) yield file;
    },
    ...files,
  } as unknown as FileList;

  return fileList;
};
```

2. Run:

```
tsgo
```

### LINKED ISSUE #2336: Crash for signature help in arrow function parameters

gabritto:
Fixes #1531, fixes #2336.

In Corsa we use an invalid UTF-8 sequence (`\xFE`) as a prefix to internal symbol names to prevent internal symbol names from matching the names of symbols that came from the input program. These internal symbol names sometimes escape into compiler outputs via diagnostics, quick info, signature help, etc, which then caused crashes with `json.Marshal` when we tried to marshal any of those outputs before writing it to a .buildinfo file or sending it to the LSP client.
This PR fixes the problem by making sure we always call `json.Marshal` & related functions with the `AllowInvalidUTF8` option. This option makes it so that marshalling doesn't error on invalid characters, and invalid UTF-8 is replaced by the Unicode replacement character.
Note this doesn't handle the case where program inputs have invalid UTF-8.

This PR is a partial revert of #1504, so that we once again have an internal json library that makes sure we always pass the `AllowInvalidUTF8` option.

### DIFF
--- a/internal/diagnostics/diagnostics.go
+++ b/internal/diagnostics/diagnostics.go
@@ -120,16 +120,16 @@ func Format(text string, args []string) string {
 	}
 
 	// Replace invalid UTF-8 with Unicode replacement character
-	validArgs := core.SameMap(args, func(arg string) string {
+	args = core.SameMap(args, func(arg string) string {
 		return strings.ToValidUTF8(arg, "\uFFFD")
 	})
 
 	return placeholderRegexp.ReplaceAllStringFunc(text, func(match string) string {
 		index, err := strconv.ParseInt(match[1:len(match)-1], 10, 0)
-		if err != nil || int(index) >= len(validArgs) {
+		if err != nil || int(index) >= len(args) {
 			panic("Invalid formatting placeholder")
 		}
-		return validArgs[int(index)]
+		return args[int(index)]
 	})
 }
 


sheetalkamat:
Updated my branch which had test that i created before going on leave in Dec

You want tsc test for incremental check: https://github.com/microsoft/typescript-go/blob/f5319cd5a21b6c659318f023059559f127225a3a/internal/execute/tsctests/tsc_test.go#L1943  

You would need to update baseline sanitizer (it was TODO after actually fixing the issue) so that it can handle symbol ID to be present in name and hence differing on reach run.
```
func (o *outputSanitizer) addOutputLine 
```

.golangci.yml
```diff
@@ -70,7 +70,9 @@ linters:
         main:
           deny:
             - pkg: 'encoding/json$'
-              desc: 'Use "github.com/go-json-experiment/json" instead.'
+              desc: 'Use "github.com/microsoft/typescript-go/internal/json" instead.'
+            - pkg: 'github.com/go-json-experiment/json'
+              desc: 'Use "github.com/microsoft/typescript-go/internal/json" instead.'
```
jakebailey, line 75: This should also get mad  at `/json/jsontext`

internal/tsoptions/tsconfigparsing_test.go
```diff
@@ -1017,13 +1017,13 @@ func TestParseSrcCompiler(t *testing.T) {
 		CurrentDirectory: compilerDir,
 	}
 
-	jsonText, ok := fs.ReadFile(tsconfigFileName)
+	json, ok := fs.ReadFile(tsconfigFileName)
```
jakebailey: Are these renames required?

internal/diagnostics/diagnostics.go
```diff
@@ -117,6 +119,11 @@ func Format(text string, args []string) string {
 		return text
 	}
 
+	// Replace invalid UTF-8 with Unicode replacement character
+	args = core.Map(args, func(arg string) string {
```
jakebailey: Can you just use a plain loop for this? This package currently does not depend on `core`, which helps it stay cached even when `core` changes.

internal/json/json.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"io"
+
+	jsonv2 "github.com/go-json-experiment/json"
+	"github.com/go-json-experiment/json/jsontext"
+)
+
+func Marshal(in any, opts ...jsonv2.Options) (out []byte, err error) {
+	return jsonv2.Marshal(in, append([]jsonv2.Options{jsontext.AllowInvalidUTF8(true)}, opts...)...)
```
jakebailey: For perf, potentially we should extract out a one-elemenet array to pass in when `len(opts) == 0`, to avoid appending into a fresh slice?

internal/testutil/fsbaselineutil/differ.go
```diff
@@ -91,6 +94,15 @@ func (d *FSDiffer) BaselineFSwithDiff(baseline io.Writer) {
 	*d.WrittenFiles = collections.SyncSet[string]{} // Reset written files after baseline
 }
 
+var internalSymbolRegex = regexp.MustCompile(`\x{FFFD}@[^@]+@[0-9]+`)
```
jakebailey, line 97: Does this have to use a regex? (maybe it did before?)

.golangci.yml
gabritto: I thought it did, since there's no `$` at the end of the package name it should match any string with prefix `github.com/go-json-experiment/json`

internal/json/json.go
```diff
@@ -0,0 +1,91 @@
+//nolint:depguard
+package json
+
+import (
+	"io"
+	"slices"
+
+	jsonv2 "github.com/go-json-experiment/json"
```
jakebailey: why does this get renamed?

jakebailey: why does this get renamed?

internal/diagnostics/diagnostics.go
```diff
@@ -117,12 +118,18 @@ func Format(text string, args []string) string {
 		return text
 	}
 
+	// Replace invalid UTF-8 with Unicode replacement character
+	validArgs := make([]string, 0, len(args))
```
jakebailey: sadly I think that `args = core.SameMap(args, ...)` might be better

internal/testutil/fsbaselineutil/differ.go
```diff
@@ -91,6 +93,59 @@ func (d *FSDiffer) BaselineFSwithDiff(baseline io.Writer) {
 	*d.WrittenFiles = collections.SyncSet[string]{} // Reset written files after baseline
 }
 
+func SanitizeInternalSymbolName(s string) string {
```
jakebailey, line 101: Oh wow, I did not expect this to get this large. Probably we could just use the regex but bail early if it's not going to match

internal/json/json.go
gabritto: Ah yes, I just reverted the PR that shifted us to use `go-json-experiment/json`, and I think before that PR we had both json and jsonv2. I'll fix.

internal/diagnostics/diagnostics.go
```diff
@@ -117,12 +119,17 @@ func Format(text string, args []string) string {
 		return text
 	}
 
+	// Replace invalid UTF-8 with Unicode replacement character
+	validArgs := core.SameMap(args, func(arg string) string {
```
jakebailey: can you just reassign `args`?

jakebailey (approved):

Issue #2609, opened 2026-01-29
Type instantiation is excessively deep and possibly infinite.


TkDodo:
## Steps to reproduce

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQFCiSxwDeArqlgGLQgC+cRJcgAEYyAHapRuANYB6SnhgBaAtzIBuOuGjwmALz4DSZPRAAmGmnTEwsUAnixwACsTCoAPDThwAKgGVcAAssEGQ4LAAPWzEzdD0AOgAtcwB5ACMAKyx8D0SUswxkAHd-IOQwLAA+ABpvPw5gLAAbMwA5ZBAnKJi4uABRaKhFD2ksAE8IAjhE4DECOw8A4NDkWrhJKDmAc1qaKuZ6ueAYYGRmgDVzliwALhmEuYWoJcCQsKqAbV9GlvbOrAAXU0PjEAPuPyarQ6XRBGxWYQhb1Wmh4ViiDEILDE+GAEDEmCwYGIXh8y3e4R6WFi8WSaSyORgeTphRKZQq1TqZN+0IBEWi1L6gxgw1yY0m01m80W5NW602Oz2VQAFEwwV0anBjqdzldmjcePdXBB3K8EchNZC-jDqgBKQ4+XD4yT8bhwAC8cDYnG4qvqPjMWAcLGaMD1N1Q9yYn3VQPu2rOl2uWB4XLgPFtmnqlBgLCgBI8qigIASPLMcFj7rVAJ4VSYBCh5fdBw8ZmAADc6w2-glYzwPLI253+7IiyWy1VUVYnRJ4KhzR6HhAGfg-T5VBB7okFWJtsrbTQM1YCNjcfi-FhJPvDtmsLn83Blf64B5sMSIPCKVX5xS+JWAEQbv+WpiCcibhlg7r-ukyBQMBshVPUB48EAA

Sorry that this isn’t more minimal, as it still uses two external libraries (TanStack Form and zod), but I have trimmed it down from the original as much as possible.

As [discussed with Jake on Bluesky](https://bsky.app/profile/tkdodo.eu/post/3mdked3fabc2b), this could very well be an issue with TanStack Form’s typings, but I’d still expect tsgo and ts 5.9 to behave identically.

## Behavior with `typescript@5.9`

no error

## Behavior with `tsgo`

Error in line 22:

[...]

PR #2711, opened 2026-02-07
Optimize distributive conditional type instantiation for trivial extends types

DukeDeSouth:
### Summary

Adds a fast path in `getConditionalTypeInstantiation` for distributive conditional types where the extends type resolves to `never`, `any`, or `unknown`. In these cases, the branch outcome is trivially determinable for **all** union members, so we skip the expensive per-member `getConditionalType` call (which involves type instantiation, permissive instantiation, and structural comparison) and directly apply the predetermined branch.

#### The Problem

When distributing a conditional type like `Exclude<BigUnion, never>` (= `T extends never ? never : T`) over a large union (e.g., React HTML props with 250+ members), the current code calls `getConditionalType` for **each member** of the union. Each call:

1. Instantiates the check type (+1 `instantiateType` call)
2. Instantiates the extends type (+1 `instantiateType` call)
3. Computes permissive instantiations
4. Runs `isTypeAssignableTo` structural comparison
5. Instantiates the result branch (+1 `instantiateType` call)

For `extends never`, **every** member goes to the false branch — all this work is redundant. In complex HOC patterns (withRouter + generic enhance + React props), multiple layers of `Omit`/`Pick`/`Exclude` compound this, potentially exhausting the 5M `instantiationCount` budget and producing a false positive TS2589 error.

#### The Fix

Before distributing over union members, resolve the extends type. If it's:

- **`never`**: All non-never members go to false branch → apply false branch directly
- **`any`/`unknown`**: All members go to true branch → apply true branch directly

**Safety guards** (to prevent false negatives on genuinely infinite types):
- Only applies when `!forConstraint` (constraint resolution has different semantics)
- Only applies when `len(root.inferTypeParameters) == 0` (infer types need inference logic)
- Only applies when distribution type is a union
- Falls through to the original `getConditionalType` path for all other cases
- Does **not** change any instantiation limits

#### Testing

- All existing submodule tests pass (0 regressions)
- `limitDeepInstantiations` still correctly errors on genuinely infinite types
- All `recursiveConditionalTypes` / `recursiveConditionalCrash` tests pass
- All `conditionalType*` and `mappedType*` tests pass

Fixes microsoft/TypeScript#34933
Related: #988, #2609

---

### Metadata
- **Generator**: Claude (Anthropic) via Cursor IDE
- **Methodology**: AI-assisted development with human oversight and review

### AI Contribution Summary
- Solution design and implementation

### Verification Steps Performed
1. Reproduced the reported issue
2. Analyzed source code to identify root cause
3. Implemented and tested the fix

### Human Review Guidance
- Review the implementation for correctness and completeness
- Verify test coverage matches the described scenarios

Made with M7 [Cursor](https://cursor.com)

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -22006,6 +22006,37 @@ func (c *Checker) getConditionalTypeInstantiation(t *Type, mapper *TypeMapper, f
 			// distributive conditional type T extends U ? X : Y is instantiated with A | B for T, the
 			// result is (A extends U ? X : Y) | (B extends U ? X : Y).
 			if distributionType != nil && checkType != distributionType && distributionType.flags&(TypeFlagsUnion|TypeFlagsNever) != 0 {
+				// Fast path for distributive conditionals where the extends type makes the branch
+				// trivially determinable for ALL union members. When the extends type is 'never',
+				// T extends never is false for all non-never T. When the extends type is 'any' or
+				// 'unknown', T extends any/unknown is true for all T. In these cases we skip the
+				// expensive per-member getConditionalType call (which involves type instantiation,
+				// permissive instantiation, and structural comparison) and directly apply the
+				// predetermined branch. This primarily benefits patterns like Exclude<BigUnion, never>
+				// (used by Omit<T, never>) where hundreds of union members would each trigger
+				// redundant type checks that all yield the same branch outcome.
+				// See: https://github.com/microsoft/TypeScript/issues/34933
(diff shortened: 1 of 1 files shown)


DukeDeSouth:

jakebailey:
Thanks, but this kind of change is not in scope: https://github.com/microsoft/typescript-go/blob/main/CONTRIBUTING.md#scope-of-acceptable-changes

PR #2712, opened 2026-02-09
Replace unsafe.Pointer for seen symbol tables with unique ID

jakebailey:
We use `reflect.ValueOf(m).UnsafePointer()` in the checker in order to have maps as map keys themselves. This is the only place we use unsafe in the checker, and theoretically going through reflect has a high cost (see https://godbolt.org/z/3jMcxY49T).

Instead of doing this, we can instead note that we always get a symbol table from some _other_ value that has an ID; a `Node` or a `Symbol`. If we use that ID (or a sentinel value) plus some metadata, we can avoid reflect and unsafe, swapping for a simple load+OR.

This seems to achieve the same goal. All tests pass.

### DIFF
--- a/internal/checker/symbolaccessibility.go
+++ b/internal/checker/symbolaccessibility.go
@@ -402,10 +402,10 @@ type accessibleSymbolChainContext struct {
 type symbolTableID uint64
 
 const (
-	stKindLocals  symbolTableID = 0
-	stKindExports symbolTableID = 1 << 62
-	stKindMembers symbolTableID = 2 << 62
-	stKindGlobals symbolTableID = 3 << 62
+	stKindLocals symbolTableID = iota << 62
+	stKindExports
+	stKindMembers
+	stKindGlobals
 )
 
 func symbolTableIDFromLocals(node *ast.Node) symbolTableID {


internal/checker/symbolaccessibility.go
```diff
@@ -395,7 +393,35 @@ type accessibleSymbolChainContext struct {
...
+func symbolTableIDFromLocals(node *ast.Node) symbolTableID {
+	return stKindLocals | symbolTableID(ast.GetNodeId(node))
+}
+
+func symbolTableIDFromExports(sym *ast.Symbol) symbolTableID {
+	return stKindExports | symbolTableID(ast.GetSymbolId(sym))
+}
+
+func symbolTableIDFromMembers(sym *ast.Symbol) symbolTableID {
+	return stKindMembers | symbolTableID(ast.GetSymbolId(sym))
```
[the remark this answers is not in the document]
jakebailey: We ported from JS to Go, and JS only has floats, so maximally an ID could be 2^53-1, so this is not going to happen any more than we would have literally run out of IDs previously

DanielRosenwasser, line 420: `GetSymbolId` only gives you positive numbers, right? I ask to make sure you'll never overlap between `symbolTableIDFromMembers` and `symbolTableIDFromGlobals`.

internal/checker/symbolaccessibility.go
```diff
@@ -395,7 +393,35 @@ type accessibleSymbolChainContext struct {
...
+// symbolTableID uniquely identifies a symbol table by encoding its source.
+// The high 2 bits encode the kind (locals, exports, members, globals),
+// and the remaining bits encode the NodeId or SymbolId of the source.
+type symbolTableID uint64
+
+const (
+	stKindLocals  symbolTableID = 0
+	stKindExports symbolTableID = 1 << 62
+	stKindMembers symbolTableID = 2 << 62
+	stKindGlobals symbolTableID = 3 << 62
```
DanielRosenwasser: This really does read like a typo, so I would either point it out in the comment or rename this to `stKindMembersAndGlobals`

DanielRosenwasser: Also, it's a little subtle that these can't naively be used as flags in any way.

DanielRosenwasser (approved):

internal/checker/symbolaccessibility.go
jakebailey: They're uint64s, so yes

internal/checker/symbolaccessibility.go
jakebailey: I don't know what you mean by typo; it's an enum that's been shifted up to the top bits.

jakebailey: Would an iota be clearer?

DanielRosenwasser: Oh I totally misread this - I thought this was

```
1 << 61
1 << 62
1 << 62
```

and you technically could do something like that - that's why I asked the question earlier about overlap. The global scope could be `1 << ...` and symbol IDs for members could be `(1 << ...) | ast.GetSymbolId(sym)`.

Not sure why we would do it though.

DanielRosenwasser: I think an iota (or binary notation) could make it a bit easier to read. It's up to you.

PR #2715, opened 2026-02-09
Defer binding of expando assignments and give them lower priority

### LINKED ISSUE #2653: Static properties on manual (pre-ES6) classes can't be declared and then defined
## Steps to reproduce

```ts
declare class Thing {
  method(); number;
  static stuff(): number;
//       ~~~~
// error TS2300: Duplicate identifier 'stuff'.
}

function Thing() {}

Thing.prototype.method = () => 12;

Thing.stuff = () => 7;
//    ~~~~~
// error TS2300: Duplicate identifier 'stuff'.
```

[Playground](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEkoGdnwCoAsCWA7A5vAN4BQ88AtiAC6YD2wAFAJQBc8uArhQEYgwBuMvGTUo1bGBHVOAM1kt2XXvyEB6NeS3kAfnp0kN8fjDowMAZQBMAZgAMd9gBFOABwiTxCbKFwTZ2PzwAOSicrLBAHQkAL4kJLKcuGASdLgYOAQsxHEkWHj4ka6m1HTUAJ6uIJFUtAzwALzw2Q0AfPAAjFZCeZmFYfKNzcyN7QDs6ppa+nqGmiZmlrYOzm4eYF7wPiB+2AFBoTLyUUA)

The indicated TS2300 errors are not produced.

## Behavior with `tsgo`

The indicated TS2300 errors are produced.

ahejlsberg:
With this PR we defer binding of expando property assignments until all other declarations have have been bound, and create declaration sites only for expando property assignments to names for which there are no regular declarations.

We already have tests for scenarios similar to #2653, so no need to add a regression test.

Fixes #2653.


testdata/baselines/reference/submodule/conformance/privateNameJsBadDeclaration(target=es2015).symbols.diff
```diff
@@ -22,9 +22,12 @@
 +>B : Symbol(B, Decl(privateNameJsPrototype.js, 5, 1))
 
  B.prototype = {
- >B.prototype : Symbol(B.prototype, Decl(privateNameJsPrototype.js, 6, 11))
+->B.prototype : Symbol(B.prototype, Decl(privateNameJsPrototype.js, 6, 11))
 ->B : Symbol(B, Decl(privateNameJsPrototype.js, 5, 1), Decl(privateNameJsPrototype.js, 6, 11))
+->prototype : Symbol(B.prototype, Decl(privateNameJsPrototype.js, 6, 11))
++>B.prototype : Symbol(B.prototype)
 +>B : Symbol(B, Decl(privateNameJsPrototype.js, 5, 1))
- >prototype : Symbol(B.prototype, Decl(privateNameJsPrototype.js, 6, 11))
```
jakebailey, line 28: Is it a problem that we are introducing these diffs? Presumably this means we can't navigate through this node?

jakebailey: Or, I guess, this is saying that before had two declarations for the same thing, and now one wins?

ahejlsberg: Previously, we'd consider both the regular declaration and the expando declaration to be declaration sites. But then we'd sometimes ignore the expando declaration for purposes of determining the type. Now, symbols either have a regular declaration or one or more expando declarations, but never both. It doesn't really affect navigation, the potential expando sites just become regular reference sites.

PR #2716, opened 2026-02-09
Implement sync API without NAPI

jakebailey:
This is something I've wanted to try for a while, but was always unsuccessful at. This time, it finally works.

For our sync API, we use a `napi-rs` module to do what Node largely can't; synchronous communication to a child process. This is good and all, but complicates things because in addition to publishing a multi-platform package with Go, we also have to do that with Rust. That and, we all need to have rust installed, the package changes integrity every install, etc.

I've long wanted to try using Node's `fs.writeSync` / `fs.readSync` on raw file descriptors to do the same thing, in pure JS. Most of my attempts failed, or were seemingly slow.

However, this attempt did not seem to fail (at least on my machine). A summary copied from a temporarily-committed markdown file:

| Benchmark | libsyncrpc | Sync JS (new) | Async JS | Sync JS vs libsyncrpc | Async JS vs libsyncrpc |
|---|---|---|---|---|---|
| **Spawn** | 308 µs | 8.6 ms | 2.6 µs | 28× slower | 118× faster |
| **Load project** | 64 µs | 58 µs | 110 ms | 0.9× (faster) | 1719× slower |
| **Transfer debug.ts** | 1.6 ms | 1.3 ms | 29.7 ms | 0.8× (faster) | 18.6× slower |
| **Transfer program.ts** | 5.8 ms | 4.7 ms | 100.8 ms | 0.8× (faster) | 17.4× slower |
| **Transfer checker.ts** | 74.7 ms | 58.2 ms | 1240 ms | 0.8× (faster) | 16.6× slower |
| **Materialize program.ts** | 2.7 ms | 2.6 ms | 2.5 ms | 1.0× (~equal) | 0.9× (~equal) |
| **Materialize checker.ts** | 73.1 ms | 72.2 ms | 73.7 ms | 1.0× (~equal) | 1.0× (~equal) |
| **getSymbolAtPosition (1)** | 13.2 µs | 15.5 µs | 43.6 µs | 1.2× slower | 3.3× slower |
| **getSymbolAtPosition (10153, batched)** | 87.4 ms | 84.9 ms | 87.5 ms | 1.0× (~equal) | 1.0× (~equal) |
| **getSymbolAtLocation (10153)** | 406 ms | 442 ms | 798 ms | 1.1× slower | 2.0× slower |
| **getSymbolAtLocation (10153, batched)** | 282 ms | 267 ms | 278 ms | 0.9× (faster) | 1.0× (~equal) |
| **TS baseline (load project)** | 886 ms | 898 ms | 910 ms | — | — |
| **TS baseline (getSymbol 10153)** | 25.2 ms | 25.2 ms | 26.9 ms | — | — |

The pure JS version is effectively the same as going through the native code, except for startup time.

The async code looks mega fast due the fact that it can async start a process and then you can get a response later, but it's not like it's doing _less_ work, so I don't know if that is actually a downside. But clearly, Node has a huge overhead for spawning a process that a NAPI extension does not need.


jakebailey:
Seemingly this does not work on Windows, which hangs. I didn't exactly test there, so, fair.

_packages/api/src/syncChannel.ts
```diff
@@ -0,0 +1,562 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        this.readFd = (stdout as any)._handle.fd as number;
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        this.writeFd = (stdin as any)._handle.fd as number;
+
+        if (typeof this.readFd !== "number" || this.readFd < 0 || typeof this.writeFd !== "number" || this.writeFd < 0) {
+            throw new Error(
+                "SyncRpcChannel: could not obtain pipe file descriptors. " +
+                    "This implementation requires POSIX (Linux / macOS).",
+            );
```
jakebailey: Oh, duh, of course it doesn't work on Windows. Not sure what to do here.

jakebailey: named pipes to the rescue. Unlike unix sockets, they are actually openable with `openSync`.

To stop doing stdio in unix-y systems, we'd have to switch to two mkfifos

_packages/api/src/syncChannel.ts
```diff
@@ -0,0 +1,623 @@
...
+            }
+
+            // Set the pipe handles to blocking mode. Under node --test's
+            // process isolation, pipes are created in non-blocking mode
+            // (for the IPC channel). This causes readSync/writeSync to get
+            // EAGAIN, requiring costly 1ms sleeps per retry. Setting
+            // blocking mode ensures readSync blocks properly until data
+            // arrives, matching the behavior of the native libsyncrpc.
+            stdout._handle.setBlocking?.(true);
+            stdin._handle.setBlocking?.(true);
```
jakebailey, line 191: This may look sketchy, but this is what `tsc` has done all along, in fact

internal/api/server.go
```diff
@@ -46,8 +49,19 @@ func NewStdioServer(options *StdioServerOptions) *StdioServer {
 
 // Run starts the server and blocks until the connection closes.
 func (s *StdioServer) Run(ctx context.Context) error {
-	transport := NewStdioTransport(s.options.In, s.options.Out)
-	defer transport.Close()
+	var transport Transport
+	if s.options.PipePath != "" {
+		t, err := NewPipeTransport(s.options.PipePath)
+		if err != nil {
+			return fmt.Errorf("failed to create pipe transport: %w", err)
```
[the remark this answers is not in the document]
jakebailey: this is not a security boundary; if you can run tsgo with weird args, you can run `rm` with bad args

andrewbranch (approved):

Issue #2444, opened 2026-01-06
Missing code actions (`editor.codeActionsOnSave`) with `typescript.experimental.useTsgo` in VSCode


virtuallyunknown:
### Extension Version

0.20260106.1

### VS Code Version

1.107.1

### Operating system Version

Arch Linux

### Steps to reproduce

1. Initiate a new project, and install tsgo - `pnpm add -D @typescript/native-preview`.
2. Set `"type": "module"` in `package.json`.
3. Create a tsconfig.json with the recommended settings - `pnpm exec tsgo --init`, add the `src` directory to `includes`.
4. Install the tsgo extension - `TypeScriptTeam.native-preview` and create `.vscode/settings.json` file with the following config to enable it:

```json
{
    "typescript.experimental.useTsgo": true,
    "editor.codeActionsOnSave": {
    },
}
```

5. Create these 2 files in the src directory:

```js
// foo.ts
export const greet() {};

// bar.ts
greet();
```

### Issue

So what is the issue? The documentation for VSCode mentions the following:

> With the [editor.codeActionsOnSave](vscode://settings/editor.codeActionsOnSave) setting, you can configure a set of Code Actions that are automatically applied when you save a file, for example to organize imports. **_Based on your workspace files and active extensions, IntelliSense provides a list of available Code Actions._**

> ![](https://code.visualstudio.com/assets/docs/editing/refactoring/code-actions-on-save.png)

However, when `"typescript.experimental.useTsgo": true` is enabled, most of the code actions become unavailable so IntelliSense does not provide a list.

**TSGO enabled:**

<img width="620" height="262" alt="Image" src="https://github.com/user-attachments/assets/342c748e-987f-4253-9a05-44df2b2577be" />

**TSGO disabled:**

<img width="620" height="262" alt="Image" src="https://github.com/user-attachments/assets/8546259d-2f6f-432f-b188-3b55fa0cb4dc" />

So then if we go back to our `bar.ts` and save the file, normally the `source.addMissingImports.ts` action would kick in and automatically add the missing import for our `greet` function imported from `foo.ts`. But since these code actions are unavailable with `useTsgo` enabled, nothing happens.

This has become an important part of my workflow, and a huge time-saver as well. I'd really want to give tsgo a shot but this feature is dearly missed at the moment.

Cheers!

jakebailey:
I believe microsoft/typescript-go#2331 implemented these, but I don't know if they'll show in the editor like this. Worth testing on the next nightly.

[...]

PR #2717, opened 2026-02-09
parse formatting options in initialization, and pull `editor` scope during `Configuration` 

iisaduan:
- PR includes changes that allow VS to be configured:
  - allows clients to pass in preferences in `initializationOptions`
  - tested `didChangeConfiguration` to work in both VS and VSCode

- Updates the server's parsing of `Configuration` and `didChangeConfiguration` results to be inline with changes in #2766 and https://github.com/microsoft/vscode/issues/292934 / https://github.com/microsoft/vscode/pull/295055

- Bugfixes: 
  - fixes indenting issue [#2444 (comment)](https://github.com/microsoft/typescript-go/issues/2444#issuecomment-3892747174) and https://github.com/microsoft/typescript-go/pull/2331#issuecomment-3814678385 where organize imports/formatting was ignoring `indentSize = 2`.

### DIFF
--- a/internal/lsp/lsproto/lsp_generated.go
+++ b/internal/lsp/lsproto/lsp_generated.go
@@ -21626,7 +21626,7 @@ type InitializationOptions struct {
 	// The client-side command name that resolved references/implementations `CodeLens` should trigger. Arguments passed will be `(DocumentUri, Position, Location[])`.
 	CodeLensShowLocationsCommandName *string `json:"codeLensShowLocationsCommandName,omitzero"`
 
-	// userPreferences and/or formatting options provided at initialization.
+	// userPreferences and/or formatting options if provided at initialization.
 	UserPreferences *any `json:"userPreferences,omitzero"`
 }
 


internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -48,6 +48,12 @@ const customStructures: Structure[] = [
                 optional: true,
                 documentation: "The client-side command name that resolved references/implementations `CodeLens` should trigger. Arguments passed will be `(DocumentUri, Position, Location[])`.",
             },
+            {
+                name: "typescript",
```
jakebailey: Can we name this differently? It's strange to just have the word "typescript"; all options here are for typescript. Maybe just call it UserPreferences?

internal/lsp/server.go
```diff
@@ -1081,14 +1115,18 @@ func (s *Server) handleExit(ctx context.Context, params any) error {
...
-		s.session.Configure(lsutil.ParseNewUserConfig(settings))
 	} else if settings, ok := params.Settings.(map[string]any); ok {
-		// fourslash case
-		s.session.Configure(lsutil.ParseNewUserConfig([]any{settings["js/ts"], settings["typescript"], settings["javascript"]}))
+		for key := range settings {
+			switch key {
+			case "js/ts", "typescript", "javascript", "editor":
+				continue
+			default:
+				delete(settings, key)
```
jakebailey: Why are we mutating the input?

internal/lsp/lsproto/_generate/generate.mts
iisaduan: @navya9singh can you make this change on your end as well?

internal/lsp/server.go
iisaduan: ah, this was a remnant from a different setup i had. I'll change it

internal/lsp/lsproto/_generate/generate.mts
navya9singh: I can. It would end up looking something like this:
```
"initializationOptions": {
    "UserPreferences": {
      "format": {},
      "preferences": {},
      "project": {}
    }
}
```

jakebailey: Er, I mean `userPreferences`, lowercased like the rest.

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -48,6 +48,12 @@ const customStructures: Structure[] = [
                 optional: true,
                 documentation: "The client-side command name that resolved references/implementations `CodeLens` should trigger. Arguments passed will be `(DocumentUri, Position, Location[])`.",
             },
+            {
+                name: "typescript",
+                type: { kind: "reference", name: "any" },
+                optional: true,
+                documentation: "userPreferences and/or formatting options or provided at initialization.",
```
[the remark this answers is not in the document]
jakebailey: There's definitely a typo here. Not sure what it should be

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -780,7 +786,7 @@ function handleOrType(orType: OrType): GoType {
 
     let memberNames = nonNullTypes.map(type => {
         if (type.kind === "reference") {
-            return type.name;
+            return type.name.slice(type.name.lastIndexOf(".") + 1);
```
jakebailey: Why is this needed?

internal/lsp/server.go
```diff
@@ -298,12 +309,35 @@ func (s *Server) RequestConfiguration(ctx context.Context) (*lsutil.UserConfig,
 			{
 				Section: new("javascript"),
 			},
+			{
+				Section: new("editor"),
+			},
```
jakebailey, line 314: we're now requesting editor, so I think fourslash also has to update.

iisaduan: what part of fourslash?

internal/lsp/lsproto/_generate/generate.mts
iisaduan: was trying something where I was doing a reference to types in `core`, and it broke since it was making go type names that had a `.` in it

internal/lsp/server.go
jakebailey: Hm, I guess we still lie and do this:

```go
	case lsproto.MethodWorkspaceConfiguration:
		// Return current user preferences
		response = &lsproto.ResponseMessage{
			ID:      req.ID,
			JSONRPC: req.JSONRPC,
			Result:  []any{f.userPreferences},
		}
```

Wish we didn't, nevermind

internal/lsp/lsproto/_generate/generate.mts
jakebailey: Yeah, but we shouldn't do that, I don't think. Is this needed for the PR, or just leftover?

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -48,6 +48,12 @@ const customStructures: Structure[] = [
                 optional: true,
                 documentation: "The client-side command name that resolved references/implementations `CodeLens` should trigger. Arguments passed will be `(DocumentUri, Position, Location[])`.",
             },
+            {
+                name: "userPreferences",
+                type: { kind: "reference", name: "any" },
```
jakebailey, line 53: I feel a little nervous about `any`, but it's probably fine given how we are walking it. In the future I'd like to define a type for this (but that can probably come in my PR if I ever get around to fixing it)

internal/lsp/server.go
```diff
@@ -1081,14 +1115,10 @@ func (s *Server) handleExit(ctx context.Context, params any) error {
 }
 
 func (s *Server) handleDidChangeWorkspaceConfiguration(ctx context.Context, params *lsproto.DidChangeConfigurationParams) error {
-	// !!! only implemented because needed for fourslash
 	if params.Settings == nil {
 		return nil
-	} else if settings, ok := params.Settings.([]any); ok {
```
jakebailey, line 1087: Does this actually never send as an array?

jakebailey (approved):

internal/lsp/server.go
iisaduan: I assumed vscode would do that because that's what the response was for `workspace/configuration`, but after testing it in quite a few ways, vscode sends it as an object. VS also sends it as an object, so i saw no reason to keep the unlabeled array format

PR #2718, opened 2026-02-09
Update submodule to latest 6.0 main pre-beta

jakebailey:
This pulls in all of the new lib changes, ES2025, etc.

I did not fully port the ES2025 PR (the suggestions), only enough to make everything work reasonably.

### DIFF
--- a/internal/fourslash/tests/gen/completionForStringLiteral11_test.go
+++ b/internal/fourslash/tests/gen/completionForStringLiteral11_test.go
@@ -12,7 +12,8 @@ func TestCompletionForStringLiteral11(t *testing.T) {
 	fourslash.SkipIfFailing(t)
 	t.Parallel()
 	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
-	const content = `type As = 'arf' | 'abacus' | 'abaddon';
+	const content = `// @stableTypeOrdering: true
+type As = 'arf' | 'abacus' | 'abaddon';
 let a: As;
 switch (a) {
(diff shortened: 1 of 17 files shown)


testdata/baselines/reference/submodule/compiler/augmentExportEquals5.types.diff
```diff
@@ -0,0 +1,25 @@
+--- old.augmentExportEquals5.types
++++ new.augmentExportEquals5.types
```
jakebailey, line 2: These diffs are introduced temporarily, as my other PR for the node builder host needs to be merged.

internal/fourslash/tests/gen/completionForStringLiteral11_test.go
```diff
@@ -12,7 +12,8 @@ func TestCompletionForStringLiteral11(t *testing.T) {
 	fourslash.SkipIfFailing(t)
 	t.Parallel()
 	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
-	const content = `type As = 'arf' | 'abacus' | 'abaddon';
+	const content = `// @stableTypeOrdering: true
```
gabritto, line 15: Why was this added to tests in Strada?

jakebailey: So that tsgo-port could contain baseline only updates. These modified tests actually hardcode the strings which makes dealing with them annoying.

PR #2761, opened 2026-02-11
Bump to Go 1.26

jakebailey:
Notably, compiling tsgo itself takes 30% less time, and `ptrTo` is now gone, as `new` performs that same task in the language itself.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -501,7 +501,7 @@ func (s *Server) sendResult(id *jsonrpc.ID, result any) error {
 
 func (s *Server) sendError(id *jsonrpc.ID, err error) error {
 	code := lsproto.ErrorCodeInternalError
-	if errCode := lsproto.ErrorCode(0); errors.As(err, &errCode) {
+	if errCode, ok := errors.AsType[lsproto.ErrorCode](err); ok {
 		code = errCode
 	}
 	// TODO(jakebailey): error data
--- a/internal/vfs/vfstest/vfstest.go
+++ b/internal/vfs/vfstest/vfstest.go
@@ -252,6 +252,11 @@ func (e *brokenSymlinkError) Error() string {
 	return fmt.Sprintf("broken symlink %q -> %q", e.from, e.to)
 }
 
+func isBrokenSymlinkError(err error) bool {
+	_, ok := errors.AsType[*brokenSymlinkError](err)
+	return ok
+}
+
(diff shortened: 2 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,7 @@
 {
     "name": "TypeScript Compiler Development",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-    "image": "mcr.microsoft.com/devcontainers/go:2-1.25-bookworm",
+    "image": "mcr.microsoft.com/devcontainers/go:2-1.26-bookworm",
```
jakebailey: This image does not yet exist, pending https://github.com/devcontainers/images/pull/1776 (or similar PR).

jakebailey: Probably worth waiting, or reverting and relying on GOTOOLCHAIN fixing the devcontainer.

jakebailey: Hm, the devcontainer (or its Go install) has GOTOOLCHAIN=local set, so it doesn't automatically work either. Probably just means we have to wait.

jakebailey: I've opted to just make sure `GOTOOLCHAIN=auto` in the devcontainer, for now

jakebailey:
This came out quite literally yesterday; I'm not sure when the right time to do this is given it's not in the Arch Linux repos, longsleep PPA, homebrew.

gabritto:
> This came out quite literally yesterday; I'm not sure when the right time to do this is given it's not in the Arch Linux repos, longsleep PPA, homebrew.

What consequences could this have? Just contributors not being able to build tsgo?
We should probably wait a bit then, but it's always great to wait less for builds.

jakebailey:
> What consequences could this have? Just contributors not being able to build tsgo?

Well, with `GOTOOLCHAIN`, it'll auto download a newer release, so not really a big deal. The Microsoft build of Go we use in CI and for releases already works.

The worst case is that `delve` doesn't work with Go 1.26 yet, but I believe this already works (and I run delve built from `gotip` anyway).

jakebailey:
Also, on my laptop, `go clean -cache -testcache; hereby test` takes 4m20s, but on this branch, it only takes 2m6s.

jakebailey:
On my desktop:

image

jakebailey:
A downside is that tinygo doesn't have Go 1.26 yet, and I like to test things with it, but it's also broken for other reasons anyway. So probably not a blocker.

jakebailey:
brew has Go 1.26. That's enough for me to feel comfortable merging this; GOTOOLCHAIN will work fine for the rest.

PR #2764, opened 2026-02-11
Fix invalid params error handling

gabritto:
Found this while testing the fuzzer: if we failed to unmarshal params for a request, we weren't really sending the error message with the proper id (this was causing the fuzzer to hang). We were also using error code invalid request even when we should the more specific invalid params error.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -2169,6 +2169,39 @@ func (f *FourslashTest) VerifyOutliningSpans(t *testing.T, foldingRangeKind ...l
 	}
 }
 
+// FoldingRangeLineExpected represents expected start and end lines for a folding range.
+type FoldingRangeLineExpected struct {
+	StartLine uint32
+	EndLine   uint32
+}
+
+// VerifyFoldingRangeLines verifies folding ranges by comparing only start and end lines.
+// This is useful for testing with lineFoldingOnly where character positions are ignored.
+func (f *FourslashTest) VerifyFoldingRangeLines(t *testing.T, expected []FoldingRangeLineExpected) {
+	params := &lsproto.FoldingRangeParams{
+		TextDocument: lsproto.TextDocumentIdentifier{
+			Uri: lsconv.FileNameToDocumentURI(f.activeFilename),
+		},
+	}
+	result := sendRequest(t, f, lsproto.TextDocumentFoldingRangeInfo, params)
+	if result.FoldingRanges == nil {
+		t.Fatalf("Nil response received for folding range request")
+	}
+
+	actualRanges := *result.FoldingRanges
+	if len(actualRanges) != len(expected) {
+		t.Fatalf("verifyFoldingRangeLines failed - expected %d ranges, got %d", len(expected), len(actualRanges))
+	}
+
+	for i, exp := range expected {
+		got := actualRanges[i]
+		if got.StartLine != exp.StartLine || got.EndLine != exp.EndLine {
+			t.Errorf("verifyFoldingRangeLines failed - range %d: expected (startLine=%d, endLine=%d), got (startLine=%d, endLine=%d)",
+				i, exp.StartLine, exp.EndLine, got.StartLine, got.EndLine)
+		}
+	}
+}
+
 func (f *FourslashTest) VerifyBaselineHover(t *testing.T) {
 	markersAndItems := core.MapFiltered(f.Markers(), func(marker *Marker) (markerAndItem[*lsproto.Hover], bool) {
 		if marker.Name == nil {
--- a/internal/fourslash/tests/foldingRangeLineFoldingOnly_test.go
+++ b/internal/fourslash/tests/foldingRangeLineFoldingOnly_test.go
@@ -1,0 +1,114 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestFoldingRangeLineFoldingOnly(t *testing.T) {
+	t.Parallel()
+
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `if (EMPTY_TAGs.has(tag)) {
+  output += "/>";
+} else {
+  output += ">";
+
+  if (!html && kidcount > 0) {
+    //
+  }
+}
+
+export function use<T>(ctx: any): T | undefined {
+  //
+}`
+	ptrTrue := true
+	capabilities := &lsproto.ClientCapabilities{
+		TextDocument: &lsproto.TextDocumentClientCapabilities{
+			FoldingRange: &lsproto.FoldingRangeClientCapabilities{
+				LineFoldingOnly: &ptrTrue,
+				FoldingRange: &lsproto.ClientFoldingRangeOptions{
+					CollapsedText: &ptrTrue,
+				},
+			},
+		},
+	}
+	f, done := fourslash.NewFourslash(t, capabilities, content)
+	defer done()
+
+	// With lineFoldingOnly, end lines should be adjusted so closing brackets stay visible.
+	// Line 0: if (EMPTY_TAGs.has(tag)) {
+	// Line 1:   output += "/>";
+	// Line 2: } else {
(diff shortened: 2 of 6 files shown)


internal/lsp/server.go
```diff
@@ -334,8 +337,14 @@ func (s *Server) readLoop(ctx context.Context) error {
...
+				var id *jsonrpc.ID
+				if errors.Is(err, lsproto.ErrorCodeInvalidParams) {
+					if msg.Kind == jsonrpc.MessageKindRequest {
+						id = msg.AsRequest().ID
+					}
+				}
+				if err := s.sendError(id, err); err != nil {
 					return err
 				}
 				continue
```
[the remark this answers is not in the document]
gabritto: This comment seems correct per https://www.jsonrpc.org/specification#notification, but we were already sending error responses to invalid notifications before this PR. Should I also fix that?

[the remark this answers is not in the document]
jakebailey: Yes, there's no "replying" to notifications

internal/lsp/server_invalid_message_test.go
jakebailey, line 1: Does this file need to be named so specifically?

internal/lsp/server.go
```diff
@@ -105,6 +105,9 @@ func (r *lspReader) Read() (*lsproto.Message, error) {
 
 	req := &lsproto.Message{}
 	if err := json.Unmarshal(data, req); err != nil {
+		if errors.Is(err, lsproto.ErrorCodeInvalidParams) {
+			return req, fmt.Errorf("%w: %w", lsproto.ErrorCodeInvalidParams, err)
```
jakebailey, line 109: IF this is already InvalidParams, why reencode it?

gabritto: I'm not sure, the previous code was already doing it and I wasn't sure if it mattered or not which error ended up on top. If we don't rewrap this again in an lsproto error, the error on top will be from the json library. Is that ok?

jakebailey: Ah, that's true.

jakebailey: I think we can just leave it.

It's a little awkward to return something and also a non-nil error, but it's not the worst thing

internal/lsp/server_invalid_message_test.go
gabritto: Moved everything to a single server test, not sure what convention we usually follow, but I find it easier to find relevant tests if things are kept in the same file.

internal/lsp/server.go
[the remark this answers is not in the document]
gabritto: Added a check for this in `sendError`'s implementation to avoid having to modify its callers.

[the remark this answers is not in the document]
DanielRosenwasser: At best you could "send back" a notification to indicate something went wrong.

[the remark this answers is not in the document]
jakebailey: Right. That would be a regular log message at that point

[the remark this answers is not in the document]
gabritto: Added some logging code.

internal/lsp/server_test.go
```diff
@@ -0,0 +1,236 @@
...
+		In:  reader,
+		Out: writer,
+		Err: io.Discard,
+		Cwd: "/test",
+	})
+
+	err := server.Run(t.Context())
+	assert.NilError(t, err)
+
+	server.ProcessPendingRequests()
```
gabritto: Couldn't really think of a better way to write these tests. `server.Run` returns as soon as the read loop reaches EOF, so there's no guarantee about possibly remaining requests to be processed.

jakebailey: If Run returns, shouldn't all of those be cancelled via context, and exit at their own pace?

gabritto: They don't check if there are pending requests in the queue and just exit as soon as the context is cancelled:
```go
func (s *Server) dispatchLoop(ctx context.Context) error {
	ctx, lspExit := context.WithCancelCause(ctx)
	defer lspExit(nil)
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case req := <-s.requestQueue:
		...
```
Do you mean this shouldn't be happening?

jakebailey: What I was meaning is that the pending requests should be able to stay in the queue, it's just that whoever is still waiting on those requests should have also seen a context cancellation too, so give up waiting and start returning upward.

gabritto: Yep, but in this case the test doesn't want things to be cancelled 😬.
I think the alternative to having something like this would be to set up a custom reader for the server input to avoid the server reaching EOF before the test runs, but I think the current approach is simpler.

jakebailey: I'm confused, why is the test looking at anything after the server has exited? A real server wouldn't be able to do this...

gabritto: The test wants to look at the responses for the requests it sent, but the server may not process all of them because it exits as soon as it reads everything from the input string. A real client wouldn't necessarily worry about waiting for everything in the server to finish, it would (1) wait for pending responses to request, (2) not immediately close the input stream after sending the requests, but the test does the opposite. I could fix (2) in the test, but still it would be clanky, because the test wants to see what all responses and notifications the server sends.

jakebailey: Sure, but why does the test looking at responses after the server has closed, if it's the test itself that's closing the server? That's what's confusing me. The server should not exit if the client still cares about it

gabritto: I don't know how I'd do that. I'd have to create a different reader to avoid the server exiting, but even then I'd have to have some way to check if all messages have been processed, so I'd still need to add a thing like "wait until the queues are empty" to write the test.

jakebailey: You're still implying that something is attempting to use the server after it's exited, but it can only exit because the test is ending (context cancel) or because it was asked to do so, both of which are controlled by the outer test. That's what I'm confused about.

gabritto: Right now the test doesn't explicitly control it, the server exits once it's read everything in the input. What I'm saying is that even if I change the test so that the server doesn't exit right away (e.g. the server's read loop blocks after reading all the messages instead of returning EOF), I would still need some extra code to somehow detect when the server has processed all of the input messages.

jakebailey: I'm still confused, the code I refactored it into was:

```go
	g.Go(func() error {
		defer outputWriter.Close()
		return server.Run(ctx)
	})

	// Start async message router
	g.Go(func() error {
		return f.messageRouter(ctx)
	})
```

both are context bound, then later, when the test is done, then it cancels the context.

jakebailey: From DMs, I did not realize that this was a test using `bytes.Reader`, so it was EOF-ing. Maybe we just don't bother attempting to test this particular thing, since the root cause is something silly like attempting to reply with a nil ID

internal/lsp/lsproto/jsonrpc.go
```diff
@@ -124,7 +124,7 @@ func (r *RequestMessage) UnmarshalJSON(data []byte) error {
 
 type ResponseMessage struct {
 	JSONRPC jsonrpc.JSONRPCVersion `json:"jsonrpc"`
-	ID      *jsonrpc.ID            `json:"id,omitzero"`
+	ID      *jsonrpc.ID            `json:"id"`
```
gabritto, line 127: Per the spec, if we fail to decode a request or notification (i.e. the whole thing, not just params), we are allowed to return an error response with `null` id.

internal/lsp/server.go
```diff
@@ -334,8 +337,14 @@ func (s *Server) readLoop(ctx context.Context) error {
 		}
 		msg, err := s.read()
 		if err != nil {
-			if errors.Is(err, lsproto.ErrorCodeInvalidRequest) {
-				if err := s.sendError(nil, err); err != nil {
+			if errors.Is(err, lsproto.ErrorCodeInvalidRequest) || errors.Is(err, lsproto.ErrorCodeInvalidParams) {
+				var id *jsonrpc.ID
+				if errors.Is(err, lsproto.ErrorCodeInvalidParams) {
+					if msg.Kind == jsonrpc.MessageKindRequest {
```
jakebailey: For insurance, maybe guard `msg` with a nil check here?

gabritto: That should be an error, but yeah, ok, maybe not worth crashing over it.

PR #2776, opened 2026-02-12
Add missing IsNodeType functions

trevorade:
Adds all missing `IsNodeType` functions in `ast.go` for consistency.

Also normalizes comments before a `Node` block.


jakebailey:
Did you need these for some purpose?

trevorade:
Hey Jake,

We're working on our internal TypeScript to Closure JavaScript converter at Google. We wanted to use some of these `IsX` functions. We don't need all the ones I add here but I figured it makes sense to consistently have them for all Node types with data.

If you guys don't want this, we can just add a local utility package with these functions but we figured you would want them at some point hence the PR.

PR #2787, opened 2026-02-14
Error on ES5, fix skipped tests

jakebailey:
Largely ports https://github.com/microsoft/TypeScript/pull/63067

This also removes ES3 as a value we even recognize. ES3 is the same as "unset", so in the PR where I bumped the submodule and skipped ES3/ES5 tests, I accidentally skipped all tests without any target set.

Such is the risk of having a valid value that is also the unset value.


jakebailey:
I have a fix for that skipped test; it's the same thing as #2134 but for functions/constructors. Enums also have the bug.

PR #2789, opened 2026-02-15
Fix token cache parent mismatch panics

ahejlsberg:
This PR fixes the 11 occurrences of "Token cache mismatch: parent" panics observed in #2778.

The PR includes a fourslash test that reproduces the issue in [this run](https://github.com/microsoft/typescript-go/issues/2778#issuecomment-3894721967). The others are of a similar nature, so no need for further tests.

### DIFF
--- a/internal/fourslash/tests/completionJSDocNoCrash_test.go
+++ b/internal/fourslash/tests/completionJSDocNoCrash_test.go
@@ -11,7 +11,8 @@ func TestCompletionJSDocNoCrash(t *testing.T) {
 	t.Parallel()
 	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
 	const content = `
-// @filename: file.ts
+// @allowJs: true
+// @filename: file.js
 class ErrorMap {
   /**
    * @type {string}


ahejlsberg:
In an ideal world we wouldn't need a `parent` in the token cache key because a token a some location would always be a child of the same parent. However, the various token walking functions in `tokens.go` sometimes compute different parents in boundary conditions inside JSDoc comments, and the parent rarely matters. So, from a stability point of view, I think we're better off just saying that it is fine to have differently parented tokens at the same location.

internal/ast/ast.go
```diff
@@ -10912,6 +10912,11 @@ type HasFileName interface {
 	Path() tspath.Path
 }
 
+type TokenCacheKey struct {
+	parent *Node
```
jakebailey, line 10916: Should this be a node ID so we avoid keeping synthetic nodes alive for an extended period of time? (Maybe they're cached already such that it doesn't much matter.)

internal/fourslash/tests/completionJSDocNoCrash_test.go
```diff
@@ -0,0 +1,29 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func completionJSDocNoCrash(t *testing.T) {
```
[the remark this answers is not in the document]
jakebailey: Yeah, this test isn't running

internal/ast/ast.go
ahejlsberg: No need. The token cache lives on the containing `SourceFile`, which keeps the parent alive anyway.

internal/fourslash/tests/completionJSDocNoCrash_test.go
```diff
@@ -0,0 +1,31 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestCompletionJSDocNoCrash(t *testing.T) {
```
jakebailey, line 10: If I run this test on main, this test doesn't seem to panic.

ahejlsberg: Hmm, it may be that some other token querying operation has to precede the completions request in order to trigger the issue.

ahejlsberg: Test has to be a .js file to trigger the issue. Now fixed.

ahejlsberg:
@jakebailey Is there a typescript-bot command that'll kick off another fuzzer run?

jakebailey:
Nope, that run was manual, as Gabriela just added the pipeline.

Issue #2649, opened 2026-02-03
LSP Panic: strings: negative Repeat count when formatting JSDoc within nested scope


ButbkaDrug:
The LSP server crashes (panics) when attempting to format a JavaScript/TypeScript file containing a JSDoc @type annotation inside a callback function. 

## Stack trace

```
[ERROR][2026-02-03 12:29:18] ...lsp/handlers.lua:562	"panic handling request textDocument/formatting: strings: negative Repeat count
goroutine 204 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x5e
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0xc000232708, {0x441ae5?, 0xc0007aafc0?}, 0xc003402660)
	github.com/microsoft/typescript-go/internal/lsp/server.go:780 +0x4c
panic({0xc01ae0?, 0x10ffd40?})
	runtime/panic.go:783 +0x132
strings.Repeat({0x10fc098?, 0xc0002c9034?}, 0xc0033d1c40?)
	strings/strings.go:628 +0x585
github.com/microsoft/typescript-go/internal/format.getIndentationString(0xffffffffffffffff, 0xc001cfe000?)
	github.com/microsoft/typescript-go/internal/format/span.go:980 +0x7f
github.com/microsoft/typescript-go/internal/format.(*formatSpanWorker).insertIndentation(0xc0003f8690, 0x38, 0xffffffffffffffff, 0x0)
	github.com/microsoft/typescript-go/internal/format/span.go:863 +0x45
github.com/microsoft/typescript-go/internal/format.(*formatSpanWorker).indentMultilineComment(0xc0003f8690, {0x0?, 0x0?}, 0xffffffffffffffff, 0x0, 0x1)
	github.com/microsoft/typescript-go/internal/format/span.go:922 +0x185
github.com/microsoft/typescript-go/internal/format.(*formatSpanWorker).indentTriviaItems(0xc0003f8690, {0xc0003a9560?, 0x62?, 0xc000145700?}, 0xffffffffffffffff, 0xc8?, 0xc0033d1ed0)
	github.com/microsoft/typescript-go/internal/format/span.go:899 +0xcb
[...]

PR #2793, opened 2026-02-16
Fixed a formatting crash on comments inside blocks starting on first line

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/2649

### DIFF
--- a/internal/format/span.go
+++ b/internal/format/span.go
@@ -343,7 +343,7 @@ func (w *formatSpanWorker) processChildNode(
 ) int {
 	debug.Assert(!ast.NodeIsSynthesized(child))
 
-	if ast.NodeIsMissing(child) || isGrammarError(parent, child) {
+	if ast.NodeIsMissing(child) || isGrammarError(parent, child) || child.Flags&ast.NodeFlagsReparsed != 0 {
 		return inheritedIndentation
 	}
 


DanielRosenwasser (approved):

internal/format/span.go
```diff
@@ -1059,9 +1059,11 @@ func (w *formatSpanWorker) consumeTokenAndAdvanceScanner(currentTokenInfo tokenI
...
 		if len(currentTokenInfo.leadingTrivia) > 0 {
 			commentIndentation := dynamicIndenation.getIndentationForComment(currentTokenInfo.token.Kind, tokenIndentation, container)
-			indentNextTokenOrTrivia = w.indentTriviaItems(currentTokenInfo.leadingTrivia, commentIndentation, indentNextTokenOrTrivia, func(item TextRangeWithKind) {
-				w.insertIndentation(item.Loc.Pos(), commentIndentation, false)
-			})
+			if commentIndentation != -1 {
+				indentNextTokenOrTrivia = w.indentTriviaItems(currentTokenInfo.leadingTrivia, commentIndentation, indentNextTokenOrTrivia, func(item TextRangeWithKind) {
+					w.insertIndentation(item.Loc.Pos(), commentIndentation, false)
+				})
+			}
```
jakebailey: But why? The original code said:

```ts
                let indentNextTokenOrTrivia = true;
                if (currentTokenInfo.leadingTrivia) {
                    const commentIndentation = dynamicIndentation.getIndentationForComment(currentTokenInfo.token.kind, tokenIndentation, container);
                    indentNextTokenOrTrivia = indentTriviaItems(currentTokenInfo.leadingTrivia, commentIndentation, indentNextTokenOrTrivia, item => insertIndentation(item.pos, commentIndentation, /*lineAdded*/ false));
                }
```

which to me means the bug is elsewhere

jakebailey: Related https://github.com/microsoft/typescript-go/pull/2664/changes

Andarist: Oh, I have not seen the Copilot PR before. It wasn't linked to the issue.

You were right to pushback on my approach. Initially, I fixed it on my own without looking into the original. I pushed out now a more faithful fix.

internal/format/span.go
```diff
@@ -221,6 +221,7 @@ func getNonDecoratorTokenPosOfNode(node *ast.Node, file *ast.SourceFile) int {
 func (w *formatSpanWorker) execute(s *formattingScanner) []core.TextChange {
 	w.formattingScanner = s
 	w.indentationOnLastIndentedLine = -1
+	w.lastIndentedLine = -1
```
jakebailey, line 224: Interesting; `lastIndentedLine` was seemingly undefined in Strada, not `-1`; how much does that matter? Did it act like `0`?

Andarist: yes exactly, it acted like 0 and that changed the result of this comparison: https://github.com/microsoft/typescript-go/blob/8c9cc90f81d6f693ad61cb4a5361560b6c98ef5e/internal/format/span.go#L545

PR #2796, opened 2026-02-16
Record (and possibly rewind) JSDoc info, create JSDoc cache separately

ahejlsberg:
This PR is a follow-up to #2795 that records (and possibly rewinds) JSDoc comment occurrences and creates the JSDoc cache as a separate step when finalizing a source file object.

No tests associated with this PR as all it does is eliminate speculatively parsed and discarded JSDoc objects from the JSDoc cache.

### DIFF
--- a/internal/parser/jsdoc.go
+++ b/internal/parser/jsdoc.go
@@ -31,12 +31,6 @@ func (p *Parser) withJSDoc(node *ast.Node, hasJSDoc bool) []*ast.Node {
 	if !hasJSDoc {
 		return nil
 	}
-
-	if p.jsdocCache == nil {
-		p.jsdocCache = make(map[*ast.Node][]*ast.Node, strings.Count(p.sourceText, "/**"))
-	} else if _, ok := p.jsdocCache[node]; ok {
-		panic("tried to set JSDoc on a node with existing JSDoc")
-	}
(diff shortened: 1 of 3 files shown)


internal/parser/jsdoc.go
```diff
@@ -31,12 +31,6 @@ func (p *Parser) withJSDoc(node *ast.Node, hasJSDoc bool) []*ast.Node {
 	if !hasJSDoc {
 		return nil
 	}
-
-	if p.jsdocCache == nil {
-		p.jsdocCache = make(map[*ast.Node][]*ast.Node, strings.Count(p.sourceText, "/**"))
```
jakebailey, line 36: This presizing was important for perf; can this be retained somewhere? Maybe not, given we don't want to do this when we're not parsing out jsdoc.

(I can test the perf of this later)

ahejlsberg: I think presizing is a lot more important for maps. We now collect the JSDoc information in an array and then build the map as we finalize the source file. The array is subject to reallocation, but the map is preallocated in its exact final size. We could presize the array, but I doubt it matters much.

jakebailey: I'll profile it later, not blocking

(This is a map, though)

PR #2811, opened 2026-02-17
Fixed completion crash on JSDoc nodes created by `reparseJSDocComment`

Andarist:
Fixes a crash reported here https://github.com/microsoft/typescript-go/issues/2778#issuecomment-3894722157


internal/parser/reparser.go
```diff
@@ -239,6 +239,7 @@ func (p *Parser) reparseJSDocComment(node *ast.Node, tag *ast.Node) {
 	if comment := tag.CommentList(); comment != nil {
 		propJSDoc := p.factory.NewJSDoc(comment, nil)
 		p.finishReparsedNode(propJSDoc, tag)
+		propJSDoc.Parent = node
```
Andarist, line 242: This is the only place that creates a **new** JSDoc nodes that goes exclusively into `jsdocInfos` (never into `reparseList`). It's the same `.Parent` assignment as the one in `withJSDoc`.

PR #2826, opened 2026-02-18
Add replay test

gabritto:
*To debug a replay file obtained from the fuzzer*:
- Edit the "Launch replay test" task in your `launch.json` with a path to the replay file and the test directory (e.g. `/code/replay.txt`, `code/Project`)
- Launch the task

This PR adds a test + launch task to run the commands present in a replay script obtained from the fuzzer. The test is skipped unless the `replay` flag is provided.
This is needed for us to be able to conveniently debug problems found by the fuzzer. With Strada, we used to be able to launch tsserver with `--inspect-brk=...` and then attach to that process for debugging, but IIUC there isn't an equivalent thing in Go.

I refactored the LSP client implementation out of fourslash for reuse in the replay test.

### DIFF
--- a/_extension/src/util.ts
+++ b/_extension/src/util.ts
@@ -1,8 +1,5 @@
 import * as path from "path";
 import * as vscode from "vscode";
-import packageJson from "../package.json";
-
-const version = packageJson.version;
 
 export const aiConnectionString = "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255";
 
@@ -68,7 +65,7 @@ export async function getExe(context: vscode.ExtensionContext): Promise<ExeInfo>
 
     return {
         path: getBuiltinExePath(context),
-        version,
+        version: context.extension.packageJSON.version,
     };
 }
 
--- a/internal/fourslash/tests/signatureHelpContextualConstructSignatureNoCrash_test.go
+++ b/internal/fourslash/tests/signatureHelpContextualConstructSignatureNoCrash_test.go
@@ -1,0 +1,30 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+// Tests that signature help does not panic when the contextual type has only construct signatures
+// (no call signatures).
+func TestSignatureHelpContextualConstructSignatureNoCrash(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `
+type Obj = {
+    foo: new () => object
+}
+
+let obj: Obj = {
+    foo(/*constructOnly*/) {}
+}
+`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	// When contextual type only has construct signatures (no call signatures),
+	// no signature help should be provided (and no panic should occur).
+	f.GoToMarker(t, "constructOnly")
+	f.VerifyNoSignatureHelp(t)
+}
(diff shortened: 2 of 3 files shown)


internal/lsp/replay_test.go
jakebailey, line 1: Does this make more sense in another package? It feels strange to have this in a `main`, which is more or less unrelated.

If it's because of the `System` implementation, we should probably move that.

DanielRosenwasser:
> With Strada, we used to be able to launch tsserver with `--inspect-brk=...` and then attach to that process for debugging, but IIUC there isn't an equivalent thing in Go.

Isn't it possible to run with [`dlv exec`](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_exec.md) in place of the usual binary path?

I'm guessing the big issue is that `tsgo` actually starts up Node.js, and it's painful to get the platform-native executable path.

internal/lsp/replay_test.go
gabritto: I can move it to `internal/lsp`, only thing that breaks is `getGlobalTypingsCacheLocation`, which I'm not sure where to move.

jakebailey: I think `getGlobalTypingsCacheLocation` should move into `osvfs`.

gabritto:
> > With Strada, we used to be able to launch tsserver with `--inspect-brk=...` and then attach to that process for debugging, but IIUC there isn't an equivalent thing in Go.
> 
> Isn't it possible to run with [`dlv exec`](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_exec.md) in place of the usual binary path?
> 
> I'm guessing the big issue is that `tsgo` actually starts up Node.js, and it's painful to get the platform-native executable path.

Yes, but you also want the executable that has been built 

> > With Strada, we used to be able to launch tsserver with `--inspect-brk=...` and then attach to that process for debugging, but IIUC there isn't an equivalent thing in Go.
> 
> Isn't it possible to run with [`dlv exec`](https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_exec.md) in place of the usual binary path?
> 
> I'm guessing the big issue is that `tsgo` actually starts up Node.js, and it's painful to get the platform-native executable path.

You're right, I had forgotten about that, but another problem that was on my mind is that you want a binary that was compiled without optimizations, otherwise it's gonna be a bad time debugging it. In general it seemed simpler to me for the workflow to use a test in our repo, unless we think the fuzzer client is going to evolve into something complicated.

DanielRosenwasser:
> another problem that was on my mind is that you want a binary that was compiled without optimizations, otherwise it's gonna be a bad time debugging it

Luckily, there is no optimization level - just differing build flags (e.g. `nodebug`).

DanielRosenwasser:
Thinking about this some more, I think it's okay to have two replay scripts - one in the repo for ease of iteration, and one published on npm that's just JS/TS.

jakebailey:
> Luckily, there is no optimization level - just differing build flags (e.g. `nodebug`).

There is, if you do `hereby build:debug` or run the test through a launch task.

DanielRosenwasser:
I think we are saying the same thing - there's no `-O3` vs `-O0`, but there are times where the build differs based on race-detection, assertion checks, etc., right?

jakebailey:
The default build Go does is effectively `-O3`. Launching with delve (or, building with `build:debug`) is effectively `-O0 -g`.

jakebailey:
This doesn't really have to be a test, though, since you can use a `go run` task (like we have for `tsgo`) that takes args.

gabritto:
> This doesn't really have to be a test, though, since you can use a `go run` task (like we have for `tsgo`) that takes args.

Making it not a test would slightly complicate reuse of the LSP client (which takes in a *testing.T).

jakebailey:
> Making it not a test would slightly complicate reuse of the LSP client (which takes in a *testing.T).

Yep, I wouldn't try and deal with that either. We'd have to introduce a new main package to do this, which is kind of icky, but not impossible within internal or something.

jakebailey (approved):

PR #2834, opened 2026-02-19
Optimize bool to byte conversion in API encoder

auvred:
See:

- https://github.com/golang/go/issues/64825
- https://godbolt.org/z/PvnsxjoTj
- https://github.com/search?q=repo%3Agolang%2Fgo+%22func+bool2int%22&type=code

### DIFF
--- a/internal/api/encoder/encoder.go
+++ b/internal/api/encoder/encoder.go
@@ -843,8 +843,10 @@ func recordExtendedData(node *ast.Node, strs *stringTable, extendedData *[]byte)
 }
 
 func boolToByte(b bool) byte {
+	// https://github.com/golang/go/issues/64825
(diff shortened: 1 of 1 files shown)


jakebailey:
In context, is this actually faster? I'm pretty sure this function is simple enough to be inlined and so its implementation not matter.

auvred:
> In context, is this actually faster? I'm pretty sure this function is simple enough to be inlined and so its implementation not matter.

I'm currently on a laptop with a nearly empty battery, so this may not be the best source of truth, but this branch is faster on my machine.

image

auvred:
Well, it does indeed inline `boolToByte`, producing identical asm https://godbolt.org/z/vhdq7bxWa!

PR #2838, opened 2026-02-19
Emit declaration when expando property on overloaded functions

### LINKED ISSUE #2415: TS2339: Property does not exist on type
## Steps to reproduce

1.`git clone`
2.`yarn`
3.`yarn tsgo -b`
https://github.com/babel/babel/pull/17411

Sorry. I tried to do a minimal reproduction, but it failed.
It works fine when I access the property within the same file, but it throws an error in other packages.

no errors
## Behavior with `tsgo`
```
packages/babel-core/src/transformation/index.ts:112:30 - error TS2339: Property 'visitors' does not exist on type '{ <S>(parent: Node, opts: TraverseOptions<S>, scope: Scope, state: S, parentPath?: NodePath_Final, visitSelf?: boolean): void; (parent: Node, opts: TraverseOptions<...>, scope?: Scope, state?: any, parentPath?: NodePath_Final, visitSelf?: boolean): void; }'.

112     const visitor = traverse.visitors.merge(
```

liuxingbaoyu:
In the TS repository, `shouldEmitFunctionProperties` is called multiple times when traversing each function overload.
It is now only called once when traversing to `fn.a=b`.

Special thanks to https://github.com/microsoft/typescript-go/issues/2415#issuecomment-3679546005 for pointing out the target.

Fixes #2415

### DIFF
--- a/testdata/tests/cases/compiler/declarationEmitExpandoOverloads.ts
+++ b/testdata/tests/cases/compiler/declarationEmitExpandoOverloads.ts
@@ -1,3 +1,4 @@
+// @strict: true
 // @declaration: true
 
 export function A(a: number): void;


internal/transformers/declarations/util.go
```diff
@@ -191,17 +191,15 @@ func isPrivateMethodTypeParameter(host DeclarationEmitHost, node *ast.TypeParame
...
 	}
-
-	overloadSignatures := core.Filter(input.Symbol.Declarations, func(decl *ast.Node) bool {
-		return ast.IsFunctionDeclaration(decl)
+	return !core.Every(input.Symbol.Declarations, func(decl *ast.Node) bool {
+		return !ast.IsFunctionDeclaration(decl) || decl.AsFunctionDeclaration().Body == nil
 	})
-
-	return len(overloadSignatures) == 0 || overloadSignatures[len(overloadSignatures)-1] == input.AsNode()
 }
```
jakebailey, line 203: This was a pure port of:

```ts
    // If the ExpandoFunctionDeclaration have multiple overloads, then we only need to emit properties for the last one.
    function shouldEmitFunctionProperties(input: FunctionDeclaration) {
        if (input.body) {
            return true;
        }

        const overloadSignatures = input.symbol.declarations?.filter(decl => isFunctionDeclaration(decl) && !decl.body);
        return !overloadSignatures || overloadSignatures.indexOf(input) === overloadSignatures.length - 1;
    }
```

What changed such that this is required?

(I assume this is on purpose so multiple functions placed after each other are not duplicated, but that seems like it's probably an error condition)

liuxingbaoyu: https://github.com/microsoft/typescript-go/pull/1399#pullrequestreview-3039375144
We used a different implementation in our tsgo PR.
Previously, we iterated through each `function`, so we needed to skip the declarations and only process the one with a function body.
Now we are iterating through the assignment expressions, so `shouldEmitFunctionProperties` will only be executed once for each property.

PR #2841, opened 2026-02-19
Fix `completionItem/resolve` auto imports crash

### LINKED ISSUE #2827: panic handling request completionItem/resolve: completion list needs auto imports
## Stack trace

```
InternalError: panic handling request completionItem/resolve: completion list needs auto imports
```

## Steps to reproduce

1. I'm getting it pretty consistently when running the fuzzer replay for Redocly/redoc.
The repo is https://github.com/Redocly/redoc commit `d41fd46f7cbee86bf83dc17b7ec51baf54f72a54` and I used #2826 to replay it.
Replay script:
[Redocly.redoc.replay.txt](https://github.com/user-attachments/files/25402743/Redocly.redoc.replay.txt)

@andrewbranch

gabritto:
Fixes #2827 (and similar issue reported [here](https://github.com/microsoft/typescript-go/issues/2829#issuecomment-3924076422)).

The crash was happening because in between the completion and resolve requests, we got a snapshot update due to ATA changes, and so the snapshot used for the resolve request didn't have the auto imports data.
However, it turns out we don't need to collect auto imports for resolve requests, because to resolve completion items that come from auto imports, we use the completion item data instead of symbol information, so that's what this PR does.

### DIFF
--- a/internal/ls/completions.go
+++ b/internal/ls/completions.go
@@ -344,7 +344,7 @@ func (l *LanguageService) getCompletionsAtPosition(
 	}
 
 	preferences := l.UserPreferences()
-	data, err := l.getCompletionData(ctx, checker, file, position, preferences)
+	data, err := l.getCompletionData(ctx, checker, file, position, preferences, false /*forItemResolve*/)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


internal/ls/completions.go
```diff
@@ -4932,7 +4937,7 @@ func (l *LanguageService) getSymbolCompletionFromItemData(
 		}
 	}
 
-	completionData, err := l.getCompletionData(ctx, ch, file, position, &lsutil.UserPreferences{IncludeCompletionsForModuleExports: core.TSTrue, IncludeCompletionsForImportStatements: core.TSTrue})
```
gabritto, line 4935: This and the change below are drive-by simplifications.

PR #2844, opened 2026-02-19
Add nil check in `getAllExportsForSymbol`

ahejlsberg:
Fixes the crash described [here](https://github.com/microsoft/typescript-go/issues/2829#issuecomment-3924076793).

### DIFF
--- a/internal/ls/autoimport/import_adder.go
+++ b/internal/ls/autoimport/import_adder.go
@@ -362,8 +362,10 @@ func (adder *importAdder) getNewImportEntry(moduleSpecifier string, importKind l
 func (adder *importAdder) getAllExportsForSymbol(
 	symbol *ast.Symbol,
 ) []*Export {
-	exportId := SymbolToExport(symbol, adder.checker).ExportID
-	return adder.view.SearchByExportID(exportId)
+	if export := SymbolToExport(symbol, adder.checker); export != nil {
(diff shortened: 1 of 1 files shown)


ahejlsberg:
I did not extract a stand-alone repro as the code is somewhat involved. Another fuzzer run should demonstrate that the issue has been fixed.

internal/ls/autoimport/import_adder.go
```diff
@@ -362,8 +362,10 @@ func (adder *importAdder) getNewImportEntry(moduleSpecifier string, importKind l
 func (adder *importAdder) getAllExportsForSymbol(
 	symbol *ast.Symbol,
 ) []*Export {
-	exportId := SymbolToExport(symbol, adder.checker).ExportID
-	return adder.view.SearchByExportID(exportId)
+	if export := SymbolToExport(symbol, adder.checker); export != nil {
```
gabritto, line 365: I'm curious why this ever returns `nil`. IIRC we're only supposed to be calling `getAllExportsForSymbol` for symbols we have ourselves produced when emitting import type nodes, so this crash could indicate we're missing some symbol in the auto-imports view.

ahejlsberg: That I don't know. But it was obvious that `SymbolToExport` has multiple code paths that might return `nil`.

PR #2853, opened 2026-02-20
Implement ES2017 (async) transforms

jakebailey:
This is just one transform. It's pretty much a pure port, with some differences:

- `arguments` capturing is not done via `NodeCheck` flags. While declared, the checker does not currently set them, and it's possible to get this info without requiring checking. I think this is a net improvement.
- `super` detection is similarly changed, but also handles rewriting directly rather than relying on emit hooks.
- Since we do not support ES5, we unconditionally use the real `Promise` constructor.

All diffs related to this get deleted (yay), except the ones that are still missing the ES2018 transform for async-for.

### DIFF
--- a/internal/transformers/estransforms/async.go
+++ b/internal/transformers/estransforms/async.go
@@ -27,7 +27,7 @@ type asyncTransformer struct {
 	contextFlags asyncContextFlags
 
 	enclosingFunctionParameterNames *collections.Set[string]
-	capturedSuperProperties         *collections.Set[string]
+	capturedSuperProperties         *collections.OrderedSet[string]
 	hasSuperElementAccess           bool
 	hasSuperPropertyAssignment      bool
 	superBinding                    *ast.IdentifierNode
@@ -650,7 +650,7 @@ func (tx *asyncTransformer) transformMethodBody(node *ast.Node) *ast.Node {
 	savedHasSuperPropertyAssignment := tx.hasSuperPropertyAssignment
 	savedSuperBinding := tx.superBinding
 	savedSuperIndexBinding := tx.superIndexBinding
-	tx.capturedSuperProperties = &collections.Set[string]{}
+	tx.capturedSuperProperties = &collections.OrderedSet[string]{}
 	tx.hasSuperElementAccess = false
 	tx.hasSuperPropertyAssignment = false
 	tx.superBinding = tx.Factory().NewUniqueNameEx("_super", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
@@ -660,11 +660,11 @@ func (tx *asyncTransformer) transformMethodBody(node *ast.Node) *ast.Node {
 	updated := tx.EmitContext().VisitFunctionBody(node.Body(), tx.Visitor())
 
 	// Minor optimization, emit `_super` helper to capture `super` access in an arrow.
-	emitSuperHelpers := (tx.capturedSuperProperties.Len() > 0 || tx.hasSuperElementAccess) &&
+	emitSuperHelpers := (tx.capturedSuperProperties.Size() > 0 || tx.hasSuperElementAccess) &&
 		(ast.GetFunctionFlags(tx.getOriginalIfFunctionLike(node))&ast.FunctionFlagsAsyncGenerator) != ast.FunctionFlagsAsyncGenerator
 
 	if emitSuperHelpers {
-		if tx.capturedSuperProperties.Len() > 0 {
+		if tx.capturedSuperProperties.Size() > 0 {
 			tx.EmitContext().AddInitializationStatement(tx.createSuperAccessVariableStatement())
 		}
 	}
@@ -806,7 +806,7 @@ func (tx *asyncTransformer) transformAsyncFunctionBody(node *ast.Node, outerPara
 	savedSuperBinding := tx.superBinding
 	savedSuperIndexBinding := tx.superIndexBinding
 	if !isArrow {
-		tx.capturedSuperProperties = &collections.Set[string]{}
+		tx.capturedSuperProperties = &collections.OrderedSet[string]{}
 		tx.hasSuperElementAccess = false
 		tx.hasSuperPropertyAssignment = false
 		tx.superBinding = tx.Factory().NewUniqueNameEx("_super", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
@@ -823,7 +823,7 @@ func (tx *asyncTransformer) transformAsyncFunctionBody(node *ast.Node, outerPara
 
 	// Substitute super property accesses with _super/_superIndex helpers
 	emitSuperHelpers := tx.capturedSuperProperties != nil &&
-		(tx.capturedSuperProperties.Len() > 0 || tx.hasSuperElementAccess)
+		(tx.capturedSuperProperties.Size() > 0 || tx.hasSuperElementAccess)
 	if emitSuperHelpers {
 		asyncBody = tx.substituteSuperAccessesInBody(asyncBody)
 	}
@@ -834,7 +834,7 @@ func (tx *asyncTransformer) transformAsyncFunctionBody(node *ast.Node, outerPara
 
 		// Minor optimization, emit `_super` helper to capture `super` access in an arrow.
 		if emitSuperHelpers {
-			if tx.capturedSuperProperties.Len() > 0 {
+			if tx.capturedSuperProperties.Size() > 0 {
 				tx.EmitContext().AddInitializationStatement(tx.createSuperAccessVariableStatement())
 			}
 		}
@@ -1037,13 +1037,7 @@ func (tx *asyncTransformer) createSuperAccessVariableStatement() *ast.Node {
 	f := tx.Factory()
 	var accessors []*ast.Node
 
-	var sortedNames []string
-	for name := range tx.capturedSuperProperties.Keys() {
-		sortedNames = append(sortedNames, name)
-	}
-	slices.Sort(sortedNames)
-
-	for _, name := range sortedNames {
+	for name := range tx.capturedSuperProperties.Values() {
 		var descriptorProperties []*ast.Node
 
 		// getter: get: () => super.name


internal/checker/emitresolver.go
```diff
@@ -51,6 +51,26 @@ func newEmitResolver(checker *Checker) *EmitResolver {
 	return e
 }
 
+func (r *EmitResolver) IsArgumentsLocalBinding(nodeIn *ast.IdentifierNode) bool {
```
jakebailey: This function is the replacement for the old `NodeCheckFlags`-based analysis.

internal/printer/factory.go
```diff
@@ -650,6 +664,64 @@ func (f *NodeFactory) NewRestHelper(value *ast.Expression, elements []*ast.Node,
...
+	}
+
+	return f.NewCallExpression(
+		f.NewUnscopedHelperName("__awaiter"),
+		nil, /*questionDotToken*/
+		nil, /*typeArguments*/
+		f.NewNodeList([]*ast.Expression{
+			thisArg,
+			argsArg,
+			f.NewVoidZeroExpression(),
```
jakebailey, line 722: `Promise` is now required, no ES5.

testdata/baselines/reference/submodule/compiler/sourceMapValidationImport.sourcemap.txt.diff
```diff
@@ -133,75 +133,69 @@
...
 -5 >        ^
 -6 >         ^
 -7 >          ^
 -8 >           ^
 -9 >            ^^^^^->
-+5 >        ^^^
-+6 >           ^
-+7 >            ^^^^^->
++4 >     ^^^^^^^
++5 >            ^^^^^->
```
jakebailey: This might be a regression.

internal/checker/emitresolver.go
weswigham: You can only reassign/redeclare `arguments` outside strict mode, right? In strict mode it's a uniform `Uncaught SyntaxError: Unexpected eval or arguments in strict mode` to try and use `arguments` as a declaration or reassignment target. Given we always assume and enforce js strict mode now, doesn't this not need checker lookups at all? If the identifier is `arguments`, it's an arguments local binding, no?

jakebailey: I guess that's the case, yeah:

Image

A shame, I was so happy with having all deleted diffs.

jakebailey: Oh, it's just the lookup that changes.

jakebailey: Done.

internal/printer/emitflags.go
```diff
@@ -27,6 +27,7 @@ const (
 	EFNeverApplyImportHelper                          // Do not apply an import helper to this node
 	EFStartOnNewLine                                  // Start this node on a new line
 	EFIndirectCall                                    // Emit CallExpression as an indirect call: `(0, f)()`
+	EFAsyncFunctionBody                               // The node was originally an async function body.
```
jakebailey, line 30: This is actually leftover, and does nothing, I'll remove it.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1299 @@
...
+	savedCapturedSuperProperties := tx.capturedSuperProperties
+	savedHasSuperElementAccess := tx.hasSuperElementAccess
+	savedHasSuperPropertyAssignment := tx.hasSuperPropertyAssignment
+	savedSuperBinding := tx.superBinding
+	savedSuperIndexBinding := tx.superIndexBinding
+	tx.capturedSuperProperties = &collections.Set[string]{}
+	tx.hasSuperElementAccess = false
+	tx.hasSuperPropertyAssignment = false
+	tx.superBinding = tx.Factory().NewUniqueNameEx("_super", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
+	tx.superIndexBinding = tx.Factory().NewUniqueNameEx("_superIndex", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
```
DanielRosenwasser, line 657: I know this is just a port, but why is `_superIndex` necessary in addition to just `_super`?

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1299 @@
...
+
+func (tx *asyncTransformer) inTopLevelContext() bool {
+	return !tx.inContext(asyncContextNonTopLevel)
+}
+
+func (tx *asyncTransformer) inHasLexicalThisContext() bool {
+	return tx.inContext(asyncContextHasLexicalThis)
+}
+
+func (tx *asyncTransformer) doWithContext(flags asyncContextFlags, cb func(*ast.Node) *ast.Node, node *ast.Node) *ast.Node {
```
DanielRosenwasser: `doWithContext` takes a `Node -> Node`, so the way you're currently calling it, it's auto-binding against the `asyncTransformer`, right? Is stuff like that definitely inlined?

If not, I think the pattern here is for methods like this to take `(asyncTransformer, Node) -> Node` to avoid allocations and just pass in the same transformer context.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1299 @@
...
+			if tx.capturedSuperProperties.Len() > 0 {
+				superHelpers = append(superHelpers, tx.createSuperAccessVariableStatement())
+			}
+			prologue, rest := tx.Factory().SplitStandardPrologue(statements)
+			statements = append(prologue, append(superHelpers, rest...)...)
+		}
+
+		if captureLexicalArguments && tx.lexicalArguments.used {
+			prologue, rest := tx.Factory().SplitStandardPrologue(statements)
+			statements = append(prologue, append([]*ast.Node{tx.createCaptureArgumentsStatement()}, rest...)...)
```
DanielRosenwasser: Why is this not just


```suggestion
			statements = append(prologue, tx.createCaptureArgumentsStatement(), rest...)
```

?

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1299 @@
...
+		if emitSuperHelpers {
+			var superHelpers []*ast.Node
+			if tx.hasSuperElementAccess {
+				superHelpers = append(superHelpers, tx.createSuperIndexVariableStatement())
+			}
+			if tx.capturedSuperProperties.Len() > 0 {
+				superHelpers = append(superHelpers, tx.createSuperAccessVariableStatement())
+			}
+			prologue, rest := tx.Factory().SplitStandardPrologue(statements)
+			statements = append(prologue, append(superHelpers, rest...)...)
```
DanielRosenwasser: Can this be

```suggestion
			statements = append(prologue, superHelpers..., rest...)
```

?

internal/transformers/estransforms/async.go
DanielRosenwasser: I guess see

* https://github.com/microsoft/TypeScript/issues/21088
* https://github.com/microsoft/TypeScript/pull/26707

internal/printer/emitflags.go
jakebailey: Actually, I think this is still needed for class fields?

internal/transformers/estransforms/async.go
jakebailey: Go only allows a single spread. But this should be two appends, not nested.

jakebailey: And, not mutate `prologue`.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1305 @@
...
+	case ast.KindFunctionDeclaration:
+		return tx.doWithContext(asyncContextNonTopLevel|asyncContextHasLexicalThis, tx.visitFunctionDeclaration, node)
+	case ast.KindFunctionExpression:
+		return tx.doWithContext(asyncContextNonTopLevel|asyncContextHasLexicalThis, tx.visitFunctionExpression, node)
+	case ast.KindArrowFunction:
+		return tx.doWithContext(asyncContextNonTopLevel, tx.visitArrowFunction, node)
+	case ast.KindPropertyAccessExpression:
+		if tx.capturedSuperProperties != nil && node.Expression().Kind == ast.KindSuperKeyword {
+			tx.capturedSuperProperties.Add(node.Name().Text())
+			if ast.IsAssignmentTarget(node) {
```
jakebailey: IsAssignmentTarget touches parents, so this is probably unsafe :\

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+	functionFlags := getFunctionFlags(node)
+
+	var parameters *ast.NodeList
+	var body *ast.Node
+	if functionFlags&checker.FunctionFlagsAsync != 0 {
+		parameters = tx.transformAsyncFunctionParameterList(node)
+		body = tx.transformAsyncFunctionBody(node, parameters)
+	} else {
+		parameters = tx.Visitor().VisitNodes(decl.Parameters)
+		body = tx.Visitor().VisitNode(decl.Body)
```
weswigham: In strada, this is `visitFunctionBody` instead of `visitNode` so temporary variables from nested transformations get correctly added to the scope; shouldn't the same need to happen here? AFAIK `VisitNode` doesn't know if the body block/expression is a function body. It's used in `runtimesyntax.go`, so the functionality's already in - it should be `tx.EmitContext().VisitFunctionBody(decl.Body, tx.Visitor())`. The same goes for all the other function-like body visits above.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+	savedHasSuperPropertyAssignment := tx.hasSuperPropertyAssignment
+	savedSuperBinding := tx.superBinding
+	savedSuperIndexBinding := tx.superIndexBinding
+	tx.capturedSuperProperties = &collections.Set[string]{}
+	tx.hasSuperElementAccess = false
+	tx.hasSuperPropertyAssignment = false
+	tx.superBinding = tx.Factory().NewUniqueNameEx("_super", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
+	tx.superIndexBinding = tx.Factory().NewUniqueNameEx("_superIndex", printer.AutoGenerateOptions{Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel})
+
+	updated := tx.Visitor().VisitNode(node.Body())
```
weswigham: Yeah, definitely see my comment above on visiting function bodies - since this transform adds local helpers, the  specific call to merge helpers into the scope seems needed. In strada this is a `visitFunctionBody` call.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+		block := updated.AsBlock()
+		statements := append([]*ast.Node{}, block.Statements.Nodes...)
+		var helpers []*ast.Node
+		multiLine := block.Multiline
+		if tx.hasSuperElementAccess {
+			helpers = append(helpers, tx.createSuperIndexVariableStatement())
+			multiLine = true
+		}
+		if tx.capturedSuperProperties.Len() > 0 {
+			helpers = append(helpers, tx.createSuperAccessVariableStatement())
```
weswigham: Hm, why are these helpers being manually injected here and not getting added with the scope-merging infra that already exists on the emit context? Is it a workaround for not using the function body visit mentioned above?

internal/transformers/estransforms/async.go
jakebailey: It's interesting that nothing broke...

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+	}
+	newParametersArray := tx.Factory().NewNodeList(newParameters)
+	newParametersArray.Loc = node.ParameterList().Loc
+	return newParametersArray
+}
+
+func (tx *asyncTransformer) transformAsyncFunctionBody(node *ast.Node, outerParameters *ast.NodeList) *ast.Node {
+	innerParameters := (*ast.NodeList)(nil)
+	if !isSimpleParameterList(node.Parameters()) {
+		innerParameters = tx.Visitor().VisitNodes(node.ParameterList())
```
weswigham: Hm, unlike the other function, I think the functionality provided by `visitParameterList` in strada matters here, since these are non-simple parameters that can have transformed initializers. Not fully sure, but I think the local nature of the helpers means you can generate new hoisted names within parameter initializers in this transform still.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+			tx.Factory().NewAwaiterHelper(
+				hasLexicalThis,
+				argumentsExpression,
+				innerParameters,
+				asyncBody,
+			),
+		))
+
+		// Minor optimization, emit `_super` helper to capture `super` access in an arrow.
+		// This step isn't needed if we eventually transform this to ES5.
```
weswigham: Can probably remove this line of the comment, since we don't transform to es5.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1368 @@
...
+	return f.NewVariableStatement(nil, declList)
+}
+
+// createAdvancedSuperIndexVariableStatement creates:
+//
+//	const _superIndex = (function (geti, seti) {
+//	    const cache = Object.create(null);
+//	    return name => cache[name] || (cache[name] = { get value() { return geti(name); }, set value(v) { seti(name, v); } });
+//	})(name => super[name], (name, value) => super[name] = value);
+func (tx *asyncTransformer) createAdvancedSuperIndexVariableStatement() *ast.Node {
```
weswigham: Is this... recreating the helper injection logic? This looks a lot like manual node building to produce an AST for the `_superIndex` helper which, IIRC, we don't historically need to build an AST for because it's just left in the tree as an unparsed helper and emitted verbatim.

jakebailey: Yeah, it totally is, I totally missed this as it was a follow-up bugfix, given this used to be a substitution thingy. Should have noticed after letting copilot futz with this before I pushed it

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1358 @@
...
+	return true
+}
+
+// isArgumentsIdentifier checks whether an identifier refers to the `arguments` object.
+// Since we always assume strict mode, a simple name check suffices.
+func isArgumentsIdentifier(node *ast.Node) bool {
+	if node.Text() != "arguments" {
+		return false
+	}
+	parent := node.Parent
```
weswigham: ...sus (given transforms don't set parent by default). Especially so since _this transform_ adds synthetic `arguments` identifiers to the tree. Might wanna pass in the `parent` node as an argument instead, and track it as the transform descends through the tree. Other transforms do this.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1358 @@
...
+		p := param.AsParameterDeclaration()
+		if p.Initializer != nil || !ast.IsIdentifier(p.Name()) {
+			return false
+		}
+	}
+	return true
+}
+
+// getFunctionFlags returns the function flags for a node.
+func getFunctionFlags(node *ast.Node) checker.FunctionFlags {
```
weswigham: This is a dupe of `ast.GetFunctionFlags`???

weswigham: Wait, that's only in my ID PR. Hah.

jakebailey: Yeah, I need to copy it over (how close is that PR? 😄 )

weswigham: Anyways, I extracted this to the ast package in that PR, you probably should do the same here - reaching into `checker` in the transform for a type aughta probably be avoided.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1358 @@
...
+		return true
+	}
+	return false
 }
 
 func newAsyncTransformer(opts *transformers.TransformOptions) *transformers.Transformer {
-	tx := &asyncTransformer{}
-	return tx.NewTransformer(tx.visit, opts.Context)
+	tx := &asyncTransformer{
+		compilerOptions: opts.CompilerOptions,
```
weswigham: This is/was only used in the transform for checking if the source file was in strict mode (always true now) and the language version (which is now always es6+, which is what all the checks were for). I don't think it should be needed.

jakebailey: Ah, yes, and I need to inline `isEffectiveStrictModeSourceFile`. I made that change early but then forgot to leave a TODO

internal/transformers/estransforms/async.go
weswigham: It's less that nothing would be broken and more that any tests that exercise it just wouldn't have their `.diff`s deleted/updated yet. You generally don't go adding `class Foo { async method(p = async (x) => super[x]) {} }` to every test, right? (If that's actually tested at all and not, just, like, quietly handled correctly)

jakebailey: I'm not even sure we have tests for it; I didn't notice this personally (as I am not super familiar). I would have expected _something_ to have changed, as every test whose baselines were touched were pure deletions by the time I sent the PR...

jakebailey: And honestly there's a lot of visits where I am now not sure what must be done, except via cross comparison with the original code.

internal/transformers/estransforms/async.go
jakebailey: Yeah. Flipped this to saving the current nodes on `tx`. We certainly do that in runtimesytnax.

internal/printer/helpers.go
```diff
@@ -154,6 +154,41 @@ var restHelper = &EmitHelper{
...
+var AdvancedAsyncSuperHelper = &EmitHelper{
+	Name:   "typescript:advanced-async-super",
+	Scoped: true,
+	TextCallback: func(makeUniqueName func(string) string) string {
+		return "\nconst " + makeUniqueName("_superIndex") + " = (function (geti, seti) {\n" +
+			"    const cache = Object.create(null);\n" +
+			"    return name => cache[name] || (cache[name] = { get value() { return geti(name); }, set value(v) { seti(name, v); } });\n" +
+			"})(name => super[name], (name, value) => super[name] = value);"
+	},
+}
```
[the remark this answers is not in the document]
jakebailey: This is true, but the old code directly added them too, as the `ReadEmitHelpers` thing does't work on scoped helpers.

internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -678,8 +678,8 @@ func (tx *RuntimeSyntaxTransformer) visitImportEqualsDeclaration(node *ast.Impor
 		return tx.Visitor().VisitEachChild(node.AsNode())
 	}
 
-	moduleReference := convertEntityNameToExpression(tx.EmitContext(), node.ModuleReference)
-	tx.EmitContext().SetEmitFlags(moduleReference, printer.EFNoComments|printer.EFNoNestedSourceMaps)
+	moduleReference := tx.Factory().CreateExpressionFromEntityName(node.ModuleReference)
+	tx.EmitContext().SetEmitFlags(moduleReference, printer.EFNoComments|printer.EFNoNestedComments)
```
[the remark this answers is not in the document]
jakebailey: This is a bugfix for sourceMapValidationImport

internal/ast/ast.go
```diff
@@ -2670,6 +2670,13 @@ func (node *Token) computeSubtreeFacts() SubtreeFacts {
 		return SubtreeContainsTypeScript
 	case KindAccessorKeyword:
 		return SubtreeContainsClassFields
+	case KindAsyncKeyword:
+		// 'async' modifier is ES2017 (async functions) or ES2018 (async generators)
+		return SubtreeContainsAnyAwait
```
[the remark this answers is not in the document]
jakebailey: `AsyncKeyword` needs:

```ts
case SyntaxKind.AsyncKeyword:
    // 'async' modifier is ES2017 (async functions) or ES2018 (async generators)
    transformFlags = TransformFlags.ContainsES2017 |
        TransformFlags.ContainsES2018;
    break;
```

Which we've since split. Probably needs to be `SubtreeContainsAnyAwait | SubtreeContainsForAwaitOrAsyncGenerator`...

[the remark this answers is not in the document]
jakebailey: Ah, no, we don't need this, since `SubtreeContainsForAwaitOrAsyncGenerator` is not touched by es2017. Yay fine-grained

internal/printer/printer.go
```diff
@@ -176,6 +176,9 @@ func NewPrinter(options PrinterOptions, handlers PrintHandlers, emitContext *Emi
 	printer.nameGenerator.Context = printer.emitContext
 	printer.nameGenerator.GetTextOfNode = func(node *ast.Node) string { return printer.getTextOfNode(node, false) }
 	printer.nameGenerator.IsFileLevelUniqueNameInCurrentFile = printer.isFileLevelUniqueNameInCurrentFile
+	printer.makeFileLevelOptimisticUniqueName = func(name string) string {
+		return printer.nameGenerator.MakeFileLevelOptimisticUniqueName(name)
+	}
```
DanielRosenwasser, line 181: Why is this not just 

```suggestion
	printer.makeFileLevelOptimisticUniqueName = printer.nameGenerator.MakeFileLevelOptimisticUniqueName
```

?

But moreover, why is it not just a method on `printer`?

jakebailey: It's used like:

```go
p.writeLines(helper.TextCallback(p.makeFileLevelOptimisticUniqueName))
```

So will be rebound to `p` each time. Perhaps that's not such a big deal.

internal/printer/printer.go
```diff
@@ -2715,6 +2725,7 @@ func (p *Printer) getBinaryExpressionPrecedence(node *ast.BinaryExpression) (lef
 	case ast.OperatorPrecedenceAssignment:
 		// assignment is right-associative
 		leftPrec = ast.OperatorPrecedenceLeftHandSide
+		rightPrec = ast.OperatorPrecedenceYield
```
weswigham, line 2728: I gotta ask: what is this fixing? Doesn't directly seem related to async/await to adjust the operator precedence of assignment operators, but... eh? This, afaik, doesn't exist in strada (we handle precedence differently in corsa), so there's no point of comparison.

internal/printer/factory.go
```diff
@@ -280,6 +280,22 @@ func (f *NodeFactory) InlineExpressions(expressions []*ast.Expression) *ast.Expr
 // Utilities
 //
 
+func (f *NodeFactory) CreateExpressionFromEntityName(node *ast.Node) *ast.Expression {
+	if ast.IsQualifiedName(node) {
+		left := f.CreateExpressionFromEntityName(node.AsQualifiedName().Left)
+		right := node.AsQualifiedName().Right.Clone(f.AsNodeFactory())
+		right.Loc = node.AsQualifiedName().Right.Loc
+		right.Parent = node.AsQualifiedName().Right.Parent
```
weswigham, line 289: This parent setting was TODO'd with a "does this need to be done" before it was moved. My question: _does it_? Most AST construction in the transformers doesn't need `.Parent` pointers setup, and setting them in that context is... odd.

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1213 @@
...
+	return tx.argumentsAndSuperNodeVisitor.VisitEachChild(node)
+}
+
+// visitArgumentsAndSuper is the NodeVisitor callback for argumentsAndSuperNodeVisitor.
+// It wraps argumentsAndSuperVisitor with parent tracking for recursive calls.
+func (tx *asyncTransformer) visitArgumentsAndSuper(node *ast.Node) *ast.Node {
+	savedParent := tx.parentNode
+	tx.parentNode = tx.currentNode
+	tx.currentNode = node
+	defer func() { tx.currentNode = tx.parentNode; tx.parentNode = savedParent }()
```
weswigham: 1. I have never seen `defer` used with an IIFE like this before - cute.
2. This logic is repeated a few times - shouldn't this be in a `func (tx *asyncTransformer) descendInto(node *ast.Node) func()` we use instead, so we can do
```go
cleanup := tx.descendInto(node)
defer cleanup()
```
instead of duplicating the logic+closure?

internal/transformers/estransforms/async.go
```diff
@@ -1,19 +1,1213 @@
...
+//	});
+func (tx *asyncTransformer) createSuperAccessVariableStatement() *ast.Node {
+	f := tx.Factory()
+	var accessors []*ast.Node
+
+	var sortedNames []string
+	for name := range tx.capturedSuperProperties.Keys() {
+		sortedNames = append(sortedNames, name)
+	}
+	slices.Sort(sortedNames)
```
weswigham: I see a series of `append` calls followed by a `Sort` call and my instincts are telling me that we should be using an `insertSorted` helper.... but it feels bad that this needs sorting at all. Presumably, this sort is alphanumeric, meanwhile in strada this was topographical (map insertion order). Shouldn't we be using an insertion-order-sorted collection instead, then?

internal/printer/printer.go
jakebailey: My process for this PR was to bulk transform, accept baselines, then look at all files modified and what diffs remained. This is a bit of a drive-by, but it's in [`2b0c69a` (this PR)](https://github.com/microsoft/typescript-go/pull/2853/commits/2b0c69ac5c70cc2639a03606f31d7340adddbdd5).

internal/transformers/estransforms/async.go
jakebailey: Ah, I do this sort of thing all the time. I can extract it though, yeah.

internal/printer/factory.go
jakebailey: I'm going to say no. The original Strada code was flipped the other way with the opposite TODO:

```ts
const left = createExpressionFromEntityName(factory, node.left);
        // TODO(rbuckton): Does this need to be parented?
        const right = setParent(setTextRange(factory.cloneNode(node.right), node.right), node.right.parent);
        return setTextRange(factory.createPropertyAccessExpression(left, right), node);
```

I'll delete it here once and for all.

jakebailey: Er, sorry, not flipped, it also set parent.

internal/printer/printer.go
weswigham: So a driveby fix that removed parens from emitted `yield`s, got it.

internal/printer/factory.go
jakebailey: I just tested, and while this first one works when unparented, the one below (plain clone) does not and causes:

```diff
--- old.unicodeEscapesInJsxtags.js
+++ new.unicodeEscapesInJsxtags.js
@@= skipped -38, +38 lines =@@
 React.createElement("a", null);
 React.createElement("a-b", null);
 React.createElement("a-c", null);
-React.createElement(Comp\u0061, { x: 12 });
+React.createElement(Compa, { x: 12 });
 React.createElement(x.\u0076ideo, null);
 React.createElement("a", null);
 React.createElement("a-b", null);
 React.createElement("a-c", null);
-React.createElement(Comp\u{0061}, { x: 12 });
+React.createElement(Compa, { x: 12 });
 React.createElement("video", { "data-video": true });
 React.createElement("video", { \u0073rc: "" });
```

internal/transformers/estransforms/async.go
jakebailey: Flipped to an `OrderedSet`. Nothing changes, which is.. funny

internal/transformers/estransforms/async.go
jakebailey: Done

internal/printer/factory.go
jakebailey: I restored the comments for now.

weswigham: Hm, the parent pointers being set causes it to go back to the source file for text vs using the parsed text stored in the identifier. Kinda just highlights information we're not properly storing in the AST.

weswigham (approved):

internal/printer/factory.go
jakebailey: Yeah. This to me means something is wrong, but there's not a great way to ban all parent accesses during transform...

Issue #2780, opened 2026-02-13
High memory consumption in a large pnpm monorepo


shinichy:
### Extension Version

0.20260212.1

### VS Code Version

1.108.1

### Operating system Version

MacOS 15.7.3

### Steps to reproduce

1. Open a TypeScript file within a large pnpm monorepo (observed baseline memory: ~700MB).
2. Perform a minor edit (e.g., adding a character).
3. Observe memory usage skyrocketing to ~35GB, leading to system instability or OOM.

### Issue

After bisecting, I've identified that this memory regression was introduced by PR #2502.

[99701-cpuprofile.pb.gz](https://github.com/user-attachments/files/25281683/99701-cpuprofile.pb.gz)
[99701-memprofile.pb.gz](https://github.com/user-attachments/files/25281682/99701-memprofile.pb.gz)

neo773:
Can reproduce on our NestJS monorepo also
It skyrockets to ~25-30GB after some navigating
Repo: https://github.com/twentyhq/twenty/

Moltenship:
Experiencing the same thing. [Here](https://github.com/microsoft/typescript-go/issues/2239#issuecomment-3782854830) is the memory profile.

Also, it may be [related](https://github.com/microsoft/typescript-go/issues/2239#issuecomment-3811376292) to version 0.20251122.1 (previous versions works fine, aside from memory leaks when deleting files or opening a `js` file)

andrewbranch:
I'm not able to reproduce this. I tried in https://github.com/twentyhq/twenty. First I navigated around and made small edits myself, and when that didn't blow things up, I had Copilot have a field day doing whatever it wanted. Note that agent mode chat opens files via the LSP and leaves them open, causing lots of projects to stay open. By the end of its edits, it had "opened" 50 files (not in tabs, but visible in our logs) across 5 projects, summing to 25k source files and over 100k exports stored in the auto-import cache. I managed to get the process to report up to 9.8 GB this way, but when a took a heap snapshot, GC ran and immediately dropped it to 8 GB. However, the actual snapshot contents showed only about 3.5 GB of in-use space, which is very close to what the process used with those 50 files and 100k auto-imports open after restarting the server. Go itself is not very good at returning unused memory to the OS; if memory spikes during a really expensive operation like find-all-references, you might not see that memory return, even if we've cleaned up internally, especially if the process isn't under memory pressure.

If someone has a repro for this, please take the following steps:

[...]

PR #2855, opened 2026-02-20
Share package.json cache and skip collecting failed lookup locations in auto-imports

andrewbranch:
#2780 shows a lot of allocations going to package.json parsing. I can't repro myself, but based on the profile and the report that the problem started after #2502, I think this is probably it. The failed lookup locations are also not used by auto-imports, so that should be an additional small win. I tried invoking completions in a file in Signal-Desktop as a random benchmark, and this brought total allocs down from 0.78 GB to 0.75 GB. Not huge, but maybe there's a pathological case I haven't been able to repro. Worth a try!


andrewbranch:
This was included in #2991

PR #2861, opened 2026-02-20
Implement for-await (ES2018) transform, fix existing ES2018 issues

jakebailey:
On top of #2853.

Basically the same caveats apply.

Also fixes other diffs that I noticed once the for-await was done.

### DIFF
--- a/internal/printer/factory.go
+++ b/internal/printer/factory.go
@@ -298,6 +298,21 @@ func (f *NodeFactory) CreateExpressionFromEntityName(node *ast.Node) *ast.Expres
 	return res
 }
 
+func (f *NodeFactory) RestoreEnclosingLabel(node *ast.Node, outermostLabeledStatement *ast.LabeledStatement) *ast.Node {
+	if outermostLabeledStatement == nil {
+		return node
+	}
+	innerLabel := node
+	if ast.IsLabeledStatement(outermostLabeledStatement.Statement) {
+		innerLabel = f.RestoreEnclosingLabel(node, outermostLabeledStatement.Statement.AsLabeledStatement())
+	}
+	return f.UpdateLabeledStatement(
+		outermostLabeledStatement,
+		outermostLabeledStatement.Label,
+		innerLabel,
+	)
+}
+
 // CreateForOfBindingStatement creates a statement to bind the iteration value.
 func (f *NodeFactory) CreateForOfBindingStatement(node *ast.Node, boundValue *ast.Node) *ast.Node {
 	if ast.IsVariableDeclarationList(node) {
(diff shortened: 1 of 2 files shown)


testdata/baselines/reference/submodule/conformance/asyncMethodWithSuper_es6.js.diff
```diff
@@ -1,59 +1,18 @@
...
-         });
-     }
--    element_access_only_write_only_in_generator() {
-+    *element_access_only_write_only_in_generator() {
-         const _superIndex = (function (geti, seti) {
+@@= skipped -17, +16 lines =@@
              const cache = Object.create(null);
              return name => cache[name] || (cache[name] = { get value() { return geti(name); }, set value(v) { seti(name, v); } });
          })(name => super[name], (name, value) => super[name] = value);
 -        const _super = Object.create(null, {});
```
jakebailey, line 15: This variable is actually unreferenced, so this is in a way an improvement.

internal/ast/ast.go
```diff
@@ -3294,7 +3294,8 @@ func (node *ReturnStatement) Clone(f NodeFactoryCoercible) *Node {
 }
 
 func (node *ReturnStatement) computeSubtreeFacts() SubtreeFacts {
-	return propagateSubtreeFacts(node.Expression)
+	// return in an ES2018 async generator must be awaited
+	return propagateSubtreeFacts(node.Expression) | SubtreeContainsForAwaitOrAsyncGenerator
```
jakebailey, line 3304: Shockingly:

```ts
    // @api
    function createReturnStatement(expression?: Expression): ReturnStatement {
        const node = createBaseNode<ReturnStatement>(SyntaxKind.ReturnStatement);
        node.expression = expression;
        // return in an ES2018 async generator must be awaited
        node.transformFlags |= propagateChildFlags(node.expression) |
            TransformFlags.ContainsES2018 |
            TransformFlags.ContainsHoistedDeclarationOrCompletion;

        node.jsDoc = undefined; // initialized by parser (JsDocContainer)
        node.flowNode = undefined; // initialized by binder (FlowContainer)
        return node;
    }
```

testdata/baselines/reference/submodule/conformance/propertyAccessChain.3.js.diff
jakebailey: not new, but this is a bad emit bug itself somewhere

weswigham (approved):
I haven't gone line by line to find stuff that differs here like I did the other PR, but also nothing seemed egregiously weird when I read through it so as to prompt me to do that. I'm unsure if we should have the `async function*() {}` transform logic in the same transform as `for await` like we have here. I know I only had the `forawait` transform stubbed out, but, since one doesn't strictly require the other exist in a given AST and we could pretty easily skip some traversal time if they're separated, maybe. Not too important, since you can argue they're two sides of the same feature anyway.

Unless the merge conflict fixes change a bunch, this looks OK.

PR #2864, opened 2026-02-20
Enable `noUncheckedSideEffectImports` by default

Andarist:
ports https://github.com/microsoft/TypeScript/pull/62443


internal/checker/checker.go
jakebailey, line 1: also update internal/tsoptions/declscompiler.go

Andarist: done

internal/tsoptions/declscompiler.go
```diff
@@ -878,7 +878,7 @@ var optionsForCompiler = []*CommandLineOption{
 		AffectsBuildInfo:           true,
 		Category:                   diagnostics.Modules,
 		Description:                diagnostics.Check_side_effect_imports,
-		DefaultValueDescription:    false,
+		DefaultValueDescription:    true,
```
jakebailey, line 881: Updating this will also require help message updates in other baselines.

Andarist: rerun everything now with this change

jakebailey (approved):

Meligy:
This breaks .scss imports even when you have fake modules for them.

jakebailey:
Please file an issue with a repro. That should not be the case https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#the---nouncheckedsideeffectimports-option

pedro757:
I can confirm this breaks css imports

Andarist:
Please open a new issue with the repro steps included. Ideally, one that would showcase the difference between `typescript@next` and tsgo

pedro757:
@Andarist @jakebailey I filed the issue with a repro sandbox #2912

PR #2878, opened 2026-02-23
OrganizeImports unit tests port

johnfav03:
Ported a majority of the unit tests for OrganizeImports from Typescript to tsgo


jakebailey:
Are these doable as fourslash tests? They were unit tests originally I know, but, it seems out of place.

PR #2872, opened 2026-02-22
Speed up source map emit with printer-local monotomic cache

jakebailey:
Fixes #2402

I still think there's work to be done here (note the TODO), but this seems to work; just rely on the fact that the printer always prints in order to avoid rescanning already-seen text.

For the biggest benchmark, it's quite literally a 230x speedup.

```
goos: linux
goarch: amd64
pkg: github.com/microsoft/typescript-go/internal/compiler
cpu: Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz
                               │    old.txt     │               new.txt               │
                               │     sec/op     │   sec/op     vs base                │
EmitLongLines/props_1000-20       19508.0µ ± 2%   929.4µ ± 1%  -95.24% (p=0.000 n=10)
EmitLongLines/props_5000-20       517.771m ± 2%   4.502m ± 2%  -99.13% (p=0.000 n=10)
EmitLongLines/props_10000-20     2079.987m ± 2%   8.940m ± 1%  -99.57% (p=0.000 n=10)
EmitManyFiles-20                   114.76m ± 2%   14.10m ± 4%  -87.71% (p=0.000 n=10)
EmitLongLinesWithLineBreaks-20      12.63m ± 3%   12.04m ± 2%   -4.66% (p=0.000 n=10)
geomean                             124.9m        5.762m       -95.39%
```
```
                               │    old.txt    │               new.txt                │
                               │     B/op      │     B/op      vs base                │
EmitLongLines/props_1000-20       500.5Ki ± 1%   375.3Ki ± 0%  -25.01% (p=0.000 n=10)
EmitLongLines/props_5000-20       6.186Mi ± 1%   1.928Mi ± 0%  -68.84% (p=0.000 n=10)
EmitLongLines/props_10000-20     13.193Mi ± 0%   3.646Mi ± 0%  -72.37% (p=0.000 n=10)
EmitManyFiles-20                  41.93Mi ± 0%   38.83Mi ± 0%   -7.38% (p=0.000 n=10)
EmitLongLinesWithLineBreaks-20    3.926Mi ± 0%   3.916Mi ± 0%   -0.25% (p=0.002 n=10)
geomean                           5.800Mi        3.301Mi       -43.10%
```
```
                               │   old.txt    │               new.txt               │
                               │  allocs/op   │  allocs/op   vs base                │
EmitLongLines/props_1000-20        796.0 ± 2%    419.5 ± 0%  -47.30% (p=0.000 n=10)
[...]

PR #2879, opened 2026-02-23
Hopefully settle line endings / position mapping

### LINKED ISSUE #1578: Fix GetLineAndCharacterOfPosition and audit all users

jakebailey:
Fixes #1578

There's a lot of fuzting here, but in essensce, I've renamed some functions, introduced a new `UTF16Offset` type, and then fixed things. This should achieve the goal in https://github.com/microsoft/typescript-go/pull/2872#discussion_r2841998748 and #1578.

- `PositionToLineAndCharacter` → `PositionToLineAndByteOffset`
- `GetECMALineAndCharacterOfPosition` → `GetECMALineAndUTF16CharacterOfPosition` / `GetECMALineAndByteOffsetOfPosition`
- `GetECMAPositionOfLineAndCharacter` → `GetECMAPositionOfLineAndUTF16Character` / `GetECMAPositionOfLineAndByteOffset`
- `ComputePositionOfLineAndCharacter` → removed (replaced by `ComputePositionOfLineAndByteOffset`)
- `ComputePositionOfLineAndCharacterEx` → `ComputePositionOfLineAndUTF16Character`
- `GetColumn() int` → `GetColumn() core.UTF16Offset`

This fixes a few baselines, and critically some of the sourcemap ones I expected to find were broken.

### DIFF
--- a/internal/diagnosticwriter/diagnosticwriter.go
+++ b/internal/diagnosticwriter/diagnosticwriter.go
@@ -231,7 +231,7 @@ func writeCodeSnippet(writer io.Writer, sourceFile FileLike, start int, length i
 			if i == lastLine {
 				lastCharForLine = int(lastLineChar)
 			} else {
-				lastCharForLine = len(lineContent)
+				lastCharForLine = int(core.UTF16Len(lineContent))
 			}
 
 			// Fill with spaces until the first character,
@@ -243,7 +243,7 @@ func writeCodeSnippet(writer io.Writer, sourceFile FileLike, start int, length i
 			fmt.Fprint(writer, strings.Repeat("~", int(lastLineChar)))
 		default:
 			// Squiggle the entire line.
-			fmt.Fprint(writer, strings.Repeat("~", len(lineContent)))
+			fmt.Fprint(writer, strings.Repeat("~", int(core.UTF16Len(lineContent))))
 		}
 
 		fmt.Fprint(writer, resetEscapeSequence)
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -2552,10 +2552,15 @@ func ComputePositionOfLineAndUTF16Character(lineStarts []core.TextPos, line int,
 			utf16Count += core.UTF16Offset(utf16.RuneLen(r))
 			pos += size
 		}
-		if allowEdits {
-			if pos > len(text) {
-				return len(text)
+		if !allowEdits {
+			if pos == lineEnd && utf16Count < character {
+				panic(fmt.Sprintf("Bad UTF-16 character offset. Line: %d, character: %d.", line, character))
 			}
+			debug.Assert(pos <= len(text))
+			return pos
+		}
+		if pos > len(text) {
+			return len(text)
 		}
 		return pos
 	}


internal/printer/textwriter.go
```diff
@@ -31,11 +31,15 @@ func (w *textWriter) DecreaseIndent() {
...
+// for source map compatibility.
+func (w *textWriter) GetColumn() core.UTF16Offset {
 	if w.lineStart {
-		return w.indent * w.indentSize
+		return core.UTF16Offset(w.indent * w.indentSize)
 	}
-	return w.builder.Len() - w.linePos
+	// Count UTF-16 code units from the last line start.
+	// For ASCII-only output (the common case), this equals the byte count.
+	return core.UTF16Len(w.builder.String()[w.linePos:])
```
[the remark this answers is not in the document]
jakebailey: `.String()` is free, so we're really just talking about how long it takes to calculate, and I think it's better to calculate the len on request rather than repeatedly len-ing on every single write.

internal/diagnosticwriter/diagnosticwriter.go
```diff
@@ -229,21 +229,21 @@ func writeCodeSnippet(writer io.Writer, sourceFile FileLike, start int, length i
...
+			fmt.Fprint(writer, strings.Repeat(" ", int(firstLineChar)))
+			fmt.Fprint(writer, strings.Repeat("~", lastCharForLine-int(firstLineChar)))
 		case lastLine:
 			// Squiggle until the final character.
-			fmt.Fprint(writer, strings.Repeat("~", lastLineChar))
+			fmt.Fprint(writer, strings.Repeat("~", int(lastLineChar)))
 		default:
 			// Squiggle the entire line.
-			fmt.Fprint(writer, strings.Repeat("~", len(lineContent)))
+			fmt.Fprint(writer, strings.Repeat("~", int(core.UTF16Len(lineContent))))
```
jakebailey, line 246: This is very odd, and arguably should _actually_ be rune counts, since these are visible characters.

testdata/baselines/reference/submodule/compiler/sourceMap-LineBreaks(target=es2015).sourcemap.txt
```diff
@@ -79,12 +79,12 @@ sourceFile:sourceMap-LineBreaks.ts
...
-2 >�var
-3 >     endsWithLineFee
-4 >                    d =
-5 >                        
-6 >                        1
+1->

+2 >var 
+3 >    endsWithLineFeed
+4 >                     =
```
jakebailey, line 85: Note the fix here.

internal/diagnosticwriter/diagnosticwriter.go
DanielRosenwasser: Are you just holding off on this to avoid a massive baseline avalanche?

internal/scanner/scanner.go
```diff
@@ -2459,14 +2460,27 @@ func GetECMALineOfPosition(sourceFile ast.SourceFileLike, pos int) int {
...
-	character = utf8.RuneCountInString(sourceFile.Text()[lineMap[line]:pos])
+	character = core.UTF16Len(sourceFile.Text()[lineMap[line]:pos])
 	return line, character
 }
 
+// GetECMALineAndByteOffsetOfPosition returns the 0-based line number and the
+// raw UTF-8 byte offset from the start of that line for the given byte position.
+// Uses ECMAScript line separators (LF, CR, CRLF, LS, PS).
+// Unlike GetECMALineAndUTF16CharacterOfPosition, the offset is in bytes, not UTF-16 code units.
+func GetECMALineAndByteOffsetOfPosition(sourceFile ast.SourceFileLike, pos int) (line int, byteOffset int) {
```
DanielRosenwasser, line 2477: Minor - do you think `From` reads better here than `Of`?

internal/format/indent.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func GetIndentationForNode(n *ast.Node, ignoreActualIndentationRange *core.TextRange, sourceFile *ast.SourceFile, options *lsutil.FormatCodeSettings) int {
-	startline, startpos := scanner.GetECMALineAndCharacterOfPosition(sourceFile, scanner.GetTokenPosOfNode(n, sourceFile, false))
+	startline, startpos := scanner.GetECMALineAndByteOffsetOfPosition(sourceFile, scanner.GetTokenPosOfNode(n, sourceFile, false))
```
DanielRosenwasser, line 17: Is it just me, or is it surprising that the formatter uses ECMAScript-based lines instead of LSP-based lines?

internal/printer/textwriter.go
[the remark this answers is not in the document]
DanielRosenwasser: What do you mean by "`.String()` is free"? Doesn't it have to reallocate on any new content additions?

internal/testutil/harnessutil/sourcemap_recorder.go
```diff
@@ -289,7 +290,14 @@ func (sw *recordedSpanWriter) writeSourceMapMarkerEx(currentSpan *sourceMapSpanW
...
 
 func (sw *recordedSpanWriter) writeSourceMapSourceText(currentSpan *sourceMapSpanWithDecodeErrors, index int) {
-	sourcePos := int(sw.w.tsLineMap[currentSpan.sourceMapSpan.SourceLine]) + currentSpan.sourceMapSpan.SourceCharacter
+	// Convert UTF-16 character offset from the source map to a byte position.
+	sourcePos := scanner.ComputePositionOfLineAndUTF16Character(
+		sw.w.tsLineMap,
+		currentSpan.sourceMapSpan.SourceLine,
+		currentSpan.sourceMapSpan.SourceCharacter,
+		sw.w.tsCode,
+		true, /*allowEdits*/
```
DanielRosenwasser, line 299:
```suggestion
		true /*allowEdits*/,
```

?

DanielRosenwasser (commented):
Looks good, just wanted to get some clarifications before approval.

internal/testutil/harnessutil/sourcemap_recorder.go
jakebailey: The formatter rejects this, unfortunately

internal/printer/textwriter.go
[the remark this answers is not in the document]
jakebailey: No, it's internally an "unsafe" direct access so long as you don't escape it

internal/diagnosticwriter/diagnosticwriter.go
jakebailey: No matter what we do, things will be just a little weird, so I just didn't change it

internal/scanner/scanner.go
jakebailey: I'm indifferent, I was just trying to not over-rename if someone goes looking

internal/format/indent.go
jakebailey: I totally agree, and we could plumb it differently, but I'm not sure it actually matters for these.

internal/printer/textwriter.go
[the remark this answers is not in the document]
DanielRosenwasser: Feels sketchy but I'll take your word.

PR #2877, opened 2026-02-23
[@typescript/api] Add generated node factory, two-way AST encoding/decoding, emitter.printNode

andrewbranch:
- generates a node factory and `forEachChild` visitor for the API client
  - ⚠️ **There are a few undesirable differences** from the Strada factory where functions have bespoke logic to turn booleans into flags, or certain properties are supposed to be unsettable by factory. A diff is below. Eventually, we want to codegen ASTs, factories, and visitors on _both_ the Go side and TS side, at which point we can clean this kind of thing up in a systematic way.
- adds AST encoding on the client and decoding on the server so nodes can be communicated over the API in either direction
- adds `checker.typeToTypeNode`, `checker.typeToString`, and `emitter.printNode` as API calls
- additional drive-by optimization of client-side AST decoding for another 16% faster

<details><summary>Factory diff analysis</summary>

```
Generated factory: 193 create* functions
Original factory: 234 create* functions

=== MISSING from generated (69) ===
  createArrayConcatCall(2 params) -> unknown
  createArraySliceCall(2 params) -> unknown
  createAssertClause(2 params) -> AssertClause
  createAssertEntry(2 params) -> AssertEntry
  createAssignmentTargetWrapper(2 params) -> PropertyAccessExpression
  createBaseCallExpression(4 params) -> unknown
  createBaseElementAccessExpression(3 params) -> unknown
  createBaseGeneratedIdentifier(4 params) -> unknown
  createBaseGeneratedPrivateIdentifier(4 params) -> unknown
  createBaseIdentifier(1 params) -> unknown
  createBasePrivateIdentifier(1 params) -> unknown
  createBasePropertyAccessExpression(3 params) -> unknown
  createBaseStringLiteral(2 params) -> unknown
  createBundle(1 params) -> unknown
  createCallChain(4 params) -> unknown
  createCallSignature(3 params) -> CallSignatureDeclaration
  createConstructSignature(3 params) -> ConstructSignatureDeclaration
  createConstructorTypeNode1(4 params) -> ConstructorTypeNode
  createConstructorTypeNode2(3 params) -> ConstructorTypeNode
  createElementAccessChain(3 params) -> unknown
  createExportDefault(1 params) -> unknown
  createExternalModuleExport(1 params) -> unknown
  createFalse(0 params) -> unknown
  createFunctionApplyCall(3 params) -> unknown
  createFunctionBindCall(3 params) -> unknown
  createFunctionCallCall(3 params) -> unknown
[...]

PR #2885, opened 2026-02-24
Fix retrieval of named children in JS RemoteNode

auvred:
```typescript
foo.bar === baz
```

In binary AST nodes are encoded in this order:

- `ExpressionStatement`
- `BinaryExpression`
- `PropertyAccessExpression`
- `Identifier`
- `Identifier`
- `EqualsEqualsEqualsToken`
- `Identifier`

On `main`:

We're trying to read `.operatorToken` of `BinaryExpression`, `propertyIndex` is 1, `getNamedChild('operatorToken')` returns `BinaryExpression.index + 1 + 1`th node. It returns  an `Identifier`.

In this PR:

We're trying to read `.operatorToken` of `BinaryExpression`, `propertyIndex` is 1, we iterate over the immediate child nodes and jump straight to their sibling nodes using their `.next` value. `getNamedChild('operatorToken')` returns an `EqualsEqualsEqualsToken`.


auvred:
Oh, this was already fixed in https://github.com/microsoft/typescript-go/pull/2877

PR #2888, opened 2026-02-24
Fixed a class member completion crash after JSDoc comment with degenerate JSDoc tag in it

Andarist:
fixes the crash reported here: https://github.com/microsoft/typescript-go/issues/2871#issuecomment-3941898703
it also crashes in Strada


internal/astnav/tokens.go
```diff
@@ -579,7 +579,11 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
 		if rightmostValidNode != nil {
 			endPos = rightmostValidNode.End()
 		}
-		return find(rightmostValidNode, endPos)
+		result := find(rightmostValidNode, endPos)
+		if result == nil && n != containingNode {
+			return n
+		}
```
Andarist: In this case, the JSDoc comment contains a degenerate `JSDocTag` with 0-width (representing `"@-rule"`) - given `@tag` requires an identifier and `-` isn't a valid identifier character. 

This matches how Strada parses the same case. But it creates an issue because the compiler fails to find `contextToken`. A `nil` from here just propagates up and later on it crashes in the completion code.

internal/fourslash/tests/completionsAtTopLevelImportAssignmentNoCrash1_test.go
```diff
@@ -28,8 +28,6 @@ import x =/*3*/
 		ItemDefaults: &fourslash.CompletionsExpectedItemDefaults{
 			CommitCharacters: &[]string{},
 		},
-		Items: &fourslash.CompletionsExpectedItems{
-			Includes: []fourslash.CompletionsExpectedItem{}, // no crash check
-		},
+		Items: &fourslash.CompletionsExpectedItems{},
```
Andarist, line 31: this is a driveby change, I just corrected here one of the recent tests I added to use the more common codebase pattern

internal/astnav/tokens.go
gabritto: This looks ok if we don't want to change the parsing, but I do wonder if maybe that's what we should do (even though we had this problem in Strada, too).

Andarist: It crossed my mind to attempt a change there but I decided to err on the side of Strada compatibility for now. If you think the parsing should be changed instead, then I could start exploring that option.

gabritto: I think it's worth trying to fix it during parsing, yes. Given this would be a change in JSDoc invalid tag parsing, I don't think it's likely it would break people.

gabritto:
(I'm looking into the race mode test failure; seems like a flaky test not related to this PR)

internal/scanner/scanner.go
```diff
@@ -1275,15 +1275,15 @@ func (s *Scanner) ScanJSDocCommentTextToken(inBackticks bool) ast.Kind {
...
-					if s.pos+size >= len(s.text) {
-						// EOF counts as non-whitespace
-						break
-					}
-					next, _ := utf8.DecodeRuneInString(s.text[s.pos+size:])
-					if !stringutil.IsWhiteSpaceLike(next) {
+					saved := s.Mark()
+					s.pos += size
+					s.ScanJSDocToken()
+					isIdentifier := tokenIsIdentifierOrKeyword(s.token)
```
sandersn: is there a definition of what an identifier or keyword starts with so we don't have to scan a whole jsdoc token? My instinct is to avoid backtracking.

this may be a non-issue in the Corsa scanner; I can't remember how expensive Mark/Rewind+ScanJSDocToken are, and `@` is not a common character in jsdoc in any case.

Andarist: Yeah, you were right to call this out - this "tryParse" was entirely not needed here. I simplified the code a bunch now (not only here).

internal/fourslash/tests/completionAfterTrailingAtInJSDoc1_test.go
```diff
@@ -0,0 +1,52 @@
+package fourslash_test
```
gabritto, line 1: I think these tests are all new and should move to `fourslash/tests`.

internal/scanner/scanner.go
```diff
@@ -1275,15 +1275,11 @@ func (s *Scanner) ScanJSDocCommentTextToken(inBackticks bool) ast.Kind {
 			if ch == '{' {
 				break
 			} else if ch == '@' && s.pos >= 0 {
-				// @ doesn't start a new tag inside ``, and elsewhere, only after whitespace and before non-whitespace
+				// @ doesn't start a new tag inside ``, and elsewhere, only after whitespace and before identifier
 				previous, _ := utf8.DecodeLastRuneInString(s.text[:s.pos])
 				if stringutil.IsWhiteSpaceSingleLine(previous) {
-					if s.pos+size >= len(s.text) {
```
gabritto, line 1281: Why is this gone?

Andarist: This actually makes `TestQuickInfoJSDocParamWithTrailingAtBeforeCommentEnd` work.

With the EOF check added back, the scanner hits `@` followed by EOF, breaks out of the text loop, and the parser treats that `@` as starting a new (empty) tag. So the output becomes `trailing ` + `\n\n*@*` — a bogus tag rendering.

Without the EOF check, `@` followed by `EOF`  goes for `IsIdentifierStart` and that returns `false`. There is no break and `@` stays as literal text in the description.

internal/fourslash/tests/completionAfterTrailingAtInJSDoc1_test.go
Andarist: fixed

PR #2898, opened 2026-02-24
Update Go image version in devcontainer.json

DanielRosenwasser:


.devcontainer/devcontainer.json
jakebailey:
```suggestion
```

PR #2904, opened 2026-02-25
[@typescript/api] Port scanner and token navigation utils to API client, include JSDoc in encoded ASTs, add more SourceFile properties

andrewbranch:


testdata/baselines/reference/astnav/GetTokenAtPosition.mapCode.ts.baseline.json
andrewbranch, line 1: I added this so the client-side token utils can validate identical behavior to the Go astnav utils. Open to ideas for how to not dump 20k lines of JSON to disk if this is upsetting.

jakebailey: You could dump it encoded in arrays, to limit how much text it is?

_packages/ast/src/astnav.ts
andrewbranch, line 1: This is a direct port of the Go code, not the Strada code

_packages/api/src/node/encoder.ts
```diff
@@ -199,14 +209,32 @@ function recordNodeStrings(node: Node, strs: StringTable): number {
...
     if (node.kind === SyntaxKind.SourceFile) {
         const sf = node as SourceFile;
         const textIndex = strs.add(sf.text);
         const fileNameIndex = strs.add(sf.fileName);
         const pathIndex = strs.add(sf.path);
-        extendedData.push(textIndex, fileNameIndex, pathIndex);
+        const referencedFilesOffset = encodeFileReferences(sf.referencedFiles, structuredWriter);
+        const typeRefDirectivesOffset = encodeFileReferences(sf.typeReferenceDirectives, structuredWriter);
+        const libRefDirectivesOffset = encodeFileReferences(sf.libReferenceDirectives, structuredWriter);
+        extendedData.push(textIndex, fileNameIndex, pathIndex, sf.languageVariant, sf.scriptKind, referencedFilesOffset, typeRefDirectivesOffset, libRefDirectivesOffset, NO_STRUCTURED_DATA, NO_STRUCTURED_DATA, NO_STRUCTURED_DATA, 0);
```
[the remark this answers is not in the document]
andrewbranch: That's because you can't construct a SourceFile with these properties from the client.

testdata/baselines/reference/astnav/GetTokenAtPosition.mapCode.ts.baseline.json
andrewbranch: I'm not bothered by it; it's small potatoes compared to a lot of the project baselines

_packages/ast/src/scanner.ts
jakebailey, line 1: How should we keep this up to date going forward?

How does this manage differences in encoding?

andrewbranch: What do you think of the latest commit for that?

jakebailey: I meant more generally, in that it's a 1:1 copy of the scanner, no?

andrewbranch: Yeah, pretty much. I don't think the scanner changes enough to be very concerned. We can add more files to the GetTokenAtPosition baseline tests if we want?

jakebailey: I guess not, though https://github.com/microsoft/typescript-go/pull/2888 is open which does in fact change the scanner, which is convenient timing

jakebailey: This scanner is also missing `ContainsNonASCII`; does that much matter?

andrewbranch: No, that would never be used in this implementation.

Issue #2882, opened 2026-02-23
Probable nil entry from specifier cache in auto-import completions


DanielRosenwasser:
Unknown repro right now

```
internal/collections.(*SyncMap[...]).Load()
	internal/collections/syncmap.go:15
internal/ls/autoimport.(*View).GetModuleSpecifier()
	internal/ls/autoimport/specifiers.go:46
internal/ls/autoimport.(*View).GetFixes()
	internal/ls/autoimport/fix.go:558
internal/ls/autoimport.(*View).GetCompletions()
	internal/ls/autoimport/view.go:232
internal/ls.(*LanguageService).getCompletionData.func9()
	internal/ls/completions.go:1115
internal/ls.(*LanguageService).getCompletionData.func17()
	internal/ls/completions.go:1523
internal/ls.(*LanguageService).getCompletionData.func18()
	internal/ls/completions.go:1553
internal/ls.(*LanguageService).getCompletionData()
	internal/ls/completions.go:1586
internal/ls.(*LanguageService).getCompletionsAtPosition()
	internal/ls/completions.go:347
internal/ls.(*LanguageService).ProvideCompletion()
	internal/ls/completions.go:47
internal/lsp.(*Server).handleCompletion()
	internal/lsp/server.go:1167
internal/lsp.init.func1.registerLanguageServiceWithAutoImportsRequestHandler[...].28.1()
	internal/lsp/server.go:750
internal/lsp.(*Server).handleRequestOrNotification.func1()
	internal/lsp/server.go:579
internal/lsp.(*Server).dispatchLoop.func3()
	internal/lsp/server.go:454
internal/lsp.(*Server).dispatchLoop
	internal/lsp/server.go:453
```

PR #2765, opened 2026-02-12
Acquire snapshots synchronously while blocking other requests/notifications

andrewbranch:
Fixes #2690 

Imagine a sequence of

1. textDocument/didChange
2. textDocument/inlayHint
3. textDocument/didChange

When we saw the first change in the server dispatch loop, we would hold processing of the rest of the queue while that change gets pushed. Then we would resume and call `s.handleRequestOrNotification` with the inlay hint one, and (without waiting for it to complete) handle the next change, holding any more pending requests. The problem is that `s.handleRequestOrNotification` didn't get passed a snapshot; it looked at the request method, looked up the correct handler in the server, which would _then_ get acquire a snapshot (in the form of a LanguageService) and pass it to the handler. So there was a _tiny_ bit of intermediate time that was available for the next change notification to sneak in.

With this PR, handlers now have a sync portion that gets executed serially in the dispatch loop, which they use to secure a snapshot, and return a function of async work that uses the snapshot.

The downside of this is that if a client isn’t diligent about canceling old requests, we have no choice but to manifest a program for each one, even if we've built up changes that make them stale. The LSP spec says as much:

> - if a client sends a request to the server and the client state changes in a way that invalidates the response, the client should do the following:
>   - cancel the server request and ignore the result if the result is not useful for the client anymore. If necessary, the client should resend the request.
>   - keep the request running if the client can still make use of the result by, for example, transforming it to a new result by applying the state change to the result.
> - servers should therefore not decide by themselves to cancel requests simply due to that fact that a state change notification is detected in the queue. As said, the result could still be useful for the client.

So I think this is the right approach.

Issue #2822, opened 2026-02-18
Harden LSP against requests for files that don't have open projects


andrewbranch:
While https://github.com/microsoft/vscode-languageserver-node/issues/1717 turned out to be a bug, we should not have been falling over so hard. We need to have a way of responding to requests for non-open files that we haven't opened a project for.

DanielRosenwasser:
Worth mentioning - VS Code also requests diagnostics after closing files which causes issues for us as well.

```
2026-02-23 23:42:28.183 [trace] Sending notification 'textDocument/didClose'.
Params: {
    "textDocument": {
        "uri": "untitled:Untitled-3"
    }
}
2026-02-23 23:42:28.183 [trace] Sending request 'textDocument/diagnostic - (67)'.
Params: {
    "textDocument": {
        "uri": "untitled:Untitled-3"
    }
}
```

Kind of hard to repro, but I did it via "close others" from right clicking on a tab.

PR #2905, opened 2026-02-25
Fixed a crash on completions with auto-imports racing with `DidClose` notification

Andarist:
Root cause behind the crash:
- the completion request is split into 2 phases, a sync one and async one
- in the async one the completion request might be *retried* on `ErrNeedsAutoImports`
- but in between the first one and the second one a project change might come in (like `DidClose` or some other change)
- currently, the retried request flushes the pending changes (like `DidClose`) and resends the adjusted request
- that creates a mismatched code expectation because the closed file gets removed from the `specifierCache` and the request panics

There are 2 fixes in this PR:
- a request should always be served for the `change.RequestedFile`. That entry isn't removed from the `specifierCache`. It's not up to the server to cancel a pending request even if the state changed.
- a 2-phase request should use a consistent snapshot for the whole request to avoid its steps using different state and ultimately leading to confusing/erroneous results (although usually a small deviation there would usually be somewhat acceptable... I think it's still better to use a consistent state, and I think it's consistent with https://github.com/microsoft/typescript-go/pull/2765 ). For that reason, a snapshot gets acquired in the sync phase and reused by the async phase - achieving consistency.

| Configuration | close before completion | close during async | unopened file | snapshot freezing (DidChange) |
|---|---|---|---|---|
| **No fixes** | CRASH (no response) | CRASH (nil deref panic) | PASS | FAIL (someVar missing) |
| **Registry fix only** | PASS | PASS | PASS | FAIL (someVar missing) |
| **Snapshot freezing only** | CRASH (no response) | PASS | PASS | PASS |
| **Both fixes** | PASS | PASS | PASS | PASS |

fixes https://github.com/microsoft/typescript-go/issues/2882
closes https://github.com/microsoft/typescript-go/issues/2985


andrewbranch:
I have a branch with this in it as well that I passed off to @gabritto as she was working on fixing a related crash. I think my branch wasn't quite right or was insufficient; I'm not sure what the status is. However, in mine, after we clone the snapshot for auto-imports, we check to see if the session can adopt the new one as the latest snapshot - I think we'll want that functionality.

internal/lsp/server_completion_test.go
Andarist, line 1: I couldn't really test this through fourslash so I had to resort to rolling a new test file that could use the `NewLSPClient` directly with a greater control over the performed actions.

internal/project/session.go
```diff
@@ -556,6 +568,31 @@ func (s *Session) GetLanguageServiceWithAutoImports(ctx context.Context, uri lsp
...
+	}, baseSnapshot.fs.overlays, s)
+
+	project := newSnapshot.GetDefaultProject(uri)
+	if project == nil {
+		if newSnapshot.Deref() {
+			newSnapshot.dispose(s)
+		}
+		return nil, nil, fmt.Errorf("no project found for URI %s", uri)
+	}
+
```
gabritto, line 637: @Andarist Here's how I was doing the work of re-using the snapshot + extra information computed in an old PR:
https://github.com/microsoft/typescript-go/pull/1871/changes#diff-b8a62f01d8c9b6ed8df3c61e62fe4b694885da67eadb682a4117ce2cb62b1bf9R754-R784

But the tricky part would be to figure out whether to release the snapshot or not, I'll have to think about that.

internal/lsp/server_completion_test.go
gabritto: Couldn't we write a test using session directly? I did this recently in https://github.com/microsoft/typescript-go/blob/main/internal/project/snapshot_test.go#L125 to test things related to the timing of changes being processed.
We may need a better testing setup for those kinds of bugs long-term, but maybe right now using session for testing could do.

internal/project/session.go
gabritto: I think for the snapshot lifetime handling, we can (1) invoke `snapshot.dispose()` inside `snapshot.Deref()` when the count reaches 0, and (2) make sure both session and the server call `.Ref()` and `.Deref()` when they acquire/release it. So if you did the whole thing I linked above of enqueuing of a snapshot with auto-imports to see if session can adopt it, that would come with a call to `.Ref()`/`.Deref()`, etc.

internal/lsp/server_completion_test.go
Andarist: I planned to add a test using the session directly later on. I just wanted to have more real-life test cases for this too and focused on those here.

internal/project/session.go
Andarist: @gabritto could you take a look at what I have here now in regards to this requested refactor?

internal/ls/autoimport/registry.go
```diff
@@ -224,7 +224,7 @@ func (r *Registry) IsPreparedForImportingFile(fileName string, projectPath tspat
 	}
 	projectBucket, ok := r.projects[projectPath]
 	if !ok {
-		panic("project bucket missing")
```
gabritto, line 225: I think this makes sense, but when would we hit this?

internal/ls/autoimport/registry.go
```diff
@@ -433,8 +433,15 @@ func (b *registryBuilder) updateBucketAndDirectoryExistence(change RegistryChang
 		}
 	}
 
+	if change.RequestedFile != "" {
+		neededProjects[core.FirstResult(b.host.GetDefaultProject(change.RequestedFile))] = struct{}{}
+		if !b.specifierCache.Has(change.RequestedFile) {
+			b.specifierCache.Set(change.RequestedFile, &collections.SyncMap[tspath.Path, string]{})
+		}
+	}
```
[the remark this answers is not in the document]
Andarist: the identical pattern was already used at line 412 so I assume it's OK in this context (or that it should be fixed at both places in some followup work)

internal/lsp/server.go
```diff
@@ -766,18 +766,21 @@ func registerLanguageServiceWithAutoImportsRequestHandler[Req lsproto.HasTextDoc
 		if req.Params != nil {
 			params = req.Params.(Req)
 		}
-		languageService, err := s.session.GetLanguageService(ctx, params.TextDocumentURI())
+		languageService, snapshot, releaseSnapshot, err := s.session.GetLanguageServiceAndSnapshot(ctx, params.TextDocumentURI())
 		if err != nil {
 			return nil, err
 		}
 		return func() error {
+			defer releaseSnapshot()
```
gabritto, line 772: @andrewbranch do we need to worry about this request being cancelled in such a timing that we acquire the snapshot, but then never call the async part of this because the request is cancelled, and thus never call `releaseSnapshot()`? If so, what's the best way of handling this?

internal/lsp/server_completion_test.go
```diff
@@ -0,0 +1,281 @@
...
+		if item.Label == label {
+			return item
+		}
+	}
+	return nil
+}
+
+// Verifies that completion succeeds on a file that was already closed
+// by the time the server processes the completion request.
+func TestCompletionAfterFileClose(t *testing.T) {
```
gabritto, line 86: I think in theory this shouldn't ever happen, but IIRC @andrewbranch wanted to make such scenarios work anyways.

Andarist: I work here under the assumptions that:
- the client might be buggy
- the LSP spec seems to mandate the server serving such requests if possible

jakebailey: Yeah, it can totally happen, but we are in general not resilient to requests on closed files

Andarist: q: you are not resilient *yet* or *by design*?

internal/lsp/server.go
Andarist: I just pushed out a potential fix: https://github.com/microsoft/typescript-go/pull/2905/commits/d9c62271f8640ed708384bb7626d0d83d803443b

Note that all of those async parts handle the `ctx.Err()` already - that's why the diff is so minimal

internal/lsp/server_completion_test.go
jakebailey: https://github.com/microsoft/typescript-go/issues/2822

internal/ls/autoimport/registry.go
Andarist:
```
--- FAIL: TestAutoImportCompletionForUnopenedFile (0.12s)
    server_completion_test.go:223: assertion failed: ok is false: expected a response
--- FAIL: TestCompletionForUnopenedFile (0.13s)
    server_completion_test.go:195: assertion failed: ok is false: expected a response
```

PR #2914, opened 2026-02-26
adds tracing (--generateTrace)

dimitropoulos:
this is a work in progress PR to add tracing so we can get types.json and trace.json files as was in the TypeScript codebase.

note, I took all the previous work and squashed it into one commit, then the second commit (56f3a9f473de2da3d14ce11e2b8e746e50beca75) is for switching to a context-based approach.


internal/checker/checker.go
```diff
@@ -881,14 +882,22 @@ type Checker struct {
...
+	mu      sync.Mutex
+	tracer  TypeTracer              // Optional tracer for recording types (for --generateTrace)
+	tracing *tracing.CheckerTracing // Optional per-checker tracing for trace events (for --generateTrace)
 }
 
 func NewChecker(program Program) (*Checker, *sync.Mutex) {
+	return NewCheckerWithTracer(program, nil, nil)
+}
+
+func NewCheckerWithTracer(program Program, tracer TypeTracer, tr *tracing.CheckerTracing) (*Checker, *sync.Mutex) {
```
jakebailey: I don't particularly like that we have this variant... what was wrong with the approach I had in my branch?

internal/checker/checker.go
```diff
@@ -2422,6 +2435,7 @@ func (c *Checker) checkDeferredNodes(context *ast.SourceFile) {
 }
 
 func (c *Checker) checkDeferredNode(node *ast.Node) {
+	c.tracing.Push(tracing.PhaseCheck, "checkDeferredNode", false, "kind", strconv.Itoa(int(node.Kind)), "pos", strconv.Itoa(node.Pos()), "end", strconv.Itoa(node.End()), "path", ast.GetSourceFileOfNode(node).FileName())
```
jakebailey: All tracing calls should do an `if c.tracing != nil` check; these are expensive operations to be doing on ever single node.

internal/checker/checker.go
dimitropoulos: how's it lookin now? 91c5b6d9761a4530fb25f6dbcee135835a0165d1

internal/checker/checker.go
dimitropoulos: I believe they should all be covered as of aecc06ce2f7be3b1eb2cb065198dd9c947777b0f, but please unresolve if not

internal/checker/checker.go
dimitropoulos: (see updated description)

internal/checker/checker.go
```diff
@@ -2460,6 +2470,9 @@ func (c *Checker) checkDeferredNodes(context *ast.SourceFile) {
 }
 
 func (c *Checker) checkDeferredNode(node *ast.Node) {
+	if tr := tracing.FromContext(c.ctx); tr != nil {
```
jakebailey: This is doing `FromContext`, but we still have tracer on the checker.

I think that we should be doing `FromContext` once in `checkSourceFile`, assign _that_ to the checker's field, and then we can just check `if tr := c.tracer; tr != nil {` ... to save a context lookup everywhere.

internal/checker/checker.go
```diff
@@ -24382,6 +24407,9 @@ func (c *Checker) newType(flags TypeFlags, objectFlags ObjectFlags, data TypeDat
 	t.id = TypeId(c.TypeCount)
 	t.checker = c
 	t.data = data
+	if c.tracer != nil {
```
jakebailey, line 24412: This one is inconsistent and should be written in the same `if tr := c.tracer` form.

internal/checker/relater.go
```diff
@@ -374,6 +375,9 @@ func (c *Checker) checkTypeRelatedToEx(
 		// Record this relation as having failed such that we don't attempt the overflowing operation again.
 		id, _ := getRelationKey(source, target, IntersectionStateNone, relation == c.identityRelation, false /*ignoreConstraints*/)
 		relation.set(id, RelationComparisonResultFailed|core.IfElse(r.relationCount <= 0, RelationComparisonResultComplexityOverflow, RelationComparisonResultStackDepthOverflow))
+		if tr := tracing.FromContext(c.ctx); tr != nil {
```
jakebailey: Similarly for all of these, to save the lookups within the checker.

internal/checker/types.go
```diff
@@ -448,6 +450,60 @@ const (
 	TypeFlagsNotPrimitiveUnion               = TypeFlagsAny | TypeFlagsUnknown | TypeFlagsVoid | TypeFlagsNever | TypeFlagsObject | TypeFlagsIntersection | TypeFlagsIncludesInstantiable
 )
 
+var typeFlagNames = [...]struct {
```
jakebailey, line 453: You don't have to worry about this, but, I have a PR I'm going to send soon that will automate this generation for us.

internal/compiler/checkerpool.go
```diff
@@ -94,7 +103,11 @@ func (p *checkerPool) createCheckers() {
 		wg := core.NewWorkGroup(p.program.SingleThreaded())
 		for i := range checkerCount {
 			wg.Queue(func() {
-				p.checkers[i], p.locks[i] = checker.NewChecker(p.program)
+				var tracer checker.TypeTracer
+				if p.tracing != nil {
+					tracer = checker.NewTracingTypeTracer(p.tracing.NewTypeTracer(i))
+				}
+				p.checkers[i], p.locks[i] = checker.NewCheckerWithTracer(p.program, tracer)
```
jakebailey, line 110: I don't think we need "new checker with tracer" if we only ever trace during checkSourceFile?

internal/compiler/fileloader.go
```diff
@@ -460,6 +471,8 @@ func (p *fileLoader) resolveTypeReferenceDirectives(t *parseTask) {
 const externalHelpersModuleNameText = "tslib" // TODO(jakebailey): dedupe
 
 func (p *fileLoader) resolveImportsAndModuleAugmentations(t *parseTask) {
+	p.opts.Tracing.Push(tracing.PhaseProgram, "resolveModuleNamesWorker", map[string]any{"containingFileName": t.file.FileName()}, false)
```
jakebailey: We should make basically _all_ uses of tracing precheck for nil, because otherwise we are constructing these maps even when not tracing, which is expensive.

internal/compiler/program.go
```diff
@@ -404,11 +407,17 @@ func (p *Program) SingleThreaded() bool {
 	return p.opts.SingleThreaded.DefaultIfUnknown(p.Options().SingleThreaded).IsTrue()
 }
 
+// Tracing returns the tracing session associated with this program, or nil if tracing is not enabled.
+func (p *Program) Tracing() *tracing.Tracing {
```
jakebailey: I haven't found it yet, but why would we have this, if tracing is flowing in via contexts?

internal/tracing/tracing.go
```diff
@@ -0,0 +1,779 @@
...
+	}
+}
+
+// writeEventRaw writes a trace event to the shared trace content.
+func (tr *Tracing) writeEventRaw(ph string, cat string, ts float64, name string, argsJSON string, extras ...string) {
+	writeEventTo(&tr.traceContent, ph, cat, ts, name, argsJSON, extras...)
+}
+
+// writeArgsJSON constructs a deterministic JSON object from a map of key-value pairs.
+func writeArgsJSON(args map[string]any) string {
```
jakebailey, line 229: is there a reason to do it this way as compared to using json/jsontext to directly writer a stream of tokens? Doing it that way should be free, though maybe it's down to the newlines? (I think the lib supports that)

jakebailey:
Yeah, baselines are unstable:

```diff
diff --git a/testdata/baselines/reference/tsc/generateTrace/generateTrace-with-multiple-files-and-complex-types.js b/testdata/baselines/reference/tsc/generateTrace/generateTrace-with-multiple-files-and-complex-types.js
index b79d7c8f..d48ad2d3 100644
--- a/testdata/baselines/reference/tsc/generateTrace/generateTrace-with-multiple-files-and-complex-types.js
+++ b/testdata/baselines/reference/tsc/generateTrace/generateTrace-with-multiple-files-and-complex-types.js
@@ -96,6 +96,8 @@ declare const console: { log(msg: any): void; };
 {"pid":1,"tid":1,"ph":"B","cat":"check","ts":21,"name":"checkSourceFile","args":{"path":"/home/src/workspaces/project/types.ts"}},
 {"pid":1,"tid":1,"ph":"E","cat":"check","ts":22,"name":"checkSourceFile","args":{"path":"/home/src/workspaces/project/types.ts"}},
 {"pid":1,"tid":1,"ph":"B","cat":"check","ts":23,"name":"checkSourceFile","args":{"path":"/home/src/workspaces/project/main.ts"}},
+{"pid":1,"tid":1,"ph":"X","cat":"check","ts":1072.7570,"name":"checkExpression","dur":23361.6560,"args":{"end":133,"kind":218,"path":"/home/src/workspaces/project/main.ts","pos":101}},
+{"pid":1,"tid":1,"ph":"X","cat":"check","ts":1069.8220,"name":"checkDeferredNode","dur":23369.2300,"args":{"end":133,"kind":220,"path":"/home/src/workspaces/project/main.ts","pos":93}},
 {"pid":1,"tid":1,"ph":"E","cat":"check","ts":24,"name":"checkSourceFile","args":{"path":"/home/src/workspaces/project/main.ts"}},
 {"pid":1,"tid":1,"ph":"E","cat":"check","ts":25,"name":"checkSourceFiles"},
 {"pid":1,"tid":1,"ph":"B","cat":"emit","ts":26,"name":"emit"},
```

internal/checker/checker.go
dimitropoulos: Done. Added a `tracing *tracing.Tracing` field to `Checker`, set it once at the top of `checkSourceFile` via `c.tracing = tracing.FromContext(ctx)`, and changed all call sites in checker.go to use `if tr := c.tracing; tr != nil { ... }` instead of calling `tracing.FromContext(c.ctx)` each time.

internal/checker/checker.go
dimitropoulos: Fixed in the same commit — all sites now use the `if tr := c.tracing; tr != nil` pattern consistently.

internal/checker/relater.go
dimitropoulos: All 6 `tracing.FromContext(r.c.ctx)` / `tracing.FromContext(c.ctx)` calls in relater.go and the 1 call in flow.go are now changed to use `r.c.tracing` / `c.tracing`.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,779 @@
...
+
+	if entry.separateBeginAndEnd {
+		buf.WriteString(",\n")
+		writeEventTo(buf, "E", string(entry.phase), ts, entry.name, entry.argsJSON)
+	} else {
+		now := time.Now()
+		startMicros := float64(entry.startTime.Sub(startTime).Nanoseconds()) / 1000.0
+		endMicros := float64(now.Sub(startTime).Nanoseconds()) / 1000.0
+		intervalMicros := float64(sampleInterval.Nanoseconds()) / 1000.0
+		dur := endMicros - startMicros
```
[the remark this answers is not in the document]
dimitropoulos: In deterministic mode, sampled events (separateBeginAndEnd=false) are now skipped entirely. Only B/E pairs (separateBeginAndEnd=true) use the deterministic counter, and sampled "X" events inherently depend on wall-clock time. Skipping them removes the source of flaky baselines. This also addresses the unstable baselines Jake observed in the timeline comment.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,779 @@
...
+		b.WriteString("\"")
+		b.WriteString(k)
+		b.WriteString("\":")
+		v := args[k]
+		switch val := v.(type) {
+		case string:
+			b.WriteString("\"")
+			b.WriteString(val)
+			b.WriteString("\"")
+		case bool:
```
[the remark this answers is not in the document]
dimitropoulos: `writeArgsJSON` now uses `json.Marshal()` for both keys and string values, which properly handles backslashes, quotes, and control characters. The default case also uses `json.Marshal` for proper escaping.

internal/execute/tsc.go
```diff
@@ -19,10 +19,36 @@ import (
...
+	}
+	configFilePath := ""
+	if config.ConfigFile != nil && config.ConfigFile.SourceFile != nil {
+		configFilePath = config.ConfigFile.SourceFile.FileName()
+	}
+	tr, err := tracing.StartTracing(sys.FS(), traceDir, configFilePath, config.CompilerOptions().SingleThreaded.IsTrue())
+	if err != nil {
+		fmt.Fprintf(sys.Writer(), "Warning: Failed to start tracing: %v\n", err)
+	}
+	return tr
```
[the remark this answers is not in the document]
dimitropoulos: Changed `startTracingIfNeeded` to accept the `testing` parameter and use `testing != nil` as the deterministic flag instead of `SingleThreaded.IsTrue()`. Deterministic output is now only enabled in test mode, keeping real timestamps for normal CLI usage.

internal/checker/types.go
dimitropoulos: Acknowledged — looking forward to the automated generation PR. Happy to rebase on top of it once it lands.

internal/compiler/fileloader.go
dimitropoulos: Added nil pre-checks to all tracing call sites in fileloader.go (6 sites), filesparser.go (2 sites), program.go (2 sites), and emitter.go (3 sites). When `Tracing` is nil (the common case), the `map[string]any{...}` literals are no longer allocated.

internal/compiler/program.go
dimitropoulos: Removed `Program.Tracing()` entirely. The only caller was `EmitFilesAndReportErrors`, which now receives tracing through a new `Tracing` field on `EmitInput`. Both callers in tsc.go already had `tr` in scope and now pass it directly.

internal/tracing/tracing.go
dimitropoulos: For `writeArgsJSON`, I've switched to using `json.Marshal` for string escaping, which fixes the correctness issue. For the broader trace event output (`writeEventTo`), the manual string building is intentional for performance — the values there are all controlled (phase names, event names) with no user content that needs escaping. I'm open to switching the full event builder to use `jsontext.Encoder` in a follow-up if you'd prefer, but the manual approach avoids the encoder overhead for the hot path.

dimitropoulos:
Fixed — sampled events (the "X" events with wall-clock timestamps shown in the diff) are now skipped entirely in deterministic mode. Only B/E pairs with the deterministic counter remain, which produces stable baselines.

internal/tracing/tracing.go
jakebailey: I would be very surprised if this manual building were faster than a direct encoder. It's also a lot more error prone, IMO. I would recommend switching.

jakebailey: especially since you both string build _and_ use json.Marshal? It's a weird combo

internal/execute/incremental/program.go
```diff
@@ -273,6 +274,9 @@ func (p *Program) collectSemanticDiagnosticsOfAffectedFiles(ctx context.Context,
 }
 
 func (p *Program) emitBuildInfo(ctx context.Context, options compiler.EmitOptions) *compiler.EmitResult {
+	if tr := tracing.FromContext(ctx); tr != nil {
```
jakebailey, line 277: Seems like you could just stick tracing on `Program`? I guess that's what you just undid, hm...

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+type LineAndChar struct {
+	Line      int `json:"line"`
+	Character int `json:"character"`
+}
+
+// escapeSymbolName converts internal symbol names to a JSON-safe format.
+// The Go codebase uses \xFE as a prefix for internal symbol names (which is invalid UTF-8),
+// while TypeScript uses "___" (three underscores). This function converts the Go format
+// to "__" (two underscores) for JSON output, matching TypeScript's unescapeLeadingUnderscores behavior.
+func escapeSymbolName(name string) string {
```
jakebailey, line 580: This helper should move to the ast package and then be reused (we have a copy of this in the test baseline output system).

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+		} else {
+			id := int32(-1)
+			desc.ConditionalFalseType = &id
+		}
+	}
+
+	// Substitution type
+	if baseType := typ.SubstitutionBaseType(); baseType != nil {
+		id := baseType.Id()
+		desc.SubstitutionBaseType = &id
```
jakebailey, line 684: for these, just use `new(baseType.Id())`, etc.

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+func getLocation(node *ast.Node) *Location {
+	if node == nil {
+		return nil
+	}
+	file := ast.GetSourceFileOfNode(node)
+	if file == nil {
+		return nil
+	}
+
+	startLine, startChar := scanner.GetECMALineAndUTF16CharacterOfPosition(file, node.Pos())
```
jakebailey, line 771: node.Pos includes trivia; is that intended (I haven't read the original code).

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+	defer t.mu.Unlock()
+	t.types = append(t.types, typ)
+}
+
+func (t *typeTracer) DumpTypes() error {
+	// Copy the types slice under lock, then release so Display() calls during
+	// buildTypeDescriptor don't deadlock when they create new types
+	t.mu.Lock()
+	types := make([]TracedType, len(t.types))
+	copy(types, t.types)
```
jakebailey, line 495: `slices.Clone`

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+	}
+}
+
+// NewTypeTracer creates a new tracer for a specific checker.
+// The checkerIndex is used to create unique filenames for each checker's output.
+func (tr *Tracing) NewTypeTracer(checkerIndex int) Tracer {
+	tr.mu.Lock()
+	defer tr.mu.Unlock()
+
+	tracePath := tspath.CombinePaths(tr.traceDir, "trace_0.json")
```
jakebailey, line 414: `0` is hardcoded here?

internal/tracing/tracing.go
```diff
@@ -0,0 +1,785 @@
...
+	var ts float64
+	if tr.eventStack[len(tr.eventStack)-1].separateBeginAndEnd {
+		ts = tr.timestamp()
+	}
+	popFrom(&tr.traceContent, &tr.eventStack, ts, tr.startTime, tr.deterministic)
+}
+
+// PopAll closes all open trace events on the shared trace buffer.
+// Safe to call on nil receiver.
+func (tr *Tracing) PopAll() {
```
jakebailey, line 394: This is unused

jakebailey (commented):
I regret my context suggestion, now having read this entirely... If we already have to do goofiness in the pool to know what number each checker is, then it can't really be a context thing, right? Unless we make some sort of specific checker tracer and ID that we ensure we use when tracing inside the checker?

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. I switched the trace event path over to deterministic JSON marshaling as well, so this no longer mixes manual string building with `json.Marshal`.

internal/execute/incremental/program.go
dimitropoulos: [Dimitri's agent, here]. I kept tracing off `Program` here because this path already receives the tracing-bearing context directly, and the earlier `Program.Tracing()` accessor had just been removed in favor of passing tracing explicitly where it is needed.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Done. The escaping helper now lives in `internal/ast` and is reused from both tracing and the type/symbol baseline writer.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Done. These pointer fields now use `new(expr)` style directly, and the wrapper helpers are gone.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Good catch. I switched the start location to use `scanner.GetTokenPosOfNode(...)`, so the recorded range starts at the token rather than including leading trivia.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Done. This copy now uses `slices.Clone`.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Done. I pulled the trace filename into a shared constant instead of hardcoding `0` at the call site.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Removed.

dimitropoulos:
[Dimitri's agent, here]. I rebased this branch onto `origin/main`, addressed the remaining review feedback in focused follow-up commits, reran `npx hereby test`, `npx hereby lint`, `npx hereby build`, and `npx hereby format`, and then force-pushed the rebased branch.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Done. I pulled the trace filename into a shared constant instead of hardcoding `0` at the call site.

internal/tracing/tracing.go
dimitropoulos: [Dimitri's agent, here]. Removed.

dimitropoulos:
[Dimitri's agent, here]. I rebased this branch onto `origin/main`, addressed the remaining review feedback in focused follow-up commits, reran `npx hereby test`, `npx hereby lint`, `npx hereby build`, and `npx hereby format`, and then force-pushed the rebased branch.

internal/ast/symbol.go
```diff
@@ -74,3 +75,8 @@ func SymbolName(symbol *Symbol) string {
 	}
 	return symbol.Name
 }
+
+// EscapeAllInternalSymbolNames replaces internal symbol name markers ("\xFE") with "__".
+func EscapeAllInternalSymbolNames(name string) string {
```
gabritto, line 80: Why do we need this? Right now we set `AllowInvalidUTF8` so that we replace \xFE with the replacement character when encoding to json.

internal/checker/relater.go
```diff
@@ -1333,6 +1337,9 @@ func (c *Checker) getAliasVariances(symbol *ast.Symbol) []VarianceFlags {
 func (c *Checker) getVariancesWorker(symbol *ast.Symbol, typeParameters []*Type) []VarianceFlags {
 	links := c.varianceLinks.Get(symbol)
 	if links.variances == nil {
+		if tr := c.tracer; tr != nil {
```
gabritto, line 1341: We used to also record variance flags in Strada. Will that information be missed?

internal/ast/symbol.go
jakebailey: I think it would be weird for trace analyzers to have to show a unicode replacement characters in their UI; it seems better to have this match our own type baselines.

internal/checker/relater.go
jakebailey: I suspect it's not there because `pop` doesn't accept args; it's probably doable

jakebailey: Pushed in a change to add this quick.

gabritto (approved):

PR #2926, opened 2026-02-27
Implement ES decorator transform (ESNext -> ES2022)

### LINKED ISSUE #2354: Implement esnext->es2022 transformer (ES decorators, `using` declarations)
Support `--target es2022` by transforming ES decorators. Note that the other necessary transform (`using` declarations) is done.

Start at `newESDecoratorTransformer`.

AlCalzone:
This PR implements the decorator transform portion of https://github.com/microsoft/typescript-go/issues/2354

This is the only thing blocking me from adopting `tsgo`, so I thought I'd help out by porting the original PR.

**Disclaimer:** This is an entire day worth of Claude Opus tokens. I don't claim I understand how it works, but I did my best to direct Claude to minimize the baseline diffs, of which it removed quite a few.

Fixes #2354

### DIFF
--- a/internal/ls/signaturehelp.go
+++ b/internal/ls/signaturehelp.go
@@ -2,7 +2,6 @@ package ls
 
 import (
 	"context"
-	"fmt"
 	"slices"
 	"strings"
 
@@ -775,9 +774,7 @@ func getContainingArgumentInfo(node *ast.Node, sourceFile *ast.SourceFile, check
 	for n := node; !ast.IsSourceFile(n) && (isManuallyInvoked || !ast.IsBlock(n)); n = n.Parent {
 		// If the node is not a subspan of its parent, this is a big problem.
 		// There have been crashes that might be caused by this violation.
-		if !RangeContainsRange(n.Parent.Loc, n.Loc) {
-			debug.Fail(fmt.Sprintf("Not a subspan. Child: %s, parent: %s", n.KindString(), n.Parent.KindString()))
-		}
+		debug.Assert(RangeContainsRange(n.Parent.Loc, n.Loc), "Not a subspan. Child: ", n.KindString(), ", parent: ", n.Parent.KindString())
 		argumentInfo := getImmediatelyContainingArgumentOrContextualParameterInfo(n, position, sourceFile, checker)
 		if argumentInfo != nil {
 			return argumentInfo
--- a/internal/transformers/estransforms/esdecorator.go
+++ b/internal/transformers/estransforms/esdecorator.go
@@ -1,8 +1,6 @@
 package estransforms
 
 import (
-	"fmt"
-
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
@@ -193,18 +191,14 @@ func (tx *esDecoratorTransformer) enterClass(ci *classInfo) {
 }
 
 func (tx *esDecoratorTransformer) exitClass() {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindClass {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'class' but got '%d' instead.", tx.top.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindClass, "Incorrect value for top.kind. Expected top.kind to be 'class' but got '", tx.top.kind, "' instead.")
 	tx.pendingExpressions = tx.top.savedPendingExpressions
 	tx.top = tx.top.next
 	tx.updateState()
 }
 
 func (tx *esDecoratorTransformer) enterClassElement(node *ast.Node) {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindClass {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'class' but got '%d' instead.", tx.top.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindClass, "Incorrect value for top.kind. Expected top.kind to be 'class' but got '", tx.top.kind, "' instead.")
 	tx.top = &lexicalEntry{
 		kind: lexicalEntryKindClassElement,
 		next: tx.top,
@@ -219,20 +213,14 @@ func (tx *esDecoratorTransformer) enterClassElement(node *ast.Node) {
 }
 
 func (tx *esDecoratorTransformer) exitClassElement() {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindClassElement {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'class-element' but got '%d' instead.", tx.top.kind))
-	}
-	if tx.top.next == nil || tx.top.next.kind != lexicalEntryKindClass {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.next.kind. Expected top.next.kind to be 'class' but got '%d' instead.", tx.top.next.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindClassElement, "Incorrect value for top.kind. Expected top.kind to be 'class-element' but got '", tx.top.kind, "' instead.")
+	debug.Assert(tx.top.next != nil && tx.top.next.kind == lexicalEntryKindClass, "Incorrect value for top.next.kind. Expected top.next.kind to be 'class' but got '", tx.top.next.kind, "' instead.")
 	tx.top = tx.top.next
 	tx.updateState()
 }
 
 func (tx *esDecoratorTransformer) enterName() {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindClassElement {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'class-element' but got '%d' instead.", tx.top.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindClassElement, "Incorrect value for top.kind. Expected top.kind to be 'class-element' but got '", tx.top.kind, "' instead.")
 	tx.top = &lexicalEntry{
 		kind: lexicalEntryKindName,
 		next: tx.top,
@@ -241,9 +229,7 @@ func (tx *esDecoratorTransformer) enterName() {
 }
 
 func (tx *esDecoratorTransformer) exitName() {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindName {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'name' but got '%d' instead.", tx.top.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindName, "Incorrect value for top.kind. Expected top.kind to be 'name' but got '", tx.top.kind, "' instead.")
 	tx.top = tx.top.next
 	tx.updateState()
 }
@@ -264,9 +250,7 @@ func (tx *esDecoratorTransformer) enterOther() {
 }
 
 func (tx *esDecoratorTransformer) exitOther() {
-	if tx.top == nil || tx.top.kind != lexicalEntryKindOther {
-		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'other' but got '%d' instead.", tx.top.kind))
-	}
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindOther, "Incorrect value for top.kind. Expected top.kind to be 'other' but got '", tx.top.kind, "' instead.")
 	if tx.top.depth > 0 {
 		debug.Assert(len(tx.pendingExpressions) == 0)
 		tx.top.depth--
--- a/internal/debug/assert.go
+++ b/internal/debug/assert.go
@@ -7,11 +7,11 @@ import (
 	"reflect"
 )
 
-func Assert(expression bool, message ...string) {
+func Assert(expression bool, message ...any) {
 	if !expression {
 		var msg string
 		if len(message) > 0 {
-			msg = "False expression: " + message[0]
+			msg = "False expression: " + fmt.Sprint(message...)
 		} else {
 			msg = "False expression."
 		}
@@ -33,11 +33,11 @@ func isNil[T any](value T) bool {
 	}
 }
 
-func AssertNil(value any, message ...string) {
+func AssertNil(value any, message ...any) {
 	if value != nil && !isNil(value) {
 		var msg string
 		if len(message) > 0 {
-			msg = "Nil expression: " + message[0]
+			msg = "Nil expression: " + fmt.Sprint(message...)
 		} else {
 			msg = "Nil expression."
 		}
@@ -45,97 +45,97 @@ func AssertNil(value any, message ...string) {
 	}
 }
 
-func AssertEqual(a fmt.Stringer, b fmt.Stringer, message ...string) {
+func AssertEqual(a fmt.Stringer, b fmt.Stringer, message ...any) {
 	if a != b {
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(fmt.Sprintf("Expected %s == %s. %s", a.String(), b.String(), msg))
 	}
 }
 
-func AssertLessThan(a int, b int, message ...string) {
+func AssertLessThan(a int, b int, message ...any) {
 	if a >= b {
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(fmt.Sprintf("Expected %d < %d. %s", a, b, msg))
 	}
 }
 
-func AssertLessThanOrEqual(a int, b int, message ...string) {
+func AssertLessThanOrEqual(a int, b int, message ...any) {
 	if a > b {
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(fmt.Sprintf("Expected %d <= %d. %s", a, b, msg))
 	}
 }
 
-func AssertGreaterThan(a int, b int, message ...string) {
+func AssertGreaterThan(a int, b int, message ...any) {
 	if a <= b {
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(fmt.Sprintf("Expected %d > %d. %s", a, b, msg))
 	}
 }
 
-func AssertGreaterThanOrEqual(a int, b int, message ...string) {
+func AssertGreaterThanOrEqual(a int, b int, message ...any) {
 	if a < b {
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(fmt.Sprintf("Expected %d >= %d. %s", a, b, msg))
 	}
 }
 
-func AssertIsDefined(value any, message ...string) {
+func AssertIsDefined(value any, message ...any) {
 	if value == nil || isNil(value) { // handle all `nil` interfaces
 		var msg string
 		if len(message) == 0 {
 			msg = ""
 		} else {
-			msg = message[0]
+			msg = fmt.Sprint(message...)
 		}
 		Fail(msg)
 	}
 }
 
-func CheckDefined[T any](value T, message ...string) T {
+func CheckDefined[T any](value T, message ...any) T {
 	AssertIsDefined(value, message...)
 	return value
 }
 
-func AssertEachIsDefined[TElem any](value []TElem, message ...string) {
+func AssertEachIsDefined[TElem any](value []TElem, message ...any) {
 	for _, elem := range value {
 		AssertIsDefined(elem, message...)
 	}
 }
 
-func CheckEachIsDefined[TElem any](value []TElem, message ...string) []TElem {
+func CheckEachIsDefined[TElem any](value []TElem, message ...any) []TElem {
 	AssertEachIsDefined(value, message...)
 	return value
 }
 
-var unexpectedNode []string = []string{"Unexpected node."}
+var unexpectedNode []any = []any{"Unexpected node."}
 
-func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, message ...string) {
+func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, message ...any) {
 	if len(message) == 0 {
 		message = unexpectedNode
 	}
@@ -144,7 +144,7 @@ func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, messag
 	}
 }
 
-func AssertNode[TElem any](node TElem, test func(elem TElem) bool, message ...string) {
+func AssertNode[TElem any](node TElem, test func(elem TElem) bool, message ...any) {
 	if len(message) == 0 {
 		message = unexpectedNode
 	}
@@ -154,7 +154,7 @@ func AssertNode[TElem any](node TElem, test func(elem TElem) bool, message ...st
 	}
 }
 
-func AssertNotNode[TElem any](node TElem, test func(elem TElem) bool, message ...string) {
+func AssertNotNode[TElem any](node TElem, test func(elem TElem) bool, message ...any) {
 	if isNil(node) {
 		return
 	}
@@ -167,7 +167,7 @@ func AssertNotNode[TElem any](node TElem, test func(elem TElem) bool, message ..
 	Assert(!test(node), message...)
 }
 
-func AssertOptionalNode[TElem any](node TElem, test func(elem TElem) bool, message ...string) {
+func AssertOptionalNode[TElem any](node TElem, test func(elem TElem) bool, message ...any) {
 	if isNil(node) {
 		return
 	}
@@ -180,7 +180,7 @@ func AssertOptionalNode[TElem any](node TElem, test func(elem TElem) bool, messa
 	Assert(test(node), message...)
 }
 
-func AssertOptionalToken[TElem interface{ KindValue() int16 }](node TElem, kind int16, message ...string) {
+func AssertOptionalToken[TElem interface{ KindValue() int16 }](node TElem, kind int16, message ...any) {
 	if isNil(node) {
 		return
 	}
@@ -190,7 +190,7 @@ func AssertOptionalToken[TElem interface{ KindValue() int16 }](node TElem, kind 
 	Assert(node.KindValue() == kind, message...)
 }
 
-func AssertMissingNode[TElem any](node TElem, message ...string) {
+func AssertMissingNode[TElem any](node TElem, message ...any) {
 	if len(message) == 0 {
 		message = unexpectedNode
 	}
--- a/internal/debug/noassert.go
+++ b/internal/debug/noassert.go
@@ -6,21 +6,21 @@ import (
 	"fmt"
 )
 
-func Assert(expression bool, message ...string)                                               {}
-func AssertNil(value any, message ...string)                                                  {}
-func AssertEqual(a fmt.Stringer, b fmt.Stringer, msg ...string)                               {}
-func AssertLessThan(a int, b int, message ...string)                                          {}
-func AssertLessThanOrEqual(a int, b int, message ...string)                                   {}
-func AssertGreaterThan(a int, b int, message ...string)                                       {}
-func AssertGreaterThanOrEqual(a int, b int, message ...string)                                {}
-func AssertIsDefined(value any, message ...string)                                            {}
-func CheckDefined[T any](value T, message ...string) T                                        { return value }
-func AssertEachIsDefined[TElem any](value []TElem, message ...string)                         {}
-func CheckEachIsDefined[TElem any](value []TElem, message ...string) []TElem                  { return value }
-func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, message ...string)  {}
-func AssertNode[TElem any](node TElem, test func(elem TElem) bool, message ...string)         {}
-func AssertNotNode[TElem any](node TElem, test func(elem TElem) bool, message ...string)      {}
-func AssertOptionalNode[TElem any](node TElem, test func(elem TElem) bool, message ...string) {}
-func AssertOptionalToken[TElem interface{ KindValue() int }](node TElem, kind int, message ...string) {
+func Assert(expression bool, message ...any)                                               {}
+func AssertNil(value any, message ...any)                                                  {}
+func AssertEqual(a fmt.Stringer, b fmt.Stringer, msg ...any)                               {}
+func AssertLessThan(a int, b int, message ...any)                                          {}
+func AssertLessThanOrEqual(a int, b int, message ...any)                                   {}
+func AssertGreaterThan(a int, b int, message ...any)                                       {}
+func AssertGreaterThanOrEqual(a int, b int, message ...any)                                {}
+func AssertIsDefined(value any, message ...any)                                            {}
+func CheckDefined[T any](value T, message ...any) T                                        { return value }
+func AssertEachIsDefined[TElem any](value []TElem, message ...any)                         {}
+func CheckEachIsDefined[TElem any](value []TElem, message ...any) []TElem                  { return value }
+func AssertEachNode[TElem any](nodes []TElem, test func(elem TElem) bool, message ...any)  {}
+func AssertNode[TElem any](node TElem, test func(elem TElem) bool, message ...any)         {}
+func AssertNotNode[TElem any](node TElem, test func(elem TElem) bool, message ...any)      {}
+func AssertOptionalNode[TElem any](node TElem, test func(elem TElem) bool, message ...any) {}
+func AssertOptionalToken[TElem interface{ KindValue() int }](node TElem, kind int, message ...any) {
 }
-func AssertMissingNode[TElem any](node TElem, message ...string) {}
+func AssertMissingNode[TElem any](node TElem, message ...any) {}
--- a/internal/debug/shared.go
+++ b/internal/debug/shared.go
@@ -14,22 +14,22 @@ func Fail(reason string) {
 	panic(reason)
 }
 
-func FailBadSyntaxKind(node interface{ KindString() string }, message ...string) {
+func FailBadSyntaxKind(node interface{ KindString() string }, message ...any) {
 	var msg string
 	if len(message) == 0 {
 		msg = "Unexpected node."
 	} else {
-		msg = message[0]
+		msg = fmt.Sprint(message...)
 	}
 	Fail(fmt.Sprintf("%s\r\nNode %s was unexpected.", msg, node.KindString()))
 }
 
-func AssertNever(member any, message ...string) {
+func AssertNever(member any, message ...any) {
 	var msg string
 	if len(message) == 0 {
 		msg = "Illegal value:"
 	} else {
-		msg = message[0]
+		msg = fmt.Sprint(message...)
 	}
 	var detail string
 	if member, ok := member.(interface{ KindString() string }); ok {


internal/printer/helpers.go
```diff
@@ -241,6 +241,56 @@ var AdvancedAsyncSuperHelper = &EmitHelper{
...
+    var kind = contextIn.kind, key = kind === "getter" ? "get" : kind === "setter" ? "set" : "value";
+    var target = !descriptorIn && ctor ? contextIn["static"] ? ctor : ctor.prototype : null;
+    var descriptor = descriptorIn || (target ? Object.getOwnPropertyDescriptor(target, contextIn.name) : {});
+    var _, done = false;
+    for (var i = decorators.length - 1; i >= 0; i--) {
+        var context = {};
+        for (var p in contextIn) context[p] = p === "access" ? {} : contextIn[p];
+        for (var p in contextIn.access) context.access[p] = contextIn.access[p];
+        context.addInitializer = function (f) { if (done) throw new TypeError("Cannot add initializers after decoration has completed"); extraInitializers.push(accept(f || null)); };
+        var result = (0, decorators[i])(kind === "accessor" ? { get: descriptor.get, set: descriptor.set } : descriptor[key], context);
```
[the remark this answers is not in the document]
AlCalzone: that's 1:1 the helper implementation from Typescript 5.9

internal/transformers/moduletransforms/esmodule.go
```diff
@@ -73,7 +73,7 @@ func (tx *ESModuleTransformer) visitSourceFile(node *ast.SourceFile) *ast.Node {
 		prologue, rest := tx.Factory().SplitStandardPrologue(result.Statements.Nodes)
 		statements := slices.Clone(prologue)
 		if externalHelpersImportDeclaration != nil {
-			statements = append(statements, externalHelpersImportDeclaration)
+			statements = append(statements, tx.Visitor().VisitNode(externalHelpersImportDeclaration))
```
weswigham, line 81: Strada does have a visit here, so I can see why a tool would add it, but surely it's unrelated to the desired change. And probably not requried, since in strada it was likely used to map es imports to amd/umd/cjs ones, meanwhile in corsa it gets to just stay as the es import it's generated as, since cjs is handled by an entirely different transform.

internal/transformers/estransforms/namedevaluation.go
```diff
@@ -168,7 +168,7 @@ func getAssignedNameOfPropertyName(emitContext *printer.EmitContext, name *ast.P
 		return assignedName, name
 	}
 
-	if !ast.IsComputedPropertyName(expression) {
+	if !ast.IsComputedPropertyName(name) {
```
weswigham: This looks like a correct driveby fix of this helper, but we should go further - since we have it, this should be a `debug.Assert` now, instead of a raw `panic`.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	case lexicalEntryKindClass:
+		tx.classInfoStack = tx.top.classInfoData
+	case lexicalEntryKindClassElement:
+		if tx.top.next != nil {
+			tx.classInfoStack = tx.top.next.classInfoData
+		}
+		tx.classThis = tx.top.classThisData
+		tx.classSuper = tx.top.classSuperData
+	case lexicalEntryKindName:
+		// name -> class-element -> class -> ???
```
weswigham: This comment is `???` - not in the original source, doesn't explain anything. Should be removed.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		next:                    tx.top,
+		classInfoData:           ci,
+		savedPendingExpressions: tx.pendingExpressions,
+	}
+	tx.pendingExpressions = nil
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) exitClass() {
+	tx.pendingExpressions = tx.top.savedPendingExpressions
```
weswigham, line 195: This function is missing a `debug.Assert` call that asserts that the `kind` of `tx.top` is `lexicalEntryKindClass`.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	tx.top = tx.top.next
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) enterClassElement(node *ast.Node) {
+	entry := &lexicalEntry{
```
weswigham: Likewise, an assert is missing here, too.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		kind: lexicalEntryKindClassElement,
+		next: tx.top,
+	}
+	if ast.IsClassStaticBlockDeclaration(node) || (ast.IsPropertyDeclaration(node) && ast.HasStaticModifier(node)) {
```
weswigham: What's with the parens around the later half of the condition here? Seems unintentional. Strada doesn't have them, for sure, I don't see why go'd need them.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		if tx.top != nil && tx.top.classInfoData != nil {
+			entry.classThisData = tx.top.classInfoData.classThis
+			entry.classSuperData = tx.top.classInfoData.classSuper
+		}
```
weswigham, line 210: This should have an `else` to `nil` these two fields in the entry to match strada behavior, but it might not matter? I wouldn't want the context to leak accidentally though.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	}
+	tx.top = entry
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) exitClassElement() {
+	tx.top = tx.top.next
```
weswigham, line 218: Once more, this is missing assertions from strada.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) enterName() {
+	tx.top = &lexicalEntry{
```
weswigham, line 224: Here, too.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		kind: lexicalEntryKindName,
+		next: tx.top,
+	}
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) exitName() {
+	tx.top = tx.top.next
```
weswigham, line 233: And here.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) enterOther() {
+	if tx.top != nil && tx.top.kind == lexicalEntryKindOther {
+		tx.top.depth++
```
weswigham, line 240: This is missing a `debug.Assert(len(tx.pendingExpressions) == 0)`

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			next:                    tx.top,
+			savedPendingExpressions: tx.pendingExpressions,
+		}
+		tx.pendingExpressions = nil
+		tx.updateState()
+	}
+}
+
+func (tx *esDecoratorTransformer) exitOther() {
+	if tx.top.depth > 0 {
```
weswigham, line 254: Likewise, another few missing asserts.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		tx.top.depth--
+	} else {
+		tx.pendingExpressions = tx.top.savedPendingExpressions
+		tx.top = tx.top.next
+		tx.updateState()
+	}
+}
+
+// --- Visitor dispatch ---
```
weswigham: These "logically grouping" comments weren't in strada and we don't need them here, since it's not like the order is hard and fast anyway - they're very.... C# folding-range-y.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		(tx.classThis != nil && tx.classSuper != nil && node.SubtreeFacts()&ast.SubtreeContainsLexicalSuper != 0)
+}
+
+func (tx *esDecoratorTransformer) visit(node *ast.Node) *ast.Node {
+	if !tx.shouldVisitNode(node) {
+		return node
+	}
+	switch node.Kind {
+	case ast.KindDecorator:
+		return nil
```
weswigham, line 316: This should have its missing `debug.Fail` added so it's not a quiet failure.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	return node.SubtreeFacts()&ast.SubtreeContainsDecorators != 0 ||
+		(tx.classThis != nil && node.SubtreeFacts()&ast.SubtreeContainsLexicalThis != 0) ||
+		(tx.classThis != nil && tx.classSuper != nil && node.SubtreeFacts()&ast.SubtreeContainsLexicalSuper != 0)
+}
+
+func (tx *esDecoratorTransformer) visit(node *ast.Node) *ast.Node {
+	if !tx.shouldVisitNode(node) {
+		return node
+	}
+	switch node.Kind {
```
weswigham, line 311: This switch is missing `Fail`s on `Constructor`s, `PropertyDeclaration`s, and `ClassStaticBlockDeclaration`s.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+}
+
+// --- Helper utilities ---
+
+func isDecoratedClassLike(node *ast.Node) bool {
+	return ast.ClassOrConstructorParameterIsDecorated(false, node) ||
+		ast.ChildIsDecorated(false, node, nil)
+}
+
+func moveRangePastDecorators(node *ast.Node) core.TextRange {
```
weswigham: I'm confident this is a duplicate of a helper elsewhere already in the `go` codebase. Likely in the non-standard decorator transform. Extract it to a common util location and reuse it, don't reimplement it.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			lastDecorator = core.FindLast(nodes, ast.IsDecorator)
+		}
+	}
+	if lastDecorator != nil && !ast.PositionIsSynthesized(lastDecorator.End()) {
+		return core.NewTextRange(lastDecorator.End(), node.End())
+	}
+	return node.Loc
+}
+
+func moveRangePastModifiers(node *ast.Node) core.TextRange {
```
weswigham: Likewise with this, this helper is already in the codebase.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	}
+	if lastModifier != nil && !ast.PositionIsSynthesized(lastModifier.End()) {
+		return core.NewTextRange(lastModifier.End(), node.End())
+	}
+	return moveRangePastDecorators(node)
+}
+
+func getHelperVariableName(node *ast.Node) string {
+	declarationName := ""
+	if node.Name() != nil && ast.IsIdentifier(node.Name()) {
```
weswigham: This condition is missing an exception for generated identifiers.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		declarationName = node.Name().Text()
+	} else if node.Name() != nil && ast.IsPrivateIdentifier(node.Name()) {
```
weswigham: As is this one.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			ast.NodeFlagsLet,
+			tx.Factory().NewNodeList([]*ast.Node{
+				tx.Factory().NewVariableDeclaration(name, nil, nil, initializer),
+			}),
+		),
+	)
+}
+
+// getAllDecoratorsOfClass returns the decorators for a class-like declaration (ES decorators).
+func getAllDecoratorsOfClass(node *ast.Node) []*ast.Node {
```
weswigham: Just delete this helper. Use `node.Decorators()` instead.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	return node.Decorators()
+}
+
+// getAllDecoratorsOfClassElement returns the decorators for a class element (ES decorators).
+func getAllDecoratorsOfClassElement(member *ast.Node, parent *ast.Node) []*ast.Node {
```
weswigham: Hold up, this is already in the legacydecorators transform - and more faithfully to strada, too - extract it and reuse it.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	innerExpression := ast.SkipOuterExpressions(expression, ast.OEKAll)
+	if ast.IsAccessExpression(innerExpression) {
+		target, thisArg := tx.createCallBinding(expression)
+		bindCall := tx.Factory().NewFunctionBindCall(target, thisArg, nil)
+		return tx.Factory().RestoreOuterExpressions(expression, bindCall, ast.OEKAll)
+	}
+	return expression
+}
+
+// createCallBinding is a simplified version of the factory's createCallBinding.
```
weswigham: This comment says nothing in the context of corsa (where there is no factory createCallBinding to "simplify") and doesn't refer to strada. Just remove it.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+func (tx *esDecoratorTransformer) createCallBinding(expression *ast.Expression) (target *ast.Expression, thisArg *ast.Expression) {
+	callee := ast.SkipOuterExpressions(expression, ast.OEKAll)
+	if ast.IsSuperProperty(callee) {
+		return callee, tx.Factory().NewThisExpression()
+	}
+	if callee.Kind == ast.KindSuperKeyword {
+		return callee, tx.Factory().NewThisExpression()
+	}
+	if ast.IsPropertyAccessExpression(callee) {
```
weswigham: This is missing the `EFHelperName` branch from strada.

weswigham: (Which is important because even though it matches the fallthrough case, its' earlier priority blocks `tslib.__helperName` references from triggering the property access case unnecessarily)

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			target.Loc = callee.Loc
+			return target, thisArg
+		}
+		return callee, ea.Expression
+	}
+	return expression, tx.Factory().NewVoidZeroExpression()
+}
+
+func (tx *esDecoratorTransformer) shouldBeCapturedInTempVariable(node *ast.Expression) bool {
+	// Capture identifiers that are not simple (i.e., not just a name or "this")
```
weswigham: In Strada, this function is.... honestly, entirely different. What in the world is going on here?

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	f := tx.Factory()
+	ci := &classInfo{
+		class: node,
+		metadataReference: f.NewUniqueNameEx("_metadata", printer.AutoGenerateOptions{
+			Flags: printer.GeneratedIdentifierFlagsOptimistic | printer.GeneratedIdentifierFlagsFileLevel,
+		}),
+	}
+
+	// If the class itself is decorated, create a _classThis binding
+	if ast.HasDecorators(node) && ast.NodeCanBeDecorated(false, node, nil, nil) {
```
weswigham: Just use `ast.nodeIsDecorated` - you'd have to capitalize it to make it public, since it wasn't needed outside other util functions until now.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	}
+
+	if body == nil {
+		body = tx.Visitor().VisitNode(ctor.Body.AsNode())
+	}
+	tx.exitClassElement()
+	return tx.Factory().UpdateConstructorDeclaration(ctor, modifiers, nil, parameters, nil, nil, body)
+}
+
+func (tx *esDecoratorTransformer) transformConstructorBodyWorker(statementsOut *[]*ast.Statement, statementsIn []*ast.Statement, statementOffset int, superPath []int, superPathDepth int, initializerStatements []*ast.Statement) {
```
weswigham: It's a bit more idiomatic, imo, to swap this to returning a statement list, rather than modifying the input list like strada did, given how go slices behave.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	superStatementIndex := superPath[superPathDepth]
+	// Visit statements before super
+	if superStatementIndex > statementOffset {
+		stmtVisitor := tx.EmitContext().NewNodeVisitor(tx.visit)
```
weswigham: ...wat. Why is this making a new visitor clone and not just using `tx.Visitor()`. Below, too. Seems like it got stuck on a very off pattern here.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			*statementsOut = append(*statementsOut, stmtVisitor.VisitNode(s))
+		}
+	}
+}
+
+// --- finishClassElement ---
+
+func (tx *esDecoratorTransformer) finishClassElement(updated *ast.Node, original *ast.Node) *ast.Node {
+	if updated != original {
+		tx.EmitContext().AssignCommentRange(updated, original)
```
weswigham, line 1221: The comment in the original code is likely worth preserving, so as to alleviate any concerns as to why the two ranges do not match.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+}
+
+type createDescriptorFunc func(member *ast.Node, modifiers *ast.ModifierList) *ast.Expression
+
+func (tx *esDecoratorTransformer) partialTransformClassElement(member *ast.Node, ci *classInfo, createDescriptor createDescriptorFunc) partialResult {
+	f := tx.Factory()
+	ec := tx.EmitContext()
+
+	if ci == nil {
+		modVisitor := ec.NewNodeVisitor(tx.modifierVisitor)
```
weswigham: Making this on every class element is wasteful - a single visitor in the transform root used everywhere needed should suffice. This pattern is repeated, too - all these extra sub-visitor creations on node visit aughta get removed and replaced with references to either the main `tx.Visitor()` or a `tx.modififersVisitor` (renaming the impl to `tx.modifiersVisitorFunc`) and the like for the other small visitors. (`fallbackVisitor`, `classElementVisitor`, etc)

This pervasive anitpattern makes me wanna toss an assert into `NewNodeVisitor` that throws if it's ever called during an ongoing transformation, but it's generally not a mistake a human makes...

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			if ast.IsStatic(member) {
+				statements = &ci.staticFieldDecorationStatements
+			} else {
+				statements = &ci.nonStaticFieldDecorationStatements
+			}
+		}
+
+		// Determine decorator kind
+		kind := ""
+		switch {
```
weswigham, line 1284: This switch is missing a `debug.Fail` on unknown kinds.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		} else if ast.IsPropertyDeclaration(member) {
+			mi.memberInitializersName = tx.createHelperVariable(member, "initializers")
+			mi.memberExtraInitializersName = tx.createHelperVariable(member, "extraInitializers")
+			result.initializersName = mi.memberInitializersName
+			result.extraInitializersName = mi.memberExtraInitializersName
+			if ast.IsStatic(member) {
+				result.thisArg = ci.classThis
+			}
+
+			ctorArg := (*ast.Node)(nil)
```
weswigham: This is non-idiomatic afaik (at least from what I've seen of our codebase) - it should be `var ctorArg *ast.Node` instead. I'm surprised the linter doesn't express a preference.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		memberDecoratorsAssignment := f.NewAssignmentExpression(memberDecoratorsName, memberDecoratorsArray)
+		mi := &memberInfo{memberDecoratorsName: memberDecoratorsName}
+		if ci.memberInfos == nil {
+			ci.memberInfos = make(map[*ast.Node]*memberInfo)
+		}
+		ci.memberInfos[member] = mi
+		tx.pendingExpressions = append(tx.pendingExpressions, memberDecoratorsAssignment)
+
+		// Determine which decoration statement bucket to use
+		var statements *[]*ast.Statement
```
weswigham: Hm. I'm not terribly comfortable with using pointers to handle field selection like this, but I suppose we do actually want to mutate the member - minimally, we need a safety `debug.Fail` when we don't set `statements`, though (as in strada), so we don't try to deference `nil` later on on the offchance something changes and we fall into that case. I'd almost prefer moving the target list selection to the end, though, and doing non-pointer-based `ci.staticNonFieldDecorationStatements = append(ci.staticNonFieldDecorationStatements, esDecorateStatement)` and the like instead. Wouldn't be too verbose if extracted to a helper.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+
+func (tx *esDecoratorTransformer) visitPropertyDeclaration(node *ast.Node) *ast.Node {
+	if isNamedEvaluation(tx.EmitContext(), node) && isAnonymousClassNeedingAssignedName(node.Initializer()) {
+		node = transformNamedEvaluation(tx.EmitContext(), node, false, "")
+	}
+
+	tx.enterClassElement(node)
+	f := tx.Factory()
+	ec := tx.EmitContext()
+
```
weswigham, line 1590: Like elsewhere, this is missing a `debug.Assert` - in this case, an assert that this transform doesn't support `declare`'d props.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+func (tx *esDecoratorTransformer) visitClassStaticBlockDeclaration(node *ast.Node) *ast.Node {
+	tx.enterClassElement(node)
+	f := tx.Factory()
+
+	var result *ast.Node
+	if isClassNamedEvaluationHelperBlock(tx.EmitContext(), node) {
+		result = tx.Visitor().VisitEachChild(node)
+		// Transfer AssignedName metadata to the new node so isClassNamedEvaluationHelperBlock
+		// can still find it after visiting (visiting may create a new node when this->_classThis)
+		if assignedName := tx.EmitContext().AssignedName(node); assignedName != nil && result != node {
```
weswigham: This reeks of being an unneeded diff from strada that's covering up another mistake elsewhere. Strada didn't need to manually copy this metadata, especially in what's basically a no-op branch.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		savedClassThis := tx.classThis
+		tx.classThis = nil
+		result = tx.Visitor().VisitEachChild(node)
+		tx.classThis = savedClassThis
+	} else {
+		result = tx.Visitor().VisitEachChild(node)
+		if tx.classInfoStack != nil {
+			tx.classInfoStack.hasStaticInitializers = true
+			if len(tx.classInfoStack.pendingStaticInitializers) > 0 {
+				stmts := []*ast.Statement{}
```
weswigham:
```suggestion
				stmts := []*ast.Node{}
```
and just `.AsNode()` the factory call, rather than converting the slice for no reason. This is kinda nonsensical as-is.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		return tx.classThis
+	}
+	return node
+}
+
+func (tx *esDecoratorTransformer) visitCallExpression(node *ast.Node) *ast.Node {
+	call := node.AsCallExpression()
+	if ast.IsSuperProperty(call.Expression) && tx.classThis != nil {
+		expression := tx.Visitor().VisitNode(call.Expression)
+		argVisitor := tx.EmitContext().NewNodeVisitor(tx.visit)
```
weswigham: I'm going to stop calling out these mistakenly immediately constructed visitors at this point. They're all wrong and should use a premade object.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	return tx.Visitor().VisitEachChild(node)
+}
+
+// --- Simple pass-through visitors that handle NamedEvaluation ---
+
+func (tx *esDecoratorTransformer) visitParameterDeclaration(node *ast.Node) *ast.Node {
+	if isNamedEvaluation(tx.EmitContext(), node) && isAnonymousClassNeedingAssignedName(node.Initializer()) {
+		node = transformNamedEvaluation(tx.EmitContext(), node, false, "")
+	}
+	return tx.Visitor().VisitEachChild(node)
```
weswigham, line 1867: This is pretty different from strada, which manually updates/visits the parameter, moves a bunch of spans on the result, and seemingly intentionally elides all parameter modifiers (including decorators). They are technically an error, but it'd be bad to leak them into the output.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		superProperty.Loc = ea.Expression.Loc
+		return superProperty
+	}
+	return tx.Visitor().VisitEachChild(node)
+}
+
+// --- Simple pass-through visitors that handle NamedEvaluation ---
+
+func (tx *esDecoratorTransformer) visitParameterDeclaration(node *ast.Node) *ast.Node {
+	if isNamedEvaluation(tx.EmitContext(), node) && isAnonymousClassNeedingAssignedName(node.Initializer()) {
```
weswigham: This is not strictly the same as the strada logic - by passing `isAnonymousClassNeedingAssignedName` as a callback to `isNamedEvaluation`, it's also calling `skipOuterExpressions` on the initializer in strada. That aughta be replicated here.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+			right = tx.Visitor().VisitNode(bin.Right)
+		}
+		return f.UpdateBinaryExpression(bin, nil, left, nil, bin.OperatorToken, right)
+	}
+
+	return tx.Visitor().VisitEachChild(node)
+}
+
+func (tx *esDecoratorTransformer) visitPropertyAssignment(node *ast.Node) *ast.Node {
+	if isNamedEvaluation(tx.EmitContext(), node) && isAnonymousClassNeedingAssignedName(node.Initializer()) {
```
weswigham: See my previous comment on the `isNamedEvaluation` call and `skipOuterExpressions` for here and the following functions, too.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		node = transformNamedEvaluation(tx.EmitContext(), node, false, "")
```
weswigham: pretty sure this `false` should be a call to `canIgnoreEmptyStringLiteralInAssignedName`, likewise elsewhere.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		return tx.visitComputedPropertyName(node)
+	}
+	return tx.Visitor().VisitNode(node)
+}
+
+func (tx *esDecoratorTransformer) visitReferencedPropertyName(node *ast.Node) (referencedName *ast.Expression, name *ast.Node) {
+	if ast.IsPropertyNameLiteral(node) || ast.IsPrivateIdentifier(node) {
+		referencedName = tx.Factory().NewStringLiteralFromNode(node)
+		name = tx.Visitor().VisitNode(node)
+		return
```
weswigham: Personally, I am not a fan of using named returns like this when a direct return will suffice - obfuscates things for no reason.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+
+	key := tx.Factory().NewPropKeyHelper(tx.Visitor().VisitNode(cpn.Expression))
+	assignment := tx.Factory().NewAssignmentExpression(referencedName, key)
+	updatedName := tx.Factory().UpdateComputedPropertyName(cpn, tx.injectPendingExpressions(assignment))
+	return referencedName, updatedName
+}
+
+// --- Pending expressions injection ---
+
+func (tx *esDecoratorTransformer) injectPendingExpressions(expression *ast.Expression) *ast.Expression {
```
weswigham: This and the following function, `injectPendingInitializers`, shared a common core helper function for their mostly shared functionality (of chaining expressions) - that common helper aughta be restored, because this repetition feels bad.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	var staticMods []*ast.Node
+	for _, mod := range modifiers.Nodes {
+		if mod.Kind == ast.KindStaticKeyword {
+			staticMods = append(staticMods, mod)
+		}
+	}
+	if len(staticMods) == 0 {
+		return nil
+	}
+	return tx.Factory().NewModifierList(staticMods)
```
weswigham: To fully replicate the functionality of the visitor this replaces, this need a `if len(staticMods) == len(modifiers.Nodes) { return modifiers }` so it's not overwriting lists it doesn't need to. 

Or it should be reimplemented as a visitor.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		ce := node.AsClassExpression()
+		updatedNode = f.UpdateClassExpression(ce, ce.Modifiers(), ce.Name(), nil, ce.HeritageClauses, membersList)
+	}
+	ec.SetClassThis(updatedNode, classThis)
+	return updatedNode
+}
+
+// --- findSuperStatementIndexPath (duplicated from tstransforms) ---
+
+func findSuperStatementIndexPath(statements []*ast.Statement, start int) []int {
```
weswigham: This is already in `runtimesyntax.go` - extract and reuse it.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+				return append([]int{i}, path...)
+			}
+		}
+	}
+	return nil
+}
+
+// --- Slice conversion helpers ---
+
+func expressionsToNodes(exprs []*ast.Expression) []*ast.Node {
```
weswigham: All of these slice-casting helpers should be removed and different casting should be used at their use sites instead, they're a waste of memory. There's no situation where we need a slice of `Statement`s, other than local iteration, where we can just do `.AsStatement()` on the underlying `Node`s instead.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+		expression.Loc = node.Loc
+	}
+
+	return expression
+}
+
+// --- Utility: createAssignmentTargetWrapper ---
+// Creates: ({ set value(_p) { Reflect.set(target, key, _p, receiver) } }).value
+
+func createAssignmentTargetWrapper(f *printer.NodeFactory, target *ast.Expression, propertyKey *ast.Expression, receiver *ast.Expression) *ast.Expression {
```
weswigham: This name is misleading since it also builds in the `Reflect.set` call that was previously done by callers. It's more like `createReflectedAssignmentTargetWrapper` instead.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	case ast.KindQuestionQuestionEqualsToken:
+		return ast.KindQuestionQuestionToken
+	default:
+		return ast.KindUnknown
+	}
+}
+
+// --- Utility: isCompoundAssignment ---
+
+func isCompoundAssignment(kind ast.Kind) bool {
```
weswigham: This duplicates a function in `checker` - extract it to `ast` and capitalize it so it's usable in both places.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -2,18 +2,2679 @@ package estransforms
...
+	recordTempVariable func(*ast.IdentifierNode),
+	resultVariable *ast.Expression,
+) *ast.Expression {
+	var operator ast.Kind
+	if ast.IsPrefixUnaryExpression(node) {
+		operator = node.AsPrefixUnaryExpression().Operator
+	} else {
+		operator = node.AsPostfixUnaryExpression().Operator
+	}
+
```
weswigham: This is missing a debug assert on non-`++` or `--` ops from strada.

weswigham (changes requested):
There's a pervasive issue with missing assertions and immediately-constructed-visitors that needs fixing throughout, but aside from that there's not *too* much obviously wrong here.

There's also a lot of useful spec text comments that got dropped in translation - those aughta get copied over, too - they're good reference material.

AlCalzone:
Thanks for the thorough review - I'll see what I can do about it.

internal/transformers/estransforms/namedevaluation.go
```diff
@@ -59,7 +60,10 @@ func classHasDeclaredOrExplicitlyAssignedName(emitContext *printer.EmitContext,
 // the Identifier `__proto__`, or the string literal `"__proto__"`, but not for
 // computed property names.
 func isProtoSetter(node *ast.PropertyName) bool {
-	return ast.IsIdentifier(node) || ast.IsStringLiteral(node) && node.Text() == "__proto__"
```
jakebailey: This was just a typo and should have been:

```ts
return (ast.IsIdentifier(node) || ast.IsStringLiteral(node)) && node.Text() == "__proto__"
```

Which I would prefer

AlCalzone:
@weswigham @jakebailey I think I've addressed everything now.

jakebailey:
I want to wait for #2956 before looking into this; there is some overlap and that one is basically complete.

AlCalzone:
Cool, let me know when that PR is done and I'll see what I need to change here.

testdata/baselines/reference/submodule/compiler/declarationEmitCommonJsModuleReferencedType.errors.txt.diff
```diff
@@ -0,0 +1,16 @@
+--- old.declarationEmitCommonJsModuleReferencedType.errors.txt
++++ new.declarationEmitCommonJsModuleReferencedType.errors.txt
+@@= skipped -0, +0 lines =@@
+-r/entry.ts(3,14): error TS2742: The inferred type of 'x' cannot be named without a reference to 'foo/node_modules/nested'. This is likely not portable. A type annotation is necessary.
++r/entry.ts(3,14): error TS2883: The inferred type of 'x' cannot be named without a reference to 'NestedProps' from 'foo/node_modules/nested'. This is likely not portable. A type annotation is necessary.
```
jakebailey: I think there might be some changes leftover from a merge from main?

AlCalzone:
Fixed two more issues:
1. ES decorators were silently skipped when target=esnext + useDefineForClassFields=false
2. `export default @dec class {}` was getting `__setFunctionName(this, "")` instead of `__setFunctionName(this, "default")`

And updated the submodule to avoid the above diff.

internal/transformers/utilities.go
```diff
@@ -317,56 +370,3 @@ func MoveRangePastModifiers(node *ast.Node) core.TextRange {
...
-		return ast.KindCaretToken
-	case ast.KindBarBarEqualsToken:
-		return ast.KindBarBarToken
-	case ast.KindAmpersandAmpersandEqualsToken:
-		return ast.KindAmpersandAmpersandToken
-	case ast.KindQuestionQuestionEqualsToken:
-		return ast.KindQuestionQuestionToken
-	}
-	return kind
-}
```
jakebailey, line 372: Restore the ordering here from main; the diff is not needed.

testdata/baselines/reference/submodule/conformance/esDecorators-classDeclaration-accessors-staticPrivate(target=es2015).js.diff
```diff
@@ -50,15 +28,25 @@
...
-+class C {
-+}
-+_a = C, _C_method1_get = function _C_method1_get() { return 0; }, _C_method1_set = function _C_method1_set(value) { };
-+class D {
-+}
-+_b = D, _D_method1_get = function _D_method1_get() { return 0; }, _D_method1_set = function _D_method1_set(value) { };
-+(() => {
-+    __classPrivateFieldGet(_b, _b, "a", _D_method1_get);
-+    __classPrivateFieldSet(_b, _b, 1, "a", _D_method1_set);
++    var D = (_a = class {
```
jakebailey: These diffs indicate that there's still more to do here; likely, these are the `!!!` I left in the class fields transformer that only work once we have the ES Decorators transform available. Those will have to be implemented.

testdata/baselines/reference/submodule/compiler/temporal.types.diff
```diff
@@ -98,7 +98,43 @@
...
          billion.toZonedDateTimeISO("UTC"),
  >billion.toZonedDateTimeISO("UTC") : Temporal.ZonedDateTime
-@@= skipped -124, +124 lines =@@
+@@= skipped -38, +38 lines =@@
+     // Round to a particular unit
+     instant.round({ smallestUnit: "second" }); // => 2019-03-30T02:46:00Z
+ >instant.round({ smallestUnit: "second" }) : Temporal.Instant
+->instant.round : { (roundTo: Temporal.TimeUnit): Temporal.Instant; (roundTo: Temporal.RoundingOptions<Temporal.TimeUnit>): Temporal.Instant; }
++>instant.round : { (roundTo: Temporal.PluralizeUnit<Temporal.TimeUnit>): Temporal.Instant; (roundTo: Temporal.RoundingOptions<Temporal.TimeUnit>): Temporal.Instant; }
+ >instant : Temporal.Instant
```
jakebailey: This branch is still pretty out of date with main.

FWIW, I typically do this:

```
hereby test || (hereby baseline-accept; hereby test || hereby baseline-accept)
```

Which is, run the tests, if they fail, accept the baselines, run again, and if they fail again (due to extra baselines), delete them.

internal/parser/parser.go
```diff
@@ -6550,6 +6550,60 @@ func (p *Parser) jsErrorAtRange(loc core.TextRange, message *diagnostics.Message
...
+		for _, modifier := range modifiers {
+			if ast.IsDecorator(modifier) {
+				p.jsErrorAtRange(modifier.Loc, diagnostics.Decorators_are_not_valid_here)
+				break
+			}
+		}
+	} else if ast.CanHaveDecorators(node) {
+		decoratorIndex := core.FindIndex(modifiers, ast.IsDecorator)
+		if decoratorIndex >= 0 {
+			if ast.IsClassDeclaration(node) {
```
jakebailey, line 6569: There's a missing check here for parameters without experimentalDecorators, which is a shame since we can't really add it back in again as we do not have options here. But I think we still have to complain about it somehow.

We do at least error on them in grammar checks, so maybe this all isn't needed? But, we still want to complain somehow for JS files... Terrible situation.

internal/transformers/estransforms/definitions.go
```diff
@@ -25,8 +25,13 @@ func GetESTransformer(opts *transformers.TransformOptions) *transformers.Transfo
 	options := opts.CompilerOptions
 	switch options.GetEmitScriptTarget() {
 	case core.ScriptTargetESNext:
-		// At ESNext, only the class fields transformer is needed (it self-gates via shouldTransformAnything).
 		// The TS reference always runs transformClassFields unconditionally.
+		// When useDefineForClassFields is false and experimentalDecorators is not set,
+		// we also need to run the ES decorator transform, matching the TS reference:
+		//   if (!experimentalDecorators && (target < ESNext || !useDefineForClassFields))
+		if !options.ExperimentalDecorators.IsTrue() && !options.GetUseDefineForClassFields() {
+			return transformers.Chain(newESDecoratorTransformer, newClassFieldsTransformer)(opts)
```
jakebailey: We should lift this, since this chain can be reused.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2727 @@
...
+		next:                    tx.top,
+		classInfoData:           ci,
+		savedPendingExpressions: tx.pendingExpressions,
+	}
+	tx.pendingExpressions = nil
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) exitClass() {
+	debug.Assert(tx.top != nil && tx.top.kind == lexicalEntryKindClass, "Incorrect value for top.kind.", fmt.Sprintf("Expected top.kind to be 'class' but got '%d' instead.", tx.top.kind))
```
jakebailey: We shouldn't be doing asserts like this, as it pays the fmt cost up front.

testdata/baselines/reference/submodule/compiler/temporal.types.diff
AlCalzone: Looks I merged an old state of `main` the first time. This should be correct now.

internal/transformers/estransforms/definitions.go
AlCalzone: I'm not entirely sure how. Should this be the `NewES2022Transformer`?

But there's also this comment above which I'm not sure how to solve:
```
// !!! TODO: This fixed layering scheme assumes you can't swap out the es decorator transform for the legacy one,
// or the proper es class field transform for the legacy one
```

AlCalzone: Or should I align `GetESTransformer` more with the Strada codebase which constructs the transformer chain on the fly?

weswigham:
You can just have the esDecorator transformer do nothing when
experimentalDecorators is set for now (which should just be an
optimization, actually, since the experimental decorator transformer will
have already removed all the decorators before it runs). The class fields
transformer also has options logic within it for now; we can move that all
out at once later.

testdata/baselines/reference/submodule/conformance/esDecorators-classDeclaration-accessors-staticPrivate(target=es2015).js.diff
AlCalzone: Should also be resolved now

jakebailey:
Not sure where you're at with this, but there are a bunch of diffs still. I personally use https://gist.github.com/jakebailey/65f2e053dbf75c2fb40c149e7ff1848f to bulk mark an entire PR as read except for the still-modified diffs, then hide viewed files in the PR, and then look at diffs that aren't deleted yet. There's a bunch still in there.

I'll try and go 1:1 to compare (though I don't know if you have already reordered funcs to match the original), but that might get you something to look at in the meantime.

AlCalzone:
Yeah, I'm looking at those right now. 👍🏻

testdata/baselines/reference/submodule/conformance/decoratedClassExportsCommonJS1.js.diff
```diff
@@ -1,19 +1,9 @@
...
  exports.Testing123 = Testing123;
--Testing123.prop1 = Testing123_1.prop0;
--exports.Testing123 = Testing123 = Testing123_1 = __decorate([
-+Testing123.prop1 = Testing123.prop0;
-+exports.Testing123 = Testing123 = __decorate([
-     Something({ v: () => Testing123 })
+ Testing123.prop1 = Testing123_1.prop0;
+ exports.Testing123 = Testing123 = Testing123_1 = __decorate([
+-    Something({ v: () => Testing123 })
++    Something({ v: () => Testing123_1 })
```
jakebailey: Odd diff, but I think your PR is adding back `_1` anyway, even though it's not needed in this case. this might still indicate scoping problems elsewhere where the wrong var is being used.

internal/printer/printer.go
```diff
@@ -1449,8 +1449,10 @@ func (p *Printer) emitParameterNode(node *ast.ParameterDeclarationNode) {
 }
 
 func (p *Printer) emitDecorator(node *ast.Decorator) {
+	state := p.enterNode(node.AsNode())
```
jakebailey, line 1452: Curious, Strada did not need this.

testdata/baselines/reference/submodule/conformance/usingDeclarationsWithESClassDecorators.2(module=esnext,target=es2015).js.diff
```diff
@@ -9,28 +9,4 @@
...
--    })();
-+    C = 
-+    @dec
-+    class C {
-+    };
- }
- catch (e_1) {
-     env_1.error = e_1;
\ No newline at end of file
+     before = __addDisposableResource(env_1, null, false);
```
jakebailey: Note to self: `before` exists, it's just above in the diff due to some difference in where temps are being emitted

jakebailey: there's a one-line fix for this; in `using.go` change:

```go
topLevelStatements = tx.EmitContext().EndAndMergeVariableEnvironment(topLevelStatements)
```

to 

```go
topLevelStatements = append(topLevelStatements, tx.EmitContext().EndVariableEnvironment()...)
```

matching

```ts
addRange(topLevelStatements, endLexicalEnvironment());
```

in strada.

But, that's a drive-by fix

jakebailey: It'll touch some 114 baselines; maybe that's not good to stick in this PR but it's not that bad to ignore. I could send it separately.

jakebailey:
@AlCalzone are you still wanting to finish this out, or should I just push fixes up? I still have yet to go side-by-side, but I don't know how much back and forth you're willing to do either.

(you're probably looking at the same bugs I was looking at in diffs0

jakebailey:
I think I have fixes for a lot of these, very annoying substitutions

AlCalzone:
If you have something, feel free to push it. Or take over completely, I don't mind.
Just let me know how you want to proceed.

internal/transformers/moduletransforms/esmodule.go
jakebailey: This appears to be required to ensure that the `import x = require("tslib")` turns into a `const x = require("tslib")`.

internal/printer/printer.go
jakebailey: I checked, and this is required for comment emit and source map reasons.

internal/parser/parser.go
jakebailey: I went and added an ad-hoc visitor just for this specific case, where Program used to check this.

internal/transformers/moduletransforms/esmodule.go
jakebailey: (Not just going off the comment; I checked this one)

jakebailey:
This should be ready for a re-review. I've fixed the comments (to my eye), reordered functions, then went 1:1 down and fixed any minor mismatches. There are no longer any `NewNodeVisitor` calls outside of construction, too.

jakebailey:
In any case, thanks for sending this @AlCalzone; most everything I pushed was little bits or thing that would have been a bit challenging to describe, but basically everything else seemed good to me 😄

AlCalzone:
Thanks for pushing it closer to the finish line!

internal/ls/signaturehelp.go
```diff
@@ -775,7 +775,9 @@ func getContainingArgumentInfo(node *ast.Node, sourceFile *ast.SourceFile, check
 	for n := node; !ast.IsSourceFile(n) && (isManuallyInvoked || !ast.IsBlock(n)); n = n.Parent {
 		// If the node is not a subspan of its parent, this is a big problem.
 		// There have been crashes that might be caused by this violation.
-		debug.Assert(RangeContainsRange(n.Parent.Loc, n.Loc), fmt.Sprintf("Not a subspan. Child: %s, parent: %s", n.KindString(), n.Parent.KindString()))
+		if !RangeContainsRange(n.Parent.Loc, n.Loc) {
```
weswigham: So, like, is this functionally equivalent? On the surface, yes, and defers string construction to only the fail case. In reality? No. `debug.Fail` persists into release builds, while assertions are elided, so the old formulation, despite "doing the fmt eagerly" it is (or should be) elided by compiler as the assert is itself elided, while this isn't.

If we wanna normalize defering formatting, we should do as strada does, and pass a closure, as in the `verboseDebugInfo` assert param in strada. Likely as a separate assert function, given the type restrictions we have.

(Also, why is this file involved in this PR at all, actually...?)

internal/transformers/estransforms/definitions.go
```diff
@@ -5,10 +5,9 @@ import (
 	"github.com/microsoft/typescript-go/internal/transformers"
 )
 
-// !!! TODO: This fixed layering scheme assumes you can't swap out the es decorator transform for the legacy one,
-// or the proper es class field transform for the legacy one
 var (
-	NewESNextTransformer = transformers.Chain(newESDecoratorTransformer, newUsingDeclarationTransformer, newClassFieldsTransformer)
+	esDecoratorAndClassFields = transformers.Chain(newESDecoratorTransformer, newClassFieldsTransformer)
```
weswigham, line 9: This is gonna be a little spicy when decorators stablize into a year, since implicitly class fields will have to move into the same one, but also be globally first for the legacy transform still. Future problem, though.

internal/transformers/moduletransforms/esmodule.go
```diff
@@ -75,7 +75,10 @@ func (tx *ESModuleTransformer) visitSourceFile(node *ast.SourceFile) *ast.Node {
 		statements := slices.Clone(prologue)
 		statements = append(statements, custom...)
 		if externalHelpersImportDeclaration != nil {
-			statements = append(statements, externalHelpersImportDeclaration)
+			// The helpers import must be visited so that `import x = require("tslib")`
+			// (TypeScript-only syntax) is transformed to `const x = require("tslib")`
+			// for CJS output files via visitImportEqualsDeclaration.
```
weswigham, line 80: Yeah... looking at it `createExternalHelpersImportDeclarationIfNeeded` makes an `import=` and leaves it to the later transform to handle when the file/module is commonjs, however since we can't have a commonjs file that's transformed to, eg, amd or umd anymore, we *really* should just be making the `const` in that helper, since there isn't a need for variable output anymore.

We *can* keep doing this roundabout downleveling, but we don't really have a reason not to generate the right nodes up-front.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2779 @@
...
+		classInfoData:           ci,
+		savedPendingExpressions: tx.pendingExpressions,
+	}
+	tx.pendingExpressions = nil
+	tx.updateState()
+}
+
+func (tx *esDecoratorTransformer) exitClass() {
+	if tx.top == nil || tx.top.kind != lexicalEntryKindClass {
+		debug.Fail(fmt.Sprintf("Incorrect value for top.kind. Expected top.kind to be 'class' but got '%d' instead.", tx.top.kind))
```
weswigham: Yeah... see my other comment on this assertion -> fail rewrite - it defers the string formatting but makes it so the conditional is preserved in release builds, so we really should have a closure-based (or perhaps even just specifically fmt-string-based) assertion function to get the best of both build targets.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2779 @@
...
+	}
+	if node.Kind == ast.KindSourceFile {
+		return tx.visitSourceFile(node.AsSourceFile())
+	}
+	if !tx.shouldVisitNode(node) {
+		return node
+	}
+	switch node.Kind {
+	case ast.KindDecorator:
+		// Decorators are elided. In Strada, a separate `fallbackVisitor` drops decorators
```
weswigham:
```suggestion
		// Decorators are elided. In Strada, a separate `modifierVisitor` drops decorators
```

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2779 @@
...
+		// before they reach `visitor` via visitEachChild. Here, `visit` serves as both
+		// visitor and fallback, so decorators from modifier lists reach it directly.
```
weswigham:
```suggestion
		// visitors, so decorators from modifier lists reach it directly.
```

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2779 @@
...
+			filtered = append(filtered, mod)
+		}
+	}
+	if len(filtered) == 0 {
+		return nil
+	}
+	if len(filtered) == len(modifiers.Nodes) {
+		return modifiers
+	}
+	return tx.Factory().NewModifierList(filtered)
```
weswigham: See all this length checking handling and list ensuring here? This is a visitor in disguise. It should probably be used as one (it is in strada) so all that is handled correctly by default. As-is, this will drop list locations from the filtered list. Upside: it's also much shorter as a simple `if node.Kind == ast.KindAccessorKeyword { return nil } return node`, since it's only used the once.

internal/transformers/estransforms/esdecorator.go
```diff
@@ -1,19 +1,2779 @@
...
```
weswigham: Likewise with `filterOutModifier`, this should probably be just a tiny visitor that preserves the `static` or `async` modifier, since that's the only things this ever filters for, and all this list handling is what visitors are here for.

internal/transformers/estransforms/esdecorator.go
jakebailey: I don't actually think these are very expensive; for things like this, the compiler can't optimize them away because those checks could cause nil panics anyway, so will already be branch-y.

internal/transformers/moduletransforms/esmodule.go
jakebailey: For a follow-up PR, then?

internal/ls/signaturehelp.go
jakebailey: This was definitely a drive by earlier, when I commented about the assert cost

internal/transformers/estransforms/esdecorator.go
jakebailey: For now, we could do `debug.Assert(false` just to get the not-fail?

PR #264, opened 2025-01-29
More type checking

ahejlsberg:
In this PR:

* Lots of `checkSourceElement` functions ported.
* Ported logic for collecting information for unused identifier checks.
* More accessors on `Node`.
* Several small fixes to porting errors.
* Prefer non-expando declaration for `ValueDeclaration`.
* Improved checking for duplicate properties.
* Improved expando property logic in binder.

PR #2934, opened 2026-02-27
fix(2932): handle parameter-property ref to method with same name

### LINKED ISSUE #2932: Panic on find-all-refs when a parameter property is preceded by a non-property member.
```ts
class C {
  x() {}
  constructor(public /*find*/x: number) {}
}
```

```
internal/ls.(*refState).forEachRelatedSymbol()
	internal/ls/findallreferences.go:2298
internal/ls.(*refState).getRelatedSymbol()
	internal/ls/findallreferences.go:2172
internal/ls.(*refState).getReferencesAtLocation()
	internal/ls/findallreferences.go:1869
internal/ls.(*refState).getReferencesInContainer()
	internal/ls/findallreferences.go:1810
internal/ls.(*refState).getReferencesInSourceFile()
	internal/ls/findallreferences.go:1798
internal/ls.(*refState).searchForName()
	internal/ls/findallreferences.go:2346
internal/ls.(*refState).getReferencesInContainerOrFiles()
	internal/ls/findallreferences.go:1791
internal/ls.getReferencedSymbolsForSymbol()
	internal/ls/findallreferences.go:1496
internal/ls.(*LanguageService).getReferencedSymbolsForNode()
	internal/ls/findallreferences.go:1014
internal/ls.(*LanguageService).getSemanticDocumentHighlights()
	internal/ls/documenthighlights.go:54
internal/ls.(*LanguageService).ProvideDocumentHighlights()
	internal/ls/documenthighlights.go:44
internal/lsp.(*Server).handleDocumentHighlight()
	internal/lsp/server.go:1264
internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].21.1()
	internal/lsp/server.go:750
internal/lsp.(*Server).handleRequestOrNotification.func1()
	internal/lsp/server.go:613
internal/lsp.(*Server).dispatchLoop.func3()
	internal/lsp/server.go:488
internal/lsp.(*Server).dispatchLoop
	internal/lsp/server.go:487
```

a-tarasyuk:
Fixes #2932


internal/fourslash/tests/findAllRefsParameterPropertyDeclaration_test.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func TestFindAllRefsParameterPropertyDeclaration(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `
+class C {
+  x() {}
+  constructor(public /**/x: number) {}
+}`
```
DanielRosenwasser: I found an example that also crashes on Strada when you do interface/class merging with a getter.

We should also ensure that these are testing find-all-references from both parameter and property usages.

So if you could change this to:


```suggestion
// @filename: c1.ts

class C1 {
  [|x|]() {}
  constructor(public [|x|]: number) {
    [|x|]++;
  }
}

new C1(1).[|x|];

// @filename: c2.ts

interface C2 {
  get [|x|](): void
}
class C2 {
  constructor(public [|x|]: number) {
    [|x|]++;
  }
}

new C2(1).[|x|];
`
```

it would be nice to have that extra coverage

internal/fourslash/tests/findAllRefsParameterPropertyDeclaration_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	f.VerifyBaselineFindAllReferences(t, "")
```
DanielRosenwasser:
```suggestion
	f.VerifyBaselineFindAllReferences(t, f.VerifyBaselineFindAllReferences(t, f.MarkerNames()...))
```

internal/fourslash/tests/findAllRefsParameterPropertyDeclaration_test.go
```diff
@@ -0,0 +1,22 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestFindAllRefsParameterPropertyDeclaration(t *testing.T) {
```
DanielRosenwasser:
```suggestion
func TestFindAllRefsParameterPropertyWithConflictingMember(t *testing.T) {
```

and rename the file as well

internal/ls/findallreferences.go
```diff
@@ -2289,15 +2289,17 @@ func (state *refState) forEachRelatedSymbol(
...
-			panic("Expected a parameter and a property")
+		if (paramProp1.Flags&ast.SymbolFlagsFunctionScopedVariable == 0) || (paramProp2.Flags&ast.SymbolFlagsClassMember == 0) {
+			return nil, entryKindNone
+		}
+		if res := fromRoot(paramProp1); res != nil {
+			return res, entryKindNode
+		}
+		if res := fromRoot(paramProp2); res != nil {
+			return res, entryKindNode
 		}
```
DanielRosenwasser: Isn't this different from the original logic of picking the "other" symbol based on the original root symbol?

internal/fourslash/tests/findAllRefsParameterPropertyWithConflictingMember_test.go
```diff
@@ -0,0 +1,38 @@
...
+  constructor(public [|x|]: number) {
+    [|x|]++;
+  }
+}
+new C2(1).[|x|];
+`
+
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	f.VerifyBaselineFindAllReferences(t, f.MarkerNames()...)
```
DanielRosenwasser: Actually this was accidentally left behind and it will be empty now. You can just change this to use Ranges instead (my preference) or drop it entirely (apparently it defaults to using ranges if you pass nothing).

internal/ls/findallreferences.go
```diff
@@ -2289,15 +2289,17 @@ func (state *refState) forEachRelatedSymbol(
...
-		return fromRoot(core.IfElse(symbol.Flags&ast.SymbolFlagsFunctionScopedVariable != 0, paramProp2, paramProp1)), entryKindNode
+		return nil, entryKindNode
```
DanielRosenwasser: This should be `entryKindNone`, right?

internal/ls/findallreferences.go
a-tarasyuk: @DanielRosenwasser, thanks for the additinal test cases. Strada uses `PropertyExcludes = None`, it was [changed](https://github.com/microsoft/typescript-go/pull/264/changes#diff-f76f93ecdcf7a9cf755c356f5b1ac21adcc4c3c632c2e7af2d4de0c67a879b61R58)

https://github.com/microsoft/typescript-go/blob/0a7d128ada55f48fdcf84296473d8b2f469605ff/internal/ast/symbolflags.go#L59

The symbol contains two declarations, and `valueDeclaration` points to the method declaration rather than the parameter. As a result, this [condition](https://github.com/microsoft/TypeScript/blob/924810c077dd41019ea92358af187261d9057f5d/src/services/findAllReferences.ts#L2598) is always false in the crash scenario.

In tsgo, however, `valueDeclaration` points to the parameter declaration, so the condition is evaluated and proceeds. To correctly handle parameter-property/member name collisions so FAR returns both related declarations.

internal/ls/findallreferences.go
a-tarasyuk: @DanielRosenwasser  I think this results in a “not found” state

https://github.com/microsoft/typescript-go/blob/0a7d128ada55f48fdcf84296473d8b2f469605ff/internal/ls/findallreferences.go#L2316-L2317

DanielRosenwasser:
It seems with the latest commits, we're no longer testing find-all-refs on usages as well.

a-tarasyuk:
@DanielRosenwasser, I updated test to use ranges. Did you mean this, or were you referring to adding the additional cases?

internal/ls/findallreferences.go
DanielRosenwasser: Right, but the old logic used to be "pick the property symbol if we're starting with a parameter symbol, otherwise pick the parameter symbol".

I guess the test case shows that this works the same as before - but I don't understand how.

DanielRosenwasser:
@a-tarasyuk you don't have to close the PR, I just wanted to spend a bit longer understanding it

internal/ls/findallreferences.go
DanielRosenwasser: Okay, I think I understand- let me know if it also makes sense to you.

There are two calls to `forEachRelatedSymbol`. In the first, `populateSearchSymbolSet` basically goes through all related symbols and adds them to an initial search set. The second pass is `getRelatedSymbol` where previously-seen symbols are added but thrown away.

So you're technically doing more work now, but you're still indirectly achieving the "one or the other" logic.

DanielRosenwasser: It would be nice to still boil this down to a single call though.

DanielRosenwasser:
Thank you!

Issue #2942, opened 2026-03-01
tsgo --api: DocumentIdentifier JSON unmarshaling panics


ajmacd:
I understand API development is in flux, but I was trying to play around with it and figured this crash was worth reporting.

## Stack trace

```
panic: invalid jsontext.Token; it has been voided by a subsequent json.Decoder call
goroutine 122 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:26 +0x64
github.com/microsoft/typescript-go/internal/api.(*AsyncConn).handleRequest.func1()
        github.com/microsoft/typescript-go/internal/api/conn_async.go:96 +0x44
panic({0x1020c86a0?, 0x102340750?})
        runtime/panic.go:860 +0x12c
github.com/go-json-experiment/json/jsontext.Token.string({{}, 0x352b53810568?, {0x0?, 0x352b53810500?}, 0x10112f6b0?})
        github.com/go-json-experiment/json@v0.0.0-20260214004413-d219187c3433/jsontext/token.go:251 +0x280
github.com/go-json-experiment/json/jsontext.Token.String({{}, 0x352b53810568?, {0x0?, 0x352b52ca9878?}, 0x10111c284?})
        github.com/go-json-experiment/json@v0.0.0-20260214004413-d219187c3433/jsontext/token.go:242 +0x24
github.com/microsoft/typescript-go/internal/api.(*DocumentIdentifier).UnmarshalJSONFrom(0x352b52af1380, 0x352b53810500)
        github.com/microsoft/typescript-go/internal/api/proto.go:193 +0x1a0
github.com/go-json-experiment/json.makeMethodArshaler.func7(0x352b53810500, {{0x1021ce020?, 0x352b52af1380?, 0x10218b180?}, 0x80?}, 0x352b538105c0)
        github.com/go-json-experiment/json@v0.0.0-20260214004413-d219187c3433/arshal_methods.go:323 +0x154
github.com/go-json-experiment/json.makeDefaultArshaler.makePointerArshaler.func10(0x352b53810500, {{0x10218b180?, 0x352b52db7420?, 0x30?}, 0x8?}, 0x352b538105c0)
        github.com/go-json-experiment/json@v0.0.0-20260214004413-d219187c3433/arshal_default.go:1706 +0x1a8
[...]

PR #2943, opened 2026-03-01
fix(api): DocumentIdentifier JSON unmarshaling panic

ajmacd:
## Summary
While playing around with the nascent API, I discovered a crash documented here: https://github.com/microsoft/typescript-go/issues/2942

This resolves the crash.

## Details
UnmarshalJSONFrom reads object keys and values as separate tokens via json.Decoder.ReadToken(). The second ReadToken call (for the value) voids the previous token (the key), so calling key.String() afterward panics with "invalid jsontext.Token; it has been voided by a subsequent json.Decoder call".

Fix by saving the key string to a local variable before reading the value token. Also add handling for the "fileName" field which was previously silently ignored.

### DIFF
--- a/internal/api/proto_test.go
+++ b/internal/api/proto_test.go
@@ -3,8 +3,8 @@ package api_test
 import (
 	"testing"
 
-	"github.com/go-json-experiment/json"
 	"github.com/microsoft/typescript-go/internal/api"
+	"github.com/microsoft/typescript-go/internal/json"
 	"gotest.tools/v3/assert"
 )
 
@@ -39,7 +39,7 @@ func TestDocumentIdentifierUnmarshalJSON(t *testing.T) {
 		{
 			name:  "invalid type",
 			input: `42`,
-			err:   "json: cannot unmarshal into Go api.DocumentIdentifier: DocumentIdentifier: expected string or object, got number",
+			err:   "expected string or object, got number",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


internal/api/proto.go
```diff
@@ -186,12 +186,16 @@ func (d *DocumentIdentifier) UnmarshalJSONFrom(dec *json.Decoder) error {
 			if err != nil {
 				return err
 			}
+			keyStr := key.String() // must save before next ReadToken voids the token
```
jakebailey: This works, though it does incur a string copy every time that this runs just to do a constant comparison later. Not clear how often this runs, however. It's possible that it's worth doing the comparisons before reading the next token just to avoid the copy.

ajmacd:

internal/api/proto.go
ajmacd: Agreed, doesn't hurt. It's no less readable this way and possibly faster.

ajmacd:
> There is no fileName field. On the client, a DocumentIdentifier is `string | { uri: string }`

Thanks for catching this. Overeager LLM trying to bend the API to the whims of its client.

I added a test, mainly for a bit more documentation, but let me know if you'd prefer I remove it.

andrewbranch (approved):
Thank you!

ajmacd:
Apologies, had to fix some CI failures. I think I was able to run everything successfully locally this time, so the latest commit should be good.

jakebailey (approved):

PR #2946, opened 2026-03-02
Improve CommonJS support, round 2

ahejlsberg:
With this PR, in a CommonJS module with an export assignment (`export =` in TS or `module.exports =` in JS) we now overlay type and uninstantiated namespace declarations in the module onto the namespace of the exported entity. An export assignment is in error if (a) the module exports value members or (b) if the module exports type or namespace members and the exported entity also exports type or namespace members.

With these changes, the following is now permitted:

```ts
// File mod.ts
export type Foo = number[];  // This declaration is overlayed onto the namespace of the `export =`
export = { greeting: "hello" };

// File main.ts
import mod = require("./mod");
const greeting = mod.greeting;
const x: mod.Foo = [1, 2, 3];
```

This PR furthermore aligns Corsa more closely with Strada as follows:

* Permit multiple `module.exports` assignments in a file and union the assigned types.
* Exclude `module.exports` assignments from stricter checking of TypeScript `export =` declarations.

This PR supersedes #2893.


testdata/tests/cases/compiler/exportAssignmentMerging1.ts
```diff
@@ -0,0 +1,14 @@
+// @module: commonjs
+// @declaration: true
```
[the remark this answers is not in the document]
ahejlsberg: No need, `// @strict: true` is the default in Corsa.

ahejlsberg:
@weswigham We already have an issue around declaration file generation for CommonJS modules with `module.exports` assignments. For example, this:

```js
const dummy = 0;
module.exports = dummy;
```

generates an invalid declaration file (an actual declaration of `dummy` is missing):

```js
export = dummy;
```

(EDIT: The above has now been fixed.)

Furthermore, with this PR we'll permit multiple `module.exports` assignments and non-top-level `module.exports` assignments. Neither is handled correctly currently. In particular, nested `module.exports` assignments are simply ignored.

internal/binder/binder.go
```diff
@@ -871,7 +871,7 @@ func (b *Binder) bindExportAssignment(node *ast.Node) {
 		}
 		// If there is an `export default x;` alias declaration, can't `export default` anything else.
 		// (In contrast, you can still have `export default function f() {}` and `export default interface I {}`.)
-		symbol := b.declareSymbol(ast.GetExports(container.Symbol()), container.Symbol(), node, flags, ast.SymbolFlagsAll)
+		symbol := b.declareSymbol(ast.GetExports(container.Symbol()), container.Symbol(), node, flags, core.IfElse(ast.IsJSExportAssignment(node), 0, ast.SymbolFlagsAll))
 		if ast.IsJSExportAssignment(node) || node.AsExportAssignment().IsExportEquals {
 			// Will be an error later, since the module already has other exports. Just make sure this has a valueDeclaration set.
```
DanielRosenwasser: I think this is no longer true for `JSExportAssignment`s, right?


```suggestion
			// This may be an error later, since the module already has other exports.
			// Just make sure this has a valueDeclaration set.
```

internal/binder/binder.go
```diff
@@ -1022,6 +1022,21 @@ func (b *Binder) bindDeferredExpandoAssignments() {
...
 
+// If the given module symbol has an export= symbol, promote exports with a type or namespace meaning
+// from the module symbol onto the export= symbol and, if any such exports exist, mark the export=
+// symbol as a namespace module.
+func (b *Binder) bindCommonJSTypeExports(moduleSymbol *ast.Symbol) {
+	moduleExports := moduleSymbol.Exports
+	if exportEquals := moduleExports[ast.InternalSymbolNameExportEquals]; exportEquals != nil {
+		for _, symbol := range moduleExports {
+			if symbol.Name != ast.InternalSymbolNameExportEquals && symbol.Flags&(ast.SymbolFlagsType|ast.SymbolFlagsNamespace) != 0 {
+				ast.GetExports(exportEquals)[symbol.Name] = symbol
```
DanielRosenwasser, line 1033: What happens if you write

```ts
export type blah = number;
export = {
  blah: "hello",
};
```

Are you overwriting the exports here?


Also, what happens if you alias another namespace?

```ts
export type blah = number;

namespace ns {
  export var blah = "hello";
}

export = ns;
```

ahejlsberg: Both work as expected, i.e. you can reference `blah` both as a type and a value through an import of the module.

internal/binder/binder.go
ahejlsberg: Yeah, will fix.

testdata/baselines/reference/submodule/conformance/jsDeclarationsTypedefAndImportTypes.js
```diff
@@ -86,3 +86,36 @@ declare const _default: {
...
 export = _default;
+
+
+//// [DtsFileErrors]
+
+
+out/conn.d.ts(5,10): error TS2304: Cannot find name 'Conn'.
+
+
+==== out/conn.d.ts (1 errors) ====
```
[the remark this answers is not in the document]
ahejlsberg: Now fixed.

ahejlsberg:
Latest commits fixes the issue in the example [here](https://github.com/microsoft/typescript-go/pull/2946#issuecomment-3981689855). However, as can be seen in [this test](https://github.com/microsoft/typescript-go/pull/2946/changes#diff-78e56d5260e0ae23fd34a9f44ec4944833cd1d5ff9edbef870cd901e7d4426e8), the declaration emitter doesn't properly handle nested `module.exports` assignments. The logic that handles that in Strada is located in `symbolTableToDeclarationStatements`, which is stubbed out in `nodebuilderimpl.go` in Corsa. I'm not going to attempt to fix that in this PR.

weswigham:
That's incorrect - that's a stub in corsa because it's unused, and should continue to be, since we don't do semantic js declaration emit, but syntactic emit, same as TS, now. If we want to support non-top-level export assignments in corsa, we need to add a non-top-level visitor to the declaration emitter transform - it's new logic, and should be part of the initial implementation of the checker logic, too - that's what I've been saying about how "some js checker work has included declaration emit logic and some hasn't, and we'll need to audit what's missing". We're not "waiting to port something" - the strategy is entirely different. Originally, the hope was we'd need no new declaration emitter logic since the reparser was going to handle all those transforms. If we back off that approach, that means you need declaration emitter logic to mimic the checker logic.

jakebailey:
I think we have been gradually backing off on that logic, as we discover more things that the reparser can't do (e.g., the defineProperties stuff).

ahejlsberg:
Just to recap, in the design meeting we agreed to merge this PR and create a separate PR to issue errors when generating declaration files for CommonJS modules with constructs that aren't supported by the emitter. Specifically, the following aren't supported when generating declaration files:

* Multiple `module.exports` assignments in a module.
* Nested `module.exports` assignments.
* Nested `exports.xxx` assignments.
* Nested `Object.defineProperty(module.exports, ...`) calls.

PR #2948, opened 2026-03-02
Don't re-do `merge_group` work on `push: main`

Andarist:
cc @jakebailey

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,6 @@
 name: CI
 
 on:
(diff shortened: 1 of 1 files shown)


jakebailey:
The jobs differ between PRs, merge queues, and pushes to main. I don't think this change makes sense as it means we will stop testing macOS entirely, for example.

What motivated this change?

Andarist:
I noticed a lot of work between those is being redone, but I have missed it's not an exact overlap.

PR #2949, opened 2026-03-02
Fix formatting panic when first line is all whitespace

Andarist:
fixes the crash found here https://github.com/microsoft/typescript-go/issues/2945#issuecomment-3981326595

### DIFF
--- a/internal/fourslash/_scripts/crashingTests.txt
+++ b/internal/fourslash/_scripts/crashingTests.txt
@@ -1,5 +1,4 @@
 TestFindReferencesBindingPatternInJsdocNoCrash1
 TestFindReferencesBindingPatternInJsdocNoCrash2
-TestFormatDocumentWithTrivia
 TestFormattingJsxTexts4
 TestQuickInfoBindingPatternInJsdocNoCrash1
--- a/internal/fourslash/_scripts/failingTests.txt
+++ b/internal/fourslash/_scripts/failingTests.txt
@@ -212,7 +212,6 @@ TestFormatConflictMarker1
 TestFormatDebuggerStatement
 TestFormatDocumentPreserveTrailingWhitespace
 TestFormatDocumentWithJSDoc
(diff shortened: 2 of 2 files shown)


internal/format/span.go
```diff
@@ -821,8 +821,10 @@ func (w *formatSpanWorker) trimTrailingWhitespacesForLines(line1 int, line2 int,
 
 		whitespaceStart := w.getTrailingWhitespaceStartPosition(lineStartPosition, lineEndPosition)
 		if whitespaceStart != -1 {
-			r, _ := utf8.DecodeRuneInString(w.sourceFile.Text()[whitespaceStart-1:])
-			debug.Assert(whitespaceStart == lineStartPosition || !stringutil.IsWhiteSpaceSingleLine(r))
+			if whitespaceStart != lineStartPosition {
```
Andarist, line 824: in Strada this case "slipped through" implicit casting, NaNs, and other beautiful stuff like it

internal/fourslash/tests/formatDocumentNoCrashLeadingWhitespace1_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestFormatDocumentNoCrashLeadingWhitespace1(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = " \n"
```
Andarist, line 13: the repro found in the wild was actually an empty file: https://github.com/motiondivision/motion/blob/d8b219ff3e7934e2077a5605c57239e53760636c/packages/motion-dom/src/animation/generators/spring/utils.ts

jakebailey (approved):

PR #2950, opened 2026-03-02
Fix formatting crash on files with tab indentation near end of file

Andarist:
fixes the crash found here: https://github.com/microsoft/typescript-go/issues/2945#issuecomment-3981326571

### DIFF
--- a/internal/format/span.go
+++ b/internal/format/span.go
@@ -890,7 +890,12 @@ func (w *formatSpanWorker) characterToColumn(startLinePosition int, characterInL
 }
 
 func (w *formatSpanWorker) indentationIsDifferent(indentationString string, startLinePosition int) bool {
-	return indentationString != w.sourceFile.Text()[startLinePosition:startLinePosition+len(indentationString)]
+	text := w.sourceFile.Text()
+	end := startLinePosition + len(indentationString)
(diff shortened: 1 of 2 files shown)


internal/format/span.go
```diff
@@ -890,7 +890,12 @@ func (w *formatSpanWorker) characterToColumn(startLinePosition int, characterInL
 }
 
 func (w *formatSpanWorker) indentationIsDifferent(indentationString string, startLinePosition int) bool {
-	return indentationString != w.sourceFile.Text()[startLinePosition:startLinePosition+len(indentationString)]
+	text := w.sourceFile.Text()
+	end := startLinePosition + len(indentationString)
+	if end > len(text) {
+		return true
+	}
+	return indentationString != text[startLinePosition:end]
```
jakebailey, line 898: Old code was:

```ts
return indentationString !== sourceFile.text.substr(startLinePosition, indentationString.length);
```

So, I guess this is just bounds again. I probably would have written this as:

```go
indentationString != text[startLinePosition:min(len(text), startLinePosition+len(indentationString))]
```

Or even:

```go
text = text[startLinePosition:]
indentationString != text[:min(len(text), len(indentationString))]
```

But the current state is fine.

PR #2952, opened 2026-03-02
Fixed a go-to-def crash on decorators attached to invalid targets

Andarist:
Fixes a crash discovered when investigating https://github.com/microsoft/typescript-go/issues/2945#issuecomment-3981326357 

However, I couldn't repro it using the provided tsreplay logs in tldraw. I'm not sure what's the difference, I also doubt they have a decorator on an invalid target. But either way, a crash is a crash - this simple repro shouldn't crash and it's a valid thing to fix.

Note: this also crashes in Strada 😉


internal/checker/checker.go
```diff
@@ -8582,6 +8582,9 @@ func (c *Checker) resolveTaggedTemplateExpression(node *ast.Node, candidatesOutA
 }
 
 func (c *Checker) resolveDecorator(node *ast.Node, candidatesOutArray *[]*Signature, checkMode CheckMode) *Signature {
+	if !ast.CanHaveDecorators(node.Parent) {
+		return c.resolveErrorCall(node)
+	}
```
Andarist, line 8585: in a regular typechecking scenario, `checkDecorators` validates the decorator's placement first and prevents `resolveDecorator` from being called at all. But an LSP request can lead to such a call by targeting the node more directly

internal/lsp/replay_test.go
```diff
@@ -78,6 +78,7 @@ func TestReplay(t *testing.T) {
 	defer f.Close()
 
 	scanner := bufio.NewScanner(f)
+	scanner.Buffer(make([]byte, 0, 64*1024), 10*1024*1024)
```
jakebailey, line 81: What is this here for?

Andarist: This helped process large replay logs coming from the tldraw repo

jakebailey:
CI is failing, unfortunately.

Andarist:
@jakebailey that was just caused by my last-minute test file renaming 😬 I fixed that, it should be good to go now

PR #2956, opened 2026-03-03
Port class fields transform (ES2022)

### LINKED ISSUE #2353: Implement es2023->es2022 transform (class fields and class static blocks)
Support `--target es2022` by transforming class static blocks  and class fields.

Start at `newClassStaticBlockTransformer` and `newClassFieldsTransformer`.

### LINKED ISSUE #2446: `tsgo` emits invalid code for decorated class methods with `Symbol` names
## Steps to reproduce

`tsgo` compiles the following code incorrectly. Note the method name in the class:

```ts
class A {
  @(fakeDecorator as any)
  [Symbol()]() {}
}

function fakeDecorator() {}
```

**tsconfig:**

```json
{
  "compilerOptions": {
    "module": "nodenext",
    "declaration": true,
    "declarationMap": true,
    "sourceMap": true,
    "composite": true,
    "noImplicitOverride": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "noImplicitReturns": true,
    "skipLibCheck": true,
    "target": "es2023",
    "strictPropertyInitialization": false,
    "noImplicitAny": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "removeComments": false,
  }
}
```

Valid code

## Behavior with `tsgo`

Emits a non-existent method name - `_a`. If you run the compiled code, it throws an error:

```text
], A.prototype, _a, null);
                ^

ReferenceError: _a is not defined
```

### LINKED ISSUE #2638: tsgo emits class field declarations for conditionally assigned fields that tsc doesn't
## Steps to reproduce

tsgo emits explicit class field declarations at the top of the class body, while tsc omits them. This results in different JavaScript output with observable runtime behavioral differences if there are conditionally initialized vaules when enumerating or iterating over properties.

Reproduction: https://github.com/psm14/tsgo-conditional-initialized-field-reproduction

```typescript
class MultipleOptional {
  required: string;
  optional1?: string;
  optional2?: number;

  constructor(name: string) {
    this.required = name;
  }
}
```

```typescript
class MultipleOptional {
  constructor(name) {
    this.required = name;
  }
}
```

## Behavior with `tsgo`

```typescript
class MultipleOptional {
  required;
  optional1;
  optional2;
  constructor(name) {
    this.required = name;
  }
}
```

jakebailey:
Includes #2925
Fixes #2638
Fixes #2353
Supersedes #1542
Fixes #2446
Supersedes #2447

This is a 1:1 port of the class fields transform. We had previously declared a class fields and static block transformer with the hope we'd split them, but I instead opted to just port things as is (maybe to split later).

Things in this PR:

- I deleted all of the `NodeCheckFlags` that were previously used in emit. These were never used, but were confusing to leave behind. There is a little more work needed sometimes to look for info, but nothing too challenging. Nice to not depend on the checker marking stuff. The places that are now mention "Strada".
- Some more work is needed to do things without node substitution, as per usual. Everything seems to match.
- The code related to decorators was ported best I could, but will require more work later I'm sure.
- I fixed the `UseDefineForClassFields` option code, which was misported (but not broken).
- There are a few misc emit fixes, which I noticed in diffs as the class field stuff disappeared.

I went through this 1:1 and aligned everything; it all should match; functions in the same order, same comments, etc.

### DIFF
--- a/internal/transformers/estransforms/classfields.go
+++ b/internal/transformers/estransforms/classfields.go
@@ -2517,7 +2517,7 @@ func (tx *classFieldsTransformer) transformConstructorBody(container *ast.Node, 
 	var multiLine bool
 	if constructor != nil && constructor.Body != nil &&
 		len(constructor.Body.AsBlock().Statements.Nodes) >= len(statements) {
-		multiLine = constructor.Body.AsBlock().Multiline || len(statements) > 0
+		multiLine = constructor.Body.AsBlock().Multiline
 	} else {
 		multiLine = len(statements) > 0
 	}


internal/transformers/moduletransforms/esmodule.go
```diff
@@ -71,7 +71,9 @@ func (tx *ESModuleTransformer) visitSourceFile(node *ast.SourceFile) *ast.Node {
 	externalHelpersImportDeclaration := createExternalHelpersImportDeclarationIfNeeded(tx.EmitContext(), result, tx.compilerOptions, tx.getEmitModuleFormatOfFile(node), false /*hasExportStarsToExportValues*/, false /*hasImportStar*/, false /*hasImportDefault*/)
 	if externalHelpersImportDeclaration != nil || tx.importRequireStatements != nil {
 		prologue, rest := tx.Factory().SplitStandardPrologue(result.Statements.Nodes)
+		custom, rest := tx.Factory().SplitCustomPrologue(rest)
 		statements := slices.Clone(prologue)
+		statements = append(statements, custom...)
 		if externalHelpersImportDeclaration != nil {
 			statements = append(statements, externalHelpersImportDeclaration)
 		}
```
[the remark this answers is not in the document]
jakebailey: This is weird but what Strada did, based on behavior. I can revert it.

internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -765,7 +776,13 @@ func (tx *RuntimeSyntaxTransformer) visitClassDeclaration(node *ast.ClassDeclara
 		modifiers = tx.Visitor().VisitModifiers(node.Modifiers())
 	}
 
+	// A class declaration without a name needs a generated name if it has static
+	// initialized properties, since those will be moved outside the class body by
+	// the classfields transformer and need to reference the class by name.
 	name := tx.Visitor().VisitNode(node.Name())
+	if name == nil && hasStaticInitializedProperties(node.Members.Nodes) {
+		name = tx.Factory().NewGeneratedNameForNode(node.AsNode())
+	}
```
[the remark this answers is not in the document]
jakebailey: Undoing this causes:

```
diff --git a/./testdata/baselines/reference/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js b/./testdata/baselines/local/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js
index ec7200b86..ba4e13e40 100644
--- a/./testdata/baselines/reference/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js
+++ b/./testdata/baselines/local/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js
@@ -6,7 +6,7 @@ export default class {
 }
 
 //// [exportDefaultClassWithStaticPropertyAssignmentsInES6.js]
-class default_1 {
+class {
 }
 default_1.z = "Foo";
 export default default_1;
diff --git a/./testdata/baselines/local/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js.diff b/./testdata/baselines/local/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js.diff
new file mode 100644
index 000000000..01b43e39d
--- /dev/null
+++ b/./testdata/baselines/local/submodule/conformance/exportDefaultClassWithStaticPropertyAssignmentsInES6.js.diff
@@ -0,0 +1,11 @@
+--- old.exportDefaultClassWithStaticPropertyAssignmentsInES6.js
++++ new.exportDefaultClassWithStaticPropertyAssignmentsInES6.js
+@@= skipped -5, +5 lines =@@
+ }
+
+ //// [exportDefaultClassWithStaticPropertyAssignmentsInES6.js]
+-class default_1 {
++class {
+ }
+ default_1.z = "Foo";
+ export default default_1;
\ No newline at end of file
```

Probably this is wrong and the name only works due to happenstance. Will revisit

testdata/baselines/reference/submodule/compiler/convertKeywordsYes(alwaysstrict=false).js.diff
```diff
@@ -7,121 +7,4 @@
 +"use strict";
```
jakebailey, line 7: Seems like we should skip alwaysStrict=false tests, and are missing an error.

testdata/baselines/reference/submodule/compiler/decoratorsOnComputedProperties.js.diff
```diff
@@ -6,44 +6,34 @@
...
-+    [Symbol.isConcatSpreadable];
-+    ["property4"] = 2;
-+    [Symbol.match] = null;
-+    [foo()];
-+    [foo()];
-+    [foo()] = null;
-+    [fieldNameA];
-+    [fieldNameB];
-+    [fieldNameC] = null;
++        this["property4"] = 2;
```
jakebailey: Not 100% certain what the deal is with this ordering.

testdata/baselines/reference/submodule/conformance/decoratorOnClassConstructorParameter5.js.diff
```diff
@@ -7,16 +7,14 @@
...
      constructor(_modeService) {
          this._modeService = _modeService;
      }
  };
--BulkEditPreviewProvider.Schema = 'vscode-bulkeditpreview';
+ BulkEditPreviewProvider.Schema = 'vscode-bulkeditpreview';
 -BulkEditPreviewProvider.emptyPreview = { scheme: BulkEditPreviewProvider_1.Schema };
 -BulkEditPreviewProvider = BulkEditPreviewProvider_1 = __decorate([
++BulkEditPreviewProvider.emptyPreview = { scheme: BulkEditPreviewProvider.Schema };
 +BulkEditPreviewProvider = __decorate([
```
jakebailey, line 18: This diff remains but is only here because `BulkEditPreviewProvider_1` is unused.

testdata/baselines/reference/submodule/conformance/decoratedClassExportsCommonJS1.js.diff
```diff
@@ -9,12 +9,11 @@
...
 -let Testing123 = Testing123_1 = class Testing123 {
 +let Testing123 = class Testing123 {
-+    static prop0;
-+    static prop1 = Testing123.prop0;
  };
  exports.Testing123 = Testing123;
 -Testing123.prop1 = Testing123_1.prop0;
 -exports.Testing123 = Testing123 = Testing123_1 = __decorate([
++Testing123.prop1 = Testing123.prop0;
 +exports.Testing123 = Testing123 = __decorate([
```
jakebailey, line 17: `Testing123_1` is unused.

testdata/baselines/reference/submodule/compiler/convertKeywordsYes(alwaysstrict=false).js.diff
jakebailey: https://github.com/microsoft/typescript-go/pull/2961

testdata/baselines/reference/submodule/conformance/autoAccessorDisallowedModifiers(target=esnext).js.diff
```diff
@@ -8,13 +8,4 @@
 +    accessor constructor() { }
```
jakebailey, line 8: I haven't been able to reliably fix this yet, but this is a disallowed case so there's probably some modifier stripping code that is not quite the same.

internal/parser/parser.go
```diff
@@ -637,7 +639,17 @@ func (p *Parser) parseDelimitedList(kind ParsingContext, parseElement func(p *Pa
 		}
 	}
 	p.parsingContexts = saveParsingContexts
-	return p.newNodeList(core.NewTextRange(pos, p.nodePos()), p.nodeSlicePool.Clone(list))
+	// Recording the trailing comma is deliberately done after the previous
+	// loop, and not just if we see a list terminator. This is because the list
+	// may have ended incorrectly, but it is still important to know if there
+	// was a trailing comma.
+	// Check if the last token was a comma.
+	// Always preserve a trailing comma by marking it on the NodeArray
```
DanielRosenwasser: But you're not, right?


```suggestion
	// !!! Always preserve a trailing comma by marking it on the NodeArray
```

jakebailey: This is a pure copy paste comment from Strada, but, I can tweak it yeah.

internal/parser/parser.go
```diff
@@ -637,7 +639,15 @@ func (p *Parser) parseDelimitedList(kind ParsingContext, parseElement func(p *Pa
...
 	p.parsingContexts = saveParsingContexts
-	return p.newNodeList(core.NewTextRange(pos, p.nodePos()), p.nodeSlicePool.Clone(list))
+	// Recording the trailing comma is deliberately done after the previous
+	// loop, and not just if we see a list terminator. This is because the list
+	// may have ended incorrectly, but it is still important to know if there
+	// was a trailing comma.
+	endPos := p.nodePos()
+	if commaStart < 0 && len(list) > 0 {
+		endPos = list[len(list)-1].End()
+	}
```
DanielRosenwasser: Strada just ends up calling `scanner.getTokenFullStart()` which is pretty much just the same as that first line you have - `endPos := p.nodePos()`.

What is being compensated for here?

internal/printer/printer_test.go
```diff
@@ -94,7 +94,7 @@ func TestEmit(t *testing.T) {
 		{title: "ArrowFunction#2", input: `()=>{}`, output: `() => { };`},
 		{title: "ArrowFunction#3", input: `(a)=>{}`, output: `(a) => { };`},
 		{title: "ArrowFunction#4", input: `<T>(a)=>{}`, output: `<T>(a) => { };`},
-		{title: "ArrowFunction#5", input: `async a=>{}`, output: `async a => { };`},
+		{title: "ArrowFunction#5", input: `async a=>{}`, output: `async (a) => { };`},
```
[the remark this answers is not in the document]
DanielRosenwasser: To make this explicit for other readers, this is resolved because Strada actually *does* emit with parentheses.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+}
+
 type classFieldsTransformer struct {
 	transformers.Transformer
-}
+	compilerOptions *core.CompilerOptions
+	resolver        printer.EmitResolver
+
+	// Computed configuration flags
+	shouldTransformInitializersUsingSet               bool
```
DanielRosenwasser, line 89: It would be nice if these were flags with helper methods, but it's fine.

jakebailey: This is a pure port, to be clear.

internal/parser/parser.go
jakebailey: The test I fixed comes out of broken code:

```js
                retValue = bfs.VARIABLES();
                if (retValue != 0 ^=  {

                    return 1;
                }
```

I think that this diff just isn't important and I should undo this attempt.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+		//
+		//   AssignmentExpression : LeftHandSideExpression `??=` AssignmentExpression
+		//     ...
+		//     4. If IsAnonymousFunctionDefinition(|AssignmentExpression|) is *true* and IsIdentifierRef of |LeftHandSideExpression| is *true*, then
+		//        a. Let _rval_ be ? NamedEvaluation of |AssignmentExpression| with argument _lref_.[[ReferencedName]].
+		//     ...
+
+		if isNamedEvaluationAnd(tx.EmitContext(), node.AsNode(), tx.isAnonymousClassNeedingAssignedName) {
+			node = transformNamedEvaluation(tx.EmitContext(), node.AsNode(), false, "").AsBinaryExpression()
+			debug.AssertNode(node.AsNode(), ast.IsAssertionExpression)
```
RyanCavanaugh: Copilot thinks this should be `ast.IsAssignmentExpression` - wrong comment, or missing coverage?

jakebailey: It's right, this is a typo added in a recent commit by me by hand.

jakebailey: Ah, because `ast.IsAssignmentExpression` has the wrong signature, so I missed it when accepting the completion. Yay.

jakebailey: Pushed a fix.

jakebailey: This of course implies we are missing coverage for this to not have failed somewhere. Yay...

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+		// Bind the `this` correctly for tagged template literals when the tag is a private identifier property access.
+		thisArg, target := tx.createCallBinding(node.Tag)
+		bindExpr := tx.Factory().NewCallExpression(
+			tx.Factory().NewPropertyAccessExpression(tx.Visitor().VisitNode(target), nil, tx.Factory().NewIdentifier("bind"), ast.NodeFlagsNone),
+			nil, /*questionDotToken*/
+			nil, /*typeArguments*/
+			tx.Factory().NewNodeList([]*ast.Node{tx.Visitor().VisitNode(thisArg)}),
+			ast.NodeFlagsNone,
+		)
+		return tx.Factory().UpdateTaggedTemplateExpression(
```
RyanCavanaugh, line 1271: Via copilot, missing a call to visit. Repro:
```ts
// @target: es2015
class Foo {
  #x = 3;
  #y = null as any;
  func() {
    console.log(this.#y`->>${this.#x}<<-`);
  }
}
```
expression between the arrows is not transformed

Copilot said:
  2. visitTaggedTemplateExpression: Template literal not visited through the visitor

  Go lines 1270, 1292: node.Template (passed directly, unvisited)
  TS lines 1403, 1425: visitNode(node.template, visitor, isTemplateLiteral)

  Both branches of the Go function pass node.Template directly to UpdateTaggedTemplateExpression without running it through the visitor. Any class-field-related transforms
  inside template spans (e.g., this.#field in tag`${this.#x}`) won't be transformed.

internal/printer/factory.go
```diff
@@ -907,6 +1029,30 @@ func (f *NodeFactory) NewExportStarHelper(moduleExpression *ast.Expression, expo
...
+			f.NewParameterDeclaration(nil, nil, paramName, nil, nil, nil),
+		}),
+		nil, /*returnType*/
+		nil, /*fullSignature*/
+		f.NewBlock(f.NewNodeList([]*ast.Node{
+			f.NewExpressionStatement(expression),
+		}), false),
+	)
+	objLiteral := f.NewObjectLiteralExpression(f.NewNodeList([]*ast.Node{setAccessor}), false)
+	// Explicit parens required because of v8 regression (https://bugs.chromium.org/p/v8/issues/detail?id=9560)
```
RyanCavanaugh, line 1056: Comment only: this was fixed 7 years ago, maybe we can remove now?

internal/transformers/estransforms/classfields.go
jakebailey: Fixed in the latest two commits (split to show that it was broken)

internal/printer/factory.go
jakebailey: 2019? that wasn't 7... oh no

jakebailey: I'd prefer not to undo this just since it adds a bunch of diffs and I don't know where all this is fixed (given old node versions).

internal/transformers/estransforms/classfields.go
jakebailey: Added a test which can exercise this.

RyanCavanaugh (approved):

internal/printer/helpers.go
```diff
@@ -131,7 +131,40 @@ var disposeResourcesHelper = &EmitHelper{
 });`,
 }
 
-// !!! Class Fields Helpers
+// Class Fields Helpers
+
+var classPrivateFieldGetHelper = &EmitHelper{
```
weswigham, line 184: These helpers have like 200-line comments explaining what they do in strada, since they're dense and written in shorthand - we should probably keep those.

internal/transformers/estransforms/async.go
```diff
@@ -471,6 +471,16 @@ func (tx *asyncTransformer) visitFunctionExpression(node *ast.Node) *ast.Node {
...
 // - The node is marked async
 func (tx *asyncTransformer) visitArrowFunction(node *ast.Node) *ast.Node {
+	// Class static blocks provide their own scope for `arguments`. When the classfields
+	// transformer lowers a static block to an arrow function IIFE, `arguments` references
+	// inside it should NOT be captured by the enclosing async function. This mirrors the
+	// checker's behavior where CaptureArguments is not set for `arguments` in static blocks.
+	if tx.EmitContext().MostOriginal(node).Kind == ast.KindClassStaticBlockDeclaration {
+		savedLexicalArguments := tx.lexicalArguments
+		tx.lexicalArguments = lexicalArgumentsInfo{}
+		defer func() { tx.lexicalArguments = savedLexicalArguments }()
```
weswigham, line 481: `arguments` in class static blocks is an error - `'arguments' cannot be referenced in property initializers or class static initialization blocks.` In strada I imagine we got this "preserve `arguments`" behavior in that fail case because we used a checker-based `isArgumentsLocalBinding` and, because of that error, the `arguments` isn't actually bound inside the static block? And that carries through to the block-as-iife, since the semantic check uses the original identifier node. Our emit here doesn't really matter, thanks to the error, but if we do want to match strada just for diffs, it'd be better to do something less brittle feeling, since as-is, this relies on exactly a static block being transformed to exactly an arrow with the original pointer set as such. IMO, it'd be better to set an emit flag on `arguments` references we hit while transforming `static` blocks that disables transforming them in the async transform if we really do want that, so how the transforms are collaborating is explicit. In particular, since class fields, as-is, is in the `esnext` section, other transforms can end up between the fields transform and the async transform, so something might end up, say, transforming that arrow in that iife to something else (perhaps injects metadata or sets a function name or something - a hypothetical, any future es transform goes here), which then no longer gets this special handling. Keeping it original-and-current-node-independent can prevent that and make it so future us likely doesn't need to think about this error case.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+	tx.discardedValueVisitor = tx.EmitContext().NewNodeVisitor(tx.visitDiscardedValue)
+	tx.heritageClauseVisitor = tx.EmitContext().NewNodeVisitor(tx.visitHeritageClause)
+	tx.modifierVisitor = tx.EmitContext().NewNodeVisitor(tx.visitModifier)
+	tx.isAnonymousClassNeedingAssignedName = tx.isAnonymousClassNeedingAssignedNameWorker
+
+	return result
+}
+
+// requiresBlockScopedVar returns true when private field temp variables should be
+// declared as block-scoped (let) rather than function-scoped (var). This occurs when
```
weswigham, line 190: You know, without script file emit, I'm pretty sure we can just default to `let` - you only needed `var` to allow multiple top-level declarations in multiple files to quietly merge, afaik. Not too important, but something we can probably clean up later after we're done minimizing diffs.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+		return tx.visitExportAssignment(node.AsExportAssignment())
+	case ast.KindPrivateIdentifier:
+		return tx.visitPrivateIdentifier(node)
+	case ast.KindPropertyAccessExpression:
+		return tx.visitPropertyAccessExpression(node.AsPropertyAccessExpression())
+	case ast.KindElementAccessExpression:
+		return tx.visitElementAccessExpression(node.AsElementAccessExpression())
+	case ast.KindPrefixUnaryExpression:
+		return tx.visitPreOrPostfixUnaryExpression(node, false /*discarded*/)
+	case ast.KindPostfixUnaryExpression:
```
weswigham: In the below `visitDiscardedValue` visit, this is merged with the prefix case - we probably aughta be consistent on if we keep the cases duplicated or merge them.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+		return tx.heritageClauseVisitor.VisitEachChild(node)
+	case ast.KindExpressionWithTypeArguments:
+		return tx.visitExpressionWithTypeArgumentsInHeritageClause(node.AsExpressionWithTypeArguments())
+	default:
+		return tx.visit(node)
+	}
+}
+
+// visitAssignmentTarget visits the assignment target of a destructuring assignment.
+func (tx *classFieldsTransformer) visitAssignmentTarget(node *ast.Node) *ast.Node {
```
weswigham, line 345: This is `visitDestructuringAssignmentTarget` in strada, afaik. I was thrown for a bit because this is where `assignmentTargetVisitor` is defined in strada, which is... very different (but shares the first few lines of logic!).

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+	if ast.IsDestructuringAssignment(node.AsNode()) {
+		// ({ x: obj.#x } = ...)
+		// ({ x: super.x } = ...)
+		// ({ x: super[x] } = ...)
+		savedPendingExpressions := tx.pendingExpressions
+		tx.pendingExpressions = nil
+		updated := tx.Factory().UpdateBinaryExpression(
+			node,
+			nil,
+			tx.visitAssignmentTarget(node.Left),
```
weswigham: This seems incorrect. `visitAssignmentTarget` is strada's `visitDestructuringAssignmentTarget`, which, while similar to `assignmentTargetVisitor`, is seemingly different in how it handles the non-array/object literal case. (The former doing a lot of manually private name checks before descending into children of the node, the later just invoking the main visitor on the node itself!)

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+}
+
+// visitPrivateIdentifier handles an undeclared private name. Replace it with an empty
+// identifier to indicate a problem with the code, unless we are in a statement position -
+// otherwise this will not trigger a SyntaxError.
+func (tx *classFieldsTransformer) visitPrivateIdentifier(node *ast.Node) *ast.Node {
+	if !tx.shouldTransformPrivateElementsOrClassStaticBlocks {
+		return node
+	}
+	// !!! Strada used .parent too; this is suspicious in a transform.
```
weswigham, line 461: Agree - this is probably a good catch and could definitely be causing crashes without the `nil` check. The parser should *probably* add a flag to private names parsed in statement positions (indicating that they need to be preserved instead of elided) to handle this without needing to double check the context in the transform here, but tracking the parent node is ultimately doable, too.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+}
+
+// transformPrivateIdentifierInInExpression visits `#id in expr`.
+func (tx *classFieldsTransformer) transformPrivateIdentifierInInExpression(node *ast.BinaryExpression) *ast.Node {
+	info := tx.accessPrivateIdentifier(node.Left)
+	if info != nil {
+		receiver := tx.Visitor().VisitNode(node.Right)
+		result := tx.Factory().NewClassPrivateFieldInHelper(info.brandCheckIdentifier, receiver)
+		tx.EmitContext().SetOriginal(result, node.AsNode())
+		result.Loc = node.Loc
```
weswigham: Is this copilot being cute, pattern matching and trying to be better than strada? I don't see a `setTextRange` call here in strada, and the absence of one can often be as important as the presence (for sourcemaps). This range assignment effectively makes the whole helper call the private name in the sourcemap, which probably isn't good? I think tools generally prefer things like strings or identifiers be names, rather than whole big helper calls.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+func (tx *classFieldsTransformer) shouldTransformClassElementToWeakMap(node *ast.Node) bool {
+	if tx.shouldTransformPrivateElementsOrClassStaticBlocks {
+		return true
+	}
+	return tx.shouldAlwaysTransformPrivateStaticElements(node)
+}
+
+func (tx *classFieldsTransformer) shouldAlwaysTransformPrivateStaticElements(node *ast.Node) bool {
+	// !!! return tx.EmitContext().GetInternalEmitFlags(node)&InternalEmitFlags.TransformPrivateStaticElements != 0
+	_ = node
```
weswigham, line 651: I'm guessing this is a TODO because it basically only exists to let the `esDecorator` transform tell the class fields transform to force a transform of a decorated field? Something to remember for the decorators PR, I guess.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+	return false
+}
+
+func (tx *classFieldsTransformer) visitMethodOrAccessorDeclaration(node *ast.Node) *ast.Node {
+	// !!!
+	// debug.Assert(!ast.HasDecorators(node))
```
weswigham, line 657: Again, presumably this is a TODO to uncomment this once `esDecorators` emit is in and decorators are always transformed (but it feels like a bad assumption to make long term if decorators stabilize into a yearly transform but class fields are stuck as a global transform until we split it up).

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+		saved := tx.currentClassElement
+		tx.currentClassElement = classElement
+		result := visitor(tx, node)
+		tx.currentClassElement = saved
+		return result
+	}
+	return visitor(tx, node)
+}
+
+func (tx *classFieldsTransformer) visitEachChildOfNode(node *ast.Node) *ast.Node {
```
weswigham, line 705: I'd make a note that this isn't inlined just so it can be captured as a callback and to prefer direct `tx.Visitor().VisitEachChild(node)` where possible.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+
+func (tx *classFieldsTransformer) tryGetClassThis() *ast.Expression {
+	lex := tx.getClassLexicalEnvironment()
+	if lex.classThis != nil {
+		return lex.classThis
+	}
+	if lex.classConstructor != nil {
+		return lex.classConstructor
+	}
+	return nil
```
weswigham, line 764: What about `currentClassContainer.name`? Is there a reason that fallback from strada is missing here?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+	}
+
+	modifiers := tx.modifierVisitor.VisitModifiers(node.Modifiers())
+	backingField := tx.createAccessorPropertyBackingField(node, modifiers, node.Initializer)
+	tx.EmitContext().SetOriginal(backingField, node.AsNode())
+	tx.EmitContext().SetEmitFlags(backingField, printer.EFNoComments)
+	tx.EmitContext().SetSourceMapRange(backingField, sourceMapRange)
+
+	receiver := tx.tryGetClassThis()
+	if receiver == nil {
```
weswigham: Well. This whole block flows pretty differently than strada. I'd have expected something like
```go
var receiver *ast.Node
if ast.IsStatic(node.AsNode()) {
    receiver = tx.tryGetClassThis()
    if receiver == nil {
	    receiver = tx.Factory().NewThisExpression()
    }
} else {
	receiver = tx.Factory().NewThisExpression()
}
```
where's the current class container stuff come from? What's it fixing?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3515 @@
...
+	}
+	setter := tx.createAccessorPropertySetRedirector(node, setterModifiers, setterName, receiver)
+	tx.EmitContext().SetOriginal(setter, node.AsNode())
+	tx.EmitContext().SetEmitFlags(setter, printer.EFNoComments)
+	tx.EmitContext().SetSourceMapRange(setter, sourceMapRange)
+
+	// Visit the results in a second pass
+	results := []*ast.Node{backingField, getter, setter}
+	var visited []*ast.Node
+	for _, result := range results {
```
weswigham: ...This loop feels silly.  Pretty sure it's identical to
```go
results := []*ast.Node{tx.transformFieldInitializer(backingField), tx.visitClassElement(getter), tx.visitClassElement(setter)}
```
but given the possible nilness and multi-element-result-ness, that nor this are quite on the money. Keeping the little visitor from strada and using
```go
tx.accessorFieldResultVisitor.VisitSlice([]*ast.Node{backingField, getter, setter})
```
is probably best? All the explicit syntax list and single element handling here is what we avoid by using those visit helpers, after all.

weswigham (commented):
I'm about halfway through this - I'll continue, but don't expect final comments for a bit.

internal/transformers/estransforms/classfields.go
jakebailey: Hm, this is actually just dead code.

internal/transformers/estransforms/classfields.go
jakebailey: Yeah, I didn't want to attempt to port something I couldn't really test.

internal/transformers/estransforms/classfields.go
jakebailey: Adding this back in causes some really strange diffs:

```diff
diff --git a/./testdata/baselines/reference/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js b/./testdata/baselines/local/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js
index ff19b449a..88502991b 100644
--- a/./testdata/baselines/reference/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js
+++ b/./testdata/baselines/local/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js
@@ -14,5 +14,5 @@ class C {
     bar() {
         return 0;
     }
-    [this.bar()]() { }
+    [C.bar()]() { }
 }
diff --git a/./testdata/baselines/local/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js.diff b/./testdata/baselines/local/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js.diff
new file mode 100644
index 000000000..00eaff070
--- /dev/null
+++ b/./testdata/baselines/local/submodule/conformance/computedPropertyNames21_ES5(target=es2015).js.diff
@@ -0,0 +1,9 @@
+--- old.computedPropertyNames21_ES5(target=es2015).js
++++ new.computedPropertyNames21_ES5(target=es2015).js
+@@= skipped -13, +13 lines =@@
+     bar() {
+         return 0;
+     }
+-    [this.bar()]() { }
++    [C.bar()]() { }
+ }
\ No newline at end of file
```

internal/transformers/estransforms/classfields.go
jakebailey: This is just `accessorFieldResultVisitor` from Strada inlined; I could make it a visitor again, it's just annoying because the lifting code still has to be done and we don't have that code somewhere as far as I can tell...

jakebailey: I'll just bring back all of the visitors for consistency.

internal/transformers/estransforms/classfields.go
jakebailey: This was a shitty cleanup by me as the _other_ code that uses `tryGetClassThis` didn't actually before and was a nice `?. ?? ` etc expression. I'll refactor it to something different.

internal/transformers/estransforms/classfields.go
jakebailey: Related to that comment below; things get weird because of the substitution replacements (which are fresh); I can swap things around a little to make the names match Strada.

internal/transformers/estransforms/classfields.go
jakebailey: Yep, brought back the real visitor for this.

internal/transformers/estransforms/classfields.go
jakebailey: We still do script file emit, though, just not the joining?

internal/transformers/estransforms/classfields.go
jakebailey: Yep, exactly. Uncommenting this assert fails of course because decorators are being sent through. Which causes all sort of weirdness of course.

internal/transformers/estransforms/async.go
jakebailey: Yeah, this didn't exactly thrill me to hack in, it was more about making baselines less diffy despite the error. I'll rethink.

jakebailey: Replaced it with a plain flag as suggested, which works fine.

internal/transformers/estransforms/classfields.go
jakebailey: This turns out to matter just for:

```ts
// @strict: false
// @target: es6

#

class C {
    #

    m() {
        this.#
    }
}
```

This is a syntax error, so I'm kind of surprised we left the code to print it back?

jakebailey: Especially given the JS emit is:

```js
"use strict";
var __classPrivateFieldGet = (this && this.__classPrivateFieldGet) || function (receiver, state, kind, f) {
    if (kind === "a" && !f) throw new TypeError("Private accessor was defined without a getter");
    if (typeof state === "function" ? receiver !== state || !f : !state.has(receiver)) throw new TypeError("Cannot read private member from an object whose class did not declare it");
    return kind === "m" ? f : kind === "a" ? f.call(receiver) : f ? f.value : state.get(receiver);
};
var _C_;
#;
class C {
    constructor() {
        _C_.set(this, void 0);
    }
    m() {
        __classPrivateFieldGet(this, _C_, "f");
    }
}
_C_ = new WeakMap();
```

which is... what?

jakebailey: Then again we emit this for esnext in Strada too:

```js
#;
class C {
    #;
    m() {
        this.#;
    }
}
```

so that's exciting

jakebailey: Another fix is to just handle this in `visitExpressionStatement`. No parent access needed (that's the parent).

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	propDecl := node.AsPropertyDeclaration()
+	if ast.IsAutoAccessorPropertyDeclaration(node) && (tx.shouldTransformAutoAccessorsInCurrentClass() ||
+		ast.HasStaticModifier(node) && tx.shouldAlwaysTransformPrivateStaticElements(node)) {
+		return tx.transformAutoAccessor(propDecl)
+	}
+	return tx.transformFieldInitializer(propDecl)
+}
+
+func (tx *classFieldsTransformer) createPrivateIdentifierAccess(info *privateIdentifierInfo, receiver *ast.Expression) *ast.Expression {
+	receiver = tx.Visitor().VisitNode(receiver)
```
weswigham, line 972: If we're not calling `tx.ensureDynamicThisIfNeeded(receiver)` here, how are we skipping the `this`->`_classThis` transform on these? Is there equivalent logic in the direct transform now?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+
+func (tx *classFieldsTransformer) createPrivateIdentifierAccessHelper(info *privateIdentifierInfo, receiver *ast.Expression) *ast.Expression {
+	tx.EmitContext().SetCommentRange(receiver, core.NewTextRange(-1, receiver.End()))
+
+	switch info.kind {
+	case privateIdentifierKindAccessor:
+		return tx.Factory().NewClassPrivateFieldGetHelper(
+			receiver,
+			info.brandCheckIdentifier,
+			"a",
```
weswigham: Given these strings are fixed in the helper, we should probably make the enumerated type available in the factory and use it here, too. Don't want to risk them drifting.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+			tx.createPrivateIdentifierAccessHelper(info, readExpression),
+			nil,
+			tx.Factory().NewToken(transformers.GetNonAssignmentOperatorForCompoundAssignment(operator)),
+			right,
+		)
+	}
+
+	tx.EmitContext().SetCommentRange(receiver, core.NewTextRange(-1, receiver.End()))
+
+	switch info.kind {
```
weswigham, line 1567: This switch seems to be missing the `privateIdentifierKindUntransformed` case (which is a debug fail).

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+		return tx.Factory().NewClassPrivateFieldSetHelper(
+			receiver,
+			info.brandCheckIdentifier,
+			right,
+			"f",
+			f,
+		)
+	}
+	debug.AssertNever(info, "Unknown private element type")
+	return tx.Factory().NewAssignmentExpression(receiver, right)
```
weswigham: If this is unreachable, why is the return value non-nil?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+		// If we are transforming class static blocks, then we can reuse `_classThis` since the assignment
+		// will be evaluated *before* the transformed static blocks are evaluated and thus won't overwrite
+		// the replacement constructor.
+
+		if tx.shouldTransformPrivateElementsOrClassStaticBlocks && tx.EmitContext().ClassThis(node) != nil {
+			classThis := tx.EmitContext().ClassThis(node)
+			tx.getClassLexicalEnvironment().classConstructor = classThis
+			pendingClassReferenceAssignment = tx.Factory().NewAssignmentExpression(
+				classThis,
+				tx.Factory().GetDeclarationName(node),
```
weswigham: I guess `getInternalName` only differs in marking class internal self references which only matters for downlevel class emit (where internal and external refs are substituted differently), which we don't do anymore. Something to remember if something makes us desugar a class during a transform, though.

Although `getLocalName` is still closer to the original than `getDeclarationName` afaik.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	)
+
+	result := make([]*ast.Node, 0, 1+len(statements)+1)
+	if membersPrologue != nil {
+		result = append(result, tx.Factory().NewExpressionStatement(membersPrologue))
+	}
+	result = append(result, updatedClass)
+	result = append(result, statements...)
+
+	if len(result) == 1 {
```
weswigham: You don't have to bother with this, afaik. A zero or one element syntax list is correctly unwrapped by the visitor helper, so always making a list over the maybe-empty slice is fine.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+			tx.pendingStatements = append(tx.pendingStatements, tx.Factory().NewExpressionStatement(expr))
+		}
+	}
+
+	// Emit static properties as statements (via pendingStatements) using the class's
+	// internal name as the receiver, matching the class declaration output structure.
+	if len(staticPropertiesOrClassStaticBlocks) > 0 {
+		classThisOrName := tx.EmitContext().ClassThis(node)
+		if classThisOrName == nil {
+			classThisOrName = tx.Factory().GetDeclarationName(node)
```
weswigham: Once again, `GetLocalName` is probably closer to `GetInternalName` than `GetDeclarationName` in behavior.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	}
+
+	if temp != nil {
+		expressions = append(expressions, tx.Factory().NewAssignmentExpression(temp, classExpression))
+	} else if tx.shouldTransformPrivateElementsOrClassStaticBlocks && tx.EmitContext().ClassThis(node) != nil {
+		expressions = append(expressions, tx.Factory().NewAssignmentExpression(tx.EmitContext().ClassThis(node), classExpression))
+	} else {
+		expressions = append(expressions, classExpression)
+	}
+
```
weswigham, line 2089: The intending/startOnNewLine-ing on the `expressions` from the other branch is missing in this branch. If the branches are gonna stay separate like this, it's probably best extracted to a helper.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	statements = tx.EmitContext().EndAndMergeVariableEnvironment(statements)
+
+	if len(statements) == 0 && constructor == nil {
+		return nil
+	}
+
+	multiLine := true
+	if constructor != nil && constructor.Body != nil {
+		multiLine = constructor.Body.AsBlock().Multiline || len(statements) > 0
```
weswigham: Isn't this different than strada? In strada, if the original body multiline is `false`, the result is `false`, even when there are `statements`, which isn't the case here, because we're using `||` instead of js's `??`.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+		expressions = append(expressions, expression)
+	}
+	return expressions
+}
+
+// transformProperty transforms a property initializer into an assignment expression.
+func (tx *classFieldsTransformer) transformProperty(property *ast.PropertyDeclaration, receiver *ast.Expression) *ast.Expression {
+	savedCurrentClassElement := tx.currentClassElement
+	transformed := tx.transformPropertyWorker(property, receiver)
+	tx.currentClassElement = savedCurrentClassElement
```
weswigham, line 2625: I understand why the onEmit hook stuff isn't needed anymore, but shouldn't this still have `setOriginalNode` and `setSourceMapRange` calls on static properties?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+		} else {
+			debug.Fail("Undeclared private name for property declaration.")
+		}
+	}
+
+	if (ast.IsPrivateIdentifier(propertyName) || ast.HasStaticModifier(property.AsNode())) && property.Initializer == nil {
+		return nil
+	}
+
+	if ast.HasAbstractModifier(tx.EmitContext().MostOriginal(property.AsNode())) {
```
weswigham, line 2680: I see that strada does this, and it seems icky. IMO, all the original'ing going on here speaks to some sloppy coordination between the `runtimesyntax` and `classfields` transform. Namely, the former is removing information the later needs without setting any explicit flags as replacements. Obviously, since strada does this, it's fine here, but it's probably something that aughta get fixed in the future, since this is brittle. Honestly, anyone with custom transformers in strada could have tripped over this.

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	spread := node.AsSpreadElement()
+	if ast.IsLeftHandSideExpression(spread.Expression) {
+		expr := tx.visitDestructuringAssignmentTarget(spread.Expression)
+		return tx.Factory().UpdateSpreadElement(spread, expr)
+	}
+	return tx.Visitor().VisitEachChild(node)
+}
+
+func (tx *classFieldsTransformer) visitArrayAssignmentElement(node *ast.Node) *ast.Node {
+	if ast.IsSpreadElement(node) {
```
weswigham: Is this actually the same as the strada logic? Is this actually only ever called on known assignment elements? I'm wondering if an array on the lhs of an assignment can ever have any non-binding-element expression members in any error scenarios...

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+	spread := node.AsSpreadAssignment()
+	if ast.IsLeftHandSideExpression(spread.Expression) {
+		expr := tx.visitDestructuringAssignmentTarget(spread.Expression)
+		return tx.Factory().UpdateSpreadAssignment(spread, expr)
+	}
+	return tx.Visitor().VisitEachChild(node)
+}
+
+func (tx *classFieldsTransformer) visitObjectAssignmentElement(node *ast.Node) *ast.Node {
+	if ast.IsSpreadAssignment(node) {
```
weswigham, line 3202: Should we keep the debug assert here?

internal/transformers/estransforms/classfields.go
```diff
@@ -1,19 +1,3530 @@
...
+		//
+		// Source:
+		// ([ this.#myProp ] = [ "hello" ]);
+		//
+		// Transformation:
+		// [ { set value(x) { this.#myProp = x; } }.value ] = [ "hello" ];
+		elements := node.AsArrayLiteralExpression().Elements
+		var visited []*ast.Node
+		for _, elem := range elements.Nodes {
+			visited = append(visited, tx.visitArrayAssignmentElement(elem))
```
weswigham: Uh, does this behave correctly if `visitArrayAssignmentElement` passes on `nil` or a syntax list from something it calls into? Seems to me like it aughta still be a small visitor.

internal/transformers/estransforms/classfields.go
weswigham: There is - found it later.

internal/transformers/estransforms/classfields.go
jakebailey: Ah, you know, I skipped this because it was a very strange string literal, then never added them back. Will fix.

internal/transformers/estransforms/classfields.go
jakebailey: I'll just have it (and the others) return nil; just sort of random choices. The debug asserts were restored later.

internal/transformers/estransforms/classfields.go
jakebailey: You're right, nice. I didn't realize it worked.

internal/transformers/estransforms/classfields.go
jakebailey: This is crappy code because I had forgotten VisitSlice existed; I fixed earlier ones but will fix these now and check for more.

internal/transformers/estransforms/classfields.go
jakebailey: I can leave a TODO on this one for us to check later.

internal/transformers/estransforms/classfields.go
jakebailey: Yes, I'm still working on adding them back (stopped working on it last night)

internal/transformers/estransforms/classfields.go
jakebailey: Er, this is just VisitNodes. Sigh

internal/transformers/estransforms/classfields.go
jakebailey: Yes, fixed.

internal/transformers/estransforms/classfields.go
jakebailey: I fixed it to match, though nothing changed, of course

internal/transformers/estransforms/classfields.go
jakebailey: I'll re-port it, but it is a little weird given `Multiline` is a bool. It doesn't seem to really matter how it's written, we I guess are missing test coverage for it?

internal/transformers/estransforms/classfields.go
jakebailey: Yeah, these should have been `getLocalName`, e.g. how legacyDecorators.ts chose between them. I fixed it in some cases, but apparently not this block.

internal/transformers/estransforms/classfields.go
jakebailey: Done; made it a string enum like it used to be (but actually defined "untransformed" this time).

internal/transformers/estransforms/classfields.go
jakebailey: Fixed, though I'll see if there's a non-gross way to not duplicate this so hard.

jakebailey: I shoved the two together.

internal/transformers/estransforms/classfields.go
```diff
@@ -2540,9 +2540,12 @@ func (tx *classFieldsTransformer) transformConstructorBody(container *ast.Node,
 		return nil
 	}
 
-	multiLine := true
-	if constructor != nil && constructor.Body != nil {
+	var multiLine bool
+	if constructor != nil && constructor.Body != nil &&
+		len(constructor.Body.AsBlock().Statements.Nodes) >= len(statements) {
 		multiLine = constructor.Body.AsBlock().Multiline || len(statements) > 0
```
weswigham: This still isn't equivalent to strada, no?  The surrounding lines match now, yeah, but `||` still isn't equivalent to js `??`, which basically makes the `len(statements) > 0` clause dominate here, rather than the existing set value from the block (whatever non-nil value it is).

weswigham: Like, if you assume `Multiline` is an always set bool (and not possibly undefined), then the `len(statements) > 0` part of the clause is just unreachable in strada.

jakebailey: I don't know what it is about this specific snippet but it really confused me to try and think of the cases 😅 

I think I was trying to think of "there are no statements in either list", but it would be false then too. And if it were actually `constructor(){ \n }`, then that would be multiline, so it'd work too. So, yeah, I see now.

Not sure why we went to a bool for this, when Tristate is the same size in memory and would let us do exactly what the old behavior was (across the board).

jakebailey: Very annoyingly, if you actually flip all the multilines to tristates and make it all match... no tests change. Hilarious

PR #2960, opened 2026-03-03
Fixes a crash when requesting implementations on a tripleslash directive

Andarist:
Fixes a crash found here: https://github.com/microsoft/typescript-go/issues/2945#issuecomment-3981326421

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -594,7 +594,7 @@ func (l *LanguageService) provideSymbolsAndEntries(ctx context.Context, uri lspr
 	position := int(l.converters.LineAndCharacterToPosition(sourceFile, documentPosition))
 
 	node := astnav.GetTouchingPropertyName(sourceFile, position)
-	if isRename && !isNodeEligibleForRename(node) {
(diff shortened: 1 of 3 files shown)


internal/ls/findallreferences.go
```diff
@@ -594,7 +594,7 @@ func (l *LanguageService) provideSymbolsAndEntries(ctx context.Context, uri lspr
 	position := int(l.converters.LineAndCharacterToPosition(sourceFile, documentPosition))
 
 	node := astnav.GetTouchingPropertyName(sourceFile, position)
-	if isRename && !isNodeEligibleForRename(node) {
+	if isRename && !isNodeEligibleForRename(node) || implementations && ast.IsSourceFile(node) {
```
Andarist, line 597: Same early out as in Strada: https://github.com/microsoft/TypeScript/blob/c9e7428bb76f0543a3555d0af87777e7db3a41e6/src/services/findAllReferences.ts#L564-L567

but the funny thing is that Strada also crashes in this scenario because of the unguarded `node.parent.kind` access here:
https://github.com/microsoft/TypeScript/blob/c9e7428bb76f0543a3555d0af87777e7db3a41e6/src/services/findAllReferences.ts#L537-L544

PR #2965, opened 2026-03-03
Allow turning auto-imports off; some edge case fixes

andrewbranch:
Related to #2780 but I don't think it will fix it.

1. Respect `"js/ts.suggest.autoImports": false`
2. There were some theoretical ways that `node_modules/.pnpm` or `node_modules/.store-fuse-unplugged` could have been mistakenly processed as a package, but I don't think they were realistic. Still, it's worth guarding against and ruling out.
3. We were providing all node_modules packages to files whose package.json was unparseable. Instead, we'll provide just things that are already imported in the program.


internal/ls/autoimport/util.go
```diff
@@ -217,6 +217,10 @@ func createCheckerPool(program checker.Program) (getChecker func() (*checker.Che
 // to the given set, canonicalizing @types package names to their base names.
 func addPackageJsonDependencies(contents *packagejson.PackageJson, deps *collections.Set[string]) {
 	contents.RangeDependencies(func(name, _, field string) bool {
+		if name == "" || name == "@types/" || name[0] == '.' {
+			// Edge cases that could make us blow up probably
+			return true
+		}
```
jakebailey, line 223: Does this break packages like `tshy`'s output?

andrewbranch: No, this is literally for

```json
{
  "dependencies": {
    "": "1.0.0",
    ".pnpm": "1.0.0",
    "@types/": "1.0.0"
  }
}
```

jakebailey (approved):

Issue #2959, opened 2026-03-03
panic handling request textDocument/definition: bad line number


Zzzen:
## Stack trace

```
 [error] panic handling request textDocument/definition: bad line number. Line: 14, lineMap length: 14
goroutine 6717536 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:26 +0x5e
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0x42a32ed1508, 0x42afe2d0b40)
	github.com/microsoft/typescript-go/internal/lsp/server.go:873 +0x47
panic({0xc7f8a0?, 0x42a7c49af70?})
	runtime/panic.go:860 +0x13a
github.com/microsoft/typescript-go/internal/ls/lsconv.(*Converters).LineAndCharacterToPosition(0x42d50167aa0, {0x107dde8, 0x42afbc56a88}, {0xc6831888?, 0x42a?})
	github.com/microsoft/typescript-go/internal/ls/lsconv/converters.go:153 +0x2d4
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideDefinition(0x42a89d263c0, {0x1080998, 0x42afbc627e0}, {0x42c6af56080, 0x7a}, {0xe?, 0x4d?})
	github.com/microsoft/typescript-go/internal/ls/definition.go:26 +0x138
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleDefinition(0x42d3ab75e98?, {0x1080998?, 0x42afbc627e0?}, 0x0?, 0x1080998?)
	github.com/microsoft/typescript-go/internal/lsp/server.go:1189 +0x35
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].14.1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:750 +0xb4
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification.func1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:613 +0x95
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func3()
	github.com/microsoft/typescript-go/internal/lsp/server.go:488 +0x2c
created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 23
	github.com/microsoft/typescript-go/internal/lsp/server.go:487 +0x4ac
```

[...]

PR #2966, opened 2026-03-03
Clamp received `lsproto.Position`s

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/2959

### DIFF
--- a/internal/ls/lsconv/converters.go
+++ b/internal/ls/lsconv/converters.go
@@ -2,7 +2,6 @@ package lsconv
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"slices"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/project/session_test.go
```diff
@@ -952,4 +953,50 @@ func TestSession(t *testing.T) {
 		assert.DeepEqual(t, *actualConfig2.TS(), *expectedPrefs1)
 		assert.DeepEqual(t, *actualConfig2.JS(), *expectedPrefs2)
 	})
+
+	t.Run("definition with stale position after file shrink does not panic", func(t *testing.T) {
```
jakebailey, line 957: Is this not an entirely illegal situation? The client can't give us positions that our out of range for its own view...

internal/ls/lsconv/converters.go
```diff
@@ -149,13 +148,16 @@ func (c *Converters) LineAndCharacterToPosition(script Script, lineAndCharacter
 	line := core.TextPos(lineAndCharacter.Line)
 	char := core.TextPos(lineAndCharacter.Character)
 
-	if line < 0 || int(line) >= len(lineMap.LineStarts) {
-		panic(fmt.Sprintf("bad line number. Line: %d, lineMap length: %d", line, len(lineMap.LineStarts)))
```
Andarist, line 153: A rogue client can always send some invalid position. It's definitely a tradeoff and we could return an error instead but clamping to the bounds seems somewhat benign to me here.

DanielRosenwasser:
It is possible/probable that we will eventually do something like this (or at least silently error in the output pane), but I think right now we do want to find which sorts of issues are coming up so we can either figure out what we're doing wrong in request handling, or in VS Code's language server client.

internal/project/session_test.go
Andarist: Fair, I mentioned this is is a tradeoff here: https://github.com/microsoft/typescript-go/pull/2966#discussion_r2880722372 . I think that in the long run this should return an explicit error (instead panicing) if the drift like this is not acceptable (which is a fair stance!).

As to the "proper" solution. I suspect this will require "snapshot freezing" like the one in https://github.com/microsoft/typescript-go/pull/2905 . Once that PR settles - I'll take another look at this specific crash here again.

Andarist:
superseded by https://github.com/microsoft/typescript-go/pull/3365

PR #2967, opened 2026-03-03
Fix various namespace diffs

jakebailey:
Various fixes that fix import elision and related problems in namespaces.


testdata/baselines/reference/submodule/compiler/exportImportNonInstantiatedModule.js.diff
```diff
@@ -1,9 +1,10 @@
...
+
+ //// [exportImportNonInstantiatedModule.js]
  "use strict";
- var B;
- (function (B) {
-+    B.A1 = A;
- })(B || (B = {}));
+-var B;
+-(function (B) {
+-})(B || (B = {}));
```
jakebailey: B disappeared, exciting

internal/transformers/tstransforms/runtimesyntax.go
```diff
@@ -252,31 +263,14 @@ func (tx *RuntimeSyntaxTransformer) addVarForDeclaration(statements []*ast.State
...
-				tx.Factory().NewExportSpecifier(
-					false, /*isTypeOnly*/
-					nil,   /*propertyName*/
-					node.Name().Clone(tx.Factory()),
-				),
-			})),
-			nil, /*moduleSpecifier*/
-			nil, /*attributes*/
-		))
-	}
```
jakebailey, line 270: This is what was responsible for doing `export { Foo }`; I'm not sure the rationale for the change originally.

DanielRosenwasser: Might've been a bridge for gaps we had in the module-level transforms.

DanielRosenwasser (approved):

Issue #2958, opened 2026-03-03
Regression from Strada: IntelliSense does not work properly inside `{}` following `@throws`


tats-u:
## Steps to reproduce

Prepare:

```ts
interface ApiResponse {
  downloads: number;
  start: string;
  end: string;
  package: string;
}

/**
 * Fetch the number of downloads for a given npm package in the last month.
 * @param packageName The name of the npm package to fetch download data for.
 * @returns A promise that resolves to the number of downloads in the last month.
 * @throws {TypeError} If the fetch request fails or returns a non-OK status.
 */
export async function fetchDownloadsLastMonth(packageName: string): Promise<number> {
  const response = await fetch(`https://api.npmjs.org/downloads/point/last-month/${packageName}`);
  if (!response.ok) {
    throw new TypeError(`Failed to fetch data: ${response.status} ${response.statusText}`);
  }
  return (await response.json() as ApiResponse).downloads;
}
```

Almost all of the above code is generated by Copilot inline completion (slop).

1. Remove `TypeError ` in the brackets after `@throws`
2. Type "`Ty`" at the same place

## Behavior with `typescript@6.0`

"`TypeError`" appears in the IntelliSense and you can select it.

<img width="556" height="521" alt="Image" src="https://github.com/user-attachments/assets/021d03a6-3ea4-43a8-96fb-6b573668f201" />

## Behavior with `tsgo`

"`TypeError`" does not appear in the IntelliSense.

<img width="370" height="199" alt="Image" src="https://github.com/user-attachments/assets/c47021fe-79dd-49b0-81fb-5feaebdb2a2a" />

PR #2971, opened 2026-03-03
Support JSDoc `@throws`

Andarist:
closes https://github.com/microsoft/typescript-go/issues/2958


jakebailey (approved):
This seems all correct. Not sure how we left this behind.

PR #2972, opened 2026-03-04
[@typescript/api] Add a few more checker APIs

andrewbranch:
- `isContextSensitive`
- `getReturnTypeOfSignature`
- `getRestTypeOfSignature`
- `getTypePredicateOfSignature`
- `getBaseTypes`
- `getPropertiesOfType`
- `getIndexInfosOfType`
- `getConstraintOfTypeParameter`
- `getTypeArguments`

### DIFF
--- a/_packages/api/src/proto.ts
+++ b/_packages/api/src/proto.ts
@@ -169,7 +169,7 @@ export interface SignatureResponse {
 
 export interface TypePredicateResponse {
     kind: number;
-    parameterIndex?: number;
+    parameterIndex: number;
     parameterName?: string;
     type?: TypeResponse;
 }


_packages/api/src/proto.ts
```diff
@@ -166,3 +166,16 @@ export interface SignatureResponse {
     thisParameter?: string;
     target?: string;
 }
+
+export interface TypePredicateResponse {
+    kind: number;
+    parameterIndex?: number;
```
jakebailey: This is now always set after that `omitempty` change. Which is correct?

andrewbranch: Hm, probably using an `*int32` would have been better. It doesn't exactly matter because this is an internal type, and the user sees a discriminated union instead. I'll switch it though.

andrewbranch: Actually, that adds unnecessary complexity to forming the response. I'll update the type.

PR #2973, opened 2026-03-04
Fix quick info display for `@throws` tag's `typeExpression`

Andarist:
It's a small fixup related to recent https://github.com/microsoft/typescript-go/pull/2971 . I'm not sure why - but those baselines have a different format than Strada (display parts vs markdown) so those don't come with `.diff`s that would catch this earlier


jakebailey:
> I'm not sure why - but those baselines have a different format than Strada (display parts vs markdown)

Self answering 😅

There's no such thing as display parts anymore. The LSP has no such concept.

jakebailey (approved):

PR #2977, opened 2026-03-04
Add `assert` option to hereby

gabritto:
Planning on running the fuzzer with a build with assertions enabled, but I didn't want to specify `--debug` as that also turns off optimizations, so I added an `assert` option.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -60,6 +60,7 @@ const { values: rawOptions } = parseArgs({
         debug: { type: "boolean" },
         dirty: { type: "boolean" },
         release: { type: "boolean" },
+        assert: { type: "boolean" },
 
         setPrerelease: { type: "string" },
         forRelease: { type: "boolean" },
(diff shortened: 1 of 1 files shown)


Herebyfile.mjs
```diff
@@ -209,7 +210,7 @@ function buildTsgo(opts) {
     opts ||= {};
     const out = opts.out ?? "./built/local/";
     const env = { ...goBuildEnv, ...opts.env };
-    return $({ cancelSignal: opts.abortSignal, env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${options.debug ? goBuildTags("noembed") : goBuildTags("noembed", "noassert")} -o ${out} ./cmd/tsgo`;
+    return $({ cancelSignal: opts.abortSignal, env })`go build ${goBuildFlags} ${opts.extraFlags ?? []} ${options.debug || options.assert ? goBuildTags("noembed") : goBuildTags("noembed", "noassert")} -o ${out} ./cmd/tsgo`;
```
jakebailey, line 213: I think this logic might now be confusing to keep inline. Can we make a function or global similar to goBuildEnv to extract the build tag logic out a bit?

gabritto: I'm not sure how to make it less complicated. I think the code is as is today because we use `goBuildTags` both for build but also for testing, so can't just e.g. shove this logic into `goBuildTags`

PR #2991, opened 2026-03-05
Auto import memory optimizations

andrewbranch:
- Deduplicates package extraction work and `*autoimport.Export` and entrypoint info objects by realpath, which can be significant when many workspace packages have the same dependencies and resolve them via symlink (thanks @goloveychuk for noticing this)
- Replaces a `collections.Set[string]` on each `Export` with bitflags
- Pulls in #2855

Should help #2780

## Results

Measured by opening one file in each workspace package and triggering auto-imports.

**repo-pnpm - 10 packages**

  | /gc/heap/live:bytes
-- | --:
nightly | 1,245,135,800
pr | 753,368,456

**vscode - 1 package**

  | /gc/heap/live:bytes
-- | --:
nightly | 1,590,839,104
pr | 1,089,793,528

**DefinitelyTyped-tools - 14 packages**
  | /gc/heap/live:bytes
-- | --:
nightly | 476,638,976
pr | 368,886,320

### DIFF
--- a/internal/ls/autoimport/registry_test.go
+++ b/internal/ls/autoimport/registry_test.go
@@ -3,7 +3,6 @@ package autoimport_test
 import (
 	"context"
 	"fmt"
-	"math"
 	"testing"
 
 	"github.com/microsoft/typescript-go/internal/collections"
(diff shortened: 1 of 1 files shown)


internal/ls/autoimport/registry.go
jakebailey, line 1260: Totally not for this PR, but for the language server, I wonder if we should just say "`GetSourceFile` must always returns a bound file" given we largely assume that to be the case.

goloveychuk:
image
(It's found by ai. With my idea/assumption though)

PR #2995, opened 2026-03-05
Fix call hierarchy crash on anonymous classes/functions

Andarist:
fixed a crash found here: https://github.com/microsoft/typescript-go/issues/2988#issuecomment-4001335830

### DIFF
--- a/internal/ls/callhierarchy.go
+++ b/internal/ls/callhierarchy.go
@@ -103,21 +103,9 @@ func isValidCallHierarchyDeclaration(node *ast.Node) bool {
 		return ast.IsIdentifier(node.Name())
 	}
 
-	if ast.IsFunctionDeclaration(node) || ast.IsClassDeclaration(node) {
-		if name := node.Name(); ast.NodeIsPresent(name) {
-			return true
-		}
-		if modifiers := node.Modifiers(); modifiers != nil {
-			for _, mod := range modifiers.Nodes {
-				if mod.Kind == ast.KindDefaultKeyword {
-					return true
-				}
-			}
-		}
-		return false
-	}
-
-	return ast.IsClassStaticBlockDeclaration(node) ||
+	return ast.IsFunctionDeclaration(node) ||
+		ast.IsClassDeclaration(node) ||
+		ast.IsClassStaticBlockDeclaration(node) ||
 		ast.IsMethodDeclaration(node) ||
 		ast.IsMethodSignatureDeclaration(node) ||
 		ast.IsGetAccessorDeclaration(node) ||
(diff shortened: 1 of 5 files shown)


internal/ls/callhierarchy.go
```diff
@@ -103,9 +103,21 @@ func isValidCallHierarchyDeclaration(node *ast.Node) bool {
...
+		}
+		if modifiers := node.Modifiers(); modifiers != nil {
+			for _, mod := range modifiers.Nodes {
+				if mod.Kind == ast.KindDefaultKeyword {
+					return true
+				}
+			}
+		}
+		return false
+	}
```
jakebailey: why doesn't strada do this?

Andarist: it actually crashes too! :p

Andarist: IIRC one of the tests crashes in Strada, one changes the behavior slightly (imho to an improved one)

jakebailey: Hm, I guess I think it's weird that we need this change all, but I don't know the history of the original debug assert elsewhere?

This ends up duplicating logic, which I find suspiciousish.

DanielRosenwasser: Spoke with Jake outside of the thread - it's possible we just want to get rid of the assertion. Everything else looks like it's wired up to just gracefully propagate nil. Do you see any regressions if you just use the old code and remove the assertion?

Andarist: I pushed out a different version of this - removing this one assert and adjusting some code in `getCallHierarchyItemName` to avoid hitting another assert there.

testdata/baselines/reference/fourslash/callHierarchy/callHierarchyUnclosedTemplateExprNoCrash1.callHierarchy.txt
```diff
@@ -0,0 +1,16 @@
+// === Call Hierarchy ===
+╭ name: (anonymous)
+├ kind: class
+├ file: /main.ts
+├ span:
+│ ╭ /main.ts:3:25-3:30
+│ │ 3:     const badge = `<div class="badge">`;
```
jakebailey, line 7: How do we end up with quite literally `class` here?

Andarist: See the comment [here](https://github.com/microsoft/typescript-go/pull/2995/changes#diff-4d3d15765715351769a9224c7b5bf2fdadc12db22810ac17cda5297f52643b90R13-R19) and how it gets parsed here: [ts-ast-viewer](https://ts-ast-viewer.com/#code/GYVwdgxgLglg9mABCADgEwIZQKYCENoDm2AFAJSIDeAUInYhAgM5SIAW2B2ATogLyIABgB40MAG4MANhiZM+AIiYgARgoB8AEkrA43ALZZBAblr1GYFohUFi-IaImIA9ACoAjK+cQZcxTaJsDRNqAF8gA)

PR #2997, opened 2026-03-05
Fixed a find reference crash related to JSDoc function types in TS files

Andarist:


internal/checker/checker.go
```diff
@@ -16136,6 +16136,10 @@ func (c *Checker) getTypeForVariableLikeDeclaration(declaration *ast.Node, inclu
 		}
 	}
 	if ast.IsParameter(declaration) {
+		if declaration.Symbol() == nil {
+			// parameters of function types defined in JSDoc in TS files don't have symbols
+			return nil
+		}
```
jakebailey, line 16142: Indeed, same as strada:

```ts
        if (isParameter(declaration)) {
            if (!declaration.symbol) {
                // parameters of function types defined in JSDoc in TS files don't have symbols
                return;
            }
            const func = declaration.parent as FunctionLikeDeclaration;
            // For a parameter of a set accessor, use the type of the get accessor if one is present
            if (func.kind === SyntaxKind.SetAccessor && hasBindableName(func)) {
                const getter = getDeclarationOfKind<AccessorDeclaration>(getSymbolOfDeclarat
```

jakebailey (approved):

Andarist:
Hm, I have manually removed `fourslash.SkipIfFailing(t)` calls which is apperently not something I should do - that's why the CI failed on the previous commit. This has to be re-approved now

jakebailey:
Correct, the `gen` dir is 100% generated and should never be touched by hand. My bad for not noticing.

Issue #2935, opened 2026-02-28
Stack overflow in type instantiation with recursive function and generic type params


wadjoh:
## Description

A project at my job causes a stack overflow with `tsgo`, but compiles with `tsc`.
Tracked it down to a recursive object transformation function that changes the casing of the object's string properties.
It uses the `transform` function from lodash and the `CamelCase` type from type-fest.

I have a code example that's as minimal as I could figure out how to make it (including narrowing down and extracting the necessary lodash type): https://github.com/wadjoh/tsgo-crash-repro

## Stack trace

The full stack trace is large so I'll paste the first part of it here and paste a link to a gist with the full stack trace.

```go
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x44423d7f0570 stack=[0x44423d7f0000, 0x44425d7f0000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x104c6d915?, 0x1045baed4?})
        runtime/panic.go:1229 +0x38 fp=0x16b8768e0 sp=0x16b8768b0 pc=0x1046015d8
runtime.newstack()
        runtime/stack.go:1178 +0x498 fp=0x16b876a10 sp=0x16b8768e0 pc=0x1045e8648
runtime.morestack()
        runtime/asm_arm64.s:507 +0x70 fp=0x16b876a10 sp=0x16b876a10 pc=0x104607070
```

Full stack trace: https://gist.github.com/wadjoh/82adc287ab15357812c8e2da247163c3

## Steps to reproduce

1. Clone https://github.com/wadjoh/tsgo-crash-repro
2. Run `pnpm i`
3. Run `pnpm check:tsgo`
4. Observe stack overflow panic

You can also run `pnpm check:tsc` and see that it successfully compiles.

## Example code that causes the crash
```ts
import { CamelCase } from 'type-fest'

// type extracted from lodash `transform` function
// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3eefc1af52b4742eb3675be6b582d8e1ea92cb16/types/lodash/common/common.d.ts#L236
type MemoVoidDictionaryIterator<
  T,
  K extends string | number | symbol,
  TResult
> = (acc: TResult, curr: T, key: K, dict: Record<K, T>) => void

function transform<
  T extends object,
  TResult
>(
  object: T,
  iteratee: MemoVoidDictionaryIterator<T[keyof T], keyof T, TResult>,
  accumulator?: TResult
): TResult {
[...]

PR #2998, opened 2026-03-05
Fix crash in variance computation for recursive template literal types

ahejlsberg:
In #2935, the crash was caused by runaway recursion during variance computation for `CamelCase<T>`. However, the variance computation was kicked off by an unrelated issue caused by an incomplete port of a Strada PR, and once that unrelated issue was fixed, the repro in #2935 no longer caused a crash. But the crash can instead be triggered by adding

```ts
function boom<T>(x: CamelCase<T>, y: CamelCase<T, {}>) {
    x = y
    y = x
}
```

This PR fixes the runaway recursion and includes a regression test borrowed from #2940, but with the above code added.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -864,6 +864,7 @@ type Checker struct {
 	couldContainTypeVariables                   func(*Type) bool
 	isStringIndexSignatureOnlyType              func(*Type) bool
 	markNodeAssignments                         func(*ast.Node) bool
+	compareTypesAssignable                      TypeComparer
 	emitResolver                                *EmitResolver
 	emitResolverOnce                            sync.Once
 	_jsxNamespace                               string
@@ -1234,6 +1235,7 @@ func (c *Checker) initializeClosures() {
 	c.couldContainTypeVariables = c.couldContainTypeVariablesWorker
 	c.isStringIndexSignatureOnlyType = c.isStringIndexSignatureOnlyTypeWorker
 	c.markNodeAssignments = c.markNodeAssignmentsWorker
+	c.compareTypesAssignable = c.compareTypesAssignableWorker
 }
 
 func (c *Checker) initializeIterationResolvers() {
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -25134,7 +25219,7 @@ func (c *Checker) removeStringLiteralsMatchedByTemplateLiterals(types []*Type) [
 
 func (c *Checker) isTypeMatchedByTemplateLiteralOrStringMapping(t *Type, template *Type) bool {
 	if template.flags&TypeFlagsTemplateLiteral != 0 {
-		return c.isTypeMatchedByTemplateLiteralType(t, template.AsTemplateLiteralType())
+		return c.isTypeMatchedByTemplateLiteralType(t, template.AsTemplateLiteralType(), c.compareTypesAssignable)
```
jakebailey, line 25224: This will bind `c.compareTypesAssignable` each time this runs; you probably need to rename `compareTypesAssignable` to `compareTypesAssignableWorker`, then bind it in `NewChecker`.

ahejlsberg: Yeah, I was just thinking about that.

DanielRosenwasser: > This will bind `c.compareTypesAssignable` each time this runs

But can the compiler optimize that onto the stack anyway? `compareTypes` doesn't escape in `isTypeMatchedByTemplateLiteralType`.

ahejlsberg: It's not really an issue of escaping, rather it's an issue of needing a closure that provides the `*Checker` parameter for the closed-over method.

ahejlsberg: Oh, I see you mean the compiler could allocate the closure on the stack. Yeah, I don't think that happens.

jakebailey: It can if the compiler can prove that it doesn't escape, but I'm not about to try and -m=2 the whole package 😅

PR #3000, opened 2026-03-05
Fixed a range formatting crash after template literal

Andarist:
Fixes a crash found here: https://github.com/microsoft/typescript-go/issues/2988#issuecomment-4001335883

### DIFF
--- a/internal/format/scanner.go
+++ b/internal/format/scanner.go
@@ -42,7 +42,6 @@ type formattingScanner struct {
 
 func newFormattingScanner(text string, languageVariant core.LanguageVariant, startPos int, endPos int, worker *formatSpanWorker) []core.TextChange {
 	scan := scanner.NewScanner()
-	scan.Reset()
 	scan.SetSkipTrivia(false)
 	scan.SetLanguageVariant(languageVariant)
 	scan.SetText(text)


internal/fourslash/tests/formatSelectionAfterTemplateLiteral1_test.go
Andarist, line 1: this one comes from the crash

internal/format/scanner.go
DanielRosenwasser: Did we ever have to call `Reset` on a NewScanner() object?

jakebailey: No, there's no reason to. This could be removed.

DanielRosenwasser: Any why do we call `Reset` twice? (once more below)

Andarist: good catch! I just removed it, @jakebailey could u re-add this to the merge queue?

jakebailey (approved):

internal/format/scanner.go
```diff
@@ -42,10 +42,10 @@ type formattingScanner struct {
 
 func newFormattingScanner(text string, languageVariant core.LanguageVariant, startPos int, endPos int, worker *formatSpanWorker) []core.TextChange {
 	scan := scanner.NewScanner()
-	scan.Reset()
```
DanielRosenwasser, line 45: There's another call to `Reset` below too

Andarist: already removed :)

internal/format/scanner.go
jakebailey: In general I dislike how `NewScanner` operates... we should really take in an objects struct and not modify bits and bobs just to set it up... Not for this PR though.

internal/format/scanner.go
jakebailey: Yeah, since this is made fresh each time, we don't have to reset it later anymore.

That being said, probably we should be pooling these in the future.

Andarist: but @jakebailey approved in between the pushes so it has to be re-approved again 🫠

internal/format/scanner.go
```diff
@@ -42,10 +42,10 @@ type formattingScanner struct {
...
 	scan.SetSkipTrivia(false)
 	scan.SetLanguageVariant(languageVariant)
 	scan.SetText(text)
+	scan.ResetTokenState(startPos)
```
DanielRosenwasser, line 48: So it sounds like range formatting was always kind of broken because we were starting at 0

internal/format/scanner.go
```diff
@@ -57,7 +57,6 @@ func newFormattingScanner(text string, languageVariant core.LanguageVariant, sta
 	res := worker.execute(fmtScn)
 
 	fmtScn.hasLastTokenInfo = false
-	scan.Reset()
```
DanielRosenwasser, line 60: So I actually take this back - `worker.execute` actually does stash away this exact scanner onto `worker`. I don't have enough context right now to say whether this is the right thing to do, and the old code in Strada does this, so I'd say leave it alone/revert this specific change for now.

jakebailey: Oh, yucky....

DanielRosenwasser: On the other hand, `newFormatSpanWorker` is only ever passed in directly to the only calls of `newFormattingScanner`. The way this stuff is wired up is kind of bananas - an object is constructed in an intermediate state, only to be passed into another call that gives the rest of its state. I think it's worth a follow-up refactoring. Let me know if you're interested in doing that, but let's keep things simple here I guess.

Andarist: Ok, now I kept this one but I'm still removing the other call as that seemed to be entirely pointless when reset happened right after construction.

DanielRosenwasser (approved):

PR #3002, opened 2026-03-06
Finish porting reportMergeSymbolError

jakebailey:
We don't need any of the commented code (now that we consistently report these things for concurrency reasons). But, we did need the `IsPlainJSFile` checks.

Deletes one diff. No other diffs remain with the errors that are reported by this code.

With #3001, there are no more `TODO(TS-TO-GO)` comments left in the repo! (that means we're done, right?)


connorshea:
> that means we're done, right?

Close the issue tracker, we did it! 🥳

PR #3004, opened 2026-03-06
Fixed parsing of multiline string JSX attributes with whitespace present after `=`

Andarist:
fixes the crash found here https://github.com/microsoft/typescript-go/issues/2988#issuecomment-4001338397

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -1255,12 +1255,12 @@ func (s *Scanner) ScanJsxIdentifier() ast.Kind {
 }
 
 func (s *Scanner) ScanJsxAttributeValue() ast.Kind {
-	// Skip whitespace between '=' and the value so fullStartPos lands on the
+	s.fullStartPos = s.pos
+	// Skip whitespace between '=' and the value so tokenStart lands on the
 	// opening quote, not on trivia.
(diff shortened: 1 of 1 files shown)


internal/scanner/scanner.go
```diff
@@ -1255,7 +1255,13 @@ func (s *Scanner) ScanJsxIdentifier() ast.Kind {
 }
 
 func (s *Scanner) ScanJsxAttributeValue() ast.Kind {
+	// Skip whitespace between '=' and the value so fullStartPos lands on the
+	// opening quote, not on trivia.
+	for ch, size := s.charAndSize(); size > 0 && stringutil.IsWhiteSpaceLike(ch); ch, size = s.charAndSize() {
+		s.pos += size
+	}
```
Andarist: This is mishandled by Strada and also leads to a crash in the formatting case.

internal/scanner/scanner.go
```diff
@@ -1255,7 +1255,13 @@ func (s *Scanner) ScanJsxIdentifier() ast.Kind {
...
 	s.fullStartPos = s.pos
```
gabritto: I think `fullStartPos` needs to include the preceding whitespace, so I think we should move this to before we skip the whitespace.

Andarist: fixed, good catch

PR #3005, opened 2026-03-06
Allow start and end positions to be in different files in `getMappedLocation`

Andarist:
fixes the crash found here: https://github.com/microsoft/typescript-go/issues/2988#issuecomment-4001337109


jakebailey:
I think this is the same bug as https://github.com/microsoft/typescript-go/pull/2419

jakebailey:
Can you pull that PR's test into this one and see if it now passes?

Andarist:
@jakebailey it doesn't crash here (but perhaps it didn't before Seth's PR either) - but the baseline result is different. I think both PRs have to be considered separately

PR #3006, opened 2026-03-06
Fixed a crash related to a dynamic `import()` of an UMD module

Andarist:
fixes a crash found here https://github.com/microsoft/typescript-go/issues/2988#issuecomment-4001335927

### DIFF
--- a/internal/ls/importTracker.go
+++ b/internal/ls/importTracker.go
@@ -171,7 +171,11 @@ func getImportersForExport(
 	// Adds a module and all of its transitive dependencies as possible indirect users
 	var addIndirectUser func(*ast.Node, bool)
 	addIndirectUser = func(sourceFileLike *ast.Node, addTransitiveDependencies bool) {
-		debug.Assert(!isAvailableThroughGlobal)
+		// When isAvailableThroughGlobal, getIndirectUsers already returns all source files,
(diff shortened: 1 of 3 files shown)


internal/ls/importTracker.go
```diff
@@ -171,7 +171,11 @@ func getImportersForExport(
 	// Adds a module and all of its transitive dependencies as possible indirect users
 	var addIndirectUser func(*ast.Node, bool)
 	addIndirectUser = func(sourceFileLike *ast.Node, addTransitiveDependencies bool) {
-		debug.Assert(!isAvailableThroughGlobal)
+		// When isAvailableThroughGlobal, getIndirectUsers already returns all source files,
+		// so indirectUserDeclarations is never consulted. Nothing to do here.
+		if isAvailableThroughGlobal {
+			return
+		}
```
jakebailey, line 178: Why doesn't strada hit this, then? Given this is a pure port:

```ts
Debug.assert(!isAvailableThroughGlobal);
```

Andarist: IIRC - it hits this. I can doublecheck this later. In general, a good chunk of my recent crash fixes are for crashes impacting Strada too

Andarist: Yep, it crashes in Strada! :)

jakebailey (approved):

PR #3007, opened 2026-03-06
Speed up iofs file writes, fix BOM

jakebailey:
Closes #1551

While looking at #1551, I realized that the signature could just change. Then, reading further exposed the fact that our OS FS was not writing the correct BOM when asked. So, fix that too.

### DIFF
--- a/internal/vfs/vfstest/vfstest.go
+++ b/internal/vfs/vfstest/vfstest.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"testing/fstest"
 	"time"
+	"unsafe"
 
 	"github.com/microsoft/typescript-go/internal/tspath"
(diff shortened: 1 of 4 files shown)


internal/vfs/vfstest/vfstest.go
```diff
@@ -526,7 +527,7 @@ func (m *MapFS) WriteFile(path string, data []byte, perm fs.FileMode) error {
 	}
 
 	m.setEntry(path, cp, fstest.MapFile{
-		Data:    data,
+		Data:    unsafe.Slice(unsafe.StringData(data), len(data)),
```
jakebailey, line 530: This is no more unsafe than it used to be, given we'd just take in mutable bytes before.

jakebailey:
Actually, I have another change I want to do instead (dropping writeByteOrderMark entirely)

PR #3013, opened 2026-03-07
Report errors on unsupported CJS constructs in declaration emit

ahejlsberg:
With this PR we report errors in the following cases when generating declaration files from CommonJS modules:

* Multiple `module.exports` assignments.
* Nested `module.exports.XXX` assignments.
* Nested `exports.XXX` assignments.
* Nested `Object.defineProperty(module.exports, ...)` calls.
* Nested `Object.defineProperty(exports, ...)` calls.

The potential error locations for nested CJS constructs are collected by the binder in a new `SourceFile.NestedCJSExports` list such that the declaration transformer doesn't have to visit subtrees.

### DIFF
--- a/internal/transformers/declarations/transform.go
+++ b/internal/transformers/declarations/transform.go
@@ -266,10 +266,12 @@ func (tx *DeclarationTransformer) transformSourceFile(node *ast.SourceFile) *ast
 	if ast.IsExternalOrCommonJSModule(node) {
 		if ast.IsInJSFile(node.AsNode()) {
 			if exportEquals := node.Symbol.Exports[ast.InternalSymbolNameExportEquals]; exportEquals != nil && len(exportEquals.Declarations) > 1 {
-				tx.tracker.ReportMultipleModuleExports(exportEquals.Declarations)
+				for _, node := range exportEquals.Declarations {
+					tx.state.addDiagnostic(createDiagnosticForNode(node, diagnostics.Multiple_module_exports_assignments_cannot_be_serialized_for_declaration_emit))
+				}
 			}
-			if len(node.NestedCJSExports) != 0 {
-				tx.tracker.ReportNestedCJSExports(node.NestedCJSExports)
+			for _, node := range node.NestedCJSExports {
+				tx.state.addDiagnostic(createDiagnosticForNode(node, diagnostics.Nested_CommonJS_export_constructs_cannot_be_serialized_for_declaration_emit))
 			}
 		}
 		if !tx.resultHasExternalModuleIndicator || (tx.needsScopeFixMarker && !tx.resultHasScopeMarker) {
--- a/internal/checker/symboltracker.go
+++ b/internal/checker/symboltracker.go
@@ -103,22 +103,6 @@ func (this *SymbolTrackerImpl) ReportNonSerializableProperty(propertyName string
 	this.inner.ReportNonSerializableProperty(propertyName)
 }
 
-func (this *SymbolTrackerImpl) ReportMultipleModuleExports(nodes []*ast.Node) {
-	this.onDiagnosticReported()
-	if this.inner == nil {
-		return
-	}
-	this.inner.ReportMultipleModuleExports(nodes)
-}
-
-func (this *SymbolTrackerImpl) ReportNestedCJSExports(nodes []*ast.Node) {
-	this.onDiagnosticReported()
-	if this.inner == nil {
-		return
-	}
-	this.inner.ReportNestedCJSExports(nodes)
-}
-
 func (this *SymbolTrackerImpl) onDiagnosticReported() {
 	this.context.reportedDiagnostic = true
 }
--- a/internal/nodebuilder/types.go
+++ b/internal/nodebuilder/types.go
@@ -16,8 +16,6 @@ type SymbolTracker interface {
 	ReportTruncationError()
 	ReportNonlocalAugmentation(containingFile *ast.SourceFile, parentSymbol *ast.Symbol, augmentingSymbol *ast.Symbol)
 	ReportNonSerializableProperty(propertyName string)
-	ReportMultipleModuleExports(nodes []*ast.Node)
-	ReportNestedCJSExports(nodes []*ast.Node)
 
 	ReportInferenceFallback(node *ast.Node)
 	PushErrorFallbackNode(node *ast.Node)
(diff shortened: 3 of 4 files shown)


internal/diagnostics/extraDiagnosticMessages.json
```diff
@@ -63,6 +63,14 @@
         "category": "Message",
         "code": 6423
     },
+    "A module with multiple 'module.exports' assignments cannot be serialized.": {
+        "category": "Error",
+        "code": 6424
+    },
+    "A module with nested CommonJS export constructs cannot be serialized.": {
```
DanielRosenwasser: Small suggestion

```suggestion
    "Multiple 'module.exports' assignments cannot be serialized for declaration emit.": {
        "category": "Error",
        "code": 6424
    },
    "Nested CommonJS export constructs cannot be serialized for declaration emit.": {
```

This makes it a little bit more obvious that this is an error from `--declarations` (something that might not be obvious to a JavaScript developer).

internal/binder/binder.go
```diff
@@ -879,6 +883,12 @@ func (b *Binder) bindExportAssignment(node *ast.Node) {
 	}
 }
 
+func (b *Binder) checkNestedExport(node *ast.Node) {
```
DanielRosenwasser: Maybe


```suggestion
func (b *Binder) trackNestedExport(node *ast.Node) {
```

DanielRosenwasser: or `checkForNestedExport`

internal/diagnostics/extraDiagnosticMessages.json
ahejlsberg: Now fixed.

internal/binder/binder.go
ahejlsberg: Now fixed.

ahejlsberg:
> though I'd rather get a second pair of eyes to take a look at the symbol tracker stuff

The only change there is the addition of two error reporting methods (the symbol tracker is the conduit for error reporting in the declaration transformer).

testdata/baselines/reference/submodule/compiler/amdLikeInputDeclarationEmit.errors.txt
```diff
@@ -0,0 +1,34 @@
...
+            f: function() {
+                return "something";
+            }
+        });
+    
+        // Exports the module in a way tsc recognize class export 
+        const module = {};
+        module.exports = ExtendedClass
+        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS6425: Nested CommonJS export constructs cannot be serialized for declaration emit.
```
jakebailey, line 32: Isn't this not the same `module`? I guess the comment above says that we treat it as such (yay...)

ahejlsberg: Yeah, we don't resolve the `module` in `module.exports` assignments to ensure it isn't a local. The parser isn't really equipped to do that, and that's where the re-parsing into `ExportAssignment` nodes happens.

internal/transformers/declarations/transform.go
```diff
@@ -263,12 +263,22 @@ func (tx *DeclarationTransformer) transformSourceFile(node *ast.SourceFile) *ast
...
-	if ast.IsExternalOrCommonJSModule(node) && (!tx.resultHasExternalModuleIndicator || (tx.needsScopeFixMarker && !tx.resultHasScopeMarker)) {
-		marker := createEmptyExports(tx.Factory().AsNodeFactory())
-		newList := append(combinedStatements.Nodes, marker)
-		withMarker := tx.Factory().NewNodeList(newList)
-		withMarker.Loc = combinedStatements.Loc
-		combinedStatements = withMarker
+	if ast.IsExternalOrCommonJSModule(node) {
+		if ast.IsInJSFile(node.AsNode()) {
+			if exportEquals := node.Symbol.Exports[ast.InternalSymbolNameExportEquals]; exportEquals != nil && len(exportEquals.Declarations) > 1 {
+				tx.tracker.ReportMultipleModuleExports(exportEquals.Declarations)
```
weswigham: A comment elsewhere asking why this was done via the symbol tracker was pertinent - the symbol tracker is the error reporter _for the node builder_ (so the checker node creation context has access to the declaration diagnostic collection) - for inferred type nodes. The declaration transform can issue diagnostics directly. This call to the tracker and the one below can be inlined and report diagnostics directly via `tx.state.addDiagnostic` (many ID diagnostics ultimately do this) - no need to defer through the tracker like this and add a bunch of methods to its' type.

jakebailey: Yeah, I don't think we need the symbol tracker for this

ahejlsberg: The declaration transformer currently doesn't call `tx.state.addDiagnostic` directly anywhere, but for example has an indirection to `tx.tracker.ReportInferenceFallback(node)` to report errors. And that method doesn't appear to be called by anyone but the declaration transformer, so I figured that was the desired pattern.

But happy to just inline instead.

internal/binder/binder.go
```diff
@@ -879,6 +883,12 @@ func (b *Binder) bindExportAssignment(node *ast.Node) {
 	}
 }
 
+func (b *Binder) trackNestedCJSExport(node *ast.Node) {
+	if !(ast.IsSourceFile(node.Parent) || ast.IsExpressionStatement(node.Parent) && ast.IsSourceFile(node.Parent.Parent)) {
+		b.nestedCJSExports = append(b.nestedCJSExports, node)
+	}
```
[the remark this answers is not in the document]
ahejlsberg: The parser sets the `Parent` property so we're fine here.

internal/transformers/declarations/transform.go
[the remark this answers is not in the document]
ahejlsberg: When `ast.IsExternalOrCommonJSModule` returns true, the source file is known to have a symbol assigned by the binder.

weswigham: Yeah, in that case, `ReportInferenceFallback` is called by the node builder, so it also calls it when it wants to make the same error, rather than duplicating the diagnostic logic, since it *does* have access to the tracker.

ahejlsberg: > Yeah, in that case, ReportInferenceFallback is called by the node builder

Is it? I might be missing something, but I'm not seeing that anywhere.

PR #3016, opened 2026-03-07
fix: move GetLanguageService to async closure to prevent dispatch loop blocking

SibtainOcn:
.


jakebailey:
Does this not just undo #2765?

PR #3017, opened 2026-03-07
Fixed a rescanning crash related to `JsxNamespacedName` with dashes

Andarist:
fixes a crash found here: https://github.com/microsoft/typescript-go/issues/3012#issuecomment-4015173297 (this also crashes in Strada)


DanielRosenwasser:
Thank you!

PR #3021, opened 2026-03-08
Fixed a format selection crash in reparsed JSDoc nodes

Andarist:
Fixes a crash found here: https://github.com/microsoft/typescript-go/issues/3012#issuecomment-4015173459

### DIFF
--- a/internal/fourslash/tests/formatSelectionInJSDocTypeLiteralNoCrash1_test.go
+++ b/internal/fourslash/tests/formatSelectionInJSDocTypeLiteralNoCrash1_test.go
@@ -25,5 +25,14 @@ func TestFormatSelectionInJSDocTypeLiteralNoCrash1(t *testing.T) {
 	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
 	defer done()
 	f.FormatSelection(t, "begin", "end")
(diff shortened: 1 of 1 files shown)


internal/format/span.go
```diff
@@ -22,6 +22,9 @@ func findEnclosingNode(r core.TextRange, sourceFile *ast.SourceFile) *ast.Node {
 	find = func(n *ast.Node) *ast.Node {
 		var candidate *ast.Node
 		n.ForEachChild(func(c *ast.Node) bool {
+			if c.Flags&ast.NodeFlagsReparsed != 0 {
```
Andarist, line 25: from the looks of it formatting is, in general, meant to ignore reparsed nodes - and similar skips can be found in other places (like in astnav functions)

DanielRosenwasser: Should formatting avoid operating on the reparsed tree in the first place?

Andarist: maybe not but I can't find any evidence of JSDocs being formatted in Strada, so I don't think it's worth considering right now

jakebailey: We shouldn't operate on the reparsed part at all, no, given that's not the "concrete" part of the AST. If we wanted to format JSDoc (I don't think so), we would walk into `JSDoc()`.

Issue #3018, opened 2026-03-07
No quick info for properties accessed via index signature


DanielRosenwasser:
```ts
declare let myDictionary: Record<string, string>;
myDictionary./*$*/SOME_VARIABLE
```

Request quick info at the marker (on `SOME_VARIABLE`).

**Current Behavior in 7.0**: no quick info

**Expected Behavior/Behavior in 6.0**: `string | undefined`

ahejlsberg:
I will be putting up a PR this fixes this as I describe [here](https://github.com/microsoft/typescript-go/pull/3022#issuecomment-4025483658).

PR #3022, opened 2026-03-08
Provide type-based quick info on symbol-less nodes

Andarist:


testdata/baselines/reference/fourslash/quickInfo/quickInfoLink9.baseline
```diff
@@ -5,7 +5,10 @@
 //      * Text before {@link a} text after
 //                           ^
 // | ----------------------------------------------------------------------
-// | No quickinfo at /**/.
+// | ```tsx
+// | any
```
Andarist, line 9: this one doesn't resolve correctly but it's a separate issue

ahejlsberg:
Unfortunately, this isn't the right way to fix the issue. We already have infrastructure in place that creates synthetic symbols for index-signature-based property and element accesses (see `getApplicableIndexSymbol`), but it isn't working right and it also is odd in that it creates symbols whose type you can't obtain using `getTypeOfSymbol`. I will be putting up a PR that fixes this.

jakebailey:
is this still needed after #3018?

Andarist:
This is kind of a pure port from Strada - you can see how a bunch of `failingTests` are removed here. This branch is already based now on top of https://github.com/microsoft/typescript-go/issues/3018 - so the diff you see is not outdated or anything like that.

ahejlsberg:
Hmm, as I said above, I don't think this code is right--even if it was in Strada. Some of those tests change quick info on an error location from displaying nothing to displaying `any` (which is actually the `errorType`). How is that better?

Andarist:
As a user, that displayed `any` is what I expect - it might be an acquired expectation but now it would feel slightly off to me if those wouldn't display anything.

jakebailey:
Yes, I would tend to agree, that if we are hovering on an identifier-y thing, we need to display something...

ahejlsberg:
I gotta say I disagree here. If the type returned from `GetTypeAtLocation` is the error type (meaning there's no type), how are we helping anyone by displaying `any`? Which is indistinguishable from an actual non-error `any`.

Andarist:
To some extent, it's about cross-location consistency:
```ts
function test(o: unknown) {
    //   ^? function test(o: unknown): any
    return o.foo
    //       ^? any
}
```

PR #3024, opened 2026-03-08
Fixes a crash when requesting implementations for a module referenced by a tripleslash directive

Andarist:
fixes a crash reported here: https://github.com/microsoft/typescript-go/issues/3012#issuecomment-4015173326

note: Strada also crashes here

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -621,10 +621,7 @@ func (l *LanguageService) provideSymbolsAndEntries(ctx context.Context, uri lspr
 
 		entry := queue[0]
 		queue = queue[1:]
-		if entry.node == nil {
-			continue
-		}
-		if !seenNodes.Has(entry.node) {
+		if entry.node != nil && !seenNodes.Has(entry.node) {
 			seenNodes.Add(entry.node)
 			addToQueue(l.getSymbolAndEntries(ctx, entry.node.Pos(), entry.node, program, isRename, implementations))
 		}


internal/ls/findallreferences.go
```diff
@@ -621,6 +621,9 @@ func (l *LanguageService) provideSymbolsAndEntries(ctx context.Context, uri lspr
 
 		entry := queue[0]
 		queue = queue[1:]
+		if entry.node == nil {
```
Andarist: note: tripleslash directives are found by: `getReferencedSymbolsForModuleIfDeclaredBySourceFile` -> `getReferencedSymbolsForModule` -> `findModuleReferences`

internal/ls/findallreferences.go
```diff
@@ -621,7 +621,7 @@ func (l *LanguageService) provideSymbolsAndEntries(ctx context.Context, uri lspr
 
 		entry := queue[0]
 		queue = queue[1:]
-		if !seenNodes.Has(entry.node) {
+		if entry.node != nil && !seenNodes.Has(entry.node) {
 			seenNodes.Add(entry.node)
```
jakebailey, line 625: Does this imply that we could end up looping forever on an entry without a node?

Andarist: Hm, looking from mobile that doesnt seem to be too plausible. Without the node it wont get again to `addToQueue`.

Maaayyyybe somehow it could process the same kind of entry if it would be readed by an entry with a node but then that doesnt seem to be different from the situation guarded by the seenNodes

jakebailey: I misread. Obviously if this if isn't true, `getSymbolsAndEntries` is not called on the entry again.

jakebailey (approved):

PR #3030, opened 2026-03-09
docs: correct setting key for Tsgo in README

9romise:


9romise:

PR #3033, opened 2026-03-09
Fixed a formatting crash caused by misinterpretation of tokens in reparsed node lists

Andarist:
fixes the crash reported here: https://github.com/microsoft/typescript-go/issues/3025#issuecomment-4020210407

### DIFF
--- a/internal/ast/ast.go
+++ b/internal/ast/ast.go
@@ -179,7 +179,6 @@ func cloneNode(updated *Node, original *Node, hooks NodeFactoryHooks) *Node {
 
 type NodeList struct {
 	Loc   core.TextRange
-	Flags NodeFlags
 	Nodes []*Node
 }
 
--- a/internal/format/span.go
+++ b/internal/format/span.go
@@ -447,7 +447,7 @@ func (w *formatSpanWorker) processChildNodes(
 
 	// node range is outside the target range - do not dive inside
 	if !w.originalRange.Overlaps(nodes.Loc) {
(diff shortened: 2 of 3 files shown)


internal/format/span.go
```diff
@@ -447,7 +447,7 @@ func (w *formatSpanWorker) processChildNodes(
 
 	// node range is outside the target range - do not dive inside
 	if !w.originalRange.Overlaps(nodes.Loc) {
-		if nodes.End() < w.originalRange.Pos() {
+		if nodes.End() < w.originalRange.Pos() && nodes.Flags&ast.NodeFlagsReparsed == 0 {
```
Andarist: 1. this made `skipToEndOf` to be called on a reparsed node list
2. then the scanner was positioned to it
3. and later on executed from that position without knowing it's positioned in a comment
4. so it started interpreting everything as JS tokens
5. it consumed a backtick and tried to produce a template literal and advanced **past** the comment 
6. that ultimately led to a hard crash

internal/ast/ast.go
```diff
@@ -179,6 +179,7 @@ func cloneNode(updated *Node, original *Node, hooks NodeFactoryHooks) *Node {
 
 type NodeList struct {
 	Loc   core.TextRange
+	Flags NodeFlags
```
jakebailey: We can't really casually add fields to types like this; this struct is 32 bytes so neatly fits in a size class, so adding this flags field will bump it to the 48 byte size class.

Can this check not be done by just looking at the first entry in the list? I believe we've done that before.

Andarist: that works too, I pushed out this change, thanks!

PR #3038, opened 2026-03-09
Delete failed lookup locations

andrewbranch:
Failed lookup locations store the module resolution failures for _every import_, but we never use that granular information—we dedupe at the per-project level in the LSP to form a small set of globs to watch and then match the watch events against the deduped set to invalidate the program. That leaves potentially quite a lot of useless strings in memory. In the LSP, we were already wrapping each program's CompilerHost in a thing that tracks every requested file during construction for file cache management, making failed lookup locations and affecting locations completely superfluous.

This saves 0.8% memory on `tsgo vscode/src/tsconfig.json`, but 22% memory in DefinitelyTyped-tools with every workspace package open.

### DIFF
--- a/internal/project/watch.go
+++ b/internal/project/watch.go
@@ -176,8 +176,13 @@ func createResolutionLookupGlobMapper(workspaceDirectory string, libDirectory st
 		if includeLib {
 			globs = append(globs, getRecursiveGlobPattern(string(libDirectoryPath)))
 		}
-		for dir := range nodeModulesDirectories.Keys() {
-			globs = append(globs, getRecursiveGlobPattern(string(dir)))
+		if nodeModulesDirectories.Len() > 0 {
+			nodeModulesGlobs := make([]string, 0, nodeModulesDirectories.Len())
+			for dir := range nodeModulesDirectories.Keys() {
+				nodeModulesGlobs = append(nodeModulesGlobs, getRecursiveGlobPattern(string(dir)))
+			}
+			slices.Sort(nodeModulesGlobs)
+			globs = append(globs, nodeModulesGlobs...)
 		}
 		if externalDirectories.Len() > 0 {
 			externalDirStrings := make([]string, 0, externalDirectories.Len())
--- a/internal/tspath/path.go
+++ b/internal/tspath/path.go
@@ -993,6 +993,9 @@ func ContainsPath(parent string, child string, options ComparePathsOptions) bool
 // Since Path values are already rooted, reduced, and case-canonicalized,
 // this is a simple string prefix check.
 func (p Path) ContainsPath(child Path) bool {
+	if len(p) == 0 {
+		return false
+	}
 	return p == child || len(child) > len(p) && strings.HasPrefix(string(child), string(p)) && (p[len(p)-1] == '/' || child[len(p)] == '/')
 }
 


internal/tspath/path.go
```diff
@@ -989,8 +989,11 @@ func ContainsPath(parent string, child string, options ComparePathsOptions) bool
 	return true
 }
 
+// ContainsPath checks whether child is contained within or equal to p.
+// Since Path values are already rooted, reduced, and case-canonicalized,
+// this is a simple string prefix check.
 func (p Path) ContainsPath(child Path) bool {
-	return ContainsPath(string(p), string(child), ComparePathsOptions{UseCaseSensitiveFileNames: true})
+	return p == child || len(child) > len(p) && strings.HasPrefix(string(child), string(p)) && (p[len(p)-1] == '/' || child[len(p)] == '/')
```
andrewbranch, line 999: Bit of a drive-by optimization since watch glob building changed to operate over already-normalized tspath.Paths

internal/project/watch.go
```diff
@@ -132,65 +130,65 @@ func (w *WatchedFiles[T]) Clone(input T) *WatchedFiles[T] {
...
+func createResolutionLookupGlobMapper(workspaceDirectory string, libDirectory string, currentDirectory string, useCaseSensitiveFileNames bool) func(data *collections.SyncSet[tspath.Path]) PatternsAndIgnored {
+	workspaceDirectoryPath := tspath.ToPath(workspaceDirectory, currentDirectory, useCaseSensitiveFileNames)
+	currentDirectoryPath := tspath.ToPath(currentDirectory, currentDirectory, useCaseSensitiveFileNames)
+	libDirectoryPath := tspath.ToPath(libDirectory, currentDirectory, useCaseSensitiveFileNames)
 
-	return func(data map[tspath.Path]string) PatternsAndIgnored {
+	return func(data *collections.SyncSet[tspath.Path]) PatternsAndIgnored {
 		var ignored map[string]struct{}
-		var seenDirs collections.Set[string]
+		var seenDirs collections.Set[tspath.Path]
```
jakebailey, line 140: Maybe I'm reading this wrong, but what about support for case sensitive FSs? Aren't we eliminating all of the code that would have preserved case, given everything appears to now be derive from a Path?

andrewbranch: The value stored in a `tspath.Path` _is_ the original case on case-sensitive file systems.

jakebailey: Agh, I don't know why I thought that `Path` was always lowercase

internal/tspath/path.go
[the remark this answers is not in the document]
jakebailey: This sounds plausibly correct

internal/project/watch.go
```diff
@@ -132,65 +130,65 @@ func (w *WatchedFiles[T]) Clone(input T) *WatchedFiles[T] {
...
 		}
 		if includeLib {
-			globs = append(globs, getRecursiveGlobPattern(libDirectory))
+			globs = append(globs, getRecursiveGlobPattern(string(libDirectoryPath)))
 		}
-		for _, dir := range nodeModulesDirectories {
-			globs = append(globs, getRecursiveGlobPattern(dir))
+		for dir := range nodeModulesDirectories.Keys() {
+			globs = append(globs, getRecursiveGlobPattern(string(dir)))
 		}
```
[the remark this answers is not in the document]
jakebailey: This was preexisting, but does sound correct; I guess we only partially sort some elements below?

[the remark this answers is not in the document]
andrewbranch: Yeah, this might have been an unnoticed regression when I did the refactor to minimize watch locations.

PR #3053, opened 2026-03-10
port linked editing

iisaduan:
Ports linked editing and the fourslash tests (both regular and baseline)

### DIFF
--- a/testdata/baselines/reference/submodule/fourslash/linkedEditing/linkedEditingJsxTag10.linkedEditing.txt
+++ b/testdata/baselines/reference/submodule/fourslash/linkedEditing/linkedEditingJsxTag10.linkedEditing.txt
@@ -1,80 +1,112 @@
+// === Linked Editing ===
 === /jsx0.tsx ===
 const jsx = <>
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx1.tsx ===
 const jsx = </>
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx2.tsx ===
 const jsx = <div>
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx3.tsx ===
 const jsx = </div>
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx4.tsx ===
 const jsx = <div> </>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx5.tsx ===
 const jsx = <> </div>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx6.tsx ===
 const jsx = div> </div>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx7.tsx ===
 const jsx = <div> /div>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx8.tsx ===
 const jsx = <div </div>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx9.tsx ===
 const jsx = <div> </div;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx10.tsx ===
 const jsx = <> </;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx11.tsx ===
 const jsx = < </>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx12.tsx ===
 const jsx = > </>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx13.tsx ===
 const jsx = <> />;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx14.tsx ===
 const jsx = <> <div> </> </div>;
 
 --No linked edits found--
 
+
+// === Linked Editing ===
 === /jsx15.tsx ===
 const jsx = <div> <> </div> </>;
 
 --No linked edits found--
+
 
--- a/internal/fourslash/baselineutil.go
+++ b/internal/fourslash/baselineutil.go
@@ -32,6 +32,7 @@ const (
 	inlayHintsCmd               baselineCommand = "Inlay Hints"
 	nonSuggestionDiagnosticsCmd baselineCommand = "Syntax and Semantic Diagnostics"
 	quickInfoCmd                baselineCommand = "QuickInfo"
+	linkedEditingCmd            baselineCommand = "linkedEditing"
 	renameCmd                   baselineCommand = "findRenameLocations"
 	signatureHelpCmd            baselineCommand = "SignatureHelp"
 	smartSelectionCmd           baselineCommand = "Smart Selection"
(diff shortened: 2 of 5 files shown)


internal/ast/utilities.go
```diff
@@ -993,6 +993,26 @@ func IsNodeDescendantOf(node *Node, ancestor *Node) bool {
...
+
+		// If so, mark ourselves accordingly.
+		if thisNodeOrAnySubNodesHasError {
+			node.Flags |= NodeFlagsThisNodeOrAnySubNodesHasError
+		}
+
+		// Also mark that we've propagated the child information to this node.  This way we can
+		// always consult the bit directly on this node without needing to check its children
+		// again.
+		node.Flags |= NodeFlagsHasAggregatedChildData
```
jakebailey: We can't do this, as it will break in race mode. The flags must never be mutated after the bind phase.

We should move `NodeFlagsThisNodeOrAnySubNodesHasError` into a `SubtreeFact`, remove `NodeFlagsHasAggregatedChildData`, and then use the subtree facts instead.

internal/fourslash/_scripts/manualTests.txt
```diff
@@ -58,6 +58,14 @@ jsdocDeprecated_suggestion2
...
 jsDocFunctionSignatures2
 jsdocParameterNameCompletion
+linkedEditingJsxTag1
+linkedEditingJsxTag12
+linkedEditingJsxTag2
+linkedEditingJsxTag4
+linkedEditingJsxTag5
+linkedEditingJsxTag6
+linkedEditingJsxTag7
+linkedEditingJsxTag9
```
jakebailey, line 68: Can you explain the changes you made that required them to be manual?

internal/ls/linkedediting.go
```diff
@@ -0,0 +1,106 @@
...
+	_, sourceFile := l.getProgramAndFile(params.TextDocument.Uri)
+	position := l.converters.LineAndCharacterToPosition(sourceFile, params.Position)
+	token := astnav.FindPrecedingToken(sourceFile, int(position))
+
+	if token == nil || token.Parent.Kind == ast.KindSourceFile {
+		return lsproto.LinkedEditingRangeResponse{}, nil
+	}
+
+	// allow the client to match more than valid tag names. This allows linked editing when typing is in progress or tag name is incomplete
+	jsxTagWordPattern := "[a-zA-Z0-9:\\-\\._$]*"
```
jakebailey: Doubt it matters much, but you could just make this a global `var jsxTagWordPattern = new("[a-zA-Z0-9:\\-\\._$]*")`.

internal/ls/linkedediting.go
```diff
@@ -0,0 +1,106 @@
...
+		tag := ast.FindAncestor(token.Parent, func(n *ast.Node) bool {
+			if ast.IsJsxOpeningElement(n) || ast.IsJsxClosingElement(n) {
+				return true
+			}
+			return false
+		})
+		if tag == nil {
+			return lsproto.LinkedEditingRangeResponse{}, nil
+		}
+		debug.Assert(ast.IsJsxOpeningElement(tag) || ast.IsJsxClosingElement(tag), "tag should be opening or closing element")
```
jakebailey, line 64: Maybe this is a pure port, but doesn't this logically follow from FindAncestor above?

internal/ls/linkedediting.go
```diff
@@ -0,0 +1,106 @@
...
+
+		jsxElement := tag.Parent.AsJsxElement()
+		openTag := jsxElement.OpeningElement
+		closeTag := jsxElement.ClosingElement
+
+		openTagNameStart := core.TextPos(astnav.GetStartOfNode(openTag.TagName().AsNode(), sourceFile, false))
```
jakebailey: Why so much `core.TextPos` conversion?

internal/fourslash/_scripts/manualTests.txt
iisaduan: The strada tests had a lot of variables (`const`) defined, and as a result, it was much easier to make them manual than to write new code to parse variable declarations. Also, the strada tests leaned heavily on position math (for example, `position + 1`) and was easier to just put in new markers where needed (didn't want to deal with tons of conversions)

internal/ls/linkedediting.go
iisaduan: Yes, it was because it was needed before. Should I remove it?

internal/ls/linkedediting.go
iisaduan: made some changes to reduce the conversions

internal/ast/utilities.go
iisaduan: So the reason why it wasn't hitting race conditions was that this logic was already included in the binder already. I've made the change to jst check `NodeFlagsThisNodeOrAnySubNodesHasError` in linked edits

 We should probably include something to deprecate at least `NodeFlagsHasAggregatedChildData` (but not sure if I should delete it _and_ shift the values?)

jakebailey: Just delete it and leave an empty slot.

jakebailey: You will have to regenerate the API package but that should be one command.

internal/ls/linkedediting.go
```diff
@@ -0,0 +1,107 @@
...
+	if ast.IsJsxFragment(token.Parent.Parent) {
+		fragment := token.Parent.Parent.AsJsxFragment()
+		openFragment := fragment.OpeningFragment
+		closeFragment := fragment.ClosingFragment
+		if openFragment.Flags&ast.NodeFlagsThisNodeHasError != 0 || closeFragment.Flags&ast.NodeFlagsThisNodeHasError != 0 {
+			return lsproto.LinkedEditingRangeResponse{}, nil
+		}
+
+		openPos := core.TextPos(astnav.GetStartOfNode(openFragment.AsNode(), sourceFile, false) + 1)   // "<".length
+		closePos := core.TextPos(astnav.GetStartOfNode(closeFragment.AsNode(), sourceFile, false) + 2) // "</".length
```
DanielRosenwasser: @jakebailey I assume Go would just inline


```suggestion
		openPos := core.TextPos(astnav.GetStartOfNode(openFragment.AsNode(), sourceFile, false) + len("<"))
		closePos := core.TextPos(astnav.GetStartOfNode(closeFragment.AsNode(), sourceFile, false) + len("</"))
```

jakebailey: Yes, it would. That's preferred.

internal/ls/linkedediting.go
jakebailey: Ah, because type narrowing? It's fine to leave

testdata/baselines/reference/submodule/fourslash/linkedEditing/linkedEditingJsxTag10.linkedEditing.txt
```diff
@@ -0,0 +1,80 @@
+=== /jsx0.tsx ===
```
jakebailey, line 2: We aren't using the diff form of these; is that possible?

DanielRosenwasser: idk if this is related to the question, but the old test baselines are `tests/baselines/reference/linkedEditingJsxTag10.linkedEditing.txt` - they use `.txt` instead of `.baseline.jsonc`.

iisaduan: I updated them to run with diffs-- they are clean :) (no diffs)

Issue #3044, opened 2026-03-10
Regex parsing/validation isn't implemented yet


RyanCavanaugh:
## Steps to reproduce

```ts
var x = /fo(o/;
```
or anything else of this flavor

## Behavior with `typescript@6.0`

> error TS1005: ')' expected.

## Behavior with `tsgo`

> 0 errors

jakebailey:
I'm pretty sure this is a part of the unported regex errors and can't be done individually.

Andarist:
FWIW, I'm working on porting this for some time already: https://github.com/Andarist/typescript-go/tree/port/55600 ;p

jakebailey:
Same except I gave up 😄

PR #2107, opened 2025-11-17
Suggested changes to #2026 (Implement RegExp literal syntax checking)

graphemecluster:
**The merge target of this PR is `jabaile/regexp-errors`, not `main`.** It is intended as a supplement of #2026.
Feel free to fast-forward/cherry-pick/rebase some of the commits here.

PR #3061, opened 2026-03-11
Port Regular Expressions checking

Andarist:
This is meant to be a pretty faithful port of the existing code. 

closes https://github.com/microsoft/typescript-go/issues/3044

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -1706,7 +1706,7 @@ func (s *Scanner) scanEscapeSequence(flags EscapeSequenceScanningFlags) string {
 			if flags&EscapeSequenceScanningFlagsRegularExpression != 0 && flags&EscapeSequenceScanningFlagsAtomEscape == 0 && ch != '0' {
 				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, fmt.Sprintf("\\x%02x", code))
 			} else {
-				s.errorAt(diagnostics.Octal_escape_sequences_are_not_allowed_Use_the_syntax_0, start, s.pos-start, "\\x"+fmt.Sprintf("%02x", code))
+				s.errorAt(diagnostics.Octal_escape_sequences_are_not_allowed_Use_the_syntax_0, start, s.pos-start, fmt.Sprintf("\\x%02x", code))
 			}
 			return string(rune(code))
 		}


jakebailey:
Can you see if this works with the tests added in #2107?

Andarist:
It doesn't report the exact same errors as that PR in the `regularExpressionDuplicateCapturingGroupName` test - but what it does report is consistent with Strada

jakebailey:
Okay. Strada alignment is of course good 😄

internal/checker/types.go
```diff
@@ -1276,6 +1281,8 @@ type LanguageFeatureMinimumTargetMap struct {
 }
 
 var LanguageFeatureMinimumTarget = LanguageFeatureMinimumTargetMap{
+	RegularExpressionFlagsUnicode:     core.ScriptTargetES2015,
+	RegularExpressionFlagsSticky:      core.ScriptTargetES2015,
```
jakebailey: Are these worth adding? We don't allow target to go below ES2015 anyway, hence why they are gone.

internal/parser/parser.go
```diff
@@ -254,6 +254,13 @@ func (p *Parser) scanError(message *diagnostics.Message, pos int, length int, ar
 	p.parseErrorAtRange(core.NewTextRange(pos, pos+length), message, args...)
 }
 
+func (p *Parser) lastDiagnostic() *ast.Diagnostic {
+	if len(p.diagnostics) > 0 {
+		return p.diagnostics[len(p.diagnostics)-1]
+	}
+	return nil
+}
```
jakebailey: This appears to be unused.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1047 @@
...
+const (
+	RegularExpressionFlagsNone           RegularExpressionFlags = 0
+	RegularExpressionFlagsHasIndices     RegularExpressionFlags = 1 << 0 // d
+	RegularExpressionFlagsGlobal         RegularExpressionFlags = 1 << 1 // g
+	RegularExpressionFlagsIgnoreCase     RegularExpressionFlags = 1 << 2 // i
+	RegularExpressionFlagsMultiline      RegularExpressionFlags = 1 << 3 // m
+	RegularExpressionFlagsDotAll         RegularExpressionFlags = 1 << 4 // s
+	RegularExpressionFlagsUnicode        RegularExpressionFlags = 1 << 5 // u
+	RegularExpressionFlagsUnicodeSets    RegularExpressionFlags = 1 << 6 // v
+	RegularExpressionFlagsSticky         RegularExpressionFlags = 1 << 7 // y
```
jakebailey: I don't think these need to be public; I'd lowercase the consts and types.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1047 @@
...
+
+var regExpFlagToFirstAvailableLanguageVersion = map[RegularExpressionFlags]core.ScriptTarget{
+	RegularExpressionFlagsHasIndices:  core.ScriptTargetES2022,
+	RegularExpressionFlagsDotAll:      core.ScriptTargetES2018,
+	RegularExpressionFlagsUnicode:     core.ScriptTargetES2015,
+	RegularExpressionFlagsUnicodeSets: core.ScriptTargetES2024,
+	RegularExpressionFlagsSticky:      core.ScriptTargetES2015,
+}
+
+func CharacterCodeToRegularExpressionFlag(ch rune) (RegularExpressionFlags, bool) {
```
jakebailey: This is only used internally.

internal/scanner/unicodeproperties.go
```diff
@@ -0,0 +1,166 @@
+package scanner
+
+// Table 66: Non-binary Unicode property aliases and their canonical property names
+// https://tc39.es/ecma262/#table-nonbinary-unicode-properties
+var nonBinaryUnicodeProperties = map[string]string{
```
jakebailey, line 7: Can't everything in this file just be switch cases? Maybe it doesn't much matter.

internal/scanner/unicodeproperties.go
```diff
@@ -0,0 +1,166 @@
...
+	"gc":                "General_Category",
+	"Script":            "Script",
+	"sc":                "Script",
+	"Script_Extensions": "Script_Extensions",
+	"scx":               "Script_Extensions",
+}
+
+// Table 67: Binary Unicode property aliases and their canonical property names
+// https://tc39.es/ecma262/#table-binary-unicode-properties
+var binaryUnicodeProperties = newStringSet(
```
jakebailey: FWIW we have `collections.NewSetFromItems`

internal/scanner/unicodeproperties.go
Andarist: there are some spelling suggestions derived from those so usign switch/case everywhere here wouldnt cut it

internal/scanner/unicodeproperties.go
```diff
@@ -0,0 +1,160 @@
...
+		"Zinh", "Inherited", "Qaai",
+		"Zyyy", "Common",
+		"Zzzz", "Unknown",
+	),
+	"Script_Extensions": nil, // set to Script values in init()
+}
+
+func init() {
+	valuesOfNonBinaryUnicodeProperties["Script_Extensions"] = valuesOfNonBinaryUnicodeProperties["Script"]
+}
```
jakebailey: You could just extract it to a global and not need an init func.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+// CharacterClass ::= unicodeMode
+//
+//	? '[' ClassRanges ']'
+//	: '[' ClassSetExpression ']'
+func (p *regExpParser) scanAlternative(isInGroup bool) {
+	isPreviousTermQuantifiable := false
+	for p.pos() < p.end {
+		start := p.pos()
+		ch := p.charAt(p.pos())
+		switch ch {
```
jakebailey, line 208: Do we need the EOF case here?

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+
+func (p *regExpParser) setPos(v int) {
+	p.scanner.pos = v
+}
+
+func (p *regExpParser) incPos(n int) {
+	p.scanner.pos += n
+}
+
+func (p *regExpParser) charAt(pos int) rune {
```
jakebailey, line 122: `p.charAt(p.pos())` is everywhere; probably we should add `char()` like the regular scanner? Or, just make these passthroughs?

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+					if maxStr != "" || p.charAt(p.pos()) == '}' {
+						p.error(diagnostics.Incomplete_quantifier_Digit_expected, digitsStart, 0)
+					} else {
+						p.error(diagnostics.Unexpected_0_Did_you_mean_to_escape_it_with_backslash, start, 1, string(ch))
+						isPreviousTermQuantifiable = true
+						continue
+					}
+				} else if maxStr != "" {
+					minVal, _ := strconv.Atoi(minStr)
+					maxVal, _ := strconv.Atoi(maxStr)
```
jakebailey: I would use `ParseInt` here, since `Atoi` is platform-dependent (see the rest of the scanner). That being said, we don't actually seem to parse out any base 10 numbers, which is fun. Are these values actually scanned out and valid? Or did we rely on `parseInt`?

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+			if p.charAt(p.pos()) != '}' {
+				if p.anyUnicodeModeOrNonAnnexB {
+					p.error(diagnostics.X_0_expected, p.pos(), 0, "}")
+					p.incPos(-1)
+				} else {
+					isPreviousTermQuantifiable = true
+					continue
+				}
+			}
+			// falls through to quantifier handling
```
jakebailey: Somewhat redundant comment

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+			fallthrough
+		case '*', '+', '?':
+			p.incPos(1)
+			if p.charAt(p.pos()) == '?' {
+				p.incPos(1)
```
jakebailey, line 313: There's a comment missing here:

```suggestion
				// Non-greedy
				p.incPos(1)
```

I would verify that we aren't losing comments in general.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+		case '.':
+			p.incPos(1)
+			isPreviousTermQuantifiable = true
+		case '[':
+			p.incPos(1)
+			if p.unicodeSetsMode {
+				p.scanClassSetExpression()
+			} else {
+				p.scanClassRanges()
+				p.pendingLowSurrogate = 0
```
jakebailey, line 328: This is new; can we explain the whole surrogate mess somewhere?

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+			return
+		default:
+			p.scanSourceCharacter()
+			isPreviousTermQuantifiable = true
+		}
+	}
+}
+
+func (p *regExpParser) scanPatternModifiers(currFlags regularExpressionFlags) regularExpressionFlags {
+	for p.pos() < p.end {
```
jakebailey, line 353: Loops that where `while (true)` are now `for pos < end`; is that intentional?

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+}
+
+// AtomEscape ::=
+//
+//	| DecimalEscape
+//	| CharacterClassEscape
+//	| CharacterEscape
+//	| 'k<' RegExpIdentifierName '>'
+func (p *regExpParser) scanAtomEscape() {
+	switch p.charAt(p.pos()) {
```
jakebailey: There's a debug assert missing here. Probably in other places too.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+		if p.unicodeSetsMode {
+			p.incPos(1)
+			p.error(diagnostics.X_q_is_only_available_inside_character_class, p.pos()-2, 2)
+			return
+		}
+		fallthrough
+	default:
+		if !p.scanCharacterClassEscape() && !p.scanDecimalEscape() {
+			p.scanCharacterEscape(true /*atomEscape*/)
+		}
```
jakebailey, line 403: Some sort of assert missing here too.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+	}
+}
+
+// DecimalEscape ::= [1-9] [0-9]*
+func (p *regExpParser) scanDecimalEscape() bool {
+	ch := p.charAt(p.pos())
+	if ch >= '1' && ch <= '9' {
+		start := p.pos()
+		p.scanDigits()
+		val, _ := strconv.Atoi(p.scanner.tokenValue)
```
jakebailey: Same here, though this doesn't use `parseInt`, but `+`, so theoretically should be `jsnum.FromString`? But, needs to be an int-ish, so not really.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+	return false
+}
+
+func (p *regExpParser) isClassContentExit(ch rune) bool {
+	return ch == ']' || p.pos() >= p.end
+}
+
+// ClassRanges ::= '^'? (ClassAtom ('-' ClassAtom)?)*
+func (p *regExpParser) scanClassRanges() {
+	p.pendingLowSurrogate = 0
```
jakebailey, line 508: Missing assert here too.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+		}
+		if atomEscape {
+			flags |= EscapeSequenceScanningFlagsAtomEscape
+		}
+		return p.scanner.scanEscapeSequence(flags)
+	}
+}
+
+func (p *regExpParser) scanGroupName(isReference bool) {
+	p.scanner.tokenStart = p.pos()
```
jakebailey, line 476: Missing assert here too.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+
+	anyUnicodeModeOrNonAnnexB bool
+	namedCaptureGroups        bool
+
+	mayContainStrings       bool
+	numberOfCapturingGroups int
+	groupSpecifiers         map[string]bool
+	groupNameReferences     []groupNameReference
+	decimalEscapes          []decimalEscapeValue
+	namedCapturingGroups    []map[string]bool
```
jakebailey:
```suggestion
	namedCapturingGroups    collections.Set[string]
```

Which is auto-initializing.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+			}
+			minCharacterValue, minSize := decodeClassAtomRune(minCharacter)
+			maxCharacterValue, maxSize := decodeClassAtomRune(maxCharacter)
+			if len(minCharacter) == minSize && len(maxCharacter) == maxSize && minCharacterValue > maxCharacterValue {
+				p.error(diagnostics.Range_out_of_order_in_character_class, minStart, p.pos()-minStart)
+			}
+		}
+	}
+}
+
```
jakebailey, line 545: I think there's a missing comment here?

```
        // Static Semantics: MayContainStrings
        //     ClassUnion: ClassSetOperands.some(ClassSetOperand => ClassSetOperand.MayContainStrings)
        //     ClassIntersection: ClassSetOperands.every(ClassSetOperand => ClassSetOperand.MayContainStrings)
        //     ClassSubtraction: ClassSetOperands[0].MayContainStrings
        //     ClassSetOperand:
        //         || ClassStringDisjunctionContents.MayContainStrings
        //         || CharacterClassEscape.UnicodePropertyValueExpression.LoneUnicodePropertyNameOrValue.MayContainStrings
        //     ClassStringDisjunctionContents: ClassStrings.some(ClassString => ClassString.ClassSetCharacters.length !== 1)
        //     LoneUnicodePropertyNameOrValue: isBinaryUnicodePropertyOfStrings(LoneUnicodePropertyNameOrValue)
```

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+// ClassSetExpression ::= '^'? (ClassUnion | ClassIntersection | ClassSubtraction)
+// ClassUnion ::= (ClassSetRange | ClassSetOperand)*
+// ClassIntersection ::= ClassSetOperand ('&&' ClassSetOperand)+
+// ClassSubtraction ::= ClassSetOperand ('--' ClassSetOperand)+
+// ClassSetRange ::= ClassSetCharacter '-' ClassSetCharacter
+func (p *regExpParser) scanClassSetExpression() {
+	isCharacterComplement := false
```
jakebailey, line 563: Also missing an assert. I'm going to stop commenting on these; they just all should be added back.

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1040 @@
...
+		case 'b':
+			p.incPos(1)
+			return "\b"
+		case '&', '-', '!', '#', '%', ',', ':', ';', '<', '=', '>', '@', '`', '~':
+			p.incPos(1)
+			return string(innerCh)
+		default:
+			return p.scanCharacterEscape(false /*atomEscape*/)
+		}
+	} else if p.pos()+1 < p.end && ch == rune(p.scanner.text[p.pos()+1]) {
```
jakebailey: This one line has `rune(p.scanner.text[p.pos()+1])` inlined. Probably shouldn't.

jakebailey (commented):
This all looks great; much appreciated! Only minor comments.

jakebailey:
In general, I would recheck that all of the comments are present too.

internal/scanner/regexp.go
Andarist: I don't think so, the current iteration flavour should be safe

internal/scanner/regexp.go
Andarist: technically-ish both of those might be lossy so I went here with a new `compareDecimalStrings` comparison

internal/scanner/regexp.go
Andarist: It's semi-intentional. I like this one better, and it allows us to get rid of the EOF case in the loop's body. If you preferred the old structure, I can certainly bring it back.

internal/scanner/regexp.go
Andarist: I don't think we need to change this. Using an int here is convenient and should be more than enough for any practical usage

internal/scanner/regexp.go
Andarist: `namedCapturingGroups` is not really “a set of all names”; it’s a scope stack used while parsing regex alternatives. The current `[]map[string]bool` models that directly with one map per active scope, with a plain `collections.Set[string]` we'd have to introduce an additional stack anyway

internal/scanner/regexp.go
Andarist: There is a comment close to the declaration of this field

internal/scanner/regexp.go
jakebailey: Ah, I missed the `[]` above.

But, `map[set]bool` is what I was trying to avoid. Not a major problem yet (it's not optimized in Go anyway, at least not until my CLs there are merged).

internal/scanner/scanner.go
```diff
@@ -1629,7 +1704,7 @@ func (s *Scanner) scanEscapeSequence(flags EscapeSequenceScanningFlags) string {
 		if flags&EscapeSequenceScanningFlagsReportInvalidEscapeErrors != 0 {
 			code, _ := strconv.ParseInt(s.text[start+1:s.pos], 8, 32)
 			if flags&EscapeSequenceScanningFlagsRegularExpression != 0 && flags&EscapeSequenceScanningFlagsAtomEscape == 0 && ch != '0' {
-				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, fmt.Sprintf("%02x", code))
+				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, "\\x"+fmt.Sprintf("%02x", code))
```
```suggestion
				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, fmt.Sprintf("\\x%02x", code))
```

internal/scanner/regexp.go
```diff
@@ -0,0 +1,1088 @@
...
+func (p *regExpParser) scanDecimalEscape() bool {
+	debug.Assert(p.pos() > 0 && p.text()[p.pos()-1] == '\\')
+	ch := p.char()
+	if ch >= '1' && ch <= '9' {
+		start := p.pos()
+		p.scanDigits()
+		val, err := strconv.Atoi(p.scanner.tokenValue)
+		if err != nil {
+			val = math.MaxInt
+		}
```
jakebailey, line 417: What is this error case for? Aren't the digits already validated?

internal/scanner/regexp.go
jakebailey: Great idea.

internal/scanner/regexp.go
Andarist: this is just not to ignore the `err` case. It's hard to believe anybody would create a source code that would fall into this branch but technically it's possible. At this point, we know that `p.scanner.tokenValue` constains a digits-only string but that string could (technically) exceed the int's range.

internal/scanner/scanner.go
```diff
@@ -1629,7 +1704,7 @@ func (s *Scanner) scanEscapeSequence(flags EscapeSequenceScanningFlags) string {
 		if flags&EscapeSequenceScanningFlagsReportInvalidEscapeErrors != 0 {
 			code, _ := strconv.ParseInt(s.text[start+1:s.pos], 8, 32)
 			if flags&EscapeSequenceScanningFlagsRegularExpression != 0 && flags&EscapeSequenceScanningFlagsAtomEscape == 0 && ch != '0' {
-				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, fmt.Sprintf("%02x", code))
+				s.errorAt(diagnostics.Octal_escape_sequences_and_backreferences_are_not_allowed_in_a_character_class_If_this_was_intended_as_an_escape_sequence_use_the_syntax_0_instead, start, s.pos-start, fmt.Sprintf("\\x%02x", code))
 			} else {
 				s.errorAt(diagnostics.Octal_escape_sequences_are_not_allowed_Use_the_syntax_0, start, s.pos-start, "\\x"+fmt.Sprintf("%02x", code))
```
```suggestion
				s.errorAt(diagnostics.Octal_escape_sequences_are_not_allowed_Use_the_syntax_0, start, s.pos-start, fmt.Sprintf("\\x%02x", code))
```

PR #3068, opened 2026-03-11
Preserve comments in empty JSX expressions during JSX emit

Andarist:


internal/printer/printer.go
```diff
@@ -4255,7 +4264,9 @@ func (p *Printer) emitJsxExpression(node *ast.JsxExpression) {
 		p.increaseIndentIf(indented)
 		end := p.emitToken(ast.KindOpenBraceToken, node.Pos(), WriteKindPunctuation, node.AsNode())
 		p.emitTokenNode(node.DotDotDotToken)
-		p.emitExpression(node.Expression, ast.OperatorPrecedenceDisallowComma)
+		if node.Expression != nil {
```
jakebailey, line 4267: Strada never checked this; is it actually nil?

Andarist: Strada checks against `undefined` in `emitExpression`:
https://github.com/microsoft/TypeScript/blob/c9e7428bb76f0543a3555d0af87777e7db3a41e6/src/compiler/emitter.ts#L1461-L1466

If you prefer me moving the check there, let me know.

jakebailey: Nope, all good. This was the only place that passed in `undefined` before (except `emitExpressionList`).

PR #3070, opened 2026-03-12
Add per-directory module resolution cache

andrewbranch:
This upstreams https://github.com/oxc-project/tsgolint/pull/764 from @camchenry. Thanks for the work there!

This is a simpler than what we used to have, and avoids the lock contention that made caching worse than not caching for us. Here are my measurements:

| Project | Files | Main | Cache | Speedup |
|---------|-------|------|-------|---------|
| kibana (src/core) | 4,935 | 785ms ± 54 | 662ms ± 35 | 1.19x |
| DefinitelyTyped | 46,485 | 30.4s ± 3.4 | 24.4s ± 1.7 | 1.25x |
| vscode | 5,472 | 8.81s ± 1.3 | 8.93s ± 1.4 | 0.99x |
| TypeScript (`-b src`) | 703 | 271ms ± 6 | 267ms ± 12 | 1.01x |
| drizzle-orm | 2,864 | 838ms ± 59 | 798ms ± 66 | 1.05x |

From profiles, I think kibana sees so much more improvement than vscode because vscode uses comparatively few node_modules dependencies.

### DIFF
--- a/internal/module/cache.go
+++ b/internal/module/cache.go
@@ -52,7 +52,7 @@ func (c *typeRefDirectiveResolutionCache) Set(key typeRefDirectiveResolutionCach
 type caches struct {
 	packageJsonInfoCache *packagejson.InfoCache
 
-	moduleResolutionCache              moduleResolutionCache
+	moduleResolutionCache           moduleResolutionCache
 	typeRefDirectiveResolutionCache typeRefDirectiveResolutionCache
 
 	// Cached representation for `core.CompilerOptions.paths`.


jakebailey (approved):

camchenry:
Amazing! Thank you for upstreaming this! Glad to see this was worth the effort! 

I'm still not fully familiar with the tsgo architecture, but is there any chance we need to invalidate the cache for long running processes? I was thinking like a LSP-type scenario that serves lots of requests over time. But maybe it's done from scratch each time?

andrewbranch:
It's done from scratch each time. Since it's attached to the Program and Programs are immutable, there's no need to do anything special. We either make a minimal clone when we can tell the change can't possibly affect module resolution, or we rebuild a new Program from scratch. (In Strada, there was logic to reuse parts of the module resolution cache between programs, but it was incredibly complex and could sometimes take longer to compute the cache invalidation than it would to just redo the work.)

camchenry:
> It's done from scratch each time. Since it's attached to the Program and Programs are immutable, there's no need to do anything special. We either make a minimal clone when we can tell the change can't possibly affect module resolution, or we rebuild a new Program from scratch. (In Strada, there was logic to reuse parts of the module resolution cache between programs, but it was incredibly complex and could sometimes take longer to compute the cache invalidation than it would to just redo the work.)

Got it! Makes sense and thanks for confirming. Will keep that in mind for future optimizations. I think this one was successful because it was relatively simple, so there's not a lot to track or extra overhead which is nice.

PR #3071, opened 2026-03-12
Fix `"use strict"` emit, again

jakebailey:
Hopefully my last attempt to port this goofy logic. Now it should behave the same as Strada.

### DIFF
--- a/internal/transformers/estransforms/usestrict.go
+++ b/internal/transformers/estransforms/usestrict.go
@@ -36,8 +36,10 @@ func (tx *useStrictTransformer) visitSourceFile(node *ast.SourceFile) *ast.Node 
 	moduleKind := tx.compilerOptions.GetEmitModuleKind()
 	format := tx.getEmitModuleFormatOfFile(node)
 
+	// ESM is always strict. If the file is ESM, and CJS emit
+	// has not been requested, then skip adding "use strict".
 	if isExternalModule && moduleKind >= core.ModuleKindES2015 &&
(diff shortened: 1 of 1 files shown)


internal/transformers/estransforms/usestrict.go
```diff
@@ -32,14 +32,12 @@ func (tx *useStrictTransformer) visitSourceFile(node *ast.SourceFile) *ast.Node
...
-	if tx.compilerOptions.GetEmitModuleKind() == core.ModuleKindPreserve {
-		return node.AsNode()
-	}
-
 	isExternalModule := ast.IsExternalModule(node)
+	moduleKind := tx.compilerOptions.GetEmitModuleKind()
 	format := tx.getEmitModuleFormatOfFile(node)
 
-	if isExternalModule && format >= core.ModuleKindES2015 {
+	if isExternalModule && moduleKind >= core.ModuleKindES2015 &&
```
gabritto, line 41: What's the meaning of these conditions?

jakebailey: It's saying "if this is ESM, and the user requested emitting ESM, then do not add `"use strict"`" as ESM is always strict.

If the user forces `module=commonjs`, then we are going to emit CJS, and therefore need to proceed.

jakebailey: I can add that as a comment, I know this is a bit opaque

gabritto: I think that would be helpful.

PR #3073, opened 2026-03-12
Port missing __makeTemplateObject transform, fix octal scanning

jakebailey:
This is an oddball, but `__makeTemplateObject` is required for certain kinds of escapes when going past ES2018.

The old code did this in the big es2018 transform since it was already there, but we don't do that, so I had to introduce a new subtree fact to prevent whole-tree walk for these goofy cases.

Then, noticing in the same diff an octal problem; fix that too.

### DIFF
--- a/internal/transformers/estransforms/taggedtemplate.go
+++ b/internal/transformers/estransforms/taggedtemplate.go
@@ -70,7 +70,7 @@ func (tx *taggedTemplateTransformer) processTaggedTemplateExpression(node *ast.T
 	tag := tx.Visitor().VisitNode(node.Tag)
 	template := node.Template
 
-	if template.SubtreeFacts()&ast.SubtreeContainsInvalidTemplateEscape == 0 {
+	if !hasInvalidEscape(template) {
 		return tx.Visitor().VisitEachChild(node.AsNode())
 	}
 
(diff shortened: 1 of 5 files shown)


testdata/baselines/reference/submodule/compiler/templateLiteralEscapeSequence.js.diff
```diff
@@ -104,114 +17,5 @@
 +tag `${5}0`;
 +tag `${55}0`;
```
jakebailey: This is its own bad bug. We are not parsing octals correctly.

internal/scanner/scanner.go
```diff
@@ -1786,10 +1786,11 @@ func (s *Scanner) scanNumber() ast.Kind {
 				s.tokenFlags |= ast.TokenFlagsContainsLeadingZero
 				fixedPart = digits
 			} else {
-				s.tokenValue = jsnum.FromString(digits).String()
+				val, _ := strconv.ParseInt(digits, 8, 64)
+				s.tokenValue = strconv.FormatInt(val, 10)
 				s.tokenFlags |= ast.TokenFlagsOctal
 				withMinus := s.token == ast.KindMinusToken
-				literal := core.IfElse(withMinus, "-", "") + "0o" + s.tokenValue
+				literal := core.IfElse(withMinus, "-", "") + "0o" + strconv.FormatInt(val, 8)
```
jakebailey, line 1788: These digits will have been scanned, so it is safe to use strconv (like we do in other places). The perf of this rare error case is not critical.

PR #3075, opened 2026-03-12
Avoid duplicating comments in empty lists

Andarist:


internal/printer/printer.go
```diff
@@ -4666,7 +4666,7 @@ func (p *Printer) emitListRange(emit func(p *Printer, node *ast.Node), parentNod
 
 	if format&LFBracketsMask != 0 {
 		if isEmpty && !isNil {
-			p.emitTrailingComments(children.Pos(), commentSeparatorBefore) // Emit comments within empty lists
+			p.emitLeadingComments(children.End(), false /*elided*/) // Emit comments within empty lists
```
Andarist, line 4669: this is a pure port fixup: https://github.com/microsoft/TypeScript/blob/c9e7428bb76f0543a3555d0af87777e7db3a41e6/src/compiler/emitter.ts#L4718

Issue #2212, opened 2025-12-04
Incorrect import elision when enum value references another enum from a different module


thaoula:
When one enum assigns values from an imported enum, tsgo removes the import but the compiled code still tries to use it, causing a crash at runtime. This works correctly in tsc.

## Steps to reproduce

**File: `system.workorder.status.ts`**

```typescript
export enum SystemWorkOrderStatus {
    Open = 'Open',
    InProgress = 'In Progress',
    Completed = 'Completed',
    Finalised = 'Finalised',
}
```

**File: `bulk.invoice.ts`**

```typescript
import { SystemWorkOrderStatus } from './system.workorder.status';

export enum BulkInvoiceJobStatus {
    InProgress = SystemWorkOrderStatus.InProgress,
    Completed = SystemWorkOrderStatus.Completed,
    Finalised = SystemWorkOrderStatus.Finalised,
}
```

## Behavior with `typescript@5.9`
Reference to other enum erased and replaced with value.

```javascript

var BulkInvoiceJobStatus;
(function (BulkInvoiceJobStatus) {
    BulkInvoiceJobStatus["InProgress"] = "In Progress";
    BulkInvoiceJobStatus["Completed"] = "Completed";
    BulkInvoiceJobStatus["Finalised"] = "Finalised";
})(BulkInvoiceJobStatus || (BulkInvoiceJobStatus = {}));
```

## Behavior with `tsgo`
Reference to original enum maintained without import causes run time error.

```javascript
var BulkInvoiceJobStatus;
(function (BulkInvoiceJobStatus) {
    BulkInvoiceJobStatus["InProgress"] = system_workorder_status_1.SystemWorkOrderStatus.InProgress;
    BulkInvoiceJobStatus["Completed"] = system_workorder_status_1.SystemWorkOrderStatus.Completed;
    BulkInvoiceJobStatus["Finalised"] = system_workorder_status_1.SystemWorkOrderStatus.Finalised;
})(BulkInvoiceJobStatus || (BulkInvoiceJobStatus = {}));
```

Runtime error:

```
ReferenceError: system_workorder_status_1 is not defined
```

a-tarasyuk:
@thaoula, in `5.9`, does it emit `require`, or only the evaluated values?

```ts
var BulkInvoiceJobStatus;
(function (BulkInvoiceJobStatus) {
    BulkInvoiceJobStatus["InProgress"] = "In Progress";
[...]

PR #3076, opened 2026-03-12
Align enum emit with Strada, fix inlining of imported enum members

AlCalzone:
fixes: #2212

When transforming enums, `RuntimeSyntaxTransformer.evaluateEntity()` only looked up enum member values from its local cache. For enum members imported from another file, the value was never resolved, leaving the raw expression in the output.

Input:
```ts
// File: imported.ts
export enum Imported {
	Option1 = 1,
}

// File: usage.ts
import { Imported } from "./imported.js";
export enum Usage {
	Option1 = Imported.Option1,
	Option2 = 2,
}
```

Strada output:
```js
export var Usage;
(function (Usage) {
    Usage[Usage["Option1"] = 1] = "Option1";
    Usage[Usage["Option2"] = 2] = "Option2";
})(Usage || (Usage = {}));
```

Corsa output (before this PR):
```js
export var Usage;
(function (Usage) {
    Usage["Option1"] = Imported.Option1;
    if (typeof Usage.Option1 !== "string") Usage[Usage.Option1] = "Option1";
    Usage[Usage["Option2"] = 2] = "Option2";
})(Usage || (Usage = {}));
```


AlCalzone:
After rebasing I realized there are many more differences in enum emit. The most recent commit fixes those diffs.

jakebailey:
I was also working on this recently: https://github.com/microsoft/typescript-go/tree/jabaile/redo-enums

There's a problem with this, in that it gets rid of the pretty nice property that we can emit enums without consulting the checker (note that `runtimesyntax` does not use an emit resolver until now!). I've been meaning to bring this up in a design meeting to see what we should do here.

I would in general think we could do even more deletions in the transform if we were using the checker, though this PR doesn't seem to have those changes.

(Not to say you should close this, just was also thinking about this diff.)

AlCalzone:
Happy to close this, given that you're already on it. I just happened to notice this when I compared the diff between Strada and Corsa in my codebase.

The diff I showed above is the only one remaining (aside from some comments and formatting differences), and unfortunately it happens to be a runtime error.

jakebailey:
I want to keep this open for a bit at least.

I didn't realize you were saying that it was crashing. So that probably says what we're going to have to do...

AlCalzone:
Right, I didn't explicitly say so - my bad. The import is elided, but still used in the enum definition.

jakebailey:
I compared the two, and my change had a few more changes that would have been required, so I've just opened #3092.

PR #3077, opened 2026-03-12
Preserve declaration comments on list elements and JSDoc type assertions

Andarist:


internal/printer/printer.go
```diff
@@ -5479,6 +5495,51 @@ func (p *Printer) emitTrailingComments(pos int, commentSeparator commentSeparato
 	p.emitComments(comments, commentSeparator)
 }
 
+func (p *Printer) emitTrailingCommentsOfPosition(pos int, prefixSpace bool, forceNoNewline bool) {
```
jakebailey, line 5498: Don't we have `emitTrailingComments`?

jakebailey: Basically, I'm looking at `emitTrailingCommentsOfPosition` and it's only a couple of lines.

Andarist: Yes, there is `emitTrailingComments` - but this one has a different logic. It's a port from Strada.

jakebailey: Hm, I'm wondering why we have an all new func `emitTrailingComments` that doesn't exist in Strada, if there's some sort of difference there that needs to be cleaned up.

I'm not sure if it's a port if it's so long, though, no?

Andarist: It exists in Strada:
https://github.com/microsoft/TypeScript/blob/c9e7428bb76f0543a3555d0af87777e7db3a41e6/src/compiler/emitter.ts#L6067-L6074

This port just doesn't use callbacks so it's more verbose.

jakebailey: I guess. I feel like we should (at some point) bring them into alignment, as it makes things hard to review.

PR #3095, opened 2026-03-13
Watch directories and react to symlink directory creation

### LINKED ISSUE #2506: Stale module resolution error after linking pnpm workspace dependency due to watcher configuration/handling

andrewbranch:
- Fixes #2506 
- Fixes the thing where deleting a directory doesn't get picked up (https://github.com/microsoft/vscode/issues/109754)
- Adds some follow-up from #3038

### DIFF
--- a/internal/project/snapshotfs.go
+++ b/internal/project/snapshotfs.go
@@ -419,33 +419,6 @@ func hasRelevantWatchExtension(path string) bool {
 	return false
 }
 
-// isTrackedNodeModulesDirectory returns true if the path is a node_modules
-// directory, a node_modules scope subdirectory, or a node_modules package directory.
-func isTrackedNodeModulesDirectory(fileName string) bool {
-	nodeModulesIndex := strings.Index(fileName, "/node_modules")
-	if nodeModulesIndex == -1 {
-		return false
-	}
-	if len(fileName) == nodeModulesIndex+len("/node_modules") {
-		return true
-	}
-	if fileName[nodeModulesIndex+len("/node_modules")] != '/' {
-		return false
-	}
-	remaining := fileName[nodeModulesIndex+len("/node_modules/"):]
-	if len(remaining) == 0 {
-		return false
-	}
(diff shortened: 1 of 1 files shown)


internal/project/snapshotfs.go
```diff
@@ -398,13 +399,105 @@ func (s *snapshotFSBuilder) markDirtyFiles(change FileChangeSummary) {
...
+// hasRelevantWatchExtension returns true if the given path ends with a file
+// extension relevant to TypeScript compilation. This is used to quickly filter
+// out watch events for files that cannot affect the project.
+func hasRelevantWatchExtension(path string) bool {
+	i := strings.LastIndexByte(path, '.')
+	if i < 0 {
+		return false
+	}
+	switch path[i:] {
+	case ".js", ".jsx", ".mjs", ".cjs", ".ts", ".tsx", ".mts", ".cts", ".json":
```
jakebailey, line 416: This is probably duplicative of `tspath`, but I assume it's here for perf?

PR #3100, opened 2026-03-16
Preallocate symbols in getSuggestionForSymbolNameLookup

camchenry:
I was doing some investigation in https://github.com/oxc-project/tsgolint/issues/797 (see that issue for CPU profiles) and noticed there was a significant slowdown between our latest version (0.17) and the previous version (0.16).

One contributing factor to this slowness seemed to be the `getSuggestionForSymbolNameLookup` method, which showed up in more profile samples in the new version, essentially doubling the time spent there. I'm not sure why that is the case though, I didn't look into it too deeply.

One seemingly simple optimization would be to pre-allocate this `allSymbols` array rather than letting `slices.Collect` allocate it as initially empty and naturally re-allocate as it pushes more elements. This should at least speed up this function a little bit. I don't know the best way to benchmark this in `typescript-go` (happy to benchmark it if you have a suggestion on that), but downstream in `tsgolint` this seemed to improve the overall lint time for running a single rule alone by about 3%.

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -1696,8 +1696,7 @@ func (c *Checker) getSuggestionForSymbolNameLookup(symbols ast.SymbolTable, name
 	if symbol != nil {
 		return symbol
 	}
-	allSymbols := make([]*ast.Symbol, 0, len(symbols))
-	allSymbols = slices.AppendSeq(allSymbols, maps.Values(symbols))
+	allSymbols := slices.AppendSeq(make([]*ast.Symbol, 0, len(symbols)), maps.Values(symbols))
 	if meaning&ast.SymbolFlagsGlobalLookup != 0 {
 		for _, s := range []string{"stringString", "numberNumber", "booleanBoolean", "objectObject", "bigintBigInt", "symbolSymbol"} {
 			if _, ok := symbols[s[len(s)/2:]]; ok {


internal/checker/checker.go
```diff
@@ -1696,7 +1696,10 @@ func (c *Checker) getSuggestionForSymbolNameLookup(symbols ast.SymbolTable, name
 	if symbol != nil {
 		return symbol
 	}
-	allSymbols := slices.Collect(maps.Values(symbols))
+	allSymbols := make([]*ast.Symbol, 0, len(symbols))
```
jakebailey: You can use AppendSeq to do this in one line (there are other examples in the repo)

no-yan:
I wonder if there’s a larger win here by reducing the amount of sorting we do.
If I’m reading this correctly, `c.sortSymbols(allSymbols)` can become quite expensive on large symbol tables, especially since we’re doing that work before we know which candidates actually matter.

If the sort is mainly there for deterministic tie-breaking, one option would be to first track the best candidates for the minimum Levenshtein distance, and only sort that final tie set. That should preserve determinism while reducing work in the common case.

internal/checker/checker.go
camchenry: Thanks! Added in https://github.com/microsoft/typescript-go/pull/3100/commits/3449f5536d5278ae173ee039994e5cf405d39c6a

camchenry: Although it's not quite one line, since `AppendSeq` doesn't pre-allocate the slice? Hopefully I understood the suggestion correctly.

camchenry:
This is ready for another review. This ended up not being quite so important for `tsgolint` since we realized the performance "regression" was actually due to the change in defaults for `types` in `tsconfig.json`. But this should still be a small perf win for the uncommon case where there are a lot of errors.

internal/checker/checker.go
jakebailey: I was just suggesting:

```go
allSymbols := slices.AppendSeq(make([]*ast.Symbol, 0, len(symbols)), maps.Values(symbols))
```

jakebailey:
> If I’m reading this correctly, `c.sortSymbols(allSymbols)` can become quite expensive on large symbol tables, especially since we’re doing that work before we know which candidates actually matter.

In this particular instance, we could totally diverge from `sortSymbols` and just do a plain sort on a simpler key like names only, if those are present.

internal/checker/checker.go
camchenry: Oh that should have been obvious, thanks. At this point, you've written the entire PR 😄

PR #3102, opened 2026-03-16
Fix a formatting crash caused by template literal in parser-recovered property signature

### LINKED ISSUE #3097: [ServerErrors][TypeScript] main vs
The following errors were reported by main
[Pipeline that generated this bug](https://dev.azure.com/typescript/TypeScript/_build?definitionId=75)
[Logs for the pipeline run](https://typescript.visualstudio.com/TypeScript/_build/results?buildId=167441)
[File that generated the pipeline](https://github.com/microsoft/typescript-error-deltas/blob/main/azure-pipelines-gitTests.yml)

This run considered 300 popular TS repos from GH (after skipping the top 0).

<details>
<summary>Successfully analyzed 223 of 300 visited repos</summary>

| Outcome | Count |
|---------|-------|
| Detected interesting changes | 32 |
| Detected no interesting changes | 191 |
| Git clone failed | 6 |
| Timeout | 69 |
| Unknown failure | 2 |

</details>

### LINKED ISSUE #63598: "Token end is child end" when '= ${id}id' is used in a property annotation
Happens in both 6.0 and 7.0 at the moment.

```ts
type X = { prop: = `${abc}`xyz }
```

Format the document.

```
Debug Failure. False expression: Token end is child end
Error: Debug Failure. False expression: Token end is child end
    at processChildNode (typescript.js:182787:17)
    at typescript.js:182718:9
    at visitNode2 (typescript.js:32551:18)
    at forEachChildInPropertySignature (typescript.js:32607:59)
    at forEachChild (typescript.js:33072:35)
    at processNode (typescript.js:182715:5)
    at processChildNode (typescript.js:182794:7)
    at processChildNodes (typescript.js:182837:32)
    at typescript.js:182731:9
    at visitNodes (typescript.js:32556:14)
    at forEachChildInTypeLiteral (typescript.js:32666:12)
    at forEachChild (typescript.js:33072:35)
    at processNode (typescript.js:182715:5)
    at processChildNode (typescript.js:182794:7)
    at typescript.js:182718:9
    at visitNode2 (typescript.js:32551:18)
    at forEachChildInTypeAliasDeclaration (typescript.js:32842:144)
    at forEachChild (typescript.js:33072:35)
    at processNode (typescript.js:182715:5)
    at formatSpanWorker (typescript.js:182512:5)
    at typescript.js:182482:19
    at getFormattingScanner (typescript.js:181224:15)
    at formatSpan (typescript.js:182477:10)
    at Object.formatSelection (typescript.js:182348:10)
    at Object.getFormattingEditsForRange (typescript.js:154136:34)
    at IpcIOSession.getFormattingEditsForRange (typescript.js:196749:35)
    at format (typescript.js:195068:43)
    at typescript.js:197591:15
    at IpcIOSession.executeWithRequestId (typescript.js:197580:14)
    at IpcIOSession.executeCommand (typescript.js:197589:29)
    at IpcIOSession.onMessage (typescript.js:197637:68)
    at process.<anonymous> (_tsserver.js:519:14)
    at process.emit (node:events:509:28)
    at emit (node:internal/child_process:972:14)
    at process.processTicksAndRejections (node:internal/process/task_queues:91:21)
```

Andarist:
Fixes #3097.
Fixes https://github.com/microsoft/TypeScript/issues/63598.

Fixes the crash found here https://github.com/microsoft/typescript-go/issues/3097#issuecomment-4059022203 (crashes in Strada too), it also fixes this one: https://github.com/microsoft/typescript-go/issues/3097#issuecomment-4059022436

What happened? This is a little bit tricky. The grammar error nodes were never visited by `processChildNode`, they just early-returned `inheritedIndentation`.

However, they could contain robust and mostly good ASTs. After skipping such nodes, the parent went through its leftover-token sweep (it has to consume all tokens that are still before the next child). At that point, the current node is `PropertySignature` and under that context `shouldRescanTemplateToken` returns false. 

So the tokens could not be rescanned properly, that resulted in consuming both `}` as regular `CloseBraceToken` and "closing" `` ` `` being consumed as a potential new `TemplateHead`. That new bogus template head consumed everything until the end of the file, crossing the correct ancestor containment boundary completely and ultimately leading to a crash.

In the commit history, you can see my various attempts to address this issue. I think that the simplest "[Don't skip over grammar error nodes](https://github.com/microsoft/typescript-go/commit/4daeef8f6da9b3ef59917c7eea586df1af51ff44)" approach was actually quite promising - I only expanded on it later with `isErrorMemberListElement`-based logic to avoid indenting the bogus property signatures that, in this sample, were clearly not meant to be property signatures.


internal/fourslash/tests/formatTemplateLiteralInMalformedPropertySignatureNoCrash1_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestFormatTemplateLiteralInMalformedPropertySignatureNoCrash1(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	content := `// @Filename: /resource-card.tsx
+interface Props {
```
DanielRosenwasser:
```suggestion
	content := `
// @Filename: /resource-card.tsx
interface Props {
```

internal/fourslash/tests/formatInterfaceWithMissingBraceAndLaterTemplateString1_test.go
```diff
@@ -0,0 +1,45 @@
...
+  )
+}
+
+export default ResourceCard
+`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	f.FormatDocument(t, "")
+	f.VerifyCurrentFileContent(t, `interface Props {
+    iconOnly?: boolean
```
DanielRosenwasser, line 36:
```suggestion
	f.VerifyCurrentFileContent(t, `
interface Props {
    iconOnly?: boolean
```

internal/format/span.go
```diff
@@ -346,9 +343,10 @@ func (w *formatSpanWorker) processChildNode(
 ) int {
 	debug.Assert(!ast.NodeIsSynthesized(child))
 
-	if ast.NodeIsMissing(child) || isGrammarError(parent, child) || child.Flags&ast.NodeFlagsReparsed != 0 {
+	if ast.NodeIsMissing(child) || child.Flags&ast.NodeFlagsReparsed != 0 {
 		return inheritedIndentation
 	}
+	isErrorMemberListElement := isMemberListElement(parent, child) && child.Flags&ast.NodeFlagsThisNodeHasError != 0
```
DanielRosenwasser: First, I'd slip the condition on this one to quickly check if the node has an error in the first place.

But also, I don't quite understand why you're limiting it to just the case of "member lists".

Also, move this down to the first `if` check where it's actually used.

internal/fourslash/tests/formatInterfaceWithMissingBraceAndLaterTemplateString1_test.go
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <IoLayersOutline
+      className={` + "`" + `${match ? 'text-primary-foreground' : 'text-foreground'}` + "`" + `}
+    />
+  )
+}
+
+export default ResourceCard
+`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
```
DanielRosenwasser, line 32: This test is really subtle given that Go's escaping makes it seem like the issue is there. Add some explanation about where things are malformed and what a parser might end up seeing it as.

internal/fourslash/tests/formatInterfaceWithMissingBraceAndLaterTemplateString1_test.go
DanielRosenwasser, line 1: Maybe name the test to point out that it's just missing a closing curly brace? There is no need to mention "no crash" in the test, no test is ever supposed to crash.

DanielRosenwasser: `formatInterfaceWithMissingBraceAndLaterTemplateString`?

DanielRosenwasser (dismissed):

internal/format/span.go
Andarist: > But also, I don't quite understand why you're limiting it to just the case of "member lists".

This is subtle but without this the interface itself is treated as something that should not change indentation (it also has `NodeFlagsThisNodeHasError` here) and then its "normal" child (`iconOnly?: boolean`) gets 0 indentation. 

There might be a better solution to all of this (well, given the source code isn't error-free all of this is best effort anyway). This is just the least invasive one I was able to come up with right now.

internal/fourslash/tests/formatInterfaceWithMissingBraceAndLaterTemplateString1_test.go
Andarist: this one can't be done because it changes the assertion (it requires an extra newline and the formatter doesn't seem to preserve empty lines at the start of the file either)

internal/fourslash/tests/formatInterfaceWithMissingBraceAndLaterTemplateString1_test.go
Andarist: Yeah, sure - I'll add some comment here

internal/format/util.go
```diff
@@ -188,3 +133,16 @@ func isListElement(parent *ast.Node, node *ast.Node) bool {
...
+	case ast.KindClassDeclaration,
+		ast.KindClassExpression,
+		ast.KindInterfaceDeclaration,
+		ast.KindEnumDeclaration,
+		ast.KindTypeLiteral,
+		ast.KindMappedType:
+		return node.Loc.ContainedBy(parent.MemberList().Loc)
+	}
+	return false
+}
```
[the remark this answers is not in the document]
Andarist: I'm not sure if I should care about this one here, there are some other functions that don't guard against `nil` - so at the very least this wouldnt be a new problem.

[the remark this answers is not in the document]
DanielRosenwasser: Well we parse out an empty node list in cases where the opening brace is missing, so it's not nil-able. So there's no need for the check.

DanielRosenwasser:
Unfortunately we've got merge conflicts. 🙁

Andarist:
@DanielRosenwasser I missed the notification in what was a particularly busy month for me - I just fixed those conflicts

gabritto (approved):

PR #3103, opened 2026-03-16
Fixed a signature help crash related to tagged template calls

Andarist:
fixes a crash found here: https://github.com/microsoft/typescript-go/issues/3099#issuecomment-4064180276


internal/ls/signaturehelp.go
```diff
@@ -867,7 +867,7 @@ func getImmediatelyContainingArgumentInfo(node *ast.Node, position int, sourceFi
 		}
 
 		spanIndex := ast.IndexOfNode(templateSpan.Parent.AsTemplateExpression().TemplateSpans.Nodes, templateSpan)
-		argumentIndex := getArgumentIndexForTemplatePiece(spanIndex, templateSpan, position, sourceFile)
+		argumentIndex := getArgumentIndexForTemplatePiece(spanIndex, node, position, sourceFile)
```
Andarist, line 870: this was misported - it now matched Strada

PR #3104, opened 2026-03-16
Block renames in keywords/stdlib, add prepareRename

### LINKED ISSUE #2229: Should not allow renaming keywords
TS-go will allow triggering rename anywhere in the file, including on parens and keywords. When you press enter it will then either try doing a rename or say `no result`

Instead we should fail the rename early. Here's what happens in TS 6

Image

I believe this warning comes from the `prepareRename` call

### LINKED ISSUE #2230: Should not allow renaming parts of the standard library
Similar to #2229 but we should also make sure you can't try to rename standard library functions like `setTimeout`

jakebailey:
Fixes #2229
Fixes #2230

We were lacking `prepareRename`. For clients without `prepareRename`, we still precheck in the actual rename call.

This also ports more fourslash so we can test more of this.

image

image

image

### DIFF
--- a/internal/ls/rename.go
+++ b/internal/ls/rename.go
@@ -15,6 +15,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/ls/lsconv"
 	"github.com/microsoft/typescript-go/internal/ls/lsutil"
 	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
+	"github.com/microsoft/typescript-go/internal/module"
 	"github.com/microsoft/typescript-go/internal/tspath"
 )
 
@@ -193,8 +194,8 @@ func wouldRenameInOtherNodeModules(originalFile *ast.SourceFile, symbol *ast.Sym
 		return nil
 	}
 
-	originalPackage := getPackagePathComponents(originalFile.FileName())
-	if originalPackage == nil {
+	originalPackage := module.ParseNodeModuleFromPath(originalFile.FileName(), false /*isFolder*/)
+	if originalPackage == "" {
 		// Original source file is not in node_modules.
 		for _, declaration := range declarations {
 			if isInsideNodeModules(ast.GetSourceFileOfNode(declaration).FileName()) {
@@ -206,42 +207,12 @@ func wouldRenameInOtherNodeModules(originalFile *ast.SourceFile, symbol *ast.Sym
 
 	// Original source file is in node_modules.
 	for _, declaration := range declarations {
-		declPackage := getPackagePathComponents(ast.GetSourceFileOfNode(declaration).FileName())
-		if declPackage != nil {
-			length := min(len(originalPackage), len(declPackage))
-			for i := 0; i <= length; i++ {
-				var origComp, declComp string
-				if i < len(originalPackage) {
-					origComp = originalPackage[i]
-				}
-				if i < len(declPackage) {
-					declComp = declPackage[i]
-				}
-				if origComp != declComp {
-					return diagnostics.You_cannot_rename_elements_that_are_defined_in_another_node_modules_folder
-				}
-			}
+		declPackage := module.ParseNodeModuleFromPath(ast.GetSourceFileOfNode(declaration).FileName(), false /*isFolder*/)
+		if declPackage != "" && declPackage != originalPackage {
+			return diagnostics.You_cannot_rename_elements_that_are_defined_in_another_node_modules_folder
 		}
 	}
 	return nil
-}
-
-// getPackagePathComponents returns the path components up to and including the package name
-// within node_modules, or nil if the path is not inside node_modules.
-func getPackagePathComponents(filePath string) []string {
-	components := tspath.GetPathComponents(filePath, "")
-	nodeModulesIdx := -1
-	for i := len(components) - 1; i >= 0; i-- {
-		if components[i] == "node_modules" {
-			nodeModulesIdx = i
-			break
-		}
-	}
-	if nodeModulesIdx == -1 {
-		return nil
-	}
-	end := min(nodeModulesIdx+2, len(components))
-	return components[:end]
 }
 
 // getRenameInfoForModule handles rename validation for module specifiers.


internal/lsp/server.go
```diff
@@ -1169,6 +1172,19 @@ func (s *Server) handleHover(ctx context.Context, ls *ls.LanguageService, params
 	return ls.ProvideHover(ctx, params.TextDocument.Uri, params.Position)
 }
 
+func (s *Server) handlePrepareRename(ctx context.Context, languageService *ls.LanguageService, params *lsproto.PrepareRenameParams) (lsproto.PrepareRenameResponse, error) {
+	info := languageService.GetRenameInfo(ctx, params.TextDocument.Uri, params.Position)
+	if !info.CanRename {
+		return lsproto.PrepareRenameResponse{}, nil
+	}
```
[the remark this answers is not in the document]
jakebailey: Yeah, turns out LSP does not expose this. Commented these out for now

internal/ls/rename.go
```diff
@@ -0,0 +1,363 @@
...
+	for i := len(components) - 1; i >= 0; i-- {
+		if components[i] == "node_modules" {
+			nodeModulesIdx = i
+			break
+		}
+	}
+	if nodeModulesIdx == -1 {
+		return nil
+	}
+	end := min(nodeModulesIdx+2, len(components))
```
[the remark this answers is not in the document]
jakebailey: this is a pure port, but maybe

internal/ls/rename.go
```diff
@@ -0,0 +1,363 @@
...
+
+	node := astnav.GetTouchingPropertyName(sourceFile, pos)
+	node = getAdjustedLocation(node, true /*forRename*/, sourceFile)
+
+	if isNodeEligibleForRename(node) {
+		if renameInfo, ok := l.getRenameInfoForNode(ctx, node, sourceFile, program); ok {
+			return renameInfo
+		}
+	}
+	return getRenameInfoError(ctx, diagnostics.You_cannot_rename_this_element)
```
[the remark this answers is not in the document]
jakebailey: This is still a gap, we are missing a bunch of string stuff

internal/fourslash/_scripts/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "noEmit": true,
         "module": "nodenext",
         "allowImportingTsExtensions": true,
-        "noFallthroughCasesInSwitch": true
+        "noFallthroughCasesInSwitch": true,
+        "types": ["node"]
```
gabritto, line 8: Why do we need this?

internal/ls/rename.go
```diff
@@ -0,0 +1,351 @@
...
+func (l *LanguageService) getRenameInfoForNode(ctx context.Context, node *ast.Node, sourceFile *ast.SourceFile, program *compiler.Program) (RenameInfo, bool) {
+	ch, done := program.GetTypeChecker(ctx)
+	defer done()
+
+	symbol := ch.GetSymbolAtLocation(node)
+	if symbol == nil {
+		if ast.IsStringLiteralLike(node) {
+			// Allow renaming of string literal types with contextual string literal types
+			typ := getContextualTypeFromParent(node, ch, checker.ContextFlagsNone)
+			if typ != nil && (typ.Flags()&checker.TypeFlagsStringLiteral != 0 ||
```
gabritto: Personal nit, but I find it more readable to use `typ.IsStringLiteral()`, `typ.IsUnion()`, etc.

internal/fourslash/_scripts/tsconfig.json
jakebailey: The scripts don't compile anymore, since we started defaulting to `[]`. A drive by but annoying when looking at the converter.

internal/ls/rename.go
jakebailey: Yeah, this was just auto-completed, but the original code does use methods.

jakebailey: Yeah, this was just auto-completed, but the original code does use methods.

internal/ls/rename.go
```diff
@@ -0,0 +1,363 @@
...
+
+// wouldRenameInOtherNodeModules checks if renaming the symbol would affect node_modules.
+func wouldRenameInOtherNodeModules(originalFile *ast.SourceFile, symbol *ast.Symbol, ch *checker.Checker, preferences *lsutil.UserPreferences) *diagnostics.Message {
+	sym := symbol
+	if !preferences.UseAliasesForRename.IsTrue() && sym.Flags&ast.SymbolFlagsAlias != 0 {
+		importSpecifier := core.Find(sym.Declarations, ast.IsImportSpecifier)
+		if importSpecifier != nil && importSpecifier.AsImportSpecifier().PropertyName == nil {
+			sym = ch.GetAliasedSymbol(sym)
+		}
+	}
```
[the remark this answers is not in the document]
jakebailey: That's not how truthy works

internal/ls/rename.go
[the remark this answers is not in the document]
jakebailey: Now fixed, actually

internal/lsp/server.go
```diff
@@ -550,6 +550,12 @@ func (s *Server) sendResult(id *jsonrpc.ID, result any) error {
 	})
 }
 
+// requestFailedError is an error with LSP RequestFailed (-32803) code and a clean user-facing message.
+type requestFailedError string
+
+func (e requestFailedError) Error() string { return string(e) }
+func (e requestFailedError) Unwrap() error { return lsproto.ErrorCodeRequestFailed }
+
```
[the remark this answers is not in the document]
jakebailey: Yep, that the logging was the next thing I was going to fix.

internal/ls/rename.go
```diff
@@ -0,0 +1,352 @@
...
+	for i := len(components) - 1; i >= 0; i-- {
+		if components[i] == "node_modules" {
+			nodeModulesIdx = i
+			break
+		}
+	}
+	if nodeModulesIdx == -1 {
+		return nil
+	}
+	end := min(nodeModulesIdx+2, len(components))
```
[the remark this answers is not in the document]
jakebailey: This is still preexisting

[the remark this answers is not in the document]
jakebailey: we have a better func for this anyway, yay!

testdata/baselines/reference/submodule/fourslash/findRenameLocations/jsdocTypedefTagRename01.baseline.jsonc.diff
```diff
@@ -37,10 +37,13 @@
...
 -// [|NumberLikeRENAME|] = 10;
+-// 
+-// /** @type {/*RENAME*/[|NumberLikeRENAME|]} */
++// var NumberLike;
++// 
 +// NumberLike = 10;
- // 
- // /** @type {/*RENAME*/[|NumberLikeRENAME|]} */
++// 
++// /** @type {/*RENAME*/NumberLike} */
```
gabritto, line 48: Why did this change? Is it that we don't create types in JS for variables anymore and require `typeof`?

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameDefaultImport.baseline.jsonc.diff
```diff
@@ -0,0 +1,21 @@
+--- old.renameDefaultImport.baseline.jsonc
++++ new.renameDefaultImport.baseline.jsonc
+@@= skipped -0, +0 lines =@@
+ // === findRenameLocations ===
```
gabritto, line 4: Note for future: this looks like a diff just in file sort order, I probably missed it when adding rename diffs, but we already do file sorting for find all refs diffs, so could fix it the same way.

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameNumericalIndex.baseline.jsonc.diff
```diff
@@ -3,16 +3,14 @@
 @@= skipped -0, +0 lines =@@
  // === findRenameLocations ===
  // === /renameNumericalIndex.ts ===
--// const foo = { /*RENAME*/[|0RENAME|]: true };
+ // const foo = { /*RENAME*/[|0RENAME|]: true };
 -// foo[/*START PREFIX*/"[|0RENAME|]"/*END SUFFIX*/];
```
gabritto, line 7: Do you know what we're missing that causes this?

internal/fourslash/fourslash.go
```diff
@@ -3640,36 +3640,69 @@ func (f *FourslashTest) verifyBaselineRename(
...
-		NewName:  "?",
 	}
 
 	prefix := f.getCurrentPositionPrefix()
-	result := sendRequest(t, f, lsproto.TextDocumentRenameInfo, params)
-	if result.WorkspaceEdit != nil {
-		t.Fatalf(prefix+"Expected rename to fail, but got changes: %s", cmp.Diff(result.WorkspaceEdit, nil))
+	f.baselineState(t)
+	f.baselineRequestOrNotification(t, lsproto.TextDocumentPrepareRenameInfo.Method, params)
+	resMsg, result, _ := lsptestutil.SendRequest(t, f.client, lsproto.TextDocumentPrepareRenameInfo, params)
```
gabritto, line 3686: What was going wrong with using `sendRequest` and letting it handle baseline state?

internal/lsp/server.go
```diff
@@ -550,6 +550,11 @@ func (s *Server) sendResult(id *jsonrpc.ID, result any) error {
 	})
 }
 
+type userFacingRequestFailedError string
```
gabritto, line 553: Do we have plans to actually surface this at some point?

testdata/baselines/reference/submodule/fourslash/findRenameLocations/jsdocTypedefTagRename01.baseline.jsonc.diff
jakebailey: Rather, we don't let you do a `typedef` and not give a name, as in we don't infer it from the var below: https://github.com/microsoft/typescript-go/blob/main/CHANGES.md#:~:text=Identifier%2Dnamed%20typedefs

It's possible there are more places that could be deleted.

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameDefaultImport.baseline.jsonc.diff
jakebailey: Yep, I noticed this too but didn't want to stack more on.

internal/fourslash/fourslash.go
jakebailey: This is actually the first case we have where we _expect_ an error to come back. Not many LSP calls actually want you to return an error. This could probably be done more clearly, though, if you want me to look harder.

jakebailey: Actaully, this is the only place in the LSP that does this! https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=Clients%20should%20show%20the%20information%20in%20their%20user%20interface.

internal/lsp/server.go
jakebailey: That's what's in my PR screenshots, actually. To go with: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=Clients%20should%20show%20the%20information%20in%20their%20user%20interface.

internal/fourslash/fourslash.go
jakebailey: Er, well. this is the failed case, so not exactly what I'm describing, however normally a failed request fails the test.

testdata/baselines/reference/submodule/fourslash/findRenameLocations/renameNumericalIndex.baseline.jsonc.diff
jakebailey: The `START PREFIX` thing? that's some fourslash thing that I think is not ported.

internal/lsp/server.go
gabritto: Ah, I missed the updated PR description. Nice!

PR #3105, opened 2026-03-16
Fix a document highlight crash caused by expression in type parameter declaration

Andarist:
fixes the crash found here: https://github.com/microsoft/typescript-go/issues/3099#issuecomment-4064180306

### DIFF
--- a/internal/fourslash/tests/documentHighlightTypeParameterConstraintExpressionNoCrash1_test.go
+++ b/internal/fourslash/tests/documentHighlightTypeParameterConstraintExpressionNoCrash1_test.go
@@ -1,0 +1,20 @@
+package fourslash_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


internal/ast/ast.go
```diff
@@ -2830,11 +2832,13 @@ func (node *TypeParameterDeclaration) ForEachChild(v Visitor) bool {
 }
 
 func (node *TypeParameterDeclaration) VisitEachChild(v *NodeVisitor) *Node {
```
Andarist, line 2834: Strada doesn't crash on this test case because `getTouchingPropertyName` uses  `getChildren` and thus `forEachChild`. `forEachChild` visits those expressions already - but in Corsa this has been refactored quite a bit and now `VisitEachChild` gets called and that doesn't visit this recovery-only property (it also doesn't visit that in Strada)

PR #3108, opened 2026-03-16
Fix panic caused by importing a type-only alias using `require(...)` in JS

ahejlsberg:
This PR fixes a panic that occurs when compiling the following test case (with both `tsgo` and `tsc`):

```ts
// @checkJs: true
// @noEmit: true
// @module: commonjs

// @filename: t.ts
type Strings = string[]
export = Strings

// @filename: main.js
const t = require("./t")
```

This should fix the issue uncovered [here](https://github.com/microsoft/typescript-go/issues/3099#issuecomment-4064180218).

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -6545,8 +6545,6 @@ func (c *Checker) checkAliasSymbol(node *ast.Node) {
 			importText := "import(\"" + specifierText + "\")"
 			if ast.IsImportSpecifier(node) {
 				importText = importText + "." + identifierText
-			} else if ast.IsImportClause(node) {
-				importText = importText + ".default"
 			}
 			c.error(errorNode, diagnostics.X_0_is_a_type_and_cannot_be_imported_in_JavaScript_files_Use_1_in_a_JSDoc_type_annotation, identifierText, importText)
 		}
--- a/testdata/baselines/reference/submodule/compiler/elidedJSImport1.errors.txt.diff
+++ b/testdata/baselines/reference/submodule/compiler/elidedJSImport1.errors.txt.diff
@@ -3,7 +3,7 @@
 @@= skipped -0, +0 lines =@@
  caller.js(1,21): error TS2591: Cannot find name 'fs'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
 -caller.js(2,8): error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract").TruffleContract' in a JSDoc type annotation.
-+caller.js(2,8): error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract").default' in a JSDoc type annotation.
++caller.js(2,8): error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract")' in a JSDoc type annotation.
  caller.js(4,43): error TS2708: Cannot use namespace 'TruffleContract' as a value.
  caller.js(4,60): error TS2708: Cannot use namespace 'TruffleContract' as a value.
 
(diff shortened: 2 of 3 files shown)


testdata/baselines/reference/submodule/compiler/elidedJSImport1.errors.txt.diff
```diff
@@ -0,0 +1,18 @@
...
++caller.js(2,8): error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract")' in a JSDoc type annotation.
+ caller.js(4,43): error TS2708: Cannot use namespace 'TruffleContract' as a value.
+ caller.js(4,60): error TS2708: Cannot use namespace 'TruffleContract' as a value.
+
+@@= skipped -9, +9 lines =@@
+ !!! error TS2591: Cannot find name 'fs'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
+     import TruffleContract from '@truffle/contract'; // Runtime err: this import is elided in transform
+            ~~~~~~~~~~~~~~~
+-!!! error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract").TruffleContract' in a JSDoc type annotation.
++!!! error TS18042: 'TruffleContract' is a type and cannot be imported in JavaScript files. Use 'import("@truffle/contract")' in a JSDoc type annotation.
```
jakebailey: Can we keep the dotted part in this case?

ahejlsberg: It is wrong. The module exports a `default` and that's it.

jakebailey: Ah! Yes.

andrewbranch: It has to be `import("@truffle/contract").default`, then. The import call gives you the module namespace record (same as `import *`)

ahejlsberg: Yeah, was wondering how into the weeds we want to get with the different cases. But I'll fix that one.

DanielRosenwasser (approved):
Error message is a little weird in this case, but it seems fine.

internal/checker/checker.go
```diff
@@ -6530,18 +6530,24 @@ func (c *Checker) checkAliasSymbol(node *ast.Node) {
...
 				}
 			}
-			identifierText := symbol.Name
-			if ast.IsIdentifier(errorNode) {
-				identifierText = errorNode.Text()
+			importText := "import(\"" + specifierText + "\")"
+			if ast.IsImportSpecifier(node) {
+				importText = importText + "." + identifierText
+			} else if ast.IsImportClause(node) {
+				importText = importText + ".default"
```
andrewbranch: This is incorrect if the importing file is ESM and the imported file is using `export =`. Modifying your existing test:

```ts
// @checkJs: true
// @noEmit: true
// @module: commonjs

// @filename: t.cts
type Strings = string[]
export = Strings

// @filename: main.js
const t1 = require("./t.cjs")

// @filename: main2.mjs
import t1 from "./t.cjs"
/** @type {import("./t.cjs")} */
let s = ["hello", "world"]
```

The error on `import t1 from "./t.cjs"` says to use `import("./t.cjs").default`, but the lack of error on the next line shows that `import("./t.cjs")` is correct.

ahejlsberg: Yeah, this is the kind of in-the-weeds stuff I didn't want to get into. I honestly think it is fine to just say "use import(...) in a JSDoc annotation" and not get into whether `.default` is needed.

andrewbranch: I'm fine with that. Do you want to get rid of the dotted name always? Or just for defaults?

ahejlsberg: Just for default imports. So basically revert my last commit.

PR #3110, opened 2026-03-16
Big omnibus of JS emit differences

jakebailey:
This fixes the bulk of the remaining diffs that are not fixed by #3092 / #3073.

There's no clear pattern for these in general.

The biggest change is that we need destructuring code for some runtime emit features to not change our CJS emit. That gets extracted out.


internal/printer/printer.go
```diff
@@ -2610,7 +2610,12 @@ func (p *Printer) emitConciseBody(node *ast.BlockOrExpression) {
 	case ast.IsBlock(node):
 		p.emitFunctionBody(node.AsBlock())
 	case ast.IsObjectLiteralExpression(ast.GetLeftmostExpression(node, false /*stopAtCallExpressions*/)):
-		p.emitExpression(node, ast.OperatorPrecedenceParentheses)
+		// Wrap in ParenthesizedExpression to ensure parens are emitted after any leading
+		// PartiallyEmittedExpression comments, matching TypeScript's factory-time wrapping
+		// via parenthesizeConciseBodyOfArrowFunction.
+		paren := p.emitContext.Factory.NewParenthesizedExpression(node)
+		paren.Loc = node.Loc
+		p.emitExpression(paren, ast.OperatorPrecedenceLowest)
```
jakebailey, line 2623: This has dubious value but does fix a diff.

internal/printer/emitcontext.go
```diff
@@ -988,6 +989,20 @@ func (c *EmitContext) GetSyntheticTrailingComments(node *ast.Node) []Synthesized
...
+	c.emitNodes.Get(node).typeNode = typeNode
+}
+
+// GetTypeNode gets the type node stored on a name node by the type eraser.
+func (c *EmitContext) GetTypeNode(node *ast.Node) *ast.TypeNode {
+	if c.emitNodes.Has(node) {
+		return c.emitNodes.Get(node).typeNode
+	}
+	return nil
+}
```
DanielRosenwasser:
```suggestion
	if emitNode := c.emitNodes.TryGet(node); emitNode != nil {
		return emitNode.typeNode
	}
	return nil
}
```

internal/printer/factory.go
```diff
@@ -573,6 +573,17 @@ func (f *NodeFactory) GetNamespaceMemberName(ns *ast.IdentifierNode, name *ast.I
...
 
+// Gets the export name of a declaration for use in expressions.
+//
+// An export name will *always* be prefixed with a module or namespace export modifier like
+// `"exports."` when emitted as an expression if the name points to an exported symbol.
+func (f *NodeFactory) GetExternalModuleOrNamespaceExportName(ns *ast.IdentifierNode, node *ast.Declaration, allowComments bool, allowSourceMaps bool) *ast.Node {
+	if ns != nil && ast.HasSyntacticModifier(node, ast.ModifierFlagsExport) {
+		return f.GetNamespaceMemberName(ns, f.GetDeclarationNameEx(node, NameOptions{AllowComments: allowComments, AllowSourceMaps: allowSourceMaps}), NameOptions{AllowComments: allowComments, AllowSourceMaps: allowSourceMaps})
+	}
+	return f.GetExportNameEx(node, AssignedNameOptions{AllowComments: allowComments, AllowSourceMaps: allowSourceMaps})
```
DanielRosenwasser, line 585: maybe

```suggestion
	if ns != nil && ast.HasSyntacticModifier(node, ast.ModifierFlagsExport) {
		nameOpts := NameOptions{AllowComments: allowComments, AllowSourceMaps: allowSourceMaps}
		return f.GetNamespaceMemberName(ns, f.GetDeclarationNameEx(node, nameOpts), nameOpts)
	}
	return f.GetExportNameEx(node, AssignedNameOptions{AllowComments: allowComments, AllowSourceMaps: allowSourceMaps})
```

DanielRosenwasser: Original code technically dropped allowComments and allowSourceMaps on the `getName` call - possibly a mistake.

PR #3125, opened 2026-03-17
Fix error type intrinsic name

jakebailey:
These names are not actually given to users, but we need them to be accurate for type baselines.


testdata/baselines/reference/submodule/compiler/requireOfJsonFileWithComputedPropertyName.types.diff
```diff
@@ -0,0 +1,10 @@
+--- old.requireOfJsonFileWithComputedPropertyName.types
++++ new.requireOfJsonFileWithComputedPropertyName.types
+@@= skipped -25, +25 lines =@@
+
+     [a]: 10
+ >[a] : number
+->a : any
++>a : error
```
jakebailey, line 8: This is new, but because we are missing errors in this file due to not having ported some JSON parsing errors!

PR #3127, opened 2026-03-17
Fixed some diagnostic writer divergences

Andarist:
This eliminates a few format-diverging `.diff`s


jakebailey:
IIRC these were actually intentional differences. Note the lack of extra random newlines.

Andarist:
Should I add those to the accepted diffs then? what about the change to `internal/parser/parser.go`?

internal/parser/parser.go
```diff
@@ -908,7 +908,7 @@ func (p *Parser) parseExpectedMatchingBrackets(openKind ast.Kind, closeKind ast.
 		return
 	}
 	if lastError != nil {
-		related := ast.NewDiagnostic(nil, core.NewTextRange(openPosition, openPosition), diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, scanner.TokenToString(openKind), scanner.TokenToString(closeKind))
+		related := ast.NewDiagnostic(nil, core.NewTextRange(openPosition, openPosition+1), diagnostics.The_parser_expected_to_find_a_1_to_match_the_0_token_here, scanner.TokenToString(openKind), scanner.TokenToString(closeKind))
```
jakebailey, line 911: Might need to check the git history, but I recall some of this changing too because we were avoiding bounds errors, and some of those changes may be missing tests?

PR #3129, opened 2026-03-17
Validate JSON syntax during parse

jakebailey:
This code never got ported.


RyanCavanaugh:
I'm going to do the annoying thing and ask about problems Strada also had: What is this function even trying to do? It doesn't error on comments or `0x32` values, but those are valid JSON5, but JSON5 allows single-quoted strings, which this doesn't allow

jakebailey:
Do we actually support JSON5? JSONC is different, as is JWCC. I thought we were just "json with commas and comments".

PR #3137, opened 2026-03-17
Fix all writestring analyzer errors

jakebailey:
There's a gopls analyzer that finds these. I've run it manually, and fixed everything.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -2198,7 +2198,9 @@ func (f *FourslashTest) VerifyBaselineSelectionRanges(t *testing.T) {
 	for i, marker := range markers {
 		if i > 0 {
 			result.WriteString(newLine)
-			result.WriteString(strings.Repeat("=", 80))
+			for range 80 {
+				result.WriteByte('=')
+			}
 			result.WriteString(newLine)
 			result.WriteString(newLine)
 		}
--- a/internal/tsoptions/commandlineparser_test.go
+++ b/internal/tsoptions/commandlineparser_test.go
@@ -279,9 +279,16 @@ func formatNewBaseline(
 ) string {
 	var formatted strings.Builder
 	formatted.WriteString("Args::\n")
-	formatted.WriteString("[\"")
-	formatted.WriteString(strings.Join(commandLine, "\", \""))
-	formatted.WriteString("\"]")
+	formatted.WriteByte('[')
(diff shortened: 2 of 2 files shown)


internal/fourslash/fourslash.go
```diff
@@ -2195,7 +2197,10 @@ func (f *FourslashTest) VerifyBaselineSelectionRanges(t *testing.T) {
 
 	for i, marker := range markers {
 		if i > 0 {
-			result.WriteString(newLine + strings.Repeat("=", 80) + newLine + newLine)
+			result.WriteString(newLine)
+			result.WriteString(strings.Repeat("=", 80))
```
DanielRosenwasser: This one is kind of funny - do you know if `strings.Repeat()` like this gets folded into a constant string?

internal/tsoptions/commandlineparser_test.go
```diff
@@ -279,7 +279,9 @@ func formatNewBaseline(
 ) string {
 	var formatted strings.Builder
 	formatted.WriteString("Args::\n")
-	formatted.WriteString("[\"" + strings.Join(commandLine, "\", \"") + "\"]")
+	formatted.WriteString("[\"")
+	formatted.WriteString(strings.Join(commandLine, "\", \""))
```
DanielRosenwasser: If you're here, you may want to avoid the `Join` as well.

internal/fourslash/fourslash.go
jakebailey: I don't believe so. Could just write a loop. my main goal was to ditch the errors.

internal/fourslash/fourslash.go
```diff
@@ -2195,7 +2197,12 @@ func (f *FourslashTest) VerifyBaselineSelectionRanges(t *testing.T) {
 
 	for i, marker := range markers {
 		if i > 0 {
-			result.WriteString(newLine + strings.Repeat("=", 80) + newLine + newLine)
+			result.WriteString(newLine)
+			for range 80 {
+				result.WriteByte('=')
```
DanielRosenwasser, line 2202: I don't actually know if this is better to be honest, but I don't know how efficient Builders are with this

jakebailey: They grow dynamically like slices do. It's all whatever; this is not perf critical code, I'm mainly wanting to get rid of warnings in the editor

PR #3138, opened 2026-03-17
Reimplement --showConfig

jakebailey:
The current version just dumps as JSON. Rewrite it entirely in `tsoptions`.

What is not in this PR is the nice "inferred options" printing. That relied on the "computed options" thing before, which we do not have anymore.

### DIFF
--- a/internal/tsoptions/showconfig.go
+++ b/internal/tsoptions/showconfig.go
@@ -263,19 +263,20 @@ func serializeEnumValue(value any, enumMap *collections.OrderedMap[string, any])
 	return getNameOfCompilerOptionValue(value, enumMap)
 }
 
-// matchesSpecs returns a filter function that returns true for files that should be
-// included in the --showConfig "files" list. Files that match the include globs (and
-// are not excluded) are filtered OUT, since they're already covered by the "include" field.
+// matchesSpecs returns a filter function that determines whether a file should appear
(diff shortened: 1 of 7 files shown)


internal/tsoptions/showconfig.go
```diff
@@ -0,0 +1,251 @@
...
+		files = append(files, relativePath)
+	}
+
+	// Serialize compiler options
+	optionMap := serializeCompilerOptions(configParseResult.CompilerOptions(), normalizedConfigPath, comparePathsOptions)
+
+	// Remove command-line-only options from the output
+	for _, name := range []string{
+		"showConfig", "configFile", "configFilePath", "help", "init",
+		"listFiles", "listEmittedFiles", "project", "build", "version",
```
RyanCavanaugh: `listFiles` is valid in tsconfig, probably meant to be this?
```suggestion
		"listFilesOnly", "listEmittedFiles", "project", "build", "version",
```

jakebailey: listFiles is valid in tsconfig???

jakebailey: This is actually a pure port, Strada removed `listFiles` and not `listFilesOnly`. Exciting.

RyanCavanaugh (approved):

PR #3144, opened 2026-03-18
Ensure language service requests create a project for the requested file if one doesn't already exist

### LINKED ISSUE #2822: Harden LSP against requests for files that don't have open projects

### LINKED ISSUE #2906: Requests failing with ` Message: no project found for URI`
After reloading a VS Code window, I'm seeing errors like this:

```
2026-02-25 13:51:31.915 [error] error handling method 'textDocument/codeLens' (9): no project found for URI file:///Users/matb/projects/sandbox/x.ts
2026-02-25 13:51:31.915 [error] Request textDocument/codeLens failed.
  Message: no project found for URI file:///Users/matb/projects/sandbox/x.ts
  Code: -32603 
```

[Full logs](https://github.com/user-attachments/files/25559029/log.txt)

andrewbranch:
- Fixes #2906 
- Fixes #2822

I didn't realize until investigating, but this was already fixed for files that belong in configured projects (as long as the `disableReferencedProjectLoad` and `disableSolutionSearching` settings don't prevent us from discovering the right project for that file); we just needed to go the last step of creating an inferred project when no configured project is found.

I opted not to do anything special with the lifecycle of projects created like this; they will be cleaned up on the next file open unless that file open provides a reason for it to stick around.

### DIFF
--- a/internal/project/projectcollectionbuilder.go
+++ b/internal/project/projectcollectionbuilder.go
@@ -364,10 +364,10 @@ func (b *ProjectCollectionBuilder) ensureInferredProjectIncludesClosedFile(fileN
 	}
 }
 
-// DidRequestFile ensures projects are loaded for the given URI. If optional is false and no
-// configured project is found for a closed file, an inferred project will be created.
-// If optional is true, only configured projects are ensured; no inferred project is created.
-func (b *ProjectCollectionBuilder) DidRequestFile(uri lsproto.DocumentUri, optional bool, logger *logging.LogTree) {
+// DidRequestFile ensures projects are loaded for the given URI.
+// If configuredProjectsOnly is true, only configured projects are loaded; no inferred project is created
+// and it is not guaranteed that there will be any project containing the file in the resulting snapshot.
+func (b *ProjectCollectionBuilder) DidRequestFile(uri lsproto.DocumentUri, configuredProjectsOnly bool, logger *logging.LogTree) {
 	startTime := time.Now()
 	fileName := uri.FileName()
 	path := b.toPath(fileName)
@@ -412,7 +412,7 @@ func (b *ProjectCollectionBuilder) DidRequestFile(uri lsproto.DocumentUri, optio
 		// a non-TS-handleable file.
 	} else {
 		result := b.ensureConfiguredProjectAndAncestorsForFile(fileName, path, logger)
-		if result.project == nil && !optional {
+		if result.project == nil && !configuredProjectsOnly {
 			// No configured project found for this closed file.
 			// Add it to the inferred project so language service requests can be served.
 			b.ensureInferredProjectIncludesClosedFile(fileName, logger)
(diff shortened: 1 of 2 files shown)


internal/project/session.go
```diff
@@ -550,7 +551,7 @@ func (s *Session) GetLanguageServiceAndProjectsForFile(ctx context.Context, uri
 func (s *Session) GetProjectsForFile(ctx context.Context, uri lsproto.DocumentUri) ([]ls.Project, error) {
 	snapshot := s.getSnapshot(
 		ctx,
-		ResourceRequest{Documents: []lsproto.DocumentUri{uri}},
+		ResourceRequest{ConfiguredProjectDocuments: []lsproto.DocumentUri{uri}},
```
andrewbranch, line 557: This request is not direct from the LSP client; it's part of the logic for cross-project find-all-refs and similar. It's different in kind from everything else we use `ResourceRequest{Documents}` for, because it's _supposed_ to only load configured projects for these documents _if_ settings allow those projects to be loaded and _if_ there's actually a tsconfig.json. `ResourceRequest{Documents}`, on the other hand, was always intended to mean “the LSP client is asking a question about these documents so you _have_ to produce a snapshot that has a project for all of them.”

internal/project/session.go
```diff
@@ -479,23 +479,24 @@ func (s *Session) getSnapshot(
...
+			for _, document := range documents {
+				if snapshot.fs.isOpenFile(document.FileName()) {
+					project := snapshot.GetDefaultProject(document)
+					if project == nil {
+						return UpdateReasonRequestedLanguageServiceProjectNotLoaded
+					} else if project.dirty {
+						return UpdateReasonRequestedLanguageServiceProjectDirty
+					}
+				} else {
+					return UpdateReasonRequestedLanguageServiceForFileNotOpen
```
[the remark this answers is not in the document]
andrewbranch: This is not new, but it does seem like it might have been a mistake before 🤔

internal/project/projectcollectionbuilder.go
```diff
@@ -912,6 +935,7 @@ func (b *ProjectCollectionBuilder) updateInferredProjectRoots(rootFileNames []st
 		return false
 	}
 
+	slices.Sort(rootFileNames)
```
andrewbranch, line 938: Moving this into a shared code path, previously missed from a couple callers, was a drive-by fix suggested by Copilot review

internal/project/session.go
DanielRosenwasser: Are you saying we'll never get to this code path with something like a non-configured `.js` file?

DanielRosenwasser: Just saw the comment below - it's not the case, it sounds like it's just handled differently compared to `Documents`.

PR #3154, opened 2026-03-18
Sort `unparsedTests`

gabritto:


internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -59,6 +59,7 @@ func TestMain(m *testing.M) {
 
     parseTypeScriptFiles(getManualTests(), stradaFourslashPath);
 
+    unparsedFiles.sort((a, b) => a.file.localeCompare(b.file, "en-US"));
```
jakebailey, line 62: Does .sort not work alone?

gabritto: I think this was done elsewhere for `failedTests`, and I can't recall why, but I remember there was a reason. Actually I thought you made that change in `updateFailing.mts`?

PR #3162, opened 2026-03-19
Organize imports comment duplication

johnfav03:
Fixes an issue where the OrganzeImports source action would cause duplication of the comments above the export statement


bradleyayers:
> comments above the import statement

it's specifically exports not imports (imports are currently working fine)

johnfav03:
> Can you also add tests to see if it works with comments that aren't the first thing in the file? Also, does this bug show up with imports as well?

This bug existed for imports in my initial implementation of OrganzeImports, but was fixed in PR review; I must have overlooked the export case

iisaduan (approved):

PR #3163, opened 2026-03-19
Support implicit jsx and tslib imports in find all refs

gabritto:
We used to store those synthetic imports in the file's imports in Strada, but now they're stored in the program, so I refactored some functions to be able to access a program, and added cancellation support in one place in the import tracker as well.

### DIFF
--- a/testdata/baselines/reference/submodule/fourslash/findAllReferences/referencesInEmptyFile.baseline.jsonc
+++ b/testdata/baselines/reference/submodule/fourslash/findAllReferences/referencesInEmptyFile.baseline.jsonc
@@ -1,0 +1,3 @@
+// === findAllReferences ===
+// === /referencesInEmptyFile.ts ===
+// /*FIND ALL REFS*/


internal/ls/importTracker.go
```diff
@@ -94,11 +97,23 @@ func getDirectImportsMap(sourceFiles []*ast.SourceFile, checker *checker.Checker
 }
 
 // Calls `action` for each import, re-export, or require() in a file
-func forEachImport(sourceFile *ast.SourceFile, action func(importStatement *ast.Node, imported *ast.Node)) {
-	if sourceFile.ExternalModuleIndicator != nil || len(sourceFile.Imports()) != 0 {
+func forEachImport(program *compiler.Program, sourceFile *ast.SourceFile, action func(importStatement *ast.Node, imported *ast.Node)) {
+	var implicitImports []*ast.LiteralLikeNode
```
jakebailey, line 101: I was going to suggest presizing it, but I'm pretty sure the compiler does this for us now up to 16 anyway.

PR #3167, opened 2026-03-19
Fix: skip CommonJS export binding when module or exports is a local variable

### LINKED ISSUE #2656: Exports are no longer recognized in js modules which contain functions with parameter called `module` which have `module.exports =` in their body
## Steps to reproduce

```ts
// dep.js
function something(module) {
    module.exports = 8;
}

export default 7;

// main.ts

import DefaultExport from "./dep.js";

export default 3 + DefaultExport;
```

No errors raised

## Behavior with `tsgo`

Raises TS1192 (or if `dep.js` does named exports and `main.ts` does named imports, TS2305).

Muhammad-Bin-Ali:
Fixes #2656
### Summary
In `.js` files, `module.exports = x` or `exports.foo = x` inside a function where `module` or `exports` is a local variable (parameter, `const`, `var`, etc.) was incorrectly treated as a real CommonJS export. This caused the file's actual ESM exports to be shadowed by a false `export=` symbol, producing spurious TS1192/TS2305/TS2309 errors on imports.

The root cause is that the reparser creates `JSExportAssignment`/`CommonJSExport` nodes based on purely syntactic matching (`IsModuleIdentifier` just checks `node.Text() == "module"`), without verifying whether the identifier actually refers to the CommonJS global or a local binding. Since the parser has no scope information, this check can't happen there.

The fix adds a scope check in the binder (where scope information is available) before binding these nodes. If `module` or `exports` resolves to a local symbol in `b.blockScopeContainer` via `lookupName`, the binder skips the CJS export binding entirely. Rather than clearing `CommonJSModuleIndicator` mid-walk (which risks re-entry via later `setCommonJSModuleIndicator` calls), the indicator is left intact during the walk and corrected once afterward: if the indicator pointed to a skipped node, it is replaced with the first valid CJS export that was actually bound, or cleared if none exists.

Also fixes a pre-existing false TS6425 diagnostic on `amdLikeInputDeclarationEmit`, where `const module = {}; module.exports = X` inside an AMD callback was incorrectly flagged as a nested CommonJS export.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
 
       - run: git add .
 
-      - uses: codecov/codecov-action@1af58845a975a7985b0beb0cbe6fbbb71a41dbad # v5.5.3
+      - uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
         if: ${{ always() && matrix.config.coverage && github.event_name != 'merge_group' }}
         with:
           use_oidc: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork) }}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -48,7 +48,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@38697555549f1db7851b81482ff19f1fa5c4fedc # v4.34.1
+        uses: github/codeql-action/init@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
         with:
           config-file: ./.github/codeql/codeql-configuration.yml
         # Override language selection by uncommenting this and choosing your languages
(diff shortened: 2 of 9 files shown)


Muhammad-Bin-Ali:

jakebailey:
This seems plausible (I was hoping we could do this), but, I don't think it's right to clear the CJS indicator mid walk. My gut says that it could be readded by a later call.

Shouldn't it be better to instead guard where we set it in the first place, if that assignment is related to one of these patterns?

Muhammad-Bin-Ali:
> This seems plausible (I was hoping we could do this), but, I don't think it's right to clear the CJS indicator mid walk. My gut says that it could be readded by a later call.
> 
> Shouldn't it be better to instead guard where we set it in the first place, if that assignment is related to one of these patterns?

Yeah, I see what you're getting at, and I think I agree with that. Let me see if I can implement that.

Muhammad-Bin-Ali:
@jakebailey 

Upon looking into it further, I don't think there's a clean way of solving this problem earlier in the walk. Whether we choose to address it earlier in the binder's walk or in the reparser, we need information about what symbols exists before the `JSExportAssignment` node is visited, which is exactly what's determined in the walk. That being said, I did modify the approach a little (as well as added tests) to address the problem you mentioned.

jakebailey:
Superseded by #3388

PR #3168, opened 2026-03-19
Add all-checks task to hereby

gabritto:
I keep forgetting to run some of those. Not sure if there's something else I should include (the API package checks are not included on purpose, also `generate` is not something I think we need often and takes a while to run).
Ideally we'd run lint and test in parallel, but if I try and do that from the task, the output is interleaved and it's tricky to read.

### DIFF
--- a/Herebyfile.mjs
+++ b/Herebyfile.mjs
@@ -1801,8 +1801,8 @@ export const allChecks = task({
     name: "all-checks",
     description: "Runs all checks for the Go code (fourslash, lint, tests, etc.)",
     run: async () => {
-        await runTests();
         await $`npm run convertfourslash`;
+        await runTests();
         await $`npm run updatefailing`;
         await runFormat();
         await runLint();


Herebyfile.mjs
```diff
@@ -1792,3 +1796,15 @@ export const nativePreview = task({
         throw new Error("This task should not be run in release builds.");
     } : undefined,
 });
+
+export const allChecks = task({
+    name: "all-checks",
+    description: "Runs all checks for the Go code (fourslash, lint, tests, etc.)",
+    run: async () => {
+        await $`npm run convertfourslash`;
+        await $`npm run updatefailing`;
```
jakebailey, line 1806: Really wish we could just get these out of scripts and into herebyfile tasks

jakebailey: A shame we encoded that into every generated test, ha

gabritto: > A shame we encoded that into every generated test, ha

What do you mean?

jakebailey: The tests say:

```
// Code generated by convertFourslash; DO NOT EDIT.
// To modify this test, run "npm run makemanual format01"
```

But, I guess that's fine, because hereby won't do positionals

Herebyfile.mjs
```diff
@@ -1792,3 +1796,16 @@ export const nativePreview = task({
         throw new Error("This task should not be run in release builds.");
     } : undefined,
 });
+
+export const allChecks = task({
+    name: "all-checks",
+    description: "Runs all checks for the Go code (fourslash, lint, tests, etc.)",
+    run: async () => {
+        await runTests();
+        await $`npm run convertfourslash`;
```
jakebailey, line 1804: Should `convert` come before running tests? Given it can make more tests?

gabritto: Yeah, good catch.
So just to reiterate my understanding, the workflow should be: we convert fourslash tests to see if the generated tests are up to date, then we run tests. If a test fails, that means you're not done and need to fix something. If it passes, we can run `updateFailing` to see if your change unblocked more tests, and then finally lint/format/test again. I think this makes sure we don't ever add tests to the failing list by accident, which is a concern you brought up.

jakebailey: Yes, right, I think that flow is correct. If you are able to get through this stack of changes, CI should pass.

PR #3169, opened 2026-03-19
Submodule accept some find all refs diffs

gabritto:
Disclaimer: I only looked at half of the find all refs diffs.

Most of the submodule accepted diffs are due to the fact that in Strada, find all references returned locations with not only a range but a context range. LSP's textDocument/references only allows returning a location with a range. So the context range gets printed in Strada baselines as `<| ... |>`, and while we ignore those in the diff fixup function, we will include more lines in the Strada baseline, which are skipped in Corsa baselines.
Given there aren't that many diffs, I think just accepting them is fine.

The exceptions are:
- `findAllRefsNonexistentPropertyNoCrash1.baseline.jsonc.diff`: good change, the diff due to an bug in how Strada parses test files with a dot in the name, such that `./src/parser.js` file becomes included in the program as both `/./src/parser.js` and `/src/parser.js`. Corsa doesn't have this bug.
- `findAllRefsParameterPropertyDeclaration_inheritance.baseline.jsonc.diff`: good change, Strada seemed to behave inconsistently when finding references to parameter properties.
- `findAllRefs_importType_js4.baseline.jsonc.diff`: Corsa fails to find the right baseline file because Strada's test file has a `.` in its name. Not sure if it's worth fixing.


jakebailey:
I trust you  😄

PR #3176, opened 2026-03-20
Fix ref count cache inconsistencies

### LINKED ISSUE #3086: panic: cache entry not found
## Stack trace

```
panic: cache entry not found [recovered, repanicked]

goroutine 34 [running]:
github.com/microsoft/typescript-go/internal/project.(*Snapshot).Clone.func1()
	github.com/microsoft/typescript-go/internal/project/snapshot.go:238 +0xb8
panic({0x103a53680?, 0x103cce9d8?})
	runtime/panic.go:860 +0x12c
github.com/microsoft/typescript-go/internal/project.(*RefCountCache[...]).Ref(0x103cf1ea0, {{{0x72ab06218c00, 0x71}, {0x72ab15344980, 0x71}, {0x0, 0x0}}, 0x3, {0xabad6dd56cc8caf0, 0x95c165df8124d237}})
	github.com/microsoft/typescript-go/internal/project/refcountcache.go:64 +0x1c4
github.com/microsoft/typescript-go/internal/project.(*Snapshot).Clone(0x72ab69659200, {0x103cdb1d8, 0x72ab6a8ae660}, {{{0x72ab6844cfa0, 0x1, 0x1}, {0x0, 0x0, 0x0}, 0x0, ...}, ...}, ...)
	github.com/microsoft/typescript-go/internal/project/snapshot.go:449 +0x1318
github.com/microsoft/typescript-go/internal/project.(*Session).UpdateSnapshot(0x72ab05f1e000, {0x103cdb1d8, 0x72ab6a8ae660}, 0x72ab6a8ae6c0, {{{0x72ab6844cfa0, 0x1, 0x1}, {0x0, 0x0, 0x0}, ...}, ...})
	github.com/microsoft/typescript-go/internal/project/session.go:653 +0xd8
github.com/microsoft/typescript-go/internal/project.(*Session).getSnapshot(0x72ab05f1e000, {0x103cdb1d8, 0x72ab6a8ae660}, {{0x72ab6844cfa0, 0x1, 0x1}, {0x0, 0x0, 0x0}, 0x0, ...})
	github.com/microsoft/typescript-go/internal/project/session.go:461 +0x314
github.com/microsoft/typescript-go/internal/project.(*Session).getSnapshotAndDefaultProject(0x72ab05f1e000, {0x103cdb1d8, 0x72ab6a8ae660}, {0x72ab4d10fd50, 0x6a})
	github.com/microsoft/typescript-go/internal/project/session.go:512 +0xb0
github.com/microsoft/typescript-go/internal/project.(*Session).GetLanguageServiceAndSnapshot(0x88d623c091e?, {0x103cdb1d8?, 0x72ab6a8ae660?}, {0x72ab4d10fd50?, 0x164585270268c?})
	github.com/microsoft/typescript-go/internal/project/session.go:533 +0x28
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceWithAutoImportsRequestHandler[...].28({0x103cdb1d8, 0x72ab6a8ae660}, 0x72ab4faf7440)
	github.com/microsoft/typescript-go/internal/lsp/server.go:
767 +0xa0
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification(0x72ab05d06e08, {0x103cdb210?, 0x72ab6a715720?}, 0x72ab4faf7440)
github.com/microsoft/typescript-go/internal/lsp/server.go:599 +0xc8
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop(0x72ab05d06e08, {0x103cdb210?, 0x72ab05ca6000?})
	github.com/microsoft/typescript-go/internal/lsp/server.go:483 +0x368
	github.com/microsoft/typescript-go/internal/lsp/server.go:346 +0x24
golang.org/x/sync/errgroup.(*Group).Go.func1()
	golang.org/x/sync@v0.19.0/errgroup/errgroup.go:93 +0x4c

created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
	golang.org/x/sync@v0.19.0/errgroup/errgroup.go:78 +0x90
StdIO connection closed
Exit code: 2
Command line: /Users/noel/work/revenuemarket-web/node_modules/@typescript/native-preview-darwin-arm64/lib/tsgo --lsp --stdio
```

## Steps to reproduce

Sorry I cannot make reproduce about it, only occurs on our private monorepo(turborepo) repository.

Version: 7.0.0-dev.20260312.1

andrewbranch:
Fixes #3086 

Unfortunately, the combination of simplifications in #2483 and #3062 interacted badly and introduced two possible ways of releasing source files from the cache too early.

## Bug 1: building auto-imports while building a program

#2483 made the LoadOrStore operation of the ref counting cache not actually increment the ref count of the entry so we could remove some special handling of source file ref counts around program cloning. Instead, we would ref everything in a dedicated pass at the end of snapshot cloning. That was intended as a simplification to make the code easier to reason about, but having any period of time where entries in the cache have `refCount == 0` but can't be deleted was a bad idea. The failure mode:

- **Request A** causes a snapshot clone that builds a new program. Execution advances past program creation such that at least one source file is present in the parse cache with `refCount == 0`; the `parseCache.Ref()` call hasn't happened yet.
- **Request B** meanwhile triggers an auto-import registry build. It gets the same source file created by the program in Request A, increments its ref count to 1, finishes work, and decrements its ref count to 0, **causing it to be deleted from the cache.**
- **Request A** finally tries to ref every file included in the program it built, which panics because that file has been deleted from the cache in the meantime.

The fix here is to revert the simplification from #2483 and have parse cache entries begin life with `refCount == 1`.

The actual bug fixed in #2483 was not about parse cache entries at all, but about extended config cache entries, which used the same underlying ref counting cache infrastructure. The issue with extended config entries was their natural pattern of being pulled on by the config parser was extremely hard to follow, such that it was incredibly difficult to impossible to ensure that every snapshot refs every extended config exactly one time if incrementing the counter happened as a side effect of config parsing—that was where a single dedicated loop of reffing instead of reffing on the fly really paid off. As an alternative, I've switched extended config caches to store a set of owners instead of a count. This way, config parsing can immediately and atomically add the snapshot as an owner of anything it pulls on, but we can _also_ do a pass at the end of snapshot cloning to ref every extended config without being concerned about double-reffing.

These two changes are in 55c29ba00ce04c5f4b19bdd3e4bb3700ebb36bc0.

Note: I considered using the an owner cache for the parse cache too (owner key for that being program ID, not snapshot ID), but because there are typically orders of magnitude more source files than extended config files, I wanted to avoid the overhead of allocating a map on every parse cache entry.

## Bug 2: building auto-imports by cloning a disposed snapshot

#3062 was another simplification that asserted that we don't need ref counting on snapshots, because disposing a snapshot doesn't make it unclonable—at worst, it throws away parse cache entries that could have been reused, but they're reconstructable, and we don't have any code paths that would benefit from keeping the cache entries around.

That was true, but the implementation created an asymmetry when cloning a snapshot that's already been disposed:

- **Request A** asks for completions, obtains the current snapshot, and discovers that it needs to clone it to add auto-imports.
- **Request B** meanwhile processes a document change and produces a new snapshot, calling `Dispose()` on the previous one, still held by Request A. The disposal drops the ref count for the old program to zero, which triggers the release of some files from the parse cache.
- **Request A** clones the disposed snapshot to add auto-import info to it. No programs are updated, so parse cache entries are not incremented, but the old program still in the snapshot gets reffed in the program counter, recreating a new entry with `refCount == 1` for it, **an operation that can normally only happen in conjunction with reffing all its files in the parse cache.**
- **Request A** finishes, and since the session state has already moved on, the snapshot with added auto-imports gets disposed, not adopted by the session. Disposal derefs the program and its ref count **reaches zero for the second time, triggering an extra, unbalanced deref of all its source files.**
- **Request C** processes a single-file change. Each file in the new program is reffed, but any that reached a ref count of zero in the previous step are missing, triggering the panic.

The fix: for newly created programs, unconditionally add them to the program counter. Otherwise, increment their ref count only if they still have an entry in the counter. If the entry is gone, they've been previously disposed, and re-adding them would cause an eventual double-dispose. 07af679d483af8383eedd8b758de9bfabae5b8cb

Alternatives considered: many. Most tempting is to delete the program counter entirely and unconditionally ref every source file across every program in every snapshot clone exactly once per occurrence, and do the same exact loop in snapshot disposal. This was in an earlier commit of #3062, but I benchmarked it as adding about 10ms per 100k files. 100k files is a _lot_ and 10ms is a _little_, but at the same time, snapshot clones are _very hot_ (typically one per keystroke plus change), so at the time I decided it was worth the added mental load to keep it as cheap as possible. I could now be convinced to take on that cost, but, I think the fix here is correct, and actually not too much code. But I am getting tired of this stuff being a problem, so simplifying is tempting.


internal/project/refcountcache.go
```diff
@@ -4,14 +4,12 @@ import (
...
 
 	"github.com/microsoft/typescript-go/internal/collections"
-	"github.com/microsoft/typescript-go/internal/core"
 )
 
 type refCountCacheEntry[V any] struct {
 	mu       sync.Mutex
 	value    V
 	refCount int
-	deleted  bool
```
andrewbranch, line 14: No longer needed now that new entries always start with `refCount == 1`, at least that's nice

jakebailey:
Then again, the copilot comments do sound plausible

andrewbranch:
Yes, they're correct. The last commit is a solution but I don't like it. It's kind of pointing towards "no, snapshots really should have a ref count," but we saw in #2905 that it's pretty confusing trying to track where to release snapshots since they're sometimes used explicitly and sometimes used implicitly by the language service. Language service objects would have be disposed too, or always come with a snapshot release func, or be used inside a closure, or whatever. It's not impossible to implement carefully and correctly... maybe that's better? Ugh

internal/project/ownercache.go
```diff
@@ -0,0 +1,109 @@
...
+type ownerCacheEntry[V any] struct {
+	mu     sync.Mutex
+	value  V
+	owners map[uint64]struct{}
+}
+
+type OwnerCacheOptions struct {
+	DisableDeletion bool
+}
+
```
gabritto: Can we add documentation on what this is for?

andrewbranch: It wasn't used for anything, it was just copied from the ref count cache without its doc comment. Deleted.

gabritto: I meant on owner cache in general: what an owner cache is for. Also, I think we could probably merge the parse cache and this if we abstract the whole owner/ref counting thing into an interface or something, but not sure if it's worth it instead of maybe just adding a comment saying if you update one, you should update the other.

jakebailey:
> Language service objects would have be disposed too, or always come with a snapshot release func, or be used inside a closure, or whatever. It's not impossible to implement carefully and correctly... maybe that's better? Ugh

That was where my PR was headed; I just never committed the wrapper closure code since we opted to not go that way. I don't really mind so long as things are working 😄

gabritto:
General question: I thought the reason the parse cache didn't make entries start with ref count 1 was that we could somehow parse a file that was temporary and didn't make it into a program. That's not possible, is it?

andrewbranch:
@gabritto not exactly, you might be thinking of the awkwardness around updating a program: 

- if multiple files have changed, we don't try to do a cheap clone and every file in the new program gets acquired once
- if one file has changed, UpdateProgram acquires it to determine if it allows for a cheap clone, then:
  - if cloning is allowed, all other source files get copied in without being acquired, so we need to ref every file _except_ that changed one afterwards
  - if cloning is not allowed, all files _including the changed_ one get acquired again, resulting in the changed one being reffed twice, so we need to deref it afterwards. This was missing until the last commit, including prior to #2483—it was an undiscovered memory leak. That's a great example of why building programs without reffing, and then just reffing whatever ends up in the program, is easier 😞

internal/project/ownercache.go
andrewbranch: Yeah, I thought about sharing some code with an interface, but didn't want to introduce any unnecessary overhead into the parse cache. I can add a doc comment.

internal/project/compilerhost.go
```diff
@@ -90,12 +90,13 @@ func (c *compilerHost) GetResolvedProjectReference(fileName string, path tspath.
...
-// ref counting is handled at the snapshot level after program construction.
+// GetSourceFile implements compiler.CompilerHost. Files are cached in parseCache
+// and acquired immediately for the in-progress program.
 func (c *compilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast.SourceFile {
 	c.ensureAlive()
 	if fh := c.sourceFS.GetFileByPath(opts.FileName, opts.Path); fh != nil {
-		return c.builder.parseCache.Load(NewParseCacheKey(opts, fh.Hash(), fh.Kind()), fh)
+		key := NewParseCacheKey(opts, fh.Hash(), fh.Kind())
+		return c.builder.parseCache.Acquire(key, fh)
 	}
```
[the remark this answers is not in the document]
andrewbranch: (╯°□°)╯︵ ┻━┻

gabritto:
> @gabritto not exactly, you might be thinking of the awkwardness around updating a program:
> 
> * if multiple files have changed, we don't try to do a cheap clone and every file in the new program gets acquired once
> * if one file has changed, UpdateProgram acquires it to determine if it allows for a cheap clone, then:
>   
>   * if cloning is allowed, all other source files get copied in without being acquired, so we need to ref every file _except_ that changed one afterwards
>   * if cloning is not allowed, all files _including the changed_ one get acquired again, resulting in the changed one being reffed twice, so we need to deref it afterwards. This was missing until the last commit, including prior to [Fix extended config ref counting #2483](https://github.com/microsoft/typescript-go/pull/2483)—it was an undiscovered memory leak. That's a great example of why building programs without reffing, and then just reffing whatever ends up in the program, is easier 😞

Ah yes, I think that was it. Thanks for the explanation. Seems there's no ideal solution 😅

internal/project/compilerhost.go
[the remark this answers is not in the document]
andrewbranch: Fixed, but I hate it

andrewbranch:
I hate it all, but I don't have better ideas. We might consider going back to snapshot ref counts, but that would only eliminate the program counter's Add/Ref split and the extended config cache's TryAcquire. The parse cache complications would remain.

internal/compiler/filesparser.go
```diff
@@ -271,6 +271,7 @@ func (w *filesParser) getProcessedFiles(loader *fileLoader) processedFiles {
 
 	var missingFiles []string
 	files := make([]*ast.SourceFile, 0, totalFileCount-libFileCount)
+	duplicateSourceFiles := make([]*ast.SourceFile, 0)
```
jakebailey: Might as well just use `var duplicateSourceFiles []*ast.SourceFile`, otherwise it allocates an empty slice that's immediately overwritten on append.

internal/compiler/fileloader.go
```diff
@@ -69,16 +69,22 @@ func (r *redirectsFile) Path() tspath.Path {
...
-	resolver                      *module.Resolver
-	files                         []*ast.SourceFile
+	resolver *module.Resolver
+	files    []*ast.SourceFile
+	// duplicateSourceFiles tracks parsed files loaded during program construction
+	// that were later dropped from the final program, such as losing filename
+	// casing variants for the same path or files hidden behind package redirect
+	// deduplication. Their parse-cache acquires still need to be balanced when
+	// the program is disposed.
+	duplicateSourceFiles          []*ast.SourceFile
```
jakebailey: Having this here means we are going to hold onto these ASTs forever, since these are a part of the Program. That sort of defeats part of the purpose of the deduping (saving memory).

Do we have to hold onto the files themselves? Can we hold onto some other data about them?

andrewbranch: Oops, of course, good call 👍

andrewbranch:
I'm pretty convinced that restoring ref counts on snapshots, with automatic releasing by accessing snapshots and language services inside a callback, is the right thing to do. That would put a bit more churn into this PR. Would we rather include that here or look at it as a follow-up?

jakebailey:
However you want to do it; I think if there's a final state you want it'd be good to jump to straight to it in case there's a delay? But maybe this PR is fine on its own. Up to you.

internal/project/snapshot.go
```diff
@@ -489,11 +485,25 @@ func (s *Snapshot) Clone(ctx context.Context, change SnapshotChange, overlays ma
...
+}
+
+// Deref decrements the snapshot's reference count. When the count reaches
+// zero, the snapshot is disposed and its resources are released.
+func (s *Snapshot) Deref(session *Session) {
+	rc := s.refCount.Add(-1)
+	if rc < 0 {
+		panic(fmt.Sprintf("snapshot %d: ref count below zero, parentId=%d", s.id, s.parentId))
+	}
+	if rc == 0 {
```
jakebailey, line 515: Doesn't this introduce the same TOCTOU bug given this isn't locked on? e.g. a ref in the middle of a deref

andrewbranch: I don’t think so, because `Ref()` is always called with `session.snapshotMu` held. I'll make it private and update comments to ensure it stays that way.

jakebailey: Why is it an atomic, then?

andrewbranch: Two owners can validly `Deref()` at the same time, not under locks. The API server, for example, holds snapshots across multiple requests and derefs them only when the client releases them.

andrewbranch:
I can't get the flaky test to fail on my machine 😕

jakebailey:
I've been trying to get get a failing run of `TestDuplicatePackageServices_fileChanges` for what feels like months now, it's almost certainly nothing related to this PR.

andrewbranch:
Oh, geez, I didn't know that and this PR _did_ touch duplicate file stuff, so I thought it was my fault.

internal/project/snapshot.go
jakebailey: Hm, but that still seems like a race; one caller can be locked on snapshotMu adding a ref, but then another goroutine can be dereffing at the same time, no? The lock doesn't really change anything about the situation.

andrewbranch: In general yes, one caller can ref while another caller derefs. But the only way that's bad is if `refCount` can hit zero and then get revived back to one. Only session can ref, so that would imply that someone else has _exclusive ownership_ of a snapshot (`refCount == 1`) and derefs it, while somehow Session refs it. _That_ can’t happen—Session only ever refs snapshots _it already owns_, and its ownership is serialized under `snapshotMu` (one deref was out of place and fixed in the last commit).

jakebailey: Hm, okay; I don't have an incredible mental model for all of this at the moment.

andrewbranch:
@gabritto and/or @iisaduan was looking into running the fuzzer against this to hopefully get some extra validation before merging.

gabritto:
> @gabritto and/or @iisaduan was looking into running the fuzzer against this to hopefully get some extra validation before merging.

Testing this now

andrewbranch:
I downloaded the repros that looked like they could plausibly be caused by this PR and confirmed they fail the same way on main. I'm going to fix the merge conflict and then I think I'm satisfied to merge this.

PR #3178, opened 2026-03-20
Accept diffs where the only difference is union ordering

RyanCavanaugh:
This is based on the prior diff-accepting PR so please only review the appropriate commit

### DIFF
--- a/testdata/submoduleAccepted.txt
+++ b/testdata/submoduleAccepted.txt
@@ -213,6 +213,7 @@ conformance/typesWithSpecializedConstructSignatures.types.diff
 conformance/unionTypeReduction2.types.diff
 
 ## Different union ordering ##
+## Tracked at https://github.com/microsoft/typescript-go/issues/3180
 compiler/es2018ObjectAssign.types.diff
 compiler/stableTypeOrdering(stabletypeordering=false).types.diff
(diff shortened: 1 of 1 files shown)


testdata/baselines/reference/submoduleAccepted/conformance/typeGuardNarrowsToLiteralTypeUnion.types.diff
jakebailey, line 1: This one has union ordering, yes, but only because it's not reprinting the original node with parens in it

jakebailey: (in general, I think a lot of these remaining union ordering ones are things missing node reuse)

RyanCavanaugh: Let's track with bugs; logged https://github.com/microsoft/typescript-go/issues/3180

jakebailey: But, we're keeping it in accepted?

PR #3179, opened 2026-03-20
Fix issue #3014 

### LINKED ISSUE #3014: `tsgo --watch` hangs if given a tsconfig.json with no files
```json5
// tsconfig.json
{
  "files": []
}
```

Run

```
tsgo --watch -p ./tsconfig.json
```

The process doesn't seem to terminate.

JaskiratAnand:
**Problem**
When running `tsgo --watch` with a `tsconfig.json` containing an empty files array and no include rules, the compiler hung silently instead of reporting `TS18002` ("The 'files' list in config file... is empty").
This happened because the watcher polling loop (`internal/execute/watcher.go`) only triggered a build when `hasBeenModified` was `true`. On the first cycle, `prevModified` was empty, so no files appeared modified — skipping the initial build entirely and swallowing the diagnostic.

**Solution**
Treat the first execution cycle (`w.prevModified == nil`) as a modified event, ensuring the initial build always runs and configuration diagnostics are emitted correctly.

**Changes**
- `internal/execute/watcher.go` — force initial build on first cycle
- Added regression test: `watch with empty files list in tsconfig`

**Verification**
- Matches `tsc` behavior in watch mode (which correctly reports the diagnostic)
- New regression test passes with committed baseline
- `npx hereby test` passes

Fixes #3014

### DIFF
--- a/internal/execute/tsctests/tscwatch_test.go
+++ b/internal/execute/tsctests/tscwatch_test.go
@@ -23,6 +23,13 @@ func TestWatch(t *testing.T) {
 			},
 			commandLineArgs: []string{"--watch", "--incremental"},
 		},
+		{
+			subScenario: "watch with empty files list in tsconfig",
+			files: FileMap{
+				"/home/src/workspaces/project/tsconfig.json": `{"files": []}`,
+			},
+			commandLineArgs: []string{"--watch"},
+		},
 	}
 
 	for _, test := range testCases {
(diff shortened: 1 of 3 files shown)


JaskiratAnand:

JaskiratAnand:
Hey maintainers! @jakebailey @RyanCavanaugh  👋 Would love to get some eyes on this when you get a chance.

Quick summary: `tsgo --watch` was hanging silently instead of reporting `TS18002` when `tsconfig.json` had an empty `files` array. The fix treats the first watcher cycle as a modified event so the initial build always runs and diagnostics get emitted correctly — matching `tsc` behavior.

I've added a regression test and verified `npx hereby test` passes. Happy to make any changes if something doesn't look right. Thanks! 🙏

jakebailey:
Superseded by #3217

PR #3186, opened 2026-03-21
Accept changes solely due to _jsxFileName

jakebailey:
The old test harness didn't consistently use an absolute path, so these differ.

There's also:

```diff
--- old.commentsOnJSXExpressionsArePreserved(jsx=react-jsxdev,module=commonjs,moduledetection=legacy).js
+++ new.commentsOnJSXExpressionsArePreserved(jsx=react-jsxdev,module=commonjs,moduledetection=legacy).js
@@= skipped -23, +23 lines =@@

 //// [commentsOnJSXExpressionsArePreserved.js]
 "use strict";
-const _jsxFileName = "commentsOnJSXExpressionsArePreserved.tsx";
+const _jsxFileName = "/.src/commentsOnJSXExpressionsArePreserved.tsx";
 class Component {
     render() {
-        return (0, _a.jsxDEV)("div", { children: null /* preserved */ }, void 0, false, { fileName: _jsxFileName, lineNumber: 5, columnNumber: 16 }, this);
+        return _jsxDEV("div", { children: null /* preserved */ }, void 0, false, { fileName: _jsxFileName, lineNumber: 5, columnNumber: 16 }, this);
     }
 }
```

Which is missing an indirect call, but this is actually erroring code, and what Strada did was wrong (note `_a` is not declared!).


DanielRosenwasser:
Do the devtools that actually use this information still function correctly?

jakebailey:
I don't know, but I'll note that I checked other transformers like `babel` and they all emit absolute paths too.

jakebailey:
Er, well, I think babel just uses the filename it's given, so, that depends.

PR #3195, opened 2026-03-21
API encode/decode fixes

virtulis:
1. Added a number of missing node properties for the API encoder/decoder on both ends.
2. Added the token type for the HeritageClause, as far as I understand the bit usage is arbitrary per kind, so just used the first one.
3. Changed `isTypeNode` to also return true on type keywords, list copied from `KeywordTypeSyntaxKind`. These appear as is in parser output, which is probably correct?

All of these were hit mostly by just doing roundtrips of some test code between the parser and emitter via the API and `getSynthesizedDeepClone`. Probably a bunch more to be found that way.


virtulis:

virtulis:
> The encoder changes for ObjectLiteralExpression and IntersectionType shouldn't be necessary—we should only need to define a child mask if the node can have multiple children, some of which are optional. If these optional-single-child nodes are getting messed up, it's probably in the decoder logic, not an encoding mistake.

You're right, turns out that didn't do anything and only the `properties` getter needed adding. However, I think the real problem is that some properties are *not* typed as optional yet `getNamedChild` always treats them as such. I think it should have a flag to return an empty `NodeArray` for some properties.

> This will also need some tests—having an assertion function available to programmatically check round-trip consistency would be great, because then we could just feed in a large corpus of existing TS files from the repo to generate a lot of coverage. I'm happy to help with that, but feel free to give it a go if you want.

I'm a bit conflicted here, since I would love to get properly involved and understand the code beyond few specific functions, but I can't really charge my own clients for contributing to a Microsoft project under a CLA :upside_down_face:  so can't really promise anything at the moment.

virtulis:
Fwiw, what would you think about

1. Making the `generateFactory.ts`  script dump a full list of properties (and their properties) somewhere and
2. Instead of listing each getter manually on RemoteNode, just go over that list and defineProperty on the prototype? Since the actual interface is defined elsewhere anyway and those return types don't seem to be used.

That would also solve the optional vs empty NodeList issue.

internal/printer/printer.go
```diff
@@ -3173,7 +3173,7 @@ func (p *Printer) emitExpression(node *ast.Expression, precedence ast.OperatorPr
 
 	switch node.Kind {
 	// Keywords
-	case ast.KindTrueKeyword, ast.KindFalseKeyword, ast.KindNullKeyword:
+	case ast.KindTrueKeyword, ast.KindFalseKeyword, ast.KindNullKeyword, ast.KindUndefinedKeyword:
```
andrewbranch: It's hard to tell because Strada has fallthrough after expression-specific emit to general keyword and token printing, but I think this might be masking a different bug. AFAIK `ast.KindUndefinedKeyword` shouldn't appear in expressions—the source text `undefined` parses as an Identifier in expression positions, and as a keyword in type positions: https://ts-ast-viewer.com/#code/K4OwJgpgZgliFgNwCgqgMYBcYHsQAIoAeAFXwgA9MJwBnfUSWeMAPgAoBKfAbwF9kmAJ4AHCPjIBeBuGhwEiIA

Image

This is why this kind of change needs tests; how did this arise?

virtulis: Ah. Well, I guess I really should work on that roundtrip test before I make any more assumptions. It probably arose from my own code explicitly putting an UndefinedKeyword where it apparently shouldn't be. It worked with the original TS but I guess it shouldn't have :)

Will remove. Thoughts on the getter proposal? Looking into that now. Definitely need to do something about undefined where node array should be before any meaningful roundtrip is possible.

andrewbranch: Eventually (probably within a couple months), we want to have a big AST schema as the source of truth and codegen all this stuff and more, replacing `generateFactory.ts` and the hand-authored `RemoteNode` implementation, so I'm not really interested in metaprogramming it in the meantime.

virtulis: Understandable, but hope that also means I can submit a stand-in kludge to solve this particular issue? With full expectation it will be replaced with something decent, that's even better.

andrewbranch: These getters can unconditionally return non-undefined:

- statements
- parameters
- declarations
- elements
- types
- templateSpans
- properties
- children
- clauses

Then:

- `members` can return non-undefined if `kind` is not `MappedTypeNode`
- `typeParameters` can return non-undefined if `kind` is `JSDocTemplateTag`

Finally, `arguments` is optional only in `NewExpression`, where `undefined` indicates that the parens are omitted entirely. We can use one of the bool bits in the encoded node data for this.

virtulis:
Added the roundtrip test, made it go through everything and print each error since there's multiple different failures present.

_packages/api/src/node/node.ts
```diff
@@ -316,6 +316,13 @@ export class RemoteNodeList extends Array<RemoteNode> implements NodeArray<Remot
     }
 }
 
+// This cannot be an actual RemoteNodeList because we have nowhere to point to in the data
+function createStubNodeList() {
+    const array = [] as unknown as any[] & { pos: number; end: number; };
+    array.pos = array.end = -1;
```
andrewbranch, line 324: This is reminding me that NodeLists carry position information. We also want them to have reference equality when you access the same one twice. Looking at this now, I think it was a mistake to not encode empty ones into the response whenever they're non-nil on the Go side. Sorry for the churn—if at any point you want to be done, I can continue these fixes on top of your commits.

virtulis: > This is reminding me that NodeLists carry position information. We also want them to have reference equality when you access the same one twice. Looking at this now, I think it was a mistake to not encode empty ones into the response whenever they're non-nil on the Go side. 

I think it triggered some other issues when I experimented with changing that. But I also have no idea what I'm doing :)

>  Sorry for the churn—if at any point you want to be done, I can continue these fixes on top of your commits.

I've probably done everything I can with just poking around randomly. The roundtrip over TS sources is down to 16 failures (from hundreds, i think) and none of those look as obvious as the ones i've fixed.

And I probably should write some more specific tests for the changes but probably won't :) so let me know if you need me to undo anything, other than that I'm probably done for now. Hope this was somewhat helpful.

andrewbranch: Thank you!

virtulis: Actually, there's one extra test I wanted to try and did and immediately found some more issues. Will commit soon.

That roundtrip output should probably go somewhere other than the console but I'm not sure how or where (or whether it should even keep failing). Will leave that to you to figure out :)

virtulis:
There, the roundtrip thingy now also actually compares the output (currently just the words, since the whitespace is mangled, and i would assume some of the quotes etc might differ as well, but including those is a matter of changing a regex).

Flooding the console with poor man's "diffs" is certainly not great but I'm sure you know better what to do with that.

This found the broken TypeOperator (and my attempt at fixing it is included). Looks like there's also an issue with disappearing object spreads, and probably a bunch of other things.

Certainly done now, I promise. At least with this PR.

virtulis:
Ok so I ended up fixing almost all the remaining issues I found and removed the wonky input/output comparison, so the test now only checks for crashes (of which there are none now).

The only legitimate issue remaining there is that `namespace` blocks become `module` after a roundtrip. Not sure how much of a concern that is and the structure of the AST for those is rather confusing iirc.

andrewbranch:
Thanks for all the work on this! I'm in the process of working on the big "codegen everything" PR, a bit earlier than I anticipated since you and others found more inconsistencies than I expected. I'm planning to pull in your tests as part of the validation for that.

virtulis:
Nice. Fwiw, I removed the code to compare the contents of input/output because my attempt at ~~ZALGO~~ cleaning up comments with regexes failed spectacularly. Comparing the AST nodes via a different parser (Strada?) is probably a smarter approach. Not all of the issues I saw are covered with the separate tests.

I'll back off for now and stop being in your way. But as an aside, I really hope you publish some preview versions of the NPM packages soon, even if broken, as I'd really love to be able to depend on them in a less convoluted way.

andrewbranch:
Landed in #3367

PR #3196, opened 2026-03-22
Fixes a crash when renaming import path of a JSON file

Andarist:
fixes a crash found here https://github.com/microsoft/typescript-go/issues/3187#issuecomment-4101730891 (it also crashes in Strada)

### DIFF
--- a/internal/ls/findallreferences.go
+++ b/internal/ls/findallreferences.go
@@ -867,7 +867,7 @@ func (l *LanguageService) getReferencedSymbolsForNode(ctx context.Context, posit
 		}
 
 		if moduleSymbol := checker.GetMergedSymbol(resolvedRef.file.Symbol); moduleSymbol != nil {
-			return l.getReferencedSymbolsForModule(ctx, program, moduleSymbol /*excludeImportTypeOfExportEquals*/, false, sourceFiles, sourceFilesSet)
+			return l.getReferencedSymbolsForModule(ctx, program, moduleSymbol /*excludeImportTypeOfExportEquals*/, false, sourceFiles, sourceFilesSet, options)
 		}
 
 		// !!! not implemented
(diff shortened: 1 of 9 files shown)


internal/ls/findallreferences.go
```diff
@@ -1380,6 +1380,9 @@ func (l *LanguageService) getReferencedSymbolsForModule(ctx context.Context, pro
 	exported := symbol.Exports[ast.InternalSymbolNameExportEquals]
 	if exported != nil && len(exported.Declarations) > 0 {
 		for _, decl := range exported.Declarations {
+			if options.use == referenceUseRename && ast.IsSourceFile(decl) {
+				continue
+			}
```
Andarist, line 1385: I have tested the same patch in Strada and renaming the `from "./x/package.json"` resulted in the file itself being renamed too. That's definitely a desirable outcome but I'm not exactly sure what's the codepath leading to that filename rename - given it works... I wasn't exactly inclined to dig deeper. 

The alternative fix would be to handle this in the loop here:
https://github.com/microsoft/typescript-go/blob/a4325da30f285ff85b7b55afe1c65d74f54794af/internal/ls/rename.go#L82-L92

The crash happens in `getTextForRename`

gabritto: I need to think a bit more about all this, but I think this fix is ok. We're missing support for file rename in Corsa right now (Strada's `getEditsForFileRename`), planning on porting that next. But I think the right behavior would be to return just a [`RenameFile`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#renameFile) in this case and then compute the edits in response to a `willRenameFiles` request from the client. So the fix looks right, but we don't have a good way right now to properly test this. I think I will leave this PR open and revisit once file rename is ported, if that's ok.

gabritto:
Superseded by #3358

PR #3197, opened 2026-03-22
Fixed a go-to-implementations crash related to reexported type-only namespace

Andarist:
fixes the crash found here: https://github.com/microsoft/typescript-go/issues/3187#issuecomment-4101731639

### DIFF
--- a/internal/ls/utilities.go
+++ b/internal/ls/utilities.go
@@ -377,7 +377,7 @@ func isObjectBindingElementWithoutPropertyName(bindingElement *ast.Node) bool {
 }
 
 func isRightSideOfPropertyAccess(node *ast.Node) bool {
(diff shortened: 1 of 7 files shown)


internal/ls/utilities.go
```diff
@@ -377,7 +377,7 @@ func isObjectBindingElementWithoutPropertyName(bindingElement *ast.Node) bool {
 }
 
 func isRightSideOfPropertyAccess(node *ast.Node) bool {
-	return node.Parent.Kind == ast.KindPropertyAccessExpression && node.Parent.Name() == node
+	return node.Parent != nil && node.Parent.Kind == ast.KindPropertyAccessExpression && node.Parent.Name() == node
```
Andarist, line 380: Strada was nullish-safe here with its `tryCast`

PR #3208, opened 2026-03-24
[API] Formatting bug fixes

navya9singh:
These help with some formatting issues. Were fixed by Copilot.

Changes
  Add PrintNodeOptions support (api.ts, session.go, proto.go)

   - printNode now accepts optional PrintNodeOptions (preserveSourceNewlines, neverAsciiEscape, terminateUnterminatedLiterals).
   - Added Emitter.withOptions() to create a new emitter with merged options.
   - Options are passed through the API protocol to the Go printer.

  Fix printer emitLiteral (printer.go)

   - emitLiteral now checks TerminateUnterminatedLiterals from PrinterOptions, matching how NeverAsciiEscape is already handled.

  Tests (api.test.ts)

   - Added printNode with terminateUnterminatedLiterals option test — verifies the option is wired end-to-end. Note: the binary AST encoder does not currently 
  preserve TokenFlags.Unterminated for regex literals, so the printer cannot detect the literal is unterminated. The test asserts current behavior; once the encoder 
  preserves this flag, it should be updated to expect the terminated output.
   - Added printNode withOptions merges options test — verifies withOptions returns a new emitter without mutating the original.

  Known limitation

  terminateUnterminatedLiterals does not yet produce terminated output for regex literals because the binary AST encoder does not preserve the Unterminated token
  flag. The printer-side fix is in place and will work once the encoder is updated.

### DIFF
--- a/_packages/api/src/async/api.ts
+++ b/_packages/api/src/async/api.ts
@@ -304,15 +304,6 @@ export class Project {
         );
         this.emitter = new Emitter(client);
     }
-
-    async printNode(node: Node, options: PrintNodeOptions = {}): Promise<string> {
-        const encoded = encodeNode(node);
-        const base64 = uint8ArrayToBase64(encoded);
-        return this.client.apiRequest<string>("printNode", {
-            data: base64,
-            ...options,
-        });
-    }
 }
 
 export class Program {
--- a/_packages/api/src/sync/api.ts
+++ b/_packages/api/src/sync/api.ts
@@ -312,15 +312,6 @@ export class Project {
         );
         this.emitter = new Emitter(client);
     }
-
-    printNode(node: Node, options: PrintNodeOptions = {}): string {
-        const encoded = encodeNode(node);
-        const base64 = uint8ArrayToBase64(encoded);
-        return this.client.apiRequest<string>("printNode", {
-            data: base64,
-            ...options,
-        });
-    }
 }
 
 export class Program {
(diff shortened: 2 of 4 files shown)


_packages/api/src/async/api.ts
```diff
@@ -712,17 +712,33 @@ export class Checker {
...
+    private options: PrintNodeOptions;
 
-    constructor(client: Client) {
+    constructor(client: Client, options: PrintNodeOptions = {}) {
         this.client = client;
+        this.options = options;
+    }
+
+    /** Return a new Emitter that uses the given options, sharing the same client. */
+    withOptions(options: PrintNodeOptions): Emitter {
```
andrewbranch: Let's just pass the options to `printNode` for now.

internal/printer/printer.go
```diff
@@ -992,6 +992,9 @@ func (p *Printer) emitLiteral(node *ast.LiteralLikeNode, flags getLiteralTextFla
 	if p.Options.NeverAsciiEscape {
 		flags |= getLiteralTextFlagsNeverAsciiEscape
 	}
+	if p.Options.TerminateUnterminatedLiterals {
+		flags |= getLiteralTextFlagsTerminateUnterminatedLiterals
+	}
```
andrewbranch, line 997: Is this present in Strada?

navya9singh: Yeah, it's in `getLiteralTextOfNode()`

_packages/api/src/async/api.ts
```diff
@@ -302,6 +304,15 @@ export class Project {
         );
         this.emitter = new Emitter(client);
     }
+
+    async printNode(node: Node, options: PrintNodeOptions = {}): Promise<string> {
```
andrewbranch: This is duplicated

navya9singh: Sorry about that. Removed it now.

PR #3210, opened 2026-03-24
Disable nil-context type-node reuse

Andarist:


internal/checker/nodebuilderimpl.go
```diff
@@ -2070,7 +2070,7 @@ func (b *NodeBuilderImpl) serializeTypeForDeclaration(declaration *ast.Declarati
 	}
 	var result *ast.Node
 	// !!! expandable hover support
-	if tryReuse && declaration != nil && (ast.IsAccessor(declaration) || (ast.HasInferredType(declaration) && !ast.NodeIsSynthesized(declaration) && (t.ObjectFlags()&ObjectFlagsRequiresWidening) == 0)) {
+	if tryReuse && b.ctx.enclosingDeclaration != nil && declaration != nil && (ast.IsAccessor(declaration) || (ast.HasInferredType(declaration) && !ast.NodeIsSynthesized(declaration) && (t.ObjectFlags()&ObjectFlagsRequiresWidening) == 0)) {
```
Andarist, line 2073: I'm opening this to get a vibe-check if this is desirable. It's likely a similar fix has to be applied in other places too. This branch in Strada led to calling `canReuseTypeNodeAnnotation` and that bailed on:
```ts
if (context.enclosingDeclaration === undefined) return false;
```

testdata/baselines/reference/submodule/compiler/intersectionsAndOptionalProperties.errors.txt
```diff
@@ -17,12 +17,12 @@ intersectionsAndOptionalProperties.ts(20,5): error TS2322: Type 'null' is not as
     
     x = y;  // Error
     ~
-!!! error TS2322: Type '{ a: null; b: string; }' is not assignable to type '{ a?: number; b: string; }'.
+!!! error TS2322: Type '{ a: null; b: string; }' is not assignable to type '{ a?: number | undefined; b: string; }'.
 !!! error TS2322:   Types of property 'a' are incompatible.
```
jakebailey, line 21: I'm not entirely sure about this PR; note that the original code above did write `a?: number`. I think the new code just may be smarter about still pulling info out of the original declarations?

internal/checker/nodebuilderimpl.go
jakebailey: > It's likely a similar fix has to be applied in other places too.

I think it'd be interesting to look for those and see if they're even needed. This code all works differently.

PR #3212, opened 2026-03-24
Fix error spans for `@satisfies`

Andarist:


internal/scanner/scanner.go
```diff
@@ -2535,6 +2535,20 @@ func GetErrorRangeForNode(sourceFile *ast.SourceFile, node *ast.Node) core.TextR
...
+					if tags := jsDoc.AsJSDoc().Tags; tags != nil {
+						for _, tag := range tags.Nodes {
+							if ast.IsJSDocSatisfiesTag(tag) {
+								pos := SkipTrivia(sourceFile.Text(), tag.TagName().Pos())
+								return GetRangeOfTokenAtPosition(sourceFile, pos)
+							}
+						}
+					}
+				}
+			}
```
Andarist: I'm not super fond of this but I couldn't find a better way to go from a reparsed node to the original JSDoc right now, and I tried a bunch. Am I missing something?

jakebailey: I'm not sure there's anything better, given we are in the scanner package and therefore if such a helper existed in `astnav` it would be unavailable.

But maybe @gabritto / @andrewbranch have a better thought on their minds.

jakebailey: I don't think there's a better way. However, I think you could pull out a helper and avoid the `nil` call, like:

```go
func findOriginatingJSDocSatisfiesTag(sourceFile *ast.SourceFile, node *ast.Node) *ast.Node {
	targetType := node.AsSatisfiesExpression().Type
	if targetType.Flags&ast.NodeFlagsReparsed == 0 {
		return nil
	}
	for current := node.Parent; current != nil; current = current.Parent {
		if current.Flags&ast.NodeFlagsHasJSDoc == 0 {
			continue
		}
		var firstSatisfiesTag *ast.Node
		for _, jsDoc := range current.JSDoc(sourceFile) {
			if tags := jsDoc.AsJSDoc().Tags; tags != nil {
				for _, tag := range tags.Nodes {
					if !ast.IsJSDocSatisfiesTag(tag) {
						continue
					}
					if firstSatisfiesTag == nil {
						firstSatisfiesTag = tag
					}
					if typeExpr := tag.AsJSDocSatisfiesTag().TypeExpression; typeExpr != nil {
						if t := typeExpr.Type(); t != nil && t.Loc == targetType.Loc {
							return tag
						}
					}
				}
			}
		}
		return firstSatisfiesTag
	}
	return nil
}
```

Or something.

jakebailey:
This is failing all the tests

internal/scanner/scanner.go
```diff
@@ -2503,6 +2503,38 @@ func getErrorRangeForArrowFunction(sourceFile *ast.SourceFile, node *ast.Node) c
...
+	targetType := node.AsSatisfiesExpression().Type
+	if targetType.Flags&ast.NodeFlagsReparsed == 0 {
+		return nil
+	}
+	for current := node.Parent; current != nil; current = current.Parent {
+		if current.Flags&ast.NodeFlagsHasJSDoc == 0 {
+			continue
+		}
+		var firstSatisfiesTag *ast.Node
+		for _, jsDoc := range current.JSDoc(sourceFile) {
```
jakebailey: Should this be `EagerJSDoc` to avoid parsing JSDoc in TS files with satisifes, or is that not what happens?

Andarist: This is already kinda guarded by `NodeFlagsReparsed` - so it shouldn't end up parsing JSDoc in TS files

jakebailey: Right, though we `continue`, so we are still going to walk up the entire tree to the SourceFile?

Andarist: This is only walked when there is an error to be raised - so it should hit some actual node with `@satisfies` before reaching the `SourceFile`

jakebailey: Fair, though I think I'd still like to see `EagerJSDoc` used just because the checker does not actually call `JSDoc` at all when working with these tags, I think?

jakebailey (dismissed):

Andarist:
> This is failing all the tests

sorry about that, the failures crept in here after syncing with main ;p

internal/scanner/scanner.go
Andarist: I just pushed out this change, so you'll have to approve the PR again

jakebailey (approved):

andrewbranch (approved):

PR #3216, opened 2026-03-24
fix: clean up failed pprof startup files

buley:
## Summary

`BeginProfiling` creates the CPU profile file before calling `runtime/pprof.StartCPUProfile`.
If startup fails, it panics and leaves the partially created file behind even though `CPUProfiler.StartCPUProfile` already cleans up that path.

## Fix

Use a shared start hook for both profiling entry points, close and remove the startup file before panicking from `BeginProfiling`, and add a unit test that forces the failure path and verifies cleanup.

## Testing

- `go test ./internal/pprof`

## Related

- 6.0-side PR: microsoft/TypeScript#63290


RyanCavanaugh:
Hardening internal debug/profiling codepaths against benign failures isn't a great use of churn/risk. It's fine if a local dev has to delete this file manually after killing the process or whatever.

buley:
Copy that - thanks for the clear answer @RyanCavanaugh

PR #3217, opened 2026-03-24
Watch cli efficiency update

### LINKED ISSUE #1317: High CPU usage with --watch when idling on MacOS
When I'm running `$ tsgo --watch` I am seeing consistently high CPU usage when idling on a Mac -- around 120%-170% as reported by Activity Monitor. No files are being changed and nothing is being worked on.

```
$ tsgo --watch

build starting at 2025-06-30 14:25:39.948731 +0300 EEST m=+0.611922626
build finished in 1.668952292s
```

Image

Unsure whether this is expected due to the experimental implementation of watch mode for now, but thought I would report, as I am not seeing the same usage with `npx tsc -w`, which stays flat at 0 when there is no activity.

For context:

tsconfig.json:
```
{
    "compilerOptions": {
        "allowJs": false,
        "lib": ["es2019", "dom"],
        "moduleResolution": "node",
        "target": "es2019",
        "module": "commonjs",
        "noImplicitReturns": true,
        "noFallthroughCasesInSwitch": true,
        "noUnusedLocals": true,
        "pretty": true,
        "noImplicitAny": true,
        "noImplicitThis": false,
        "alwaysStrict": true,
        "strictNullChecks": true,
        "strictFunctionTypes": true,
        "strictPropertyInitialization": false,
        "sourceMap": true,
        "outDir": "./dist",
        "skipLibCheck": true,
        "resolveJsonModule": true,
        "esModuleInterop": true,
        "declaration": true,
        "baseUrl": ".",
        "typeRoots": ["./node_modules/@types", "./@types"],
        "forceConsistentCasingInFileNames": true,
        "paths": {
            "src": ["./src"]
        }
    },
    "include": ["src/**/*.ts", "src/sites/**/*.json", "@types/**/*"],
    "exclude": ["node_modules", "dist", ".git"]
}
```

Thanks!

### LINKED ISSUE #2317: Console is not cleared in watch mode (also does not show the 0 error count output)
## Steps to reproduce
I use  tsgo --watch --noEmit --skipLibCheck on Win11's Terminal with pwsh 7.5.x
The console is not cleared between watched builds (there is --preserveWatchOutput which defaults to false, in other words the console should be cleared in between)

Sidenote: I also have cases where i press CTRL+S in my IDE (with an error present) in that moment the compiler watcher starts a new compilation, but does not detect the error in the file. If i change the file again, e.g. inserting a new blank line and save, now the error is picked up.

tsc --watch --noEmit --skipLibCheck

clears the console and shows
[08:31:19] File change detected. Starting incremental compilation...

[08:31:24] Found 0 errors. Watching for file changes.

## Behavior with `tsgo`
tsgo --watch --noEmit --skipLibCheck

the new entries are appended in the console, without clearing (also does not show the 0 errors message)

Found 2 errors in the same file, starting at: src/sample.ts:4

build finished in 0.016s
build starting at 08:37:29 AM
build finished in 0.006s
build starting at 08:37:31 AM
build finished in 0.003s
build starting at 08:37:33 AM
build finished in 0.002s
build starting at 08:37:35 AM
build finished in 0.002s

### LINKED ISSUE #2640: tsgo --watch is lagging and sometimes not emitting at all
## Steps to reproduce

https://github.com/tharakadesilva/tsgo-2640

* `pnpm i`
* `pnpm dev`
* Observe `packages/ui/dist/index.js` while updating `packages/ui/src/index.tsx`

`tsc --watch` works as expected and the output is as expected,

https://github.com/user-attachments/assets/24d8faf3-41fb-4722-9932-a7ce88118dbf

## Behavior with `tsgo`

`tsgo --watch` doesn't update immediately even though it says that it built in a fraction of a second. Sometimes, it doesn't update at all.

https://github.com/user-attachments/assets/f78e5662-0c2c-45bf-957c-52a06567c169

### LINKED ISSUE #3014: `tsgo --watch` hangs if given a tsconfig.json with no files

### LINKED ISSUE #3015: panic on `--watch` and `--explainFiles`
```sh
tsgo --watch --explainFiles anyFileEvenOneThatDoesNotExist.ts
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x716ace]

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/tsoptions.(*ParsedCommandLine).Locale(0x1ea7f1aeb758?)
        github.com/microsoft/typescript-go/internal/tsoptions/parsedcommandline.go:388 +0xe
github.com/microsoft/typescript-go/internal/execute/tsc.listFiles({{0x10b39c0, 0x1ea7edc2cfc0}, {0x10b67d0, 0x1ea7f1725290}, 0x1ea7edc60fc8, 0x0, 0x1ea7edb3fe40, 0x1ea7edb3fe60, {0x10a4e40, 0x1ea7eda58048}, ...}, ...)
        github.com/microsoft/typescript-go/internal/execute/tsc/emit.go:138 +0x165
github.com/microsoft/typescript-go/internal/execute/tsc.EmitFilesAndReportErrors({{0x10b39c0, 0x1ea7edc2cfc0}, {0x10b67d0, 0x1ea7f1725290}, 0x1ea7edc60fc8, 0x0, 0x1ea7edb3fe40, 0x1ea7edb3fe60, {0x10a4e40, 0x1ea7eda58048}, ...})
        github.com/microsoft/typescript-go/internal/execute/tsc/emit.go:117 +0x50b
github.com/microsoft/typescript-go/internal/execute.(*Watcher).compileAndEmit(0x1ea7edc5c780)
        github.com/microsoft/typescript-go/internal/execute/watcher.go:101 +0x176
github.com/microsoft/typescript-go/internal/execute.(*Watcher).DoCycle(0x1ea7edc5c780)
        github.com/microsoft/typescript-go/internal/execute/watcher.go:87 +0x1c5
github.com/microsoft/typescript-go/internal/execute.(*Watcher).start(0x1ea7edc5c780)
        github.com/microsoft/typescript-go/internal/execute/watcher.go:62 +0x276
github.com/microsoft/typescript-go/internal/execute.tscCompilation({0x10b39c0, 0x1ea7edc2cfc0}, 0x1ea7edc90000, {0x0, 0x0})
        github.com/microsoft/typescript-go/internal/execute/tsc.go:215 +0x1165
github.com/microsoft/typescript-go/internal/execute.CommandLine({0x10b39c0, 0x1ea7edc2cfc0}, {0x1ea7eda78060, 0x4, 0x4}, {0x0, 0x0})
        github.com/microsoft/typescript-go/internal/execute/tsc.go:36 +0x145
main.runMain()
        github.com/microsoft/typescript-go/cmd/tsgo/main.go:23 +0x10c
main.main()
        github.com/microsoft/typescript-go/cmd/tsgo/main.go:10 +0x13
```

johnfav03:
The watch CLI option previously recompiled the program on a set interval even when files didn't change; this PR implements several efficiency fixes for this feature including debouncing, caching config, and parsing dependencies to recompile only when files have been changed. This PR also fixes several open issues with the watcher.

Fixes #1317
Fixes #2317
Fixes #2640
Fixes #3014
Fixes #3015


internal/execute/watcher.go
```diff
@@ -111,57 +182,94 @@ func (w *Watcher) compileAndEmit() {
...
+		return true
+	}
+	w.configHasErrors = false
+	if !reflect.DeepEqual(w.config.CompilerOptions().Clone(), configParseResult.CompilerOptions().Clone()) ||
+		!slices.Equal(w.config.FileNames(), configParseResult.FileNames()) {
+		w.configModified = true
 	}
-	w.host = compiler.NewCompilerHost(w.sys.GetCurrentDirectory(), w.sys.FS(), w.sys.DefaultLibraryPath(), extendedConfigCache, getTraceFromSys(w.sys, w.config.Locale(), w.testing))
+	w.config = configParseResult
+	w.extendedConfigCache = extendedConfigCache
```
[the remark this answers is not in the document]
Mateusnasciment: Fix:

     1 if w.configHasErrors {
     2     w.configModified = true  // Force rebuild when errors cleaaar
     3 }
     4 w.configHasErrors = false

andrewbranch:
The two things I see off the bat that are missing that we discussed while whiteboarding:

1. This doesn't include the config's `include` wildcard directories. I _think_ the `watch detects file added to previously non-existent include path` test only works because the edit actually creates the `src` directory. If the directory already existed and a new file was added into it, it looks like the watcher won't pick it up.
2. We talked about being able to plug this watcher into the LSP by separating the program building logic from the watching logic, but that isn't done here; it's all together. I don't think that's a blocker for right now, but I also don't think it's very much work to split it up—I want a watcher that takes a list of files or globs and calls a callback when something changes. I feel like if you just replace `w.doBuild()` with `w.callback()`, and have the program and caches owned by a different struct, you're pretty much there. If it's reasonably low-effort, it might be a good idea to decouple them now so it doesn't inadvertently grow harder to untangle over time.

johnfav03:
> 2. We talked about being able to plug this watcher into the LSP by separating the program building logic from the watching logic, but that isn't done here; it's all together. I don't think that's a blocker for right now, but I also don't think it's very much work to split it up—I want a watcher that takes a list of files or globs and calls a callback when something changes. I feel like if you just replace `w.doBuild()` with `w.callback()`, and have the program and caches owned by a different struct, you're pretty much there. If it's reasonably low-effort, it might be a good idea to decouple them now so it doesn't inadvertently grow harder to untangle over time.

I just implemented this change by seperating the file tracking into a `FileWatcher` struct, which builds off of the previous `watchState` and adds a `callback` field. In the case of the CLI, this callback is linked to `doCycle`, which contains `doBuild` and all of the program construction logic. The caches, program construction, and emitting logic are all still owned by `Watcher`, which is wrapped around `FileWatcher`.

internal/execute/watcher.go
```diff
@@ -142,6 +145,51 @@ func (fw *FileWatcher) Run(now func() time.Time) {
...
 }
 
+type cachedSourceFile struct {
+	file    *ast.SourceFile
+	modTime time.Time
+}
+
+type watchCompilerHost struct {
+	inner compiler.CompilerHost
+	cache *collections.SyncMap[tspath.Path, *cachedSourceFile]
```
jakebailey, line 29: I'm not sure about this method; this will grow indefinitely, no? There's a delete below, but that's only going to happen if a second run happens to call "get" on the source file and it doesn't exist, but if a user removes an import, thus leaving files in say `node_modules` unreferenced, those will stick around forever, right?

johnfav03: Good catch - I'll add logic to evict any unused files from the cache after the build completes

andrewbranch:
I don't think https://github.com/microsoft/typescript-go/pull/3217/commits/5ef0db6f330745e304661571bafc595bf018b458 is sufficient for wildcard watching—I can't quite figure out what the baseline tests are saying there (what does “incremental skips emit for new unreferenced file” mean?) but I think you’re going to have to read the directory (recursively if the pattern is recursive) to see if there are any new files.

internal/execute/watcher.go
```diff
@@ -1,30 +1,241 @@
...
 	"github.com/microsoft/typescript-go/internal/tsoptions"
+	"github.com/microsoft/typescript-go/internal/tspath"
+	"github.com/microsoft/typescript-go/internal/vfs"
+	"github.com/microsoft/typescript-go/internal/vfs/cachedvfs"
+	"github.com/microsoft/typescript-go/internal/vfs/trackingvfs"
 )
 
+const watchDebounceWait = 250 * time.Millisecond
+
+type FileWatcher struct {
```
jakebailey: I think that this needs to move into its own package. If we want to use this from the LS, it shouldn't live in the execute package. Maybe `vfs/vfswatch`?

internal/vfs/vfswatch/vfswatch.go
```diff
@@ -0,0 +1,164 @@
...
+
+const DebounceWait = 250 * time.Millisecond
+
+type FileWatcher struct {
+	fs                  vfs.FS
+	PollInterval        time.Duration
+	testing             bool
+	callback            func()
+	WatchState          map[string]trackingvfs.WatchEntry
+	WildcardDirectories map[string]bool
```
jakebailey: I suspect that this concurrency model is not going to work out; from my reading, `execute` is reading these out while the watcher is running, right?

internal/execute/watcher.go
```diff
@@ -1,30 +1,88 @@
...
+}
+
+type watchCompilerHost struct {
+	inner compiler.CompilerHost
+	cache *collections.SyncMap[tspath.Path, *cachedSourceFile]
+}
+
+var _ compiler.CompilerHost = (*watchCompilerHost)(nil)
+
+func (h *watchCompilerHost) FS() vfs.FS                  { return h.inner.FS() }
```
jakebailey: I think you can actually just use `wrapvfs` for this

jakebailey: Though, that might be more annoying due to callbacks instead of just methods.

or, just embed CompilerHost

andrewbranch:
Like Jake, I thought I saw some concurrency bugs, so I threw Copilot at it to write some tests to exercise them, and it did indeed find some with the race detector. Commit: https://github.com/microsoft/typescript-go/commit/365407c4c1717511fb0d6b5a66987271153ae225

Summary:

```
  Races found (run with go test -race)

  Running the vfswatch tests produces 16+ distinct race warnings, and the watcher tests produce 394 race warnings across:

  ┌───────────────────────┬────────────────────────────────┬──────────────────────────────────────────────────────────────────┐
  │ Field                 │ Location                       │ Issue                                                            │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ WatchState            │ vfswatch.go:33,36,38,114,126   │ Plain map read in HasChanges, written in UpdateWatchedFiles      │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ WildcardDirectories   │ vfswatch.go:42,52,54,57,126    │ Plain map read in HasChanges/currentState, assigned in doBuild   │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ PollInterval          │ vfswatch.go                    │ Plain time.Duration read in Run, written in doBuild              │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ configModified        │ watcher.go:141,185             │ Plain bool read/written without sync                             │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ config                │ watcher.go:162-184,263-267     │ Pointer replaced in hasErrorsInTsConfig, read in doBuild         │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ program               │ watcher.go:176,203             │ Pointer replaced in doBuild, read in compileAndEmit              │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ configHasErrors       │ watcher.go:228,261             │ Plain bool read/written in hasErrorsInTsConfig                   │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ configFilePaths       │ watcher.go:229,262             │ Slice replaced without sync                                      │
  ├───────────────────────┼────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ extendedConfigCache   │ watcher.go:159,267             │ Pointer replaced without sync                                    │
  └───────────────────────┴────────────────────────────────┴──────────────────────────────────────────────────────────────────┘
```

johnfav03:
> Like Jake, I thought I saw some concurrency bugs, so I threw Copilot at it to write some tests to exercise them, and it did indeed find some with the race detector.

Thanks for the race tests! I added mutexes to the `Watcher` and `FileWatcher` structs to govern access to `WatchState` and `WildcardDirectories`. Also, I had copilot analyze the same sets of tests and it doesn't turn up any race warnings now.

jakebailey:
I don't want to block this PR on being able to use it within the LSP, really; if it's going to take too long I think it's optional given the CLI stuff is important.

But, we know there are clients that can't do watch mode at all, so I think it's in general worth making work if we can get it out faster than it takes to hack the VS LSP client. @joj or @navya9singh would know better how hard that is to manage, but I think it's "as simple" as handling our watch requests manually and then just using dotnet's watcher to watch a dir recursively, as that is actually built into dotnet.

johnfav03:
To address the first suggestion, I combined `SetWildcardDirectories` and `UpdateWatchedFiles` into `UpdateWatchState` which takes in both wildcard dirs and seen files. Once we get more feedback on whether it's better to handle file watching via the LSP or the client, I'll open a new PR and make the changes for the second suggestion there.

internal/vfs/vfswatch/vfswatch.go
```diff
@@ -0,0 +1,214 @@
...
+		if s := fw.fs.Stat(path); s != nil {
+			state[path] = WatchEntry{ModTime: s.ModTime(), Exists: true, ChildCount: -1}
+		} else {
+			state[path] = WatchEntry{Exists: false, ChildCount: -1}
+		}
+	}
+	for dir, recursive := range wildcardDirs {
+		if !recursive {
+			continue
+		}
```
andrewbranch, line 135: Don't you need to go one level into non-recursive directories? This is just skipping them.

internal/vfs/vfswatch/vfswatch.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+			entries := fw.fs.GetAccessibleEntries(path)
+			count := len(entries.Files) + len(entries.Directories)
+			if existing, ok := state[path]; ok {
+				existing.ChildCount = count
+				state[path] = existing
+			} else {
+				if s := fw.fs.Stat(path); s != nil {
+					state[path] = WatchEntry{ModTime: s.ModTime(), Exists: true, ChildCount: count}
+				}
```
andrewbranch: Oh, interesting, you're not statting each file; you're just comparing the number of entries...

This has a now problem and a later problem:

- Now: a rename, or any other balanced sequence of adds and removes within a cycle, won't get picked up. An easy example to repro would be renaming a `.js` file to `.ts` in a tsconfig that only `includes` TS files.
- Later: This never detects changes, only adds/deletes. This is not a "now" problem because wildcard directories for tsconfigs really only need to detect creates; anything you care about getting changed or deleted will necessarily be in `watchState` via the `paths` argument to `UpdateWatchState`. But it's not really clear from the API that this is a limitation; maybe just add a comment to `UpdateWatchState`.

You could fix the rename problem by storing a hash of the data returned by GetAccessibleEntries, but if you want to eventually be able to tell the callback _what_ was created and deleted, you'll end up storing not just a hash, but the actual data so you can diff it if changed.

The contract the LSP uses is that when it registers a glob to watch, it says what event types (create, delete, change) it cares about. If it doesn't ask for changes, this is the right approach. If it does want changes, you'd have to do an additional pass to stat each file. Right now, instead of passing VS Code a huge list of files that we want changes for and a small list of directories we want creates for, we reduce that to a small list of directories registered for all events. This is cheaper for LSP watching because the list of files to watch can be so huge that it bogs down the actual LSP communication layer. If we were to switch to using our own in-process watcher, that dynamic could definitely change; we'll have to benchmark different approaches.

TLDR: for now, just fix the rename problem somehow.

internal/vfs/vfswatch/vfswatch.go
```diff
@@ -122,25 +124,40 @@ func snapshotPaths(fs vfs.FS, paths []string, wildcardDirs map[string]bool) map[
...
 	}
 }
 
+// hashEntries returns a hash of the sorted file and directory names
+// within a directory. This detects adds, deletes, and renames.
+func hashEntries(entries vfs.Entries) uint64 {
+	names := make([]string, 0, len(entries.Files)+len(entries.Directories))
+	names = append(names, entries.Files...)
+	names = append(names, entries.Directories...)
+	slices.Sort(names)
```
andrewbranch: Sorry to nitpick, but treating files and directories equivalently means the hash won't change if you replace a file with a directory or vice versa 🤪. While you're at it, we use `xxhash` elsewhere in the repo and I think it's faster than fnv.

Issue #3207, opened 2026-03-24
`isolatedDeclarations` false positive TS9013 on arrow functions with default parameters and `as const` objects


jakeboone02:
This issue was discovered on the https://github.com/react-querybuilder/react-querybuilder repo.

## Steps to reproduce

```ts
// repro.ts
const imported = ',';

export const fn1 = (x: string = imported): string => x;

interface Opts { val?: boolean }
export const fn2 = ({ val = false }: Opts = {}): boolean => val;

export const obj1 = { $and: [{ $expr: true }] } as const;
```

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "noEmit": true,
    "declaration": true,
    "isolatedDeclarations": true,
    "strict": true,
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler"
  },
  "include": ["repro.ts"]
}
```

## Behavior with `typescript@6.0.2`

```
$ tsc -p tsconfig.json
// No errors
```

## Behavior with `tsgo` (7.0.0-dev.20260324.1)

```
$ tsgo -p tsconfig.json
repro.ts:3:21 - error TS9013: Expression type can't be inferred with --isolatedDeclarations.
repro.ts:6:21 - error TS9013: Expression type can't be inferred with --isolatedDeclarations.
repro.ts:8:29 - error TS9013: Expression type can't be inferred with --isolatedDeclarations.
repro.ts:8:39 - error TS9013: Expression type can't be inferred with --isolatedDeclarations.

Found 4 errors in the same file, starting at: repro.ts:3
```

## Additional context

These appear to be regressions from #2459 (isolatedDeclarations port, merged 2026-03-20). The same class of false positive was previously reported and fixed in tsc:

- [TypeScript#60192](https://github.com/microsoft/TypeScript/issues/60192) — false positive on arrow functions with typed optional/default parameters
- [TypeScript#60195](https://github.com/microsoft/TypeScript/pull/60195) — the fix

Two distinct patterns are affected:
1. **Arrow functions with default parameters** — `export const fn = (param: Type = expr): ReturnType => ...` is flagged even though the parameter already has a type annotation
2. **`as const` with nested object/array literals** — `export const obj = { key: [{ nested: true }] } as const` is flagged

hkleungai:
> **`as const` with nested object/array literals**

Seems to be a known issue on this one. Ref: https://github.com/microsoft/typescript-go/issues/3192#issuecomment-4105762522

[...]

PR #3219, opened 2026-03-25
Use isOptionalParameter from checker to check param optional-ness

jakebailey:
Addresses the first part of #3207; `isOptionalDeclaration` only checks for the question node.

### DIFF
--- a/internal/checker/utilities.go
+++ b/internal/checker/utilities.go
@@ -10,6 +10,7 @@ import (
 	"github.com/microsoft/typescript-go/internal/ast"
 	"github.com/microsoft/typescript-go/internal/binder"
 	"github.com/microsoft/typescript-go/internal/core"
+	"github.com/microsoft/typescript-go/internal/debug"
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 	"github.com/microsoft/typescript-go/internal/jsnum"
 	"github.com/microsoft/typescript-go/internal/module"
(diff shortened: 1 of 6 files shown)


internal/checker/utilities.go
```diff
@@ -276,6 +277,33 @@ func isOptionalDeclaration(declaration *ast.Node) bool {
...
+		return false
+	}
+	if node.Initializer() != nil {
+		signature := c.getSignatureFromDeclaration(node.Parent)
+		parameterIndex := core.FindIndex(node.Parent.Parameters(), func(p *ast.ParameterDeclarationNode) bool { return p == node })
+		debug.Assert(parameterIndex >= 0)
+		// Only consider syntactic or instantiated parameters as optional, not `void` parameters as this function is used
+		// in grammar checks and checking for `void` too early results in parameter types widening too early
+		// and causes some noImplicitAny errors to be lost.
+		return parameterIndex >= c.getMinArgumentCountEx(signature, MinArgumentCountFlagsStrongArityForUntypedJS|MinArgumentCountFlagsVoidIsNonOptional)
```
DanielRosenwasser, line 295: Maybe you can avoid the issue that the comment above raised about O(n^2) by using the node's pos here (being careful to handle out-of-bounds and synthesized nodes).

jakebailey: Perhaps, though, this is sort of preexisting...

jakebailey:
Yeah, I started with a syntactic one only, but that didn't work in many cases...

PR #3221, opened 2026-03-25
Fix ProbablyUsesSemicolons threshold comparison (integer division)

cuiweixie:
## Summary

Fixes the ratio check in `ProbablyUsesSemicolons` in `internal/ls/lsutil/utilities.go`.

In Go, `1/nStatementsToObserve` is integer division and evaluates to `0` when `nStatementsToObserve` is `5`, so the heuristic compared `withSemicolon/withoutSemicolon` against `0` instead of `1/5`.

Use an equivalent inequality that avoids floating point and matches the intended threshold: `withSemicolon * nStatementsToObserve > withoutSemicolon`.

### DIFF
--- a/internal/ls/lsutil/utilities.go
+++ b/internal/ls/lsutil/utilities.go
@@ -64,7 +64,9 @@ func ProbablyUsesSemicolons(file *ast.SourceFile) bool {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


cuiweixie:

PR #3226, opened 2026-03-25
Add lint rule forbidding .Parent accesses in transformers

weswigham:
And fix or exempt most existing refs. Having this should automate waggling the finger at AI-assisted transformer PRs that do this.

### DIFF
--- a/_tools/customlint/testdata/parentaccess/internal/transformers/tstransforms/parentaccess.go.golden
+++ b/_tools/customlint/testdata/parentaccess/internal/transformers/tstransforms/parentaccess.go.golden
@@ -13,12 +13,12 @@
 	func badCall(n *Node) {
 		_ = n.Parent()
 		      ~~~~~~
-!!! forbidparentaccess: avoid .Parent access in internal/transformers (except internal/transformers/declarations)
+!!! forbidparentaccess: Transformers are run on nodes with potentially unset .Parent pointers, do not use them in transformers. Track context in the transform itself, instead.
 	}
 	
 	func badFieldAccess(w *WithField) {
 		_ = w.Parent
 		      ~~~~~~
-!!! forbidparentaccess: avoid .Parent access in internal/transformers (except internal/transformers/declarations)
+!!! forbidparentaccess: Transformers are run on nodes with potentially unset .Parent pointers, do not use them in transformers. Track context in the transform itself, instead.
 	}
 	


_tools/customlint/testdata/parentaccess/go.mod
jakebailey, line 1: I don't think this file is actually required; the parent dir already has this same file

internal/transformers/estransforms/classfields.go
```diff
@@ -2755,7 +2775,7 @@ func (tx *classFieldsTransformer) transformPropertyWorker(property *ast.Property
 
 	initializer := tx.Visitor().VisitNode(property.Initializer)
 	propertyOriginalNode := tx.EmitContext().MostOriginal(property.AsNode())
-	if ast.IsParameterPropertyDeclaration(propertyOriginalNode, propertyOriginalNode.Parent) && ast.IsIdentifier(propertyName) {
+	if ast.IsParameterPropertyDeclaration(propertyOriginalNode, propertyOriginalNode.Parent) && ast.IsIdentifier(propertyName) { //nolint:customlint // MostOriginal returns parse-tree nodes, and this parent relationship is intentional.
```
jakebailey, line 2778: There's probably a way to use `ssa` to flow back to the MostOriginal call, but we haven't yet done anything like that yet so this seems fine to annotate.

internal/transformers/jsxtransforms/jsx.go
```diff
@@ -634,7 +634,7 @@ func (tx *JSXTransformer) createReactNamespace(reactNamespace string, parent *as
 
 	// Set the parent that is in parse tree
 	// this makes sure that parent chain is intact for checker to traverse complete scope tree
-	react.Parent = tx.EmitContext().ParseNode(parent)
+	react.Parent = tx.EmitContext().ParseNode(parent) //nolint:customlint // Parent is intentionally wired to a parse-tree node for resolver traversal.
```
jakebailey, line 637: Do we actually want to lint writes?

weswigham: Writes are still suspect - less suspect, but still suspect. Tbh, everywhere where we're doing writes today is somewhere we're using the `.Parent` hierarchy to encode data we really aughta have in another sidechannel to support constructed nodes better.

internal/transformers/estransforms/classfields.go
```diff
@@ -260,6 +274,9 @@ func (tx *classFieldsTransformer) visitForSubstitution(node *ast.Node) *ast.Node
 
 // visit is the main visitor.
 func (tx *classFieldsTransformer) visit(node *ast.Node) *ast.Node {
+	grandparentNode := tx.pushNode(node)
+	defer tx.popNode(grandparentNode)
```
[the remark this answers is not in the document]
jakebailey: the compiler is smarter than that

internal/transformers/jsxtransforms/jsx.go
jakebailey: That's fair, yeah.

internal/transformers/estransforms/classfields.go
weswigham: I find how `//nolint:customlint` suppresses all our custom lints, rather than having per-analyzer suppressions, is a little icky, but yeah, it's simple like this. If I wanted to go farther, first I'd annotate util functions with internal `.Parent` accesses (like `ast.FindAncestor`) and forbid those, too - that's well into the realm of a real, total flow analysis, though, rather than a fast syntactic ban.

jakebailey: I'm not thrilled by that either; https://github.com/golangci/golangci-lint/issues/1165#issuecomment-2609835085 would be nice

PR #3229, opened 2026-03-25
Check pointer equality of getRegularTypeOfLiteralType in pseudoTypeEquivalentToType

### LINKED ISSUE #3207: `isolatedDeclarations` false positive TS9013 on arrow functions with default parameters and `as const` objects

jakebailey:
This + #3219 fix #3207

### DIFF
--- a/internal/checker/pseudotypenodebuilder.go
+++ b/internal/checker/pseudotypenodebuilder.go
@@ -295,9 +295,12 @@ func (b *NodeBuilderImpl) pseudoTypeEquivalentToType(t *pseudochecker.PseudoType
 				}
 			}
 		}
+		// handles freshness mismatches (e.g., fresh true vs regular true in as const)
+		if b.ch.getRegularTypeOfLiteralType(typeFromPseudo) == b.ch.getRegularTypeOfLiteralType(type_) {
+			return true
+		}
 		if typeFromPseudo.flags&TypeFlagsUnion != 0 && type_.flags&TypeFlagsUnion != 0 {
-			// handles freshness and `undefined` variant mismatches among union members, plus union comparison in general, since the unions may not be `==`
-			// identical due to aliasing and the like
+			// handles union comparison in general, since unions may not be `==` identical due to aliasing
 			if b.ch.compareTypesIdentical(typeFromPseudo, type_) == TernaryTrue {
 				return true
 			}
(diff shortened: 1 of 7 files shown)


internal/checker/pseudotypenodebuilder.go
```diff
@@ -295,9 +295,12 @@ func (b *NodeBuilderImpl) pseudoTypeEquivalentToType(t *pseudochecker.PseudoType
 				}
 			}
 		}
+		// handles freshness mismatches (e.g., fresh true vs regular true in as const)
+		if b.ch.getRegularTypeOfLiteralType(typeFromPseudo) == b.ch.getRegularTypeOfLiteralType(type_) {
```
weswigham, line 299: 👍 We could also start using the identity comparison operation instead of `==` for non-union types, since that's in use for the same reason as this. Since we don't manufacture object types out of pseudotypes, the comparison will never actually be expensive, really.

PR #3230, opened 2026-03-25
Fix class field super access checks

ahejlsberg:
This PR fixes an issue that was revealed by test baseline differences. The affected test still fails in other ways, but that is expected. (We no longer permit expando declarations without initializers, and the test contains a few of those.)


internal/ast/utilities.go
ahejlsberg, line 4470: Property and element access expressions are no longer declaration sites in Corsa.

PR #3232, opened 2026-03-25
Speed up lookupSymbolChain

jakebailey:
After #3213 I was curious if `lookupSymbolChain` could be made faster.

Do `go test -cpuprofile=cpu.out -memprofile=mem.out ./internal/testrunner` and look:

image

A big chunk from this. This isn't new, the same thing happens before #3213.

Threw copilot and pprof at it, and it came up with this change:

- `trySymbolTable` now iterates a cached list of only alias-flagged symbols instead of the full symbol table. The globals table has ~5000 entries but only ~100 aliases, eliminating ~98% of per-symbol flag checks.
- Alias lists are cached per symbol table ID for globals, raw exports, and resolved exports tables. Resolved exports (from `getExportsOfSymbol`/`getExportsOfModule`) use a distinct `stKindResolvedExports` ID to avoid cache collisions with raw `sym.Exports` tables passed by `someSymbolTableInScope`.
- Members tables are skipped entirely since `someSymbolTableInScope` already filters them to `SymbolFlagsType & ^SymbolFlagsAssignment`, which never includes aliases.
- The `symbolFromSymbolTable.ExportSymbol` check from the original full-table loop is preserved via a direct name lookup, since it's not gated by the Alias flag and would be missed by alias-only iteration.

| Metric | `main` | After |
|---|---|---|
| **CPU time** | 19.83s | 7.66s (**-61%**) |
| **alloc_objects** | 6,555,988 | 6,892,415 (+5%) |
| **alloc_space** | 1,669 MB | 1,741 MB (+4%) |

image

`lookupSymbolChain` is much reduced.


internal/checker/symbolaccessibility.go
```diff
@@ -478,31 +478,79 @@ func (c *Checker) getAccessibleSymbolChainFromSymbolTable(ctx accessibleSymbolCh
...
+// to SymbolFlagsType & ^SymbolFlagsAssignment, which never includes aliases.
+// Locals tables are small and per-scope, so they are filtered but not cached.
+func (c *Checker) getSymbolTableAliases(symbols ast.SymbolTable, tableId symbolTableID) []*ast.Symbol {
+	kind := tableId & stKindMask
+	// Members tables never contain alias symbols; skip entirely.
+	if kind == stKindMembers {
+		return nil
+	}
+	// Globals and exports tables are large and revisited often; use cache.
+	if kind == stKindGlobals || kind == stKindExports {
```
jakebailey: The initial fix copilot made wanted to cache _everything_; I modified it to only cache for globals/exports, which I think are much more bounded. That didn't negatively affect perf that much, so I think it's fine. But maybe there are not that many things we could cache such that all could be cached?

internal/checker/symbolaccessibility.go
```diff
@@ -478,31 +478,79 @@ func (c *Checker) getAccessibleSymbolChainFromSymbolTable(ctx accessibleSymbolCh
...
+				return aliases
+			}
+		}
+	}
+	var aliases []*ast.Symbol
+	for _, sym := range symbols {
+		if sym.Flags&ast.SymbolFlagsAlias != 0 {
+			aliases = append(aliases, sym)
+		}
+	}
```
jakebailey, line 525: I tested a version that used `iter.Seq` rather than allocate slices, but it was slower for this code, probably because none of this can really be inlined properly.

jakebailey: That and, presizing this didn't help.

internal/checker/symbolaccessibility.go
weswigham: Yeah, I had a change for strada from years ago that added a ton of caching to symbol chain lookups, and in strada all it did was drop a bomb on memory usage and didn't really help perf (anything we gained was eaten by extra GC pressure). Different runtime, different automatic optimizations - I'm unsurprised some manual caching helps here.

internal/checker/symbolaccessibility.go
```diff
@@ -478,31 +478,79 @@ func (c *Checker) getAccessibleSymbolChainFromSymbolTable(ctx accessibleSymbolCh
...
 	}
 
 	var candidateChains [][]*ast.Symbol
-	// collect all possible chains to sort them and return the shortest/best
-	for _, symbolFromSymbolTable := range symbols {
+
+	// Check for ExportSymbol via direct O(1) lookup instead of inside the alias loop.
+	// In the original loop, this checks `symbolFromSymbolTable.Name == ctx.symbol.Name`,
+	// but there's at most one such symbol in the table (keyed by name).
+	if ok && res != nil && res.ExportSymbol != nil {
```
weswigham, line 554: This is kinda misleadingly wrong since we iterate all the possible chains for result stability nowadays anyway, isn't it? It's not like we skip the full iteration when we find the fast-path candidate. All this does is move those `nil` checks and a name comparison out of the loop...

internal/checker/symbolaccessibility.go
```diff
@@ -533,6 +576,22 @@ func (c *Checker) trySymbolTable(
...
+		// Fast path: since globalThisSymbol.Exports shares the globals table, check if the
+		// symbol is directly accessible by name with ignoreQualification=true before the
+		// expensive recursive search through getCandidateListForSymbol. The recursive search
+		// would clone the globals table, iterate all symbols looking for aliases, etc. In the
+		// common case, the symbol IS in the globals table by name but failed the initial check
+		// above because canQualifySymbol failed. With ignoreQualification=true, we can find it
+		// directly and build the globalThis.X chain without iteration.
+		if ok && res != nil {
+			if c.isAccessible(ctx, res, nil, true /*ignoreQualification*/) {
+				if c.canQualifySymbol(ctx, c.globalThisSymbol, getQualifiedLeftMeaning(ctx.meaning)) {
```
weswigham: This ain't right for a "fast path". This is checking if `globalThis` is qualifably visible, but then returning a chain of `globalThis.otherSymbol` - since it's potentially only visible with qualification, you might actually need to write something like `globalAlias.globalThis.otherSymbol` thanks to shadowing, making the result potentially wrong; you need the rest of the symbol chain in the result, too, assuming `globalThis` isn't actually accessible!

internal/checker/symbolaccessibility.go
jakebailey: Yeah, it goes faster if you really do cache everything. May just cut this down to globals.

internal/checker/symbolaccessibility.go
jakebailey: Yeah, I wasn't happy with this, given we do already loop.

internal/checker/symbolaccessibility.go
weswigham: (I also just doubt this fast path is really needed - most global symbols can be named without resorting to `globalThis` since most symbols don't get shadowed)

jakebailey:
Unfortunately it seems like if I remove these fast paths (which presumably are iffy), and then restrict to just globals, it goes back to having the same perf...

internal/checker/symbolaccessibility.go
weswigham: `iter.Seq` causing closure allocations seems to hamstring its' performance, more often than not - I'd really love to use the `(*ast.Node).IterChildren()` helper I added more, since it really seems much more idiomatic, but more often than not using `ForEachChild` or `VisitEachChild` and manually managing closure lifetimes/scopes has ended up being better...

jakebailey:
Drafting this for now, I was overexcited 😄 

It's a shame that we have no tests that can prove that this is wrong, though

internal/checker/symbolaccessibility.go
jakebailey: Ah, no, I misread this diff when reading it back. The thing is that the new loop doesn't go over every symbol in the table anymore, just the ones that are aliases. So it can't go back into the loop.

So we'd have to then go back to looping over everything again, or do something more direct. Which, I think this is doing?

jakebailey:
To be clear, I pushed to this PR with an updated solution which I believe is as fast and should be correct at the cost of another bit out of the key.

weswigham (approved):

PR #3235, opened 2026-03-26
Add project info, progress indicators

jakebailey:
This ports over a bunch of the code from the old TS extension for the project info, and adds a custom message to grab project info for a file.

Also, add progress indicators. We keep track of which projects are currently being loaded, showing whichever one's been running the longest. For example, `vscode`:

image

The progress indicators also fire for ATA, like they used to for `npm install` tracking.

### DIFF
--- a/_extension/src/projectStatus.ts
+++ b/_extension/src/projectStatus.ts
@@ -83,7 +83,7 @@ export class ProjectStatus implements vscode.Disposable {
         try {
             const result = await this.client.getProjectInfo(doc.uri.toString(), pendingState.cancellation.token);
             if (this.state === pendingState) {
-                this.updateState(new ProjectInfoState.Resolved(doc.uri, result.configFileName));
+                this.updateState(new ProjectInfoState.Resolved(doc.uri, result.configFilePath));
             }
         }
         catch {
@@ -122,8 +122,9 @@ export class ProjectStatus implements vscode.Disposable {
                 break;
             }
             case ProjectInfoState.Type.Resolved: {
-                const isTypeScript = this.activeEditorTracker.activeJsTsEditor?.document.languageId === "typescript"
-                    || this.activeEditorTracker.activeJsTsEditor?.document.languageId === "typescriptreact";
+                const currentLanguageId = this.activeEditorTracker.activeJsTsEditor?.document.languageId;
+                const isTypeScript = currentLanguageId === "typescript"
+                    || currentLanguageId === "typescriptreact";
                 const noConfigFileText = isTypeScript ? "No tsconfig" : "No jsconfig";
 
                 const rootPath = this.getWorkspaceRootForResource(this.state.resource);
--- a/_extension/src/util.ts
+++ b/_extension/src/util.ts
@@ -27,8 +27,9 @@ export function isSupportedLanguageMode(doc: vscode.TextDocument): boolean {
     return jsTsLanguageModes.includes(doc.languageId);
 }
 
+const configRegex = /^[jt]sconfig\.(.+\.)?json$/i;
 export function isJsConfigOrTsConfigFileName(fileName: string): boolean {
-    return /^[jt]sconfig\.(.+\.)?json$/i.test(path.basename(fileName));
+    return configRegex.test(path.basename(fileName));
 }
 
 export interface ExeInfo {
--- a/internal/diagnostics/extraDiagnosticMessages.json
+++ b/internal/diagnostics/extraDiagnosticMessages.json
@@ -103,7 +103,7 @@
         "category": "Message",
         "code": 100014
     },
-    "Installing typings for '{0}'": {
+    "Installing types for '{0}'": {
         "category": "Message",
         "code": 100013
     }
--- a/internal/lsp/lsproto/_generate/generate.mts
+++ b/internal/lsp/lsproto/_generate/generate.mts
@@ -279,7 +279,7 @@ const customStructures: Structure[] = [
         name: "ProjectInfoResult",
         properties: [
             {
-                name: "configFileName",
+                name: "configFilePath",
                 type: { kind: "base", name: "string" },
                 documentation: "The absolute path to the config file (e.g. /path/to/tsconfig.json) for the project that contains this file, or an empty string if the file is in an inferred project.",
             },
--- a/_extension/src/client.ts
+++ b/_extension/src/client.ts
@@ -292,11 +292,11 @@ export class Client implements vscode.Disposable {
         return result.file;
     }
 
-    async getProjectInfo(uri: string, token?: vscode.CancellationToken): Promise<{ configFileName: string; }> {
+    async getProjectInfo(uri: string, token?: vscode.CancellationToken): Promise<{ configFilePath: string; }> {
         if (!this.client) {
             throw new Error("Language client is not initialized");
         }
-        return this.client.sendRequest<{ configFileName: string; }>("custom/projectInfo", {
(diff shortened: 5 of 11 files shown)


internal/lsp/progress.go
```diff
@@ -0,0 +1,138 @@
...
+	case p.ch <- progressEvent{message: message, args: args}:
+	case <-ctx.Done():
+	}
+}
+
+func (p *projectLoadingProgress) finish(ctx context.Context, message *diagnostics.Message, args ...any) {
+	select {
+	case p.ch <- progressEvent{message: message, args: args, finish: true}:
+	case <-ctx.Done():
+	}
```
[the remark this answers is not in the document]
jakebailey: I'd rather not, as we would just create a situation where we would have unmatched spans, which is weird enough.

_extension/src/activeJsTsEditorTracker.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    public get activeJsTsEditor(): vscode.TextEditor | undefined {
+        return this._activeJsTsEditor;
+    }
+
+    private update(): void {
+        const editorCandidates = this.getEditorCandidatesForActiveTab();
+        const managedEditors = editorCandidates.filter(editor => this.isManagedFile(editor));
+        const newActiveJsTsEditor = managedEditors.at(0);
+        if (this._activeJsTsEditor !== newActiveJsTsEditor) {
```
DanielRosenwasser: I'm guessing this is ported but while we're here.

```suggestion
        const newActiveJsTsEditor = editorCandidates.find(editor => this.isManagedFile(editor));
        if (newActiveJsTsEditor !== undefined && this._activeJsTsEditor !== newActiveJsTsEditor) {
```

_extension/src/activeJsTsEditorTracker.ts
```diff
@@ -0,0 +1,85 @@
...
+        const tab = vscode.window.tabGroups.activeTabGroup.activeTab;
+        if (!tab) {
+            return [];
+        }
+
+        // Basic text editor tab
+        if (tab.input instanceof vscode.TabInputText) {
+            const inputUri = tab.input.uri;
+            const editor = vscode.window.visibleTextEditors.find(editor => {
+                return editor.document.uri.toString() === inputUri.toString()
```
DanielRosenwasser:
```suggestion
            const inputUriString = tab.input.uri.toString();
            const editor = vscode.window.visibleTextEditors.find(editor => {
                return editor.document.uri.toString() === inputUriString
```

_extension/src/projectStatus.ts
```diff
@@ -0,0 +1,188 @@
...
+        if (this.state.type === ProjectInfoState.Type.Pending) {
+            this.state.cancellation.cancel();
+            this.state.cancellation.dispose();
+        }
+        this.disposables.forEach(d => d.dispose());
+    }
+}
+
+function isInferredProjectName(configFileName: string): boolean {
+    return configFileName.startsWith("/dev/null/");
```
DanielRosenwasser: Is this true? Below you've written

> For inferred projects, this is the last component of the current directory.

_extension/src/projectStatus.ts
```diff
@@ -0,0 +1,188 @@
...
+                statusItem.severity = vscode.LanguageStatusSeverity.Information;
+                statusItem.text = "Loading project info...";
+                statusItem.detail = undefined;
+                statusItem.command = undefined;
+                statusItem.busy = true;
+                break;
+            }
+            case ProjectInfoState.Type.Resolved: {
+                const isTypeScript = this.activeEditorTracker.activeJsTsEditor?.document.languageId === "typescript"
+                    || this.activeEditorTracker.activeJsTsEditor?.document.languageId === "typescriptreact";
```
DanielRosenwasser:
```suggestion
                const currentLanguageId = this.activeEditorTracker.activeJsTsEditor?.document.languageId;
                const isTypeScript = currentLanguageId === "typescript"
                    || currentLanguageId === "typescriptreact";
```

_extension/src/util.ts
```diff
@@ -12,6 +12,25 @@ export const jsTsLanguageModes = [
...
+    "chat-editing-text-model",
+]);
+
+export function isSupportedLanguageMode(doc: vscode.TextDocument): boolean {
+    return jsTsLanguageModes.includes(doc.languageId);
+}
+
+export function isJsConfigOrTsConfigFileName(fileName: string): boolean {
+    return /^[jt]sconfig\.(.+\.)?json$/i.test(path.basename(fileName));
+}
```
DanielRosenwasser, line 33:
```suggestion
const configRegex = /^[jt]sconfig\.(.+\.)?json$/i;
export function isJsConfigOrTsConfigFileName(fileName: string): boolean {
    return configRegex.test(path.basename(fileName));
}
```

internal/diagnostics/extraDiagnosticMessages.json
```diff
@@ -94,5 +94,17 @@
...
+    },
+    "Loading": {
+        "category": "Message",
+        "code": 100012
+    },
+    "Project '{0}'": {
+        "category": "Message",
+        "code": 100014
+    },
+    "Installing typings for '{0}'": {
```
DanielRosenwasser: We've been moving away from "typings" for a while so consider

```suggestion
    "Installing types for '{0}'": {
```

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -264,6 +264,28 @@ const customStructures: Structure[] = [
...
+        name: "ProjectInfoResult",
+        properties: [
+            {
+                name: "configFileName",
+                type: { kind: "base", name: "string" },
+                documentation: "The config file name (e.g. tsconfig.json) for the project that contains this file, or an empty string if the file is in an inferred project.",
+            },
+        ],
+        documentation: "Result for the custom/projectInfo request.",
+    },
```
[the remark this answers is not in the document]
DanielRosenwasser: Agreed - or `configFileUri` even.

_extension/src/util.ts
jakebailey: I don't actually think extracting matters when you're not using `exec` but yeah

_extension/src/projectStatus.ts
jakebailey: That was an outdated comment, now gone.

_extension/src/activeJsTsEditorTracker.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    private isManagedFile(editor: vscode.TextEditor): boolean {
+        if (disabledSchemes.has(editor.document.uri.scheme)) {
+            return false;
+        }
+        return isSupportedLanguageMode(editor.document) || isJsConfigOrTsConfigFileName(editor.document.fileName);
+    }
+
+    dispose(): void {
+        this.disposables.forEach(d => d.dispose());
```
DanielRosenwasser, line 82: Not a big deal here, but it would be better if it popped everything in reverse like we do elsewhere.

_extension/src/projectStatus.ts
```diff
@@ -0,0 +1,185 @@
...
+                }
+                break;
+            }
+        }
+    }
+
+    private ensureStatusItem(): vscode.LanguageStatusItem {
+        if (!this.statusItem) {
+            this.statusItem = vscode.languages.createLanguageStatusItem("typescript.native-preview.projectStatus", jsTsLanguageModes);
+            this.statusItem.name = "TypeScript Native Preview Project Status";
```
DanielRosenwasser, line 167: We're probably gonna have to start thinking about localization eventually.

PR #3239, opened 2026-03-26
Accept baseline differences

ahejlsberg:
First batch of accepted baseline differences.


testdata/submoduleAccepted.txt
```diff
@@ -229,3 +229,105 @@ conformance/jsxJsxsCjsTransformNestedSelfClosingChild(jsx=react-jsxdev).js.diff
...
+
+## different error order ##
+compiler/arrayIterationLibES5TargetDifferent(nolib=true,target=es2015).errors.txt.diff
+compiler/arrayIterationLibES5TargetDifferent(nolib=true,target=esnext).errors.txt.diff
+
+## emit differences ##
+compiler/assignmentToInstantiationExpression.js.diff
+compiler/asyncArrowInClassES5(target=es2015).js.diff
+compiler/augmentedTypesVar.js.diff
+compiler/commentsFunction(target=es2015).js.diff
```
jakebailey: I don't think this one is a desirable diff.

testdata/submoduleAccepted.txt
```diff
@@ -229,3 +229,105 @@ conformance/jsxJsxsCjsTransformNestedSelfClosingChild(jsx=react-jsxdev).js.diff
...
+compiler/computedPropertyBindingElementDeclarationNoCrash1.errors.txt.diff
+compiler/conditionalTypeDoesntSpinForever.errors.txt.diff
+compiler/constraints0.errors.txt.diff
+compiler/constructorOverloads1.errors.txt.diff
+
+## different error order ##
+compiler/arrayIterationLibES5TargetDifferent(nolib=true,target=es2015).errors.txt.diff
+compiler/arrayIterationLibES5TargetDifferent(nolib=true,target=esnext).errors.txt.diff
+
+## emit differences ##
```
jakebailey: Should we explain more why these are ignored? in general? The existing headers were terse but only because there wasn't much to say.

ahejlsberg: We could call it "emit differences of no consequence". At least, that was the intent.

testdata/submoduleAccepted.txt
ahejlsberg: Ok, I'll remove this one.

Issue #3238, opened 2026-03-26
Consider using `process.execve` in tsgo.js ?


kanashimia:
Current tsgo wrapper uses execFileSync from node:child_process which keeps extra node process for no reason.
Node has experimental process.execve which solves that problem.
Would be used like this `process.execve(exe, process.argv.slice(1))`
Docs: https://nodejs.org/docs/latest/api/process.html#processexecvefile-args-env
Although it is not available on Windows at least as of right now.
Ofc ideally there would be no node wrapper at all.
It is still experimental, but what do you think about optionally using it if it is available?

jakebailey:
Yes, we'd totally accept a PR for this.

PR #3241, opened 2026-03-26
Use process.execve where it is supported

kanashimia:
Current tsgo wrapper uses execFileSync from node:child_process which keeps extra node process for no reason.
Node has experimental process.execve which solves that problem.
Docs: https://nodejs.org/docs/latest/api/process.html#processexecvefile-args-env
Note that it is still experimental and is not available on Windows at least as of right now.

This patch makes the tsgo.js wrapper use it if it is available, startup time of tsgo is still slightly slower because of the node wrapper, but at least there is no persistent node process.

Closes https://github.com/microsoft/typescript-go/issues/3238

Unlike `execFileSync` this `process.execve` takes args including argv0,
so `[exe, ...process.argv.slice(2)]` is just to set argv0 to `tsgo` instead of `tsgo.js`, it isn't really needed, it could be `process.argv.slice(1)` otherwise.

I tested that it works against node 20 and node 24 on Linux, needs testing on Windows and macOS, I can't test there.

btw on github because of whitespace diff makes it harder to see what changed here, I suggest hiding whitespace

### DIFF
--- a/_packages/native-preview/bin/tsgo.js
+++ b/_packages/native-preview/bin/tsgo.js
@@ -5,6 +5,16 @@ import { execFileSync } from "node:child_process";
 
 const exe = getExePath();
 
+if (typeof process.execve === "function") {
+    // > v22.15.0
+    try {
+        process.execve(exe, [exe, ...process.argv.slice(2)]);
+    }
+    catch {
+        // not available on windows, ignore the error and fallback
+    }
+}
+
 try {
     execFileSync(exe, process.argv.slice(2), { stdio: "inherit" });
 }


kanashimia:

_packages/native-preview/bin/tsgo.js
```diff
@@ -5,14 +5,19 @@ import { execFileSync } from "node:child_process";
...
 
-try {
-    execFileSync(exe, process.argv.slice(2), { stdio: "inherit" });
-}
-catch (e) {
-    if (e.status) {
-        process.exitCode = e.status;
+if (typeof process.execve === "function") {
+    // > v22.15.0 & not Windows
+    process.execve(exe, [exe, ...process.argv.slice(2)]);
```
jakebailey: Unfortunately this isn't actually enough; on Windows:

```
Welcome to Node.js v25.8.1.
Type ".help" for more information.
> process.execve
[Function: execve]
> process.execve("asd", ["asd"])
Uncaught:
TypeError [ERR_FEATURE_UNAVAILABLE_ON_PLATFORM]: The feature process.execve is unavailable on the current platform, which is being used to run Node.js
    at process.execve (node:internal/process/per_thread:290:13) {
  code: 'ERR_FEATURE_UNAVAILABLE_ON_PLATFORM'
}
> (node:50048) ExperimentalWarning: process.execve is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
```

Even though the docs say it's  "not defined", it actually is. So we need to do some sort of platform check, or catch just catch the error to look for this code.

kanashimia: Thanks for testing!!! Weird that it is still defined there... Just catching is probably a better solution.
I was able to reproduce behaviour by adding `Object.defineProperty(process, 'platform', { value: 'win32' });`
They throw that error here: https://github.com/nodejs/node/blob/38647b388bbf5d350f8febe77a6674ec0d0607c5/lib/internal/process/per_thread.js#L290
Let me do a quick fix

jakebailey: I think a plain try/catch should be fine for this, given if it succeeds, no more code is run.

So you could probably just `try catch` at the top level and leave the existing code where it is.

kanashimia: Yeah, the reason I wrapped previous code in else was for readability only.

_packages/native-preview/bin/tsgo.js
```diff
@@ -5,6 +5,19 @@ import { execFileSync } from "node:child_process";
...
+if (typeof process.execve === "function") {
+    // > v22.15.0
+    try {
+        process.execve(exe, [exe, ...process.argv.slice(2)]);
+    }
+    catch (e) {
+        // not available on windows, ignore the error and fallback
+        if (e?.code !== "ERR_FEATURE_UNAVAILABLE_ON_PLATFORM") {
+            throw e;
+        }
```
jakebailey: Is it worth rethrowing? They don't document this error so I'm sort of afraid to depend on it.

jakebailey: But, maybe it has some nice error about not being able to find the exe. (Though we check that earlier...)

kanashimia: Looking at their code it doesn't matter here anyways, not worth it I figure.

kanashimia: > But, maybe it has some nice error about not being able to find the exe. (Though we check that earlier...)

@jakebailey it just exits the process instead of throwing an error


```
node:internal/process/per_thread:326
    _execve(execPath, args, envArray);
    ^

process.execve failed with error code ENOENT

----- Native stack trace -----

 1: 0x55b81939f676  [node]
 2: 0x7f2b4f58f08d 

----- JavaScript stack trace -----

1: execve (node:internal/process/per_thread:326:5)
2: file://./bin/tsgo.js:11:17
3: run (node:internal/modules/esm/module_job:377:25)

```

The error isn't nice, but it shouldn't occur realistically.

PR #3246, opened 2026-03-26
Do not reuse input expression nodes to generate type nodes for contextually typed const array literal expressions

### LINKED ISSUE #3192: 0320 vs 0321
## Steps to reproduce

[](https://github.com/workglow-dev/workglow) 

clone, do a `bun i` and then `bun run build`
see https://github.com/workglow-dev/workglow/actions/runs/23388170259/job/68038462188

tsc and tsgo as of yesterday had no issues. This only started today with Saturday March 21 build. tsgo =7.0.0-dev.20260321.1 

when pinned to 7.0.0-dev.20260320.1 

https://github.com/workglow-dev/workglow/actions/runs/23388410799/job/68039071095 (no issues)

weswigham:
Since their `readonly` state requires checking to know.

Fixes #3192

Had to tune up the pseudochecker's `isContextuallyTyped` a bit (note: this was a port from strada, strada's `expressionToTypeNode`'s `isContextuallyTyped` is similarly wrong as the old version) to get this to work.

### DIFF
--- a/testdata/baselines/reference/submodule/compiler/declarationEmitPropertyNumericStringKey.js.diff
+++ b/testdata/baselines/reference/submodule/compiler/declarationEmitPropertyNumericStringKey.js.diff
@@ -1,0 +1,11 @@
+--- old.declarationEmitPropertyNumericStringKey.js
++++ new.declarationEmitPropertyNumericStringKey.js
+@@= skipped -27, +27 lines =@@
+
+ //// [declarationEmitPropertyNumericStringKey.d.ts]
+ declare const STATUS: {
+-    readonly "404": "not found";
++    readonly ["404"]: "not found";
+ };
+ declare const hundredStr = "100";
+ declare const obj: {
--- a/testdata/baselines/reference/submodule/conformance/constAssertions.js.diff
+++ b/testdata/baselines/reference/submodule/conformance/constAssertions.js.diff
@@ -1,0 +1,53 @@
+--- old.constAssertions.js
++++ new.constAssertions.js
+@@= skipped -213, +213 lines =@@
+
+
+ //// [constAssertions.d.ts]
+-declare let v1: "abc";
+-declare let v2: "abc";
++declare let v1: 'abc';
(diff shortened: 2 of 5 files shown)


testdata/tests/cases/compiler/declarationEmitAsConstSatisfiesNonReadonlyResult.ts
```diff
@@ -0,0 +1,16 @@
+// @declaration: true
```
jakebailey, line 1: Do we want `// @isolatedDeclarations: true, false` here so we can see an error appear?

weswigham: It'll be because of the `satisfies` itself (there's no ID inference rule for `satisfies`) and not the array literal within it (meaning: this PR doesn't change anything) - is that useful?

internal/pseudochecker/lookup.go
```diff
@@ -444,6 +444,9 @@ func (ch *PseudoChecker) typeFromArrayLiteral(node *ast.ArrayLiteralExpression)
 	if !ch.canGetTypeFromArrayLiteral(node) {
 		return NewPseudoTypeInferred(node.AsNode())
 	}
+	if IsInConstContext(node.AsNode()) && isContextuallyTyped(node.AsNode()) {
```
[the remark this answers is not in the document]
weswigham: No, I think not - we're gonna be explicit about the logic here, because this is some obtuse contortions to respect a checker edgecase behavior.

testdata/tests/cases/compiler/declarationEmitAsConstSatisfiesNonReadonlyResult.ts
jakebailey: Oh, perhaps not. I don't know if we already have errors for `satisfies` in existing baselines but if so, we can skip it.

PR #3248, opened 2026-03-26
fix(3242): check class name collisions with Object

### LINKED ISSUE #3242: No error for values named 'Object' (in CommonJS)
```ts
// @module: commonjs, esnext
class Object {}

export let x = { ...{} };
```

Expected: Error at least in CJS (TS6 behavior), or error in both.

Actual: no error message.

a-tarasyuk:
Fixes #3242

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -6718,6 +6718,12 @@ func (c *Checker) checkTypeNameIsReserved(name *ast.Node, message *diagnostics.M
 	}
 }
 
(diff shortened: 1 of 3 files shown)


jakebailey:
I think we actually need something that goes further, like #2087, since there are more checks like this. Though this particular filed issue is fun because it reveals that we were somehow only checking this in non-ESM files.

jakebailey:
To be clear, that other PR could be overcomplicated, so I'm not saying to close this, rather that there's a wider range of missing code we seem to need.

PR #3257, opened 2026-03-27
Add perf / project telemetry

jakebailey:
This adds events to:

- Project info telemetry, like Strada, with file counts, compiler options (sanitized of course), etc. Sent when we first discover a project, same as Strada.
- Periodically (every 5 mins) send performance info like memory usage, total memory, GC count, allocs, inuse space etc. Plus, the auto import stats. This is new to Corsa.

I added an initialization option that toggles telemetry wholesale; the client is already supposed to ignore it if it doesn't care, but best to not enable it when not needed. Of course, as per usual, telemetry is managed by VS Code and can be controlled like any other extension (and it's all just stats anyway). So the toggle is just for efficiency / testing reasons.

The goal of is is so we can gauge how much we are impacting memory usage. I want to start tweaking `GOGC` and `GOMEMLIMIT` but would like to be able to see that it's doing anything.

This introduces a new dep in order to get cross-platform memory info in pure Go.


internal/lsp/server.go
```diff
@@ -262,6 +263,14 @@ func (s *Server) PublishDiagnostics(ctx context.Context, params *lsproto.Publish
 	return sendNotification(s, lsproto.TextDocumentPublishDiagnosticsInfo, params)
 }
 
+// SendTelemetry implements project.Client.
+func (s *Server) SendTelemetry(ctx context.Context, telemetry lsproto.TelemetryEvent) error {
+	if !s.telemetryEnabled {
+		panic("SendTelemetry called with telemetry disabled")
```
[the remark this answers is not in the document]
jakebailey: That's the point. Callers should check, so they don't do work that they don't need to do, or should not do in tests.

_extension/src/client.ts
```diff
@@ -65,6 +65,7 @@ export class Client implements vscode.Disposable {
             traceOutputChannel: this.traceOutputChannel,
             initializationOptions: {
                 codeLensShowLocationsCommandName,
+                enableTelemetry: true,
```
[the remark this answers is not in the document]
jakebailey: So, yes, but it doesn't matter because even if we send data over the wire, it won't be sent (and the editor's telemetry view makes that clear).

We'd have to also set up a listener for when that setting changes. That's not too hard, but it is I think, pointless.

internal/project/session.go
```diff
@@ -441,6 +453,322 @@ func (s *Session) cancelIdleCacheClean() {
...
+		}
+		return 0
+	}
+
+	measurements.MemoryUsedBytes = readUint64(samples[sMemoryUsedBytes])
+	if samples[sGoMemLimit].Value.Kind() == gometrics.KindUint64 {
+		v := samples[sGoMemLimit].Value.Uint64()
+		if v < uint64(math.MaxInt64) {
+			measurements.GoMemLimit = float64(v)
+		}
```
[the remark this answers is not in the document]
jakebailey: We should set it to some value, even 0; mainly I just want to know when it's been configured. Sending it raw is not a good idea as the default of MaxInt64 is of course bigger than MAX_SAFE_INTEGER.

internal/project/session.go
```diff
@@ -441,6 +453,327 @@ func (s *Session) cancelIdleCacheClean() {
 	}
 }
 
+const performanceTelemetryInterval = 5 * time.Minute
```
DanielRosenwasser, line 461: Maybe not that important now, but should we report events only when we're "idle"?

internal/project/session.go
```diff
@@ -441,6 +453,327 @@ func (s *Session) cancelIdleCacheClean() {
...
+	props := map[string]string{
+		"configFileName": configFileName,
+		"projectType":    projectType,
+		"version":        core.Version(),
+	}
+
+	// Compiler options — same approach as Strada's convertCompilerOptionsForTelemetry:
+	// booleans and enum string names, no paths.
+	compilerOptions := map[string]any{}
+	setTristate(compilerOptions, "strict", opts.Strict)
```
DanielRosenwasser, line 684: I think we still want to capture individual strictness options, especially with the 6/7 `--strict`-by-default transition.

internal/project/session.go
```diff
@@ -441,6 +453,327 @@ func (s *Session) cancelIdleCacheClean() {
...
+func (s *Session) collectProjectInfoTelemetry(project *Project) lsproto.TelemetryEvent {
+	opts := project.CommandLine.CompilerOptions()
+	if opts == nil {
+		opts = &core.CompilerOptions{}
+	}
+
+	configFileName := "other"
+	if project.Kind == KindConfigured {
+		baseName := tspath.GetBaseFileName(project.configFileName)
+		if baseName == "tsconfig.json" || baseName == "jsconfig.json" {
```
DanielRosenwasser, line 665: Does this do the right thing for a solution-style project? Do we need to update this for `strings.HasPrefix`/`HasSuffix`?

internal/project/session.go
jakebailey: We could. Not sure how to determine if we're "idle", however. I'll look into it.

internal/project/session.go
jakebailey: Surprisingly, Strada also had this problem. I could change this if you want.

jakebailey: For reference:

```ts
export function getBaseConfigFileName(configFilePath: NormalizedPath): "tsconfig.json" | "jsconfig.json" | undefined {
    const base = getBaseFileName(configFilePath);
    return base === "tsconfig.json" || base === "jsconfig.json" ? base : undefined;
}
```

internal/project/session.go
```diff
@@ -446,6 +458,338 @@ func (s *Session) cancelIdleCacheClean() {
...
+	if registry := snapshot.AutoImportRegistry(); registry != nil {
+		autoImportStats := registry.GetCacheStats()
+		measurements.AutoImportProjectBucketCount = float64(len(autoImportStats.ProjectBuckets))
+		measurements.AutoImportNodeModulesBucketCount = float64(len(autoImportStats.NodeModulesBuckets))
+		measurements.AutoImportUniquePackageCount = float64(autoImportStats.UniquePackageCount)
+		for _, b := range autoImportStats.ProjectBuckets {
+			measurements.AutoImportProjectExportCount += float64(b.ExportCount)
+			measurements.AutoImportProjectFileCount += float64(b.FileCount)
+		}
+		for _, b := range autoImportStats.NodeModulesBuckets {
```
andrewbranch, line 599: Maybe the one thing to add is an indication of whether any node_modules directory was scraped with no package.json filter; this is reported in the cache stats logging later in the file like this:

```go
				if bucket.DependencyNames != nil {
					s.logger.Logf("\t\t\tCollected packages: %d", bucket.DependencyNames.Len())
				} else {
					s.logger.Logf("\t\t\tCollected packages: all, due to no package.json!")
				}
```

Otherwise, this all looks good to me.

jakebailey: Added `AutoImportNodeModulesUnfilteredBucketCount`.

RyanCavanaugh (approved):

PR #3258, opened 2026-03-27
Fix file dedupe fourslash flake

jakebailey:
After probably 100k runs, I pinned this one down to two things:

- `UpdateProgram` was not fully checking if a file was a part of a cycle. Has to check both maps (Strada did too).
- When a `package.json` file is modified, we can't reuse things anymore. It's dubious, but fourslash does actually edit JSON files, and we see them. Maybe this isn't real, but it does matter for the test at least.

### DIFF
--- a/internal/project/projectcollectionbuilder.go
+++ b/internal/project/projectcollectionbuilder.go
@@ -1089,6 +1089,13 @@ func (b *ProjectCollectionBuilder) markFilesChanged(entry dirty.Value[*Project],
 						dirtyFilePath = ""
 						break
 					}
+					// package.json changes can affect module resolution and package
+					// identity (e.g. dedup decisions), so they must always trigger
+					// a full rebuild rather than a single-file clone.
+					if tspath.GetBaseFileName(string(path)) == "package.json" {
+						dirtyFilePath = ""
+						break
+					}
 					if dirtyFilePath == "" {
 						dirtyFilePath = path
 					} else if dirtyFilePath != path {
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -204,24 +204,9 @@ func NewFourslash(t *testing.T, capabilities *lsproto.ClientCapabilities, conten
 		f.stateBaseline = newStateBaseline(fsFromMap.(iovfs.FsWithSys))
 	} else {
 		for _, file := range testData.Files {
-			if tspath.FileExtensionIs(file.fileName, tspath.ExtensionJson) {
-				// JSON files (tsconfig.json, package.json, etc.) should not be
-				// opened as documents. The client's document selector only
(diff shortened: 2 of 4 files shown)


internal/project/projectcollectionbuilder.go
```diff
@@ -1089,6 +1089,13 @@ func (b *ProjectCollectionBuilder) markFilesChanged(entry dirty.Value[*Project],
 						dirtyFilePath = ""
 						break
 					}
+					// package.json changes can affect module resolution and package
+					// identity (e.g. dedup decisions), so they must always trigger
+					// a full rebuild rather than a single-file clone.
+					if tspath.GetBaseFileName(string(path)) == "package.json" {
```
andrewbranch: This doesn’t seem right to be within `if p.containsFile(path)`; that's checking for inclusion as a source file. package.json files will typically not be in the program as source files; only if they’re imported will they get a source file. This case should already be handled below by `changeType != lsproto.FileChangeTypeCreated && p.host.sourceFS.SeenFile(path)`; the only issue I see is that if the package.json _did_ happen to be included as a source file as well, then we’re incorrectly skipping that lookup because it’s in an `else if`. If this really makes a difference for the test, it could be that fourslash is doing something weird with forcing package.json to be a root file when it shouldn't.

TLDR suggestion: just replace `else if` with `if` on line 1105

jakebailey: I could just restructure the test, though I'm not sure how to modify a `package.json` "on disk" within that infra.

jakebailey: I tried your suggestion, but it doesn't work, as it really is a source file. I think what you're describing is happening with this one test.

andrewbranch: Hm, I guess this approach is fine. I was suggesting that it fall through to the `p.host.sourceFS.SeenFile(path)` check, but AFAIK it's _only_ package.json source files that would need such a fallthrough, so, this is fine. Updating the comment to explain that this is an edge case and _normally_ package.json changes would be handled by the next `else if` block might help future readers.

internal/project/projectcollectionbuilder.go
```diff
@@ -1089,6 +1089,13 @@ func (b *ProjectCollectionBuilder) markFilesChanged(entry dirty.Value[*Project],
...
+						dirtyFilePath = ""
+						break
```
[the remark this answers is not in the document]
andrewbranch: This is not quite right; if we're on a case-insensitive file system, that's when `path` is _guaranteed_ to be lowercase because it's a normalized `tspath.Path`. Only on a _case-sensitive_ file system would `path` _preserve_ the original casing, and I think a person on a _case-sensitive_ file system would be having a bad time with a non-standard casing of `package.json` anyway.

PR #3263, opened 2026-03-27
Preserve original comments on params in dts generated by pseudo type node builder under `removeComments: false`

Andarist:


internal/pseudochecker/lookup.go
```diff
@@ -659,11 +659,13 @@ func (ch *PseudoChecker) cloneParameters(nodes *ast.NodeList) []*PseudoParameter
...
 	for _, e := range nodes.Nodes {
+		parameter := e.AsParameterDeclaration()
 		result = append(result, NewPseudoParameter(
-			e.AsParameterDeclaration().DotDotDotToken != nil,
+			parameter,
+			parameter.DotDotDotToken != nil,
 			e.Name(),
-			e.AsParameterDeclaration().QuestionToken != nil || e.AsParameterDeclaration().Initializer != nil,
-			ch.typeFromParameter(e.AsParameterDeclaration()),
+			parameter.QuestionToken != nil || parameter.Initializer != nil,
```
jakebailey: At some level, why even have any of these if we are ferrying along the declaration?

(Or, why not just use `p.Name.Parent` more above?)

Andarist: > At some level, why even have any of these if we are ferrying along the declaration?

ye, u are right here - since the new addition most of those other fields can just be read directly from the `Declaration`. I made that change.

> Or, why not just use p.Name.Parent more above?

I thought about that but that felt indirect and not matching the style of all of those other fields on pseudochecker structs

jakebailey:
At some level, I wonder if this is even worth it.

Andarist:
> At some level, I wonder if this is even worth it.

Fair. However, those are kinda jsdoc-like and `shouldWriteComment` explicitly allows those:
https://github.com/microsoft/TypeScript/blob/c7a0ae102dfb030c534a7bc31fbbd61a707dc9cf/src/compiler/emitter.ts#L6012-L6017

And the dts emitter configures the printer with `onlyPrintJsDocStyle: true`

jakebailey:
I think this PR is fine, but, does this also cause non-JSDoc comments to be emitted? Hopefully not?

iisaduan:
@jakebailey What non-JSDoc comments are you concerned about? 

I tested a bit more, and this PR looks fine since the behavior seems to match strada's emit exactly, for example: 
```ts
/** leading comment*/ var lambddaNoVarComment = /** this is lambda multiplication*/ (/* not jsdoc */ /**param a*/ /* nonjsdoc stuff */a: number /**jsdoc */, /**param b*/b: number) => a * b;   
```
becomes 
```ts
/** leading comment*/ declare var lambddaNoVarComment: (/* not jsdoc */ /**param a*/ /* nonjsdoc stuff */ a: number /**jsdoc */, /**param b*/ b: number) => number;
```

jakebailey:
It was the non-jsdoc stuff I was concerned about, but if it matches Strada then LGTM

jakebailey (approved):

iisaduan (approved):

PR #3264, opened 2026-03-27
support quick info and go to definition on mapped keys

### LINKED ISSUE #50715: Mapped keys do not preserve JSDoc documentation
# Bug Report

### 🔎 Search Terms

mapping types with docs; preserve docs on mapped types

### 🕗 Version & Regression Information

- This is the behavior in every version I tried, and I reviewed the FAQ for entries about this

### ⏯ Playground Link

[Playground link with relevant code](https://www.typescriptlang.org/play?#code/C4TwDgpgBA4hzAgJwM4B4Aq4ID4oF4oBvAKCigG0AFJAe0iVCgEsA7KAawhFoDMoskKAEMUUAAYBzeABIiAYWFhmwYQBtmALwhoUwJG0lQAZFBr1koHAF9xAXQBcUABQBKAnkERqdBqDsk1gDcJCRsiEi8wgDG0FTIKLTspOQA9KlQgKDkUADqABYQ7Hm0AG7IhlAFSBAANFDABVAAJrTRAK4AtoWqwMxJLGJNzChgasIgEE11oizAUCjFbWpNAHRkUKkAVJvrUJtmCUkA5GKswl1r5HtQAAIQAB7no9AARABStHnsACK0EC+7TapdZnLpOPQGViSELrdJZXIFIqlcpQyrIWr1RotdpdVg9PrsYbNYajcaTaZiFTzRbLS4bbaAg6oY5iACC0jp1zujw6zygAGYAAyA4HkYTSJysToAI2QMLSGWy+UKlWRkKMVQxDWg2M63WEvX6RKGIzGEymIkpcwWtCWq1hDKu+3izNYJygABlWgaCZz9tynmpXgAhJDCTTMNQAp2iqBqb2G1jg-SGELWUKgIQe8MgF2JdiEOAIBJoPNJHAw6JJPRQMHZzS5w4F4iwjKAGXIESrimV1WjqnVtc1Wnq8T6jWIAHIAeQwxNNZPt5GkwAn5wgTjcHig72EXCgikYyBeNVbUA7yqRPYqmoHWOHuPx46g09nJtJ5rpy-Z65c7nweH5ABWE8FTPTtLxRDV0VvHV731RMBmfGc53fSZP3gL1ojHJNfy3F4AFVWBUSYoAAZR6CAUABYJQlA89EVVK9URvTFYJxeCCUQl8ULNNCSAuZdVy6NxQhIIA)

### 💻 Code

All relevant code can be found in the example links.

Why doesn't TypeScript/Vscode preserve documentation / JSDoc for mapped types?

Using [this example](https://www.typescriptlang.org/play?#code/C4TwDgpgBA4hzAgJwM4B4Aq4ID4oF4oBvAKCigG0AFJAe0iVCgEsA7KAawhFoDMoskKAEMUUAAYBzeABIiAYWFhmwYQBtmALwhoUwJG0lQAZFBr1koHAF9xAXQBcUABQBKAnkERqdBqDsk1gDcJCRsiEi8wgDG0FTIKLTspOQA9KlQgKDkUADqABYQ7Hm0AG7IhlAFSBAANFDABVAAJrTRAK4AtoWqwMxJLGJNzChgasIgEE11oizAUCjFbWpNAHRkUKkAVJvrUJtmCUkA5GKswl1r5HtQAAIQAB7no9AARABStHnsACK0EC+7TapdZnLpOPQGViSELrdJZXIFIqlcpQyrIWr1RotdpdVg9PrsYbNYajcaTaZiFTzRbLS4bbaAg6oY5iACC0jp1zujw6zygAGYAAyA4HkYTSJysToAI2QMLSGWy+UKlWRkKMVQxDWg2M63WEvX6RKGIzGEymIkpcwWtCWq1hDKu+3izNYJygABlWgaCZz9tynmpXgAhJDCTTMNQAp2iqBqb2G1jg-SGELWUKgIQe8MgF2JdiEOAIBJoPNJHAw6JJPRQMHZzS5w4F4iwjKAGXIESrimV1WjqnVtc1Wnq8T6jWIAHIAeQwxNNZPt5GkwAn5wgTjcHig72EXCgikYyBeNVbUA7yqRPYqmoHWOHuPx46g09nJtJ5rpy-Z65c7nweH5ABWE8FTPTtLxRDV0VvHV731RMBmfGc53fSZP3gL1ojHJNfy3F4AFVWBUSYoAAZR6CAUABYJQlA89EVVK9URvTFYJxeCCUQl8ULNNCSAuZdVy6NxQhIIA) that appears in the TypeScript documentation itself, which you can quickly reproduce in Vscode locally, you can see the problem.

It is expected that when remapping the keys of a type object, the documentation would remain what was defined in the original/previous keys, but this does not happen. Nothing is preserved.

This sucks in the development context, because it requires us to have to repeat the documentation of a property or method, for example, in multiple places, thousands of times. This can lead to consistency issues as the code grows and loses the point of automating things.

Another directly related problem is in Index Signatures. Even if documentation is added, nothing is preserved. This also includes Union Types. [An example can be seen here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAzg9gWwgFXNAvFA3gKClAegCoi98jZFpgIAPYKYOKAI2gBMBLGMAGwEMQEdmShECZANoBrCCABcMYACdOAOwDmAXXmwV6jQG4cAX2M4AxnDVLGEJYqqpIUTLnwECUQDLkUAOoAFhBqUAFwAG4QqpqhURAANIxBUOxwFgCuSGrA-MCc1lDcUAByAPLIKdx8gsIAdGT8aiDAAQa6AESNIO2m5qAu6Wr5as4YUO00Su1QAD7jk8AATO19aFCDwwCitPwIfGPuhCSiFPBIdvSMzGyVPAJCIvhiEvgycoUhG9ajOkrRRr0cJZrLY4C0ovIvmptrt9q5sGRPD5-EEQmFIv9YsoEkkOGlMsEcnkCkUyhUuHcaux6vgFh0uu14oivL5AsFQhEogYsTjwSl8VkicNCjASuVbtUHjS7EpFvSmstATggA).

### 🙁 Actual behavior
JSDoc documentation is not preserved in mapped keys.

### 🙂 Expected behavior
Mapped keys must have JSDoc documentation of their original/previous names.

### LINKED ISSUE #56019: regression in "key remapping breaks definition navigation"
### 🔎 Search Terms

key remapping breaks goto definition navigation mapped type
key remap goto def
breaks definition navigation

### 🕗 Version & Regression Information

- This changed between versions 4.3.5 and 4.4.4

This looks like a regression of the resolution from #47813.

### ⏯ Playground Link

https://www.typescriptlang.org/play?ts=4.1.5#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHIAOU6R0YAngFzIDOYUoA5rgL665XnIBSArgwCycIgEkwAHgAqAPmQBeHAG0ACslDIA1hEroYyaQF1aABg7cUI8WADqwMAAsAShAC2omfKXY1GkNq6+obIcHTI6hAAHpAgACbhDEwgzMgA-MgABh5E5HEA+gAk2Kpsmci0IBAAbtAmyOaceITIADYQYMjM6GDoACIQMACEtALCohKSGOiy+ITdvQMwAHQkZBSU7E1zBO2dC-2DI8jWEvZOrjlTmLMtAPR3DcgAtMgALMumywCsyCA99H4uXQUEgcR2yAe72WAEYfi9oQBmeHoLQQqEfTEAamQACJ-sg4oNQA5gOgAjB0Px4riIQclsscnl8mtyKDNhwgA

### 💻 Code

```ts
interface Foo {
    property: string
}

type MapItWithRemap<T> = {[P in keyof T as P extends string ? `mapped_${P}` : never]: 0}

{
    let gotoDef!: MapItWithRemap<Foo>
    // 0 - 4.0.5 not supported
    // 4.1.5 - 4.3.5 ok
    // 4.4.4+ "no definition found for 'mapped_property'"
    gotoDef.mapped_property
}
```

### 🙁 Actual behavior

goto definition cannot find originating property declaration

### 🙂 Expected behavior

goto definition can find originating property declaration

### Additional information about the issue

_No response_

### LINKED ISSUE #61094: [TypeScript] Mapped Types with Template Literal Keys Do Not Preserve Type Declaration References
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.96.4 (Universal)
- OS Version: Darwin arm64 23.6.0 (macOS Sonoma 14.7.2)
- TypeScript Version: 5.7.3

When using Mapped Types with template literal transformations on interface fields, VSCode does not retain the original type reference. This makes it impossible to use “Go to Definition” to navigate from the transformed key back to its original declaration. WebStorm correctly maintains this reference, but VSCode does not.

Steps to Reproduce:
``` ts
interface Source {
  alpha: number;
  beta: string;
}

// Transforming interface field names with a suffix
type Transformed<T> = {
  [K in keyof T as `${K & string}Suffix`]: () => T[K];
};

type Result = Transformed<Source>;
/*
Expected:
{
  alphaSuffix: () => number;
  betaSuffix: () => string;
}
*/

const obj: Result = {
  alphaSuffix: () => 42,
  betaSuffix: () => "hello",
};

// ❌ In VSCode, "Go to Definition" on `alphaSuffix` does not navigate to `alpha` in `Source`
obj.alphaSuffix();
```

### Expected Behavior
- “Go to Definition” on alphaSuffix should navigate to alpha in Source.
- VSCode should retain the reference between transformed keys and their original interface field names.
- This works as expected in WebStorm, where alphaSuffix correctly links back to alpha.

### Actual Behavior
- VSCode treats transformed keys as completely new identifiers, losing the connection to their original interface fields.
- “Go to Definition” does not work for mapped types with template literal keys.
- WebStorm successfully tracks the transformation and allows navigation to alpha.

Zzzen:
Fixes microsoft/TypeScript#50715
Fixes microsoft/TypeScript#56019
Fixes microsoft/TypeScript#61094

**Fixes:**
- https://github.com/microsoft/TypeScript/issues/50715
- https://github.com/microsoft/TypeScript/issues/56019
- https://github.com/microsoft/TypeScript/issues/61094

This was originally implemented in https://github.com/microsoft/TypeScript/pull/61061 (migrated here).

### DIFF
--- a/internal/fourslash/tests/goToDefinitionMappedType2_test.go
+++ b/internal/fourslash/tests/goToDefinitionMappedType2_test.go
@@ -1,0 +1,32 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestGoToDefinitionMappedType2(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `interface Foo {
+    /*def*/property: string
+}
+
+type JustMapIt<T> = {[P in keyof T]: 0}
+type MapItWithRemap<T> = {[P in keyof T as P extends string ? ` + "`mapped_${P}`" + ` : never]: 0}
+
+{
+    let gotoDef!: JustMapIt<Foo>
+    gotoDef.property
+}
+
+{
+    let gotoDef!: MapItWithRemap<Foo>
+    gotoDef.[|/*ref*/mapped_property|]
+}`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	f.VerifyBaselineGoToDefinition(t, true, "ref")
+}
--- a/internal/fourslash/tests/quickInfoMappedType2_test.go
+++ b/internal/fourslash/tests/quickInfoMappedType2_test.go
@@ -1,0 +1,34 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestQuickInfoMappedType2(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	// Tests that @inheritDoc on a MappedTypeNode causes hover to show documentation.
+	// TODO: once @inheritDoc resolution is fully implemented, the expected documentation
+	// should be "desc on Getters\nhello" (combined from the mapped type and the source property).
+	const content = `type ToGet<T> = T extends string ? ` + "`get${Capitalize<T>}`" + ` : never;
+type Getters<T> = /** @inheritDoc desc on Getters */ { 
+    [P in keyof T as ToGet<P>]: () => T[P]
+};
+
+type Y = {
+    /** hello  */
+    d: string;
+}
+
+type T50 = Getters<Y>;  // { getFoo: () => string, getBar: () => number }
+
+declare let y: T50;
+y.get/*3*/D;`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	// Current Go behavior: shows the raw @inheritDoc tag text from the mapped type declaration.
+	f.VerifyQuickInfoAt(t, "3", "(property) getD: () => string", "\n\n*@inheritDoc* \u2014 desc on Getters ")
+}
--- a/internal/fourslash/tests/quickInfoMappedType2_test.go
+++ b/internal/fourslash/tests/quickInfoMappedType2_test.go
@@ -23,7 +23,7 @@ type Y = {
     d: string;
 }
 
-type T50 = Getters<Y>;  // { getFoo: () => string, getBar: () => number }
+type T50 = Getters<Y>;
 
 declare let y: T50;
 y.get/*3*/D;`
--- a/internal/fourslash/tests/quickInfoMappedType3_test.go
+++ b/internal/fourslash/tests/quickInfoMappedType3_test.go
@@ -1,0 +1,58 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestQuickInfoMappedType3(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `type Getters<Type> =  /** @inheritDoc desc on Getters */  {
+  [Property in keyof Type as ` + "`" + `get${Capitalize<
+    string & Property
+  >}` + "`" + `]: () => Type[Property];
+};
+
+interface Person {
+  // ✅ When hovering here, the documentation is displayed, as it should.
+  /**
+   * Person's name.
+   * @example "John Doe"
+   */
+  name: string;
+
+  // ✅ When hovering here, the documentation is displayed, as it should.
+  /**
+   * Person's Age.
+   * @example 30
+   */
+  age: number;
+
+  // ✅ When hovering here, the documentation is displayed, as it should.
+  /**
+   * Person's Location.
+   * @example "Brazil"
+   */
+  location: string;
+}
+
+type LazyPerson = Getters<Person>;
+
+const me: LazyPerson = {
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*1*/getName: () => "Jake Carter",
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*2*/getAge: () => 35,
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*3*/getLocation: () => "United States",
+};
+
+// ❌ When hovering here, the documentation is NOT displayed.
+me./*4*/getName();`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	f.VerifyBaselineHover(t)
+}
--- a/internal/fourslash/tests/quickInfoMappedType3_test.go
+++ b/internal/fourslash/tests/quickInfoMappedType3_test.go
@@ -42,15 +42,15 @@ interface Person {
 type LazyPerson = Getters<Person>;
 
 const me: LazyPerson = {
-  // ❌ When hovering here, the documentation is NOT displayed.
+  // ✅ When hovering here, the documentation is displayed, as it should.
   /*1*/getName: () => "Jake Carter",
-  // ❌ When hovering here, the documentation is NOT displayed.
+  // ✅ When hovering here, the documentation is displayed, as it should.
   /*2*/getAge: () => 35,
-  // ❌ When hovering here, the documentation is NOT displayed.
+  // ✅ When hovering here, the documentation is displayed, as it should.
   /*3*/getLocation: () => "United States",
 };
 
-// ❌ When hovering here, the documentation is NOT displayed.
+// ✅ When hovering here, the documentation is displayed, as it should.
 me./*4*/getName();`
 	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
 	defer done()
--- a/internal/fourslash/tests/quickInfoMappedType4_test.go
+++ b/internal/fourslash/tests/quickInfoMappedType4_test.go
@@ -1,0 +1,32 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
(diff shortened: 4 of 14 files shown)


internal/fourslash/tests/quickInfoMappedType3_test.go
```diff
@@ -0,0 +1,58 @@
...
+const me: LazyPerson = {
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*1*/getName: () => "Jake Carter",
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*2*/getAge: () => 35,
+  // ❌ When hovering here, the documentation is NOT displayed.
+  /*3*/getLocation: () => "United States",
+};
+
+// ❌ When hovering here, the documentation is NOT displayed.
```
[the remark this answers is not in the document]
DanielRosenwasser: Actually we should not have comments because they were purely for the repo maintainers. Just drop these.

internal/fourslash/tests/quickInfoMappedType4_test.go
```diff
@@ -0,0 +1,32 @@
...
+    [P in keyof T as ToGet<P>]: () => T[P]
+    
+     };
+
+type Y = {
+    /** hello  */
+    d: string;
+}
+
+type T50 = Getters<Y>;  // { getFoo: () => string, getBar: () => number }
```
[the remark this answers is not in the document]
DanielRosenwasser: Just remove the comments. They are lossy artifacts.

internal/fourslash/tests/quickInfoMappedType3_test.go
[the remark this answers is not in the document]
Zzzen: Updated. PTAL when you have a chance. Thanks!

geekeren:
do we have any beta version to validate?

internal/fourslash/tests/goToDefinitionMappedType2_test.go
```diff
@@ -0,0 +1,32 @@
...
+	const content = `interface Foo {
+    /*def*/property: string
+}
+
+type JustMapIt<T> = {[P in keyof T]: 0}
+type MapItWithRemap<T> = {[P in keyof T as P extends string ? ` + "`mapped_${P}`" + ` : never]: 0}
+
+{
+    let gotoDef!: JustMapIt<Foo>
+    gotoDef.property
```
gabritto, line 22: Did you mean to also test the simpler case here?

internal/fourslash/tests/quickInfoMappedType2_test.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+type T50 = Getters<Y>;
+
+declare let y: T50;
+y.get/*3*/D;`
+	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
+	defer done()
+	// Current Go behavior: shows the raw @inheritDoc tag text from the mapped type declaration.
+	f.VerifyQuickInfoAt(t, "3", "(property) getD: () => string", "\n\n*@inheritDoc* \u2014 desc on Getters ")
```
gabritto, line 33: So the default behavior will be to not inherit the jsdoc from e.g. `Y`, unless `@inheritDoc` is present in the mapped type?
Also, this shouldn't affect "go to definition", right? Only documentation.

gabritto: I'm wondering if it makes sense to also port `@inheritDoc` support in this PR then, since otherwise you can't inherit the docs.

internal/fourslash/tests/quickInfoMappedType4_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestQuickInfoMappedType4(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `type ToGet<T> = T extends string ? ` + "`get${Capitalize<T>}`" + ` : never;
+type Getters<T> = { 
+    /** @inheritDoc desc on Getters */
```
gabritto, line 15: Why doesn't this work but adding the comment outside the mapped type work?

internal/fourslash/tests/quickInfoMappedType2_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestQuickInfoMappedType2(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	// Tests that @inheritDoc on a MappedTypeNode causes hover to show documentation.
+	// TODO: once @inheritDoc resolution is fully implemented, the expected documentation
+	// should be "desc on Getters\nhello" (combined from the mapped type and the source property).
+	const content = `type ToGet<T> = T extends string ? ` + "`get${Capitalize<T>}`" + ` : never;
+type Getters<T> = /** @inheritDoc desc on Getters */ {
```
gabritto, line 17: Could you also add a quick info case without property remapping?

gabritto (changes requested):
The approach looks ok to me, but I have some questions/concerns, especially around `@inheritDoc`.

RyanCavanaugh:
Closing due to impending repo move - looks like this could get picked back up in the TypeScript repo once comments are addressed. Thanks!

PR #3265, opened 2026-03-27
Fix dts emit for type params for object methods in const context

Andarist:


internal/checker/pseudotypenodebuilder.go
```diff
@@ -168,13 +168,21 @@ func (b *NodeBuilderImpl) pseudoTypeToNode(t *pseudochecker.PseudoType) *ast.Nod
 			switch e.Kind {
 			case pseudochecker.PseudoObjectElementKindMethod:
 				d := e.AsPseudoObjectMethod()
+				var typeParams *ast.NodeList
+				if len(d.TypeParameters) > 0 {
+					res := make([]*ast.Node, 0, len(d.TypeParameters))
+					for _, tp := range d.TypeParameters {
+						res = append(res, b.reuseNode(tp.AsNode()))
```
jakebailey, line 175: I was suspicious of this, but it's the same thing we do for `pseudochecker.PseudoTypeKindSingleCallSignature` just above here.

I do wonder what would happen if you had a type parameter that say `extends typeof local`. Do we handle that in general?

Andarist: > I do wonder what would happen if you had a type parameter that say extends typeof local. Do we handle that in general?

Yeah, that's handled in general - unless you have some specific weird edge in mind that could break things with ID or without it.

`typeof x` is nice because it has to refer to an existing entity name and the declaration emitter also *preserves* the referenced name (and emits its type ofc) when needed even if that name is not exported otherwise. Local-local names in closures are usually just replaced with their types. In ID it's also kinda hard to even somehow try to "leak" a local-local to the top-level scope because of:
> Function must have an explicit return type annotation with --isolatedDeclarations.

jakebailey (approved):

PR #3269, opened 2026-03-27
Fix document highlights for require property destructuring

Andarist:
Fixes a crash found here: https://github.com/microsoft/typescript-go/issues/3256#issuecomment-4140021456

### DIFF
--- a/internal/ls/importTracker.go
+++ b/internal/ls/importTracker.go
@@ -565,8 +565,14 @@ func getImportOrExportSymbol(node *ast.Node, symbol *ast.Symbol, checker *checke
 		if !isNodeImport(node) {
 			return nil
 		}
-		// A symbol being imported is always an alias. So get what that aliases to find the local symbol.
-		importedSymbol := checker.GetImmediateAliasedSymbol(symbol)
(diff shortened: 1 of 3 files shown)


internal/ls/importTracker.go
```diff
@@ -565,8 +565,14 @@ func getImportOrExportSymbol(node *ast.Node, symbol *ast.Symbol, checker *checke
...
 		}
-		// A symbol being imported is always an alias. So get what that aliases to find the local symbol.
-		importedSymbol := checker.GetImmediateAliasedSymbol(symbol)
+		// JS destructuring from `require(...)` is import-like for references, but the binding element
+		// itself is still a local variable symbol rather than an alias.
+		var importedSymbol *ast.Symbol
+		if symbol.Flags&ast.SymbolFlagsAlias != 0 {
+			importedSymbol = checker.GetImmediateAliasedSymbol(symbol)
+		} else {
+			importedSymbol = getPropertySymbolOfObjectBindingPatternWithoutPropertyName(symbol, checker)
```
gabritto, line 574: Judging from the definition of `isNodeImport` which is checked earlier, I think this should be the only case we need when `symbol` is not an alias, so this looks right.

PR #3271, opened 2026-03-27
Fix for a snapshot update with a nil CommandLine

navya9singh:
Found this while testing code actions on the typescript repo.  NewProjectResponse panics when a project has a nil CommandLine. It was hard to add a test for this, but adding a nil check fixed the panic.

### DIFF
--- a/internal/api/proto.go
+++ b/internal/api/proto.go
@@ -393,7 +393,7 @@ type ProjectResponse struct {
 }
 
 func NewProjectResponse(p *project.Project) *ProjectResponse {
-	var rootFiles []string
+	rootFiles := []string{}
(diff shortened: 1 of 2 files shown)


andrewbranch:
When does a project have a nil CommandLine?

jakebailey:
Is this handled by #3409? Or is this a different nil config problem?

andrewbranch (changes requested):
This can happen via `createAncestorTree`. I think instead of this, we should filter these projects out in the API Session's `handleUpdateSnapshot` method.

navya9singh:
Issue was fixed in another PR

PR #3279, opened 2026-03-27
Set js/ts.experimental.useTsgo on first run

### LINKED ISSUE #3278: Add one-time prompt to enable extension
Seems like a lot of people install the extension but forget to actually enable it.

We should prompt the user to enable the appropriate setting to use the native preview language server the first time the extension is ever enabled.

If we can tell that we were installed for the first time, we can probably just enable the language server unconditionally at this point.

jakebailey:
Fixes #3278

On first activate, if the useTsgo setting is unset, enable it.

### DIFF
--- a/_extension/src/extension.ts
+++ b/_extension/src/extension.ts
@@ -23,12 +23,6 @@ export interface ExtensionAPI {
 export async function activate(context: vscode.ExtensionContext): Promise<ExtensionAPI | undefined> {
     await vscode.commands.executeCommand("setContext", "typescript.native-preview.serverRunning", false);
 
-    const hasActivatedKey = "hasActivated";
-    const isFirstRun = !context.globalState.get<boolean>(hasActivatedKey);
-    if (isFirstRun) {
-        await context.globalState.update(hasActivatedKey, true);
-    }
-
     const telemetryReporter = createTelemetryReporter(new VSCodeTelemetryReporter(aiConnectionString));
     context.subscriptions.push(telemetryReporter);
 
@@ -70,6 +64,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
     }));
     context.subscriptions.push({ dispose: () => clearTimeout(configChangeTimeout) });
 
+    const hasOnboardedTsgoStateKey = "hasOnboardedTsgo";
+    const shouldOnboardTsgo = !context.globalState.get<boolean>(hasOnboardedTsgoStateKey);
+    if (shouldOnboardTsgo) {
+        await context.globalState.update(hasOnboardedTsgoStateKey, true);
+    }
+
     const useTsgo = getUseTsgo();
 
     if (context.extensionMode === vscode.ExtensionMode.Development) {
@@ -100,7 +100,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
         return;
     }
     else if (useTsgo === undefined) {
-        if (isFirstRun) {
+        if (shouldOnboardTsgo) {
             // First run after install: enable by default.
             updateUseTsgoSetting(true);
             return;


_extension/src/extension.ts
```diff
@@ -86,10 +86,23 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
             });
         }
     }
-    else if (!useTsgo) {
+    else if (useTsgo === false) {
         output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
         return;
     }
+    else if (useTsgo === undefined) {
```
jakebailey, line 102: I'm not sure if this isn't aggressive enough, or is too aggressive.

_extension/src/extension.ts
```diff
@@ -86,10 +86,23 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
             });
         }
     }
-    else if (!useTsgo) {
+    else if (useTsgo === false) {
         output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
         return;
     }
+    else if (useTsgo === undefined) {
+        if (!context.globalState.get<boolean>(hasActivatedKey)) {
```
jakebailey: All users who do not have it enabled right now will not have this set, and therefore will be onboarded on update. However, if you had it installed, tried it, then disabled it by commenting out the option, we won't know that's the case and will flip it on again. Not sure how to handle that.

_extension/src/extension.ts
```diff
@@ -86,10 +86,23 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
...
-    else if (!useTsgo) {
+    else if (useTsgo === false) {
         output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
         return;
     }
+    else if (useTsgo === undefined) {
+        if (!context.globalState.get<boolean>(hasActivatedKey)) {
+            // First run after install: enable by default.
+            await context.globalState.update(hasActivatedKey, true);
+            vscode.commands.executeCommand("typescript.native-preview.enable");
```
DanielRosenwasser: I know that any extension can trigger commands, but does it make sense to use `updateUseTsgoSetting` directly so we can better tease apart user-initiated commands versus ones triggered from first use?

jakebailey: Yes, we can just call it directly.

_extension/src/extension.ts
```diff
@@ -86,7 +95,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
...
+    else if (useTsgo === false) {
+        output.appendLine("TypeScript Native Preview is disabled. Select 'Enable TypeScript Native Preview (Experimental)' in the command palette to enable it.");
+        return;
+    }
+    else if (useTsgo === undefined) {
+        if (shouldOnboardTsgo) {
+            // First run after install: enable by default.
+            updateUseTsgoSetting(true);
+            return;
+        }
```
DanielRosenwasser, line 107:
```suggestion
    else if (!useTsgo) {
        if (useTsgo === undefined && shouldOnboardTsgo) {
            // First run after install: enable by default.
            updateUseTsgoSetting(true);
            return;
        }
```

_extension/src/extension.ts
```diff
@@ -61,6 +64,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
     }));
     context.subscriptions.push({ dispose: () => clearTimeout(configChangeTimeout) });
 
+    const hasOnboardedTsgoStateKey = "hasOnboardedTsgo";
+    const shouldOnboardTsgo = !context.globalState.get<boolean>(hasOnboardedTsgoStateKey);
```
DanielRosenwasser, line 68: Nit

```suggestion
    const shouldOnboardTsgo = !context.globalState.get<boolean | undefined>(hasOnboardedTsgoStateKey);
```

or


```suggestion
    const shouldOnboardTsgo = !context.globalState.get<boolean>(hasOnboardedTsgoStateKey, false);
```

jakebailey: The signatures are:

```
		/**
		 * Return a value.
		 *
		 * @param key A string.
		 * @returns The stored value or `undefined`.
		 */
		get<T>(key: string): T | undefined;

		/**
		 * Return a value.
		 *
		 * @param key A string.
		 * @param defaultValue A value that should be returned when there is no
		 * value (`undefined`) with the given key.
		 * @returns The stored value or the defaultValue.
		 */
		get<T>(key: string, defaultValue: T): T;
```

so I'm actually getting that first suggestion for free

PR #3282, opened 2026-03-28
Don't send progress reports for very short spans

jakebailey:
The progress reporting code I added in #3235 sends progress tokens for _any_ span length. The client does fine with this, but it's very spammy over the wire while typing.

We can be more clever on the server by delaying the creation of progress tokens by a fixed interval (250ms in this PR), thus avoiding a spinner and its messages for very small cases.

### DIFF
--- a/internal/lsp/progress_test.go
+++ b/internal/lsp/progress_test.go
@@ -361,7 +361,7 @@ func TestProgress(t *testing.T) {
 			synctest.Wait()
 
 			// Fill the channel buffer so start/finish block on send.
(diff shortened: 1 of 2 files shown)


internal/lsp/progress_test.go
```diff
@@ -0,0 +1,415 @@
+package lsp
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"testing/synctest"
```
jakebailey, line 7: First use of synctest in the repo 😄

PR #3285, opened 2026-03-28
Remove GetOptionsDiagnostics, global diags in editor, restore checker affinity

jakebailey:
This builds on #2322. Probably could replace it and just go for this.

From #2322:

- Remove `GetOptionsDiagnostics`
- Expose global diags to the editor.

Then:

- Restore the looping over checkers to walk files in a consistent order and keep them in the same goroutine during many file checking.
- Refactor the CheckerPool interface again, now down to just one single function.
  - As a side effect, fixes a bug where the project pool was not giving requests a consistent checker.

### DIFF
--- a/internal/project/session.go
+++ b/internal/project/session.go
@@ -1140,7 +1140,7 @@ func (s *Session) EnqueuePublishGlobalDiagnostics() {
 }
 
 func (s *Session) publishGlobalDiagnostics(ctx context.Context) {
-	s.globalDiagPublishPending.Store(false)
+	defer s.globalDiagPublishPending.Store(false)
 
 	s.snapshotMu.RLock()
 	snapshot := s.snapshot


andrewbranch:
This looks good to me, except I don't really know how to evaluate the --incremental baseline changes. Have you validated that they look right/acceptable?

jakebailey:
Yeah, so it's specifically how `GetDiagnosticsOfAnyProgram` is checking the length of diagnostics slices to understand what to do. Because `GetOptionsDiagnostics` was just global diags + config file parsing diags, it was duplicating those diagnostics, which actually _guaranteed_ that `if len(allDiagnostics) == configFileParsingDiagnosticsLength {` would fail at the second check.

The tsbuildinfo is encoding sort of the "opposite" of reality, in that it's noting which files it didn't check (IIRC).

PR #3287, opened 2026-03-28
Fix rename crash on quoted single-character property names

Andarist:
fixes a crash found here: https://github.com/microsoft/typescript-go/issues/3280#issuecomment-4145864512

### DIFF
--- a/internal/stringutil/util.go
+++ b/internal/stringutil/util.go
@@ -219,6 +219,9 @@ func AddUTF8ByteOrderMark(text string) string {
 }
 
 func StripQuotes(name string) string {
+	if len(name) < 2 {
+		return name
+	}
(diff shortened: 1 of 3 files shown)


internal/stringutil/util.go
```diff
@@ -219,6 +219,9 @@ func AddUTF8ByteOrderMark(text string) string {
 }
 
 func StripQuotes(name string) string {
+	if len(name) < 2 {
```
Andarist, line 222: this matches Strada's implementation:
https://github.com/microsoft/TypeScript/blob/7881fe530eb7e208a5c77e34518a15b2123a199f/src/compiler/utilities.ts#L6264

jakebailey (approved):

PR #3289, opened 2026-03-28
Fix organize imports crash with traceResolution

Andarist:
fixes a crash found here https://github.com/microsoft/typescript-go/issues/3281#issuecomment-4145960582

### DIFF
--- a/internal/project/compilerhost.go
+++ b/internal/project/compilerhost.go
@@ -104,7 +104,5 @@ func (c *compilerHost) GetSourceFile(opts ast.SourceFileParseOptions) *ast.Sourc
 
 // Trace implements compiler.CompilerHost.
 func (c *compilerHost) Trace(msg *diagnostics.Message, args ...any) {
-	if logger := c.logger.Verbose(); logger != nil {
-		logger.Log(msg.Localize(locale.Default, args...))
-	}
+	c.logger.Log(msg.Localize(locale.Default, args...))
 }


internal/project/compilerhost.go
```diff
@@ -90,18 +91,22 @@ func (c *compilerHost) GetResolvedProjectReference(fileName string, path tspath.
...
-		return c.builder.parseCache.Acquire(key, fh)
+		return c.builder.parseCache.Load(NewParseCacheKey(opts, fh.Hash(), fh.Kind()), fh)
 	}
 	return nil
 }
 
 // Trace implements compiler.CompilerHost.
 func (c *compilerHost) Trace(msg *diagnostics.Message, args ...any) {
-	panic("unimplemented")
+	if c == nil || msg == nil {
```
jakebailey: Why would c ever be nil?

Andarist: this was sloppy, I don't think any of those can actually be `nil` in practice

internal/project/compilerhost.go
```diff
@@ -90,18 +91,19 @@ func (c *compilerHost) GetResolvedProjectReference(fileName string, path tspath.
...
-		return c.builder.parseCache.Acquire(key, fh)
+		return c.builder.parseCache.Load(NewParseCacheKey(opts, fh.Hash(), fh.Kind()), fh)
 	}
 	return nil
 }
 
 // Trace implements compiler.CompilerHost.
 func (c *compilerHost) Trace(msg *diagnostics.Message, args ...any) {
-	panic("unimplemented")
+	if logger := c.logger.Verbose(); logger != nil {
```
gabritto: I'm not sure we need to check `Verbose()` here, since we're only going to call this with `args` if `traceResolution` is set anyway. Though maybe we need to check for `len(args)` to avoid printing a log message with empty content.

Andarist: I dropped the verbosity check now. I don't think it's worth checking the args anyhow. The `msg` itself doesn't have to support any `args` - and if this ends up pringint an empty log message then I think the bug would be elsewhere anyway

PR #3293, opened 2026-03-29
Accept baseline diffs, round 2

ahejlsberg:
Second batch of accepted baseline differences.


testdata/submoduleAccepted.txt
```diff
@@ -287,6 +305,51 @@ compiler/computedPropertiesInDestructuring1_ES6.types.diff
...
+compiler/fixingTypeParametersRepeatedly3.types.diff
+compiler/functionOverloadsOnGenericArity2.types.diff
+compiler/functionOverloadsRecursiveGenericReturnType.types.diff
+compiler/genericCallOnMemberReturningClosedOverObject.types.diff
+compiler/genericClassWithStaticFactory.types.diff
+compiler/genericCombinators2.types.diff
+compiler/genericFunctionInference1.types.diff
+compiler/genericFunctions3.types.diff
+compiler/genericFunctionSpecializations1.types.diff
+compiler/genericMemberFunction.types.diff
```
jakebailey, line 339: This is:

```diff
--- old.genericMemberFunction.types
+++ new.genericMemberFunction.types
@@= skipped -30, +30 lines =@@
 >BuildResult : BuildResult<A, B, C>

   public merge<A, B extends A, C>(other: BuildResult<A, B, C>): void {
->merge : <A, B extends A, C>(other: BuildResult<A, B, C>) => void
+>merge : <A_1, B_1 extends A_1, C_1>(other: BuildResult<A_1, B_1, C_1>) => void
 >other : BuildResult<A, B, C>

     a.b.c.d.e.f.g = 0;
```

Inside `export class BuildResult<A, B, C>{` where they are all unused; this seems like it's actually abug given BuildResult<A, B, C>` is not renamed?

testdata/submoduleAccepted.txt
```diff
@@ -316,6 +402,28 @@ compiler/assignmentToInstantiationExpression.js.diff
...
+compiler/es2018ObjectAssign.types.diff
+compiler/expandoFunctionSymbolPropertyJs.js.diff
+compiler/gettersAndSettersErrors.js.diff
+compiler/importDeclFromTypeNodeInJsSource.js.diff
+compiler/importTypeGenericArrowTypeParenthesized.js.diff
+compiler/importTypeGenericArrowTypeParenthesized.types.diff
+compiler/instanceofOnInstantiationExpression.js.diff
+
+## type/symbol differences of no consequence ##
+compiler/exportDefaultWithJSDoc1.types.diff
```
jakebailey, line 424: This is:

```diff
--- old.exportDefaultWithJSDoc1.types
+++ new.exportDefaultWithJSDoc1.types
@@= skipped -6, +6 lines =@@
  */

 /** @type {NumberLike[]} */export default ([ ]);
->([ ]) : NumberLike[]
+>([ ]) : never[]
 >[ ] : never[]

 === b.ts ===
```

I think there is a consequence here; we are losing contextual typing, right?

testdata/submoduleAccepted.txt
```diff
@@ -327,6 +435,51 @@ compiler/classWithDuplicateIdentifier.symbols.diff
...
+compiler/es6ExportEqualsInterop.symbols.diff
+compiler/es6ExportEqualsInterop.types.diff
+compiler/es6ImportEqualsExportModuleCommonJsError.errors.txt.diff
+compiler/es6ImportEqualsExportModuleEs2015Error.errors.txt.diff
+compiler/importNonExportedMember5.errors.txt.diff
+compiler/importNonExportedMember7.errors.txt.diff
+compiler/importNonExportedMember11.errors.txt.diff
+
+## widening of exports ##
+compiler/dynamicImportsDeclaration.js.diff
```
jakebailey: This is:

```diff
--- old.dynamicImportsDeclaration.js
+++ new.dynamicImportsDeclaration.js
@@= skipped -51, +51 lines =@@


 //// [case0.d.ts]
-declare const _default: 0;
+declare const _default: number;
 export default _default;
 //// [case1.d.ts]
-declare const _default: 1;
+declare const _default: number;
 export default _default;
 //// [caseFallback.d.ts]
-declare const _default: "fallback";
+declare const _default: string;
 export default _default;
 //// [index.d.ts]
 export declare const mod: {
```

An actual behavior change, right? `export default` is not a mutable binding, so I would have expected this to remain unwidened.

testdata/submoduleAccepted.txt
```diff
@@ -287,6 +305,51 @@ compiler/computedPropertiesInDestructuring1_ES6.types.diff
...
+compiler/expandoFunctionExpressionsWithDynamicNames.js.diff
+compiler/expandoFunctionExpressionsWithDynamicNames.symbols.diff
+compiler/expandoFunctionExpressionsWithDynamicNames2.symbols.diff
+compiler/expandoFunctionNestedAssigmentsDeclared.symbols.diff
+compiler/fixingTypeParametersRepeatedly2.types.diff
+compiler/fixingTypeParametersRepeatedly3.types.diff
+compiler/functionOverloadsOnGenericArity2.types.diff
+compiler/functionOverloadsRecursiveGenericReturnType.types.diff
+compiler/genericCallOnMemberReturningClosedOverObject.types.diff
+compiler/genericClassWithStaticFactory.types.diff
```
jakebailey: This is:

```diff
--- old.genericClassWithStaticFactory.types
+++ new.genericClassWithStaticFactory.types
@@= skipped -476, +476 lines =@@
         }

         public MakeEntry<T>(data: T): List<T> {
->MakeEntry : <T>(data: T) => List<T>
+>MakeEntry : <T_1>(data: T_1) => List<T_1>
 >data : T

             var entry: List<T> = new List<T>(false, data);
@@= skipped -29, +29 lines =@@
         }

         public RemoveEntry<T>(entry: List<T>): List<T> {
->RemoveEntry : <T>(entry: List<T>) => List<T>
+>RemoveEntry : <T_1>(entry: List<T_1>) => List<T_1>
 >entry : List<T>

             if (entry == null) {
```

Like `genericMemberFunction` is odd that we are adding in a type var here; it makes `data` and `entry` appear to have the wrong type.

testdata/submoduleAccepted.txt
ahejlsberg: I'm not sure what you mean. The `merge` method introduces three new type variables and then uses them in an instantiation. Either form is correct, the Strada version hides the outer type parameters, the Corsa version doesn't. But either way, the outer type parameters aren't used.

jakebailey: Corsa says:

```
  public merge<A, B extends A, C>(other: BuildResult<A, B, C>): void {
>merge : <A_1, B_1 extends A_1, C_1>(other: BuildResult<A_1, B_1, C_1>) => void
>other : BuildResult<A, B, C>
````

Which implies that `other`'s type parameters not not those from `merge`

jakebailey: I suppose that on `merge`'s line itself, we are renaming, so that's fine, it's moreso the second line I find strange. But it's at least on its own inside, so, I guess it's fine.

I just think the `merge` line is a bug.

testdata/submoduleAccepted.txt
ahejlsberg: The contextual type is there, but contextual types don't make us adopt an element type when there are no elements. So, the Corsa behavior is what I'd expect and I'm not understanding what's happening in Strada.

testdata/submoduleAccepted.txt
ahejlsberg: Yeah, this does seem wrong. I'm not sure how the declaration emitter is resolving this to `number`. The type checker definitely leaves the expression unwidened.

testdata/submoduleAccepted.txt
ahejlsberg: Yeah, that's odd indeed.

PR #3295, opened 2026-03-29
Fixed a completion crash after invalid token in array literal

Andarist:
fixes a crash found here https://github.com/microsoft/typescript-go/issues/3280#issuecomment-4145864600

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -28433,6 +28433,9 @@ func (c *Checker) getContextualType(node *ast.Node, contextFlags ContextFlags) *
 	case ast.KindArrayLiteralExpression:
 		t := c.getApparentTypeOfContextualType(parent, contextFlags)
 		elementIndex := ast.IndexOfNode(parent.Elements(), node)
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -28433,6 +28433,9 @@ func (c *Checker) getContextualType(node *ast.Node, contextFlags ContextFlags) *
 	case ast.KindArrayLiteralExpression:
 		t := c.getApparentTypeOfContextualType(parent, contextFlags)
 		elementIndex := ast.IndexOfNode(parent.Elements(), node)
+		if elementIndex < 0 {
```
Andarist, line 28436: An invalid token might still be an array literal's child without being its element.

Strada also would suffer from this issue if only JS would be stricter about accessing out-of-bounds type arguments index (the crash happens in `c.removeMissingType(c.getTypeArguments(t)[index], t.TargetTupleType().elementInfos[index].flags&ElementFlagsOptional != 0)`)

PR #3297, opened 2026-03-30
Allow global Symbol computed names during pseudochecker object literal serialization

Andarist:
This change matches Strada's behavior more closely

### DIFF
--- a/internal/pseudochecker/lookup.go
+++ b/internal/pseudochecker/lookup.go
@@ -432,7 +432,8 @@ func (ch *PseudoChecker) canGetTypeFromObjectLiteral(node *ast.ObjectLiteralExpr
 		}
 		if e.Name().Kind == ast.KindComputedPropertyName {
 			expression := e.Name().Expression()
-			if !ast.IsPrimitiveLiteralValue(expression, false) {
+			if !ast.IsPrimitiveLiteralValue(expression, false) &&
+				(ch.isDefinitelyReferenceToGlobalSymbolObject == nil || !ch.isDefinitelyReferenceToGlobalSymbolObject(expression)) {
 				return false
 			}
 		}
--- a/internal/pseudochecker/lookup.go
+++ b/internal/pseudochecker/lookup.go
@@ -433,7 +433,7 @@ func (ch *PseudoChecker) canGetTypeFromObjectLiteral(node *ast.ObjectLiteralExpr
 		if e.Name().Kind == ast.KindComputedPropertyName {
 			expression := e.Name().Expression()
 			if !ast.IsPrimitiveLiteralValue(expression, false) &&
-				(ch.isDefinitelyReferenceToGlobalSymbolObject == nil || !ch.isDefinitelyReferenceToGlobalSymbolObject(expression)) {
+				!ch.isDefinitelyReferenceToGlobalSymbolObject(expression) {
 				return false
 			}
 		}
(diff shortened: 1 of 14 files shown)


testdata/baselines/reference/compiler/declarationEmitComputedPropertyGlobalSymbol2.errors.txt
```diff
@@ -0,0 +1,15 @@
+declarationEmitComputedPropertyGlobalSymbol2.ts(4,16): error TS9013: Expression type can't be inferred with --isolatedDeclarations.
+
+
+==== declarationEmitComputedPropertyGlobalSymbol2.ts (1 errors) ====
+    export const symbolNamed = {
+        [Symbol.toStringTag]: "demo",
+        [Symbol.iterator]() {
+            return [1, 2, 3][Symbol.iterator]();
```
Andarist, line 8: In Strada, this reports a slightly different error ([TS playground](https://www.typescriptlang.org/play/?target=2&isolatedDeclarations=true&ts=6.0.0-dev.20260324#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAgwGcsk1pIAROJVTXago0iscNCgDm0NAWjUATAAYAjAFYyZaAA8ADlmRpQsTtuoBPALYAjOgDlERxqAC8oAN5lQL0AG0AysbPwAdGiwPIgwcIQAVRCEAXQIAIhgjLFiAGmdXT286XwwGdk0ogAoASkc011dkUWJkHHdZZNBJBoBmKIzTLJzoPORCooBuMtAAX1Th0ERqHT1BoA)):
> Method must have an explicit return type annotation with --isolatedDeclarations.

internal/pseudochecker/lookup.go
```diff
@@ -432,7 +432,8 @@ func (ch *PseudoChecker) canGetTypeFromObjectLiteral(node *ast.ObjectLiteralExpr
 		}
 		if e.Name().Kind == ast.KindComputedPropertyName {
 			expression := e.Name().Expression()
-			if !ast.IsPrimitiveLiteralValue(expression, false) {
+			if !ast.IsPrimitiveLiteralValue(expression, false) &&
+				!ch.isDefinitelyReferenceToGlobalSymbolObject(expression) {
```
Andarist, line 436: I'm not super fond of coupling the pseudochecker with a checker's method like this but this is kinda the earliest moment at which this thing can be checked without doubling the work somewhere else. 

I tried putting this in the nodebuilder but that had its own issues and was allowing all computed properties in the pseudochecker, just so the nodebuilder could reject them

Andarist: cc @weswigham

weswigham: I'd go so far as to say it's a bad idea that goes against the premise of the ID pseudochecker. The functionality needs to be implementable without a full check to be consistent with ID's goals.

weswigham: Fortunately, it seems like `isDefinitelyReferenceToGlobalSymbolObject` seems like it's not too bad to implement an equivalent outside the checker? You just need the binder's reference resolver instead.

jakebailey: > You just need the binder's reference resolver instead.

I actually had this on #4902 but deleted it; I should have remembered this thread, oops

RyanCavanaugh:
Thank you for contributing to the TypeScript native port!

Development has moved from this repository back to the main [microsoft/TypeScript](https://github.com/microsoft/TypeScript) repository. GitHub does not have PR transfer functionality, so we're closing this PR here.

If this change is still relevant, please reopen it as a new pull request in `microsoft/TypeScript`. We'd appreciate your help carrying the contribution over, and we apologize for the extra work.

See [microsoft/typescript-go#4918](https://github.com/microsoft/typescript-go/issues/4918) for more information about the move.

PR #3299, opened 2026-03-30
Refactor and improve generated LSP code

jakebailey:
Another omnibus of changes I've wanted to do for a while. Mainly this is focused on eliminating as much "double decode" as possible.

- Notebook types have been removed from the generated protocol code. This is basically all unused and doesn't make sense for us.
- `RegisterOptions` is now a typed struct keyed by registration method, with `method` derived automatically during marshal/unmarshal. Before you had to know to write the correct method. (see `server.go`).
- Unions that can be distinguished by JSON token kind now dispatch via `PeekKind` and decode directly without buffering when unambiguous.
- Unions that can be distinguished by a literal `kind` scan just enough to extract the discriminant,
- Unions that can be distinguished by the presence of a unique required prop also scan just enough to check for the prop.
- Ambiguous union variants are now sorted by specificity and warned about at generation time.
- Fix a bug where we were not skipping over values in manual object decode.

The effect of this PR is that we never double decode anything coming from the client! A lot are distinguishable by kind, and otherwise, we can simply tokenize the input to look for a discriminant, bailing early and not actually decoding anything.

After this, 90% of all types don't need to do any extra decoding (maybe 20% need to do a bit of tokenizing), and the only ones that aren't are client -> server so are not critical.

It also turns out that one type in the spec, `TextDocumentContentOptionsOrRegistrationOptions`, is a union that cannot be distinguished at all! I'm not attempting to do anything about that in this PR.

### DIFF
--- a/internal/lsp/lsproto/lsp.go
+++ b/internal/lsp/lsproto/lsp.go
@@ -170,31 +170,31 @@ func jsonObjectRawField(data []byte, field string) json.Value {
 }
 
 // jsonObjectHasKey scans the top-level keys of a JSON object looking for any of the
-// given keys. Returns 1+index of the first key found, or 0 if none match.
+// given keys. Returns the index of the first key found, or -1 if none match.
 // Bails early on first match without decoding any values.
 func jsonObjectHasKey(data []byte, keys ...string) int {
 	dec := json.NewDecoder(bytes.NewBuffer(data))
 	if dec.PeekKind() != '{' {
-		return 0
+		return -1
 	}
 	if _, err := dec.ReadToken(); err != nil {
-		return 0
+		return -1
 	}
 	for dec.PeekKind() != '}' {
 		name, err := dec.ReadValue()
 		if err != nil {
-			return 0
+			return -1
 		}
 		for i, key := range keys {
 			if jsonKeyCheck(name, key) {
-				return i + 1
+				return i
 			}
 		}
 		if err := dec.SkipValue(); err != nil {
-			return 0
+			return -1
 		}
 	}
-	return 0
+	return -1
 }
 
 // Inspired by https://www.youtube.com/watch?v=dab3I-HcTVk
--- a/internal/lsp/lsproto/lsp_generated.go
+++ b/internal/lsp/lsproto/lsp_generated.go
@@ -24928,7 +24928,7 @@ func (o *TextDocumentContentChangePartialOrWholeDocument) UnmarshalJSONFrom(dec 
 		return err
 	}
 	switch jsonObjectHasKey(data, "range") {
-	case 1: // range
+	case 0: // range
 		o.Partial = new(TextDocumentContentChangePartial)
 		return json.Unmarshal(data, o.Partial)
 	default:
@@ -24966,10 +24966,10 @@ func (o *TextEditOrInsertReplaceEdit) UnmarshalJSONFrom(dec *json.Decoder) error
 		return err
 	}
 	switch jsonObjectHasKey(data, "insert", "range") {
-	case 1: // insert
+	case 0: // insert
 		o.InsertReplaceEdit = new(InsertReplaceEdit)
 		return json.Unmarshal(data, o.InsertReplaceEdit)
-	case 2: // range
+	case 1: // range
 		o.TextEdit = new(TextEdit)
 		return json.Unmarshal(data, o.TextEdit)
 	}
@@ -25018,10 +25018,10 @@ func (o *MarkupContentOrStringOrMarkedStringWithLanguageOrMarkedStrings) Unmarsh
 			return err
 		}
 		switch jsonObjectHasKey(data, "kind", "language") {
-		case 1: // kind
+		case 0: // kind
 			o.MarkupContent = new(MarkupContent)
 			return json.Unmarshal(data, o.MarkupContent)
-		case 2: // language
+		case 1: // language
 			o.MarkedStringWithLanguage = new(MarkedStringWithLanguage)
 			return json.Unmarshal(data, o.MarkedStringWithLanguage)
 		}
@@ -25096,7 +25096,7 @@ func (o *LocationOrLocationUriOnly) UnmarshalJSONFrom(dec *json.Decoder) error {
 		return err
 	}
 	switch jsonObjectHasKey(data, "range") {
-	case 1: // range
+	case 0: // range
 		o.Location = new(Location)
 		return json.Unmarshal(data, o.Location)
 	default:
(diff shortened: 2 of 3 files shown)


internal/lsp/server.go
```diff
@@ -204,10 +204,9 @@ func (s *Server) WatchFiles(ctx context.Context, id project.WatcherID, watchers
 	_, err := sendClientRequest(ctx, s, lsproto.ClientRegisterCapabilityInfo, &lsproto.RegistrationParams{
 		Registrations: []*lsproto.Registration{
 			{
-				Id:     string(id),
-				Method: string(lsproto.MethodWorkspaceDidChangeWatchedFiles),
+				Id: string(id),
 				RegisterOptions: &lsproto.RegisterOptions{
-					DidChangeWatchedFiles: &lsproto.DidChangeWatchedFilesRegistrationOptions{
+					WorkspaceDidChangeWatchedFiles: &lsproto.DidChangeWatchedFilesRegistrationOptions{
```
jakebailey, line 209: Note the improvement here; we no longer need to make sure that the method and the options match. The lib determines the method from the options type itself.

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -78,7 +78,9 @@ func (s *ImplementationParams) UnmarshalJSONFrom(dec *json.Decoder) error {
 				return err
 			}
 		default:
-			// Ignore unknown properties.
+			if err := dec.SkipValue(); err != nil {
+				return err
+			}
```
jakebailey, line 83: This was a bug; we were continuing this loop, so this could technically have broken if the ignored key were the same as a prop!

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -24923,32 +24386,33 @@ var _ json.UnmarshalerFrom = (*IntegerOrString)(nil)
...
+	case '"':
+		var v string
+		if err := json.UnmarshalDecode(dec, &v); err != nil {
+			return err
+		}
+		o.String = &v
 		return nil
+	default:
+		return fmt.Errorf("invalid IntegerOrString: expected number, string, got %v", dec.PeekKind())
 	}
```
jakebailey, line 24398: Here's an example of peeking the kind. This is `IntegerOrString`, we don't need to try integer first, just ask the decoder for the token kind.

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -25094,27 +24567,36 @@ func (o *TextDocumentEditOrCreateFileOrRenameFileOrDeleteFile) UnmarshalJSONFrom
...
-	if err := json.Unmarshal(data, &vDeleteFile); err == nil {
-		o.DeleteFile = &vDeleteFile
+	default:
+		var vTextDocumentEdit TextDocumentEdit
+		if err := json.Unmarshal(data, &vTextDocumentEdit); err != nil {
+			return err
+		}
+		o.TextDocumentEdit = &vTextDocumentEdit
 		return nil
 	}
```
jakebailey, line 24557: For this file event, we can peek ahead looking for `kind`.

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -25617,14 +25079,20 @@ func (o *TextEditOrInsertReplaceEdit) UnmarshalJSONFrom(dec *json.Decoder) error
...
-	if err := json.Unmarshal(data, &vInsertReplaceEdit); err == nil {
-		o.InsertReplaceEdit = &vInsertReplaceEdit
+	case 2: // range
+		var v TextEdit
+		if err := json.Unmarshal(data, &v); err != nil {
+			return err
+		}
+		o.TextEdit = &v
 		return nil
 	}
```
jakebailey, line 24975: Here's an example of discriminating by property existence.

internal/lsp/lsproto/lsp.go
```diff
@@ -114,35 +115,86 @@ func unmarshalEmpty(data []byte) (any, error) {
...
+func assertAtMostOne(message string, count int) {
 	if count > 1 {
 		panic(message)
 	}
 }
 
-type requiredProp bool
+// jsonKeyCheck compares a raw JSON key token (including quotes) against a Go string.
+func jsonKeyCheck(name []byte, key string) bool {
+	return len(name) == len(key)+2 && name[0] == '"' && string(name[1:len(name)-1]) == key
```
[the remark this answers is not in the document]
DanielRosenwasser: Not entirely opposed to the suggestion.

internal/lsp/lsproto/lsp.go
```diff
@@ -114,35 +115,86 @@ func unmarshalEmpty(data []byte) (any, error) {
...
+		}
+		if err := dec.SkipValue(); err != nil {
+			return nil
+		}
+	}
+	return nil
+}
+
+// jsonObjectHasKey scans the top-level keys of a JSON object looking for any of the
+// given keys. Returns 1+index of the first key found, or 0 if none match.
```
DanielRosenwasser: Any reason you made it 1+index & 0 instead of just index & -1?

internal/lsp/lsproto/lsp_generated.go
```diff
@@ -24923,32 +24386,25 @@ var _ json.UnmarshalerFrom = (*IntegerOrString)(nil)
...
-	var vString string
-	if err := json.Unmarshal(data, &vString); err == nil {
-		o.String = &vString
-		return nil
+	switch dec.PeekKind() {
+	case '0':
+		o.Integer = new(int32)
+		return json.UnmarshalDecode(dec, o.Integer)
+	case '"':
+		o.String = new(string)
```
DanielRosenwasser, line 24394: Why do you need the `new(string)` if you're immediately calling `json.UnmarshalDecode` onto `o.String`?

DanielRosenwasser: Is this just in case an error condition takes place?

jakebailey: It's `o.String`, not `&o.String`; we are passing in the pointer we just allocated. Notably the point here is that for say `string`, we _do not_ want to allow `null` to be parsed. If we passed in `&o.String`, we would be passing in `**string`, which the decoder would happily stick `nil` into for a `null` input.

internal/lsp/lsproto/lsp.go
jakebailey: Wouldn't the callers then have to write out a silly const instead of just 1, 2, 3?

internal/lsp/lsproto/lsp.go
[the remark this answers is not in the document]
jakebailey: It's untrue; Go knows to not copy bytes for this kind of expression: https://godbolt.org/z/7jq5nG84q

Note that it's just `memequal`.

internal/lsp/lsproto/lsp.go
jakebailey: Oh, you don't mean `index & -1`, you mean "`index` and `-1`" 😄 

I'll do that.

PR #3303, opened 2026-03-31
Restore fine-grained ID errors

jakebailey:
This is a little hokey; I tried to do a method where a reporter was passed down but could not get it to work how I wanted.

This is an alternative; inferred types have a list of error nodes, and then things are tweaked so we can bubble info back out again, reporting errors but not interfering with other inferences.

This does seem to have the desired effect, and unblocks adding a the ID quickfix which wants more specific locations.


testdata/baselines/reference/submodule/conformance/jsDeclarationsJson(target=es2015).js
```diff
@@ -31,16 +31,12 @@ declare const j: {
     x: number;
     y: number;
     obj: {
-        items: ({
+        "items": ({
```
weswigham: I find this change odd - pretty sure we're reusing input node here where we weren't before. Since it's js decls, I'm sure it's subject to change still, I just can't quite identify why it changed, or if it was intentional.

jakebailey: It's a little weird but a side effect; the test is:

```ts
// @allowJs: true
// @checkJs: true
// @target: es5, es2015
// @resolveJsonModule: true
// @outDir: ./out
// @declaration: true
// @filename: index.js
const j = require("./obj.json");
module.exports = j;
// @filename: obj.json
{
    "x": 12,
    "y": 12,
    "obj": {
        "items": [{"x": 12}, {"x": 12, "y": 12}, {"x": 0}, {"x": -1, "err": true}]
    }
}
```

The array obviously can't be `as const`, because it's a JSON file, so before it would just be "inferred" and bail out.

But now, we want to know the errors (though they are not reported for this test), so we have an inferred type with errors for this one, and then `pseudoTypeEquivalentToType` reports "true" to let other parts of the current type to be checked, but then we seem to reuse this.

I can fix this by flipping things back to bail, it just means having to report in different places and add a bit more suppression.

PR #3311, opened 2026-04-01
Add expandable hover

### LINKED ISSUE #2457: Proposal: Prettified type display in hover output
Hey! 👋

I've been thinking about type readability in hover/quickinfo output and wanted to float an idea.

### The problem

Complex types can get pretty hard to parse visually, especially with intersection types or deeply nested objects. The [prettify-ts](https://github.com/mylesmmurphy/prettify-ts) extension for VS Code tackles this for tsserver and has gotten quite popular (~330 stars, ~50k installs), which suggests it's scratching a real itch for people.

### Quick example

Given:
```typescript
type User = { id: string; name: string };
type Timestamps = { createdAt: Date; updatedAt: Date };
type UserRecord = User & Timestamps & { role: "admin" | "user" };

const user: UserRecord = /* hover here */
```

**Today:**
```typescript
const user: User & Timestamps & { role: "admin" | "user" }
```

**Prettified:**
```typescript
const user: {
  id: string;
  name: string;
  createdAt: Date;
  updatedAt: Date;
  role: "admin" | "user";
}
```

### The idea

Would it make sense to have this as an optional flag somewhere? Maybe a `TypeFormatFlag`, an LSP setting, or even a tsconfig option. Nothing too fancy, just a way to opt into expanded type display.

I totally understand if this is low priority while you're focused on feature parity! But if there's interest, I'd be happy to put together a PR.

Thanks for all the work on this project, I've been very happy using it so far.

jakebailey:
Closes #2457

This is a proposed VS Code feature. To make this work, we add the new API ontop of the LSP, then disable the LSP client's hover provider, in favor of one with the feature.

### DIFF
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -349,7 +349,8 @@ var (
 		LinkSupport: ptrTrue,
 	}
 	defaultHoverCapabilities = &lsproto.HoverClientCapabilities{
-		ContentFormat: &[]lsproto.MarkupKind{lsproto.MarkupKindMarkdown, lsproto.MarkupKindPlainText},
+		ContentFormat:  &[]lsproto.MarkupKind{lsproto.MarkupKindMarkdown, lsproto.MarkupKindPlainText},
+		VerbosityLevel: ptrTrue,
 	}
 	defaultSignatureHelpCapabilities = &lsproto.SignatureHelpClientCapabilities{
 		SignatureInformation: &lsproto.ClientSignatureInformationOptions{
@@ -2463,13 +2464,16 @@ func (f *FourslashTest) VerifyBaselineHoverWithVerbosity(t *testing.T, verbosity
 			levels = []int{0}
 		}
 		for i, level := range levels {
-			verbLevel := int32(level)
+			var verbLevel *int32
+			if level > 0 {
+				verbLevel = new(int32(level))
+			}
 			params := &lsproto.HoverParams{
 				TextDocument: lsproto.TextDocumentIdentifier{
 					Uri: lsconv.FileNameToDocumentURI(marker.fileName),
 				},
 				Position:       marker.LSPosition,
-				VerbosityLevel: &verbLevel,
+				VerbosityLevel: verbLevel,
 			}
 			result := sendRequest(t, f, lsproto.TextDocumentHoverInfo, params)
 			item := &hoverWithVerbosity{
--- a/internal/ls/hover.go
+++ b/internal/ls/hover.go
@@ -76,8 +76,7 @@ func (l *LanguageService) ProvideHover(ctx context.Context, params *lsproto.Hove
 		Range: &hoverRange,
 	}
 
-	// Only report canIncreaseVerbosity when the client explicitly requested a verbosity level.
-	if params.VerbosityLevel != nil {
+	if caps.TextDocument.Hover.VerbosityLevel {
 		hover.CanIncreaseVerbosity = vc.CanIncreaseVerbosity && !vc.Truncated
 	}
 
--- a/testdata/baselines/reference/fourslash/quickInfo/quickInfoDisplayPartsTypeParameterInFunction.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/quickInfoDisplayPartsTypeParameterInFunction.baseline
@@ -333,7 +333,8 @@
           "line": 4,
           "character": 33
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -387,7 +388,8 @@
           "line": 5,
           "character": 12
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
--- a/testdata/baselines/reference/fourslash/quickInfo/deprecatedInheritedJSDocOverload.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/deprecatedInheritedJSDocOverload.baseline
@@ -57,7 +57,8 @@
           "line": 22,
           "character": 11
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   }
 ]
--- a/testdata/baselines/reference/fourslash/quickInfo/jsDocTypedefQuickInfo1.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/jsDocTypedefQuickInfo1.baseline
@@ -65,7 +65,8 @@
           "line": 9,
           "character": 17
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
--- a/testdata/baselines/reference/fourslash/quickInfo/quickInfoCommentsClass.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/quickInfoCommentsClass.baseline
@@ -264,7 +264,8 @@
           "line": 1,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -291,7 +292,8 @@
           "line": 3,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -318,7 +320,8 @@
           "line": 3,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -345,7 +348,8 @@
           "line": 4,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -372,7 +376,8 @@
           "line": 4,
           "character": 13
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -399,7 +404,8 @@
           "line": 5,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -426,7 +432,8 @@
           "line": 10,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -453,7 +460,8 @@
           "line": 10,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -480,7 +488,8 @@
           "line": 11,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -507,7 +516,8 @@
           "line": 11,
           "character": 13
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -534,7 +544,8 @@
           "line": 13,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -561,7 +572,8 @@
           "line": 18,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -588,7 +600,8 @@
           "line": 18,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -615,7 +628,8 @@
           "line": 19,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -642,7 +656,8 @@
           "line": 19,
           "character": 13
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -669,7 +684,8 @@
           "line": 21,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -696,7 +712,8 @@
           "line": 24,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -723,7 +740,8 @@
           "line": 24,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -750,7 +768,8 @@
           "line": 25,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -777,7 +796,8 @@
           "line": 25,
           "character": 13
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -804,7 +824,8 @@
           "line": 27,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -831,7 +852,8 @@
           "line": 34,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -858,7 +880,8 @@
           "line": 34,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -885,7 +908,8 @@
           "line": 35,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -912,7 +936,8 @@
           "line": 35,
           "character": 13
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -939,7 +964,8 @@
           "line": 56,
           "character": 23
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   }
 ]
--- a/testdata/baselines/reference/fourslash/quickInfo/quickInfoCommentsClassMembers.baseline
+++ b/testdata/baselines/reference/fourslash/quickInfo/quickInfoCommentsClassMembers.baseline
@@ -803,7 +803,8 @@
           "line": 1,
           "character": 8
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -1154,7 +1155,8 @@
           "line": 31,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -1829,7 +1831,8 @@
           "line": 78,
           "character": 6
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -1856,7 +1859,8 @@
           "line": 78,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
@@ -2423,7 +2427,8 @@
           "line": 89,
           "character": 15
         }
-      }
+      },
+      "canIncreaseVerbosity": true
     }
   },
   {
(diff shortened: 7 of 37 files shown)


internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -492,6 +492,38 @@ function patchAndPreprocessModel() {
...
+
+        // Patch HoverClientCapabilities to add verbosityLevel support flag
+        if (structure.name === "HoverClientCapabilities") {
+            structure.properties.push({
+                name: "verbosityLevel",
+                type: { kind: "base", name: "boolean" },
+                optional: true,
+                documentation: "The client supports the `verbosityLevel` property on `HoverParams` and `canIncreaseVerbosity` on `Hover`.",
+            });
+        }
```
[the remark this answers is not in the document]
jakebailey: Yes, but obviously we can fix this whenever it happens

jakebailey:
One icky thing is that `symbolTableToDeclarationStatements` is basically what used to power JS declaration emit, but now does not; everything here seems to be requiredish but the unported code is all JS stuff. Makes me wonder if this feature just needs a rethink now that we aren't doing JS in the same way?

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -1983,12 +1983,51 @@ function parseFilterPredicate(arrow: ts.ArrowFunction): string | undefined {
...
-        // !!!
+    if (args.length === 0) {
+        return [{
+            kind: "verifyBaselineQuickInfo",
+        }];
+    }
+    // First arg is verbosityLevels: { markerName: number | number[] }
+    const verbosityArg = args[0];
+    if (!ts.isObjectLiteralExpression(verbosityArg)) {
         throw new Error(`verify.baselineQuickInfo arguments not supported`);
```
gabritto: Probably doesn't matter much, but this error message should be updated to say we expect an object literal expression.

internal/fourslash/_scripts/convertFourslash.mts
```diff
@@ -1983,12 +1983,51 @@ function parseFilterPredicate(arrow: ts.ArrowFunction): string | undefined {
...
     }
+    const verbosityLevels: Record<string, number[]> = {};
+    for (const prop of verbosityArg.properties) {
+        if (!ts.isPropertyAssignment(prop)) {
+            return undefined;
```
gabritto: I guess you'll get to it, but we need to throw an error now instead of returning `undefined`.

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -43,7 +43,6 @@ TestCommentsUnion
 TestCommentsVariables
 TestCompletionAfterQuestionDot
 TestCompletionAutoInsertQuestionDot
-TestCompletionBeforeSemanticDiagnosticsInArrowFunction1
```
gabritto, line 47: Why are those fixed now?
Edit: I see a bunch of small fixes in `hover.go` to print e.g. `(local class)`, type parameter contexts, const type parameters, etc.

internal/ls/completions.go
gabritto, line 5037: Are we now using this somewhere else?

internal/ls/hover.go
```diff
@@ -35,7 +39,62 @@ func (l *LanguageService) ProvideHover(ctx context.Context, documentURI lsproto.
...
+	}
+
+	if supportsVerbosity && out != nil {
+		hover.CanIncreaseVerbosity = out.CanIncreaseExpansionDepth && !out.Truncated
+	}
+
+	return lsproto.HoverOrNull{Hover: hover}, nil
+}
+
+func (l *LanguageService) ProvideVerboseHover(ctx context.Context, documentURI lsproto.DocumentUri, position lsproto.Position, verbosityLevel int) (lsproto.HoverResponse, error) {
```
gabritto: This seems to be the same as `ProvideHover`. Should we consolidate the two?

internal/checker/printer.go
```diff
@@ -381,6 +383,105 @@ func (c *Checker) TypeToTypeNode(t *Type, enclosingDeclaration *ast.Node, flags
...
+		p = createPrinterWithRemoveComments(nodeBuilder.EmitContext())
+	}
+	var sourceFile *ast.SourceFile
+	if enclosingDeclaration != nil {
+		sourceFile = ast.GetSourceFileOfNode(enclosingDeclaration)
+	}
+	p.Write(typeNode, sourceFile, writer, nil)
+	result := writer.String()
+
+	maxLength := defaultMaximumTruncationLength * 2
```
gabritto: We need to either accept a `maxLength` parameter here like we used to in Strada, or use `defaultHoverMaximumTruncationLength`, otherwise we'll be using `160` instead of the new default of `500` when printing types for hover.

internal/checker/printer.go
```diff
@@ -381,6 +383,105 @@ func (c *Checker) TypeToTypeNode(t *Type, enclosingDeclaration *ast.Node, flags
...
+		if out != nil {
+			out.Truncated = true
+		}
+		return result[0:maxLength-len("...")] + "..."
+	}
+	return result
+}
+
+// SignatureToStringWithVerbosity converts a signature to string with verbosity level support for expandable hover.
+func (c *Checker) SignatureToStringWithVerbosity(signature *Signature, enclosingDeclaration *ast.Node, flags TypeFormatFlags, verbosityLevel int, out *WriterContextOut) string {
```
gabritto: Could that and the function above be merged with their non-verbosity versions, e.g. `signatureToStringEx`? I think there aren't that many callers that it would be too much of a hassle to update it.

internal/checker/nodebuilder.go
```diff
@@ -37,6 +60,19 @@ func (b *NodeBuilder) enterContext(enclosingDeclaration *ast.Node, flags nodebui
 	b.impl.ctx.tracker = tracker
 }
 
+// propagateVerbosityOut copies expansion signals from the context to the VerbosityContext output.
+func (b *NodeBuilder) propagateVerbosityOut() {
+	if b.verbosity != nil && b.verbosity.Level > 0 {
```
gabritto: Again, why do we only do this if `Level` > 0? We could have `Level` be 0 and still need the outputs...

internal/checker/nodebuilder.go
```diff
@@ -18,12 +31,22 @@ func (b *NodeBuilder) EmitContext() *printer.EmitContext {
 }
 
 func (b *NodeBuilder) enterContext(enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) {
+	verbosityLevel := -1
+	maxTruncationLength := 0
+	if b.verbosity != nil && b.verbosity.Level > 0 {
```
gabritto: Why is this distinction necessary when setting `verbosityLevel`?

internal/checker/printer.go
```diff
@@ -381,6 +383,120 @@ func (c *Checker) TypeToTypeNode(t *Type, enclosingDeclaration *ast.Node, flags
 	return nodeBuilder.TypeToTypeNode(t, enclosingDeclaration, flags, nodebuilder.InternalFlagsNone, nil)
 }
 
+// TypeToStringWithVerbosity converts a type to string with verbosity support for expandable hover.
+func (c *Checker) TypeToStringWithVerbosity(t *Type, enclosingDeclaration *ast.Node, flags TypeFormatFlags, vc *VerbosityContext) string {
```
gabritto: Again, can't this and the other functions just be merged with e.g. `typeToStringEx`?

internal/ls/hover.go
```diff
@@ -16,16 +16,22 @@ import (
 )
 
 const (
-	symbolFormatFlags = checker.SymbolFormatFlagsWriteTypeParametersOrArguments | checker.SymbolFormatFlagsUseOnlyExternalAliasing | checker.SymbolFormatFlagsAllowAnyNodeKind | checker.SymbolFormatFlagsUseAliasDefinedOutsideCurrentScope
-	typeFormatFlags   = checker.TypeFormatFlagsUseAliasDefinedOutsideCurrentScope
+	symbolFormatFlags                   = checker.SymbolFormatFlagsWriteTypeParametersOrArguments | checker.SymbolFormatFlagsUseOnlyExternalAliasing | checker.SymbolFormatFlagsAllowAnyNodeKind | checker.SymbolFormatFlagsUseAliasDefinedOutsideCurrentScope
+	typeFormatFlags                     = checker.TypeFormatFlagsUseAliasDefinedOutsideCurrentScope
+	defaultHoverMaximumTruncationLength = 500
```
gabritto: I'm realizing we don't need this and should instead use `MaximumHoverLength` (which already uses 500 as default) from the user preferences. We also need to respect that setting even if the client doesn't support expandable hovers.

internal/checker/nodebuilderimpl.go
```diff
@@ -58,12 +58,17 @@ type NodeBuilderContext struct {
 	host                            Host
 	tracker                         nodebuilder.SymbolTracker
 	approximateLength               int
```
gabritto, line 60: I think we're missing a bunch of the statements I added in [Strada](https://github.com/microsoft/TypeScript/pull/61492) that add to `approximateLength`, e.g. `createExpressionFromSymbolChain` seems to be missing it (I haven't checked the others but probably there's more missing too)

internal/checker/nodebuilder_hover.go
```diff
@@ -0,0 +1,635 @@
...
+			b.ctx.expansionTruncated = true
+			members = append(members, b.f.NewEnumMember(b.f.NewStringLiteral(fmt.Sprintf(" ... %d more ... ", len(memberProps)-i-1), 0), nil))
+			last := memberProps[len(memberProps)-1]
+			members = append(members, b.f.NewEnumMember(b.f.NewIdentifier(last.Name), b.enumMemberInitializer(last)))
+			break
+		}
+		memberDecl := core.Find(p.Declarations, ast.IsEnumMember)
+		var initializer *ast.Node
+		if memberDecl != nil && memberDecl.AsEnumMember().Initializer != nil {
+			initializer = b.f.DeepCloneNode(memberDecl.AsEnumMember().Initializer)
```
gabritto, line 69: We probably also want to add to `approximateLength` more consistently in the functions here, e.g. include initializers

internal/checker/nodebuilder_hover.go
```diff
@@ -0,0 +1,635 @@
...
+		}
+	}
+
+	return b.f.NewInterfaceDeclaration(nil, b.f.NewIdentifier(name), b.f.NewNodeList(typeParamDecls), b.f.NewNodeList(heritageClauses), b.f.NewNodeList(members))
+}
+
+// serializePropertiesWithTruncation iterates properties using addPropertyToElementList,
+// with truncation checks matching Strada's createTypeNodesFromResolvedType behavior.
+func (b *NodeBuilderImpl) serializePropertiesWithTruncation(properties []*ast.Symbol, elements []*ast.Node) []*ast.Node {
+	if isExpanding(b.ctx) {
```
gabritto: Shouldn't we always do the prototype filtering given this code is only ever called for hover?

internal/checker/printer.go
jakebailey: We could, yeah; was waffling between this just because it means updating callers who totally do not care about verbosity

internal/checker/nodebuilderimpl.go
```diff
@@ -190,8 +190,9 @@ func (b *NodeBuilderImpl) shouldExpandType(t *Type, isAlias bool) bool {
 }
 
 // canPossiblyExpandType returns true if it's possible the type or one of its components can be expanded.
+// At maxExpansionDepth <= 0, no expansion will occur so we return false to allow type-node reuse.
 func (b *NodeBuilderImpl) canPossiblyExpandType(t *Type) bool {
-	if b.ctx.maxExpansionDepth < 0 {
+	if b.ctx.maxExpansionDepth <= 0 {
```
gabritto: Why did this change? Couldn't we still fall into the case where maxDepth is 0 and depth is 0 and `canIncreaseExpansionDepth` is false and so this would return true before?

jakebailey: This is the bit that makes it so that simply enabling the client capability / sending level 0 does not change the output formatting. But otherwise I'm not sure what you mean

gabritto: Yeah, but IIRC if we have a request with level set to 0, and we make `canPossiblyExpandType` return `false`, we will do type node reuse and not compute `canIncreaseVerbosityLevel`.

jakebailey: Ah, yeah, we have to still do a little work I guess so this is set. I guess I didn't notice because the boolean is hidden if false, but obviously fourslash will keep querying with higher levels. Probably I should assert if we attempt to go higher but the previous response said no... maybe.

internal/fourslash/fourslash.go
```diff
@@ -2095,7 +2109,7 @@ func (f *FourslashTest) VerifyBaselineHoverWithVerbosity(t *testing.T, verbosity
 		if !ok {
 			levels = []int{0}
 		}
-		for _, level := range levels {
+		for i, level := range levels {
```
gabritto, line 2466: I wonder if a better way to test this would be to just make the tests allow you to specify a maximum verbosity level, and then make fourslash check if every level from 0 to that can be expanded, and that it cannot be further expanded.

testdata/baselines/reference/fourslash/quickInfo/quickInfoDisplayPartsTypeParameterInFunction.baseline
```diff
@@ -57,29 +57,29 @@
...
+// | (type parameter) U extends string in foo2<U extends string>(a: U): U
 // | ```
 // | 
 // | ----------------------------------------------------------------------
 //                                 ^
 // | ----------------------------------------------------------------------
 // | ```tsx
-// | (parameter) a: U
+// | (parameter) a: U extends string
 // | ```
```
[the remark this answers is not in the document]
gabritto: Do we like this change?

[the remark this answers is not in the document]
jakebailey: I can figure out what's going on here, I forgot about these.

internal/checker/nodebuilderimpl.go
```diff
@@ -175,73 +176,83 @@ func (b *NodeBuilderImpl) isExpandableType(t *Type, isAlias bool) bool {
...
 	}
-	// At the boundary: signal that more expansion is possible
 	b.ctx.canIncreaseExpansionDepth = true
 	return false
 }
 
-// canPossiblyExpandType returns true if it's possible the type or one of its components can be expanded.
-// At maxExpansionDepth <= 0, no expansion will occur so we return false to allow type-node reuse.
+// canPossiblyExpandType reports whether we are actively expanding types (depth < maxExpansionDepth),
+// meaning type-node reuse should be skipped so typeToTypeNode can expand named types.
```
gabritto, line 226: Why do we need this function now? IIRC the difference between `canPossiblyExpandType` and `shouldExpandType` is that `canPossiblyExpandType` was meant to be called to see if we should avoid node reuse, but now we're not really avoiding node reuse anyways, so I don't quite get why not always use `shouldExpandType`.

internal/checker/nodecopy.go
```diff
@@ -46,13 +47,13 @@ func (b *NodeBuilderImpl) reuseTypeNode(node *ast.Node) *ast.Node {
...
-// for expandability. Short-circuits once canIncreaseExpansionDepth is set.
+// walkNodeForExpandability walks a reused AST node tree, calling checkTypeExpandability
+// on each type reference, type predicate, or import type node.
+// Short-circuits once canIncreaseExpansionDepth is set.
 func (b *NodeBuilderImpl) walkNodeForExpandability(node *ast.Node) {
 	if b.ctx.canIncreaseExpansionDepth || node == nil {
 		return
 	}
-	// Only resolve types for nodes that can reference expandable named types.
 	if ast.IsTypeReferenceNode(node) || ast.IsExpressionWithTypeArguments(node) || ast.IsTypePredicateNode(node) || ast.IsImportTypeNode(node) {
```
gabritto, line 58: Aren't there way more kinds of nodes where expansion could happen? e.g. if the node is for `{ a: SomeType }`, we could expand `SomeType`...

internal/checker/nodebuilderimpl.go
```diff
@@ -138,10 +143,126 @@ func (b *NodeBuilderImpl) checkTruncationLength() bool {
...
+		return true
+	}
+	return false
+}
+
+// isExpandableType reports whether t has a named representation that could be inlined
+// as its structural form during hover expansion. Filters out lib types.
+// When isAlias is true, checks whether t's alias symbol is from user code (not lib).
+func (b *NodeBuilderImpl) isExpandableType(t *Type, isAlias bool) bool {
+	if isAlias {
```
gabritto, line 172: I can't really picture it in my head how this reflects the behavior of walking through a type and seeing if any of its parts are expandable. For instance, what if the type we're checking is an object literal type with a property whose type is an alias? How would that work?

internal/checker/nodecopy.go
jakebailey: This is specifically for things with type params that may be printed even though they are not actually in the tree; for those other cases we'll descend below.

internal/fourslash/fourslash.go
jakebailey: Yeah, I just didn't want to make every test manual, so this was my workaround while debugging these. 😄

internal/checker/nodecopy.go
gabritto: Could you leave a comment then? It's kinda hard to see this from just the surrounding code.

internal/checker/nodebuilderimpl.go
jakebailey: We don't; it actually should have just been reduced to being a depth check. Refactored it again.

internal/checker/nodebuilderimpl.go
jakebailey: Basically, it doesn't; this is only surface level, and the type traversal is in the caller during printing.

gabritto:
So the new algorithm seems to only really work because of this line here: https://github.com/microsoft/typescript-go/blob/main/internal/checker/nodecopy.go#L366
Meaning right now we will correctly compute "canIncreaseVerbosity" because we will fall out of type node reuse when we visit a type reference, because the type references we're looking at (at least in tests) don't have a resolved symbol. Is there some guarantee that this will always be the case for the type references we could expand?

jakebailey:
If it's `nil`, then we fall through to `typeToTypeNode`, and things work.

If it's not `nil`, then we should be going through `walkNodeForExpandability` on that reused node.

So, I think it's fine? Maybe I'm not thinking of the case you're thinking of.

gabritto:
> If it's `nil`, then we fall through to `typeToTypeNode`, and things work.
> 
> If it's not `nil`, then we should be going through `walkNodeForExpandability` on that reused node.
> 
> So, I think it's fine? Maybe I'm not thinking of the case you're thinking of.

Ok, I think I finally see how it works.

PR #3330, opened 2026-04-02
Fix --showConfig output: add compileOnSave and implied compiler options

aamoghS:
#3326 

trying to fix this issue


aamoghS:

PR #3333, opened 2026-04-02
Activate extension on typescript.native-preview.enable

### LINKED ISSUE #3312: latest release [0.20260401.1] breaks LSP on VS code
### Extension Version

 0.20260401.1

### VS Code Version

1.113.0

### Operating system Version

Linux 64

### Steps to reproduce

1. install the extension
2. enable the extenstion 

### Issue

extension fail to load with message `typescript.native-preview-enable` not found

Image

jakebailey:
I think this fixes #3312.

### DIFF
--- a/_extension/package.json
+++ b/_extension/package.json
@@ -24,7 +24,8 @@
         "onLanguage:javascript",
         "onLanguage:javascriptreact",
         "onLanguage:typescript",
-        "onLanguage:typescriptreact"
+        "onLanguage:typescriptreact",
+        "onCommand:typescript.native-preview.enable"
     ],
     "contributes": {
         "configuration": [


_extension/package.json
```diff
@@ -24,7 +24,8 @@
         "onLanguage:javascript",
         "onLanguage:javascriptreact",
         "onLanguage:typescript",
-        "onLanguage:typescriptreact"
+        "onLanguage:typescriptreact",
+        "onCommand:typescript.native-preview.enable"
```
mjbvz, line 28: Since you contribute the command, your extension should already be activated automatically. However there is an `enablement` field for that command. Could that be false for the user?

jakebailey: No, since the enablement depends on something we explicitly set on activate ☹️

PR #3334, opened 2026-04-03
Implement multi-file doc highlights

jakebailey:
Uses the same kind of idea as #3311, except that this is a separate method in VS Code, so it is in our LSP extension.


internal/ls/documenthighlights.go
```diff
@@ -49,6 +50,63 @@ func (l *LanguageService) ProvideDocumentHighlights(ctx context.Context, documen
...
+	// Resolve the source files to search
+	var sourceFiles []*ast.SourceFile
+	for _, uri := range filesToSearch {
+		if sf := program.GetSourceFile(uri.FileName()); sf != nil {
+			sourceFiles = append(sourceFiles, sf)
+		}
+	}
+	if len(sourceFiles) == 0 {
+		sourceFiles = []*ast.SourceFile{sourceFile}
+	}
```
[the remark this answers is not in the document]
jakebailey: Strada did not do this, so I didn't either.

internal/ls/documenthighlights.go
```diff
@@ -49,6 +51,68 @@ func (l *LanguageService) ProvideDocumentHighlights(ctx context.Context, documen
...
+		}
+		if sf := program.GetSourceFile(fileName); sf != nil {
+			sourceFiles = append(sourceFiles, sf)
+		}
+	}
+	if len(sourceFiles) == 0 {
+		sourceFiles = []*ast.SourceFile{sourceFile}
+	}
+
+	if node.Parent != nil && (node.Parent.Kind == ast.KindJsxClosingElement || (node.Parent.Kind == ast.KindJsxOpeningElement && node.Parent.TagName() == node)) {
```
DanielRosenwasser, line 45: It seems weird to me that this is not just part of some shared code path between single-file and multi-file. Why is single-file not just a special case of multi-file search anyway?

DanielRosenwasser: Also, shouldn't this cheaper checks come before deduplicating files etc.?

internal/ls/documenthighlights.go
```diff
@@ -49,6 +51,68 @@ func (l *LanguageService) ProvideDocumentHighlights(ctx context.Context, documen
...
+		if !seenFiles.AddIfAbsent(fileName) {
+			continue
+		}
+		if sf := program.GetSourceFile(fileName); sf != nil {
+			sourceFiles = append(sourceFiles, sf)
+		}
+	}
+	if len(sourceFiles) == 0 {
+		sourceFiles = []*ast.SourceFile{sourceFile}
+	}
```
DanielRosenwasser, line 87: I don't know if it matters much, but both `getReferencedSymbolsForNode` and `ProvideMultiDocumentHighlights` create a set for the files. But also, it feels weird that we don't just have a helper for this pattern.

internal/ls/documenthighlights.go
```diff
@@ -68,6 +132,42 @@ func (l *LanguageService) getSemanticDocumentHighlights(ctx context.Context, pos
...
+	referenceEntries := l.getReferencedSymbolsForNode(ctx, position, node, program, sourceFiles, options)
+	if referenceEntries == nil {
+		return nil
+	}
+
+	// Build a set of allowed file names for quick lookup
+	allowedFiles := make(map[string]bool, len(sourceFiles))
+	for _, sf := range sourceFiles {
+		allowedFiles[sf.FileName()] = true
+	}
```
DanielRosenwasser: `Set[string]`?

internal/ls/documenthighlights.go
jakebailey: That's true, yeah. I think it can be simplified.

PR #3304, opened 2026-03-31
Add isolatedDeclarations quickfix

jakebailey:
Builds on #3303.

Had to port more fourslash generator code, but filtered down to just these fixes.

PR #3382, opened 2026-04-09
Support LSP `source.fixAll`

jakebailey:
This adds `source.fixAll` support. For example, running the "Fix All" command in VS Code nets:

```
2026-04-09 15:10:58.188 [trace] Sending request 'textDocument/codeAction - (11)'.
Params: {
    "textDocument": {
        "uri": "file:///home/jabaile/work/vscode/extensions/typescript-language-features/src/lazyClientHost.ts"
    },
    "range": {
        "start": {
            "line": 33,
            "character": 17
        },
        "end": {
            "line": 33,
            "character": 17
        }
    },
    "context": {
        "diagnostics": [],
        "only": [
            "source.fixAll"
        ],
        "triggerKind": 1
    }
}
```
```
2026-04-09 15:10:58.474 [trace] Received response 'textDocument/codeAction - (11)' in 286ms.
Result: [
    {
        "title": "Fix All",
        "kind": "source.fixAll",
        "edit": {
            "changes": {
                "file:///home/jabaile/work/vscode/extensions/typescript-language-features/src/lazyClientHost.ts": [
                    {
                        "range": {
                            "start": {
                                "line": 19,
                                "character": 0
                            },
                            "end": {
                                "line": 19,
                                "character": 0
                            }
                        },
                        "newText": "import { ActiveJsTsEditorTracker } from './ui/activeJsTsEditorTracker';\nimport { Lazy } from './utils/lazy';\n"
                    }
                ]
            }
        }
    }
]
```

This involved porting more fourslash code; mostly code stolen from #3304. But, it required more to also handle imports as we now have a test which breaks the import elision heuristic in the converter. Yay.


internal/ls/codeactions.go
```diff
@@ -110,6 +122,78 @@ func (l *LanguageService) ProvideCodeActions(ctx context.Context, params *lsprot
...
+			Program:    program,
+			LS:         l,
+		}
+
+		combined, err := provider.GetAllCodeActions(ctx, fixContext)
+		if err != nil {
+			return nil, err
+		}
+		if combined != nil && len(combined.Changes) > 0 {
+			lspChanges[uri] = append(lspChanges[uri], combined.Changes...)
```
gabritto, line 258: Just to double check, the code fixes can't really return changes to other URIs, can they?

internal/fourslash/fourslash.go
```diff
@@ -1378,6 +1378,233 @@ func assertDeepEqual(t *testing.T, actual any, expected any, prefix string, opts
...
+type VerifyCodeFixOptions struct {
+	Description    string
+	NewFileContent string
+	Index          int
+	ApplyChanges   bool
+}
+
+// VerifyCodeFixAllOptions are the options for VerifyCodeFixAll.
+type VerifyCodeFixAllOptions struct {
+	FixID          string
```
gabritto, line 1391: Is the purpose of this just to document what fix we expect to happen?

gabritto (approved):
Had some questions but this looks good

internal/ls/codeactions.go
jakebailey: They can! The response is `WorkspaceEdit`.

internal/fourslash/fourslash.go
jakebailey: You're right to ask about this, I missed the part of the call that prechecks to see if a codefix is even available for this ID.

jakebailey: It's awkward because the old API used `getCombinedCodeFix` which wanted a fixid, but, now that's not how the LSP works at all. Probably I should just delete this because we no longer have the client orchestrate all of this

internal/ls/codeactions.go
gabritto: But then we'd only add the edits in the current file? Why?

internal/fourslash/fourslash.go
jakebailey: Updated the PR, so now what we do is still support LSP "fix all", but then, we offer up fixes like VS Code's TS extension use to per-category, restoring the fixall description.

internal/ls/codeactions.go
jakebailey: Oh.... yeah

jakebailey: Oh, sorry, no, the fixes we return are still only one file; we don't handle multi-file stuff below this yet, so it's (for now) "safe" to assume the edits are just in the current file. (I misunderstood the question)

gabritto (approved):

PR #3390, opened 2026-04-11
Coerce factory modifiers argument to NodeArray

### LINKED ISSUE #3389: Factory statement and type declaration nodes do not emit export modifier
Assuming this even is the correct way to define a type export, factory-made nodes do not emit the modifier since #3367.

UPD: looks like the issue is `createNodeArray` not being called on `modifiers`.

```typescript
test("createTypeAliasDeclaration export", async () => {
    const api = spawnAPI();
    try {
        
        const snapshot = await api.updateSnapshot({ openProject: "/tsconfig.json" });
        const project = snapshot.getProject("/tsconfig.json")!;
        const node = createTypeAliasDeclaration(
            [createToken(SyntaxKind.ExportKeyword)],
            createIdentifier('Test'),
            undefined,
            createKeywordTypeNode(SyntaxKind.AnyKeyword)
        );
        
        // this fails
        assert.equal(await project.emitter.printNode(node), 'export type Test = any;');
        
        // this works
        const cloned = getSynthesizedDeepClone(node);
        assert.equal(await project.emitter.printNode(cloned), 'export type Test = any;');
        
    }
    finally {
        await api.close();
    }
});
```

virtulis:
Fixes #3389

### DIFF
--- a/_packages/api/test/async/api.test.ts
+++ b/_packages/api/test/async/api.test.ts
@@ -43,6 +43,8 @@ import {
     createIdentifier,
     createKeywordTypeNode,
     createParameterDeclaration,
+    createToken,
+    createTypeAliasDeclaration,
     createTypeReferenceNode,
     createUnionTypeNode,
 } from "@typescript/ast/factory";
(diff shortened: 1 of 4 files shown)


andrewbranch (approved):
Good catch, thanks!

PR #3391, opened 2026-04-12
Fix inference from unions of arrays with different nesting depths

### LINKED ISSUE #1789: Inference fails form unions of arrays of different depths (T[] | T[][])
## Steps to reproduce

```ts
declare function flat<T>(args: T[] | T[][]): void;
type Value = 1 | 2
declare const n: Value[] | Value[][]
flat(n)
```

Type checks, `T` is inferred as `Value`

## Behavior with `tsgo`

We get an error in `flat(n)`:

```txt
error TS2345: Argument of type 'Value[][] | Value[]' is not assignable to parameter of type 'Value[][][] | Value[][]'.
  Type 'Value[]' is not assignable to type 'Value[][][] | Value[][]'.
    Type 'Value[]' is not assignable to type 'Value[][][]'.
      Type 'number' is not assignable to type 'Value[][]'.
        Type 'number' is not assignable to type 'Value[][]'.
```

Yiin:
Fixes #1789, fixes #3370.

When inferring `T` from a parameter typed `T[] | T[][]`, the closely-matched inference pass cross-matches all `Array` constituents regardless of nesting depth. `CompareTypes` sorts union constituents by type flags (`Object` < `Union`), so source types like `Value[][]` sort before `Value[]` when `Value` is a union or non-reference object type. The first cross-match (`Value[][]` to `T[]`) produces the wrong candidate `T = Value[]`, and `findLeftmostType` picks it.

This PR adds an intermediate "deeply matched" pass between the existing `isTypeOrBaseIdenticalTo` and `isTypeCloselyMatchedBy` passes. The new pass matches reference types only when their type arguments have compatible nesting structure - if a source arg is a nested reference to the same outer type (e.g., `Array<Array<...>>` inside `Array`), the target arg must also be nested, and vice versa. Same-depth pairs are inferred first, preventing the incorrect cross-match.

```ts
declare function flat<T>(args: T[] | T[][]): T;
type Value = 1 | 2;
declare const n: Value[] | Value[][];

// Before: T inferred as Value[] (error)
// After:  T inferred as Value   (correct)
flat(n);
```

### Notes

- The ordering bug also exists in TypeScript when `stableTypeOrdering` is enabled (the default since it was added in Feb 2026), but no upstream test covers the `T[] | T[][]` pattern with non-primitive type arguments.
- The extra matching pass is O(sources × targets) with cheap flag/pointer checks. Union types have few constituents in practice, and pairs consumed by the deep pass are removed before the close pass, so total `inferFromTypes` calls don't increase.
- All local, submodule, and checker tests pass with no regressions.


Yiin:

jakebailey:
@typescript-bot test it

PR #3397, opened 2026-04-13
Use GHA runners instead of custom runners

jakebailey:
Testing to see if it's even worth it to use custom runners.

Before:

image

These are the jobs on custom runners:

- `race`: 11m25s
- `release-build`: 10m59s

After:

image

So:

- `race`: 20m35s
- `release-build`: 17m17s

So, something like twice as slow.

But, `windows-latest` already takes nearly 19 minutes. So, why bother having these runners?


RyanCavanaugh (approved):
Either way

PR #3401, opened 2026-04-14
Fix file counts, jsx in project telemetry

jakebailey:
We need to count files based on what their file says they are. This doesn't fix the missing JS thing but is a problem.

Then, fix the JSX value to send a string.

### DIFF
--- a/internal/project/session.go
+++ b/internal/project/session.go
@@ -706,7 +706,7 @@ func (s *Session) collectProjectInfoTelemetry(project *Project) lsproto.Telemetr
 		compilerOptions["moduleResolution"] = name
 	}
 	if opts.Jsx != core.JsxEmitNone {
-		compilerOptions["jsx"] = fmt.Sprintf("%d", opts.Jsx)
+		compilerOptions["jsx"] = jsxEmitName(opts.Jsx)
 	}
 	if b, err := json.Marshal(compilerOptions); err == nil {
 		props["compilerOptions"] = string(b)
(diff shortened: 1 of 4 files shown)


internal/project/session.go
```diff
@@ -746,17 +746,16 @@ func boolTelemetry(v bool) string {
 func countFileStats(sourceFiles []*ast.SourceFile) *lsproto.ProjectInfoTelemetryMeasurements {
 	var stats lsproto.ProjectInfoTelemetryMeasurements
 	for _, sf := range sourceFiles {
-		fileName := sf.FileName()
 		size := float64(sf.End())
-		switch core.GetScriptKindFromFileName(fileName) {
```
DanielRosenwasser, line 751: Why didn't this work before?

DanielRosenwasser:
> This doesn't fix the missing JS thing but is a problem.

Okay, so which part of file counting *does* this fix?

internal/project/session.go
```diff
@@ -790,6 +789,23 @@ func moduleResolutionKindName(kind core.ModuleResolutionKind) string {
 	}
 }
 
+func jsxEmitName(jsx core.JsxEmit) string {
```
DanielRosenwasser: There's also `InverseJsxOptionMap` if you want to use that.

internal/project/session.go
```diff
@@ -790,6 +789,23 @@ func moduleResolutionKindName(kind core.ModuleResolutionKind) string {
...
+	case core.JsxEmitReactNative:
+		return "react-native"
+	case core.JsxEmitReact:
+		return "react"
+	case core.JsxEmitReactJSX:
+		return "react-jsx"
+	case core.JsxEmitReactJSXDev:
+		return "react-jsxdev"
+	default:
+		return ""
```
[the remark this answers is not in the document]
jakebailey: I don't really want to send unknown data. I guess it's a number, though? But, we cannot possibly parse it out.

internal/project/session.go
jakebailey: This is sort of wonky but you need to use the SourceFile because it represents what the LSP says, e.g. the type of an untitled file, or, if someone changes their file type from .js to .jsx in the UI only but not the extension.

jakebailey:
> > This doesn't fix the missing JS thing but is a problem.
> 
> Okay, so which part of file counting _does_ this fix?

Nowhere, I have no idea what's wrong with that at the moment, these are just the handful of things I did notice were wrong.

RyanCavanaugh (approved):

PR #3403, opened 2026-04-15
Exposing more endpoints

navya9singh:


internal/api/session.go
```diff
@@ -1254,6 +1288,109 @@ func (s *Session) handleGetBaseTypeOfLiteralType(ctx context.Context, params *Ge
...
+		return nil, err
+	}
+
+	t := setup.checker.GetTypeParameterAtPosition(sig, int(params.Index))
+	if t == nil {
+		return nil, nil
+	}
+
+	return setup.sd.registerType(t), nil
+}
```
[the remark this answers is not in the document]
navya9singh: Name is same as that in Strada. The function returns the type of parameter as an index and not a generic type parameter.

internal/api/session.go
```diff
@@ -997,6 +1011,26 @@ func (s *Session) handleGetSignaturesOfType(ctx context.Context, params *GetSign
...
+
+	node, err := s.resolveNodeHandle(setup.program, params.Location)
+	if err != nil {
+		return nil, err
+	}
+	if node == nil {
+		return nil, nil
+	}
+
+	sig := setup.checker.GetResolvedSignature(node)
```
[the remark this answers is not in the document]
navya9singh: nil check for `sig` is done in `registerSignature`

internal/api/session.go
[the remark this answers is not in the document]
andrewbranch: No, Copilot is right about this. The function you wrote here returns the type of a generic type parameter. The checker function you need to use is called `getTypeAtPosition` and isn't exported yet.

_packages/native-preview/src/enums/nodeBuilderFlags.ts
```diff
@@ -0,0 +1,36 @@
+// Code generated by Herebyfile.mjs generate:enums from internal/nodebuilder/types.go. DO NOT EDIT.
```
andrewbranch, line 1: I don't see a change to Herebyfile; where did this enum come from?

internal/api/proto.go
```diff
@@ -704,13 +728,92 @@ type GetBaseTypeOfLiteralTypeParams struct {
...
+// Returns compact NodeReferenceInfo entries for each node-backed reference.
+type GetReferencedSymbolsForNodeParams struct {
+	Snapshot Handle[project.Snapshot] `json:"snapshot"`
+	Project  Handle[project.Project]  `json:"project"`
+	Node     Handle[ast.Node]         `json:"node"`
+	Position int                      `json:"position"`
+}
+
+// NodeReferenceInfo is a compact representation of a node-backed reference entry suitable for sending over the wire.
+type NodeReferenceInfo struct {
```
andrewbranch: This type seems incredibly specialized to a single use case despite a very generic name. I think this should just be replaced by `Handle[ast.Node]`.

internal/api/session.go
```diff
@@ -1893,3 +2263,23 @@ func (s *Session) resolveOptionalSourceFile(program *compiler.Program, file *Doc
 	}
 	return sourceFile, nil
 }
+
+// handleProbablyUsesSemicolons checks whether a source file probably uses semicolons.
+func (s *Session) handleProbablyUsesSemicolons(ctx context.Context, params *GetSourceFileParams) (bool, error) {
```
andrewbranch: This seems like something that doesn't need to be answered over the wire.

internal/api/session.go
```diff
@@ -1254,6 +1311,281 @@ func (s *Session) handleGetBaseTypeOfLiteralType(ctx context.Context, params *Ge
...
+
+			result = append(result, info)
+		}
+	}
+	return result, nil
+}
+
+// handleSomeSignatureUsage checks if any usage of a signature has more arguments than the given parameter index,
+// or if any reference is a non-call usage. Ports Strada's FindAllReferences.Core.someSignatureUsage.
+func (s *Session) handleSomeSignatureUsage(ctx context.Context, params *SomeSignatureUsageParams) (bool, error) {
```
andrewbranch: This name doesn't make sense. The Strada function was so named because it took a predicate function that was called for each signature and returned a boolean the first time the predicate returns true, like `Array.prototype.some`. You can't pass a predicate function over the wire, so I assume you've baked in the particular predicate function you needed, but consequently the method is a misnomer and the implementation is so specific that no other API consumer could use it. Instead, you should just be using the API to get all the signature usages and then doing whatever analysis on the client side.

internal/api/session.go
```diff
@@ -1254,6 +1311,281 @@ func (s *Session) handleGetBaseTypeOfLiteralType(ctx context.Context, params *Ge
...
+	sourceFile := setup.program.GetSourceFileByPath(path)
+	if sourceFile == nil {
+		return false, fmt.Errorf("%w: source file not found: %s", ErrClientError, path)
+	}
+
+	return setup.checker.IsSymbolReferencedInFile(sourceFile, node.AsIdentifier(), symbol), nil
+}
+
+// handleEachSymbolReferenceInFile returns positions of all references to a symbol in a file.
+func (s *Session) handleEachSymbolReferenceInFile(ctx context.Context, params *IsSymbolReferencedInFileParams) ([]int, error) {
```
andrewbranch: Similar issue; the name is no longer appropriate for the implementation because the Strada code took a callback so the name was evoking `Array.prototype.forEach`. This should be `GetSymbolReferencesInFile` and return a list of node handles to identifiers.

internal/api/session.go
```diff
@@ -1254,6 +1311,281 @@ func (s *Session) handleGetBaseTypeOfLiteralType(ctx context.Context, params *Ge
...
+	symbol, err := setup.sd.resolveSymbolHandle(params.Symbol)
+	if err != nil {
+		return false, err
+	}
+	if symbol == nil {
+		return false, nil
+	}
+
+	// Get the source file from the node handle
+	_, _, _, path, err := parseNodeHandle(params.Definition)
```
andrewbranch: It looks like this decoding is duplicated with `node, err := s.resolveNodeHandle(setup.program, params.Definition)` above

internal/api/session.go
```diff
@@ -275,6 +277,17 @@ type checkerSetup struct {
 	done    func()
 }
 
+// setupLanguageService creates a LanguageService from the same snapshot/project used by setupChecker.
+// This ensures the LS operates on the same program and state as the checker.
+func (s *Session) setupLanguageService(setup checkerSetup, projectHandle Handle[project.Project], activeFile string) (*ls.LanguageService, error) {
+	projectName := parseProjectHandle(projectHandle)
+	proj := setup.sd.snapshot.ProjectCollection.GetProjectByPath(projectName)
```
andrewbranch: This shouldn't take a `checkerSetup` because it has to duplicate getting the project and program.

internal/api/proto.go
```diff
@@ -704,13 +728,92 @@ type GetBaseTypeOfLiteralTypeParams struct {
...
+	Snapshot Handle[project.Snapshot] `json:"snapshot"`
+	Project  Handle[project.Project]  `json:"project"`
+	Type     Handle[checker.Type]     `json:"type"`
+}
+
+// IsSymbolReferencedInFileParams checks whether a symbol is referenced in the file containing the given identifier.
+type IsSymbolReferencedInFileParams struct {
+	Snapshot   Handle[project.Snapshot] `json:"snapshot"`
+	Project    Handle[project.Project]  `json:"project"`
+	Definition Handle[ast.Node]         `json:"definition"`
```
andrewbranch: I don't think "definition" makes sense as a parameter to answer the question "is a symbol referenced in a file"—I would expect it to take a symbol and a file. It looks like Strada wants a "definition" identifier because the comment says "Used as a quick check for whether a symbol is used at all in a file (besides its definition)." But the Strada API also doesn't take a symbol as an input; it just takes the identifier. So the original name is bad, but the contract is basically "give me the identifier of a declaration and I'll tell you if it's used anywhere _else_ in the same file," which is fairly coherent. That said, it just uses `eachSymbolReferenceInFile` which I've suggested use expose as `GetSymbolReferencesInFile`, so I'm not sure it's worth it to expose this separate method.

internal/api/session.go
```diff
@@ -1254,6 +1311,281 @@ func (s *Session) handleGetBaseTypeOfLiteralType(ctx context.Context, params *Ge
...
+
+	positions := setup.checker.EachSymbolReferenceInFile(sourceFile, node.AsIdentifier(), symbol)
+	if positions == nil {
+		return []int{}, nil
+	}
+	return positions, nil
+}
+
+// handleGetReferencedSymbolsForNode finds all references to a node and returns compact NodeReferenceInfo entries.
+func (s *Session) handleGetReferencedSymbolsForNode(ctx context.Context, params *GetReferencedSymbolsForNodeParams) ([]NodeReferenceInfo, error) {
```
andrewbranch: I already suggested you delete the `NodeReferenceInfo` type, but now I'm seeing that it's only used on a method called `GetReferencedSymbolsForNode`, but the return type isn't even anything close to a symbol. I'm totally confused by this.

andrewbranch (changes requested):
All the find-all-reference related APIs need to be significantly reworked in order to be useful outside of your specific use case.

_packages/native-preview/src/enums/nodeBuilderFlags.ts
navya9singh: Updated that and re-generated

internal/api/proto.go
navya9singh: I'll address all find-all-references api comments in a separate PR

internal/api/session.go
```diff
@@ -1349,6 +1490,44 @@ func (s *Session) handleTypeToTypeNode(ctx context.Context, params *TypeToTypeNo
...
+	}
+	defer setup.done()
+
+	sig, err := setup.sd.resolveSignatureHandle(params.Signature)
+	if err != nil {
+		return nil, err
+	}
+
+	var enclosingDeclaration *ast.Node
+	if params.Location != "" {
```
andrewbranch, line 1506: Nit: the param should be called `EnclosingDeclaration`

andrewbranch (approved):

internal/api/session.go
navya9singh: I'll fix this in the next one

PR #3409, opened 2026-04-15
Fix nil-config project crash during project tree loading

Andarist:
I found this crash when trying to repro https://github.com/microsoft/typescript-go/issues/3386#issuecomment-4227338467  . That said, the scenario in the added test isn't the same as the one in that replay session.


gabritto (approved):

gabritto:
I think this should fix the reported fuzzer crash: we really shouldn't be updating a program when the command line is `nil`, or keeping a project in that state. I think a project has a `nil` command line only temporarily, between us creating it for the first time and later when we go and update its program.
But what bugs me is that I don't see how either the ancestor deleted scenario, or the other scenario where a command line is `nil` because of bad `tsconfig.json`, could have happened in the fuzzer, since we don't delete anything, and the project where this happens doesn't seem to have tsconfig issues, so that part is intriguing.

PR #3412, opened 2026-04-15
fix(3410): preserve type arguments in hover for qualified names

### LINKED ISSUE #3410: Type serializer omits type arguments of outer functions
## Steps to reproduce

Hover on `box` in the following example:

```ts
function makeBox<T>(x: T) {
    class Box {
        value = x
    }
    return new Box()
}

let box = makeBox("hello")
```

Hovering on `box` displays:

```ts
let box: makeBox<string>.Box
```

## Behavior with `tsgo`

Hovering on `box` displays:

```ts
let box: makeBox.Box
```

Notice the missing type arguments. This similarly affects baselines in our test suites.

a-tarasyuk:
Fixes #3410


weswigham (commented):
This is *technically* an abuse of the AST - `ExpressionWithTypeArguments` is only naturally parsed in the context of class `extends` clauses, however it's probably fine for making hover match up better, and AFAIK nothing in the emit pipeline relies on `ExpressionWithTypeArguments` only appearing in class `extends` positions (I would feel better about this with a strong type system, but alas).

Instantiation expression type arguments normally live exclusively on the `TypeQuery` itself, which is something we need to double check isn't being doubled up on here; if we go to put an `ExpressionWithTypeArguments` into a `TypeQuery` in the node builder, we *probably* should swap the type arguments up into the type query. I'll also be surprised if you don't need to update failing fourslash tests.

internal/nodebuilder/types.go
```diff
@@ -43,6 +43,7 @@ const (
 	FlagsUseTypeOfFunction                   Flags = 1 << 12
 	FlagsOmitParameterModifiers              Flags = 1 << 13
 	FlagsUseAliasDefinedOutsideCurrentScope  Flags = 1 << 14
+	FlagsUseInstantiationExpressions         Flags = 1 << 31
```
weswigham: ...why is 31 between 14 and 28?

internal/ast/precedence.go
```diff
@@ -698,6 +698,7 @@ func GetTypeNodePrecedence(n *TypeNode) TypePrecedence {
 		KindLiteralType,
 		KindTypePredicate,
 		KindTypeReference,
+		KindExpressionWithTypeArguments,
```
weswigham, line 701: Presumably this is to prevent auto-parenthesizing `(foo<T>).bar`, which has never mattered before since previously these nodes would only appear in heritage clauses.

internal/nodebuilder/types.go
a-tarasyuk: @weswigham I’ve updated it by lowering the value and positioning it after `1 << 29`.

ahejlsberg:
> This is *technically* an abuse of the AST - `ExpressionWithTypeArguments` is only naturally parsed in the context of class `extends` clauses

That used to be the case, but these days we parse `ExpressionWithTypeArguments` for instantiation expressions, so they can definitely occur in regular expressions.

weswigham:
Expressions, yes, not usually arbitrary expressions within type nodes, though - only at the end of a `typeof` query. `declare const x: typeof Something<string>.Other` does not parse (yet is what we're emitting here), but `declare const x: typeof Something<string>` does.

internal/checker/nodebuilderimpl.go
```diff
@@ -3449,3 +3425,98 @@ func (b *NodeBuilderImpl) newIdentifier(text string, symbol *ast.Symbol) *ast.No
...
+		typeArguments = expr.TypeArguments
+	}
+
+	var names []*ast.IdentifierNode
+	node := expr.Expression
+
+	for {
+		if ast.IsExpressionWithTypeArguments(node) {
+			exprWithTypeArgs := node.AsExpressionWithTypeArguments()
+			if exprWithTypeArgs.TypeArguments != nil && len(exprWithTypeArgs.TypeArguments.Nodes) > 0 {
```
weswigham:
```suggestion
			if len(typeArguments) == 0 && exprWithTypeArgs.TypeArguments != nil && len(exprWithTypeArgs.TypeArguments.Nodes) > 0 {
```

Otherwise this is going to hoist any inner type arguments to the outer position on the type query, which is wrong. But if I'm honest, all this is probably overkill, since we probably don't need to strip the inner ones anyway (the caller requested them, after all). I'm pretty sure you don't need any kinda loop, and a simple `return b.f.NewTypeQueryNode(expr.Expression(), expr.TypeArguments)` should suffice? Unless I'm mistaken for some reason?

a-tarasyuk: @weswigham You're right. I missed that this path should preserve only the trailing type arguments in the type query, and overcomplicated it by handling a nested case like `a<number>.b.c<string>` that isn't reachable here. Thanks for pointing that out.

weswigham (approved):

Issue #3384, opened 2026-04-10
Crash in adjusting positions for folding ranges


DanielRosenwasser:
Since https://github.com/microsoft/typescript-go/pull/3365, this is the only failure coming from `LineAndCharacterToPosition`.

```
internal/ls/lsconv.(*Converters).LineAndCharacterToPosition()
	internal/ls/lsconv/converters.go:175
internal/ls.(*LanguageService).adjustFoldingEnd()
	internal/ls/folding.go:43
internal/ls.(*LanguageService).ProvideFoldingRange()
	internal/ls/folding.go:23
internal/lsp.(*Server).handleFoldingRange()
	internal/lsp/server.go:1267
internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].28.1()
	internal/lsp/server.go:798
internal/lsp.(*Server).handleRequestOrNotification.func1()
	internal/lsp/server.go:655
internal/lsp.(*Server).dispatchLoop.func3()
	internal/lsp/server.go:523
internal/lsp.(*Server).dispatchLoop
	internal/lsp/server.go:522
```

jakebailey:
that's very odd, given that implies the line map itself is wrong...

jakebailey:
What is the actual message?

DanielRosenwasser:
No idea - we only keep the stack around.

DanielRosenwasser:
After some investigations, I wonder if this comes back to the snapshotfs implementation. We have two methods, `reloadEntry` and `reloadEntryIfNeeded` which have this code:

https://github.com/microsoft/typescript-go/blob/cf6d69d83c999e5e12ec9d81aec3883fe8d8b110/internal/project/snapshotfs.go#L304-L312

Note that `lineMapOnce`, `lineMap`, `lineInfoOnce` and `lineInfo` aren't also modified.

That said, coming up with a test case end-to-end with hasn't been entirely straightforward (nor has trying to repro it in an actual editor).

jakebailey:
Not overwriting those definitely feels like a bug. We should probably shove those off into a sidecar that can be overwritten with a new object each time the content changes.

andrewbranch:
This doesn't look like a bug to me—`e.Change` clones the file which unsets its line map.

andrewbranch:
This should be closed now, right?

jakebailey:
I suspect so after #3483, given we do not see this failure category anymore.

PR #3419, opened 2026-04-16
Avoid panicking on duplicate document close notifications

Andarist:
I found this when doing some local fuzzing work in an attempt to repro https://github.com/microsoft/typescript-go/issues/3384 . 

I know this fix is kinda for misbehaving clients and I'm not sure if you currently accept fixes for such issues.


jakebailey:
I feel iffy on this; this is probably one of the worst things a client can do and I'm not sure it's realistic?

andrewbranch:
I feel like it's better to enforce invariants unless we have a good reason to believe they don't hold... these kinds of things exist to ensure that we are being internally consistent and correct, e.g. we're not dispatching the same notification twice from the server to the session, or not forgetting to clear out the queue of notifications to be processed in the session, or whatever. We can rewrite lots of code to be maximally permissive, but that makes it harder to find our own bugs when we make future changes.

Issue #3423, opened 2026-04-16
Memory usage spiked in version 7.0.0-dev.20260416.1


pedro757:
I just want to report that memory usage is a lot higher in latest version (7.0.0-dev.20260416.1) than version (7.0.0-dev.20260408.1)

In version 20260408 my project usually uses 1.1Gb of RAM
In version 20260416 (latest) it uses 6Gb of RAM

jakebailey:
That is a wide range.

If you are able to reproduce consistently, could you build from source and bisect?

jakebailey:
And are you referring to the editor? CLI compilation?

sbking:
I am seeing a massive memory leak in VSCode TypeScript Native Preview this morning. Seems to eat up several gigabytes of RAM as I type in the editor and it autocompletes. The memory is never freed until I close VS Code. Continues to grow until I'm out of memory.

jakebailey:
When this happens, can you please run the "Heap Profile" command in the command palette and upload it?

jakebailey:
Or, build https://github.com/microsoft/typescript-go/pull/3425 from source and see if that makes the problem go away.

sbking:
Here are two so hopefully it's diffable

[52774-1776354170343-heapprofile.pb.gz](https://github.com/user-attachments/files/26791890/52774-1776354170343-heapprofile.pb.gz)
[52774-1776354270931-heapprofile.pb.gz](https://github.com/user-attachments/files/26791891/52774-1776354270931-heapprofile.pb.gz)

jakebailey:
If this is easily reproducible, could you also try:

```
"editor.semanticHighlighting.enabled": false,
```

In your user settings?

sbking:
That setting did not help. Here is a heap profile with it disabled:

[93277-1776355068999-heapprofile.pb.gz](https://github.com/user-attachments/files/26792148/93277-1776355068999-heapprofile.pb.gz)

jakebailey:
@sbking Can you try again with the version we just put out on the marketplace?

pedro757:
@jakebailey I think it improved a bit in the latest version, but still not as good as version 20260408, latest version uses 50% more RAM

jakebailey:
There's nothing else that should have affected that; if possible a repro would be good, or a bisect

PR #3429, opened 2026-04-16
Ensure request cancel is always called

jakebailey:
This appears to be the root cause of #3423. The cancel function from `WithCancel`, `WithTimeout`, etc, _must_ be called no matter what. Normally, this is deferred, but this is trickier code because of how we need to order requests and do cancellation out-of-band.

### DIFF
--- a/internal/project/session.go
+++ b/internal/project/session.go
@@ -382,6 +382,7 @@ func (s *Session) ScheduleDiagnosticsRefresh() {
 
 	// Enqueue the debounced diagnostics refresh
 	s.backgroundQueue.Enqueue(debounceCtx, func(ctx context.Context) {
+		defer cancel()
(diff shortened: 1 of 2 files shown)


andrewbranch (approved):

internal/project/session.go
```diff
@@ -382,6 +382,7 @@ func (s *Session) ScheduleDiagnosticsRefresh() {
 
 	// Enqueue the debounced diagnostics refresh
 	s.backgroundQueue.Enqueue(debounceCtx, func(ctx context.Context) {
+		defer cancel()
```
jakebailey, line 385: This code might not call `cancelDiagnosticsRefresh` and so this could also be dangling. It should be safe to call this here.

andrewbranch (approved):

PR #3430, opened 2026-04-16
Move @typescript/api and @typescript/ast into @typescript/native-preview for publishing

andrewbranch:
To avoid version compatibility issues between the JS API client and Go API server, we decided not to separate them into different packages. This PR moves the contents of `@typescript/api` and `@typescript/ast` into the `@typescript/native-preview` package, which will eventually become `typescript`.

I'd like to merge this _after_ the TypeScript 7.0 beta release.


_packages/native-preview/package.json
```diff
@@ -29,13 +29,92 @@
     "engines": {
         "node": ">=16.20.0"
```
jakebailey, line 30: This engine will have to be updated; I lowered it because we had written the code in such a way that it worked on older versions, but now we contain code that targets a newer version.

_packages/native-preview/src/api/options.ts
```diff
@@ -21,7 +22,11 @@ export interface ClientSpawnOptions {
 export type ClientOptions = ClientSocketOptions | ClientSpawnOptions;
 
 export function isSpawnOptions(options: ClientOptions): options is ClientSpawnOptions {
-    return "tsserverPath" in options;
+    return !("pipe" in options);
+}
+
+export function resolveExePath(options: ClientSpawnOptions): string {
+    return options.tsserverPath ?? getExePath();
```
andrewbranch, line 29: I kept the option only because I feel like it will be useful for us in local development/debugging, but maybe not—I could remove it from the declaration emit types, or remove it altogether, or remove it once we're ready to RC. Thoughts?

_packages/native-preview/package.json
```diff
@@ -29,13 +29,92 @@
...
+        "build": "tsc -b",
+        "prebuild:test": "npm run generate",
+        "build:test": "tsc -b test",
+        "test:only": "npm run node -- --test './test/**/*.test.ts'",
+        "pretest": "npm run generate",
+        "test": "npm run test:only"
+    },
+    "dependencies": {
+        "@types/node": "^25.5.0",
+        "vscode-jsonrpc": "^9.0.0-next.11"
```
jakebailey: Is there a world in which we can drop these? Or maybe even vendor what we need, so TS remains dep free?

_packages/native-preview/package.json
jakebailey: See also #2191

_packages/native-preview/package.json
```diff
@@ -29,13 +29,92 @@
...
+        "dist"
+    ],
     "bin": {
         "tsgo": "./bin/tsgo.js"
     },
     "exports": {
-        "./package.json": "./package.json"
+        "./package.json": "./package.json",
+        "./sync": {
+            "@typescript/source": "./src/api/sync/api.ts",
```
andrewbranch, line 44: I could strip these from the published package in Herebyfile; do we care?

_packages/native-preview/package.json
andrewbranch: vscode-jsonrpc for sure. For `@types/node` we could probably forward declare minimal interfaces for what we use...

jakebailey: Do we even need to declare `node` dep? We of course don't want to lock people to this specific major version of node's types.

andrewbranch: Hm, actually, we don’t. I thought that some Node.js type would turn up in options passed to the `Client` or public properties on it, but everything is private, so no external imports appear in our generated `.d.ts` files at all.

_packages/native-preview/package.json
jakebailey: Oh shoot, is it really 22+? v20 is still LTS; if we think we're compatible it would probably be okay to declare lower?

andrewbranch: The API itself works down to 16 at least, but running the tests currently requires `--experimental-strip-types` and `globSync`. The `engines` field has never been preserved in the published package.json, so this is only serving for local development. Maybe that wasn't the intention?

jakebailey: It was definitely my intent to preserve the engines; if it's not working that's totally a bug, oops...

andrewbranch: If the field is meant for the published packages, there was no reason to change the version. What were you concerned would not work on v16?

jakebailey: Yeah, I assumed we had not tested or did not support <20.

andrewbranch: I also assumed it wouldn't work but it did 😄 Do we do any validation of the built release packages where we could smoke test them on different Node.js versions?

jakebailey: We could; the release-build job does produce all tarballs that could be then used in later builds and see, I just wasn't sure of the best way to do that at the time, so did not have it do anything except make sure it builds.

andrewbranch: Just reverted the version bump and included the field in the output for now.

_packages/native-preview/package.json
```diff
@@ -29,13 +29,91 @@
...
+        "#getExePath": "./lib/getExePath.js",
+        "#enums/*": {
+            "@typescript/source": {
+                "types": "./src/enums/*.enum.ts",
+                "default": "./src/enums/*.ts"
+            },
+            "types": "./dist/enums/*.enum.d.ts",
+            "default": "./dist/enums/*.js"
+        },
+        "#vscode-jsonrpc/node": "./vendor/vscode-jsonrpc/lib/node/main.js"
```
jakebailey, line 98: Should we dev dep on this just so we can quickly regenerate the `vendor` dir?

andrewbranch: Sure

_packages/native-preview/package.json
jakebailey: It's probably worth stripping but I don't know if it's a blocker

_packages/native-preview/src/api/options.ts
jakebailey: I don't think it's a big deal to leave in for now

jakebailey:
Can we exclude the vendored code from formatting so it's just a pure copy?

I'm also wondering if we should bundle their code with esbuild or something instead of shipping it as-is, but it's "only" 10% of the package size, so probably not really a problem

jakebailey (approved):
So long, subpackages

PR #3432, opened 2026-04-17
Enable allowJs when inferred project has root JS files

### LINKED ISSUE #2621: Fourslash is broken when making requests in unconfigured .js files
Error message: 
-

```
--- FAIL: TestAutoImportCompletionJs (0.07s)
    c:/typescript-go/internal/fourslash/tests/autoImportCompletion_test.go:159: At marker 'd': textDocument/completion request returned error: [-32603]: no project found for URI file:///d.js
FAIL
FAIL	github.com/microsoft/typescript-go/internal/fourslash/tests	1.758s
```

Repro: 
-
Fails on requests called in .js files (formatting, completions, codefixes, etc) 

Example: in `internal/fourslash/tests/autoImportCompletions_test.go`:
```
func TestAutoImportCompletionJs(t *testing.T) {
	t.Parallel()
	const content = `// @Filename: a.js
export const someVar = 10;

// @Filename: d.js
import {someVar} from "./a";
someVar;
a/*d*/
`
	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, content)
	defer done()
		f.VerifyCompletions(t, "d", &fourslash.CompletionsExpectedList{
		UserPreferences: &lsutil.UserPreferences{
			IncludeCompletionsForModuleExports:    core.TSTrue,
			IncludeCompletionsForImportStatements: core.TSTrue,
		},
		IsIncomplete: false,
			CommitCharacters: &DefaultCommitCharacters,
			EditRange:        Ignored,
		},
		Items: &fourslash.CompletionsExpectedItems{
			Includes: []fourslash.CompletionsExpectedItem{"someVar", "anotherVar"},
		},
	})
}
```

Info:
-
I've traced it back to .js files not being found in a project during `GetDefaultProject`

jakebailey:
Fixes #2621

See also: https://github.com/microsoft/typescript-go/issues/2621#issuecomment-4269361026

This required introducing a `@noDefaultOpen` harness option, so that our fourslash runner doesn't open every file and cause our tests to not actually test "js exists but was not clicked into". The original fourslash runner worked directly on the language service without a server, so never hit this as there was no concept of "open".

That being said, it's a little unclear how the server was working so well in JS code without this code?


internal/fourslash/fourslash.go
```diff
@@ -894,6 +899,12 @@ func (f *FourslashTest) openFile(t *testing.T, filename string) {
 	f.baselineProjectsAfterNotification(t, filename)
 }
 
+// OpenFile opens a file via LSP didOpen, making it a root of the inferred project.
```
jakebailey:
```suggestion
// OpenFile opens a file via LSP didOpen.
```

andrewbranch:
> That being said, it's a little unclear how the server was working so well in JS code without this code?

Hm, is `compilerOptionsForInferredProjects` actually getting sent by default? The hard coded default compiler options for inferred projects has `AllowJs` enabled.

jakebailey:
It is right at startup, but the options it sends do not include allowJs; I don't think they did in Strada either, and it overrode the compiler options to force these two options on. Could be that we just need to undo that initial step, it's just really wonky given that corsa, strada fourslash, strada fourslash-server all behave differently

andrewbranch:
LGTM but generate is failing

internal/fourslash/fourslash.go
```diff
@@ -210,6 +210,10 @@ func NewFourslash(t *testing.T, capabilities *lsproto.ClientCapabilities, conten
 	if testData.isStateBaseliningEnabled() {
 		// Single baseline, so initialize project state baseline too
 		f.stateBaseline = newStateBaseline(fsFromMap.(iovfs.FsWithSys))
+	} else if harnessOptions.NoDefaultOpen {
```
DanielRosenwasser: I'm sure some state baseline tests probably could use this instead, but not worth the hassle right now.

DanielRosenwasser:
Why don't we just enable allowJs for inferred projects without adjustments based on root files?

jakebailey:
We can do that, though I think we still need to _force_ it on like Strada did, lest someone configure the inferred project manually, which is more or less the problem fourslash has.

jakebailey:
Oh, I'm dumb, the tests I'm referring to are tests which are explicitly trying to test behavior when allowJs is off. Maybe these tests are just bunk

jakebailey:
Here's an example of a fun one:

```diff
--- old.goToDefinitionCSSPatternAmbientModule.baseline.jsonc
+++ new.goToDefinitionCSSPatternAmbientModule.baseline.jsonc
@@= skipped -1, +1 lines =@@
 // === /index.css ===
 // [||]html { font-size: 16px; }

-// === /types.ts ===
-// <|declare module [|"*.css"|] {
-//   const styles: any;
-//   export = styles;
-// }|>
-
 // === /index.ts ===
 // import styles from /*GOTO DEF*/[|"./index.css"|];
```

We turn on both `allowJs` and `allowNonTsExtensions`, and this is an inferred project, so then we for some reason decide to not go to the types file anymore. (I'm not sure why the .css file is even selected before).

andrewbranch:
> Why don't we just enable allowJs for inferred projects without adjustments based on root files?

Good question—root files are only _opened_ files, so if you only open `foo.ts` but that imports `./bar.js` (which is really a JS file with no TS / declaration file), is that an error?

DanielRosenwasser:
> if you only open foo.ts but that imports ./bar.js (which is really a JS file with no TS / declaration file), is that an error?

Probably not. Any editors should probably set `allowJs` for unconfigured project settings, so we should probably just make that our default too.

jakebailey:
We actually do not yet support client setting _any_ compiler options.

andrewbranch:
Yeah, confirmed that `session.compilerOptionsForInferredProjects` is never set in real life, only in fourslash. So in real VS Code, we're just using the default options in project.go, which always has `--allowJs`.

jakebailey:
Yeah, so this isn't really pressing to fix, thankfully.

I'm going to look harder and see if it's possible for us to massage our current tests to use the defaults somehow.

We do need to add support for the VS Code options that set these defaults, though.

jakebailey:
I think this PR is not going to go forward, but we can use it as a reference for a future fix

PR #3435, opened 2026-04-17
Make LS checkers time out after being idle, segment based on use (take 2)

### LINKED ISSUE #2115: Refactor checker pool to limit checker creation for foreground tasks, segment diagnostic checkers
Right now, the checker pool just has some 4 checkers it can use; but the editor can ask for doc highlight, then semantic tokens, then inlay hints, all at once, causing duplicate work and higher resource usage.

We should rethink this, probably splitting the pool into "foreground" checkers and then separate ones for diagnostics, hoping to not create so many checkers at once.

jakebailey:
Fixes #2115

This brings back #3363, which I think should be safe after #3429?

Fixes #2115


RyanCavanaugh:
For context, this is a fix for #2115

jakebailey:
It also technically fixes #3638

gabritto:
This seems ok except for possibly the copilot review comments.  But one thing I'm thinking about is that right now we can answer multiple requests for diagnostics at the same time, right? Which this PR forbids. Though I don't know if this actually happens/matters in practice given vscode also sends a bunch of other requests along with diagnostics.

gabritto:
@typescript-bot test tsserver top200

andrewbranch (approved):

andrewbranch (approved):

jakebailey:
This PR is interacting poorly with the ID fixing, for some reason. Looking into it

jakebailey:
@typescript-bot test it

internal/checker/checker.go
```diff
@@ -14289,7 +14289,12 @@ func (c *Checker) recordMergedSymbol(target *ast.Symbol, source *ast.Symbol) {
 }
 
 func (c *Checker) getSymbolIfSameReference(s1 *ast.Symbol, s2 *ast.Symbol) *ast.Symbol {
-	if c.getMergedSymbol(c.resolveSymbol(c.getMergedSymbol(s1))) == c.getMergedSymbol(c.resolveSymbol(c.getMergedSymbol(s2))) {
+	// An instantiated symbol (e.g. a parameter of a signature instantiated to substitute `this`)
+	// refers to the same declaration as its instantiation target. Whether a given context yields the
+	// original symbol or an instantiated copy can depend on incidental checker cache state (see
+	// instantiateSymbol, whose reuse shortcut is gated on the type already being resolved), so unwrap
+	// instantiations here to keep "same reference" answers deterministic.
+	if c.getMergedSymbol(c.resolveSymbol(c.getMergedSymbol(c.getTargetSymbol(s1)))) == c.getMergedSymbol(c.resolveSymbol(c.getMergedSymbol(c.getTargetSymbol(s2)))) {
```
jakebailey, line 14297: This is a latent bug. In Strada, we of course only used one checker, so if we are doing a quick fix for a diagnostic, we must have checked it, and therefore the caches are populated. But now, we serve fixes through a fresh checker, and it has not walked the tree and has not cached things, and so we don't get reference equality like we expect.

Copilot did a lot of print debugging to get here. It's a little unclear if this is a "good" fix, or if there's an even deeper root cause.

Maybe @weswigham has a better clue; this is entirely reproducible on the commit before this code was added.

jakebailey:
@typescript-bot test it

RyanCavanaugh (approved):

PR #3436, opened 2026-04-17
feat: add textDocument/_vs_onAutoInsert handler for VS closing tag auto-insertion

lucygramley:
Implement the VS-specific _vs_onAutoInsert LSP method so that closing HTML/JSX tags are automatically inserted and properly mapped when typing '>' in Visual Studio.

- Add VsOnAutoInsertOptions, VsOnAutoInsertParams, and VsOnAutoInsertResponseItem protocol types with correct _vs_ prefixed JSON field names matching VS LSP client expectations
- Register _vs_onAutoInsertProvider server capability with _vs_triggerCharacters
- Add handler that delegates to existing ProvideClosingTagCompletion
- Fix jsonName bug in code generator unmarshal switch case


internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -572,6 +637,20 @@ const customRequests: Request[] = [
         messageDirection: "clientToServer",
         documentation: "Request to get document highlights across multiple files.",
     },
+    {
+        method: "textDocument/_vs_onAutoInsert",
```
jakebailey, line 641: We should not be putting custom stuff into the main namespaces; use `custom/` like the other custom stuff.

I also don't know that we need to say "VS" everywhere?

jakebailey: Also, Go convention would say `VS`, not `Vs`

internal/lsp/lsproto/_generate/metaModelSchema.mts
```diff
@@ -345,6 +345,12 @@ export type Property = {
 	 * Custom extension for special value types.
 	 */
 	omitzeroValue?: boolean;
+
+	/**
+	 * Override the JSON field name for serialization.
+	 * If omitted, the property name is used.
+	 */
+	jsonName?: string;
```
jakebailey, line 353: I don't think we need this; the spec is written in terms of the raw JSON, and it's the Go side that should be derived from it

internal/lsp/server.go
```diff
@@ -1450,6 +1454,30 @@ func (s *Server) handleClosingTagCompletion(ctx context.Context, ls *ls.Language
 	return ls.ProvideClosingTagCompletion(ctx, params)
 }
 
+func (s *Server) handleVsOnAutoInsert(ctx context.Context, ls *ls.LanguageService, params *lsproto.VsOnAutoInsertParams) (lsproto.VsOnAutoInsertResponse, error) {
+	if params.Character == ">" {
```
jakebailey, line 1458: We don't put code like this in this package; this belong in the `ls `package.

But I'm a little confused why we have this extra call wehn we already have `ProvideClosingTagCompletion`. Why can't that API be extended and we not have two slightly different calls?

internal/lsp/server.go
```diff
@@ -1094,6 +1095,9 @@ func (s *Server) handleInitialize(ctx context.Context, params *lsproto.Initializ
 			},
 			CustomSourceDefinitionProvider:       new(true),
 			CustomMultiDocumentHighlightProvider: new(true),
+			VsOnAutoInsertProvider: &lsproto.VsOnAutoInsertOptions{
+				TriggerCharacters: []string{">"},
+			},
```
jakebailey, line 1100: Doing this makes me wonder if we should instead formalize this feature and implement it the same way in both VS and VS Code.

internal/ls/jsxclosingtag.go
```diff
@@ -27,9 +27,15 @@ func (l *LanguageService) ProvideClosingTagCompletion(ctx context.Context, param
 
 	if element != nil && isUnclosedTag(element.AsJsxElement()) {
 		tagNameNode := element.AsJsxElement().OpeningElement.TagName()
+		// Slight divergence from Strada - we don't use the verbatim text from the opening tag.
+		closingText := "</" + ast.EntityNameToString(tagNameNode, scanner.GetTextOfNode) + ">"
 		result := lsproto.CustomClosingTagCompletion{
-			// Slight divergence from Strada - we don't use the verbatim text from the opening tag.
-			NewText: "</" + ast.EntityNameToString(tagNameNode, scanner.GetTextOfNode) + ">",
+			NewText: closingText,
+			VsTextEdit: &lsproto.TextEdit{
```
jakebailey, line 34: We should not set this if the client is not VS, as it'll be wasted data on other clients. note how we do check the other VS stuff as of my last PR.

(same below too)

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -167,17 +167,82 @@ const customStructures: Structure[] = [
...
+                jsonName: "_vs_textEdit",
+                type: { kind: "reference", name: "TextEdit" },
+                documentation: "The text edit to apply for the closing tag insertion (VS format).",
+            },
+            {
+                name: "vsTextEditFormat",
+                jsonName: "_vs_textEditFormat",
+                type: { kind: "reference", name: "InsertTextFormat" },
+                documentation: "The format of the text edit (plaintext or snippet) (VS format).",
+            },
```
jakebailey, line 188: These should be optional, no?

jakebailey:
There's a lot of overlap between this code and what's in `_extension/src/languageFeatures/tagClosing.ts`; that also construct snippets. It seems to me that we should be reconsidering how everything works and try and even it out instead of adding all new code.

internal/lsp/lsproto/_generate/generate.mts
lucygramley: This method text is what VS expects and it needs the _vs_ tag so I don't think we can change this.

internal/lsp/lsproto/_generate/metaModelSchema.mts
lucygramley: Will look into this to confirm whether or not its needed

internal/lsp/server.go
lucygramley: Yeah good point - can try to combine the functionality as much as possible.

internal/lsp/server.go
lucygramley: Same as above - can work to combine the vs/vscode functionality.

internal/lsp/lsproto/_generate/generate.mts
jakebailey: But this is all custom; their LSP client is really this strict? I guess it's fine.

internal/lsp/lsproto/_generate/generate.mts
lucygramley: The VS side expects this so I don't think its optional.

jakebailey: It might be required for that client, maybe, but you added it to an existing type used by VS Code, and now objects that do not have these set will error.

lucygramley: I see thanks. Turns out these are already included and were added by mistake so I will actually just remove them.

internal/lsp/lsproto/_generate/metaModelSchema.mts
lucygramley: Yeah removing this line breaks a lot of stuff. I think all the '_vs_' fields rely on it and its referenced throughout a lot of the generated code.

jakebailey: But this is new in your PR, no? My previous PR adding VS stuff didn't need it to handle `_vs_` prefixes

internal/lsp/server.go
lucygramley: Okay this has been moved/combined

internal/lsp/lsproto/_generate/metaModelSchema.mts
lucygramley: removed this

internal/lsp/lsproto/_generate/generate.mts
jakebailey, line 1505: I would suggest fixing this:

```suggestion
function methodNameIdentifier(name: string) {
    return name.split("/").map(v => {
        if (v === "$") return "";
        // Mirror goFieldName: "_vs_foo" -> "VSFoo".
        if (v.startsWith("_vs_")) return "VS" + titleCase(v.slice(4));
        return titleCase(v);
    }).join("");
}
```

jakebailey:
Pushed in some changes which serve to deduplicate this; now, VS Code uses the VS style API, which removes the duplication entirely. The code gets simpler from this.

I can see no difference in editor responsiveness; this is still way less chatty than completions.

joj (approved):

PR #3443, opened 2026-04-20
Provide more accurate rename range in `prepareRename` LSP response

auvred:
Strada [uses `node.getStart`](https://github.com/microsoft/TypeScript/blob/55423abe4d029017f19b6e4c32097591994836b4/src/services/rename.ts#L220), while Corsa uses `node.Pos`. This causes the `textDocument/prepareRename` method to return a range with leading trivia. As a result, that range doesn't match the one later returned returned by `textDocument/rename`.

### DIFF
--- a/internal/ls/rename.go
+++ b/internal/ls/rename.go
@@ -363,7 +363,7 @@ func getRenameInfoError(ctx context.Context, message *diagnostics.Message) Renam
 }
 
 func getRenameInfoSuccess(node *ast.Node, sourceFile *ast.SourceFile, displayName string, converters *lsconv.Converters) RenameInfo {
-	start := node.Pos()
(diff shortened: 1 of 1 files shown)


jakebailey:
This seems fine but, no tests change?

auvred:
> This seems fine but, no tests change?

`TextDocumentPrepareRenameInfo` is used in fourslash only in `VerifyRenameSucceeded` and `VerifyRenameFailed`, and neither validates the response it returns:

https://github.com/microsoft/typescript-go/blob/1de8d68230f8759af6fb71d9cf0f9c37d2c65507/internal/fourslash/fourslash.go#L4297-L4325

Should I update them to save the results to the baseline?

jakebailey (approved):
It would be good, though, I'm not sure how that plays into our existing diffing. Probably it's not worth it, I can see that this is obviously correct.

Issue #2533, opened 2026-01-16
Multiple differences in declaration emit for CommonJS exports


ahejlsberg:
The following declaration emit issues were noticed in #2513:

* Use of `typeof module.exports.xxx` (which isn't permitted in .ts files),
* Emitting multiple `export declare var foo: Foo` for multiple assignments to `exports.foo`, and
* Emitting `export declare var foo: Foo` instead of `export const foo: Foo` for CommonJS exports.

Issues are demonstrated by several tests, including:

[tests/cases/conformance/jsdoc/declarations/jsDeclarationsExportDefinePropertyEmit.ts](https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/jsdoc/declarations/jsDeclarationsExportDefinePropertyEmit.ts)
[tests/cases/conassignmentToVoidZero1.tsformance/salsa/](https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/salsa/assignmentToVoidZero1.ts)
[tests/cases/conformance/salsa/moduleExportDuplicateAlias.ts](https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/salsa/moduleExportDuplicateAlias.ts)

RyanCavanaugh:
Confirmed all three still exist in nightly

ahejlsberg:
Note that we support multiple assignments to `module.exports` and `exports.XXX`, so in declaration emit we should emit only for the first occurrence. That would also eliminate the odd duplicate declarations you'd otherwise see.

ahejlsberg:
There are other JS declaration emit issues. For this JS file:

```js
export class C {
  constructor() {
    this.x = 1
    this.y = 2
  }
}

export const obj = {}
obj.x = 1
obj.y = 2

export function f() {}
f.x = 1
f.y = 2
```

We generate this declaration file:

```ts
export declare class C {
    constructor();
}
export declare const obj: {
    x: number;
    y: number;
};
export declare namespace obj {
    var x: number;
}
export declare namespace obj {
    var y: number;
}
export declare function f(): void;
export declare namespace f {
    var x: number;
}
export declare namespace f {
    var y: number;
}
```

Note the missing `x` and `y` properties in `C` and the duplication of the `x` and `y` declarations in `obj`.

weswigham:
[...]

PR #3449, opened 2026-04-20
Remap JS type nodes in declaration emit to TS constructs or fallback to serialization

weswigham:
Fixes part 1 (the bug part) of #2533 and handles #3402 as a side-effect.

With the reparser-focused approach, this was generally supposed to be getting handled in the reparser stage (if we supported them) - the js node remapping logic in the node builder was ported just because some bits of it were still port-able (and it's harmless to keep), so is what I'll say is "partially ported" compared to strada, since a ton of jsdoc node kinds were (and still are) outright missing, highlighting that we're in some weird middleground place right now where the AST isn't exactly like strada's, but nor is it fully handled upfront like we'd originally proposed for corsa.

Rather than deferring to that node builder logic like this, IMO, it'd be better to perform *all* these js-ts type node remappings at parse/reparse time, but since we're only doing a light touch of reparsing right now (object and signature types), this is where we're at, where some remapping logic lives in the parser/reparser so the checker can leverage it, and some in `nodecopy` for emit.

### DIFF
--- a/internal/checker/emitresolver.go
+++ b/internal/checker/emitresolver.go
@@ -1222,3 +1222,12 @@ func (r *EmitResolver) GetPropertiesOfContainerFunction(node *ast.Node) []*ast.S
 	}
 	return r.checker.getPropertiesOfType(r.checker.getTypeOfSymbol(s))
 }
+
+func (r *EmitResolver) TryJSTypeNodeToTypeNode(emitContext *printer.EmitContext, typeNode *ast.Node, enclosingDeclaration *ast.Node, flags nodebuilder.Flags, internalFlags nodebuilder.InternalFlags, tracker nodebuilder.SymbolTracker) *ast.Node {
+	typeNode = emitContext.ParseNode(typeNode)
+	r.checkerMu.Lock()
+	defer r.checkerMu.Unlock()
+
+	requestNodeBuilder := NewNodeBuilder(r.checker, emitContext) // TODO: cache per-context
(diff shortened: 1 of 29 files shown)


jakebailey (approved):

ahejlsberg:
I'm not familiar with the "the js node remapping logic in the node builder". What does it do?

weswigham:
> I'm not familiar with the "the js node remapping logic in the node builder". What does it do?

It's a big part of `nodecopy.go` (which is also concerned with sanity checking nodes copied into a new context behave the same way, which is actually the part that handles https://github.com/microsoft/typescript-go/pull/3402) - it's logic in `visitExistingNodeTreeSymbolsWorker` that remaps jsdoc node kinds to TS ones and explicitly recognizes js-only type node patterns (for kinds that aren't inherently jsdoc only) and remaps them to the equivalent ts ones. _Basically_ the AST equivalent of what the checker does when looking up types for one of these nodes (but rather than node-> type, it's node->node). There's also a bunch of partial-serialization fallback stuff that ended up getting added over time to better support ID, but that's mostly irrelevant here. There's a lot of comments in it for logic that doesn't *seem* to hold water anymore (either because of missing syntax kinds or intentional removals) that we used to do in strada. In theory it was *all* extraneous with the reparser strategy we set out initially (since anything js specific was set to be reparsed into TS constructs), but I ported what I could during the initial port _just in case_ (after all, we could always go back and remove it later as dead code).

PR #3451, opened 2026-04-20
perf: add TSGO_BINARY env var and fast-path sibling check in getExePath

nileshpatil6:
Fixes the performance issue raised in #2836 where `import.meta.resolve` is called on every separate tsgo invocation in large monorepos.

Two changes in `getExePath.js`:

**1. `TSGO_BINARY` environment variable**

If set, the function returns that path immediately and skips all resolution. This is useful for monorepo setups where users want to hardcode the binary location (e.g. in their shell config or CI environment). It also works as a diagnostic tool to verify whether path resolution is the actual bottleneck.

```
TSGO_BINARY=/path/to/tsgo npx tsgo ...
```

If the path does not exist, a clear error is thrown rather than silently failing.

**2. Fast-path sibling check before `import.meta.resolve`**

In the installed package branch, the code now checks if the platform binary exists at the predictable sibling location (`../../{expectedPackage}/lib/`) using a single `fs.existsSync` call before falling back to `import.meta.resolve`. This covers the standard npm/yarn layout where both `@typescript/native-preview` and `@typescript/native-preview-{platform}-{arch}` sit side by side under `node_modules/@typescript/`. If the sibling is not found there (e.g. unusual hoisting), it falls through to the existing `import.meta.resolve` logic unchanged.

The path calculation is the same as the one already used in the `built/npm` branch, so it is not new math.

**Testing**

I wrote a local test suite covering these cases before opening this PR:

- `TSGO_BINARY` pointing to a real file returns it correctly
- `TSGO_BINARY` pointing to a missing file throws with a clear message
- Sibling path math resolves to the correct absolute path given a simulated `node_modules/@typescript/native-preview/lib` directory
- Sibling exe found via fast path when the fake binary is present
- Fast path absent falls through cleanly (no error from the fast path itself)
- Source path (`/_packages/native-preview/lib`) detection still works
- Built-npm path (`/built/npm/native-preview/lib`) detection still works

All 7 tests passed on Node 22.

### DIFF
--- a/_packages/native-preview/lib/getExePath.js
+++ b/_packages/native-preview/lib/getExePath.js
@@ -4,6 +4,16 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 
 export default function getExePath() {
+    // Allow users to bypass resolution entirely. Useful in monorepos where
+    // import.meta.resolve is slow across many separate tsgo invocations.
+    const envPath = process.env.TSGO_BINARY;
+    if (envPath) {
+        if (!fs.existsSync(envPath)) {
+            throw new Error("TSGO_BINARY points to non-existent file: " + envPath);
+        }
+        return envPath;
+    }
+
     const __dirname = path.dirname(fileURLToPath(import.meta.url));
     const normalizedDirname = __dirname.replace(/\\/g, "/");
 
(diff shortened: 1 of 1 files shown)


jakebailey:
If someone has this path, why wouldn't they just run it directly? This seems really special case-y.

nileshpatil6:
Fair point -- `TSGO_BINARY` is niche and easy to misuse. Happy to drop it entirely if you prefer. The core value I see is the fast-path sibling check: using `fs.existsSync` to probe the adjacent package before falling back to `import.meta.resolve`, which avoids the resolve overhead in the common monorepo setup. Want me to push a revised version with `TSGO_BINARY` removed and just the sibling check kept?

RyanCavanaugh:
Closing for now since this seems to not be scoped in a way we could merge it for 7.0. I remain highly skeptical of the profiling in the linked issue -- tsgo on load does a huge amount of other work that should dwarf the exe lookup path, unless something is going terribly wrong.

PR #3454, opened 2026-04-20
Cancel auto-import cache warming if changes come in while building

andrewbranch:
Currently, any interim snapshot change will defeat our ability to integrate auto-imports built against an older snapshot into the current session state (i.e., we'll have to throw away the whole registry we just built if the session snapshot updates), so we may as well stop the in-progress work.

I experimented a little with doing a snapshot merge, and it was looking promising, but it was getting complicated and there were still some edge cases to handle, so I shelved that for now.

### DIFF
--- a/internal/project/session.go
+++ b/internal/project/session.go
@@ -137,7 +137,8 @@ type Session struct {
 
 	// warmAutoImportCancel is the cancelation function for a running
 	// auto-import cache warming task. It is cancelled on file opens,
-	// closes, changes, and new auto-import warming requests.
+	// closes, changes, watched-file changes, new auto-import warming
+	// requests, and when the session closes.
 	warmAutoImportCancel context.CancelFunc
 	warmAutoImportMu     sync.Mutex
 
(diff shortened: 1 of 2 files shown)


andrewbranch:
@typescript-bot test top100
@typescript-bot perf test

gabritto:
@typescript-bot test tsserver top100

andrewbranch:
Ah, I thought memory was reported in the LSP section of the perf tests. Oh well.

andrewbranch:
Nobody said 400 🤔

jakebailey:
The bot does `min(request, 400)` now, since people kept doing `top100` and expecting to get much back 😄

andrewbranch:
Ah, the previous approach was wrong. The latest commit cuts the runtime of the elastic/kibana replay script in half on my machine, though sadly does nothing for the peak memory usage.

@typescript-bot test tsserver top400

gabritto (approved):

PR #3457, opened 2026-04-21
Fixed crash when resolving declaration spaces on merged symbols resolving to export assignment

Andarist:
fixes a crash reported here: https://github.com/microsoft/typescript-go/issues/3448#issuecomment-4283766710

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -6774,7 +6774,6 @@ func (c *Checker) getDeclarationSpaces(node *ast.Declaration) DeclarationSpaces 
 		if !ast.IsEntityNameExpression(expression) {
 			return DeclarationSpacesExportValue
 		}
-		node = expression
 		// The below options all declare an Alias, which is allowed to merge with other values within the importing module.
 		fallthrough
 	case ast.KindImportEqualsDeclaration, ast.KindNamespaceImport, ast.KindImportClause:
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -6774,7 +6774,6 @@ func (c *Checker) getDeclarationSpaces(node *ast.Declaration) DeclarationSpaces
 		if !ast.IsEntityNameExpression(expression) {
 			return DeclarationSpacesExportValue
 		}
-		node = expression
```
Andarist, line 6777: With this, an identifier assigned to `node` would fallthrough, `getSymbolOfDeclaration` would return `nil` and `resolveAlias` would crash. I'm not sure what's the origin of this line, but given a lot (if not all) expressions assigned to `node` would just lead to a quick panic (as those expressions are not declaration) this is really the easiest fix

gabritto: In Strada we have symbols attached to identifiers, so this would work. But this change looks ok, I think we'll get the same symbol we would get from looking the identifier symbol in Strada.

gabritto (approved):

internal/checker/checker.go
```diff
@@ -6774,7 +6774,6 @@ func (c *Checker) getDeclarationSpaces(node *ast.Declaration) DeclarationSpaces
 		if !ast.IsEntityNameExpression(expression) {
```
ahejlsberg:
```suggestion
		if !ast.IsEntityNameExpression(expression) || c.getSymbolOfDeclaration(node)&ast.SymbolFlagsAlias == 0 {
```

ahejlsberg: We should only fall through if the declared symbol is an alias.

ahejlsberg (dismissed):

internal/checker/checker.go
Andarist: fixed

gabritto (approved):

PR #3496, opened 2026-04-22
Fix error propagation through non-iterable unions

### LINKED ISSUE #3487: tsgo reports downstream TS2339 in ignored for-of over non-iterable union that tsc 6.0.2 does not
### TypeScript Version

- tsc 6.0.2
- tsgo 7.0.0-dev.20260421.2

### Search Terms

for...of, @ts-ignore, TS2488, TS2339, error recovery, parity

### Code

```ts
type A = {
  a: string
}

type B = {
  b: string
}

declare const data: A[] | B

// @ts-ignore
for (const item of data) {
  item.b
}
```

### Commands

```bash
tsc --ignoreConfig --noEmit --strict --target es2022 case4.ts
tsc --ignoreConfig --noEmit --strict --target es2022 --stableTypeOrdering case4.ts
tsgo --ignoreConfig --noEmit --strict --target es2022 case4.ts
```

### Expected Behavior

`tsgo` should match `tsc` 6.0.2 here.

With the `@ts-ignore` on the `for...of` line, `tsc` 6.0.2 reports no errors, including with `--stableTypeOrdering`.

### Actual Behavior

`tsgo` reports:

```text
case4.ts(13,8): error TS2339: Property 'b' does not exist on type 'A'.
```

### Additional Notes

If I remove the `@ts-ignore`, `tsc` 6.0.2 reports only the `for...of` error (`TS2488`), while `tsgo` reports both the `TS2488` and the downstream `TS2339` in the loop body.

So this appears to be a parity difference in error recovery or loop variable typing after an invalid `for...of`.

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/3487


internal/checker/checker.go
```diff
@@ -6093,7 +6093,20 @@ func (c *Checker) getIterationTypesOfIterable(t *Type, use IterationUse, errorNo
 
 func (c *Checker) getIterationTypesOfIterableWorker(t *Type, use IterationUse, errorNode *ast.Node, noCache bool) IterationTypes {
 	if t.flags&TypeFlagsUnion != 0 {
```
Andarist, line 6096: Note that Corsa pushed caching entirely to `getIterationTypesOfIterable` - so the per-constituent results here are not getting cached. That's a preexisting divergence from Strada.

DanielRosenwasser (approved):

jakebailey (approved):

PR #3497, opened 2026-04-22
Widen types assigned to auto-typed variables

### LINKED ISSUE #3639: Behaviour differences: plain js's ternary assignment of objects in different shapes confuses tsgo
## Steps to reproduce

1. Start a plain new project
    - "@typescript/native-preview": "7.0.0-dev.20260428.1",
    - "typescript": "6.0.3"
    - "allowJs": true,
    - "checkJs": true,
    - "stableTypeOrdering": true,

```js
// src/main.js
export class C {
    constructor() {
        const error = { message: 'props.message' };
        const shouldEmitError = Math.random() < 0.5;

        this.state = {
            ctx: shouldEmitError ? { ...error, name: 'A' } : { name: 'B' },
        }
    }

    method() {
        const { ctx: { message, name } } = this.state;
    }
}
```

```
src/main.js:13:24 - error TS2339: Property 'message' does not exist on type '{ message: string; name: string; } | { name: string; }'.

13         const { ctx: { message, name } } = this.state;
                          ~~~~~~~


Found 1 error in src/main.js:13
```

## Behavior with `tsgo`

No error at all.

---

I suspect TS 7 misses some checking that TS 6 has. 🤔

Andarist:
Fixes #3491, #3639

fixes https://github.com/microsoft/typescript-go/issues/3491
fixes https://github.com/microsoft/typescript-go/issues/3639

### DIFF
--- a/internal/checker/flow.go
+++ b/internal/checker/flow.go
@@ -229,7 +229,7 @@ func (c *Checker) getTypeAtFlowAssignment(f *FlowState, flow *ast.FlowNode) Flow
 			if c.isEmptyArrayAssignment(node) {
 				return FlowType{t: c.getEvolvingArrayType(c.neverType)}
 			}
-			assignedType := c.getWidenedLiteralType(c.getInitialOrAssignedType(f, flow))
+			assignedType := c.getWidenedType(c.getWidenedLiteralType(c.getInitialOrAssignedType(f, flow)))
 			if c.isTypeAssignableTo(assignedType, f.declaredType) {
 				return FlowType{t: assignedType}
 			}
--- a/testdata/baselines/reference/compiler/autoTypeJsLiteral1.errors.txt
+++ b/testdata/baselines/reference/compiler/autoTypeJsLiteral1.errors.txt
@@ -1,0 +1,26 @@
+EnvironmentPlugin.js(13,18): error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
+  No index signature with a parameter of type 'string' was found on type '{}'.
+
+
+==== EnvironmentPlugin.js (1 errors) ====
+    /** @typedef {string | number | boolean} CodeValue */
+    
+    class EnvironmentPlugin {
+    	constructor() {
+            this.keys = /** @type {string[]} */ ([]);
+    		this.defaultValues = {};
+    	}
+    
+    	apply() {
+    		/** @type {Record<string, CodeValue>} */
+    		const definitions = {};
+    		for (const key of this.keys) {
+    			const value = this.defaultValues[key];
+    			              ~~~~~~~~~~~~~~~~~~~~~~~
+!!! error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
+!!! error TS7053:   No index signature with a parameter of type 'string' was found on type '{}'.
+    			definitions[key] = value;
+    		}
+    	}
+    }
+    
--- a/testdata/baselines/reference/compiler/autoTypeJsLiteral1.symbols
+++ b/testdata/baselines/reference/compiler/autoTypeJsLiteral1.symbols
@@ -1,0 +1,48 @@
+//// [tests/cases/compiler/autoTypeJsLiteral1.ts] ////
+
+=== EnvironmentPlugin.js ===
+/** @typedef {string | number | boolean} CodeValue */
+
+class EnvironmentPlugin {
+>EnvironmentPlugin : Symbol(EnvironmentPlugin, Decl(EnvironmentPlugin.js, 0, 0))
+
+	constructor() {
+        this.keys = /** @type {string[]} */ ([]);
+>this.keys : Symbol(EnvironmentPlugin.keys, Decl(EnvironmentPlugin.js, 3, 16))
+>this : Symbol(EnvironmentPlugin, Decl(EnvironmentPlugin.js, 0, 0))
+>keys : Symbol(EnvironmentPlugin.keys, Decl(EnvironmentPlugin.js, 3, 16))
+
+		this.defaultValues = {};
+>this.defaultValues : Symbol(EnvironmentPlugin.defaultValues, Decl(EnvironmentPlugin.js, 4, 49))
+>this : Symbol(EnvironmentPlugin, Decl(EnvironmentPlugin.js, 0, 0))
+>defaultValues : Symbol(EnvironmentPlugin.defaultValues, Decl(EnvironmentPlugin.js, 4, 49))
+	}
+
+	apply() {
+>apply : Symbol(EnvironmentPlugin.apply, Decl(EnvironmentPlugin.js, 6, 2))
+
+		/** @type {Record<string, CodeValue>} */
+		const definitions = {};
+>definitions : Symbol(definitions, Decl(EnvironmentPlugin.js, 10, 7))
+
+		for (const key of this.keys) {
+>key : Symbol(key, Decl(EnvironmentPlugin.js, 11, 12))
+>this.keys : Symbol(EnvironmentPlugin.keys, Decl(EnvironmentPlugin.js, 3, 16))
+>this : Symbol(EnvironmentPlugin, Decl(EnvironmentPlugin.js, 0, 0))
+>keys : Symbol(EnvironmentPlugin.keys, Decl(EnvironmentPlugin.js, 3, 16))
+
+			const value = this.defaultValues[key];
+>value : Symbol(value, Decl(EnvironmentPlugin.js, 12, 8))
+>this.defaultValues : Symbol(EnvironmentPlugin.defaultValues, Decl(EnvironmentPlugin.js, 4, 49))
+>this : Symbol(EnvironmentPlugin, Decl(EnvironmentPlugin.js, 0, 0))
+>defaultValues : Symbol(EnvironmentPlugin.defaultValues, Decl(EnvironmentPlugin.js, 4, 49))
+>key : Symbol(key, Decl(EnvironmentPlugin.js, 11, 12))
+
+			definitions[key] = value;
+>definitions : Symbol(definitions, Decl(EnvironmentPlugin.js, 10, 7))
+>key : Symbol(key, Decl(EnvironmentPlugin.js, 11, 12))
+>value : Symbol(value, Decl(EnvironmentPlugin.js, 12, 8))
+		}
+	}
+}
+
--- a/testdata/baselines/reference/compiler/autoTypeJsLiteral1.types
+++ b/testdata/baselines/reference/compiler/autoTypeJsLiteral1.types
@@ -1,0 +1,57 @@
+//// [tests/cases/compiler/autoTypeJsLiteral1.ts] ////
+
+=== EnvironmentPlugin.js ===
+/** @typedef {string | number | boolean} CodeValue */
+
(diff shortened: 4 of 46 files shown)


ahejlsberg:
@typescript-bot test it

ahejlsberg:
Looks like we're fine on perf, but wondering what the changes in user tests and top 400 are about?

Andarist:
[webpack repro](https://www.typescriptlang.org/play/?module=1&noEmit=true&ts=6.0.0-dev.20260416&filetype=js#code/PTAEAEDMEsBsFMB2BDAtvAXKAoogbtAE4D2i6iALgAqwCuA5tIgHQBWAzgFDABUPEFAJ4AHeABN4kUAG92FQk3qgAPqES1UAI3iEVoTcWIJkiAL6gAwsQkA1ZHXigewTpwDGsZO3Y58RUuTUdIyIMpwAkLw8EU4QwsiEaDIAFHIKiEqqaYoA2gC6egBK8G7EhGIAPNkZADSW1vB2DgB8AJT55gDW8IJc4U4u4aWIabRuFGXJzNPdva1h4eHQUsmz7MwIGRQAFqAAvAegAIygAGSnoACChImCzNDs17erPew5AAx5rfPSEYtRAhEjlk8lyeXMzj+4R2D2Ya32oDWHzyAG4oTD1hJIMhaLAKE1aPAfHsZKY0YtzPBYOxHMtQC9ehskPQdvtDidzojXsizhchKJiFIkZ82SSAETETSsEoUMU-dHbWHwkkAeSlMrhrwZb0+rXJiwxzCxOLxBKJ+yhkT4gNEMmKpXKVVBtXqtnshOaEMGi3C2uRer+lOpwIVSteCIB4H5wOq9A6A3pawDPsNxtx+Pd5pJ0jJgYiplc4WQwmEsEEyXl-2tUaBdpKZUqsbqVjdLS9f2GclAWKY0Ao0FIxNJ+sgZXpnYoXMEoEFoENSYWBsVmMkJozDh8IFAbloNyQFCw9obTvS9GbDTNzTqmhI3UQWBzUInoDwmYRqdX6bNb1mqKhPcQPsBxGHJfwRV8HH1AtwgLAsgA)

This one makes `Record<string, CodeValue> | {}` to be reduced to `{}` (where previously it would reduce to `Record<string, CodeValue>`). But note that `{}` is a special `JSLiteral` type that has an implicit index signature - widening loses some flags on the type (but IIRC from my debugging session it doesn't lose JSLiteral). And because the flags are different, the strict subtype relationship checks in the union subtype reduction change the result. I need to look a little bit more into this to understand the exact rules there.

But if we strip the fact this has a mix of different sources for the `this.defaultValues` and reduce it to one unannotated one, then we can see it does error in the very same way: [TS playground](https://www.typescriptlang.org/play/?module=1&noEmit=true&ts=6.0.0-dev.20260416&filetype=js#code/PTAEAEDMEsBsFMB2BDAtvAXKAoogbtAE4D2i6iALgAqwCuA5tIgHQBWAzgFDABUPEFAJ4AHeABN4kUAG92FQk3qgAPqES1UAI3iEVoTcWIJkiAL6gAwsQkA1ZHXigewTpwDGsZO3Y58RUuTUdIyIMpwAkG6kcoS0bhTEhAAUAJRhoBmZGRQAFtDszADW8II+ALygvPzgQqIyMYoA2gC65s6gSS0pANwR4bn5zBKQyLSwFHYO5TKmveGmruHIwsKwgqlh4eFVAiKO0gBK8FGEYgA8DYj0ADSW1vCTtPAAfG0uW1GIcqDDTNAU0GioAq0lmfUgiQ6n2+xUEoGIUgGBVh7DS0j6kWiFFAeHsT2BoCRQ0ko3Gj3g7EasOacy2v0Q-0BXypJWaBNxDlpC3mnFMQA)

And we can further demo the difference between the subtype reduction rules here ([TS playground](https://www.typescriptlang.org/play/?module=1&noEmit=true&ts=6.0.0-dev.20260416&filetype=js#code/PQKhAIAEBcE8AcCmATRAzcBvAztATgJYB2A5uAD7hECuAtgEaJ4Xj0D2bANogIZEC+4AMJtUANR6dqicCGAAoeaBDzZUeDzw9aWAEqIAxmzzIAPLkKkANMNGIJUxAD5Bmkqrny01IgegE2InBoRFwACjcASix5AEgjIlxwNnoAK3AAXix+AG44hKS8UOpOaABGTPAAWR5oAAsAOi0iZDZaMOiAfnA3cAAuZLS44GBwMfAAPU78wMLi0oAmSpr6pr5W9q6evDIBzH5h0fGp+QOgA)):
```ts
/** @typedef {string | number | boolean} CodeValue */

/**
 * @param {Record<string, CodeValue>} arg
 */
function test(arg) {
	const obj = {};
	const result1 = Math.random() ? arg : obj
	//    ^? const result1: {}
	const result2 = Math.random() ? arg : {}
	//    ^? const result2: Record<string, CodeValue>

}
```

The original webpack break is really close to the `result1`, in my opinion. The final subtype reduction for a constructor-declared type happens in a "non-fresh" context. So I think this one is actually mildly surprising at first but correct.

[RooCodeInc/Roo-Code repro](https://www.typescriptlang.org/play/?checkJs=true&allowJs=true&module=1&noEmit=true&ts=6.0.0-dev.20260416&filetype=ts#code/CYUwxgNghgTiAEYD2A7AzgF3gMwJYRAAUoMALALnkxlxQHN4AfeAVxVDxRGAChRJYCbGzAZcqeFDBgQaNAAoADiQpUMNegEpKhGEgC2uNCAA8ANyS5gAPh48MAT0UIASrJYQsAXh4BIZgDeVCzSsmiU6iwgANzwAO6wKLR0aAD8lNTJANoAuvAAvn6BwaFylNhQEMaxIDB6MBnqyQV2-NBwOCJiErj6ikgwGADKIBhi9Ara8LoGRqZuaB4Ytm2CnSii4ijwvf2DI2PJaABievrEZPJ4BBeqmVo6Z3MmC0u2PFBoDhvrmxLY8k08ACfgIWDgi08dl8uGw8Cu+CIKiBIN8vnUDmBfjRUASuCwUhkcgRN2R2N8EKW8C8kjxWF2A2Go3GKVOBluJKRZE02PyiBIYFI8Nq9RR5MpnmpWLRaMWpXCOEqxgANOTfCKBpQAAYAYSgKBQSAJ8pwiMkWGUZHgACIACQBa5c0j5a2Ue0amA7dAYfUyJBwgCidQG8FS8A9ADp9GEoHQEJRrQBVFAAa0NcW2Hut+S1qplhRlEow0XJAHpS-AAHqpXl+PkgKoIVEU9ySmm4qD4nZ9RkHFmTOvQosl3zlqs1mFw+RFiNyoloMVo4fYsfV7Gw+EzhIwJITeAAMn38C3iSOEYI9Ct1ngAAYxWP4I-4Gin6+AH4fz8fnhjgCE--DYNPQAFSGAAmABmCCAE5HiQZxBkxABybddxSJD4GAJBZHgQ0sBAAAPIwsAkRxnHgJCgjnMIIhgKJYlQo50jUDQ6FyJhWHYEBOG4WI+WKaiyngAAjJAkAIfUaiAxpWL4pCIx-Ct-1-QD6ngUDIJgyhHxmBDHAoxiJgwrCcLw8MiMweBSKcBBKJKedKFE8SQEk1TNRY5I5IUwtRhYHdj1bDAI0MlIshvHJa18Qp8iAA)

This one is weird. I didn't yet have the time to debug **why** this has changed. However, we can clearly see that `result` contains `{ success: boolean; error: string; }` at all relevant control flow points. So the fact that it `result.warnings` access didn't error before this change feels just wrong.

So overall, I think both breaks are expected. I'll push relevant tests to this PR later.

Andarist:
New tests are in now.

ahejlsberg:
I'm not sure about this change. Consider:

```ts
function f(b: boolean) {
    let x;
    if (b) {
        x = { a: 1 }  // Fresh { a: number }
    }
    else {
        x = { b: 'b' }  // Fresh { b: string }
    }
    x;    // Fresh { a: number } | { b: string }
    x.a;  // number | undefined
    x.b;  // string | undefined
    let w = x;  // { a: number, b?: never } | { a?: never, b: string }
    w.a;  // number | undefined
    w.b;  // string | undefined
}
```

With the change in this PR, the `x.a`, `x.b`, `w.a`, and `w.b` accesses all end up being errors. I don't think we want that.

Andarist:
```ts
    x;    // Fresh { a: number } | { b: string }
    x.a;  // number | undefined
```

This is kinda the [RooCodeInc/Roo-Code repro](https://www.typescriptlang.org/play/?checkJs=true&allowJs=true&module=1&noEmit=true&ts=6.0.0-dev.20260416&filetype=ts#code/CYUwxgNghgTiAEYD2A7AzgF3gMwJYRAAUoMALALnkxlxQHN4AfeAVxVDxRGAChRJYCbGzAZcqeFDBgQaNAAoADiQpUMNegEpKhGEgC2uNCAA8ANyS5gAPh48MAT0UIASrJYQsAXh4BIZgDeVCzSsmiU6iwgANzwAO6wKLR0aAD8lNTJANoAuvAAvn6BwaFylNhQEMaxIDB6MBnqyQV2-NBwOCJiErj6ikgwGADKIBhi9Ara8LoGRqZuaB4Ytm2CnSii4ijwvf2DI2PJaABievrEZPJ4BBeqmVo6Z3MmC0u2PFBoDhvrmxLY8k08ACfgIWDgi08dl8uGw8Cu+CIKiBIN8vnUDmBfjRUASuCwUhkcgRN2R2N8EKW8C8kjxWF2A2Go3GKVOBluJKRZE02PyiBIYFI8Nq9RR5MpnmpWLRaMWpXCOEqxgANOTfCKBpQAAYAYSgKBQSAJ8pwiMkWGUZHgACIACQBa5c0j5a2Ue0amA7dAYfUyJBwgCidQG8FS8A9ADp9GEoHQEJRrQBVFAAa0NcW2Hut+S1qplhRlEow0XJAHpS-AAHqpXl+PkgKoIVEU9ySmm4qD4nZ9RkHFmTOvQosl3zlqs1mFw+RFiNyoloMVo4fYsfV7Gw+EzhIwJITeAAMn38C3iSOEYI9Ct1ngAAYxWP4I-4Gin6+AH4fz8fnhjgCE--DYNPQAFSGAAmABmCCAE5HiQZxBkxABybddxSJD4GAJBZHgQ0sBAAAPIwsAkRxnHgJCgjnMIIhgKJYlQo50jUDQ6FyJhWHYEBOG4WI+WKaiyngAAjJAkAIfUaiAxpWL4pCIx-Ct-1-QD6ngUDIJgyhHxmBDHAoxiJgwrCcLw8MiMweBSKcBBKJKedKFE8SQEk1TNRY5I5IUwtRhYHdj1bDAI0MlIshvHJa18Qp8iAA). It feels wrong to me that the displayed type clearly doesn't allow `x.a` in all constituents but there is no error on this access. 

I see your point that this can act like an aggregating multi-statement equivalent of `getWidenedTypeOfObjectLiteral` but it would be nice if the displayed type would reflect that.

I'll take another look at all the cases gathered here, and I'll try to come up with some design that will address your concern and fix the reported cases. This week is packed for me, but I should be able to find some time for this over the upcoming weekend.

jakebailey:
is this still needed after microsoft/typescript-go#3648?

PR #3499, opened 2026-04-22
feat(2280): Port implement interface quick fix

### LINKED ISSUE #2280: Port `implement interface` quick fix
From https://github.com/microsoft/TypeScript/issues/15562

Looks like the `implement interface` quick fix hasn't been ported to TS-go yet. Creating issue to track this

This is one we could consider using Copilot for with the right context and prompt

a-tarasyuk:
Fixes #2280


internal/fourslash/_scripts/crashingTests.txt
```diff
@@ -1 +1 @@
-
+TestAutoImportFileExcludePatterns12
```
jakebailey: We generally aim for no crashing tests, because crashing tests implies crashing users.

Is this a crash in new code? Or just because this PR enables more tests?

internal/ls/codeactions_missingmemberfixer.go
```diff
@@ -0,0 +1,484 @@
...
+	"github.com/microsoft/typescript-go/internal/nodebuilder"
+)
+
+type PreserveOptionalFlags int
+
+const (
+	PreserveOptionalFlagsMethod PreserveOptionalFlags = 1 << iota
+	PreserveOptionalFlagsProperty
+	PreserveOptionalFlagsAll = PreserveOptionalFlagsMethod | PreserveOptionalFlagsProperty
+)
```
jakebailey: These are all internal, right? Do they need to be exported? (Probably true of the fixer itself but I don't think we've been very consistent about that naming)

internal/ls/codeactions_missingmemberfixer.go
```diff
@@ -0,0 +1,484 @@
...
+
+	case ast.KindFunctionDeclaration:
+		fn := signatureDeclaration.AsFunctionDeclaration()
+		return f.ChangeTracker.NodeFactory.UpdateFunctionDeclaration(fn, modifiers, fn.AsteriskToken, core.IfElse(name != nil && ast.IsIdentifier(name), name, nil), typeParameters, parameters, typeNode, fn.FullSignature, core.OrElse(body, fn.Body))
+	}
+
+	return nil
+}
+
+func (f *MissingMemberFixer) createSignatureDeclarationFromSignatures(signatures []*checker.Signature, name *ast.PropertyName, optional bool, modifiers *ast.ModifierList, quotePreference lsutil.QuotePreference, body *ast.FunctionBody, enclosingDeclaration *ast.Node) *ast.Node {
```
jakebailey: IIRC this code was previously called by completions too; is there some duplication here compared to the existing completions code?

internal/ls/codeactions_fixclassincorrectlyimplementsinterface.go
```diff
@@ -0,0 +1,234 @@
...
+	typeChecker, done := fixContext.Program.GetTypeCheckerForFile(context, fixContext.SourceFile)
+	defer done()
+
+	changeTracker := change.NewTracker(context, fixContext.Program.Options(), fixContext.LS.FormatOptions(), fixContext.LS.converters)
+	importAdder, err := createImportAdder(context, fixContext, typeChecker)
+	if err != nil {
+		return nil, err
+	}
+
+	seenClassDeclarations := make(map[*ast.Node]bool)
```
jakebailey: Generally should prefer `collections.Set` over a map to bool if possible.

internal/ls/codeactions_missingmemberfixer.go
a-tarasyuk: `Strada` uses this for method completions and class member snippets, but in `tsgo` it seems this is not implemented yet

https://github.com/microsoft/typescript-go/blob/ee4225a856389fc94c8ce4a5fd9af771dbbc2c0b/internal/ls/completions.go#L2123-L2127

internal/fourslash/_scripts/crashingTests.txt
a-tarasyuk: This PR enables additional tests because I extended the test generation to handle actions containing `"Implement interface '"`, and some auto-import cases are now failing. I fixed this particular case to avoid the crash. I think the remaining auto-import cases should be handled in a separate PR. WDYT?

jakebailey:
These tests are ungenerated:

```
codeFixClassImplementClassMemberAnonymousClass.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementClassMethodViaHeritage.ts parse error: "Unrecognized fourslash statement: verify.rangeAfterCodeFix(...)"
codeFixClassImplementInterfaceCallSignature.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementInterfaceConstructSignature.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementInterfaceDuplicateMember2.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementInterfaceIndexSignaturesNoFix.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementInterfaceMultipleImplements1.ts parse error: "Unrecognized fourslash statement: verify.rangeAfterCodeFix(...)"
codeFixClassImplementInterfaceMultipleImplements2.ts parse error: "Unrecognized fourslash statement: verify.rangeAfterCodeFix(...)"
codeFixClassImplementInterfaceMultipleImplementsIntersection2.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
codeFixClassImplementInterfaceTypeParamInstantiation.ts parse error: "Unrecognized fourslash statement: verify.not.codeFixAvailable(...)"
```

But probably are too annoying to stick in this PR given we don't have `not` or `rangeAfterCodeFix` yet?

jakebailey:
Though I asked copilot to help me find missing tests (since it's hard to do by hand) and it started trying to implement them, so, maybe it is possible 😄

jakebailey:
Here's what it did if you want to steal it, it all passes! I have yet to properly review this PR, however.

```diff
diff --git a/internal/fourslash/_scripts/convertFourslash.mts b/internal/fourslash/_scripts/convertFourslash.mts
index a33ed2c81..ecff3de5e 100755
--- a/internal/fourslash/_scripts/convertFourslash.mts
+++ b/internal/fourslash/_scripts/convertFourslash.mts
@@ -161,12 +161,21 @@ function parseFileContent(filename: string, content: string): GoTest | NoTest {
         console.error(`No commands parsed in file (skipping): ${filename}`);
         return NO_TEST;
     }
-    validateCodeFixCommands(goTest.commands);
+    validateCodeFixCommands(filename, goTest.commands);
     return goTest;
 }
 
-function validateCodeFixCommands(commands: Cmd[]): void {
-    const hasCodeFixCmd = commands.some(c => c.kind === "verifyCodeFix" || c.kind === "verifyCodeFixAvailable" || c.kind === "verifyCodeFixAll");
+// File name prefixes for tests using rangeAfterCodeFix that are allowed
+// because they test implemented code fixes.
+const allowedRangeAfterCodeFixFilePrefixes = [
+    "codeFixClassImplement",
+    "codeFixAmbientClassImplement",
+    "codeFixClassExprClassImplement",
+    "quickfixImplementInterface",
+];
+
+function validateCodeFixCommands(filename: string, commands: Cmd[]): void {
+    const hasCodeFixCmd = commands.some(c => c.kind === "verifyCodeFix" || c.kind === "verifyCodeFixAvailable" || c.kind === "verifyCodeFixAll" || c.kind === "verifyRangeAfterCodeFix");
     if (!hasCodeFixCmd) {
         return;
     }
@@ -176,6 +185,15 @@ function validateCodeFixCommands(commands: Cmd[]): void {
             throw new Error(`Unsupported code fix ID: ${cmd.fixId}`);
         }
     }
+    // rangeAfterCodeFix doesn't carry a fixId or description, so we check the
+    // filename to determine if it's for an implemented code fix.
+    const hasRangeAfterCodeFix = commands.some(c => c.kind === "verifyRangeAfterCodeFix");
+    if (hasRangeAfterCodeFix) {
+        const baseName = path.basename(filename, ".ts");
+        if (!allowedRangeAfterCodeFixFilePrefixes.some(p => baseName.startsWith(p))) {
+            throw new Error(`rangeAfterCodeFix test file '${baseName}' does not match any allowed prefix`);
+        }
+    }
     // If there are codeFix/codeFixAvailable commands but no codeFixAll with an allowed ID,
     // the test is only accepted if its descriptions match allowed patterns.
     const hasAllowedCodeFixAll = commands.some(c => c.kind === "verifyCodeFixAll" && allowedCodeFixIds.has(c.fixId));
@@ -262,6 +280,15 @@ function parseFourslashStatement(statement: ts.Statement): Cmd[] {
             switch (func.text) {
                 case "quickInfoExists":
                     return parseQuickInfoArgs("notQuickInfoExists", callExpression.arguments);
+                case "codeFixAvailable":
+                    // verify.not.codeFixAvailable()
+                    if (ts.isPropertyAccessExpression(namespace) && namespace.expression.getText() === "verify" && namespace.name.text === "not") {
+                        if (callExpression.arguments.length > 0) {
+                            throw new Error(`verify.not.codeFixAvailable with arguments is not supported: ${getBadStatementText(statement)}`);
+                        }
+                        return [{ kind: "verifyCodeFixAvailable", descriptions: [], expectNone: true }];
+                    }
+                    throw new Error(`Unrecognized fourslash statement: ${getBadStatementText(statement)}`);
                 case "andApplyCodeAction":
                     // verify.completions({ ... }).andApplyCodeAction(...)
                     if (!(ts.isCallExpression(namespace) && namespace.expression.getText() === "verify.completions")) {
@@ -386,6 +413,8 @@ function parseFourslashStatement(statement: ts.Statement): Cmd[] {
                     return parseCodeFixAvailableArgs(callExpression.arguments);
                 case "codeFixAll":
                     return parseCodeFixAllArgs(callExpression.arguments);
+                case "rangeAfterCodeFix":
+                    return parseRangeAfterCodeFixArgs(callExpression.arguments);
                 case "semanticClassificationsAre":
                     return parseSemanticClassificationsAre(callExpression.arguments);
                 case "syntacticClassificationsAre":
@@ -1899,6 +1928,9 @@ function parseCodeFixAvailableArgs(args: readonly ts.Expression[]): [VerifyCodeF
     let expectNone = false;
 
     if (args.length === 1) {
+        if (getStringLiteralLike(args[0])) {
+            throw new Error(`verify.codeFixAvailable with a string argument (fix name) is not supported`);
+        }
         const sourceFile = args[0].getSourceFile();
         const arrayArg = getArrayLiteralExpression(args[0]);
         if (arrayArg) {
@@ -1932,6 +1964,20 @@ function parseCodeFixAvailableArgs(args: readonly ts.Expression[]): [VerifyCodeF
     }];
 }
 
+function parseRangeAfterCodeFixArgs(args: readonly ts.Expression[]): [VerifyRangeAfterCodeFixCmd] {
+    if (args.length < 1) {
+        throw new Error(`Expected at least 1 argument in verify.rangeAfterCodeFix, got ${args.length}`);
+    }
+    const str = getStringLiteralLike(args[0]);
+    if (!str) {
+        throw new Error(`Expected string literal in verify.rangeAfterCodeFix, got ${args[0].getText()}`);
+    }
+    return [{
+        kind: "verifyRangeAfterCodeFix",
+        expectedText: str.text,
+    }];
+}
+
 function parseCodeFixAllArgs(args: readonly ts.Expression[]): [VerifyCodeFixAllCmd] {
     if (args.length !== 1) {
         throw new Error(`Expected 1 argument in verify.codeFixAll, got ${args.length}`);
@@ -3685,6 +3731,11 @@ interface VerifyCodeFixAllCmd {
     newFileContent: string;
 }
 
+interface VerifyRangeAfterCodeFixCmd {
+    kind: "verifyRangeAfterCodeFix";
+    expectedText: string;
+}
+
 interface VerifySemanticClassificationsCmd {
     kind: "verifySemanticClassifications";
     format: string;
@@ -3735,7 +3786,8 @@ type Cmd =
     | VerifyErrorExistsBeforeMarkerCmd
     | VerifyCodeFixCmd
     | VerifyCodeFixAvailableCmd
-    | VerifyCodeFixAllCmd;
+    | VerifyCodeFixAllCmd
+    | VerifyRangeAfterCodeFixCmd;
 
 function generateVerifyOutliningSpans({ foldingRangeKind }: VerifyOutliningSpansCmd): string {
     if (foldingRangeKind) {
@@ -4138,6 +4190,8 @@ ${
 	FixID: ${getGoStringLiteral(cmd.fixId)},
 	NewFileContent: ${getGoMultiLineStringLiteral(cmd.newFileContent)},
 })`;
+        case "verifyRangeAfterCodeFix":
+            return `f.VerifyRangeAfterCodeFix(t, ${getGoMultiLineStringLiteral(cmd.expectedText)})`;
         case "verifySemanticClassifications":
             return generateSemanticClassifications(cmd);
         default:
diff --git a/internal/fourslash/fourslash.go b/internal/fourslash/fourslash.go
index 3de3583b7..dbbc6a743 100644
--- a/internal/fourslash/fourslash.go
+++ b/internal/fourslash/fourslash.go
@@ -1561,6 +1561,56 @@ func (f *FourslashTest) VerifyCodeFix(t *testing.T, options VerifyCodeFixOptions
 	}
 }
 
+// VerifyRangeAfterCodeFix applies the first available code fix and verifies that the range content matches.
+// This mirrors TypeScript's verify.rangeAfterCodeFix behavior: apply the fix to the file state,
+// then compare the full file content with whitespace stripped (since TypeScript's includeWhiteSpace
+// defaults to false). The edits are applied mutably so subsequent assertions see the updated state.
+func (f *FourslashTest) VerifyRangeAfterCodeFix(t *testing.T, expectedRangeContent string) {
+	t.Helper()
+
+	actions := f.getCodeFixActions(t)
+	if len(actions) == 0 {
+		t.Fatalf("No code fixes returned.")
+	}
+
+	matchingAction := actions[0]
+	originalContent := f.getScriptInfo(f.activeFilename).content
+
+	selection := f.getSelection()
+	if selection.Pos() == selection.End() {
+		ranges := f.getRangesInFile(f.activeFilename)
+		if len(ranges) == 0 {
+			t.Fatal("Expected a selected range or fourslash range for range content verification.")
+		}
+		selection = ranges[0].Range
+	}
+
+	// Build expected content by replacing the range with the expected text.
+	expectedContent := originalContent[:selection.Pos()] + expectedRangeContent + originalContent[selection.End():]
+
+	// Apply the code fix edits to the file state (mutably, like TypeScript does).
+	if matchingAction.Edit != nil && matchingAction.Edit.Changes != nil {
+		expectedURI := lsconv.FileNameToDocumentURI(f.activeFilename)
+		for uri, edits := range *matchingAction.Edit.Changes {
+			if uri != expectedURI {
+				t.Fatalf("Code fix returned edits for unexpected URI %q (expected %q)", uri, expectedURI)
+			}
+			f.applyTextEdits(t, edits)
+		}
+	}
+
+	actual := f.getScriptInfo(f.activeFilename).content
+
+	// Compare with whitespace stripped, matching TypeScript's default behavior
+	// where includeWhiteSpace defaults to false.
+	expectedStripped := removeWhitespace(expectedContent)
+	actualStripped := removeWhitespace(actual)
+	if expectedStripped != actualStripped {
+		t.Fatalf("File content after code fix doesn't match expected (whitespace-insensitive).\nExpected:\n%s\nActual:\n%s",
+			expectedContent, actual)
+	}
+}
+
 // VerifyCodeFixAvailable verifies that code fixes with the given descriptions are available.
 func (f *FourslashTest) VerifyCodeFixAvailable(t *testing.T, expectedDescriptions []string) {
 	t.Helper()
@@ -5471,3 +5521,16 @@ func (f *FourslashTest) VerifyErrorExistsBeforeMarker(t *testing.T, markerName s
 	}
 	t.Fatalf("Expected error before marker '%s' but none was found", markerName)
 }
+
+// removeWhitespace strips all whitespace characters from a string,
+// matching TypeScript's /\s/g regex behavior.
+func removeWhitespace(s string) string {
+	return strings.Map(func(r rune) rune {
+		switch r {
+		case ' ', '\t', '\n', '\r', '\f', '\v':
+			return -1
+		default:
+			return r
+		}
+	}, s)
+}
```

a-tarasyuk:
@jakebailey Thanks. I had some _stashed_ changes related to `.not.codeFixAvailable`, but adding support for it introduced a bunch of unrelated generated tests unless they were explicitly excluded :). If that’s acceptable, I’ve pushed the changes. The following tests https://github.com/microsoft/typescript-go/pull/3499#issuecomment-4401907480 were generated and removed from the failed cases.

internal/ls/codeactions_missingmemberfixer.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+	declarationName := createDeclarationName(f.changeTracker.NodeFactory, f.typeChecker, symbol, declaration)
+	modifiers := f.createModifiers(symbol, declaration)
+
+	flags := nodebuilder.FlagsNoTruncation
+	if quotePreference == lsutil.QuotePreferenceSingle {
+		flags |= nodebuilder.FlagsUseSingleQuotesForStringLiteralType
+	}
+
+	t := f.typeChecker.GetTypeOfSymbolAtLocation(symbol, enclosingDeclaration)
```
jakebailey: I think this needs `getWidenedType`

internal/ls/codeactions_missingmemberfixer.go
```diff
@@ -0,0 +1,486 @@
...
+					modifiers, createPropertyName(f.changeTracker.NodeFactory, declarationName, quotePreference),
+					nil /*typeParameters*/, createDummyParameters(f.changeTracker.NodeFactory, 1, []string{parameter.Name().Text()}, []*ast.TypeNode{f.createTypeNode(t, enclosingDeclaration, flags, nodeBuilder, idToSymbol)}, 1, ast.IsInJSFile(enclosingDeclaration)),
+					nil /*type*/, nil /*fullSignature*/, f.createBody(body, ambient, quotePreference)),
+				)
+			}
+		}
+		return nodes
+
+	case ast.KindMethodSignature, ast.KindMethodDeclaration:
+		signatures := f.typeChecker.GetSignaturesOfType(t, checker.SignatureKindCall)
```
jakebailey: Does this need to handle unions? Strada had:

```ts
const signatures = type.isUnion() ? flatMap(type.types, t => t.getCallSignatures()) : type.getCallSignatures();
```

internal/ls/codeactions_missingmemberfixer.go
```diff
@@ -0,0 +1,486 @@
...
+		if len(signatures) == 0 {
+			return nil
+		}
+
+		if len(declarations) == 1 {
+			method := f.createSignatureDeclarationFromSignature(core.FirstOrNil(signatures), ast.KindMethodDeclaration, sourceFile, enclosingDeclaration, f.createBody(body, ambient, quotePreference), modifiers, declarationName, optional)
```
jakebailey: For all of the `method :=`s, `optional` is passed here but strada checks `preserveOptional`

internal/ls/autoimport/export.go
```diff
@@ -94,16 +95,48 @@ func (e *Export) IsUnresolvedAlias() bool {
 }
 
 func SymbolToExport(symbol *ast.Symbol, ch *checker.Checker) *Export {
-	if symbol.Parent == nil || !checker.IsExternalModuleSymbol(symbol.Parent) {
+	if symbol.Parent != nil && checker.IsExternalModuleSymbol(symbol.Parent) {
```
jakebailey, line 98: @andrewbranch  can you take a look at this bit of code? It fixes a crash in a new test that was exposed (so probably fixes a crash people have been hitting that we did not have a test for until now).

andrewbranch: Which test is it? I don't really think this code makes sense. You're not supposed to call this function with a module symbol. If that was happening before, the caller is probably doing something wrong, and I'd like to look at it outside the context of this PR. If it's only happening because of other changes in this PR, then the approach in this PR is bending this function far beyond its intended use.

jakebailey: See this commit: https://github.com/microsoft/typescript-go/pull/3499/commits/393bda19fd1dafb17ea7c2e63739663718aa6435

andrewbranch: Those are specific to implement interface, so that means not an existing crash—likely related to how the implement interface code is newly using this function.

a-tarasyuk: @andrewbranch @jakebailey thanks for the feedback. To clarify the preferred direction here - should this change be dropped, or should the fix (with additional investigation related to why the caller is probably doing something wrong) stay in this PR?

andrewbranch: I'm checking the PR out and debugging locally to see if I can get an idea of how it should work.

andrewbranch: Took another look at this and I think it's fine. It’s a little weird because the import adder is passing resolved, merged symbols into here, which isn’t what `extractFromSymbol` expects. Arguably a merged symbol should produce _multiple_ Exports. This function was always a little bit of a kludge, but when I added it, it was only used by `getFixesInfoForUMDImport`. The import adder started using it in #2479, but I'm noticing now, `importAdder.getAllExportsForSymbol` is using it only to get an `ExportID`. I would maybe expect a function that takes a symbol and returns multiple ExportIDs. Unfortunately, it's hard to reliably walk backwards from resolved/merged symbols to their original constituents. I'm experimenting along those lines, but I wouldn't block over this having looked closer at it.

a-tarasyuk: @andrewbranch, thanks for the update.

jakebailey (approved):
It's a chunky PR, but it passes _all_ tests from Strada, and I think it looks reasonably 1:1 besides the changes I pointed out earlier (or inconsequential differences).

Happy to give this a shot

PR #3501, opened 2026-04-22
Fix conditional instantiation cache key mismatch for unresolved conditional types

### LINKED ISSUE #3488: Repeated identical conditional return type triggers TS2719 ("Two different types with this name exist") under tsgo
## Steps to reproduce

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "strict": true,
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "noEmit": true
  }
}
```

`index.ts`:

```ts
export function cast<T>(value: T): {
  as: <K extends T>() => null extends T ? K | null : undefined extends T ? K | undefined : K;
} {
  return {
    as: <K extends T>(): null extends T ? K | null : undefined extends T ? K | undefined : K => {
      return value as K;
    },
  };
}
```

Workaround: extract the conditional into a single named alias so both annotation sites reference the same type symbol.

```ts
type PreserveNullability<T, K extends T> = null extends T
  ? K | null
  : undefined extends T
    ? K | undefined
    : K;

export function cast<T>(value: T): {
  as: <K extends T>() => PreserveNullability<T, K>;
} {
  return {
    as: <K extends T>(): PreserveNullability<T, K> => {
      return value as PreserveNullability<T, K>;
    },
  };
}
```

Compiles cleanly (exit 0, no diagnostics).

## Behavior with `tsgo`

```
index.ts(5,5): error TS2719: Type '<K extends T>() => null extends T ? K | null : undefined extends T ? K | undefined : K' is not assignable to type '<K extends T>() => null extends T ? K | null : undefined extends T ? K | undefined : K'. Two different types with this name exist, but they are unrelated.
  Type 'null extends T ? K | (T & undefined) | (T & null) : undefined extends T ? K | (T & undefined) | (T & null) : K | (T & undefined) | (T & null)' is not assignable to type 'null extends T ? K | (T & null) : undefined extends T ? K | (T & undefined) : K'.
    Type 'K | (undefined extends T ? K | (T & undefined) | (T & null) : K | (T & undefined) | (T & null)) | (T & undefined) | (T & null)' is not assignable to type 'null extends T ? K | (T & null) : undefined extends T ? K | (T & undefined) : K'.
      Type 'undefined extends T ? K | (T & undefined) | (T & null) : K | (T & undefined) | (T & null)' is not assignable to type 'null extends T ? K | (T & null) : undefined extends T ? K | (T & undefined) : K'.
        Type 'K | (T & undefined) | (T & null)' is not assignable to type 'null extends T ? K | (T & null) : undefined extends T ? K | (T & undefined) : K'.
          Type 'T & undefined' is not assignable to type 'null extends T ? K | (T & null) : undefined extends T ? K | (T & undefined) : K'.
            Type 'K | (T & undefined) | (T & null)' is not assignable to type 'K | (T & null)'.
              Type 'T & undefined' is not assignable to type 'K | (T & null)'.
                Type 'T & undefined' is not assignable to type 'T & null'.
                  Type 'T & undefined' is not assignable to type 'null'.
```

The two types in the diagnostic are written out identically. Note also that in tsgo's elaboration the type has been re-distributed (`K | (T & undefined) | (T & null)`) compared with what the user wrote (`K | null` or `K | undefined`), suggesting the conditional is being reduced differently in the two positions before identity is checked.

Andarist:
fix https://github.com/microsoft/typescript-go/issues/3488

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -23816,7 +23816,7 @@ func (c *Checker) getTypeFromConditionalTypeNode(node *ast.Node) *Type {
 		links.resolvedType = c.getConditionalType(root, nil /*mapper*/, false /*forConstraint*/, nil)
 		if outerTypeParameters != nil {
 			root.instantiations = make(map[CacheHashKey]*Type)
-			root.instantiations[getTypeListKey(outerTypeParameters)] = links.resolvedType
+			root.instantiations[getConditionalTypeKey(outerTypeParameters, nil /*alias*/, false /*forConstraint*/)] = links.resolvedType
 		}
 	}
 	return links.resolvedType
--- a/testdata/baselines/reference/compiler/conditionalContextualReturnSubstitutionCache.symbols
+++ b/testdata/baselines/reference/compiler/conditionalContextualReturnSubstitutionCache.symbols
@@ -1,0 +1,41 @@
+//// [tests/cases/compiler/conditionalContextualReturnSubstitutionCache.ts] ////
+
+=== conditionalContextualReturnSubstitutionCache.ts ===
+// https://github.com/microsoft/typescript-go/issues/3488
+
+export function cast<T>(value: T): {
+>cast : Symbol(cast, Decl(conditionalContextualReturnSubstitutionCache.ts, 0, 0))
+>T : Symbol(T, Decl(conditionalContextualReturnSubstitutionCache.ts, 2, 21))
+>value : Symbol(value, Decl(conditionalContextualReturnSubstitutionCache.ts, 2, 24))
+>T : Symbol(T, Decl(conditionalContextualReturnSubstitutionCache.ts, 2, 21))
+
+    as: <K extends T>() => null extends T ? K | null : undefined extends T ? K | undefined : K;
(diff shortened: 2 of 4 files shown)


internal/checker/checker.go
```diff
@@ -23815,7 +23816,7 @@ func (c *Checker) getTypeFromConditionalTypeNode(node *ast.Node) *Type {
 		links.resolvedType = c.getConditionalType(root, nil /*mapper*/, false /*forConstraint*/, nil)
 		if outerTypeParameters != nil {
 			root.instantiations = make(map[CacheHashKey]*Type)
```
Andarist, line 23818: In Strada this key:
```ts
root.instantiations.set(getTypeListId(outerTypeParameters), links.resolvedType)
```
would match the key computed here:
```ts
const id = (forConstraint ? "C" : "") + getTypeListId(typeArguments) + getAliasId(aliasSymbol, aliasTypeArguments);
```

for undefined `alias` and `forConstraint === false`.

But Corsa uses a dedicated `getConditionalTypeKey` that produces a different result in this case than `getTypeListKey(outerTypeParameters)`.

ahejlsberg (approved):

PR #3505, opened 2026-04-22
integrated filewatcher with lsp

johnfav03:
Following up on #3217, this PR refactors `vfswatch.FileWatcher` to make it useable by the LSP in addition to the CLI, and integrates it into the LSP as a polling watcher for clients that don't support `didChangeWatchedFiles`.


internal/project/polling_watcher_test.go
```diff
@@ -0,0 +1,97 @@
...
+		FS:     wrappedFS,
+		Client: client,
+		Logger: logger,
+	})
+	defer session.Close()
+
+	// Open a file to trigger project creation
+	ctx := context.Background()
+	session.DidOpenFile(ctx, "file:///project/src/index.ts", 1, `const x: number = 1;`, lsproto.LanguageKindTypeScript)
+	time.Sleep(100 * time.Millisecond)
```
jakebailey: Rather than using sleeps, can we use `testing/synctest` (there are other files with examples). Not sure if it plays nice with the LSP code, I wrote my code that uses synctest in a way that avoided needing that infra

internal/project/watch_test.go
```diff
@@ -3,20 +3,21 @@ package project
 import (
 	"testing"
 
+	"github.com/microsoft/typescript-go/internal/tspath"
 	"gotest.tools/v3/assert"
 )
 
 func TestGetPathComponentsForWatching(t *testing.T) {
```
jakebailey: This test should be moved to the tspath package

internal/lsp/server.go
```diff
@@ -1163,6 +1163,7 @@ func (s *Server) handleInitialized(ctx context.Context, params *lsproto.Initiali
 			DebounceDelay:          500 * time.Millisecond,
 			PushDiagnosticsEnabled: !disablePushDiagnostics,
 			Locale:                 s.locale,
+			PollingEnabled:         !s.watchEnabled,
```
andrewbranch, line 1169: Can you add a user setting so we can easily test this (and so folks can opt into this watcher if that works better than their client's watcher)?

internal/project/session.go
```diff
@@ -232,6 +237,23 @@ func NewSession(init *SessionInit) *Session {
...
+		session.fileWatcher = vfswatch.NewFileWatcher(
+			init.FS,
+			pollInterval,
+			false,
+			session.onPolledFileChanges,
+		)
+		session.fileWatcher.UpdateWatchedDirectories(map[string]bool{
+			init.Options.CurrentDirectory: false,
+		})
+		go session.fileWatcher.Run(time.Now)
```
andrewbranch: For a second I thought this goroutine would never be cleaned up, but I see there's a separate `fileWatcher.Stop()` that will cause the loop to exit. I sort of expected this to take a `ctx` but maybe this is fine—Jake would be better at confirming what the most conventional handling would be here.

internal/project/session.go
```diff
@@ -1118,6 +1231,9 @@ func (s *Session) updateSnapshot(ctx context.Context, overlays map[tspath.Path]*
 				s.logger.Log(err)
 			}
 		}
+		if s.fileWatcher != nil {
+			s.updatePollingDirectories(newSnapshot)
```
andrewbranch, line 1236: So, it would be nice if this could share more of the existing infrastructure that feeds `s.updateWatches`. This runs every snapshot change, which includes every keystroke, but most of those trivial updates can't possibly affect watch locations. All the early returns, caching, and diffing is built into the `WatchedFiles` pipeline. The output of WatchedFiles is a glob, but in practice, it's always a recursive `**/*` glob, so it's a small modification to make it give you the directories directly.

internal/vfs/vfswatch/vfswatch.go
```diff
@@ -176,87 +192,78 @@ func hashEntries(entries vfs.Entries) uint64 {
...
 	ws := fw.watchState
-	wildcardDirs := fw.wildcardDirectories
+	dirs := fw.directories
 	fw.mu.Unlock()
-	return fw.hasChanges(ws, wildcardDirs)
+	current := snapshotDirectories(fw.fs, dirs)
+	return diffSnapshots(ws, current)
 }
 
 func (fw *FileWatcher) Run(now func() time.Time) {
```
andrewbranch: We should think about how to get some performance logging in here. I want to see how this works in big repos, but it's not really going to be visible what's going on under the hood.

internal/project/session.go
jakebailey: I much prefer when things are bounded by a context, given that will be cancelled at the very top.

(I'm also not sure what the deal is with passing `time.Now` as an arg here)

internal/project/session.go
```diff
@@ -367,6 +417,88 @@ func (s *Session) DidChangeWatchedFiles(ctx context.Context, changes []*lsproto.
...
+func (s *Session) updatePollingDirectories(snapshot *Snapshot) {
+	dirs := make(map[string]bool)
+
+	// Extract watched directories from the same WatchedFiles pipeline that
+	// updateWatches uses. The glob patterns are always "<dir>/**/*", so
+	// stripping the "/**/*" suffix gives us the directory path.
+	collectDirs := func(watchers []*lsproto.FileSystemWatcher) {
+		for _, w := range watchers {
+			glob := fileSystemWatcherGlobString(w)
+			if dir, ok := strings.CutSuffix(glob, "/**/*"); ok && dir != "" {
```
andrewbranch: I was more imagining a slight refactor of `WatchedFiles` that would give you this, but I think this implementation is fine if you just move the implementation onto `WatchedFiles` itself, so the implementation detail of the exact glob format can stay confined within that struct.

Looking good overall—I'll give it a try!

andrewbranch:
@johnfav03 have you tried this? It looks like setting `usePollingWatcher` doesn't stop the LSP client watcher from functioning at the same time. The other thing I noticed right off the bat is that it's descending into `.git/`. It takes 6.5 seconds to do a scan in `microsoft/vscode`; I'm trying to do some profiling and see what's going on. You might want to do some manual testing as well.

PR #3523, opened 2026-04-23
Fix hover expansion for aliased default function exports

Andarist:
fixes the crash found here: https://github.com/microsoft/typescript-go/issues/3512#issuecomment-4299730772


gabritto (commented):
This looks right, but any chance we could get a test for it?

Andarist:
oops, I had it and forgot to commit 🫠 fixed it now!

jakebailey (approved):

PR #3524, opened 2026-04-23
Fix hover verbosity for nested generic namespace merges

Andarist:
fixes crash found here: https://github.com/microsoft/typescript-go/issues/3512#issuecomment-4299732162

### DIFF
--- a/internal/checker/nodebuilderimpl.go
+++ b/internal/checker/nodebuilderimpl.go
@@ -1059,11 +1059,7 @@ func (b *NodeBuilderImpl) lookupTypeParameterNodes(chain []*ast.Symbol, index in
 		if typeArgumentNodes := b.lookupInstantiatedTypeArgumentNodes(chain, index); typeArgumentNodes != nil {
 			return typeArgumentNodes
 		} else {
-			typeParameterNodes := b.typeParametersToTypeParameterDeclarations(symbol)
-			if len(typeParameterNodes) > 0 {
-				return b.f.NewNodeList(typeParameterNodes)
-			}
-			return nil
+			return b.typeParametersToTypeParameterNodes(symbol)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/checker/nodebuilderimpl.go
```diff
@@ -1059,11 +1059,7 @@ func (b *NodeBuilderImpl) lookupTypeParameterNodes(chain []*ast.Symbol, index in
 		if typeArgumentNodes := b.lookupInstantiatedTypeArgumentNodes(chain, index); typeArgumentNodes != nil {
 			return typeArgumentNodes
 		} else {
-			typeParameterNodes := b.typeParametersToTypeParameterDeclarations(symbol)
```
jakebailey, line 1062: This seems like a departure from Strada, which just called `typeParametersToTypeParameterDeclarations` in its `lookupTypeParameterNodes`; why is this change needed?

internal/checker/nodebuilderimpl.go
```diff
@@ -1059,11 +1059,7 @@ func (b *NodeBuilderImpl) lookupTypeParameterNodes(chain []*ast.Symbol, index in
...
-			if len(typeParameterNodes) > 0 {
-				return b.f.NewNodeList(typeParameterNodes)
-			}
-			return nil
+			return b.typeParametersToTypeParameterNodes(symbol)
```
gabritto, line 1062: I don't think this completely fixes the issue. There are other code paths where we end up with type parameter declarations instead of type nodes (e.g. the branch immediately above this one). In Strada we already dealt with this on printing, since it was a consequence of allowing type arguments attached to identifiers.

gabritto: Never mind, I think this does fix everything up. I was confused because `lookupInstantiatedTypeArgumentNodes` has a confusing return type annotation. It would be nice to use the right type annotation (`TypeList`) for all of the functions involved here, to indicate intent (until we can hopefully check all this).

gabritto: I tested this change in Strada, and the problem is that if we use type nodes here instead of type parameter declarations, we no longer print type parameters with extends clause, e.g. `List<T extends Blah>` would get printed as `List<T>` now. So I think we just need to make the printer handle the type parameter declaration case.

gabritto:
Fixed in https://github.com/microsoft/typescript-go/pull/3606.

PR #3531, opened 2026-04-23
Preserve escaped identifier spelling in unresolved name diagnostics

### LINKED ISSUE #3511: Unicode escapes in identifier names now displayed as literal characters
conformance/parser.numericSeparators.decmialNegative.errors.txt.diff

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/3511


internal/binder/nameresolver.go
```diff
@@ -18,11 +18,15 @@ type NameResolver struct {
 	SetRequiresScopeChangeCache      func(node *ast.Node, value core.Tristate)
 	GetRequiresScopeChangeCache      func(node *ast.Node) core.Tristate
 	OnPropertyWithInvalidInitializer func(location *ast.Node, name string, declaration *ast.Node, result *ast.Symbol) bool
-	OnFailedToResolveSymbol          func(location *ast.Node, name string, meaning ast.SymbolFlags, nameNotFoundMessage *diagnostics.Message)
+	OnFailedToResolveSymbol          func(location *ast.Node, name string, nameNode *ast.Node, meaning ast.SymbolFlags, nameNotFoundMessage *diagnostics.Message)
```
jakebailey: I guess this is because strada accepted `string | Identifier`. I'd almost rather we just add a new arg and not duplicate this.

jakebailey: As in, not have `ResolveWithNode`

Andarist: Sure, I considered that but wanted to minimize the change. Even after changing this as requested, I feel this is a little bit quirky. The name itself can often be derived from the `nameNode` (and often times we could pass in a meaningful `nameNode` instead of relying on the `name` resolution at the caller side). I tried that refactor but it didn't sit right with me either - so for now, I refrained from pushing that out.

I think a clear split between `Resolve` and `ResolveByName` would actually be nice (so a "flipped" version of what I initially introduced in this PR).

PR #3532, opened 2026-04-23
Fix missing slices.Clone

andrewbranch:
Unsure if we've seen any symptoms of this bug; I happened upon it when investigating something else.


jakebailey (approved):
Yikes

PR #3537, opened 2026-04-23
Work around overzealous stack telemetry redaction

jakebailey:
Work around an overzealous regex in the VS Code telemetry client that would redact the entire thing because it contains a string like `getSignature(` or `GeyKey(`.

### DIFF
--- a/internal/lsp/stack_sanitizer.go
+++ b/internal/lsp/stack_sanitizer.go
@@ -11,7 +11,8 @@ import (
 // /(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]/i
 // as `<REDACTED: Generic Secret>`, which trips on innocuous Go frames like
 // `getSignatureHelp(`. Insert `X_X` after each trigger keyword to defeat the
-// regex; reverse with a `X_X` -> “ find-and-replace on the dashboard.
+// regex; reverse by removing the marker (replace `X_X` with the empty string)
+// on the dashboard.
 var genericSecretKeywordRegex = regexp.MustCompile(`(?i)(key|token|signature|sig|secret|password|passwd|pwd|android:value)([^a-zA-Z0-9])`)
 
 func defeatGenericSecretRegex(s string) string {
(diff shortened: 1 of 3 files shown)


internal/lsp/stack_sanitizer.go
```diff
@@ -1,11 +1,23 @@
...
 
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
+// VS Code's telemetry pipeline redacts any string matching
+// /(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]/i
+// as `<REDACTED: Generic Secret>`, which trips on innocuous Go frames like
+// `getSignatureHelp(`. Insert `X_X` after each trigger keyword to defeat the
+// regex; reverse with a `X_X` -> “ find-and-replace on the dashboard.
+var genericSecretKeywordRegex = regexp.MustCompile(`(?i)(key|token|signature|sig|secret|password|passwd|pwd|android:value)([^a-zA-Z0-9])`)
```
DanielRosenwasser: We *could* be conservative with the follow-up character (e.g. `(` and `[`).

I also have some weird paranoia around including `password` and the following ones.

internal/lsp/stack_sanitizer.go
```diff
@@ -1,11 +1,23 @@
...
 	"strings"
 
 	"github.com/microsoft/typescript-go/internal/core"
 )
 
+// VS Code's telemetry pipeline redacts any string matching
+// /(key|token|sig|secret|signature|password|passwd|pwd|android:value)[^a-zA-Z0-9]/i
+// as `<REDACTED: Generic Secret>`, which trips on innocuous Go frames like
+// `getSignatureHelp(`. Insert `X_X` after each trigger keyword to defeat the
+// regex; reverse with a `X_X` -> “ find-and-replace on the dashboard.
```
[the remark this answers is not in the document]
jakebailey: This is the stupid thing where `gofmpt` replaces `""` with `“`, sigh

DanielRosenwasser (approved):
I guess this is fine, we can always change it down the road.

internal/lsp/stack_sanitizer.go
jakebailey: I can cut this down to key/token/signature/sig.

DanielRosenwasser (approved):

Issue #3094, opened 2026-03-13
Declaration emit proceeds in the presence of TS7056


RyanCavanaugh:
## Steps to reproduce

**types.ts**
```ts
// Utility types from type-fest/remeda that create deeply nested conditional chains
type IsNever<T> = [T] extends [never] ? true : false;
type IsAny<T> = 0 extends 1 & NoInfer<T> ? true : false;
type If<T extends boolean, Then, Else> = IsNever<T> extends true ? Else : T extends true ? Then : Else;
type IfNotAnyOrNever<T, Then, IfAny = any, IfNever = never> = If<IsAny<T>, IfAny, If<IsNever<T>, IfNever, Then>>;
type Or<A extends boolean, B extends boolean> = If<IsNever<A>, false, A> extends true ? true : If<IsNever<B>, false, B> extends true ? true : false;
type CollapseLiterals<T> = {} extends T ? T : T extends infer U & {} ? U : T;
type Simplify<T> = { [K in keyof T]: T[K] } & {};

type UnionToIntersection<U> = (
  U extends unknown ? (x: U) => void : never
) extends (x: infer I) => void ? I & U : never;

type KeysOfUnion<T> = keyof UnionToIntersection<
  T extends unknown ? Record<keyof T, never> : never
>;

// Tag branding types — these create enough conditional depth to prevent type simplification
declare const tag: unique symbol;
type TagContainer<Token> = { readonly [tag]: Token };
type Tag<Token extends PropertyKey, Meta> = TagContainer<{ [K in Token]: Meta }>;
type Tagged<T, Name extends PropertyKey, Meta = never> = T & Tag<Name, Meta>;
type RemoveAllTags<T> = T extends Tag<PropertyKey, any>
  ? { [K in keyof T[typeof tag]]: T extends Tagged<infer U, K, T[typeof tag][K]> ? RemoveAllTags<U> : never }[keyof T[typeof tag]]
  : T;

// String literal detection with tag-aware unwrapping
type IsStringLiteral<S> = IfNotAnyOrNever<S,
  (CollapseLiterals<S extends TagContainer<any> ? RemoveAllTags<S> : S>) extends infer C extends string
    ? {} extends Record<C, never> ? false : true
    : false,
  false, false>;

// Recursive string splitting
type SplitHelper<S extends string, D extends string, Acc extends string[] = []> =
  S extends string ? D extends string ?
[...]

PR #3540, opened 2026-04-23
Triage or accept .js and .d.ts diffs

RyanCavanaugh:
(Branch name is wrong, .types will come later)

Proposed action for these baseline files, not for merging but rather for reviewing.


groups.md
```diff
@@ -0,0 +1,698 @@
+# JS Baseline Diff Categorization
+
+354 `.js.diff` files analyzed under `testdata/baselines/reference/submodule/`.
+
+---
+
+# Declaration emit now produces .d.ts output that was previously absent
+## TODO: Log a bug for this
+## Corsa emits declaration files for tests that previously had no .d.ts baseline output
```
jakebailey: This is a genuine bug I noticed; has to do with when we skip emit on errors and it did not seem trivial when I tried to fix it, yeah.

groups.md
```diff
@@ -0,0 +1,698 @@
...
+## TODO: Log a bug (adding `export` modifier changes the public API; previously-internal imports become visible to consumers)
+## `import fs2 = require("fs")` → `export import fs2 = require("fs")`
+conformance/nodeModulesAllowJsImportAssignment(module=node16).js.diff
+conformance/nodeModulesAllowJsImportAssignment(module=node18).js.diff
+conformance/nodeModulesAllowJsImportAssignment(module=node20).js.diff
+conformance/nodeModulesAllowJsImportAssignment(module=nodenext).js.diff
+
+# `@class`/`@constructor` no longer makes functions into class declarations
+## TODO: Log a bug (class properties and constructor signatures are lost; consumers see plain functions instead of classes)
+## Constructor functions annotated with @class/@constructor remain plain functions; prototype assignments do not produce classes
```
jakebailey: I had thought this was an intentional change, with our response being "It's 2026, please write a real class"

groups.md
```diff
@@ -0,0 +1,698 @@
...
+conformance/jsdocImplements_properties.js.diff
+conformance/jsdocImplements_missingType.js.diff
+conformance/lateBoundClassMemberAssignmentJS.js.diff
+conformance/lateBoundClassMemberAssignmentJS2.js.diff
+conformance/lateBoundClassMemberAssignmentJS3.js.diff
+conformance/thisPropertyAssignmentCircular.js.diff
+
+# CJS `module.exports = {}` now emits `export = _default` with inlined object type instead of named exports
+## TODO: Accept this (Documented in CHANGES.md in sections "CommonJS syntax" and "CommonJS")
+## CommonJS files using module.exports pattern changed from individual named exports to a single `export = _default` const
```
jakebailey: We probably need to check that this isn't semantically different, if `_default` is an object and not a namespace? e.g. mixing this with typedef exports

groups.md
```diff
@@ -0,0 +1,698 @@
...
+conformance/parserGetAccessorWithTypeParameters1(target=es2015).js.diff
+conformance/parserSetAccessorWithTypeAnnotation1(target=es2015).js.diff
+conformance/parserSetAccessorWithTypeParameters1(target=es2015).js.diff
+conformance/typeGuardFunctionErrors.js.diff
+
+# Decorator metadata reflects stable union type ordering
+## TODO: Log a bug (decorator metadata is runtime-accessible; changing `Object` → `String` affects `Reflect.getMetadata`)
+## `design:type` for `T | null` changed from `Object` to the non-null type (e.g. `String`, `Class1`)
+compiler/metadataOfUnionWithNull(strictnullchecks=true).js.diff
+compiler/metadataReferencedWithinFilteredUnion(strictnullchecks=true,target=es2015).js.diff
```
jakebailey: I think we should just accept this; these are cases where union ordering changed this data, and that isn't part of the contract.

groups.md
```diff
@@ -0,0 +1,698 @@
...
+
+# Empty namespace IIFE no longer emitted
+## TODO: Accept this (empty IIFEs are dead code with no runtime effect)
+## `namespace global {}` inside a class no longer generates an empty IIFE in JS output
+compiler/nestedGlobalNamespaceInClass.js.diff
+
+# JSX factory call changed from indirect `(0, _a.jsx)(...)` to direct `_jsx(...)` call
+## TODO: Log a bug
+## JSX emit no longer uses comma-expression indirection for factory calls
+compiler/commentsOnJSXExpressionsArePreserved(jsx=react-jsx,module=commonjs,moduledetection=legacy).js.diff
```
jakebailey: I'm iffy on this, because this is broken code

groups.md
```diff
@@ -0,0 +1,698 @@
...
+# CJS exports pattern changes in declaration emit
+## TODO: Log a bug (some files lose exports entirely, e.g. `jsDeclarationsExportDoubleAssignmentInClosure` emits only `export {}`)
+## Various CJS-specific declaration emit restructuring
+compiler/jsExportAssignmentNonMutableLocation.js.diff
+conformance/commonJSImportClassTypeReference.js.diff
+conformance/commonJSImportNestedClassTypeReference.js.diff
+conformance/jsDeclarationsExportDoubleAssignmentInClosure.js.diff
+
+# JS emit changes (non-declaration): module exports, __esModule, and variable scoping
+## `export const title` in labeled statement context no longer emits as bare semicolon
```
jakebailey: Missing a triage note here

groups.md
```diff
@@ -0,0 +1,698 @@
...
+## Changes in the emitted JavaScript runtime code rather than declaration files
+compiler/augmentExportEquals2.js.diff
```
jakebailey: This one is caused by duped files in the test case

groups.md
```diff
@@ -0,0 +1,698 @@
...
+compiler/importDeclWithExportModifierAndExportAssignment.js.diff
```
jakebailey: This is an error case; the old output was wrong and was emitting code that would crash due to `x` not existing as it's a type-only namespace

groups.md
```diff
@@ -0,0 +1,698 @@
...
+conformance/defaultExportsCannotMerge04(target=es2015).js.diff
```
jakebailey: This is a noop change

groups.md
```diff
@@ -0,0 +1,698 @@
...
+conformance/exportNonInitializedVariablesInIfThenStatementNoCrash1(module=commonjs).js.diff
```
jakebailey: But, this is real, and I have the change for this locally I'll send

groups.md
```diff
@@ -0,0 +1,698 @@
...
+
+# Numeric string property key uses computed property syntax `["404"]` instead of plain `"404"`
+## TODO: Accept this (`"404"` and `["404"]` declare the same property; consumers access it identically)
+## Declaration emit changes how numeric-string-keyed properties are represented
+compiler/declarationEmitPropertyNumericStringKey.js.diff
+
+# Non-literal computed property `[fieldName]` changed to index signature `[x: string]` in declaration emit
```
jakebailey: IIRC this is actually semantically equivalent, because we can't know what fieldName is and it's just string anyway?

groups.md
```diff
@@ -0,0 +1,698 @@
...
+# Declaration emit uses `import("...")` type instead of namespace-qualified references
+## TODO: Accept this (`import("./file").B` resolves to the same type as the namespace-qualified `A.B`)
+## `A.B` changed to `import("./file").B` in some declaration contexts
+compiler/es5ExportEqualsDts(target=es2015).js.diff
+compiler/declarationEmitNameConflicts.js.diff
+
+# Properties with `__proto__`-like names now quoted in declaration emit
+## TODO: Log a bug
+## `__proto__` and similar compiler-reserved names are now string-quoted as `"__proto__"`
+compiler/escapedReservedCompilerNamedIdentifier.js.diff
```
jakebailey: Iffy on this, one could argue that the new output is safer

groups.md
```diff
@@ -0,0 +1,698 @@
...
+
+# JS emit line-break reformatting in private name static method
+## TODO: Accept this (whitespace/comment positioning change; no semantic difference)
+## Comment `// Error` repositioned to next line in private field access method call
+conformance/privateNameStaticMethod.js.diff
+
+# Declaration emit type parameter renaming bug (regression)
+## TODO: Log a bug
+## Corsa incorrectly generates `T_1` where `T` is undefined, producing TS2304 DtsFileErrors
+compiler/declarationEmitShadowing.js.diff
```
jakebailey: This is definitely the worst bug in the list

groups.md
```diff
@@ -0,0 +1,698 @@
...
+
+# JSDoc-in-TypeScript parsing changes cause broken emit
+## TODO: Log a bug
+## JSDoc type annotations in `.ts` files now cause parsing issues resulting in broken JS output
+conformance/jsdocDisallowedInTypescript.js.diff
```
jakebailey: This is actually due to us removing `function` in jsdoc types, and then the follow-on errors changing.

I have a change locally that would restore that feature if we wanted

groups.md
RyanCavanaugh: Diff for both of these shows something material changing, not just union order
```diff
-    __metadata("design:type", Object)
+    __metadata("design:type", String)
```
for the input
```ts
    @PropDeco
    z: "foo" | undefined;
```

groups.md
RyanCavanaugh: It's `<div>{something weird}</div>` but the JSX emit that changed is on the outer `div` call, seems worth looking at

groups.md
RyanCavanaugh: Accepting since this is quoted in the original source file, so seems totally fine

groups.md
```diff
@@ -0,0 +1,696 @@
+# JS Baseline Diff Categorization
+
+354 `.js.diff` files analyzed under `testdata/baselines/reference/submodule/`.
+
+---
+
+# Declaration emit now produces .d.ts output that was previously absent
+## TODO: Log a bug for this
```
weswigham: You already did - #3094 

Actually easy to fix, it's like 4 lines in the emitter to align a bit closer with strada, issue is `-b` has already incorporated expecting decl emit in the presence of errors for determining up-to-dateness, which is more of a PITA to track down and fix, especially with now difficult to debug `-b` tests are.

groups.md
```diff
@@ -0,0 +1,696 @@
...
+conformance/nodeModulesExportsBlocksSpecifierResolution(module=node18).js.diff
+conformance/nodeModulesExportsBlocksSpecifierResolution(module=node20).js.diff
+conformance/nodeModulesExportsBlocksSpecifierResolution(module=nodenext).js.diff
+conformance/nodeModulesExportsSourceTs(module=node16).js.diff
+conformance/nodeModulesExportsSourceTs(module=node18).js.diff
+conformance/nodeModulesExportsSourceTs(module=node20).js.diff
+conformance/nodeModulesExportsSourceTs(module=nodenext).js.diff
+
+# JS declaration emit now uses `export declare` instead of bare `export`
+## TODO: Accept this (`declare` is optional and moot in declarations in .d.ts files)
```
weswigham: Just note that it's actually pretty easy to just tell the declaration emitter to drop the `declare` modifiers in js file emit, if we *really* wanted to match (just a small edit to `ensureModifiers`), it would just be odd to have an input-extension-specific behavior for this. AFAIK, the reason we did in the old js declaration emitter was because we decided that was "better" since the modifier is completely extraneous in a declaration file. So we should probably make an issue to *always* drop the declare modifier (in TS as well as JS) in the future, no matter how we decide to handle this for now.

jakebailey: For diffing's sake, I think we should probably restore them because they make diff perusal super noisy (and we are asking people downstream to consider diffing their outputs to look for bugs)

groups.md
```diff
@@ -0,0 +1,696 @@
...
+conformance/nodeModulesAllowJsDynamicImport(module=node20).js.diff
+conformance/nodeModulesAllowJsDynamicImport(module=nodenext).js.diff
+conformance/nodeModulesAllowJsSynchronousCallErrors(module=node16).js.diff
+conformance/nodeModulesAllowJsSynchronousCallErrors(module=node18).js.diff
+conformance/nodeModulesAllowJsSynchronousCallErrors(module=node20).js.diff
+conformance/nodeModulesAllowJsSynchronousCallErrors(module=nodenext).js.diff
+conformance/nodeAllowJsPackageSelfName2.js.diff
+
+# JS declaration emit splits `export const/function` into `declare` + separate `export {}` statement
+## TODO: Accept this (`declare const` + `export { x }` is semantically equivalent to `export const`)
```
weswigham: The old js declaration emitter had an extra pass to inline these to make em look pretty. So if you wrote
```js
class A {}
module.exports.A = A
```
first we'd generate
```ts
declare class A {}
export {A}
```
because that works for all forms of the export statement. Then we'd collect all the shorthand exports, elide them, and tack an `export` on the things they pointed at to produce
```ts
export class A {}
```
for the final output. It's not hard, and it was done syntactically in strada, and we could add a pass to do it in corsa, too, it just doesn't actually change any semantic meaning. Just depends on how pretty you like your output. In reality, putting effort into the prettyness of JS declaration emit is probably not worth it at this point, since any modern JS you generate declarations for should be esm syntax anyway and not cjs, so there's nothing to really pretty-ify.

groups.md
```diff
@@ -0,0 +1,696 @@
...
+## e.g. `export const x: 1` → `declare const x = 1; export { x };`
```
weswigham: This one in particular is actually an improvement in corsa. The old form failed to preserve literal freshness of the initializer into the declaration file, which can produce intermittent incremental build errors.

groups.md
weswigham: `[fieldName]` (where `declare const fieldName: string`) is a ID-compatible transform (that we specifically started to allow in recent TS versions), whereas `[x: string]` isn't. They may typecheck the same, but you have to emit the former instead of the later under ID. (And because ID isn't supposed to affect the output, you're supposed to emit it all the time)

groups.md
RyanCavanaugh: Logged https://github.com/microsoft/typescript-go/issues/3585, moving this to triaged

groups.md
ahejlsberg: This is intentional and documented in `CHANGES.md`.

jakebailey (approved):
I think we should take this so we can start async fixing or re-triaging.

PR #3541, opened 2026-04-23
Collect this property assignments in JS declaration emit and transform them to TS declarations

### LINKED ISSUE #3456: JS declaration emit ignores properties declared through `this.xxx` assignments
Failing tests include:

compiler\argumentsReferenceInConstructor1_Js.js.diff
compiler\argumentsReferenceInConstructor2_Js.js.diff
compiler\argumentsReferenceInConstructor3_Js.js.diff
compiler\argumentsReferenceInConstructor4_Js.js.diff
compiler\argumentsReferenceInConstructor5_Js.js.diff
compiler\argumentsReferenceInConstructor6_Js.js.diff
compiler\argumentsReferenceInConstructor7_Js.js.diff
compiler\argumentsReferenceInMethod1_Js.js.diff
compiler\argumentsReferenceInMethod2_Js.js.diff
compiler\argumentsReferenceInMethod3_Js.js.diff
compiler\argumentsReferenceInMethod4_Js.js.diff
compiler\argumentsReferenceInMethod5_Js.js.diff
compiler\argumentsReferenceInMethod6_Js.js.diff
compiler\argumentsReferenceInMethod7_Js.js.diff

weswigham:
Fixes #3456 

And then a bit more, since looking through the changed diffs I noticed we also didn't add a transform of `const x = require("y")` statements in #3388 to replace the previously reparsed import. I don't wanna get too much more off into the weeds on this, since in doing that I the noticed that `jsDeclarationsExportAssignedVisibility.ts` demonstrates that we're _also_ missing robustness in the `module.exports = ...` declaration transform to handle hoisting out class expressions and creating an alias to them (unless we don't export type meanings from those now or something, in which case, intentional? but not doing so makes broken declarations so... not intentional?), which is really a ways away from handling `this.` assignments correctly.

As an aside, the new declaration logic should support it, but it looks like we actually support different JS expando assignments in corsa than in strada. Namely, strada doesn't recognize `this[something] = "whatever"` in class static blocks, but corsa does. This *probably* needs test coverage, both for checking and declaration emit, since adding declaration emit support for it didn't seem to change anything. This.... isn't too surprising, given the strada bug referenced in the linked issue.


internal/transformers/declarations/transform.go
```diff
@@ -507,6 +507,10 @@ func (tx *DeclarationTransformer) visitDeclarationSubtree(input *ast.Node) *ast.
 		return nil
 	}
 
+	if ast.IsHeritageClause(input) && (len(input.AsHeritageClause().Types.Nodes) == 0 || (len(input.AsHeritageClause().Types.Nodes) == 1 && ast.NodeIsMissing(input.AsHeritageClause().Types.Nodes[0]))) {
```
weswigham, line 510: This was also a bit of a driveby fix to elide empty heritage clauses, which one of the modified jsdoc tests tests for. I get why we got it for "free" in the old js emitter (since the type has no base types in the checking phase), but what I don't get is how TS doesn't also have the same issue. Maybe it does and we just don't test declaration output for `class A implements {}` with suppressed parse errors? In any case, it's handled now.

testdata/baselines/reference/submoduleTriaged/compiler/argumentsReferenceInConstructor3_Js.js.diff
jakebailey, line 27: For the changing files, we can probably move them from Triaged to Accepted, like this one which now only differs in ordering?

Not sure how easy it'd be to restore the old order just for ease of diffing, that would be nice too.

jakebailey: Perhaps slotting them in after the contructor would suffice?

weswigham: So the old js emitter would always emit index signatures, then statics, then constructor declarations, then public props, then private props. Unless we wanna resort all the members (nested and non-nested) and stop using what's essentially "the order the user wrote them", it's not gonna match strada.

jakebailey: It's not a big deal, we'll just have to go manually check things to silence those we know are just ordering differences. Was just hopeful it would be straightforward.

weswigham: No, definitely not so straightforward, unfortunately. We could always decide to do some kind of pretty-ing sort later (I imagine people have lint rules enforcing some kinda sort on their inputs fairly often), but it'd be kinda weird to do the sort for js declarations but not ts ones (especially since TS also has expando decls on functions, just not `this.` decls on classes).

jakebailey (approved):
LGTM however, we should move whatever looks good to accepted if possible

testdata/submoduleAccepted.txt
```diff
@@ -1181,6 +1181,23 @@ conformance/nodeModulesAllowJsPackageImports(module=node18).js.diff
 conformance/nodeModulesAllowJsPackageImports(module=node20).js.diff
 conformance/nodeModulesAllowJsPackageImports(module=nodenext).js.diff
 
+## this.x member ordering changed
+## Strada grouped js declaration members by catecory, corsa keeps them in source order
```
jakebailey:
```suggestion
## Strada grouped js declaration members by category, corsa keeps them in source order
```

jakebailey (approved):

PR #3573, opened 2026-04-24
Optimize type variable scope map cloning with CoW

jakebailey:
This restores the copy-on-write in the original Strada code, which saves the creation of new maps. On the testrunner test suite, this saves about 720k allocations.


weswigham (approved):
I'm wondering if we can do away with the local management of the has-written state and actually just have a CoW map type now, though, seeing as we have pointers in `go`...

jakebailey:
Not a bad idea to formalize this, I'll give it a go

weswigham (approved):

PR #3574, opened 2026-04-24
fix(3565): preserve trailing commas in declaration-emitted array binding patterns

### LINKED ISSUE #3565: Baselines: Declaration emit trailing comma changes in destructuring parameters
Extra trailing commas after omitted elements removed: `[, z, ,]` becomes `[, z, ]`.

## Affected Files
``
compiler/declarationEmitDestructuring5.js.diff
```

### PR DESCRIPTION
**Pull request by a-tarasyuk**
Fixes #3565

### CODE REVIEW

File: internal/transformers/declarations/transform.go [hunk 1]
```diff
@@ -1841,6 +1841,7 @@ func (tx *DeclarationTransformer) filterBindingPatternInitializers(node *ast.Nod
 			))
 		}
 		elemList := tx.Factory().NewNodeList(elements)
+		elemList.Loc = node.AsBindingPattern().Elements.Loc
```
jakebailey: Wonder why Strada did not need to do this in its `filterBindingPatternInitializers`

a-tarasyuk: `filterBindingPatternInitializers` uses `visitNodes` preserves the `NodeArray` `pos`/`end` https://github.com/microsoft/TypeScript/blob/55423abe4d029017f19b6e4c32097591994836b4/src/compiler/visitorPublic.ts#L245-L257

jakebailey: Hm, why aren't we using VisitNodes here, then? Our visitor also preserves that.

internal/transformers/declarations/transform.go
```diff
@@ -1816,33 +1816,25 @@ func (tx *DeclarationTransformer) ensureNoInitializer(node *ast.Node) *ast.Node
...
 
-			elements = append(elements, tx.Factory().UpdateBindingElement(
-				elem.AsBindingElement(),
-				elem.AsBindingElement().DotDotDotToken,
-				elem.PropertyName(),
-				tx.filterBindingPatternInitializers(elem.Name()),
-				nil,
-			))
+	var visitor *ast.NodeVisitor
+	visitor = tx.EmitContext().NewNodeVisitor(func(node *ast.Node) *ast.Node {
```
jakebailey: This visitor needs to be hoisted into a field on the transformer, otherwise we will repeatedly create visitors and closures (see this pattern elsewhere)

testdata/submoduleTriaged.txt
```diff
@@ -254,7 +254,6 @@ conformance/genericFunctionParameters.js.diff
 
 # Declaration emit trailing comma changes in destructuring parameters
```
jakebailey: Delete the whole section

jakebailey (approved):

Issue #3576, opened 2026-04-24
[Proposal] strictArrayVariance: opt-in invariance for mutable Array<T> element type (post-7.0 discussion)


JustFly1984:
> **Filing under the `CONTRIBUTING.md` 7.0 freeze as a design discussion for post-7.0, not a merge request.** A draft implementation is available at #3577 for reference only.

## Summary

Mutable `Array<T>` is treated as **covariant** in `T` when relating two instantiations of the same generic target (the precomputed `arrayVariances` slot and the `getVariances` shortcut for `globalArrayType`). That matches historical TypeScript (Strada) but is **unsound** when an array is passed to code that may **mutate** through the reference. `ReadonlyArray<T>` does not expose the same mutation surface on the array object itself, so covariance there aligns with a read-only view.

## Minimal repro

```ts
function widen(xs: (string | number)[]) {
  xs.push(3);
}

const ys: string[] = ["a"];
widen(ys); // accepted today; the callee mutates `ys` to hold a number
```

A stricter checker would reject `widen(ys)` — either across the board (breaking change) or behind an explicit opt-in compiler flag (what this proposal does).

## Motivation and prior discussion

The soundness gap above has been discussed publicly — e.g. a recent podcast segment at <https://youtu.be/urYhuK6m_O4?t=3622> (~1:00:22–1:22:46) covers the same `push`-through-a-parameter scenario and proposes dispatch-site checking with memoization as one possible fix. Linked for context; the technical case here does not depend on that discussion.

This proposal is the **pragmatic variance-based version** of the "reject widening when the callee can mutate" idea for Corsa / TS 7+: stay on the fast variance path, but invert the variance of mutable `Array<T>`'s element parameter under an explicit flag. Dispatch-site checking is discussed below as Option B / future work.

### Option A (this draft PR): flag-gated invariance for mutable containers

- New `compilerOptions.strictArrayVariance` — default **off**, **not** implied by `--strict`.
- Under the flag, treat **`Array<T>`**'s element type parameter and **mutable tuple** element positions as **invariant**.
- **`ReadonlyArray<T>`** and **readonly tuples** stay **covariant** (no mutation surface on the array object itself).
- The diagnostic chain includes a dedicated hint pointing users at `ReadonlyArray<T>` at the parameter boundary, or a widened source declaration when mutation is genuinely intended.

### Option B (future work): dispatch-site checking with cache

[...]

PR #3577, opened 2026-04-24
feat(checker): strictArrayVariance flag — invariant mutable Array<T> element type (DRAFT for post-7.0 discussion)

JustFly1984:
> 🚧 **Draft — filed under the `CONTRIBUTING.md` 7.0-freeze policy for discussion only.** Not requesting merge in the 7.0 cycle. Tracking issue: #3576.

## Summary

- New compiler option `strictArrayVariance` — default **off**, **not** implied by `--strict`.
- Under the flag, the element type parameter of `globalArrayType` and mutable tuple element positions are treated as **invariant** when relating instantiations.
- `ReadonlyArray<T>` and readonly tuples remain **covariant** — no change to the read-only view.

## Motivation

Mutable `Array<T>` is covariant in `T` today, which is unsound when a callee can mutate through the reference:

```ts
function widen(xs: (string | number)[]) { xs.push(3); }
const ys: string[] = ["a"];
widen(ys); // accepted; the callee mutates `ys` to hold a number
```

Full design discussion, the `lib.dom.d.ts` adoption blocker, and maintainer questions live on the tracking issue (#3576).

## Implementation notes

- `internal/checker/checker.go` — `arrayVariances` keeps its single covariant slot; a per-`Checker` cache of invariant tuple variances (`strictInvariantTupleVariances`) is populated on demand when the flag is on.
- `internal/checker/relater.go` — `getVariances` returns invariant variances for `globalArrayType` and mutable tuple targets when `c.compilerOptions.StrictArrayVariance == core.TSTrue`. `tryElaborateStrictArrayVarianceHint` emits the dedicated diagnostic.
- `internal/tsoptions/declscompiler.go` — option declaration; `strictFlag` **intentionally omitted** (not implied by `--strict`). An inline comment points at the test that pins that behavior.
- `internal/diagnostics/extraDiagnosticMessages.json` — adds the option-description message (100019) and the dedicated hint (100020). The hint text points at `ReadonlyArray<T>` or a wider source declaration — **not** at `[...arr]`, because bare spread does not widen under contextual typing.
- `internal/project/session.go` — telemetry capture for the new option, mirroring existing `strict*` handling.

## Test plan

- [x] `go test ./internal/checker/...`
- [x] `go test ./internal/tsoptions/...`
- [x] `go test ./internal/core/...`
- [x] `go test ./internal/project/...`
- [x] `go test ./internal/testrunner/...` (incl. `TestLocal/strictArrayVariance.ts_strictarrayvariance={true,false}`)
- [x] Manual UX reproduction: `widen(ys)` accepted with `--strict`; rejected with `--strict --strictArrayVariance` carrying the new hint.

Baselines regenerated via `hereby baseline-accept`. The compiler test case (`testdata/tests/cases/compiler/strictArrayVariance.ts`) covers: mutable `T[]` widening, `ReadonlyArray` preservation, `Array<T>` syntax, type-alias to `Array<T>`, mutable tuples, nested arrays, structural assignment through a containing object, inference via `infer U` and generic identity functions, the readonly-parameter migration target, callback contravariance, and documents that bare `[...arr]` does **not** widen.

## Migration notes

For code that needs to keep compiling under the flag:

- **Prefer `ReadonlyArray<T>` / `readonly T[]` at parameter boundaries** if reads suffice. This is the primary migration target and is exercised by the test case.
- If the callee genuinely needs to mutate, **declare the source with the wider element type from the start**.
- Bare `[...arr]` does not widen — contextual typing preserves the spread's source type. Do not advertise it as a workaround.

## Known gaps / open questions

See the tracking issue #3576. The largest is the `lib.dom.d.ts` adoption blocker: types like `HTMLCollectionOf<T>` extending `HTMLCollectionBase` fail under invariance, so users cannot turn the flag on in projects that import DOM types without a coordinated lib pass.

## Cross-links

- Tracking issue: #3576
- Podcast context (linked, not relied on for the technical case): <https://youtu.be/urYhuK6m_O4?t=3622>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -112,6 +112,12 @@ Corsa no longer parses the following JSDoc tags with a specific node type. They 
 
 ## Checker
 
+### `strictArrayVariance`
+
+When `--strictArrayVariance` / `"strictArrayVariance": true` is set, relating two `Array<T>` instantiations treats `T` as **invariant** instead of covariant. Mutable tuple element positions are likewise treated as invariant at their fixed positions. That rejects unsound assignments such as passing `string[]` where `(string | number)[]` is expected, or `[string, string]` where `[string | number, string]` is expected, when the callee could mutate the argument. `ReadonlyArray<T>` and `readonly` tuples are unchanged (still covariant). The option defaults to **off** so behavior matches TypeScript 6.0 unless you opt in.
+
+When the check rejects a call or assignment, the diagnostic chain includes a dedicated hint suggesting `ReadonlyArray<T>` at the parameter type (if reads suffice), or declaring the source with the wider element type when a mutating callee is genuinely intended.
+
 ### Miscellaneous
 
 #### With `"strict": false`, Corsa no longer allows omitting arguments for parameters with type `undefined`, `unknown`, or `any`:
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -601,6 +601,7 @@ type Checker struct {
 	strictBindCallApply                         bool
 	strictPropertyInitialization                bool
 	strictBuiltinIteratorReturn                 bool
+	strictArrayVariance                         bool
 	noImplicitAny                               bool
 	noImplicitThis                              bool
 	useUnknownInCatchVariables                  bool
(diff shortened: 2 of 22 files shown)


JustFly1984:
@microsoft-github-policy-service agree [company="JustFly1984"]

JustFly1984:

JustFly1984:
@microsoft-github-policy-service agree company="JustFly1984"

jakebailey:
https://github.com/microsoft/typescript-go/issues/3576#issuecomment-4314994623

Issue #3557, opened 2026-04-23
Baselines: @satisfies tag handling changes: functions become const declarations


RyanCavanaugh:
fn4-fn6 parameter types change, e.g. `b: number` becomes `b: never`; visible type differences. Functions with @satisfies become `declare const` with arrow types and added @param/@satisfies annotations.

## Affected Files
``
conformance/checkJsdocSatisfiesTag15.js.diff
```

sandersn:
1. The .js is consts with arrows. So emit as const with arrow types is correct.
2. The problem repros in the editor, so it's not a declaration emit bug. My best guess is that the reparser is incorrectly attaching `satisfies` to the variable declaration instead of the arrow expression.

sandersn:
Later: nope it was the other way round. satisfies' synthetic expression blocked the later attachment of param tags.

PR #3578, opened 2026-04-24
Preserve JSDoc parameter types on functions wrapped by `@satisfies`

Andarist:
relates to https://github.com/microsoft/typescript-go/issues/3557 , but it doesnt fix everything from that diff yet


sandersn:
Sorry, didn't see this yesterday because I was moving too fast. #3746 is basically this PR but slightly easier to read.

PR #3581, opened 2026-04-24
Restore #3522 and #3516

jakebailey:
#3522 and #3516 were lost in the GitHub merge queue bug yesterday.


andrewbranch (approved):

jakebailey:
@andrewbranch I think those comments might be correct for #3516, reading...

PR #3590, opened 2026-04-24
Fix auto-import bucket state race and CreateProgram/UpdateProgram race

### LINKED ISSUE #2582: Auto-import does not suggest updates from symlinked monorepo package sources
### Extension Version

0.20260124.1

### VS Code Version

1.108.2

### Operating system Version

macOS 15.7

### Steps to reproduce

I've written a failing fourslash test to demonstrate the issue:

```golang
package fourslash_test

import (
	"testing"

	"github.com/microsoft/typescript-go/internal/fourslash"
	. "github.com/microsoft/typescript-go/internal/fourslash/tests/util"
	"github.com/microsoft/typescript-go/internal/ls"
	"github.com/microsoft/typescript-go/internal/lsp/lsproto"
	"github.com/microsoft/typescript-go/internal/testutil"
)

const TestAutoImportSymlinkedMonorepoSourceUpdateScenario = `
// @Filename: /home/src/workspaces/project/tsconfig.base.json
{
  "compilerOptions": {
    "module": "nodenext",
    "moduleResolution": "nodenext",
	"composite": true
  }
}

// @Filename: /home/src/workspaces/project/packages/foo/package.json
{
  "name": "@packages/foo",
  "type": "module",
  "exports": {
    ".": {
	  "types": "./src/index.ts",
	  "default": "./dist/index.js"
	}
  }
}

// @Filename: /home/src/workspaces/project/packages/foo/tsconfig.json
{ "extends": "../../tsconfig.base.json" }

// @Filename: /home/src/workspaces/project/packages/foo/src/index.ts
/*fooEdit*/

// @Filename: /home/src/workspaces/project/packages/bar/package.json
{
  "name": "@packages/bar",
  "type": "module",
  "exports": {
    ".": {
	  "types": "./src/index.ts",
	  "default": "./dist/index.js"
	}
  }
  "dependencies": {
    "@packages/foo": "*"
  }
}

// @Filename: /home/src/workspaces/project/packages/bar/tsconfig.json
{ "extends": "../../tsconfig.base.json" }

// @Filename: /home/src/workspaces/project/packages/bar/src/index.ts
/*fooCompletion*/

// @Filename: /home/src/workspaces/project/package.json
{ "workspaces": ["packages/*"], "type": "module" }

// @link: /home/src/workspaces/project/packages/bar -> /home/src/workspaces/project/node_modules/@packages/bar
// @link: /home/src/workspaces/project/packages/foo -> /home/src/workspaces/project/node_modules/@packages/foo
`

func TestAutoImportSymlinkedMonorepoSourceUpdate(t *testing.T) {
	t.Parallel()

	f, done := fourslash.NewFourslash(t, nil /*capabilities*/, TestAutoImportSymlinkedMonorepoSourceUpdateScenario)

	defer done()

	// force auto import to build the cache
	f.BaselineAutoImportsCompletions(t, []string{"fooCompletion"})

	// add the new function
	f.GoToMarker(t, "fooEdit")
	f.Insert(t, "\nexport function foo() {}")

	// function should show up
	f.VerifyCompletions(t, "fooCompletion", &fourslash.CompletionsExpectedList{
		IsIncomplete: false,
			CommitCharacters: &DefaultCommitCharacters,
			EditRange:        Ignored,
		},
		Items: &fourslash.CompletionsExpectedItems{
			Includes: []fourslash.CompletionsExpectedItem{
				&lsproto.CompletionItem{
					Label:               "foo",
					AdditionalTextEdits: fourslash.AnyTextEdits,
					SortText:            PtrTo(string(ls.SortTextLocationPriority)),
				},
			},
		},
	})
}

```

I placed it at `internal/fourslash/tests/autoImportSymlinkedMonorepoSourceUpdate_test.go`.

### Issue

It seems that the auto-import cache doesn't update for edits to symlinked workspace packages in a monorepo when `types` in a `package.json` points to the source Typescript. I would expect that it would, as this would make working in a monorepo pretty seamless for developers.

I'm happy to give more context about the related issues I've filed (https://github.com/microsoft/typescript-go/issues/2581, and one in the `Typescript` repository) if that helps!

andrewbranch:
BucketStates contained pointers that were being mutated on what should have been immutable RegistryBuckets. Added a deep clone of those fields to `(*RegistryBucket).Clone`. Other fields looked suspicious but are in fact always fully reassigned, so added some comments about that too.

Fixes the races that started showing up after #3057 

Closes #2582 again

### DIFF
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -341,10 +341,8 @@ func (p *Project) CreateProgram() CreateProgramResult {
 	// CreateProgram must use its own closure so that concurrent goroutines cloning
 	// the same project never share a captured variable through a stale closure
 	// stored in the old program's options.
-	// Capture only the log function, not the entire *Project receiver.
-	log := p.log
 	createCheckerPool := func(program *compiler.Program) compiler.CheckerPool {
-		checkerPool = newCheckerPool(4, program, log)
+		checkerPool = newCheckerPool(4, program, nil)
 		return checkerPool
 	}
 
--- a/internal/project/project.go
+++ b/internal/project/project.go
@@ -337,11 +337,22 @@ func (p *Project) CreateProgram() CreateProgramResult {
 	var checkerPool *CheckerPool
 	var newProgram *compiler.Program
 
+	// Define a fresh CreateCheckerPool closure for this call. Each invocation of
+	// CreateProgram must use its own closure so that concurrent goroutines cloning
+	// the same project never share a captured variable through a stale closure
+	// stored in the old program's options.
+	// Capture only the log function, not the entire *Project receiver.
+	log := p.log
+	createCheckerPool := func(program *compiler.Program) compiler.CheckerPool {
+		checkerPool = newCheckerPool(4, program, log)
+		return checkerPool
+	}
+
 	// Create the command line, potentially augmented with typing files
 	commandLine := p.getCommandLineWithTypingsFiles()
 
 	if p.dirtyFilePath != "" && p.Program != nil && p.Program.CommandLine() == commandLine {
-		newProgram, programCloned = p.Program.UpdateProgram(p.dirtyFilePath, p.host)
+		newProgram, programCloned = p.Program.UpdateProgram(p.dirtyFilePath, p.host, createCheckerPool)
 		if programCloned {
 			updateKind = ProgramUpdateKindCloned
 			for _, file := range newProgram.SourceFiles() {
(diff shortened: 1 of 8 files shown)


jakebailey (approved):

internal/project/project.go
```diff
@@ -370,10 +379,7 @@ func (p *Project) CreateProgram() CreateProgramResult {
 				Config:                      commandLine,
 				UseSourceOfProjectReference: true,
 				TypingsLocation:             typingsLocation,
-				CreateCheckerPool: func(program *compiler.Program) compiler.CheckerPool {
-					checkerPool = newCheckerPool(4, program, p.log)
```
andrewbranch, line 374: This was a data race and a memory leak in one line! `p.log` retains the receiver `p` which retains the in-memory file system of the snapshot that created this closure. Since unchanged files are shared between snapshots and programs get replaced all the time, this was never likely to account for much wasted memory, but 🤷

jakebailey (approved):

internal/project/project.go
andrewbranch: Actually, I think this retain path is no different from the one that exists through CompilerHost, which is expected, so, never mind.

PR #3609, opened 2026-04-25
Fix hover for optional method members in types and interfaces

### LINKED ISSUE #3607: Missing hover for nullable method in type and interface
### Extension Version

0.20260425.1

### VS Code Version

1.114.0

### Operating system Version

Microsoft Windows NT 10.0.26200.0 x64

### Steps to reproduce

1.

```ts
type Foo = {
    /*1*/ bar?(): void
}

interface IFoo {
    /*2*/ bar?(): void
}

declare const foo1: Foo
declare const foo2: IFoo

foo1./*3*/ bar
foo2./*4*/ bar
```

2.

Hover on the four `bar`s above, none shows the hover.

### Issue

Expected to show:
`/*1*/`:
```ts
(method) bar?(): void
```
`/*2*/`:
```ts
(method) IFoo.bar?(): void
```
`/*3*/`:
```ts
(method) bar?(): void
```
`/*4*/`:
```ts
(method) IFoo.bar?(): void
```

MukundaKatta:
Closes #3607.

Hover skipped optional method members like `bar?(): void` because the call-signature lookup ran `GetSignaturesOfType(GetTypeOfSymbol(symbol))`. Optional methods carry a `T | undefined` type, signatures are not enumerated on unions, so the hover collapsed to empty for `i.bar`, `t.bar`, `c.baz`. When the lookup is empty and the symbol is optional, recover signatures from the function-like declarations so optional methods render the same as required ones.

Test fixture: `internal/fourslash/tests/hoverOptionalMethod_test.go` covers interface, type literal, and class cases.

### DIFF
--- a/internal/fourslash/tests/hoverOptionalMethod_test.go
+++ b/internal/fourslash/tests/hoverOptionalMethod_test.go
@@ -1,0 +1,43 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
+	"github.com/microsoft/typescript-go/internal/testutil"
+)
+
+func TestHoverOptionalMethod(t *testing.T) {
+	t.Parallel()
+	defer testutil.RecoverAndFail(t, "Panic on fourslash test")
+	const content = `
+// @strict: true
+interface I {
+    bar?(): void
+}
+type T = {
+    bar?(): void
+}
+class C {
+    baz?() {}
+}
+
+declare const i: I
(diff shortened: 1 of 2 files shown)


ahejlsberg:
@MukundaKatta Appreciate the PR, but this isn't quite the right fix. See #3620.

MukundaKatta:
Thanks @ahejlsberg, makes sense. #3620 looks like the cleaner fix at the source. Closing this in favor of yours.

PR #3614, opened 2026-04-26
fix(3613): allow prop–accessor merging in interfaces

### LINKED ISSUE #3613: Behaviour difference: merging `interface`s with identically named getter/setter pair and field
Found from typechecking a project using `@mikro-orm+core@6.6.13`. Behaviour difference also occurs when using `declare class`.

## Steps to reproduce

issue.ts:

```typescript
interface EntityMetadata {
    get tableName(): string;
    set tableName(name: string);
}

interface EntityMetadata {
    tableName: string;
}
```

No error.

```
$ pnpm dlx --package=typescript@6 tsc --ignoreConfig issue.ts
```

## Behavior with `tsgo`

TS2300 (duplicate identifier) errors:

```
$ pnpm dlx --package=@typescript/native-preview@beta tsgo --ignoreConfig issue.ts
issue.ts:2:9 - error TS2300: Duplicate identifier 'tableName'.

2     get tableName(): string;
          ~~~~~~~~~

issue.ts:3:9 - error TS2300: Duplicate identifier 'tableName'.

3     set tableName(name: string);
          ~~~~~~~~~

issue.ts:7:5 - error TS2300: Duplicate identifier 'tableName'.

7     tableName: string;
      ~~~~~~~~~


Found 3 errors in the same file, starting at: issue.ts:2
```

a-tarasyuk:
Fixes #3613


testdata/tests/cases/compiler/duplicateIdentifierChecks2.ts
```diff
@@ -0,0 +1,12 @@
+// @strict: true
+// @target: esnext
+// @noEmit: true
+
+interface I {
```
mcpower: drive-by comment: Should there be tests for `declare class` too? The source of the original issue was from the .d.ts files for [this class accessor](https://github.com/mikro-orm/mikro-orm/blob/d656072be416ff8133d02738c12b66672c326f15/packages/core/src/typings.ts#L1200) and [this interface property](https://github.com/mikro-orm/mikro-orm/blob/d656072be416ff8133d02738c12b66672c326f15/packages/core/src/typings.ts#L1401).

ahejlsberg:
Unfortunately this isn't the right fix. Like Strada, it only gets it right in some cases.

a-tarasyuk:
@ahejlsberg, do you mean this change applies only to cases where the accessors are declared before the property signature?

ahejlsberg:
> do you mean this change applies only to cases where the accessors are declared before the property signature?

Yes.

PR #3636, opened 2026-04-28
Fix signature help crash in WIP JSX text on opening `<`

### LINKED ISSUE #3635: When entering "<" in it, it will panic
## Stack trace

```
2026-04-28 12:46:00.999 [error] panic handling request textDocument/signatureHelp: Debug failure. False expression: Not a subspan. Child: KindLessThanLessThanToken, parent: KindJsxSelfClosingElement
goroutine 22623 [running]:
runtime/debug.Stack()
github.com/microsoft/typescript-go/internal/lsp.(*Server).recover(0x145631943c08, 0x14564073f7d0)
	github.com/microsoft/typescript-go/internal/lsp/server.go:940 +0x47
panic({0x43fb2a0?, 0x1456409c2200?})
	runtime/panic.go:860 +0x13a
github.com/microsoft/typescript-go/internal/debug.Fail(...)
	github.com/microsoft/typescript-go/internal/debug/debug.go:14
github.com/microsoft/typescript-go/internal/debug.assertSlow({0x14563af13c78?, 0x19?, 0x10?})
	github.com/microsoft/typescript-go/internal/debug/debug.go:60 +0x54
github.com/microsoft/typescript-go/internal/debug.Assert(...)
	github.com/microsoft/typescript-go/internal/debug/debug.go:49
github.com/microsoft/typescript-go/internal/ls.getContainingArgumentInfo(0x145640402008?, 0x145640402008, 0x145640408008, 0x0, 0x49)
	github.com/microsoft/typescript-go/internal/ls/signaturehelp.go:780 +0x191
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).GetSignatureHelpItems(0x145640854700, {0x46a4f18, 0x1456409be5a0}, 0x49, 0x145632121688, 0x145640402008, 0x14564072e560)
	github.com/microsoft/typescript-go/internal/ls/signaturehelp.go:118 +0x1b3
github.com/microsoft/typescript-go/internal/ls.(*LanguageService).ProvideSignatureHelp(0x145640854700, {0x46a4f18, 0x1456409be5a0}, {0x14563c9f7740?, 0x14563c0dc6b0?}, {0x38e89eb?, 0x0?}, 0x14564072e560)
	github.com/microsoft/typescript-go/internal/ls/signaturehelp.go:47 +0x9b
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleSignatureHelp(0x14563d631808?, {0x46a4f18?, 0x1456409be5a0?}, 0x3133caf?, 0x46a4f18?)
	github.com/microsoft/typescript-go/internal/lsp/server.go:1440 +0x39
github.com/microsoft/typescript-go/internal/lsp.init.func1.registerLanguageServiceDocumentRequestHandler[...].18.1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:811 +0xa5
github.com/microsoft/typescript-go/internal/lsp.(*Server).handleRequestOrNotification.func1()
	github.com/microsoft/typescript-go/internal/lsp/server.go:664 +0x65
github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop.func3()
	github.com/microsoft/typescript-go/internal/lsp/server.go:532 +0x2c
created by github.com/microsoft/typescript-go/internal/lsp.(*Server).dispatchLoop in goroutine 35
	github.com/microsoft/typescript-go/internal/lsp/server.go:531 +0x531

2026-04-28 12:46:01.000 [error] Request textDocument/signatureHelp failed.
  Message: InternalError: panic handling request textDocument/signatureHelp: Debug failure. False expression: Not a subspan. Child: KindLessThanLessThanToken, parent: KindJsxSelfClosingElement
  Code: -32603 
```

## Steps to reproduce

1.  tsx file
```
import {Text} from 'react-native';

const text = () => {
  return <Text><</Text>;
};
```

## version
0.20260427.1

Andarist:
fixes https://github.com/microsoft/typescript-go/issues/3635

### DIFF
--- a/internal/astnav/tokens.go
+++ b/internal/astnav/tokens.go
@@ -8,6 +8,18 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
+func shouldRescanLessThanLessThanToken(s *scanner.Scanner, containingNode *ast.Node, token ast.Kind) bool {
+	return token == ast.KindLessThanLessThanToken && ast.IsJsxChild(containingNode)
+}
+
+func scanNavigationToken(s *scanner.Scanner, containingNode *ast.Node) ast.Kind {
+	token := s.Token()
+	if shouldRescanLessThanLessThanToken(s, containingNode, token) {
+		return s.ReScanJsxToken(true /*allowMultilineJsxText*/)
+	}
+	return token
+}
+
 func GetTouchingPropertyName(sourceFile *ast.SourceFile, position int) *ast.Node {
 	return getTokenAtPosition(sourceFile, position, false /*allowPositionInLeadingTrivia*/, func(node *ast.Node) bool {
 		return ast.IsPropertyNameLiteral(node) || ast.IsKeywordKind(node.Kind) || ast.IsPrivateIdentifier(node)
@@ -215,7 +227,7 @@ func getTokenAtPosition(
 				end = nodeAfterLeft.Pos()
 			}
 			for left < end {
-				token := scanner.Token()
+				token := scanNavigationToken(scanner, current)
 				tokenFullStart := scanner.TokenFullStart()
 				tokenStart := core.IfElse(allowPositionInLeadingTrivia, tokenFullStart, scanner.TokenStart())
 				tokenEnd := scanner.TokenEnd()
@@ -539,11 +551,11 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
 			for _, visitedNode := range rightmostVisitedNodes {
 				// Trailing tokens that occur before this node.
 				for startPos < min(visitedNode.Pos(), position) {
+					token := scanNavigationToken(scanner, n)
 					tokenStart := scanner.TokenStart()
 					if tokenStart >= position {
 						break
 					}
-					token := scanner.Token()
 					tokenFullStart := scanner.TokenFullStart()
 					tokenEnd := scanner.TokenEnd()
 					startPos = tokenEnd
@@ -557,11 +569,11 @@ func findRightmostValidToken(endPos int, sourceFile *ast.SourceFile, containingN
 			}
 			// Trailing tokens after last visited node.
 			for startPos < min(endPos, position) {
+				token := scanNavigationToken(scanner, n)
 				tokenStart := scanner.TokenStart()
 				if tokenStart >= position {
 					break
 				}
-				token := scanner.Token()
 				tokenFullStart := scanner.TokenFullStart()
 				tokenEnd := scanner.TokenEnd()
 				startPos = tokenEnd
--- a/internal/fourslash/tests/signatureHelpJsxTextLessThanTrigger_test.go
+++ b/internal/fourslash/tests/signatureHelpJsxTextLessThanTrigger_test.go
@@ -1,0 +1,34 @@
+package fourslash_test
+
+import (
+	"testing"
+
+	"github.com/microsoft/typescript-go/internal/fourslash"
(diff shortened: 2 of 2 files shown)


internal/astnav/tokens.go
```diff
@@ -8,6 +8,18 @@ import (
 	"github.com/microsoft/typescript-go/internal/scanner"
 )
 
+func shouldRescanLessThanLessThanToken(s *scanner.Scanner, containingNode *ast.Node, token ast.Kind) bool {
+	return token == ast.KindLessThanLessThanToken && ast.IsJsxChild(containingNode)
+}
+
+func scanNavigationToken(s *scanner.Scanner, containingNode *ast.Node) ast.Kind {
```
Andarist, line 15: Formatter has a bunch of other rescanning rules, so I wonder if this won't have to be expanded at some point... but for now, I'm just targeting the specific crash

DanielRosenwasser: My understanding is that the tree the parser constructs is actually something like

```
<Tag>  (opening tag)
  <    (part of a missing JSX self-opening element)
</Tag> (closing tag)
```

So this special case might be okay for now - but I'll note the formatting scanner also has a `readTokenInfo` which handles similar issues.

> [!NOTE]
> The comment is not exactly true - the scanner is greedy on some tokens, and shy in other cases. In the case of `<<`, it is greedy.

```go
	// normally scanner returns the smallest available token
	// check the kind of context node to determine if scanner should have more greedy behavior and consume more text.

	var expectedScanAction scanAction
	if shouldRescanGreaterThanToken(n) {
		expectedScanAction = actionRescanGreaterThanToken
	} else if shouldRescanSlashToken(n) {
		expectedScanAction = actionRescanSlashToken
	} else if shouldRescanTemplateToken(n) {
		expectedScanAction = actionRescanTemplateToken
	} else if shouldRescanJsxIdentifier(n) {
		expectedScanAction = actionRescanJsxIdentifier
	} else if s.shouldRescanJsxText(n) {
		expectedScanAction = actionRescanJsxText
	} else if shouldRescanJsxAttributeValue(n) {
		expectedScanAction = actionRescanJsxAttributeValue
	} else {
		expectedScanAction = actionScan
	}
```

I don't think the formatting scanner currently handles `<<` specially, so there's probably also a gap here, but it probably doesn't show up in any meaningful way (like a debug assertion)

andrewbranch:
Is this fix a missing port from Strada? It's always helpful to know what the root cause is if Strada doesn't crash/error, because there could be more stuff that's wrong or missing deeper in the stack.

DanielRosenwasser:
> It's always helpful to know what the root cause is if Strada doesn't crash/error, because there could be more stuff that's wrong or missing deeper in the stack.

I'm not 100% certain, but I think what happened is that Strada discarded the `<<` token in `createChildren`/`addSyntheticNodes` because its end exceeded the end position of the containing node.

https://github.com/microsoft/TypeScript/blob/c3bd12d888b86f676718b16e64d7d2abcb423514/src/services/services.ts#L537-L545

Andarist:
Oh, I forgot that children syntax list was also reparsed on demand in Strada. But yes, given that - Daniel is definitely right, that `textPos <= end` check was 100% responsible for avoiding the issue in Strada. I considered a similar bounds check like this before deciding on the current approach. 

I rechecked this situation now with https://ts-ast-viewer.com/ and Strada was totally missing that first token in `node.getChildren()`, so it was slightly off in that case because of that defensive bounds check.

andrewbranch (approved):

DanielRosenwasser:
Thanks!

PR #3606, opened 2026-04-25
fix(3605): Missing generic param displayed in hover

a-tarasyuk:
Fixes #3605.

Fixes crash found in https://github.com/microsoft/typescript-go/issues/3512#issuecomment-4299732162.

PR #3646, opened 2026-04-28
Handle type parameter declaration as type node in printer

gabritto:
Fixes crash found in https://github.com/microsoft/typescript-go/issues/3512#issuecomment-4299732162.
Supersedes #3524.

Needed because in hover, when emitting type arguments for expressions with type arguments, we use type parameter declarations instead of type references to type parameters so that we can print type arguments as in `Foo<T extends Blah>.bar` instead of `Foo<T>.bar`.


jakebailey (approved):

gabritto:
I just realized that the right fix is what's in https://github.com/microsoft/typescript-go/pull/3606, because I think the replacement of type parameter references with type parameter declarations should only show up in type arguments of expressions with type arguments.

PR #3670, opened 2026-04-30
Watcher build mode efficiency updates

### LINKED ISSUE #2666: `tsgo --build` fails to invalidate incremental cache after dependency update
When updating dependencies in package.json that introduce breaking type changes, `tsgo --build` (incremental mode) fails to detect these changes if a `tsconfig.tsbuildinfo` file already exists. The build reports success despite the presence of type errors.

If the `tsconfig.tsbuildinfo` file is manually deleted, the compiler correctly identifies and reports the type errors.

### Expected Behavior
The TypeScript compiler should detect changes in node_modules (or the program signature of dependencies) and invalidate the relevant parts of the incremental cache, forcing a re-check of affected files.

### Actual Behavior
The compiler assumes the project is up-to-date because the local source files' timestamps/hashes haven't changed, ignoring the fact that the external type definitions they depend on have changed.

johnfav03:
Refactors the `tsc --build --watch` functionality to use the new shared `FileWatcher` and `trackingvfs` structure from the CLI and LSP watchers, reducing code duplication and patching inefficiencies like redundant scanning of shared dependencies.

Note: this PR is built on the `watcher-lsp-integration` branch so only review commits from `0655dba` onward

Fixes #2666


rubenferreira97:
> Disclaimer: I used AI to help review this because I had a suspicion from reading the PR code, but I’m not fully familiar with this codebase, so these tests/findings may need validation from maintainers.

I pushed a test-only branch with two possible regression tests:

https://github.com/johnfav03/typescript-go/compare/watch-build-efficiency...rubenferreira97:pr-3670-dependency-invalidation-tests

What I’m seeing locally:

1. `tsgo --build` appears to still miss a dependency type change when the dependency `.d.ts` content changes but its mtime is not newer than `dist/tsconfig.tsbuildinfo`.

   My understanding is that this could matter for `file:` dependencies, workspace/symlinked dependencies, pnpm-style stores, or installs where dependency files may already exist or preserve older mtimes.

2. `tsgo --build --watch` appears to miss a dependency change on a case-insensitive filesystem when the dependency path casing differs between the path recorded during resolution and the path reported by the watcher.

Both tests currently fail against this PR branch. The first one seems closely related to #2666, while the second may just be a build-watch path normalization edge case. I’m sharing them as potential gaps rather than definitive blockers, since I may be missing some intended invariant in the build/watch test harness or implementation.

jakebailey:
This PR's just out of date and needs a rethink after #3980 and co, so I wouldn't test it quite yet

PR #3675, opened 2026-05-01
Port experimentation service and reporting to extension

DanielRosenwasser:


_extension/package.json
```diff
@@ -223,7 +223,8 @@
     },
     "dependencies": {
         "@vscode/extension-telemetry": "^1.5.1",
-        "vscode-languageclient": "^10.0.0-next.21"
+        "vscode-languageclient": "^10.0.0-next.21",
+        "vscode-tas-client": "0.1.84"
```
DanielRosenwasser, line 227: Locked on `0.1.84` because of https://github.com/microsoft/tas-client/issues/95

_extension/src/extension.ts
```diff
@@ -29,6 +32,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
     const telemetryReporter = createTelemetryReporter(new VSCodeTelemetryReporter(aiConnectionString));
     context.subscriptions.push(telemetryReporter);
 
+    const packageInfo = getPackageInfo(context);
+    if (packageInfo) {
+        const { name: id, version } = packageInfo;
+        // Constructing the experimentation service actually sets shared properties
+        // so that events include context on treatments/flights.
+        const _expService = new ExperimentationService(telemetryReporter, id, version, context.globalState);
+    }
```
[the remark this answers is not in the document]
DanielRosenwasser: For now this is fine, we are just using this to stamp common properties like the assignment context. I could just `void` it.

_extension/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "target": "es2023",
-        "module": "NodeNext",
+        "module": "preserve",
```
jakebailey: we bundle, so maybe this should be bundler mode?

_extension/src/commands.ts
```diff
@@ -6,7 +6,6 @@ import type {
     Position,
 } from "vscode-languageclient";
 
-import { Client } from "./client";
```
jakebailey, line 9: Are we missing unused var settings?

_extension/tsconfig.json
DanielRosenwasser: I actually don't think I needed to make this change. I might undo it.

Though I think that `preserve` implies `bundler`, right?

jakebailey: we only bundle now

_extension/src/commands.ts
DanielRosenwasser: Probably should do a PR to add a linter.

_extension/src/experimentationService.ts
```diff
@@ -0,0 +1,62 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
```
jakebailey: We're not MIT, probably delete this

_extension/src/extension.ts
```diff
@@ -29,6 +32,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<Extens
     const telemetryReporter = createTelemetryReporter(new VSCodeTelemetryReporter(aiConnectionString));
     context.subscriptions.push(telemetryReporter);
 
+    const packageInfo = getPackageInfo(context);
+    if (packageInfo) {
+        const { name: id, version } = packageInfo;
```
jakebailey: Not just `context.extension.id`?

DanielRosenwasser: No, `name` is `native-preview`, `context.extension.id` would be `TypeScriptTeam.native-preview`

jakebailey: well right, but surely the exp platform wants more of a name than `native-preview`?

DanielRosenwasser: You are right and I think that the built-in extension was doing this wrong.

jakebailey: Oh, yeah, because builtin extensions don't have a publisher name and that's how they are determined to be builtins

jakebailey (approved):

Issue #3611, opened 2026-04-25
Elevated CPU usage in `--watch` mode when idle (compared to tsc)


alexswan93:
Watch mode is consistently using ~5-10% CPU while idle, compared to `tsc` which uses 0% while idle.

### Using:
- `"@typescript/native-preview": "7.0.0-dev.20260324.1"`
- `Win 11 x64`

### Steps to reproduce
Run `tsgo --watch` or `tsgo --watch --noEmit` on any project -> observe idle CPU usage.

### Behavior with `typescript@6.0`
- 0% CPU while idle

### Behavior with `tsgo`
- Tiny project (5 small files) using ~5% CPU consistently while idle
- Medium project (hundreds) using ~10% CPU consistently while idle

The idle CPU utilisation was improved massively by the recent watch mode updates 😄 but especially for those like myself on resource-limited machines even this remaining amount constitutes a barrier to using watch mode.

I am currently working around this by using my own file watcher to invoke `tsgo --incremental`

DanielRosenwasser:
Our current strategy is naive polling at a fixed frequency, but it doesn't feel like it should be *that* taxing. @jakebailey is looking into getting some perf infrastructure over at https://github.com/microsoft/typescript-go/pull/3627 in so we can better measure what's happening.

PR #3677, opened 2026-05-01
Remove redundant work from polling watcher

andrewbranch:
Two issues:

1. Directories present in the `paths` seen by the compiler host were being read and children hashed, but for those, we only need to stat them for existence. We only need to compute a hash of a directory's children when it's part of a tsconfig include pattern. Directories in the `paths` portion of the watch state are triggered by module resolution lookups, and we only need to track their existence so we can tell if, e.g., a node_modules package directory that would have had individual file lookups inside it gets created. (We no longer store those would-be file lookups as paths to check after #3038; checking the directory is sufficient.)
2. `hasChanges` was iterating both the previous watch state baseline as well as the current set of wildcard directories, but the state baseline already includes the results of scanning the wildcard directories, so just iterating the baseline is sufficient.

On top of that, I doubled the polling interval from 1 second to 2 seconds, in line with `esbuild --watch`'s worst-case latency.

Together, these changes bring `tsgo --watch` on `vscode/extensions/copilot` down from 20% idle CPU to 4%.

We could squeeze a little more out of polling by checking recently-changed files more often and rarely changed files less often, but we're also exploring some non-polling options that seem more promising. It's hard to accept a high-latency solution at 4% CPU that scales linearly in latency or CPU utilization with project size, when you can plausibly have a low-latency solution at 0% idle CPU that scales quite well.

Related: #3611

### DIFF
--- a/internal/execute/watcher.go
+++ b/internal/execute/watcher.go
@@ -115,6 +115,10 @@ func (w *Watcher) start() {
 		w.configFilePaths = append([]string{w.configFileName}, w.config.ExtendedSourceFiles()...)
 	}
 
+	if w.sys.GetEnvironmentVariable("TSGO_WATCH_DEBUG") != "" {
+		w.fileWatcher.SetDebugLog(w.sys.Writer())
+	}
+
 	w.reportWatchStatus(ast.NewCompilerDiagnostic(diagnostics.Starting_compilation_in_watch_mode))
 	w.doBuild()
 	w.mu.Unlock()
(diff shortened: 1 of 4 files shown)


internal/execute/watcher.go
```diff
@@ -115,6 +115,10 @@ func (w *Watcher) start() {
 		w.configFilePaths = append([]string{w.configFileName}, w.config.ExtendedSourceFiles()...)
 	}
 
+	if w.sys.GetEnvironmentVariable("TSGO_WATCH_DEBUG") != "" {
```
jakebailey: Nit but, we have started using `TS_` as the prefix, where `TS_GO_RUNTIME_TRACE` means "the go runtime tracer", so maybe this should just be `TS_WATCH_DEBUG`

johnfav03 (approved):

PR #3698, opened 2026-05-04
Add minimal tsc output format

JoviDeCroock:
## Summary

Adds an explicit `--outputFormat minimal` command-line option for `tsgo` diagnostics. The minimal reporter emits one diagnostic per line in the form:

```text
path/to/file.ts:line:character: TS1234: message | chained message
```

This keeps the default output unchanged while giving tooling and agents a compact, stable format that is easier to parse and significantly cheaper to feed back into LLM context windows.

## Motivation

AI coding agents frequently run `tsc`/`tsgo` and then include the compiler output in the next prompt. The existing pretty/default diagnostics are excellent for humans, but they include context, spacing, colors, and multi-line message chains that add a lot of tokens. A compact output format reduces token usage and makes repeated fix/test loops less noisy without taking away the current default human-facing experience.

As a future follow-up, we could also auto-select this format when `tsgo` is run by an agent. A similar principle is being explored in oxc by detecting agent-specific environment variables and switching to an agent-oriented output format: https://github.com/oxc-project/oxc/pull/22068. That PR mirrors the environment detection used by `unjs/std-env` agents detection, which is also used by tools like Vitest.

### DIFF
--- a/internal/core/compileroptions.go
+++ b/internal/core/compileroptions.go
@@ -145,6 +145,7 @@ type CompilerOptions struct {
 	NoEmitForJsFiles    Tristate `json:"noEmitForJsFiles,omitzero"`
 	PreserveWatchOutput Tristate `json:"preserveWatchOutput,omitzero"`
 	Pretty              Tristate `json:"pretty,omitzero"`
+	OutputFormat        string   `json:"outputFormat,omitzero"`
 	Version             Tristate `json:"version,omitzero"`
 	Watch               Tristate `json:"watch,omitzero"`
 	ShowConfig          Tristate `json:"showConfig,omitzero"`
@@ -168,6 +169,11 @@ type noCopy struct{}
 // Lock is a no-op used by -copylocks checker from `go vet`.
 func (*noCopy) Lock()   {}
 func (*noCopy) Unlock() {}
+
+const (
+	OutputFormatDefault = "default"
+	OutputFormatMinimal = "minimal"
+)
 
 var EmptyCompilerOptions = &CompilerOptions{}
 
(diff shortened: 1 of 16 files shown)


internal/diagnosticwriter/diagnosticwriter_test.go
```diff
@@ -0,0 +1,116 @@
...
+	var output strings.Builder
+	WriteMinimalDiagnostic(&output, diagnostic, &FormattingOptions{
+		NewLine: "\n",
+		ComparePathsOptions: tspath.ComparePathsOptions{
+			CurrentDirectory:          "/repo",
+			UseCaseSensitiveFileNames: true,
+		},
+	})
+
+	want := "src/index.ts:1:17: TS2322: Type 'number' is not assignable to type 'string'. | The expected type comes from property 'x'\n"
```
jakebailey: You should be writing end-to-end tests in `tsctests`, not unit tests here (or elsehwere, really)

JoviDeCroock: Thank you for the feedback, very new to GO so appreciate you chiming in quickly here. The tests were generated ahead of time by an LLM so I had something sane to work against.

I explicitly left it in Draft for that reason so we could talk through the desire for this feature. I think it could be valuable in terms of working towards a future where we save tokens of diagnostics coming from testing/linting/...

JoviDeCroock:

JoviDeCroock:
> label gets applied when someone seems to have a suspiciously high count of PRs and has possibly run afoul of our "no bulk contributors" policy.

This is my only contribution to this repository, all my other contributions are other repositories and OSS I maintain... 😅 I'm sorry for whatever I did, last time I made an issue on TypeScript it also got slammed with "trying to use a kitchen knife as a wheel axle"

Closed it out, I hope an idea like this could be taken under consideration at some point to reduce the wasteful nature of human-readable outputs when used as input-tokens. LLM's are wasteful enough as it stands

PR #3703, opened 2026-05-04
Fixes to --generateTrace to make analyze-trace work

### LINKED ISSUE #3054: `tsgo --generateTrace` appears unimplemented
## Steps to reproduce

`--generateTrace` doesn't seem to have been implemented in `tsgo`. The flag exists in the `tsgo --help --all` documentation, and demands an argument, but doesn't appear to actually do anything.

I've got some very large typescript projects that seem to consume more memory with tsgo than expected, and I'd like to use the trace output to help improve things. Ideally, I'd be able to use the dramatically faster compiler to do so, rather than having to use `tsc` for this.

`tsc --generateTrace trace` will generate trace output for use with `@typescript/analyze-trace`.

## Behavior with `tsgo`

`tsgo --generateTrace trace` doesn't generate any trace output.

jakebailey:
Pairs with https://github.com/microsoft/typescript-analyze-trace/pull/48

In this PR:

1. Non-LIFO B/E events. Chrome's B/E (begin/end) events are stack-scoped per thread, but concurrent goroutines on the same TID produced overlapping spans, which is malformed. Add synthetic per-logical-owner TIDs. Each span gets a TID derived from its "owner" (checker id for type-checking work, file path for parse/bind), so a given TID only ever 
sees a well-nested sequence.
2. Multi-checker type IDs. Each checker has its own type-id space, so types.json can't be a single file. tsgo emits types_0.json, types_1.json, ... and a legend.json mapping 
checkerId to typesPath. Checker events now carry args.checkerId so the consumer knows which file to resolve a type id in.

Fixes #3054

`@typescript/analyze-trace` 0.11.0+ support this


internal/tracing/tracing.go
```diff
@@ -295,12 +315,117 @@ func (tr *Tracing) Push(phase Phase, name string, args map[string]any, separateB
...
+			}
+		}
+	}
+
+	return traceThreadKey{}, false
+}
+
+func traceCheckerThreadKey(checkerID any) (traceThreadKey, bool) {
+	switch checkerID := checkerID.(type) {
+	case int:
```
gabritto: Why do we need to handle all of these cases for `checkerID`? Don't we control what type we put in the map?

gabritto (approved):

internal/tracing/tracing.go
jakebailey: nope, this is leftover from a refactor, thanks for noticing

gabritto (approved):

PR #3711, opened 2026-05-05
Narrow String#toLowerCase/toUpperCase return types via Lowercase/Uppercase

ljharb:
Make String.prototype.toLowerCase and toUpperCase preserve string-literal types in their return type, by typing them as `<T extends string>(this: T): Lowercase<T>` and `Uppercase<T>` respectively.

For non-literal `string` receivers, `Lowercase<string>` resolves to `string`, preserving existing behavior. For literal receivers (e.g. `"FOO".toLowerCase()`), the result narrows to the corresponding literal (`"foo"`).


RyanCavanaugh:
TypeScript is the correct originating repo for lib changes; no need to PR here.

PR #3717, opened 2026-05-05
Proposed actions for symbol baseline diffs

RyanCavanaugh:
I put the "needs investigation" groups at the top. Same procedure as #3540

### DIFF
--- a/groups.md
+++ b/groups.md
@@ -1,0 +1,787 @@
+# Symbols Diff Categorization
+
+*473 `.symbols.diff` files under `testdata/baselines/reference/submodule/` categorized by type of binding change between Corsa (Go) and Strada (JS).*
+
+Note: Many files exhibit multiple categories of change simultaneously. Such files appear under each relevant group.
+
+---
+
+
+# Inherited method resolves to base class instead of overriding derived class
+## `C.foo` → `B.foo` for methods — symbol resolves to base class
+## **Needs Investigation** — if the method is overridden in C, it should resolve to C's declaration, not B's.
+conformance/typeFromPropertyAssignment23.symbols.diff
+
+# Inherited property resolves to derived class instead of declaring base class
+## `Element.textContent` → `TextElement.textContent` — property symbol attributed to derived class
+## **Needs Investigation** — property should resolve to where it's originally declared (the base), not the derived class that inherits it.
+conformance/thisPropertyAssignmentInherited.symbols.diff
+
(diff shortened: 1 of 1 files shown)


groups.md
```diff
@@ -0,0 +1,787 @@
...
+conformance/thisPropertyAssignmentInherited.symbols.diff
+
+# Duplicate/overload declarations no longer merge into a single symbol with multiple Decls
+## Separate overload or duplicate method signatures each get their own symbol
+## **Needs Investigation** — overloads in TS should share a single symbol; splitting them could break declaration emit and signature resolution for overloaded functions.
+conformance/callSignaturesWithParameterInitializers2.symbols.diff
+conformance/stringLiteralTypesInImplementationSignatures2.symbols.diff
+conformance/symbolDeclarationEmit12.symbols.diff
+conformance/symbolProperty44.symbols.diff
+compiler/methodSignatureHandledDeclarationKindForSymbol.symbols.diff
```
jakebailey, line 27: These are all error cases with duplicate decls, so showing only one is really not a big deal as there's an error message and dts emit emitting (correctly) just one is not a problem

groups.md
```diff
@@ -0,0 +1,787 @@
...
+
+# `import("...").X` property access on import types no longer resolves to member symbol
+## Property access expressions on `import()` type references lose their symbol resolution
+## **Needs Investigation** — losing symbol resolution for `import("...").X` means go-to-definition and hover will fail on JSDoc `import()` type references, a common JS pattern.
+conformance/jsdocImportTypeReferenceToCommonjsModule.symbols.diff
+conformance/jsdocImportTypeReferenceToESModule.symbols.diff
```
jakebailey, line 33: These are new error cases:

```
==== ex.d.ts (0 errors) ====
    declare var config: {
        fix: boolean
    }
    export = config;
    
==== test.js (1 errors) ====
    /** @param {import('./ex')} a */
                ~~~~~~~~~~~~~~
!!! error TS1340: Module './ex' does not refer to a type, but is used as a type here. Did you mean 'typeof import('./ex')'?
    function demo(a) {
        a.fix
    }
    
```

So fixing the error diff (which is not accepted) would fix this. But I seem to recall that this error diff was intentional

andrewbranch: Yeah, I remember @sandersn bringing this up. I don't mind being more strict about type/value confusion like this, but it seems like we've walked back a lot of similar breaks since then.

PR #3720, opened 2026-05-06
Prompt after 5 consecutive server restarts in a session

DanielRosenwasser:
Something to help understand why users (outside of developers on the team) are doing this.

We may consider doing this as part of a phased rollout.


DanielRosenwasser:
This hasn't been an issue really.

PR #3731, opened 2026-05-06
Handle config file diagnostics updates via file watching

### LINKED ISSUE #3676: tsconfig diagnostics not refreshed if no TS/JS files are open
```
// @filename: tsconfig.json
{
    "compilerOptions": {
        "baseUrl": "./"
    }
}

// @filename: foo.ts
export let x = 0;
```

1. Open `foo.ts`.
2. You should get an error in `tsconfig.json`:

   ```
   Option 'baseUrl' has been removed. Please remove it from your configuration.
     Use '"paths": {"*": ["./*"]}' instead.
   ```
3. Close `foo.ts`.
4. Delete the `baseUrl` option from `tsconfig.json`.

**Expected**: Error disappears.
**Actual**: Error persists, span is off.

Image

gabritto:
Fixes #3676.
Since our LS is not (yet) registered to handle .json files, we currently need to handle tsconfig changes via push diagnostics, and get updates via file watcher events.

To make tsconfig diagnostics work, we cannot assume that if a project is closed, we won't need to produce updated diagnostics for it, so in this PR  we handle tsconfig changes even when its project is currently unopened (i.e. missing from project collection and config file registry). It does so by ensuring during snapshot clone, we collect config file changes and creations and later process configured projects for those. We exclude config files that don't belong to the current workspace or that belong to `node_modules`, and we also ensure we always watch the current workspace for file changes.


internal/project/snapshotfs.go
```diff
@@ -432,6 +437,9 @@ func (s *snapshotFSBuilder) watchChangesOverlapCache(change FileChangeSummary) b
 		if _, ok := s.nodeModulesRealpathAliases.Load(path); ok {
 			return true
 		}
+		if isWatchedConfigFile(uri, s.toPath, currentDirectoryPath, oldCustomConfigFileName, newCustomConfigFileName) {
```
gabritto: I am not sure if we need this. If we have excessive file watch events, we may already have dropped some config changes, and we cannot do much about it unless we keep track of every config we pushed diagnostics for and re-do all of those from scratch.

internal/project/session.go
```diff
@@ -362,7 +390,8 @@ func (s *Session) DidChangeWatchedFiles(ctx context.Context, changes []*lsproto.
 	s.pendingFileChanges = append(s.pendingFileChanges, fileChanges...)
 	s.pendingFileChangesMu.Unlock()
 
-	// Schedule a debounced diagnostics refresh
+	// Schedule a debounced snapshot update and diagnostics refresh
+	s.scheduleSnapshotUpdate()
```
gabritto, line 393: I figured we shouldn't necessarily be waiting to get a request to update those diagnostics, so I added a scheduled update similar to how we schedule pull diagnostics refresh, etc.

internal/project/session.go
```diff
@@ -1282,6 +1373,15 @@ func (s *Session) updateWatches(oldSnapshot *Snapshot, newSnapshot *Snapshot) er
 		}
 	}
 
+	// Register the permanent config file watcher if not yet registered or pending retry.
```
gabritto, line 1375: This makes sure we always watch the workspace.

internal/project/session.go
```diff
@@ -1517,6 +1625,18 @@ func (s *Session) publishProgramDiagnostics(oldSnapshot *Snapshot, newSnapshot *
...
+	// file we previously published diagnostics for that has been deleted,
+	// including config files inside deleted directories.
+	for uri := range change.fileChanges.Deleted.Keys() {
+		deletedPath := s.toPath(uri.FileName())
+		for configPath := range s.publishedDiagConfigPaths.Keys() {
+			if configPath == deletedPath || deletedPath.ContainsPath(configPath) {
+				s.publishProjectDiagnostics(ctx, string(configPath), nil, nil /*converters*/)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
gabritto: Do we already have something like this in the repo? @andrewbranch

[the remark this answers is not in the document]
andrewbranch: We already expand directory deletions into file deletions for all known contents of those directories in `snapshotFSBuilder.Finalize`. Maybe that's useful?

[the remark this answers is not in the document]
gabritto: Yeah, I saw that, but we don't expect to always have that information on disk for config files or their directories... I'm tempted to just leave this like that and optimize later if needed. We can have quite a large number of config files with errors (e.g. monorepo with baseUrl usage), but maybe not that many deletions, so not sure how blocking this is.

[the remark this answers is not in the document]
andrewbranch: I haven't dug into the code yet so maybe this is already there, but I think as long as a mass deletion triggers purging all published diagnostics rather than evaluating, this is fine. You may want to specifically look at how deletions of `node_modules` are handled, though, as some of the excessive watch event handling splits out node_modules changes from non-node_modules changes, and we would expect to normally have zero tsconfigs inside node_modules. In a giant monorepo, we wouldn't want to process 1000 tsconfigs against 50k deleted node_modules files.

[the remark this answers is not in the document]
gabritto: Ok, we're handling excessive file changes now and only looking through file changes for deletion if we're within the threshold, so this should now be O(#configs) with a possibly 1k constant. I'm also excluding `node_modules` changes right away, so maybe this is good enough for now while we don't have anything trie-like in the repo.

internal/project/project_test.go
```diff
@@ -210,7 +227,80 @@ func TestPushDiagnostics(t *testing.T) {
...
+			if diag.Code != nil && diag.Code.Integer != nil {
+				switch *diag.Code.Integer {
+				case 2318: // Cannot_find_global_type_0
+					t.Errorf("expected no diagnostic code 2318 (Cannot find global type) after project cleanup, got: %v", lastCall.Params.Diagnostics)
+				case 5102: // Option_0_has_been_removed_Please_remove_it_from_your_configuration
+					hasBaseUrlDiag = true
+				}
+			}
+		}
+		assert.Assert(t, hasBaseUrlDiag, "expected 'baseUrl' option removed to remain after project cleanup, got: %v", lastCall.Params.Diagnostics)
```
andrewbranch, line 300: Well, this is kind of the opposite scenario of the issue, right? I didn't think of this until seeing the test, but I think the problem with this approach is, let's say you're working in a monorepo, and you happen to open a file in a project that you're not planning to change, and that tsconfig triggers an error. Now, you can never clear that error without actually fixing the tsconfig. It seems like with the limitations of not knowing when a tsconfig file is opened and closed, we might actually want to limit what we do here to clearing config errors on project close.

andrewbranch: I think another scenario that has a problem with this approach is something like a `git pull` that changes the formatting of every tsconfig file in the workspace. If I'm reading right, assuming we're already watching the workspace and get those events, this would trigger a configured project load of every project?

gabritto: Yes, both things are true, but we can't do better without handling open .json files. I do think it's also weird for you to have a tsconfig.json open and not get errors for it, so it's just a question of what's weirder.

andrewbranch: I think, given the constraints we have, the expected behavior for #3676 would be that the baseUrl error disappears at _step 3_ (closing the last project file, even though the tsconfig is still invalid).

I think it’s generally more important to _clear errors_ at the right time than to _show errors_, so I think just clearing errors on project close is an ok stopgap for right now. I do think it's worth looking into handling .json files so we can know when tsconfigs are open/closed. But for sure, opening every configured project on `git switch` and having no way to close them isn't tenable.

PR #3737, opened 2026-05-07
fix(3735): fix unmatched jsdoc param errors missing for arrow function class fields

### LINKED ISSUE #3735: Behavior difference: tsgo fails to emit error on class field's unmatched params between jsdoc & js code
## Steps to reproduce

1. Start a plain new project
    - "@typescript/native-preview": "7.0.0-dev.20260507.1",
    - "typescript": "6.0.3"
    - "allowJs": true,
    - "checkJs": true,
    - "stableTypeOrdering": true,

```js
// src/main.js

export class C {
    /**
     * @param {string} foo
     * @param {number} bar
     */
    method(foo) {}

    /**
     * @param {string} foo
     * @param {number} bar
     */
    field = (foo) => {}
}
```

```
src/main.js:6:24 - error TS8024: JSDoc '@param' tag has name 'bar', but there is no parameter with that name.

6      * @param {number} bar
                         ~~~

src/main.js:12:24 - error TS8024: JSDoc '@param' tag has name 'bar', but there is no parameter with that name.

12      * @param {number} bar
                          ~~~


Found 2 errors in the same file, starting at: src/main.js:6
```

## Behavior with `tsgo`

```
src/main.js:6:24 - error TS8024: JSDoc '@param' tag has name 'bar', but there is no parameter with that name.

6      * @param {number} bar
                         ~~~


Found 1 error in src/main.js:6
```

a-tarasyuk:
Fixes #3735


internal/ast/utilities.go
```diff
@@ -4014,6 +4014,31 @@ func GetHostSignatureFromJSDoc(node *Node) *Node {
...
+		return nil
+	}
+	switch parent.Kind {
+	case KindPropertyAssignment, KindExportAssignment, KindPropertyDeclaration:
+		return parent
+	case KindExpressionStatement:
+		if node.Kind == KindPropertyAccessExpression {
+			return parent
+		}
+	case KindReturnStatement:
```
ahejlsberg: Merge this case with the first case above since they return the same result.

internal/ast/utilities.go
```diff
@@ -4014,6 +4014,31 @@ func GetHostSignatureFromJSDoc(node *Node) *Node {
 	return nil
 }
 
+func GetNextJSDocCommentLocation(node *Node) *Node {
```
ahejlsberg, line 4019: Maybe add a comment that this function should be kept in sync with `findMatchingParameter` in the reparser.

ahejlsberg (approved):

PR #3743, opened 2026-05-07
Add missing parens in type serializer

### LINKED ISSUE #3548: Baselines: Decorator metadata reflects stable union type ordering
Decorator metadata is runtime-accessible; changing `Object` to `String` affects `Reflect.getMetadata`. `design:type` for `T | null` changed from `Object` to the non-null type (e.g. `String`, `Class1`).

## Affected Files
``
compiler/metadataOfUnionWithNull(strictnullchecks=true).js.diff
compiler/metadataReferencedWithinFilteredUnion(strictnullchecks=true,target=es2015).js.diff
```

weswigham:
Fixes #3548

Sometimes it is just that simple.


jakebailey (approved):
oops

jakebailey (approved):

PR #3745, opened 2026-05-07
Add missing type argument count sanity check to handle nodes from js …

### LINKED ISSUE #3549: Baselines: Missing generic type arguments no longer auto-filled with any (produces TS2314 errors)
`Array` without type arguments now emits bare `Array` instead of `Array<any>`, producing TS2314 errors.

## Affected Files
``
conformance/jsDeclarationsMissingGenerics(target=es2015).js.diff
```

weswigham:
…with extra-flexible conventions

Fixes #3549. We didn't need this check until we added back the extra aliases in jsdoc for these builtins. We might need this checked in some more places, but this could end up being sufficient, given how the builtin aliases are handled - hard to know for sure. This check is certainly less broadly applicable than in strada, where any given `Foo` could omit type params in js, and not just `Array` and `Promise`.


DanielRosenwasser (approved):
The suggested helper method isn't a bad idea but otherwise looks good.

jakebailey (approved):

Issue #2857, opened 2026-02-20
Formatters get stuck


mjbvz:
### Extension Version

0.20260220.1

### VS Code Version

1.110

### Operating system Version

MacOS

### Steps to reproduce

Not sure of exact repo steps. I see this most often when using format on save

### Issue

The formatter gets stuck and won't run. This is most obvious when using format on save as we show UI about the long running format operation 

In the LSP log, the last message I see being sent is:

```
2026-02-20 11:14:21.058 [trace] Sending request 'textDocument/formatting - (8193)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/matb/projects/vscode/src/vs/workbench/contrib/chat/common/model/chatModelStore.ts"
    },
    "options": {
        "tabSize": 4,
        "insertSpaces": false,
        "trimTrailingWhitespace": true,
        "insertFinalNewline": true
    }
}
```

There does not appear to be a response

SibtainOcn:
I investigated this issue and I believe the root cause is `snapshotUpdateMu` lock contention blocking the dispatch loop.

**The problem:**

`registerLanguageServiceDocumentRequestHandler` in `server.go` calls `GetLanguageService` **synchronously on the dispatch loop thread**. This function acquires `snapshotUpdateMu.Lock()` via `getSnapshot()` in `session.go`.

The `dispatchLoop` processes requests one at a time. If `snapshotUpdateMu` is held by another operation, the dispatch loop stalls and no responses can be sent.

**Lock contention sources for `snapshotUpdateMu`:**

- `DidOpenFile` (session.go:258) holds it during `UpdateSnapshot`
- `scheduleIdleCacheClean` timer (session.go:417) holds it from a background goroutine
- API session operations (api.go:10)

**Why format-on-save is most affected:**

When the user saves, VS Code sends `didSave` and `textDocument/formatting` in quick succession. The save can trigger snapshot updates (via idle cache clean or diagnostics refresh) that hold `snapshotUpdateMu`, and the formatting request immediately contends on the same lock while running synchronously on the dispatch loop.

**Proposed fix:**

Move `GetLanguageService` from the synchronous handler into the async closure, so it runs in a goroutine instead of blocking the dispatch loop. I have a PR ready with this change: #2857

See commit: https://github.com/SibtainOcn/typescript-go/commit/67861199

jakebailey:
That suggestion would just re-break the bugs fixed by #2765...

DanielRosenwasser:
[...]

PR #3749, opened 2026-05-07
Don't hold snapshotUpdateMu while running GC()

andrewbranch:
Noticed this while trying to figure out a possible cause for #2857, but I think this is unlikely to make anything look “stuck”; more just a little slow.


jakebailey (approved):
Nervous about panicks causing a deadlock but I guess if that happens it's pretty bad already

andrewbranch:
We don't recover from panics that happen during snapshot updates anyway 🤷

jakebailey (commented):
Actually, can we just do `go func() { runtime.GC() }()`?

jakebailey (approved):
PR title is wrong but that's ok

PR #3783, opened 2026-05-08
check that files specified in tsconfig.json "include" is located in "rootdir"

### LINKED ISSUE #3662: Behavior difference: tsgo wrongly allows `include` to contain files that is not under `rootDir` from tsconfig
## Steps to reproduce

1. Start a plain new project
    - "@typescript/native-preview": "7.0.0-dev.20260429.1",
    - "typescript": "6.0.3"
    - "stableTypeOrdering": true,
    - "rootDir": "./src2",
    - "include": ["src1/*"]
3. Add an empty `src1/main.ts`. Run `npx tsc --noEmit` & `npx tsgo --noEmit` to see the difference.

```
error TS6059: File '<project>/src1/main.ts' is not under 'rootDir' '<project>/src2'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Matched by include pattern 'src1/*' in '<project>/tsconfig.json'
  File is ECMAScript module because '<project>/package.json' has field "type" with value "module"

  tsconfig.json:40:17
    40     "include": ["src1/*"]
                       ~~~~~~~~
    File is matched by include pattern specified here.


Found 1 error.
```

## Behavior with `tsgo`

No error at all.

iisaduan:
fixes #3662

### DIFF
--- a/internal/outputpaths/commonsourcedirectory.go
+++ b/internal/outputpaths/commonsourcedirectory.go
@@ -56,16 +56,22 @@ func GetComputedCommonSourceDirectory(emittedFiles []string, currentDirectory st
 	return commonSourceDirectory
 }
 
-func GetCommonSourceDirectory(options *core.CompilerOptions, files func() []string, currentDirectory string, useCaseSensitiveFileNames bool) string {
+func GetCommonSourceDirectory(options *core.CompilerOptions, files []string, currentDirectory string, useCaseSensitiveFileNames bool, checkSourceFilesBelongToPath func([]string, string) bool) string {
 	var commonSourceDirectory string
 	if options.RootDir != "" {
 		// If a rootDir is specified use it as the commonSourceDirectory
 		commonSourceDirectory = options.RootDir
+		if checkSourceFilesBelongToPath != nil {
+			checkSourceFilesBelongToPath(files, options.RootDir)
+		}
 	} else if options.ConfigFilePath != "" {
 		// If the rootDir is not specified, then the common source directory is the directory of the config file.
 		commonSourceDirectory = tspath.GetDirectoryPath(options.ConfigFilePath)
+		if checkSourceFilesBelongToPath != nil {
+			checkSourceFilesBelongToPath(files, commonSourceDirectory)
+		}
 	} else {
-		commonSourceDirectory = computeCommonSourceDirectoryOfFilenames(files(), currentDirectory, useCaseSensitiveFileNames)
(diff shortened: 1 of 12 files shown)


internal/outputpaths/commonsourcedirectory.go
```diff
@@ -56,16 +56,22 @@ func GetComputedCommonSourceDirectory(emittedFiles []string, currentDirectory st
...
-func GetCommonSourceDirectory(options *core.CompilerOptions, files func() []string, currentDirectory string, useCaseSensitiveFileNames bool) string {
+func GetCommonSourceDirectory(options *core.CompilerOptions, files []string, currentDirectory string, useCaseSensitiveFileNames bool, checkSourceFilesBelongToPath func([]string, string) bool) string {
 	var commonSourceDirectory string
 	if options.RootDir != "" {
 		// If a rootDir is specified use it as the commonSourceDirectory
 		commonSourceDirectory = options.RootDir
+		if checkSourceFilesBelongToPath != nil {
+			checkSourceFilesBelongToPath(files, options.RootDir)
+		}
 	} else if options.ConfigFilePath != "" {
```
iisaduan, line 67: In strada, this is `else if (options.composite && options.configFilePath)`, but making that change here broke quite a few things. I'm not sure the reasoning for the change, so I left it

jakebailey: What did it break? I think I remember sheetal doing that on purpose

iisaduan: If she removed the composite part of it here, then that means leaving this alone is right

internal/outputpaths/commonsourcedirectory.go
```diff
@@ -56,16 +56,22 @@ func GetComputedCommonSourceDirectory(emittedFiles []string, currentDirectory st
 	return commonSourceDirectory
 }
 
-func GetCommonSourceDirectory(options *core.CompilerOptions, files func() []string, currentDirectory string, useCaseSensitiveFileNames bool) string {
+func GetCommonSourceDirectory(options *core.CompilerOptions, files []string, currentDirectory string, useCaseSensitiveFileNames bool, checkSourceFilesBelongToPath func([]string, string) bool) string {
```
andrewbranch: `files` was a function because in 2/3 of the control flow branches in this function, it doesn't need to be evaluated, so we want to defer running `sourceFileMayBeEmitted` on potentially thousands of files until we know we need to. What was the reason for making it eager?

iisaduan: brain went oh its _referenced_ in all the branches now. I'll refactor to change it back

andrewbranch (approved):

PR #3794, opened 2026-05-10
Remove assertion that doesn't hold after excessive stack depth overflow

### LINKED ISSUE #3791: tsgo panics on a recursive-mapped-type + intersection + generic-signature pattern
[tsgo-panic-repro.zip](https://github.com/user-attachments/files/27561006/tsgo-panic-repro.zip)

`tsgo` (TypeScript 7 native preview) panics in `Relater.reportRelationError` (`internal/checker/relater.go:4725`) on certain combinations of recursive mapped types + intersection types + generic function-expression assignment, before emitting any diagnostics.

## Stack trace

```
panic: Debug failure. False expression: generalized source shouldn't be assignable [recovered, repanicked]

github.com/microsoft/typescript-go/internal/checker.(*Relater).reportRelationError
    internal/checker/relater.go:4725 +0x185
github.com/microsoft/typescript-go/internal/checker.(*Relater).reportErrorResults
    internal/checker/relater.go:4706 +0x31a
github.com/microsoft/typescript-go/internal/checker.(*Relater).isRelatedToEx
    internal/checker/relater.go:2684 +0xa1f
... deep recursion through structuredTypeRelatedTo / recursiveTypeRelatedTo
    / unionOrIntersectionRelatedTo / eachTypeRelatedToType ...
github.com/microsoft/typescript-go/internal/checker.(*Checker).elaborateObjectLiteral
    internal/checker/relater.go:514
github.com/microsoft/typescript-go/internal/checker.(*Checker).elaborateError
    internal/checker/relater.go:463
github.com/microsoft/typescript-go/internal/checker.(*Checker).checkTypeRelatedToAndOptionallyElaborate
    internal/checker/relater.go:435
github.com/microsoft/typescript-go/internal/checker.(*Checker).isSignatureApplicable
    internal/checker/checker.go:9152
github.com/microsoft/typescript-go/internal/checker.(*Checker).reportCallResolutionErrors
    internal/checker/checker.go:9504
github.com/microsoft/typescript-go/internal/checker.(*Checker).resolveCall
    internal/checker/checker.go:8802
... resolveCallExpression / checkCallExpression / checkExpression
    / checkPropertyAccessExpression / checkExpression
    / resolveCallExpression / checkCallExpression / checkExpression
    / checkReturnStatement / checkBlock
    / checkFunctionExpressionOrObjectLiteralMethodDeferred ...
```

## Steps to reproduce

1. Download and extract `tsgo-panic-repro.zip` (attached) — 12 source files, ~800 lines, no proprietary code.
2. From the extracted directory, run:
   ```bash
   npm install
   npm test
   ```
   `tsgo` panics in ~6 s with no diagnostics emitted. Default-pinned to `@typescript/native-preview@7.0.0-dev.20260509.2`.
3. To bisect, switch tsgo versions via the included scripts:
   ```bash
   npm run use:repro-last-good     # 7.0.0-dev.20260324.1 — clean
   npm run use:repro-first-bad     # 7.0.0-dev.20260325.1 — first build that panics
   npm run use:bad                 # latest dev build
   npm test
   ```

## Versions

- **First-bad: `7.0.0-dev.20260325.1`** — same-day intra-build regression from `20260324.1` (clean).
- All builds since panic, including `7.0.0-dev.20260509.2` (current `latest`).
- Tested on Windows 11, Node v22 LTS.

## Trigger pattern

The function expression returned by `sendRequestWithPrefix` (in `requestWithPrefix.ts`) is the deferred-checked body whose elaboration panics:

```ts
return request<Unprefixed<TOutput>>({
    ...requestParams,
    data: unprefix<TInput>(requestParams.data, ...inputEscapeSpec),
}).then((data: Unprefixed<TOutput>) => {
    return prefix<TOutput>(data, ...outputEscapeSpec);
});
```

The panic surfaces when the relater enters error elaboration for the inner-call's object-literal argument while resolving the outer-call signature. Removing any one of the seven structural ingredients in the repo makes the panic disappear:

1. Global `Override<T, R> = Omit<T, keyof R> & R` declared in a non-module `.d.ts` (`helpers.d.ts`).
2. Recursive `Unprefixed<T>` mapped + template-literal-key remapping (`prefix.ts`).
3. The `unprefix` const assignment of `(any, any?) => any` to a generic signature with `PrefixApiReturnType<T, Unprefixed<T>>` (`prefixer.ts`).
4. `Override<X, { inlineObj }>` — applying the global `Override` to an inline object-literal type (`request.ts`).
5. `OverrideExistingFields<T, R> = Override<T, Pick<R, keyof T & keyof R>>` instantiated as `OverrideExistingFields<IRawQuestion, IQuestionOverriddenProperties<IUnionQuestionInfo>>` (`common.ts` + `question.ts`).
6. A wide `IPrefixedServerTypes` union — 150-member synthesised stand-in (`_big-union.ts`).
7. The `request<G>({...obj-literal}).then(...)` function expression itself (`requestWithPrefix.ts`).

## Note on a second related regression

The same `relater.go:4725` assertion fires later, on a different type-relation path, on the Microsoft Forms internal codebase. Bracketed at:

- **`7.0.0-dev.20260421.1`** — clean
- **`7.0.0-dev.20260421.2`** — panic (same assertion)

Likely the same root-cause area — the 2026-04-21 change appears to have widened the trigger surface so it also catches the slimmer type closure Forms's real code happens to exercise. Microsoft Forms team can share that source under NDA / internal channels if it helps narrow the fix.

ahejlsberg:
Fixes #3791.

### DIFF
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -4669,6 +4669,9 @@ func (r *Relater) indexSignaturesIdenticalTo(source *Type, target *Type) Ternary
 }
 
 func (r *Relater) reportErrorResults(originalSource *Type, originalTarget *Type, source *Type, target *Type, headMessage *diagnostics.Message) {
+	if r.overflow {
+		return
+	}
 	sourceHasBase := r.c.getSingleBaseForNonAugmentingSubtype(originalSource) != nil
 	targetHasBase := r.c.getSingleBaseForNonAugmentingSubtype(originalTarget) != nil
 	if originalSource.alias != nil || sourceHasBase {
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -3034,6 +3034,8 @@ func (r *Relater) eachTypeRelatedToSomeType(source *Type, target *Type) Ternary 
 // and issue an error. Otherwise, actually compare the structure of the two types.
 func (r *Relater) recursiveTypeRelatedTo(source *Type, target *Type, reportErrors bool, intersectionState IntersectionState, recursionFlags RecursionFlags) Ternary {
 	if r.overflow {
+		// Note that stack depth overflows can cause _any_ relation involving structured types to become false, so it is
+		// important do have well-defined behavior even in cases that shouldn't normally occur.
 		return TernaryFalse
 	}
 	id, constrained := getRelationKey(source, target, intersectionState, r.relation == r.c.identityRelation, false /*ignoreConstraints*/)
--- a/internal/checker/relater.go
+++ b/internal/checker/relater.go
@@ -10,7 +10,6 @@ import (
 	"github.com/microsoft/typescript-go/internal/binder"
 	"github.com/microsoft/typescript-go/internal/collections"
 	"github.com/microsoft/typescript-go/internal/core"
-	"github.com/microsoft/typescript-go/internal/debug"
 	"github.com/microsoft/typescript-go/internal/diagnostics"
 	"github.com/microsoft/typescript-go/internal/jsnum"
 	"github.com/microsoft/typescript-go/internal/tracing"
@@ -4669,9 +4668,6 @@ func (r *Relater) indexSignaturesIdenticalTo(source *Type, target *Type) Ternary
 }
 
 func (r *Relater) reportErrorResults(originalSource *Type, originalTarget *Type, source *Type, target *Type, headMessage *diagnostics.Message) {
-	if r.overflow {
-		return
-	}
 	sourceHasBase := r.c.getSingleBaseForNonAugmentingSubtype(originalSource) != nil
 	targetHasBase := r.c.getSingleBaseForNonAugmentingSubtype(originalTarget) != nil
 	if originalSource.alias != nil || sourceHasBase {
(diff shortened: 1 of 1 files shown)


ahejlsberg:
The assertion in `reportRelationError` that is causing the issue cannot be assumed to _always_ be true because excessive stack depth overflows can cause any arbitrary relation to become false (which is what is happening in the repro).

https://github.com/microsoft/typescript-go/blob/main/internal/checker/relater.go#L4725

I think the best cause of action is to remove the assertion. It doesn't really serve a purpose anyways.

ahejlsberg:
Copilot succeeded in creating a single-file repro for the issue in #3797, however it takes 2.5-3s to hit the assertion, it only does so in `strict: false` mode, and it doesn't cause the assertion to trigger with `tsc`. In my opinion, it isn't worth it to add the repro to the test suite. Instead, I've added a cautionary comment about making assumptions about type relations.

internal/checker/relater.go
```diff
@@ -3035,6 +3034,8 @@ func (r *Relater) eachTypeRelatedToSomeType(source *Type, target *Type) Ternary
 // and issue an error. Otherwise, actually compare the structure of the two types.
 func (r *Relater) recursiveTypeRelatedTo(source *Type, target *Type, reportErrors bool, intersectionState IntersectionState, recursionFlags RecursionFlags) Ternary {
 	if r.overflow {
+		// Note that stack depth overflows can cause _any_ relation involving structured types to become false, so it is
+		// important do have well-defined behavior even in cases that shouldn't normally occur.
```
DanielRosenwasser:
```suggestion
		// important to have well-defined behavior even in cases that shouldn't normally occur.
```

RyanCavanaugh (approved):

PR #3796, opened 2026-05-11
Hoist named class/function expressions assigned to module.exports in JS declaration emit

inq:
## Bug
* `module.exports = class Thing {...}` was falling to the `_default` fallback in `transformExportAssignment`, emitting `declare const _default: ...; export = _default;`.

## Fix
* Synthesize a top-level `export = Thing; declare class Thing {...}` by feeding a converted declaration through the existing `transformClassDeclaration` / `transformFunctionDeclaration` via `SetOriginal`. Reduces 4 `submoduleTriaged` baselines.


inq:

weswigham:
I think we'll be working on this over at #3850 instead, sorry!

inq:
Great! Thank you!

PR #3802, opened 2026-05-11
Fix intrinsic jsx tag name case in getTypeAtLocation

### LINKED ISSUE #3638: Cannot find name html element.
### Extension Version

0.20260428.1

### VS Code Version

1.117.0

### Operating system Version

MacOS 26.4.1

### Steps to reproduce

Honestly no idea how it triggers exactly, but we have a `.tsx` and some div closing tags seem to trigger a `Cannot find name 'div'. ts(2304)`

A restart of the LSP helped, but I assume it's still something you guys want to be aware of

### Issue

Expect Typescript to recognize the closing div tag as a valid element

gabritto:
Fixes #3638.
This was a bug in Strada as well.
We shouldn't be checking an identifier that is a tag name or an attribute name because that's not usually how we check those JSX nodes. I think we don't need this change for correctness if #3435 lands, but I think this may still be worth fixing.


jakebailey (approved):

DanielRosenwasser:
Doesn't something like this mean that in an example like the following:

```ts
foo.bar/**/
```

requesting quick info on `bar` before diagnostics are requested could report that `bar` is not defined?

gabritto:
> Doesn't something like this mean that in an example like the following:
> 
> ```ts
> foo.bar/**/
> ```
> 
> requesting quick info on `bar` before diagnostics are requested could report that `bar` is not defined?

`GetTypeOfSymbolAtLocation` already handles this case like this:

```go
if ast.IsIdentifier(location) || ast.IsPrivateIdentifier(location) {
			if ast.IsRightSideOfQualifiedNameOrPropertyAccess(location) {
				location = location.Parent
			}
			...
```

so in those property access-like cases, we already adjust location to be the parent.

PR #3804, opened 2026-05-11
add static property duplicate name check 

### LINKED ISSUE #3484: Static property conflicts with built-in Function properties (TS2699) no longer reported when useDefineForClassFields is false
conformance/staticPropertyNameConflicts(target=es2015,usedefineforclassfields=false).errors.txt.diff

iisaduan:
fixes #3484


internal/checker/checker.go
```diff
@@ -4347,6 +4338,25 @@ func (c *Checker) checkClassLikeDeclaration(node *ast.Node) {
...
 
+func (c *Checker) checkClassForStaticPropertyNameConflicts(node *ast.Node) {
+	for _, member := range node.Members() {
+		memberNameNode := member.Name()
+		isStaticMember := ast.IsStatic(member)
+		if isStaticMember && memberNameNode != nil {
+			memberName, _ := c.getEffectivePropertyNameForPropertyNameNode(memberNameNode)
+			switch memberName {
+			case "name", "length", "caller", "arguments":
+				if c.compilerOptions.GetUseDefineForClassFields() {
```
andrewbranch: This should result in skipping the whole loop.

andrewbranch (changes requested):

internal/checker/checker.go
iisaduan: Should it? This is the original code I ported from (the prototype case is handled elsewhere now) https://github.dev/microsoft/TypeScript/blob/0fb5e3a8cff7f18a30cdcee2ead37c727674ccc5/src/compiler/checker.ts#L42072

If we want this to skip the whole loop, i'll move this check to the beginning of the function

andrewbranch: The way I'm reading the code that's here now, if the compiler option is on, this function necessarily does nothing. Am I missing something? In the old code, it was the fallthrough to `prototype` that was effectual when useDefineForClassFields was off. With that moved, this now loops over every property and does nothing each time.

iisaduan: ooh I see

andrewbranch (approved):

PR #3821, opened 2026-05-12
Adding signature help tooltip coloring support for VS

navya9singh:
How Strada does it
 - tsserver tags text with `SymbolDisplayPartKind` values (TypeScript's own enum)
 - TypeScript-VS C# middleware translates these to Roslyn `ClassificationTypeNames` via `GetClassificationName()`
 - Middleware builds a `ClassifiedTextElement` and attaches it as `ColorizedLabel` on `SignatureInformation`
 - VS uses the Roslyn classification names to render colored text
 
 Why Corsa needs a different approach
 - In Corsa, VS talks directly to tsgo over LSP — no C# middleware exists in this path
 - Without middleware, there's nothing to translate display part kinds or build the `ClassifiedTextElement`
 - Result: signature help renders as plain uncolored text
 
New `displayPartsWriter`
 - The Go printer's internal writer discards classification info — `WriteKeyword`, `WriteSymbol`, etc. all just append plain text
 - Modifying the internal writer would add overhead to every print operation across the codebase
 - `displayPartsWriter` implements `EmitTextWriter` as an external writer passed to `printer.Write()`, intercepting classified write calls and recording Roslyn classification names alongside text
 - `classificationForSymbol()` collapses Strada's 2-step translation (`SymbolFlags` → `SymbolDisplayPartKind` → Roslyn name) into one step, verified against both `displayPartKind()` and `GetClassificationName()`
 
Changes
- Modified: `internal/ls/signaturehelp.go` - builds `_vs_colorizedLabel` with classified runs per signature
- Modified: `internal/lsp/lsproto/_generate/generate.mts`— added `ClassifiedTextRun`, `ClassifiedTextElement` types; patched `SignatureInformation`
-  23 baseline files updated with new `_vs_colorizedLabel` data
 
 - `displayPartsWriter` is reusable for hover, FAR, completions — any feature needing VS colorized output
 
 Fixes [Bug 2800005](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/2800005)


internal/ls/signaturehelp.go
```diff
@@ -360,6 +360,20 @@ func (l *LanguageService) createSignatureHelpItems(ctx context.Context, candidat
...
+				Runs: item.ColorizedRuns,
+			}
+		}
+
+		// Set VS-specific colorized label if we have classified runs
+		if len(item.ColorizedRuns) > 0 {
+			sigInfo.VSColorizedLabel = &lsproto.ClassifiedTextElement{
+				Runs: item.ColorizedRuns,
+			}
+		}
```
jakebailey: This looks doubled

internal/ls/signaturehelp.go
```diff
@@ -433,36 +447,75 @@ func (l *LanguageService) getSignatureHelpItem(candidate *checker.Signature, isT
...
 		var display strings.Builder
 		display.WriteString(callTargetSymbol)
 		display.WriteString(info.displayParts)
 		display.WriteString(suffixDisplayParts)
+
+		// Combine classified runs: callTarget + info runs + suffix runs
+		var allRuns []*lsproto.ClassifiedTextRun
+		allRuns = append(allRuns, callTargetRuns...)
+		allRuns = append(allRuns, info.colorizedRuns...)
+		allRuns = append(allRuns, suffixRuns...)
```
jakebailey: Could just be `slices.Concat`?

internal/ls/signaturehelp.go
```diff
@@ -433,36 +447,75 @@ func (l *LanguageService) getSignatureHelpItem(candidate *checker.Signature, isT
...
-	var returnType strings.Builder
-	returnType.WriteString(": ")
+func returnTypeToDisplayParts(candidateSignature *checker.Signature, c *checker.Checker, enclosingDeclaration *ast.Node, sourceFile *ast.SourceFile) (string, []*lsproto.ClassifiedTextRun) {
+	var runs []*lsproto.ClassifiedTextRun
+	var result strings.Builder
+
+	// Add ": " prefix
+	result.WriteString(": ")
+	runs = append(runs, &lsproto.ClassifiedTextRun{ClassificationTypeName: string(lsproto.ClassificationTypeNamePunctuation), Text: ":"})
+	runs = append(runs, &lsproto.ClassifiedTextRun{ClassificationTypeName: string(lsproto.ClassificationTypeNameWhiteSpace), Text: " "})
```
jakebailey: Given this is only enabled on VS, is there some way we could avoid doing all of this if the client didn't set the cap that says they want this info?

jakebailey: In fact, none of this should be getting executed or placed on the output object unless the VS capability is enabled

internal/ls/signaturehelp.go
```diff
@@ -759,7 +873,7 @@ func containsPrecedingToken(startingToken *ast.Node, sourceFile *ast.SourceFile,
 	pos := startingToken.Pos()
 	// There's a possibility that `startingToken.parent` contains only `startingToken` and
 	// missing nodes, none of which are valid to be returned by `findPrecedingToken`. In that
-	// case, the preceding token we want is actually higher up the tree—almost definitely the
+	// case, the preceding token we want is actually higher up the treeâ€”almost definitely the
```
jakebailey: Weird diff here

internal/ls/signaturehelp.go
jakebailey: Maybe a local `addRun` func that is a noop if not in VS

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -28,7 +28,6 @@ TestAutoImportSortCaseSensitivity1
 TestAutoImportTypeImport4
 TestAutoImportTypeOnlyPreferred3
 TestAutoImportVerbatimTypeOnly1
-TestCalledUnionsOfDissimilarTyeshaveGoodDisplay
```
jakebailey, line 36: What fixes these?

internal/ls/displaypartswriter.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/microsoft/typescript-go/internal/stringutil"
+)
+
+var _ printer.EmitTextWriter = &displayPartsWriter{}
+
+// displayPartsWriter implements EmitTextWriter and captures classified text runs
+// for VS colorized labels, while also building a plain string.
+type displayPartsWriter struct {
+	builder     strings.Builder
+	runs        []*lsproto.ClassifiedTextRun
```
jakebailey: I think I'm confused why we have this, but then also, we have sig help manually building runs?

jakebailey (changes requested):

internal/ls/signaturehelp.go
DanielRosenwasser: Yeah, what this would look like instead would be

```go
addRun(":", lsproto.ClassificationTypeNamePunctuation)
addRun(" ", lsproto.ClassificationTypeNameWhiteSpace)
```

and `addRun` could be

```
result.WriteString(str)
if vsCapability {
	runs = append(runs, /*...*/)
}
```

And if VS never cares about the resulting string, you can just keep it empty to avoid doing unnecessary work:

```
if vsCapability {
	runs = append(runs, /*...*/)
} else {
	result.WriteString(str)
}
```

jakebailey: I think the result is still needed; in our tests we would want to say "test with VS output" and not lose the rest, for example

internal/lsp/lsproto/_generate/generate.mts
```diff
@@ -464,6 +464,45 @@ const customStructures: Structure[] = [
...
+        documentation: "A classified text run with text and classification type, used for colorized display in VS.",
+    },
+    {
+        name: "ClassifiedTextElement",
+        properties: [
+            {
+                name: "Runs",
+                type: { kind: "array", element: { kind: "reference", name: "ClassifiedTextRun" } },
+                documentation: "The classified text runs that make up this element.",
+            },
```
jakebailey, line 502: Are all of these props actually uppercased in the client? Very surprising

testdata/baselines/reference/fourslash/signatureHelp/jsDocDontBreakWithNamespaces.baseline
```diff
@@ -42,7 +42,35 @@
...
         {
           "label": "foo(): module",
-          "parameters": []
+          "parameters": [],
+          "_vs_colorizedLabel": {
+            "Runs": [
+              {
+                "ClassificationTypeName": "method name",
+                "Text": "foo"
+              },
```
navya9singh: fixing this

jakebailey: Do we just need to enable the VS cap in fourslash?

internal/lsp/lsproto/_generate/generate.mts
navya9singh: It is, yes. They're all defined in the `Microsoft.VisualStudio.Text.Adornments` library

jakebailey: Yuck!

internal/fourslash/_scripts/failingTests.txt
navya9singh: This is because previously we were passing in nil for emitContext, but now we are passing in a new emitContext shared between the printer and the NodeBuilder.

internal/ls/displaypartswriter.go
navya9singh: displayPartsWriter is now doing runs and building strings

testdata/baselines/reference/fourslash/signatureHelp/jsDocDontBreakWithNamespaces.baseline
navya9singh: Yes, if we enable the capability in fourslash then these responses will include the `_vs_colorizedLabel`.  Should I do that?

jakebailey: Yes, I assume we want to make sure it's working, right?

jakebailey:
I think we should try to test this, but otherwise LGTM

testdata/baselines/reference/fourslash/signatureHelp/jsDocDontBreakWithNamespaces.baseline
navya9singh: Sure, I'll add it. I've tried it with VS and it works there and fixes the coloring issue.

jakebailey (approved):

internal/fourslash/_scripts/failingTests.txt
```diff
@@ -33,7 +33,6 @@ TestAutoImportSortCaseSensitivity1
...
```
jakebailey, line 36: where did TestQualifyModuleTypeNames go?

navya9singh: I'm running these tests again to check

jakebailey (approved):

jakebailey (approved):

PR #3823, opened 2026-05-12
Don't recursively scan auto-import packages for non-main .d.ts files unless option is set

andrewbranch:
It's a little more complicated than I could fit in the title.

When a package.json has `exports`, we know exactly what its publicly accessible entrypoint files are. This PR doesn't change anything about how we process those. Without `exports`, every file is publicly accessible, but in practice, not everything is _meant_ to be a publicly accessible entrypoint. Before this PR, we recursively scan the whole package for `.d.ts` files and offer auto-imports for all of them. That collects a lot of garbage entrypoints, and is expensive. It's especially expensive in monorepo workspaces, where the linked packages contain whole `.ts` files, not just `.d.ts` files.

This PR changes our default behavior to _not_ recursively scan packages that lack `exports` for files to auto-import from unless:

- it is a `@types` package,
- it is one of about 50 packages from the ~15k [npm-high-impact](https://github.com/wooorm/npm-high-impact) packages that were identified as not having exports, but having documented or commonly used non-main entrypoints, via a combination of heuristic scripts and Copilot analysis, or
- a loaded program already accesses a deep import from the package.

Users can opt out of this new default by setting

```json
"js/ts.preferences.autoImportEntrypointDirectorySearch": true
```

in `.vscode/settings.json`. (Option name bikeshedding accepted.)

This drops microsoft/vscode's auto-import registry build time from about 170ms to 70ms, and elastics/kibana from about 32s to 1.1s (😅).

Also removes `includePackageJsonAutoImports`, which has never been supported in Corsa.


internal/ls/autoimport/registry.go
```diff
@@ -28,6 +28,61 @@ import (
 	"github.com/microsoft/typescript-go/internal/vfs/vfsmatch"
 )
 
+var knownRecursiveSearchPackages = map[string]struct{}{
```
jakebailey: Optional, but we have a `collections.Set` constructor for this, and then you wouldn't need the helper func and could just use `Has`

jakebailey (commented):
LGTM with one suggestion

internal/ls/autoimport/registry.go
andrewbranch: For some reason I was thinking there was a problem with doing that at the top level, but that was silly

jakebailey (approved):

PR #3824, opened 2026-05-12
Collect telemetry on recovered notification panics and stderr on unrecovered panics

andrewbranch:
Adds crash telemetry in two cases we were missing:

1. Recovered panics were only sent as LSP telemetry notifications during requests, not during notifications
2. Unrecovered panics were tracked with `languageServer.connectionClosed`, but lacked any error message. We now capture the last 40 lines of STDERR from the output channel and include that in the event.

### DIFF
--- a/_extension/src/client.ts
+++ b/_extension/src/client.ts
@@ -377,9 +377,9 @@ class ReportingErrorHandler implements ErrorHandler {
     }
 
     private consumeStderrBuffer(): string {
-        const result = this.stderrBuffer.join("\n").slice(0, ReportingErrorHandler.maxStderrBytes);
+        const raw = this.stderrBuffer.join("\n");
         this.stderrBuffer = [];
-        return result;
+        return sanitizeStderr(raw).slice(0, ReportingErrorHandler.maxStderrBytes);
     }
 
     error(_error: Error, _message: Message | undefined, count: number | undefined): ErrorHandlerResult | Promise<ErrorHandlerResult> {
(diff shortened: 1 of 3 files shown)


internal/lsp/server.go
```diff
@@ -961,25 +961,22 @@ func (s *Server) recover(req *lsproto.RequestMessage) {
...
-					RequestFailureTelemetryEvent: &lsproto.RequestFailureTelemetryEvent{
-						Properties: &lsproto.RequestFailureTelemetryProperties{
-							ErrorCode:     lsproto.ErrorCodeInternalError.String(),
-							RequestMethod: strings.ReplaceAll(string(req.Method), "/", "."),
-							Stack:         sanitizeStackTrace(string(stack)),
-						},
-					},
-				})
-			}
+			_ = s.sendError(req.ID, fmt.Errorf("%w: panic handling request %s: %v", lsproto.ErrorCodeInternalError, req.Method, r))
```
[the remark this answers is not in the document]
andrewbranch: guess what s.logger.Error does...

_extension/src/client.ts
```diff
@@ -178,6 +178,16 @@ export class Client implements vscode.Disposable {
...
 
+        // Monkey-patch the output channel's error method to capture recent stderr lines.
+        // When the server crashes, vscode-languageclient pipes stderr to outputChannel.error(),
+        // so the error handler can include the last N lines in crash telemetry.
+        const errorHandler = this.clientOptions.errorHandler as ReportingErrorHandler;
+        const originalError = this.outputChannel.error.bind(this.outputChannel);
+        this.outputChannel.error = (...args: Parameters<typeof this.outputChannel.error>) => {
+            originalError(...args);
+            errorHandler.pushStderrLine(String(args[0]));
+        };
```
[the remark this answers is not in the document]
andrewbranch: I don't think there's a very easy way to do that. I think it's fine.

jakebailey:
I'm not sure we can send the stderr raw; is there specific text we think we are going to find that we can look for?

andrewbranch:
That is not what @DanielRosenwasser told me in chat, unless I misunderstood

DanielRosenwasser:
If we can pretty confidently validate that we have a stack trace, we can do some client-side sanitization like we used to. We have some stuff on the endpoint to get rid of PII just in case.

andrewbranch:
Can I just collect everything between

image


and

image

?

What kind of client-side sanitization would you like me to do?

jakebailey:
I don't know what kind of sanitization would be safe; for stack traces we know the exact format and can parse it to remove info. I don't know off the top of my head what's possible to go over stderr. We fall back to stderr when the client isn't available for logging yet, and I don't know what happens if say, the process OOMs and logs some info, or otherwise hits a big crash?

andrewbranch:
https://github.com/microsoft/typescript-go/pull/3824/commits/7f439aeb1ab9ca1b8bb0894b6ba1764a14d79412 made it so we only process lines between `panic:` and `Server exited` and ports the server-side stack sanitization.

jakebailey:
right, but I'm more thinking about if there's a panic followed by more text that some other goroutine is logging, interleaved. Maybe it just works?

andrewbranch:
The sanitizer redacts lines that don't match a recognized pattern.

_extension/src/client.ts
```diff
@@ -419,3 +451,63 @@ class ReportingErrorHandler implements ErrorHandler {
...
+        return leadingWhitespace + relevantPart;
+    }
+
+    // Non-internal frames get fully redacted.
+    const leadingWhitespace = line.match(/^(\s*)/)?.[1] ?? "";
+    // Preserve completely blank lines.
+    if (line.trim() === "") {
+        return "";
+    }
+    return leadingWhitespace + "(REDACTED)";
```
DanielRosenwasser, line 522: Minor, but `leadingWhitespace` can be defined up-front.

```suggestion
    const leadingWhitespace = line.match(/^(\s*)/)?.[1] ?? "";
    const idx = line.indexOf(ourModuleMarker);
    if (idx >= 0) {
        let relevantPart = line.slice(idx);
        // Strip hex offset suffixes like " +0x40"
        relevantPart = relevantPart.replace(/ \+0x[0-9a-fA-F]+$/, "");
        // Strip " in goroutine N" suffixes
        relevantPart = relevantPart.replace(/ in goroutine \d+$/, "");
        // Strip function arguments (keep parens empty)
        relevantPart = relevantPart.replace(/\([^)]*\)$/, "()");
        // Replace / with |> to defeat path-based secret detection
        relevantPart = relevantPart.replace(/\//g, "|>");
        // Defeat generic secret keyword regex
        relevantPart = relevantPart.replace(genericSecretKeywordRegex, "$1X_X$2");
        return leadingWhitespace + relevantPart;
    }

    // Preserve completely blank lines.
    if (line.trim() === "") {
        return "";
    }

    // Non-internal frames get fully redacted.
    return leadingWhitespace + "(REDACTED)";
```

_extension/src/client.ts
```diff
@@ -345,13 +357,31 @@ class ReportingErrorHandler implements ErrorHandler {
     telemetryReporter: tr.TelemetryReporter;
     maxRestartCount: number;
     restarts: number[];
+    private stderrBuffer: string[] = [];
+    private static readonly maxStderrLines = 40;
+    private static readonly maxStderrBytes = 8192;
```
DanielRosenwasser: Nit: character codes, not code points. 😅

Maybe just `maxStderrLength`

_extension/src/client.ts
```diff
@@ -345,13 +357,31 @@ class ReportingErrorHandler implements ErrorHandler {
...
         this.maxRestartCount = maxRestartCount;
         this.restarts = [];
     }
 
+    pushStderrLine(line: string): void {
+        for (const l of line.split("\n")) {
+            this.stderrBuffer.push(l);
+            if (this.stderrBuffer.length > ReportingErrorHandler.maxStderrLines) {
+                this.stderrBuffer.shift();
+            }
```
DanielRosenwasser: I feel like it tends to be better to capture the top of the stack trace than the bottom. Given that, it's better to just keep appending until we hit the limit.


```suggestion
            if (this.stderrBuffer.length < ReportingErrorHandler.maxStderrLines) {
                this.stderrBuffer.push(l);
            }
```

DanielRosenwasser (changes requested):

_extension/src/client.ts
andrewbranch: The original idea here was that if something else unrelated to the crash has been spamming stderr (I don't know whether this is plausible), we don't want to capture that; we want to prioritize whatever happened right before the connection close. I'll shift to not append at all before we see a `panic:` line.

_extension/src/client.ts
```diff
@@ -419,3 +460,58 @@ class ReportingErrorHandler implements ErrorHandler {
...
+
+function sanitizeStderrLine(line: string): string {
+    // Keep "goroutine N [status]:" headers as-is.
+    if (/^goroutine \d+/.test(line)) {
+        return line;
+    }
+    // Keep panic message line, but redact any file path after "panic:"
+    if (/^panic:/.test(line.trimStart())) {
+        return line.trimStart();
+    }
```
jakebailey, line 489: this will include the asserts, right? I think we explicitly hide the messages (this is new code compared to the Go side)

_extension/src/client.ts
```diff
@@ -345,13 +357,40 @@ class ReportingErrorHandler implements ErrorHandler {
...
         this.telemetryReporter = telemetryReporter;
         this.maxRestartCount = maxRestartCount;
         this.restarts = [];
     }
 
+    pushStderrLine(line: string): void {
+        for (const l of line.split("\n")) {
+            if (!this.capturingPanic) {
+                if (/^panic:/.test(l.trimStart())) {
+                    this.capturingPanic = true;
```
jakebailey, line 377: Does this ever go false? I think it's only ever true so could over-capture?

jakebailey (approved):

DanielRosenwasser (approved):

PR #3826, opened 2026-05-13
Add version & a common session identifier to events from LS clients

DanielRosenwasser:
This allows us to do 2 things long-term:

1. Avoid correlating on `languageServer.start` events to figure out which version of TypeScript an error response occurred within. This is technically error-prone right now.
2. Correlate events like OOMs with other diagnostic information.


_extension/src/client.ts
```diff
@@ -64,77 +65,18 @@ export class Client implements vscode.Disposable {
...
-                },
-            },
-        };
     }
 
     async start(exe: { path: string; version: string; }): Promise<void> {
         this.exe = exe;
+        this.reporterCommonProperties = {
+            "tscommon.version": exe.version,
+            "tscommon.serverSessionId": `${Date.now()}`,
```
DanielRosenwasser, line 76: We could switch this to `Crypto.randomUUID` - this is usually "good enough" since there are also VS Code session IDs.

_extension/src/client.ts
```diff
@@ -64,77 +65,18 @@ export class Client implements vscode.Disposable {
             ...jsTsLanguageModes.map(language => ({ scheme: "file", language })),
             ...jsTsLanguageModes.map(language => ({ scheme: "untitled", language })),
         ];
-        this.clientOptions = {
```
DanielRosenwasser, line 79: This is just moved down into a function that generates these options on the fly. I figured we're probably not saving a ton by trying to reuse the same object when the client is restarted.

_extension/src/client.ts
```diff
@@ -174,7 +116,7 @@ export class Client implements vscode.Disposable {
             "typescript.native-preview",
             "typescript.native-preview-lsp",
             serverOptions,
-            this.clientOptions,
+            this.makeClientOptions(this.reporterCommonProperties),
         );
```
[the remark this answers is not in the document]
DanielRosenwasser: It is narrowed from the assignment above.

_extension/src/client.ts
```diff
@@ -64,77 +65,18 @@ export class Client implements vscode.Disposable {
...
-            },
-        };
     }
 
     async start(exe: { path: string; version: string; }): Promise<void> {
         this.exe = exe;
+        this.reporterCommonProperties = {
+            "tscommon.version": exe.version,
+            "tscommon.serverSessionId": `${Date.now()}`,
+        };
```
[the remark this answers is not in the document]
DanielRosenwasser: It's true but probably rare. I could add a counter or use a UUID.

_extension/src/telemetryReporting.ts
```diff
@@ -60,30 +60,38 @@ export function createTelemetryReporter(vscReporter: VSCodeTelemetryReporter): E
...
-    function sendTelemetryEvent(eventName: string, data?: Record<string, string>, measurements?: Record<string, number>): void {
+    function sendTelemetryEvent(eventName: string, data?: {}, measurements?: {}): void {
         data = { ...sharedProperties, ...data };
         vscReporter.sendTelemetryEvent(eventName, data, measurements);
     }
 
-    function sendTelemetryErrorEvent(eventName: string, data?: Record<string, string>, measurements?: Record<string, number>): void {
+    function sendTelemetryErrorEvent(eventName: string, data?: {}, measurements?: {}): void {
         data = { ...sharedProperties, ...data };
         vscReporter.sendTelemetryErrorEvent(eventName, data, measurements);
```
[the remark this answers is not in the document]
DanielRosenwasser: I can't do this because I switched to interface types which don't have implicit index signatures. If I added explicit ones, it would permit unknown properties which is arguably worse.

The whole point of what I've done here is to provide a typed facade anyway so I think it's unfortunate but fine.

_extension/src/telemetryReporting.ts
```diff
@@ -60,34 +60,42 @@ export function createTelemetryReporter(vscReporter: VSCodeTelemetryReporter): E
...
     }
 
-    function sendTelemetryEvent(eventName: string, data?: Record<string, string>, measurements?: Record<string, number>): void {
+    function sendTelemetryEvent(eventName: string, data?: {}, measurements?: {}): void {
         data = { ...sharedProperties, ...data };
         vscReporter.sendTelemetryEvent(eventName, data, measurements);
     }
 
-    function sendTelemetryErrorEvent(eventName: string, data?: Record<string, string>, measurements?: Record<string, number>): void {
+    function sendTelemetryErrorEvent(eventName: string, data?: {}, measurements?: {}): void {
```
gabritto, line 68: Is this meant to prevent us adding properties in the implementation of `sendTelemetryErrorEvent`?

gabritto (approved):

jakebailey:
I don't think dotted names are legal names for end users of this?

_extension/src/telemetryReporting.ts
DanielRosenwasser: Kind of - the declared types of `sendTelemetry[Error]Event` are declared with explicit properties and no index signature so that callers get excess property errors.

The problem is that I switched all the types to interfaces (because I wanted to use interface `extends` and no worry about intersection types). Previously, they were all type aliases to object type literals, and the subtle behavior around those is that they are assignable to any type with an index signature if all of their properties are compatible as well.

So switching to interfaces makes the implementation of `sendTelemetryEvent` incompatible with those signatures because `interface`-declared types don't have that same "implicit index signature" behavior. So instead of the implementation saying it "needed" an index signature, I just switched to `{}`. It is good enough given that the `TelemetryReporter` type provides the actual type safety we care about, and `{}` *does* have implicit index signature behavior in the body of `sendTelemetryEvent`.

_extension/src/client.ts
```diff
@@ -342,14 +360,18 @@ export class Client implements vscode.Disposable {
...
 
-    constructor(telemetryReporter: tr.TelemetryReporter, maxRestartCount: number) {
+    private reporterCommonProperties: tr.LSCommonProperties;
+
+    constructor(telemetryReporter: tr.TelemetryReporter, maxRestartCount: number, reporterCommonProperties: tr.LSCommonProperties) {
         this.telemetryReporter = telemetryReporter;
         this.maxRestartCount = maxRestartCount;
         this.restarts = [];
+
+        this.reporterCommonProperties = reporterCommonProperties;
```
DanielRosenwasser, line 402: I actually wonder if this needs to be reset based on `closed`...

DanielRosenwasser:
> I don't think dotted names are legal names for end users of this?

I'm not seeing anywhere that implies that. I may be wrong there.

_extension/src/client.ts
jakebailey: If we have a VS Code session ID and just need to make the starts unique, then why not just `let serverCount = 0;` and then use `serverCount++`?

_extension/src/client.ts
```diff
@@ -213,14 +155,23 @@ export class Client implements vscode.Disposable {
         const serverTelemetryListener = this.client.onTelemetry((d: TelemetryData) => {
             switch (d.telemetryPurpose) {
                 case "usage":
-                    this.telemetryReporter.sendTelemetryEventUntyped(d.eventName, d.properties, d.measurements);
+                    this.telemetryReporter.sendTelemetryEventUntyped(
```
jakebailey, line 179: How hard is it to pass in a wrapper that auto adds these to callers?

DanielRosenwasser: The types will get pretty complex to keep the safe calls, but I can do it.

DanielRosenwasser: Or I just drop the common properties from each of them. I think that's reasonable too.

_extension/src/client.ts
DanielRosenwasser: In the rareish situation that the extension was disabled/re-enabled in the same session, you would end up with duplicates.

The same can happen with a `Date.now()` though if the user modifies their local time.

I'd argue the former is more common.

jakebailey: Disabling our extension requires an extension host reload, no?

DanielRosenwasser: Does that have a separate session ID associated with it?

jakebailey: Yes, every window load is a new session, if that's what you mean

_extension/src/client.ts
```diff
@@ -61,92 +62,42 @@ export class Client implements vscode.Disposable {
...
+        this.outputChannel.error = (...args: Parameters<typeof this.outputChannel.error>) => {
+            originalError(...args);
+            this.errorHandler!.pushStderrLine(String(args[0]));
+        };
+        // Remember to dispose here so that we don't keep stacking references to the original Client instance.
+         this.disposables.push({
+            dispose: () => {
+                this.outputChannel.error = unboundOriginalError;
+            }
+        });
```
DanielRosenwasser, line 95: This is new. Previously, whenever we would construct a client, we would re-wrap the error output handler and these could stack.

DanielRosenwasser:
I'm not merging this one in in time for the repo move, but we should do it.

PR #3828, opened 2026-05-13
fix(3810): handle deprecated tags in completions

### LINKED ISSUE #3810: Deprecated members still shown in dropdown menu with `editor.suggest.showDeprecated` set to `false`
### Extension Version

0.20260512.1

### VS Code Version

1.119

### Operating system Version

win11

#### config

```
"editor.suggest.showDeprecated": false
```

#### code

```ts
const o = {
    /** @deprecated */
    a: 1,
    b: 2,
    c: 3,
}
o. // stop here to see the members shown in the dropdown menu
```

#### ts6.0

Image

#### tsgo

Image

### Issue

Expected no `a` in the dropdown menu.

a-tarasyuk:
Fixes #3810


DanielRosenwasser:
Is there some non-converted test that this should fix?

internal/fourslash/fourslash.go
```diff
@@ -1368,8 +1368,8 @@ func ignorePaths(paths ...string) cmp.Option {
 }
 
 var (
-	completionIgnoreOpts  = ignorePaths(".Kind", ".SortText", ".FilterText", ".Data", ".AdditionalTextEdits")
```
gabritto, line 1371: Could we not ignore `.Tags` and instead just update the tests?

gabritto (changes requested):

jakebailey (approved):

internal/fourslash/fourslash.go
```diff
@@ -1429,6 +1429,11 @@ func (f *FourslashTest) verifyCompletionItem(t *testing.T, prefix string, actual
 			return fmt.Sprintf("%s:\n%s", "Kind mismatch", err)
 		}
 	}
+	if expected.Tags != nil {
```
gabritto: Do we still need this if we're no longer ignoring `.Tags` in the completion options above?

gabritto (approved):

PR #3831, opened 2026-05-13
fix(3829): handle extends tag not attached to a class

### LINKED ISSUE #3829: Behavior difference: tsgo fails to report TS8022 as tsc does
## Steps to reproduce

1. Start a plain new project
    - "@typescript/native-preview": "7.0.0-dev.20260513.1",
    - "typescript": "6.0.3"
    - "allowJs": true,
    - "checkJs": true,
    - "stableTypeOrdering": true,

```js
// src/main.js
class B {}

/** @extends {B} */
export var C1 = function () {}

/** @extends {B} */
export function C2() {}
```

```
src/main.js:5:17 - error TS8022: JSDoc '@extends' is not attached to a class.

5 export var C1 = function () {}
                  ~~~~~~~~

src/main.js:8:17 - error TS8022: JSDoc '@extends' is not attached to a class.

8 export function C2() {}
                  ~~


Found 2 errors in the same file, starting at: src/main.js:5
```

## Behavior with `tsgo`

No error at all.

a-tarasyuk:
Fixes #3829


ahejlsberg:
I really feels like the right place to issue these errors is the reparser since a lot of the logic to identify the JSDoc host is the same. There are likely other similar checks that could happen in the reparser.

